PR #364, opened 2021-11-03
wip: proof of concept for request / response hooks

jacob-ebey:
Adds `beforeRequest` and `beforeResponse` to createRequestHandler to allow a place to implement things like request rewrites, SWR caches, request / response logging, etc...

Questions:
- How composable should this be?
- Is the throw pattern in beforeRequest a good idea?

Current idea:

```tsx
let { createRequestHandler } = require("@remix-run/express");
let build = require("./build");

let swrCache = require("./swr-cache");

async function beforeRequest({ request }) {
  try {
    let cachedResponse = swrCache.get(request.clone());
    if (cachedResponse) {
      throw cachedResponse;
    }
  } catch (err) {
    console.error(err);
  }
  return request;
}

async function beforeResponse({ request, response }) {
  try {
    return await swrCache.write(request, response.clone());
  } catch (err) {
    console.error(err);
    return response;
  }
}

let handler = createRequestHandler({ build, beforeRequest, beforeResponse });

// etc.....
```

### DIFF
--- a/packages/remix-architect/server.ts
+++ b/packages/remix-architect/server.ts
@@ -12,7 +12,8 @@ import type {
 import type {
   AppLoadContext,
   ServerBuild,
-  ServerPlatform
+  ServerPlatform,
+  CreateRequestHandlerOptions
 } from "@remix-run/server-runtime";
 import { createRequestHandler as createRemixRequestHandler } from "@remix-run/server-runtime";
 import type { Response as NodeResponse } from "@remix-run/node";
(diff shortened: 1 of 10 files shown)


romansndlr:
> Can we push this back until we actually have a cache implementation we can offer people? Until then it seems like we're just speculating about how people will actually use this and we have thus far been very deliberate about avoiding middleware.

Can you maybe elaborate on this? Why not middleware?

chaance:
@jacob-ebey I'm trying to clean up GitHub pretty hardcore today, gonna close this for now until we come back with something actionable. Let me know if you think we should reconsider!

PR #383, opened 2021-11-16
feat: add support for multipart / file uploads

jacob-ebey:
feat: added global Blob and File polyfill
feat: utlize abort controller to close connection early
chore: added onFile handler to node Reqeust.formData
feat: add file and memory upload storage handlers
feat: enable multipart form data in node package
chore: import jest globals from @remix-run/node
fix: fix server entry not catching due to missing await


docs/api/remix.md
```diff
@@ -1295,6 +1295,211 @@ return new Response("", {
...
+
+export let action: ActionFunction = async ({ request }) => {
+  let formData = await parseMultipartFormData({
+    request,
+    uploadHandler
+  });
+
+  let file = formData.get("avatar");
+
+  // file is a "NodeFile" which has a similar API to "File"
```
MichaelDeBoey, line 1373: Why is `file` of type `NodeFile` here if we're saying it's of type `File` with a memoryHandler and it should be `File`, `string` or `undefined` when creating a custom handler?

> Your job is to do whatever you need with the `stream` and return a value that's a valid [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) value: [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File), `string`, or `undefined`.

jacob-ebey: NodeFile is an implementation of the File interface that keeps things out of memory and on disk until you need it. It is infact a valid FormData value :)

packages/remix-node/form-data.ts
```diff
@@ -1,32 +1,144 @@
...
+      abortController?.abort();
+      reject(error || new Error("failed to parse form data"));
+    }
+
+    busboy.on("field", (name, value) => {
+      fields[name] = fields[name] || [];
+      fields[name].push(value);
+    });
+
+    if (uploadHandler) {
```
mjackson, line 52: We need to warn them when they don't have an upload handler and the request body contains files.

packages/remix-node/fetch.ts
```diff
@@ -1,27 +1,58 @@
...
+    this.abortController =
+      anyInput?.abortController || anyInit?.abortController;
   }
 
-  async formData() {
-    let body = await this.clone().text();
-    return new RemixFormData(body);
+  async formData({
+    uploadHandler
+  }: RemixFormDataOptions = {}): Promise<FormData> {
```
mjackson, line 35: Change to `request.formData(uploadHandler?)`

mjackson (changes requested):

fixtures/gists-app/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "@exampledev/new.css": "^1.1.3",
     "@mdx-js/react": "^1.6.22",
+    "abort-controller": "^3.0.0",
```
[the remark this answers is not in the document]
jacob-ebey: Real applications will not, but our fixtures does because of how the repo is setup.

packages/remix-node/form-data.ts
```diff
@@ -1,32 +1,144 @@
...
+}
+
 export class RemixFormData implements FormData {
-  private _params: URLSearchParams;
+  private _fields: Record<string, (File | string)[]>;
 
-  constructor(body?: string) {
-    this._params = new URLSearchParams(body);
+  constructor(fields: Record<string, (File | string)[]> = {}) {
+    this._fields = fields;
```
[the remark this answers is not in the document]
jacob-ebey: turns out FormData lib does not even conform to the `set` method in the interface.... Looked into it but sticking with a small custom one for now.

mjackson (changes requested):

packages/remix-node/upload/memoryUploadHandler.ts
```diff
@@ -0,0 +1,83 @@
...
+  filter,
+  maxFileSize = 3000000
+}: MemoryUploadHandlerOptions): UploadHandler {
+  return async ({ name, stream, filename, encoding, mimetype }) => {
+    if (filter && !(await filter({ filename, encoding, mimetype }))) {
+      stream.resume();
+      return;
+    }
+
+    let bufferStream = new BufferStream();
```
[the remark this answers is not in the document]
jacob-ebey: Right below ;)

packages/remix-node/upload/meter.ts
```diff
@@ -23,6 +23,6 @@ export class Meter extends Transform {
 
 export class MeterError extends Error {
   constructor(public field: string, public maxBytes: number) {
-    super(`Field ${field} exceeded upload size of ${maxBytes} bytes.`);
+    super(`Field "${field}"" exceeded upload size of ${maxBytes} bytes.`);
```
mjackson, line 26: There are two closing "

mjackson (changes requested):

packages/remix-node/upload/meter.ts
jacob-ebey: Oops

packages/remix-node/upload/fileUploadHandler.ts
```diff
@@ -88,22 +92,32 @@ export function createFileUploadHandler({
       return;
     }
 
-    let filedir = resolvePath(
+    let dir =
       typeof directory === "string"
         ? directory
-        : directory({ filename, encoding, mimetype })
-    );
+        : directory({ filename, encoding, mimetype });
```
mjackson, line 98: The `directory` option should just be a static string. No need for a function here.

packages/remix-node/upload/fileUploadHandler.ts
```diff
@@ -88,22 +92,32 @@ export function createFileUploadHandler({
...
+    }
+
+    let filepath = resolvePath(filedir, path);
 
     if (avoidFileConflicts) {
       filepath = await uniqueFile(filepath);
     }
 
-    await mkdir(dirname(filepath), { recursive: true });
+    await mkdir(dirname(filepath), { recursive: true }).catch(() => {});
```
mjackson, line 120: I thought we agreed to just let this thing throw?

jacob-ebey: It throws if the directory exists, that's the issue. The message from the file creation will mention directory doesn't exist so we are good.

packages/remix-node/upload/fileUploadHandler.ts
jacob-ebey: I disagree, I might want to use the same filename function but store them in different directories if they are images or PDF's for example. This enables that without creating two separate handlers and having to compose them.

mjackson: Actually, we should probably default it to a tmp directory that we create on node. No need to make `directory` a required option here.

jacob-ebey: Good call!

mjackson (approved):

mikeybinnswebdesign:
The docs need to be updated now this has been merged, specifically here:
https://remix.run/docs/en/v1/api/remix#form-enctype

PR #390, opened 2021-11-17
Make <Scripts> render nothing on the client

developit:
This makes `<Scripts>` do nothing on the client. In Preact, the mismatched `<script>` and `<link>` tags in the SSR'd HTML will be skipped over during hydration, and I'm fairly certain this is also the case in React.

The existing implementation could be further simplified now that useMemo() is no longer needed to prevent reinsertion of scripts after hydration. Also, if there's a compile-time equivalent of the `typeof document !== 'undefined'` check (`process.browser` is a common constant in build systems), that would make it possible to avoid shipping `<Scripts>` to the client at all.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -673,7 +673,7 @@ type ScriptProps = Omit<
 >;
 
 export const Scripts =
-  typeof document === "undefined" ? ServerScripts : ClientScripts;
+  typeof window === "undefined" ? ServerScripts : ClientScripts;
 
 /**
  * Browser implementation of <Scripts>, which does nothing.


ryanflorence:
I think this would be performance regression because `<Scripts>` renders `<link rel="modulepreload">` for all of the *next matches* when a transition starts.

https://github.com/remix-run/remix/blob/2facf6983c1ef402ed2feccad320ac817e8df30f/packages/remix-react/components.tsx#L755-L776

Without that, when we dynamically `import()` the next route module we'll have an import waterfall of all of its dependencies, but the way the code is today the entire module graph for the next page starts preloading in parallel even before we `import()` the next routes.

So it's not just a server thing, it matters for client transitions too.

We already memoize the initial `<scripts>` that are only needed on the server and they never change after the initial render: https://github.com/remix-run/remix/blob/2facf6983c1ef402ed2feccad320ac817e8df30f/packages/remix-react/components.tsx#L702-L741

So it's not clear to me what problem we have and what this PR is attempting to solve? Please re-open (but include the module preloads ofc) if I'm missing something :)

PR #419, opened 2021-11-21
Jokes App example

kentcdodds:
Here's what we need for this to be launch-ready:

- [x] react-router-dom needs to be added as a regular dep to remix
- [x] remix needs to re-export everything from react-router-dom
- [x] We need to update the ~~StackBlitz~~ Codesandbox link in the tutorial to a prepared starter for people
- [x] We need to check the ~~StackBlitz~~ Codesandbox link in the README.md to make sure this runs in ~~stackblitz~~ Codesandbox
- [x] Update the code for the login `redirectTo` feature
- [x] Right before merging, check over all code line highlights to make sure they make sense (sometimes the code moves around)


kumard3:
how can i use react-router-dom   Route  function , in remix ?

PR #422, opened 2021-11-22
Skip script rendering after hydration

developit:
This is a small tweak to @jacob-ebey's #403 that removes the window property. It achieves the same result as #390 but with minimal changes to Remix and no hydration mismatch concerns. Remix's `components` module is only instantiated once, so a module variable is sufficient for tracking `isHydrated` state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,5 @@ yarn-error.log
 /fixtures/test
 /fixtures/my-remix-app
 /fixtures/deno-app
-/fixtures/my-remix-app
 
 .eslintcache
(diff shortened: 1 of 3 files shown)


ryanflorence:
Can ya'll help me understand what this is solving? The script entries only ever render once and then they never change because of the `useMemo`. Why do we need to remove them from the page after hydration?

developit:
@ryanflorence the script elements do re-render when the whole tree gets remounted due to an error or redirect. Memoization doesn't help there.

ryanflorence:
well crap ... tests pass locally for me

PR #424, opened 2021-11-22
feat: re-export react-router-dom utilities

jacob-ebey:


sergiodxa (approved):

packages/remix-react/package.json
```diff
@@ -5,11 +5,13 @@
   "repository": "https://github.com/remix-run/packages",
   "main": "index.js",
   "browser": "browser/index.js",
+  "dependencies": {
+    "react-router-dom": "^6.0.2"
+  },
   "peerDependencies": {
     "history": "^5.1.0",
```
mjackson, line 12: We should get rid of this one too. history is now a regular dep of the router.

mjackson (changes requested):
Almost there, thank you 👍

MichaelDeBoey (approved):

PR #428, opened 2021-11-22
fixed diff highlighting in docs tutorial

SCasarotto:


contributors.yml
```diff
@@ -1 +1,2 @@
 - mcansh
+- SCasarotto
```
MichaelDeBoey, line 2: Should be lowercase for some reason 🤷‍♂️

```suggestion
- scasarotto

```

SCasarotto:
@MichaelDeBoey Thanks for the help there.

contributors.yml
mcansh: ah shoot, i fixed this in the inverse 😂

kentcdodds (commented):
Thanks! We've got other tutorials that are going to replace this one though. Appreciate you taking the time though!

PR #436, opened 2021-11-22
fix: use `remix` exports instead of `react-router-dom`

MichaelDeBoey:


examples/blog-tutorial/.gitignore
```diff
@@ -3,3 +3,5 @@ node_modules
 /.cache
 /build
 /public/build
+/package-lock.json
```
ryanflorence, line 6: we commit our lock file (that's the point!) can you remove these changes here?

MichaelDeBoey: I removed them because of 86b6b1cdec026a521413381033d941a08926ac38

I can revert if needed

chaance: @ryanflorence I ignored them in our examples because I didn't want to end up with both npm and Yarn lockfiles depending on how they were installed, and they aren't a part of our overall workspaces. I don't have a strong opinion here but was hoping to avoid an easy footgun!

kentcdodds: Yeah, I'm fine with that. We wouldn't want this in our templates though 👍

kentcdodds (approved):
Thanks!

PR #438, opened 2021-11-22
Add support for `remix.config.ts`

jensmeindertsma:
Hello 👋.

I love using Remix, and what better way to show than to implement support for a feature I'd really love to see in Remix? It's a minor change but one that I quite like.

This pull requests implements support for `remix.config.ts` in addition to the existing `remix.config.js`. Writing your config files in TypeScript can be nice for projects which try to use as much TypeScript as possible, it increases the confidence you can have that your app still works after a dependency change. It also provides autocompletion ( I'm aware JSDoc also does this ). It also updates the `gists-app` fixture to verify that this feature works. Please let me know if you'd prefer to see this fixture change be reverted and would rather see a new fixture app created for this feature. 

Thanks for all the awesome work on Remix! 🙏


packages/remix-dev/config.ts
```diff
@@ -162,10 +164,37 @@ export async function readConfig(
...
+    let configFileExists = fs.existsSync(configFile);
+    let tsConfigFileExists = fs.existsSync(tsConfigFile);
+    if (tsConfigFileExists) {
+      if (tsConfigFileExists && configFileExists) {
+        throw new Error(
+          "Both `remix.config.ts` and `remix.config.js` exist. Please delete either as Remix doesn't know which one to use."
+        );
+      }
+
+      let temporaryFile = path.resolve(os.tmpdir(), "_remix.config.js");
```
egoist, line 180: this would cause problem if you write `path.join(__dirname, './some-path')` in your config, `__dirname` will be the tmpdir.

jensmeindertsma: I could also generate a _remix.config.js file in the project root just temporarily, then unlink it immediately. Would that be better?

jensmeindertsma: If I do that I might not need the bundle option either

egoist: you do need the bundle option since you might `import` another ts file in your config

jensmeindertsma: I could probably generate the temporary file in the project root instead, then __dirname would still work?

egoist: maybe check out https://github.com/egoist/bundle-require/blob/main/README.md#how-it-works, it's how Vite loads vite.config.ts

jensmeindertsma: Yeah, I might use that instead. Hand rolling my own solution is likely going to cause some problems down the line

jensmeindertsma:
Let me know if there's anything I can do to help get this merged 😊

ryanflorence:
Thanks Jens. We don't want to add the complexity and a new dependency right now, the remix config is pretty basic.

(You might want to check your editor to make sure it's using the project's prettier config because this PR has a bunch of formatting changes in unrelated files.)

jensmeindertsma:
Would you accept this PR if it didn't add any new dependencies ( just use `esbuild` ), or not?

donavon:
> We don't want to add the complexity and a new dependency right now, the remix config is pretty basic.

I don't care how basic is it, this is a bad call, IMO. Maybe not this PR, but is there any way we can get the concept of a `remix.config.ts` reopened? JavaScript is killing me.

donavon:
Here's an example of why it's important to have a type-safe config file. Here `depth` can only be 1 through 6.

image


p.s. I've patched my Remix to allow `remix.config.ts`.

sjc5:
It would be lovely if this were revisited, given that the config potentially includes writing functions to define custom routes and things of that nature. Not to rehash why typescript even exists, but doing that in pure JS is painful and buggy. Got bit by this last night.

Issue #97, opened 2021-03-14
Infer types of the loader function


mellson:
It would be nice to infer the type of the loader function when using `useRouteData`.
I'm not sure how the internals of remix work, so I don't know if there's anywhere one could hook (pun intended) into the place where the actual loading occurs.

Anyway, here is a small utility that could help someone else needing an easy way to get the types.

This code (placed in a helper file somewhere):
```typescript
import { LoaderFunction } from "@remix-run/data";
import { useRouteData } from "@remix-run/react";

type Awaited<T> = T extends PromiseLike<infer U> ? Awaited<U> : T;
export type LoaderType<T extends LoaderFunction> = Awaited<ReturnType<T>>;

export function useTypedRouteData<T extends LoaderFunction>(loader: T) {
  return useRouteData<LoaderType<typeof loader>>();
}

```

allows the user to get the types in one of two ways:
1 - Add the type to the existing hook `let data = useRouteData<LoaderType<typeof loader>>()`
2 - Use the provided wrapper `let data = useTypedRouteData(loader)`

This should work equally well with promises and responses. But there's probably a lot of edge cases I'm not thinking of right now.

joshmedeski:
I got something working. I made the following changes.

1. Drop the loader `LoadFunction` type on the `loader` function and use the `DataFunctionArgs` argument type. This is so Typescript can infer the return type of `loader`.

```ts
import { DataFunctionArgs } from '@remix-run/server-runtime'

export const loader = async ({ params }: DataFunctionArgs) => {
  invariant(params.movieId, 'expected params.movieId')
  const details = await new Tmdb().movie.getDetails(params.movieId)
  return { details }
}
```

2. Create a `useInferredRouteData` reusable type. This extends the `useLoaderData` function and infers the returned type from the promise.

```ts
export function useInferredRouteData<
  T extends (args: DataFunctionArgs) => any
>() {
  return useLoaderData<Awaited<ReturnType<T>>>()
}
```

3. Include `typeof loader` as the generic on the `useInferredRouteData` hook.

```ts
export default function Index() {
  const { details } = useInferredRouteData<typeof loader>()
```

**Warning:** I realize that this is incomplete. If you have conditional response types on the loader function you may run into errors with this code. But it's great for a happy path and the `useInferredRouteData` type can be separated so it can be easily reused.

**Type Name Suggestion:** I'd also suggest to the remix team that `DataFunctionArgs` be changed to `LoaderFunctionArgs` to better match the API.

[...]

PR #439, opened 2021-11-22
Support generic in json (aka typed json)

sergiodxa:
Right now, if you want to ensure you pass the correct data to `json` when returning a response in a loader/action you need to do something like this:

```ts
type LoaderData = { ... }

export let loader: LoaderFunction = async ({ request }) => {
  // variable created just to ensure the object passed to json match the type
  let data: LoaderData = { ... }
  return json(data);
}

export default function Screen() {
  let data = useLoaderData<LoaderData>()
  // ...
}
```

With this super small change, you don't need that extra variable for type safety anymore

```ts
type LoaderData = { ... }

export let loader: LoaderFunction = async ({ request }) => {
  // no more variables, just pass the generic to json, TS will check if data match the type
  return json<LoaderData>(data);
}

export default function Screen() {
  let data = useLoaderData<LoaderData>()
  // ...
}
```

Super simple change, a little less boilerplate, and if you don't pass the generic it will infer it from the data object.

This makes https://github.com/sergiodxa/remix-utils#typed-json useless.


packages/remix-server-runtime/responses.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * A JSON response. Converts `data` to JSON and sets the `Content-Type` header.
  */
-export function json(data: any, init: number | ResponseInit = {}): Response {
+export function json<Data>(data: Data, init: number | ResponseInit = {}): Response {
```
MichaelDeBoey, line 4: Defaulting `Data` to `unknown` makes sure we don't **have** to pass the type

```suggestion
export function json<Data = unknown>(data: Data, init: number | ResponseInit = {}): Response {
```

MichaelDeBoey (approved):

packages/remix-server-runtime/responses.ts
sergiodxa: That's not really needed, TS infer it because it's being used in the param (just tried again)

MichaelDeBoey: Forget what I said then 🙈

eps1lon:
Alternatives: https://github.com/remix-run/remix/issues/97

ryanflorence (approved):
@mjackson I'd like this, you?

PR #440, opened 2021-11-22
Add responseBack response helper

sergiodxa:
While now you can return data from the action, if you need to return a redirect for some reason and want to ensure the user is on the exact same page (pathname and query params) there's a header Referer (the typo is correct) you can use to know from what URL the user came from.

This PR adds a response helper called `responseBack` which receives the Request object to get the header and ask for a fallback URL, because the Referer may not always be there based on user privacy settings (most of the time it's there), this fallback should be something you are ok if the user go there, for example if you are on a filtered list page, submit an action, the Referer will keep the same search params, the fallback can be just the pathname and remove the params, which is a nice fallback in case user has stricter privacy settings.

This is how you could use it:

```ts
export let action: ActionFunction = async ({ request }) => {
  // do something
  return redirectBack(request, { fallback: "/list" });
};
```

This makes https://github.com/sergiodxa/remix-utils#redirect-back useless. 🥳

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -3,3 +3,4 @@
 - jacob-ebey
 - ryanflorence
 - kentcdodds
+- sergiodxa
--- a/packages/remix-server-runtime/responses.ts
+++ b/packages/remix-server-runtime/responses.ts
@@ -41,3 +41,24 @@ export function redirect(
     headers
   });
 }
+
+/**
+ * Create a new Response with a redirect set to the URL the user was before.
+ * It uses the Referer header to detect the previous URL. It asks for a fallback
+ * URL in case the Referer couldn't be found, this fallback should be a URL you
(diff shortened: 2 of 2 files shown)


ryanflorence:
I've messed around with this myself and I'm not sure how to resolve the issue that the fetch request's "Referer" header is the URL of the new page (the page that would have a redirect).

Go open up a remix app and look at the "referer" header on the data requests, they point to the page they are fetching because the browser history API changes the URL immediately.

So, I'm just not ready to rely on that header in a general way in the context of Remix because it's a little bit strange.

It of course works great for the primary use case: a form submission and then a redirect (usually in an auth workflow). I think it's correct every time for that (document or script), but as soon as you try to use it outside of that use case there could be some unexpected behavior.

PR #449, opened 2021-11-23
update blog quick start tutorial to add import statement for processMarkdown 

M0nica:
I think it would be helpful to include the addition of the import statement in the code example, similar to how it's included in the tutorial when other new packages are added.


ryanflorence:
oops, thanks monica, got this one in a different PR (I really should start from the bottom of the list cause those were sooner!)

PR #457, opened 2021-11-23
chore: typo

jesseflorig:
Fixed typo from "compact disk" to "[compact disc](https://en.wikipedia.org/wiki/Compact_disc)"


ryanflorence:
This is very important. Thank you!

PR #470, opened 2021-11-23
Quick Start incorrect example code

vipulbhj:
`Post` type is not exported and wrong file imported in with `~/posts`, where it should just be `~/post`.


ryanflorence:
thanks! got this in later PRs

PR #472, opened 2021-11-23
docs: developer blog fixes

mattmazzola:
I was learning about remix and going through the Developer Blog tutorial and noticed some typos and errors in the code examples.

List of changes
- Various typos
- `Post` was not exported from post.ts but attempted to be imported in next code block
- The instructions said to create singular `post.ts` module, but the code was attempting to import using plural `posts.ts`
- `getPost` function was updated to use `processMarkdown` but that function was not available imported yet
- errors object type was `{}` so adding/accessing properties was error that was not fixed
image

### DIFF
--- a/docs/tutorials/blog.md
+++ b/docs/tutorials/blog.md
@@ -73,7 +73,7 @@ Data loading is built in to Remix.
 
 If your web dev background is primarily in the last few years you're probably used to creating two things here: an API route to provide data and a frontend component that consumes it. In Remix your frontend component is also it's own API route and it already knows how to talk to itself on the server from the browser. That is, you don't have to fetch it.
 
-If your background is a bit farther back than that with Rails, PHP, etc. Then you can think of your Remix routes as backend views using React for templating, but then they know how to seamlessly hydrate in the browser to add some flair. It's progressive enhacement realized in its fullest.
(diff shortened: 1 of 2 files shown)


docs/tutorials/blog.md
```diff
@@ -882,13 +886,17 @@ export default function NewPost() {
 
 TypeScript is still mad, so let's add some invariants to make it happy.
 
-```tsx filename=app/routes/admin/new.tsx lines=[11-14,16-18]
```
mattmazzola, line 885: I'm not 100% on these line numbers. I didn't actually run the build and run the docs to observe. If someone can verify these that would be nice!

ryanflorence:
thank you! we got these fixes in other PR though. As for the type errors, the tutorial adds code as plain JS first, then TypeScript as not to distract the JS users too much :)

mattmazzola:
> we got these fixes in other PR though

Oh ok. No worries

> the tutorial adds code as plain JS first, then TypeScript

Yes, I was aware. I had only intended to fix type errors that remained even after they were supposed to have been fixed by the following code.  For example it would say "Typescript is still mad" then following code snippet would add types as if it would fix all the errors, but it didn't.  I didn't change the tsconfig so I assume we would have same compilation errors.

Although if your sure they were addressed without these changes then can disregard

PR #477, opened 2021-11-23
Update jokes.md

Shumuu:
Although it is just an example, prisma would'nt be defined as we use db for the connection


MichaelDeBoey (approved):

Shumuu:
I signed the CLA, it can now be merged

PR #482, opened 2021-11-23
chore: tsconfig path mappings require baseUrl property

mskoroglu:
Path mappings require `baseUrl` property.
TSConfig Reference: https://www.typescriptlang.org/tsconfig#paths

```js
{
  "compilerOptions": {
    "baseUrl": ".", // this must be specified if "paths" is specified.
    "paths": {
      ...
    }
  }
}

```


packages/create-remix/templates/_shared_ts/tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "resolveJsonModule": true,
     "target": "ES2019",
     "strict": true,
+    "baseUrl": ".",
```
MichaelDeBoey, line 11: This should also be changed in the basic example

MichaelDeBoey (approved):

packages/create-remix/templates/_shared_ts/tsconfig.json
mskoroglu: Thanks Michael!

I changed the files: https://github.com/remix-run/remix/pull/482/commits/41fe1de30f92755711390ee26abb00f308360f24

MichaelDeBoey (approved):

PR #486, opened 2021-11-23
add engines to package.json and set engines-strict=true

UsamaHameed:
- adds the engines property to package.json 
- sets `engines-strict=true` in `.npmrc` to throw an error if deps are installed with an incompatible node version 

more on this here: 
https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines


kentcdodds (approved):
Thanks!

PR #487, opened 2021-11-23
clarifies what file the user is supposed to view

matsdahlin:
Even though the file name is mentioned by Derrick a bit further down, this sentence was throwing me off when trying the guide.


MichaelDeBoey (approved):

mcansh:
thanks! looks like the example docs were updated to be more clear in #560

PR #505, opened 2021-11-23
documentation: fixed some path finding issues that leads into 404

chandelieraxel:
Some links within the documentation were leading to 404 pages, because the paths were wrong

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,5 +1,6 @@
 - abereghici
 - chaance
+- chandelieraxel
 - goncy
 - ianduvall
 - jacob-ebey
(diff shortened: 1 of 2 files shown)


chandelieraxel:
Still missing one, couldn't find the proper markdown file 😅

ryanflorence:
Thanks for the work here, you mind rebasing with main and verifying if the links still need to be fixed?

If so, please use markdown link identifiers in the bottom of the document. We started adding them all there so that it's easier to update/fix links across the project. If you scroll to the bottom you can see how it works.

kentcdodds:
I think these have been fixed. If not feel free to open a new PR. Thanks!

PR #506, opened 2021-11-23
docs(jokes): fix injection flaw in RSS feed

dunglas:
The generated RSS feed contains [unescaped user-generated content](https://cyber.harvard.edu/rss/encodingDescriptions.html). This can lead to bugs and even to [security vulnerabilities](https://www.insecure.in/papers/HackingFeeds.pdf). This PR should fix the issue (to be double-checked).

See also https://github.com/remix-run/remix-jokes/pull/6


kentcdodds (approved):
Good call 👏

kentcdodds (approved):
Perfect. Thanks!

PR #508, opened 2021-11-23
docs(jokes): fix RSS date format

dunglas:
[According to the spec](https://validator.w3.org/feed/docs/rss2.html), `pubDate` must be a valid RFC 822 date. It's not currently the case, and weird strings can be injected depending on the user's local (in French for instance: `Tue Nov 23 2021 13:44:16 GMT+0100 (heure normale d’Europe centrale)`).


MichaelDeBoey (approved):

kentcdodds (approved):
Thanks 👏

PR #523, opened 2021-11-23
Fix #517: Fetcher action errors not going to error boundary

### LINKED ISSUE #517: [Bug]: Errors when submitting fetcher.Form don't fall back to ErrorBoundary

- [X] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [x] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/vercel`

1.0.4

### Steps to Reproduce

Add a route with the following code: ([stackblitz link](https://stackblitz.com/edit/github-6ofazo?file=app/routes/fetcher.tsx))

```jsx
import { useFetcher } from "remix";

export const action = () => {
  throw new Error("Whoops!");
};

export default function Component() {
  const fetcher = useFetcher();

  return (
    <div>
      <fetcher.Form method="post">
        <button type="submit">Submit fetcher form</button>
      </fetcher.Form>
      <form method="post">
        <button type="submit">Submit regular form</button>
      </form>
    </div>
  );
}
```

Open up the route and click the `Submit fetcher form` button.

### Expected Behavior

The error from the action should be handled by falling back to the nearest `ErrorBoundary`. The docs [here](https://remix.run/docs/en/v1.0.4/api/remix#usefetcher) back up this expectation.

> Handles uncaught errors by rendering the nearest ErrorBoundary (just like a normal navigation from \<Link\> or \<Form\>)

### Actual Behavior

The form submit fails (500 in network tab), but it does not fall back to the `ErrorBoundary`. If you submit the regular form, this is what happens.

### Failing Test

I've implemented a failing test in #523.

airjp73:
Fixes #517.


fixtures/gists-app/app/root.jsx
```diff
@@ -66,7 +66,7 @@ export default function Root() {
         <div data-test-id="content" id="content">
           <Outlet />
         </div>
-        {data.enableScripts && (
+        {data?.enableScripts && (
```
airjp73, line 69: This seems unrelated, but I was getting an error reading `enableScripts` of undefined.

packages/remix-react/transition.ts
```diff
@@ -811,7 +811,11 @@ export function createTransitionManager(init: TransitionManagerInit) {
       update({
         fetchers: new Map(state.fetchers),
         error: result.value,
-        errorBoundaryId
+        errorBoundaryId,
+        location: {
+          ...state.location,
+          state: { isRedirect: false, type: "error" }
```
airjp73, line 817: The issue seems to be that `errorBoundaryId` was being set here, but the error boundary only updates its error state when the `location` changes. Removing that check caused other tests to fail, so I added a location change.

ryanflorence:
See https://github.com/remix-run/remix/issues/517#issuecomment-995007982

PR #524, opened 2021-11-23
docs: fix Property 'title' does not exist on type '{}'

sunadoi:
TypeScript is mad, so let's add  Record type to make it happy.

```
let errors = {};
if (!title) errors.title = true; // Property 'title' does not exist on type '{}'
```

fixed
```
let errors: Record<string, boolean> = {};
if (!title) errors.title = true;
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -10,3 +10,4 @@
 - morinokami
 - meetbryce
 - ryanflorence
+- sunadoi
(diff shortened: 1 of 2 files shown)


kentcdodds (commented):
Thanks for this! Looks like two of you found some needed improvements to this bit. I think I prefer the fix offered here: #757 

Thanks a lot for taking the time anyway.

Issue #532, opened 2021-11-23
[Bug]: Demo - URL Params and Error Boundaries - Nested `<p>` console error in `CatchBoundary`


joshuadotson:
### Which Remix packages are impacted?

- [X] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.0.4

### Steps to Reproduce

* nav to `./examples/basic`
* install dependencies `npm i`
* start dev server `npm dev`
* open `http://localhost:3000/`
* choose `URL Params and Error Boundaries` under `Demos In This App`
* open dev tools console
* select `This will be a 404` under `Click these Links`
* observe error in console

### Expected Behavior

No error occurs because the HTML structure is valid.

### Actual Behavior

`Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.`

Above error occurs because the HTML structure is not valid.

machour:
Fixed in #535

PR #533, opened 2021-11-23
Fixes nested `<p>` tag error on URL Params and Error Boundaries demo

joshuadotson:
Fixes https://github.com/remix-run/remix/issues/532

Removed `<p>` that wraps message to avoid nested `<p>` error and allow for `message` to remain of type `ReactNode` for future flexibility in error messaging.

The other option was to remove all the `<p>` tags within `CatchBoundary` and leave the tag around `{message}` but that would imply that the message content should be (mostly) of type `string`.


ryanflorence:
Looks like we've already fixed this. Thank you for your help!

PR #536, opened 2021-11-23

MichaelDeBoey:
To make @kentcdodds happy

https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

### DIFF
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -273,13 +273,15 @@ export default function Signup() {
       <Form method="post">
         <p>
           <input type="text" name="email" />
-          {errors?.email && <span>{errors.email}</span>}
+          {errors?.email ? (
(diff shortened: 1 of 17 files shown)


krszwsk:
Wondering if casting to a boolean and keeping the `&&` is an equally good solution 🤔

kentcdodds (approved):
Hee hee hee. And because I'm the docs owner I get to do this :)

kentcdodds:
Oh rats. I just realized a few of these have highlighted lines that will be off based on this change. @MichaelDeBoey, do you think you could take a minute to check and fix any highlighted lines that broke as a result of this change?

MichaelDeBoey:
@kentcdodds I'll look into this and create a separate PR 👍

PR #545, opened 2021-11-23
feat: add fastify adapter

mcansh:
still need to write ~~some tests and~~ an example/template, but using this puppy on a side project: https://shredder.mcan.sh


mcansh:
> Thank you for signing the Contributor License Agreement. Let's get this merged! 🥳

hm, thought i fixed sending a comment to an open PR if they've signed already...

packages/remix-fastify/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/remix-run/remix",
+    "directory": "packages/remix-fastify"
+  },
+  "bugs": {
+    "url": "https://github.com/remix-run/remix/issues"
+  },
+  "dependencies": {
+    "@remix-run/node": "1.0.4"
```
MichaelDeBoey, line 14: Shouldn't this package also have `@remix-run/server-runtime` as dependency? 🤔

mcansh: `@remix-run/node` has `@remix-run/server-runtime` as a dependency :)

MichaelDeBoey: I think all of them should have `@remix-run/server-runtime` as a dependency then or at least a `peerDependency`

packages/remix-fastify/__tests__/server-test.ts
```diff
@@ -0,0 +1,272 @@
...
+        method: "GET",
+        url: "/"
+      });
+
+      expect(response.statusCode).toBe(200);
+    });
+
+    it("handles status codes", async () => {
+      mockedCreateRequestHandler.mockImplementation(() => async () => {
+        return new Response("", { status: 204 });
```
MichaelDeBoey, line 79:
```suggestion
        return new Response(null, { status: 204 });
```

See #847

packages/remix-fastify/__tests__/server-test.ts
```diff
@@ -0,0 +1,272 @@
...
+        );
+        headers.append(
+          "Set-Cookie",
+          "second=two; MaxAge=1209600; Path=/; HttpOnly; Secure; SameSite=Lax"
+        );
+        headers.append(
+          "Set-Cookie",
+          "third=three; Expires=Wed, 21 Oct 2015 07:28:00 GMT; Path=/; HttpOnly; Secure; SameSite=Lax"
+        );
+        return new Response("", { headers });
```
MichaelDeBoey, line 106:
```suggestion
        return new Response(null, { headers });
```

See #847

MichaelDeBoey (changes requested):

ryanflorence:
This is awesome.

But we've got to figure out a better way to support a bunch of adapters without all of them being part of the remix source code. Let's figure this out after the break. I made a linear issue.

mcansh:
yeah i felt the same when i opened it haha 👍

rsbh:
What is the status of fastify integration?
I am fine in using custom solution but i prefer official support.
Can you folks share the status or timeline regarding fastify support

chaance:
@rsbh This isn't a huge priority for us at the moment, sorry! Official support means for more adapters is out of scope at the moment, but if we revisit this in the future we can revisit this discussion.

PR #546, opened 2021-11-23
Update template to use correct property name for assets directory

nielsdB97:
Since `browserBuildDirectory` is deprecated in favor of `assetsBuildDirectory`, let's update our template 🙂


ryanflorence:

nielsdB97:
I'm afraid this is not yet fixed [on `main`](https://github.com/nielsdB97/remix/blob/main/packages/create-remix/templates/_shared_js/remix.config.js#L6), so I think this should be revisited.

PR #547, opened 2021-11-23
Set body to null for null body status code responses

jamiedavenport:
According to the [fetch spec](https://fetch.spec.whatwg.org/#statuses)

> A null body status is a status that is 101, 204, 205, or 304.

So we should be setting the body to `null`. This is also causing errors when using Miniflare:

```
GET /admin/new?_data=routes%2Fadmin%2Fnew: TypeError: Response with null body status cannot have body
```


MichaelDeBoey (approved):

jamiedavenport:
Would it be possible to get this merged in since it's been approved? @MichaelDeBoey @ryanflorence

MichaelDeBoey:
@jamiedavenport I'm not from the Remix team, so I can't merge this one.
I'm only helping out a bit by doing reviews.

You'll have to be patient until somebody from the team finds some time to go over this.

kentcdodds:
Hey @jamiedavenport thanks for making this PR. We were a bit overwhelmed with PRs when we open sourced and @jacob-ebey fixed this one without realizing you'd made this PR. Should be fixed in the next release: https://github.com/remix-run/remix/pull/743

Thanks for taking the time!

PR #549, opened 2021-11-23
Joke App tutorial: (docs) should not have the prefetch at this point.

gon250:


MichaelDeBoey (approved):

kentcdodds (approved):
Ah, good catch. Thanks!

PR #551, opened 2021-11-23
Add displayName to RouteChangeAnnouncement

andrelandgraf:
The displayName is used by React in debugging messages. Adding the display name satisfies the eslint rule [react/display-name](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md).

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,3 +1,4 @@
+- andrelandgraf
 - BasixKOR
 - chaance
 - jacob-ebey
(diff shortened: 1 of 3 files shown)


MichaelDeBoey:
> [...] you don’t need to set it explicitly because it’s inferred from the name of the function or class that defines the component.

https://reactjs.org/docs/react-component.html#displayname

kentcdodds:
We removed this from the template. Thanks though!

PR #556, opened 2021-11-23
Fix form links in conventions.md

ledenis:
Hi, this fixes two "see also" links that were incorrect


ledenis:
Closing since it was fixed by https://github.com/remix-run/remix/pull/890

PR #566, opened 2021-11-24
Remove reference to REMIX_TOKEN from fly.io readme

juhanakristian:
REMIX_TOKEN was removed from the package.json deploy script in #519 but the README.md still referenced it.


mcansh:
thanks! Must've missed a few when I updated the cli 🤦

PR #568, opened 2021-11-24
chore(docs): add missing id to textarea

sitek94:
Hey there, I noticed that even though we're using `htmlFor` there is no `id` on textarea :)


kentcdodds (approved):
Super, thanks!

PR #569, opened 2021-11-24
feat: prefetch link preloads

ryanflorence:
If a route exports `<link rel="preload">`, those are added as prefetches for `<Link prefetch>`. 

If the route says they are important enough to make high priority when the route becomes active, then we can speed it up even more by prefetching before the user gets there.

This enables routes to tell links to prefetch any kind of asset without the linking component needing to know.


packages/remix-react/links.ts
```diff
@@ -8,6 +8,7 @@ import type { RouteMatch } from "./routeMatching";
 // import { matchClientRoutes } from "./routeMatching";
 import type { RouteModules, RouteModule } from "./routeModules";
 import { loadRouteModule } from "./routeModules";
+import invariant from "./invariant";
```
sergiodxa, line 11: This is not being used

PR #589, opened 2021-11-24
Update blog.md

atrimn:
Fixed type from React app server to Remix app server


ryanflorence:

PR #607, opened 2021-11-24
Fix: ESlint

Angelfire:
Remove unused imports in the `create-remix` package


MichaelDeBoey (approved):

MichaelDeBoey:
This is already fixed in #611

ryanflorence:
Thanks for your work here, the code has changed a lot since this and I think it addresses the issues.

PR #620, opened 2021-11-25
fix: set isolatedModules in template tsconfig.json

tylerbrostrom:
Remix compiles TypeScript using esbuild. esbuild compiles each file in isolation; it can’t tell if an import is a type or a value. This can lead to miscompilation of certain TypeScript features.

Enabling the isolatedModules option in tsconfig.json safeguards against such potential miscompilations.

References:

- [esbuild docs](https://esbuild.github.io/content-types/#isolated-modules)


mcansh (approved):

kentcdodds (approved):
Solid 👍

PR #623, opened 2021-11-25
`remix dev` - Custom port note

TheRealFlyingCoder:
I think this should be a config variable, but for now at least we can show how to set it manually


docs/other-api/dev.md
```diff
@@ -52,3 +52,4 @@ Same as `watch` but also boots the [Remix app server](serve.md) in development m
 ```sh
 $ remix dev
 ```
```
ryanflorence, line 54: you wanna add some whitespace under this?

ryanflorence:
does it need the `cross-env`?

Also might want to note that `REMIX_DEV_SERVER_WS_PORT` if you want to change that too (which you will if you're trying to run two apps with `remix dev` at the same time.

TheRealFlyingCoder:
> does it need the cross-env?

It did not, i've updated to be more generic :)

>Also might want to note that REMIX_DEV_SERVER_WS_PORT if you want to change that too (which you will if you're trying to run two apps with remix dev at the same time.

Mentioned, and linked to the relevant doc 👍

machour (approved):

Issue #557, opened 2021-11-23
[Bug]: Live reload does not work with Netlify starter


jurajkrivda:
### Which Remix packages are impacted?

- [X] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [X] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [X] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.0.4

### Steps to Reproduce

Create the app with `npx create-remix@latest` and select Netlify.

### Expected Behavior

Content should be changed

### Actual Behavior

The content has not changed

uwuru:
What commands are you running? Are you running both commands as per the instructions?

It sounds like you are not also running `yarn dev` which deals with the hot reloading.

jurajkrivda:
Hi. Thank you for your response.
My command: `"dev": "concurrently \"remix watch\" \"yarn dev:netlify\"",`
but I tried separately in two CLI windows. 

In Chrome dev console I see 
```
(index):19 WebSocket connection to 'ws://localhost:8002/socket' failed: 
(anonymní) @ (index):19
(index):31 Remix dev asset server web socket error:
```

gabrielvincent:
I'm having the same problem. It happens in both cases:

### Case 1
```
npm run remix watch # on terminal window 1
npm run dev:netlify # on terminal window 2
```

### Case 2

```
npm run remix dev
# if I run `npm run dev:netlify` on another terminal window the command fails because port 3000 is already in use
```

I both cases I see the following error on the browser console:

```
(index):19 WebSocket connection to 'ws://localhost:8002/socket' failed: 
(index):31 Remix dev asset server web socket error:
(index):32 
Event {isTrusted: true, type: 'error', target: WebSocket, currentTarget: WebSocket, eventPhase: 2, …}
isTrusted: true
bubbles: false
cancelBubble: false
cancelable: false
composed: false
currentTarget: WebSocket {url: 'ws://localhost:8002/socket', readyState: 3, bufferedAmount: 0, onopen: null, onerror: ƒ, …}
defaultPrevented: false
eventPhase: 0
path: []
returnValue: true
[...]

PR #624, opened 2021-11-25
create-remix -  update devServerPort to 8002 instead of 8888

kalijonn:
BugFix for https://github.com/remix-run/remix/issues/557

while creating the WebSocket Server, remix reads devServerPort from remix.config.js in the root directory. By default it's 8888 while the websocket is trying to connect to 8002. By changing devServerPort on remix.config.js to 8002, LiveReload works.


MichaelDeBoey:
@kalijonn Seems like this one can be closed

Issue #599, opened 2021-11-24
[Bug]: Loader data gets undefined when an error is thrown in a child route


Alarid:
### Which Remix packages are impacted?

- `remix` (Remix core)

### What version of Remix are you using?

1.0.3

### Steps to Reproduce

I couldn't reproduce the bug with the remix template on Codesandbox, but this happens 100% of the time with [the Remix Jokes tutorial](https://remix.run/docs/en/v1/tutorials/jokes). You can watch the bug in action [at this moment](https://youtu.be/hsIWJpuxNj0?t=17563) in @kentcdodds's lifestream yesterday. He asked if someone could open up an issue for it, so here I am.

I have created a public repo on Github with the code you are supposed to have when you have been following along the Jokes tutorial, just before beginning the "SEO with Meta tags" part. Here's how to see the bug in action:

1. Clone [this repo](https://github.com/Alarid/remix-error-boundary-bug-repro)
2. Run `npm i`
3. Create a .env file and add `DATABASE_URL="file:./dev.db"` in it, along with `SESSION_SECRET="whatever"`
4. Seed the DB: `npx prisma db seed`
5. Run the app: `npm run dev`
6. Click "Read jokes", then "Add your own" (or go to http://localhost:3000/jokes/new)
7. Submit the form
8. Kaboum!

### Expected Behavior

An error thrown in a route action should be caught by the route's error boundary (or the closest one above). The parent's loader data shouldn't become undefined when this happens.

### Actual Behavior

The error occurring in the `/jokes/new` route `action` function isn't handled by the route's error boundary, instead, the `/jokes` route loader data becomes undefined, triggering an error that is caught by the root error boundary:

```
TypeError: Cannot read property 'user' of undefined
    at `{data.user ? (` (./app/routes/jokes.tsx:50:16)
    [...]
```

andmayorov:
I've added empty `CatchBoundary` to `routes/jokes.tsx` like this and it fixed it for me.

```ts
export function CatchBoundary() {}
```

Not sure why

Alarid:
Indeed, it fixed it for me too. But I'm definitely keen to know **why** the loader data becomes undefined when a child route throw an error in the action, if someone got an explanation.

Xiphe:
The bug does not need database and secret to be reproduced. Here are slightly faster steps:


1. check out `git@github.com:Xiphe/remix-error-action-loader-undefined.git`
2. run `npm install && npm run dev`
[...]

PR #635, opened 2021-11-25
fix: `useLoaderData` in parent route without boundary handlers should return data when child throws an error

andmayorov:
Currently there is a bug which easier to explain by this example:

1. We have 3 nested routes
```ts
let matches = ["root", "/jokes", "/jokes/$jokeId"];
```
2. `root` and `/jokes/$jokeId` modules export `CatchBoundary` function while `/jokes` doesn't
3. Now we throw an error inside action in `/jokes/$jokeId` module and expect  `CatchBoundary` to handle it.
4. But instead we get some different kind of error because `useLoaderData` inside `/jokes` returns `undefined`. And error is handles by `ErrorBoundary` inside `root` module instead.

The reason why it happens is that at the moment we use `getMatchesUpToDeepestBoundary` to filter out `matchesToLoad`. We expect `matchesToLoad` to contain `["root", "/jokes"]` but instead it contains `["root"]`.

Since `/jokes` doesn't have `CatchBoundary` `getMatchesUpToDeepestBoundary` ignores it. Therefore it doesn't initialise `/jokes`'s `loader` even though it didn't throw any errors and Remix still tries to render it.

Related issue: https://github.com/remix-run/remix/issues/599

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -15,3 +15,4 @@
 - morinokami
 - msutkowski
 - ryanflorence
+- andmayorov
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -363,25 +363,13 @@ async function handleDocumentRequest(
   // If we did not match a route, there is no need to call any loaders
   let matchesToLoad = requestState !== "ok" ? [] : matches;
   switch (responseState) {
+    // get rid of the action, we don't want to call it's loader either
+    // because we'll be rendering the catch/error boundary, if you can get access
+    // to the loader data in the catch boundary then how the heck is it
+    // supposed to deal with thrown responses?
     case "caught":
-      matchesToLoad = getMatchesUpToDeepestBoundary(
-        // get rid of the action, we don't want to call it's loader either
-        // because we'll be rendering the catch boundary, if you can get access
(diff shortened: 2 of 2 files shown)


ryanflorence:
@andmayorov thanks for your work here! I think we've fixed this though.

@jacob-ebey I think your recent changes fix this?

If not, please re-open, either one of you :)

sbking:
@ryanflorence This is still happening when following the official Jokes tutorial as of today. It happens in the [Expected Errors](https://remix.run/docs/en/v1/tutorials/jokes#expected-errors) section, when you try to delete a joke that you don't own.

JudeTejada:
any updates on this?

llwt:
I just ran into this as well working through the tutorial today.

machour:
Too many things have changed, another PR needs to be produced if the problem persists.
Meanwhile, #599 is still open to track the bug, please comment on the issue instead 🙌🏼

nikshirobokov:
> Problem definitely still exists as of June 2022.

Yep! Confirmed. 
Remix version `1.6.2`

machour:
Please read through https://github.com/remix-run/remix/issues/599 and open a new issue if the problem still exists.
Commenting on a closed PR isn't likely to get you anywhere.

PR #637, opened 2021-11-25
chore(create-remix): simplify Netlify template

ascorbic:
Hello!
We've [updated the Netlify CLI](https://github.com/netlify/cli/pull/3662) to make the DX better for Remix, removing the need to run two commands. This PR updates the template to add the new instructions and configuration.

We will soon be adding auto-detection for Remix too, which will remove the need for most of the configuration too. I will do another PR when that is ready.

Keep up the great work!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -15,3 +15,4 @@
 - morinokami
 - msutkowski
 - ryanflorence
+- ascorbic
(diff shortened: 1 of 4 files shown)


MichaelDeBoey (approved):

mcansh:
awesome @ascorbic! I missed this quickly combing through the issues this morning, we'll get this released soon!

mcansh (approved):

PR #611, opened 2021-11-24
chore: fix ESLint errors & warnings

MichaelDeBoey:
Fixes #625 
Fixes #805

PR #638, opened 2021-11-25
refactor(examples/basic/app): remove unused imports

marcelscr:
Remove unused imports from `examples/basic/app`.


MichaelDeBoey:
This is already fixed in #611

PR #641, opened 2021-11-25
docs: Adapt link targets for posts in admin ui

HenryVogt:
This fix addresses the tutorial docs for the blog example and the respective example sources. The links in the Admin UI for the existing posts are broken. This adaption adds `/posts/` as prefix to the slug so that the links lead to the correct documents.


HenryVogt:
Closes #644

kentcdodds (approved):
Thanks!

PR #642, opened 2021-11-25
chore: add support of worker-types v3 on cf adapter

edmundhung:
Cloudflare released an [update](https://github.com/cloudflare/workers-types/releases/tag/v3.0.0) for `@cloudflare/workers-types` last month.
It is now automatically generated from the source as announced [here](https://blog.cloudflare.com/automatically-generated-types/).
This is really helpful for development as it ensures typing to be always up-to-dated with the latest runtime.

However, the current CF adapter has a peerDependency of v2 only. It would be nice if we could add support of v3 as well.


edmundhung:
Closing this in favour of #645 as it should be branching off from dev based on the CLA

PR #648, opened 2021-11-25
Add Koyeb as a deployment option

edouardb:
Hi there 👋 

Added Koyeb as a deployment option.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -15,3 +15,4 @@
 - morinokami
 - msutkowski
 - ryanflorence
+- edouardb
--- a/docs/guides/envvars.md
+++ b/docs/guides/envvars.md
@@ -21,6 +21,7 @@ If you're experience with web development is primarily with the JS frameworks in
 Environment variables on your server will be handled by your host, for example:
 
 - [Fly.io](https://fly.io/docs/reference/secrets/)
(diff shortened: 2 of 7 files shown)


sergiodxa:
Since the template doesn't come with anything different than the Remix App Server, what's the benefit of having another template? Why templates ask for the deployment target, they are actually for the adapter you want to use, any deployment service that doesn't need an special adapter would be the same as Remix App Server or Express if you want control of the server.

edouardb:
Hi @sergiodxa 

Thanks for your reply!
I took inspiration from the Fly template which doesn’t seem to have any adapter. Did  I miss something?

sergiodxa:
🤔 I think Fly used to require extra config and it doesn't anymore.

Anyway, I'm not a part of the team, so they would decide if they want it 💪

edmundhung:
As an alternative, maybe we can consider making an extra options on create-remix which bootstrap with the express template.

chaance:
Hi @edouardb, thanks for the PR! I'm not sure we are quite ready to start extending our packaged adapters just yet, but we have a new feature coming that will be very community driven and will be perfect for stuff like this. Stay tuned!

kentcdodds:
As @chaance said, we have something even better coming soon, so I'm going to go ahead and close this PR. Stay tuned though!

PR #662, opened 2021-11-25
Use superjson to stringify and parse data between server and client

### LINKED ISSUE #656: [Feature]: Use Blitz.js superjson instead of native JSON.serialize/parse

Use [superjson](https://github.com/blitz-js/superjson) to serialize and parse data returned from loaders/actions.

Right now Remix uses JSON.stringify and JSON.parse, this means types like Date (see #634) are going to be converted to strings when serializing it but not converted back to Date on the revive step, and other types like BigInt are not even compatible with JSON.stringify, so if an ORM returns one the user needs to manually convert them to a string before returning the data/response inside the loader.

This superjson library created by the Blitz.js team have differents stringify and parse method that it's capable of serializing more types of data and reviving them back to the original class (see the table here https://github.com/blitz-js/superjson#parse), this way if a loader does this:

```ts
return json({ date: new Date() })
```

Then the useLoaderData will get this

```ts
let { date } = useLoaderData<{ date: Date }>()
let fullYear = date.getFullYear() // this will work
```

And other native objects like BigInt, RegExp, Set, Map and Error will also work as valid return values.

---

Additionally, superjson can be extended to serialize and revive other objects, it may be cool if Remix let you customize that somehow, but I think just the default config would be better than the native JSON methods.

sergiodxa:
Fixes #656 

This implements the Blitz.js superjson library to serialize and parse values.

This way if a loader or action returns `undefined`, BigInt, Date, RegExp, Set, Map or Error objects the useLoaderData will receive the instance of those objects instead of a string or on error because JSON.stringify can't serialize them.


kentcdodds:
This seems pretty sweet. Would be a breaking change so we'd probably want to start with a config option to enable this. Probably would want it to be unstable.

Even before that, a remix-superjson lib would be good for folks to try it out in real apps.

In any case, I can't say that we're going to merge this, but those are just a few of my thoughts.

ryanflorence:
Let's open a discussion for this. I think I like it?

sergiodxa:
> Let's open a discussion for this. I think I like it?

There’s an issue here https://github.com/remix-run/remix/issues/656, maybe convert it to discussion?

PR #663, opened 2021-11-25
docs: add jit compiler to tailwind config

### LINKED ISSUE #581: [Feature]: Add Tailwind JIT mode in your docs

Add Tailwind JIT mode method at https://remix.run/docs/en/v1/guides/styling

Cause jit is a lot fast, that means remix will be a lot faster too!

mnlfischer:
The JIT compiler generates styles on-demand and increases the browser performance in development.

Reference:
[Tailwind JIT](https://tailwindcss.com/docs/just-in-time-mode)

fixes #581

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -15,3 +15,4 @@
 - morinokami
 - msutkowski
 - ryanflorence
+- mnlfischer
--- a/docs/guides/styling.md
+++ b/docs/guides/styling.md
@@ -86,6 +86,7 @@ Here's a quick guide to getting it set up. We encourage you to read the official
 
     ```js filename=tailwind.config.js
     module.exports = {
+      mode: 'jit', // enables just-in-time compiler
       purge: [
         "./app/**/*.tsx",
         "./app/**/*.jsx",


MichaelDeBoey:
You can always do this yourself if you want to.

V3 will use JIT automatically btw

mnlfischer:
> You can always do this yourself if you want to.
> 
> V3 will use JIT automatically btw

I understand your point but for now tailwind v3 is in alpha state and I have no insights into it. The majority is using jit mode and why not forcing a recommendation to use it. Using the inspector with all classes can end up in some annoying developer experience. 
At the end this is addressed in v3 and I have no emotions on it.

MichaelDeBoey:
I think it's fair the Remix team only pushes the defaults Tailwind is providing (by doing `tailwindcss init`), but I also don't have a strong opinion on this

mnlfischer:
already done in [Commit](https://github.com/remix-run/remix/commit/5a0108827f0d1ac001138dfb5e28624450a642f9)

PR #664, opened 2021-11-25
feat(remix-dev): add `devServerRebuildingIndicator` config option

ikhattab:
add spinner in LiveReload component to show spinning 💿 when any change that require live reload happens

https://user-images.githubusercontent.com/401304/143497677-80f48d9e-c9eb-44e8-bfbe-8a729ad44d87.mov

**UPDATE 1:**

I made a few updates in [cc8b0d2](https://github.com/remix-run/remix/pull/664/commits/cc8b0d2a21b7b78f2cab6561cc516e0661d46b86) to allow adding `devServer` key to `remix.config` that can configure a few options for now. e.g:

```
devServer:  {
  broadcastDelay: 0,
  port: 8002,
  rebuildingIndicator: true,
}
```

I took the same approach as deprecating browserBuildDirectory added `@deprecated` flag and still alow `devServerPort` and `devServerBroadcastDelay` to be used.

**UPDATE 2:**

I created a separate PR for refactoring devServer options in https://github.com/remix-run/remix/pull/923

### DIFF
--- a/packages/remix-dev/__tests__/readConfig-test.ts
+++ b/packages/remix-dev/__tests__/readConfig-test.ts
@@ -28,6 +28,7 @@ describe("readConfig", () => {
         "cacheDirectory": Any<String>,
         "devServerBroadcastDelay": 0,
         "devServerPort": Any<Number>,
+        "devServerRebuildingIndicator": true,
         "entryClientFile": "entry.client.tsx",
         "entryServerFile": "entry.server.tsx",
         "mdx": undefined,
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -202,6 +202,9 @@ export async function watch(
       start = Date.now();
       onRebuildStart?.();
       log("Rebuilding...");
+      if (config.devServerRebuildingIndicator) {
+        broadcast({ type: "REBUILDING" });
+      }
     },
     onRebuildFinish() {
       log(`Rebuilt in ${prettyMs(Date.now() - start)}`);
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1437,36 +1437,70 @@ export const LiveReload =
       }) {
         let js = String.raw;
         return (
-          <script
-            nonce={nonce}
-            suppressHydrationWarning
-            dangerouslySetInnerHTML={{
-              __html: js`
-                (() => {
-                  let protocol = location.protocol === "https:" ? "wss:" : "ws:";
-                  let host = location.hostname;
-                  let socketPath = protocol + "//" + host + ":" + ${String(
-                    port
-                  )} + "/socket";
+          <>
+            <span
+              id="__remix-livereload-spinner"
+              aria-label="loading..."
+              role="img"
+            >
+              💿
+            </span>
+            <script
+              nonce={nonce}
+              suppressHydrationWarning
+              dangerouslySetInnerHTML={{
+                __html: js`
+                  (() => {
+                    let protocol = location.protocol === "https:" ? "wss:" : "ws:";
+                    let host = location.hostname;
+                    let socketPath = protocol + "//" + host + ":" + ${String(
+                      port
+                    )} + "/socket";
 
-                  let ws = new WebSocket(socketPath);
-                  ws.onmessage = (message) => {
-                    let event = JSON.parse(message.data);
-                    if (event.type === "LOG") {
-                      console.log(event.message);
-                    }
-                    if (event.type === "RELOAD") {
-                      console.log("💿 Reloading window ...");
-                      window.location.reload();
-                    }
-                  };
-                  ws.onerror = (error) => {
-                    console.log("Remix dev asset server web socket error:");
-                    console.error(error);
-                  };
-                })();
+                    let ws = new WebSocket(socketPath);
+                    let spinner = document.getElementById('__remix-livereload-spinner');
+                    ws.onmessage = (message) => {
+                      let event = JSON.parse(message.data);
+                      if (event.type === "LOG") {
+                        console.log(event.message);
+                      }
+                      if(event.type === "REBUILDING") {
+                        spinner.classList.add("loading");
+                      }
+                      if (event.type === "RELOAD") {
+                        console.log("💿 Reloading window ...");
+                        window.location.reload();
+                      }
+                    };
+                    ws.onerror = (error) => {
+                      console.log("Remix dev asset server web socket error:");
+                      console.error(error);
+                    };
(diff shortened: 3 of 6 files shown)


jensmeindertsma:
I love the little spinning disk! Really nice!

packages/remix-dev/cli/commands.ts
```diff
@@ -107,6 +107,7 @@ export async function watch(
       start = Date.now();
       onRebuildStart && onRebuildStart();
       log("Rebuilding...");
+      broadcast({ type: "REBUILDING" });
```
jensmeindertsma, line 110: Here you could maybe make this conditional based on whether it's enabled in `remix.config.js`? It could be enabled by default.

jensmeindertsma (changes requested):
Could you make it configurable? That's something I think would be nice.

packages/remix-dev/cli/commands.ts
ikhattab: this would be nice, was thinking to keep broadcasting the event I believe it will be useful as could be used in a different context to show rebuilding and instead to control the behavior in the LiveReload component, what do you think?

packages/remix-react/components.tsx
```diff
@@ -1281,14 +1284,39 @@ export function LiveReload({ port = 8002 }: { port?: number }) {
...
+          bottom: 10px;
+          right: 10px;
+          font-size: 20px;
+          user-select: none;
+        }
+        #__remix-livereload-spinner.loading {
+          visibility: visible;
+          animation: spin 1s linear infinite;
+        }
+        @keyframes spin {
```
garand, line 1313: I'd recommend adding a prefix to the animation name to prevent any possible conflicts.

ikhattab: good point

ikhattab: I checked and it has really good support https://caniuse.com/mdn-css_at-rules_keyframes also searched the repo found multiple uses of it without prefix I believe it's good to leave it without one

garand: Sorry I mean prefix the name of the animation. Incase someone using Remix has another animation called spin.

ikhattab: aha got it!

packages/remix-dev/__tests__/readConfig-test.ts
```diff
@@ -27,6 +27,7 @@ describe("readConfig", () => {
         "cacheDirectory": Any<String>,
         "devServerBroadcastDelay": 0,
         "devServerPort": 8002,
+        "devServerShouldShowLoadingSpinner": true,
```
jacob-ebey, line 30: Let's be careful about adding root keys to the config. Can we try to come up with a key name for development overlay configuration? We will probably need a few more keys in there in the future for surfacing uncaught errors and whatnot.

jacob-ebey (changes requested):

packages/remix-react/components.tsx
ikhattab: addressed in https://github.com/remix-run/remix/pull/664/commits/23a0df466a6ed23f515a4df2afdd173563e8562c will probably squash all before merging

packages/remix-dev/__tests__/readConfig-test.ts
ikhattab: not sure if I got this correctly, do you mean something like `devServerOptions` for example that combines `boadcastDelay`,  `port`, `shouldShowLoadingSpinner` options?

mcansh (approved):

packages/remix-dev/__tests__/readConfig-test.ts
garand: I wonder if we can take some naming inspiration from things like Vite also.

I would do something like this:

```
devServer: {
  rebuildingIndicator: true
}
```

ikhattab:
I made a few updates in https://github.com/remix-run/remix/pull/664/commits/cc8b0d2a21b7b78f2cab6561cc516e0661d46b86 to allow adding `devServer` key to `remix.config` that can configure a few options for now. e.g:

```js
devServer:  {
  broadcastDelay: 0,
  port: 8002,
  rebuildingIndicator: true,
}
```

I took the same approach as deprecating `browserBuildDirectory` added `@deprecated` flag and still alow `devServerPort` and `devServerBroadcastDelay` to be used.

jensmeindertsma (approved):

MichaelDeBoey:
@ikhattab I think the new `devServer` option together with deprecating the `devServerBroadcastDelay` & `devServerPort` options would best be done in a separate PR first

ikhattab:
@MichaelDeBoey good call 👍🏽  will work on it

jacob-ebey:
> @ikhattab I think the new `devServer` option together with deprecating the `devServerBroadcastDelay` & `devServerPort` options would best be done in a separate PR first

We also don't want a breaking change this early, so let's make sure the existing options are proxied onto the config with a console.warn of deprecation.

ikhattab:
@jacob-ebey @MichaelDeBoey I created a separate PR for refactoring devServer options in https://github.com/remix-run/remix/pull/923

MichaelDeBoey:
@ikhattab Since all changes from #923 are currently revoked, could you please revert them here too and just add the `devServerRebuildingIndicator` config + fix all conflicts?

kiliman:
The `<LiveReload/>` component is very simple. I think if you really want this feature, just create your own version of it and include it in your *root* route.

https://github.com/remix-run/remix/blob/63c5eff12661e1982a37cef8e1bc919523f93c78/packages/remix-react/components.tsx#L1375

MichaelDeBoey:
@ikhattab Since all changes from #923 are currently revoked, could you please revert them here too and just add the `devServerRebuildingIndicator` config + fix all conflicts?

ikhattab:
Hi @MichaelDeBoey, that'd be great. I'll resume working on this and probably will start with a new branch from `dev` with the same name and apply my changes to it

ikhattab:
updated, also updated the docs.

chaance:
I'm going to close this for now. I'm not convinced this is something we need to expand our config API for. As @kiliman points out, this is mostly achievable today by using your own `LiveReload` component. You don't get the rebuild hook, but esbuild is so fast I'm not sure that adds a ton of value.

If we do offer any sort of rebuild indicator, I suspect it will be built-in eventually. We have lots of ideas on how to improve the DX while running `remix dev`, and we have tasks in the backlog that may require a big overhaul of the `LiveReload` component. But we'd prefer to make it so freaking fast that you don't have to think about it!

PR #665, opened 2021-11-25
Fix PrefetchBehavior comment

andrelandgraf:
@ryanflorence: "In the spirit of “you own the network tab” we don’t prefetch by default"

Fixes code comment at type PrefetchBehavior to correctly describe the default behavior of the Remix Link and NavLink components.


ryanflorence:
Sorry for not getting to this sooner, looks like another commit changed the wording here. Thanks!

PR #666, opened 2021-11-25
A couple of minor typos I noticed when going through the jokes tutorial

Matthew-Burfield:
I'm a bit nervous about creating a PR because I've never done it before, and I think someone said the docs are auto generated on a build? So not sure if this will even be helpful, but figured it can't hurt..

The update on line 2579 does get applied in the tutorial further down, but not until the next section in "User Registration". I'd recommend moving it to here because after finishing the "Build The Login Form" section, it left me with a login form that wouldn't actually go anywhere which I thought was an error in something I had done, and I didn't realise I had to continue to the next section to fix it.

The other typos I think are self explanatory, mostly just making sure everything matches up/is consistent.

If you're interested at all in overall feedback of the tutorial I have to say it is really well done! I found all the concepts were really well explained and I had no troubles moving through the tutorial. I especially love how all the code snippets are hidden behind the "click to expand" sections.


kentcdodds (approved):
Super! Thanks!

PR #670, opened 2021-11-26
docs: fix missing link and backticks in Conventions/link

plastic041:
1. Added missing mdn reference link of [`Link` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) in `Coventions/links/HtmlLinkDescriptor`.
2. Fixed `Coventions/links/HtmlMetaDescriptor` not wrapped with backticks.


kentcdodds (approved):
Thanks!

Issue #677, opened 2021-11-26
[Bug]: cache-control to forever not being set for build files in template created with express deployment


airondumael:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [X] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.0.5

### Steps to Reproduce

- run `npx create-remix@latest`
- choose express server as deployment 
- run the app
- see network build files are cache for 1hr instead of "forever" as intended in the code. 

Built files have fingerprints so it's better to just aggressively cache them

### Expected Behavior

Cache the built files for longer time

### Actual Behavior

Cached for 1 hour only because it conflicts with another express static route

<img width="1086" alt="image" src="https://user-images.githubusercontent.com/4447799/143587389-4cd798bd-16b8-423c-b7a9-95453f794ed8.png">

machour:
Fixed by https://github.com/remix-run/remix/pull/1861, please reopen if I'm wrong.

PR #678, opened 2021-11-26
fix caching on express server deployment

airondumael:
resolves:  #677

cache remix built files for longer time
image

### DIFF
--- a/packages/create-remix/templates/express/server/index.js
+++ b/packages/create-remix/templates/express/server/index.js
@@ -44,7 +44,7 @@ function purgeRequireCache() {
   }
 }
 
-let remixBuildPath = path.join(process.cwd(), 'public', 'build');
+const remixBuildPath = path.join(process.cwd(), 'public', 'build');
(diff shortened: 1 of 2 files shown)


packages/create-remix/templates/express/server/index.js
```diff
@@ -46,3 +43,11 @@ function purgeRequireCache() {
     }
   }
 }
+
+let remixBuildPath = path.join(process.cwd(), 'public', 'build');
```
MichaelDeBoey, line 47: As of #809, we're using `const` in the docs

```suggestion
const remixBuildPath = path.join(process.cwd(), 'public', 'build');
```

MichaelDeBoey (changes requested):

packages/create-remix/templates/express/server/index.js
airondumael: updated.

ryanflorence:

#1861

Issue #673, opened 2021-11-26
[Bug]: Jokes app returns password in pages


gaetansnl:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

latest

### Steps to Reproduce

Login to https://remix-jokes.lol/jokes, open dev tools, reload the page

### Expected Behavior

No password in the page

### Actual Behavior

Password appear in the page
![IXwkhWosorQLWJnkUS1vL40yP](https://user-images.githubusercontent.com/8348807/143545508-a423abb2-01e9-40d8-8dc0-25742c3d43d2.png)

matmilbury:
What does that mean that passwords appear 'in the page'? Where exactly?

gaetansnl:
It appears in the page source
![image](https://user-images.githubusercontent.com/8348807/143549995-2779efed-760b-4f8e-b543-a72f42c8381b.png)

airondumael:
This happens because the whole user data is being return in the loader function. Can be fix by limiting the data in the loader.  Created a [PR](#679) for it.

ryanflorence:
thanks, follow #679 for this :)

PR #679, opened 2021-11-26
docs(jokes): fix passwordHash included in body on /jokes/* response in

airondumael:
resolves: #673

Jokes example returns `passwordHash` in body. 
this happens because the whole user data was returned in the loader function.

image

### DIFF
--- a/examples/jokes/app/routes/jokes.tsx
+++ b/examples/jokes/app/routes/jokes.tsx
@@ -2,12 +2,11 @@ import type { LoaderFunction, LinksFunction } from "remix";
 import { Form } from "remix";
 import { Outlet, useLoaderData, Link } from "remix";
 import { db } from "~/utils/db.server";
-import type { User } from "@prisma/client";
 import { getUser } from "~/utils/session.server";
 import stylesUrl from "../styles/jokes.css";
 
 type LoaderData = {
-  user: User | null;
+  user: Awaited<ReturnType<typeof getUser>>;
   jokeListItems: Array<{ id: string; name: string }>;
 };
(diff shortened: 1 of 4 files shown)


examples/jokes/app/routes/jokes.tsx
```diff
@@ -21,7 +21,10 @@ export let loader: LoaderFunction = async ({ request }) => {
 
   let data: LoaderData = {
     jokeListItems,
-    user,
+    user: user && {
+      id: user.id,
+      username: user.username,
+    },
```
Dev-CasperTheGhost, line 27: The `getUser` function could possibly use [Prisma's `select`](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#select) property:


```js
async function getUser() {
  return db.user.findUnique({
    where: { /* ... */ },
    select: {
      id: true,
      username: true
    }
  })
}
```

airondumael: Yes that will also fix it but will limit the usage of getUser.

Dev-CasperTheGhost: true! Although since it's only 1 property of the user not being selected this could be a parameter in the function, but this might make the function too complex for a simple example 😅.

ryanflorence:
@kentcdodds let me know if the changes look good. I didn't run the app but tried to verify all the spots in the tutorial this affected. Please verify and merge :)

kentcdodds (approved):
Thanks a bunch for this fix! Silly of me to miss it 😅

PR #691, opened 2021-11-26
Update external links to open in new tabs.

joshcawthorne:
A minor annoyance when using the starter kit the first time, clicking on links to guides etc replaced the starter. Makes sense to update these to open in a new tab :)


favealo (approved):

ryanflorence:
the starter template changed significantly, thanks for sending this though!

PR #694, opened 2021-11-27
Add option to change routes directory

mjr:
Add property `routesDirectory` in `remix.config.js` to allow change routes directory.

```es6
module.exports = {
  appDirectory: 'src',
  routesDirectory: 'pages',
  browserBuildDirectory: 'public/build',
  publicPath: '/build/',
  serverBuildDirectory: 'build',
  devServerPort: 8002,
};
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -16,3 +16,4 @@
 - msutkowski
 - ryanflorence
 - ascorbic
+- mjr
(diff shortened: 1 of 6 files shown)


ryanflorence:
Thanks for taking the time to add this but we would like to keep it the way it is :)

PR #695, opened 2021-11-27
feat(config): add API to modify esbuild config before it is used

nicksrandall:
This API borrows inspiration from the Next.js config api allowing users to modify the esbuild config for the various targets before they are used to build the application.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -16,3 +16,4 @@
 - msutkowski
 - ryanflorence
 - ascorbic
+- nicksrandall
--- a/contributors.yml
+++ b/contributors.yml
@@ -15,3 +15,4 @@
 - morinokami
 - msutkowski
 - ryanflorence
+- ascorbic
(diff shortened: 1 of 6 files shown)


jongbelegen:
This change in config api would help me a lot as well.
I'm currently trying to use the official Postgres npm package "pg".

Esbuild gives me the following error:
```
 > node_modules/pg/lib/native/client.js:4:21: error: Could not resolve "pg-native" (mark it as external to exclude it from the bundle, or surround it with try/catch to handle the failure at run-time)
    4 │ var Native = require('pg-native')
      ╵                      ~~~~~~~~~~~
```

This would make me able to use the external config: https://esbuild.github.io/api/#external

nicksrandall:
Another immediate use case for this is to support the "worker" package exports conditional when bundling for cloudfare workers. Same thing applies for bundling for Deno as well. 

We need a way to tell esbuild (via the conditionals config option) to prefer builds for these environments over the default.

ryanflorence:
Thanks for taking the time to send this PR, but please see https://github.com/remix-run/remix/pull/1529#issuecomment-1035696128

PR #702, opened 2021-11-27
Update params demo page to break on switch case

rowinbot:
Now the right message should be set inside the switch case 💯


favealo (approved):
🤝

nimaa77:
LOL this took me ten minutes to find out what's wrong.

PR #711, opened 2021-11-27
Fix broken links in the default template

pbteja1998:


MichaelDeBoey (approved):

ryanflorence:
we've changed the template, but thank you for the help!

PR #715, opened 2021-11-27
Fix Variable Used in Conventions.md Code Example

mpetrus001:
The "Throwing Responses in Loaders" code example appears to use an undefined variable in the CatchBoundary function. The return value of useCatch() is defined as "let caught", but the JSX returned by the boundary tries to use "invoiceCatch" as the variable; "invoiceCatch" does not appear to be defined in the code example. The commit replaces "invoiceCatch" with "caught".

I've also added myself to the Contributors.yml as has been requested by others.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -15,3 +15,4 @@
 - morinokami
 - msutkowski
 - ryanflorence
+- mpetrus001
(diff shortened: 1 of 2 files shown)


ryanflorence:
Sorry for not getting to this sooner but it looks like another commit already fixed this. Thanks for taking the time to send the PR!

PR #720, opened 2021-11-27
Fix React JSX Transform Warnings in VS Code

AHBruns:
To use the React JSX transform with VS Code this must be in the jsconfig. Got warnings/errors while following the blog tutorial for the first time without this.


kentcdodds (approved):
Thanks! Since this is in the templates, I'll put it in `dev`. Thanks!

PR #721, opened 2021-11-27
Fix issue in accessibility Doc page

josh-degraw:
The `NavLink` link wasn't escaped properly in [this section](https://remix.run/docs/en/v1/guides/accessibility#links), so it wasn't showing it in the rendered Doc.


ryanflorence:
Sorry for not getting to this sooner, but we got this in a different commit. Thank you!

PR #734, opened 2021-11-27
Update spelling of client-side and server-side

zachdtaylor:
I noticed that the spelling of "client-side" and "server-side" was inconsistent across pages in the docs. I did some research and I think the correct way to spell it is "client side" if the phrase is being used as a noun, and "client-side" if it is used as an adjective (ditto for "server-side").


docs/guides/data-writes.md
```diff
@@ -191,13 +191,13 @@ export default function NewProject() {
...
 
 Of course, we started complicating things to try to create better user experiences than the default browser behavior. Keep going, we'll get there, but we won’t have to change any of the code we've already written to get the core functionality.
 
 ### Form Validation
 
-It's common to validate forms both clientside and serverside. It's also (unfortunately) common to only validate clientside, which leads to various issues with your data that we don't have time to get into right now. Point is, if your validating in only one place, do it on the server. You find with Remix that's the only place you care to anymore (the less you send to the browser the better!).
+It's common to validate forms both client side and server side. It's also (unfortunately) common to only validate client side, which leads to various issues with your data that we don't have time to get into right now. Point is, if your validating in only one place, do it on the server. You find with Remix that's the only place you care to anymore (the less you send to the browser the better!).
```
kentcdodds, line 200: Guessing this was an oversight 😅
```suggestion
It's common to validate forms both client-side and server-side. It's also (unfortunately) common to only validate client-side, which leads to various issues with your data that we don't have time to get into right now. Point is, if your validating in only one place, do it on the server. You find with Remix that's the only place you care to anymore (the less you send to the browser the better!).
```

kentcdodds (approved):
Super, thanks!

docs/guides/data-writes.md
zachdtaylor: Ah, yeah, thanks!

PR #736, opened 2021-11-27
docs: fix typo

drazik:
Example code was relying on a non existing `prisma` variable. It was actually intended to reference the `db` variable.


kentcdodds:
Thanks for this. I merged another PR that fixed this earlier before I noticed your PR. Sorry about that. Thanks anyway!

PR #738, opened 2021-11-28
fix: remove process.env check for LiveReload from docs, examples and templates

ZiiMakc:
When using:
```
{process.env.NODE_ENV === "development" ? (
   <LiveReload />
) : null}
```
in `root` component as stated in docs or by using express default example, browser will show multiple `Uncaught ReferenceError: process is not defined` errors caught by `ErrorBoundary`.

This check is already included in `LiveReload` [component](https://github.com/remix-run/remix/blob/c56761e09d1aeb27c4d3d5374759fd1bb5f79dcf/packages/remix-react/components.tsx) itself `if (process.env.NODE_ENV !== "development") return null;` so it's not necessary to duplicate it. Removing `process...` ckeck will remove this errors from appearing. 

Only thing I'm unsure of, is why there is no errors from `process...` usage inside `LiveReload` component.

chrome_7SlCkQVedg

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -16,3 +16,4 @@
 - msutkowski
 - ryanflorence
 - ascorbic
+- ZiiMakc
--- a/contributors.yml
+++ b/contributors.yml
@@ -15,3 +15,4 @@
 - morinokami
 - msutkowski
 - ryanflorence
+- ascorbic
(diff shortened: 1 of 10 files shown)


MichaelDeBoey:
Duplicate of #1352

PR #742, opened 2021-11-28
fix(TS): EntryRouteModule default export optional

kentcdodds:
Because of Resource Routes, the `EntryRouteModule`'s `default` property is optional.

### DIFF
--- a/packages/remix-server-runtime/entry.ts
+++ b/packages/remix-server-runtime/entry.ts
@@ -7,7 +7,11 @@ import type {
 } from "./routes";
 import type { RouteData } from "./routeData";
 import type { RouteMatch } from "./routeMatching";
-import type { RouteModules, EntryRouteModule } from "./routeModules";
+import type {
+  RouteModules,
+  EntryRouteModule,
+  ServerRouteModule
+} from "./routeModules";
 
 export interface EntryContext {
   componentDidCatchEmulator: ComponentDidCatchEmulator;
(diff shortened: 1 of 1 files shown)


kentcdodds:
I made this edit in GitHub, so I'm not sure of the impact on everything else here. Just figured this was more helpful than an issue (though maybe not quite as helpful as a more legit PR 😅)

kentcdodds:
Whoops. jk. I just realized that there's already a distinction between `ServerRouteModule` and `EntryRouteModule`. I'll have to figure out what I have wrong in my own code that triggered this error. It will probably be a different fix.

packages/remix-server-runtime/entry.ts
```diff
@@ -7,15 +7,19 @@ import type {
...
+} from "./routeModules";
 
 export interface EntryContext {
   componentDidCatchEmulator: ComponentDidCatchEmulator;
   manifest: AssetsManifest;
   matches: RouteMatch<EntryRoute>[];
   routeData: RouteData;
   actionData?: RouteData;
-  routeModules: RouteModules<EntryRouteModule>;
+  routeModules: RouteModules<EntryRouteModule | ServerRouteModule>;
```
kentcdodds, line 22: Just realized that a `ServerRouteModule` extends `EntryRouteModule` so this isn't quite right. We probably want to have a type definition for a ResourceRouteModule or something 🤔

kentcdodds:
If this comes up again I'll look at it.

PR #743, opened 2021-11-28
fix: null for body to conform to standards

jacob-ebey:
fix: pass full request to handleDataRequest, not stripped


kentcdodds (approved):
I don't know anything about this code yet, but I don't see anything concerning 😁

MichaelDeBoey (approved):

PR #744, opened 2021-11-28
chore(docs): fix typo

stardustrain:
Fix typo in API -> convention documetation.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -46,3 +46,4 @@
 - ascorbic
 - IAmLuisJ
 - weavdale
+- stardustrain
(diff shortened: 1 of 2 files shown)


kentcdodds:
@stardustrain this PR has an additional commit in it that I don't think you intended to include. Do you want to fix that/make a new one?

stardustrain:
> @stardustrain this PR has an additional commit in it that I don't think you intended to include. Do you want to fix that/make a new one?

@kentcdodds Yes you are right. I will be fix this PR soon.

stardustrain:
@kentcdodds fix at 0d87b45

kentcdodds (commented):
Ah! Thanks for fixing it. Now that I see what change you made, in afraid to tell you I already merged another PR that fixed this. Thanks anyway.

PR #748, opened 2021-11-28
docs: fix type annotation in jokes tutorial

drazik:
The `useActionData` takes a type parameter `T` and returns something of type `T | undefined`, so we don't need to specify `ActionData | undefined` in the type parameter.


kentcdodds:
Thanks for this, but I've merged another PR that fixed this. Cheers!

PR #755, opened 2021-11-28
fix(docs): fix docs use title 

veritem:
use title variable in error handling section of the tutorial


kentcdodds (approved):
Sweet, thanks!

Issue #760, opened 2021-11-28
[Feature]: Add `.env` to `.gitignore` in Remix  App Server template


JasonEtco:
### What is the new or updated feature that you are suggesting?

Hello! In trying out Remix and following [this doc on setting up a database](https://remix.run/docs/en/v1/tutorials/jokes#connect-to-the-database), I realized that the `.env` file is not automatically included in the `.gitignore` file generated by the **Remix App Server** template. I think it should be, since `.env` is recognized as a sensitive file, intended for storing development secrets. After adding some GitHub App credentials (for user authentication), I almost committed those secrets 😱 

I could not find the source for the Remix App Server so I'm opening an issue, happy to open a PR to do this if you'll point me in the right direction. Also not sure if this is a bug or a feature 😅 feel free to re-classify it as needed 👍 

### Why should this feature be included?

Noted above, but: `.env` files are typically used to store sensitive credentials, and _not_ including it in `.gitignore` might give folks the wrong impression that it was included already, since Remix uses `.env` automatically.

fre-ben:
Noticed this as well while doing the tutorial today - should be an easy, userfriendly fix :)
Good idea

IAmLuisJ:
Added this to the dev branch, it should merge with main on the next version update. Sorry, I think I should have said "fixed #768 " for it to close automatically when it merges to main.

JasonEtco:
Oh awesome, thanks for doing that @IAmLuisJ!

PR #768, opened 2021-11-29
added .env to gitignore

IAmLuisJ:
Updated gitignore template files to include .env files by default so they are not committed to a repo. Adds feature request #760


kentcdodds (approved):
Good call. Thanks!

PR #770, opened 2021-11-29
chore(docs): Update Jokes tutorial to use the latest project structure

mattstobbs:
The starter project structure has been updated so that it no longer includes `utils.server.tsx` and `remix.css` so I've updated the tutorial to remove those files.


MichaelDeBoey (approved):

kentcdodds (approved):
Thanks!

PR #785, opened 2021-11-29
chore: update dependencies to latest minor

MichaelDeBoey:


packages/create-remix/templates/arc/package.json
```diff
@@ -6,7 +6,8 @@
     "start": "arc sandbox"
   },
   "dependencies": {
+    "@architect/architect": "^8.7.4",
```
MichaelDeBoey, line 9: `@remix-run/architect` has this as a `peerDependency`

packages/create-remix/templates/vercel/package.json
```diff
@@ -6,6 +6,7 @@
   },
   "dependencies": {
     "@remix-run/serve": "*",
-    "@remix-run/vercel": "*"
+    "@remix-run/vercel": "*",
+    "@vercel/node": "^1.12.1"
```
MichaelDeBoey, line 10: `@remix-run/vercel ` has this as a `peerDependency`

packages/remix-cloudflare-workers/package.json
```diff
@@ -12,7 +12,7 @@
     "url": "https://github.com/remix-run/remix/issues"
   },
   "dependencies": {
-    "@cloudflare/kv-asset-handler": "^0.1.3",
+    "@cloudflare/kv-asset-handler": "^0.2.0",
     "@remix-run/server-runtime": "1.0.6"
```
MichaelDeBoey, line 16: `@remix-run/server-runtime` has `react` & `react-dom` as `peerDependency`.
Should we add them as dependencies or peerDependencies in this package too?

MichaelDeBoey:
@kentcdodds Is there something I should still do to get this one merged?

kentcdodds (changes requested):
Thanks for this, but we're a bit concerned about this in the `npm-check-updates` docs:

> npm-check-updates upgrades your package.json dependencies to the latest versions, ignoring specified versions

We don't want to ignore specified versions. So I don't think we're going to merge this. We'll update deps manually on an as-needed basis. Thanks for taking the time anyway!

MichaelDeBoey:
@kentcdodds Since I used the `--minor` flag, these weren't ignored

I can remove the script if you want btw.

kentcdodds:
I think we'll come back to this when we're more confident in our testing infra. Thank you for writing the script. It may be needed in the future! Thanks!

PR #789, opened 2021-11-29
docs: .server filename in constraints

RossJHagan:
The `.server` naming convention is documented in the jokes app tutorial.  It seems sensible to also put this information in module constraints file convention when talking about patterns to isolate server code.


MichaelDeBoey (approved):

kentcdodds (approved):
Thanks!

PR #792, opened 2021-11-30
Do not use Prisma instance, use db instance instead

d4vsanchez:
In this part of the J🤪kes example we're using a `prisma` instance instead of using the `db`. This is a bit weird in this place because we're not creating the Prisma instance anywhere in this code, if you take a look at the previous examples, we created the Prisma instance inside the `db.server.ts` file and is being exported there.


kentcdodds:
Hey @d4vsanchez! Sorry but I missed your PR and merged someone else's first. Thanks anyway!

Issue #793, opened 2021-11-30
[Bug]: minor docs issue


Graham42:
### Which Remix packages are impacted?

N/A this is a docs issue

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

v1

### Steps to Reproduce

Go to the [docs page for the developer blog](https://remix.run/docs/en/v1/tutorials/blog#creating-the-project) and there's a confusing sentence that I think has an extra word.

> There's a bit going on in the file, find into the Layout component and right after the link to "Home", add a new link to "/posts"

### Expected Behavior

I think the wording could be clearer here. 
- "the file" doesn't have any context yet, so it's not clear which file is being referred to. 
- "find into the `Layout` component" sounds like "into" is an extra word.

Possible update:

```diff
- There's a bit going on in the file, find into the `Layout` component and right after the link to "Home", add a new link to "/posts"
+ There's a bit going on in the `app/root.tsx` file. Find the `Layout` component, and right after the link to "Home", add a new link to "/posts"
```

### Actual Behavior

![The "Your First Route" section of the blog tutorial with the problematic sentence](https://user-images.githubusercontent.com/2476733/143968876-188d1927-6c98-4640-82bf-5f2fed1ea489.png)

Graham42:
Also I just found the docs are on the `dev` branch, so I'm going to open a PR for this ✋

PR #794, opened 2021-11-30
docs: clarify instruction in blog tutorial

Graham42:
"the file" didn't have any context yet, so it was not clear which file
was being referred to.

In the phrase "find into the Layout component" it appears "into" is an
extra word.

fixes: https://github.com/remix-run/remix/issues/793


kentcdodds (approved):
Thanks!

Issue #774, opened 2021-11-29
[Bug]: prefetch not working properly, mismatched prefetched URL and clicked URL 


airondumael:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [X] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.0.5

### Steps to Reproduce

Prefetch a link with search params. 

```
<Link to="/search?q=search&p=2" prefetch="render">Link</Link>
```

The URL loaded during prefetch have the `_data` at the end but when you click the link it will be at the start of search params causing the prefetched cache to not hit. 

### Expected Behavior

The URLs for prefetch and the actual click should be consistent for the prefetch to work. 

<img width="1075" alt="image" src="https://user-images.githubusercontent.com/4447799/143908270-7e33ef0e-6fd8-490d-9369-8190aa741ad4.png">

Was able to get the expected behavior using this patch but there is also an existing issue related to prefetched url not matching clicked url [here](#484) that may have a different issue. Maybe this needs to be fix somewhere else. 

```
diff --git a/node_modules/@remix-run/react/browser/links.js b/node_modules/@remix-run/react/browser/links.js
index 967ead0..9e190fd 100644
--- a/node_modules/@remix-run/react/browser/links.js
+++ b/node_modules/@remix-run/react/browser/links.js
@@ -153,8 +153,12 @@ function getDataLinkHrefs(page, matches, manifest) {
       search
     } = path;
     let searchParams = new URLSearchParams(search);
-    searchParams.append("_data", match.route.id);
-    return `${pathname}?${searchParams}`;
+    let dataSearchParams = new URLSearchParams();
+    dataSearchParams.append("_data", match.route.id);
+    for (let [key, value] of searchParams) {
+      dataSearchParams.append(key, value);
+    };
+    return `${pathname}?${dataSearchParams}`;
   }));
 }
 function getModuleLinkHrefs(matches, manifestPatch) {

```

### Actual Behavior

URL are not matching causing the data to be loaded twice and not gain the benefit of prefetch.

[...]

PR #797, opened 2021-11-30
fix prefetch url params sorting to improve caching

airondumael:
resolves: #774 

url params on prefetch and on fetcing data does not match because fetchData is sorting url params to make it more predictable (improve caching) to match this we also need to sort url params inside getDataLinkHrefs.

https://github.com/remix-run/remix/blob/main/packages/remix-react/data.ts#L39-L40


ryanflorence:
fixed by #1396. thanks!

PR #798, opened 2021-11-30
fix(docs): fix post page link href in admin page for blog tutorial

Hopsken:
In admin page, link href for post pages should be `/posts/${post.slug}` rather than `post.slug`


kentcdodds (commented):
Thanks! This link is actually correct. Thanks to nested routing you don't have to construct full URLs anymore 🎉 Relative links FTW!

Hopsken:
@kentcdodds Thanks for the reply. However, because of "Nested Routing", without the full URL, the link would be treated as `/admin/posts/${post.slug}`, which I believe is not implement.

kentcdodds:
Gotcha! Thanks for this!

dylancom:
@kentcdodds @Hopsken it seems this error was reintroduced again, since switching to the indie stack?
I just followed the blog tutorial and the posts inside the admin link to: "http://localhost:3000/posts/admin/my-first-post", which should be "http://localhost:3000/posts/my-first-post"

PR #802, opened 2021-11-30
Docs: show how to use tailwind @apply

fergusmeiklejohn:
I've changed the docs to show how to use the Tailwind `@apply` feature to generate custom classes. It needs the user to build tailwind in a slightly different way.


kentcdodds (approved):
Awesome 👍 Thank you.

PR #808, opened 2021-11-30
Added .DS_STORE to template .gitignores

nialldbarber:
Added `.DS_STORE` to all the relevant templates' `.gitignore`. This might need to changed to `**/.DS_Store` if they pop up in deeper directories, but hopefully not! 

This came about when I tried using the Vercel template and found quite a few `.DS_STORE`(s): https://github.com/nialldbarber/ndb-remix

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -32,3 +32,4 @@
 - tylerbrostrom
 - ascorbic
 - IAmLuisJ
+- nialldbarber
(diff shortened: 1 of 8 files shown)


airjp73:
Looks like this would address #811.

revelt:
The root-level gitignore `.DS_Store` entry will cover all sub-level ds_stores too. I checked on my monorepo, there's no need for `**/.DS_Store`. On other hand, this PR won't solve #811 but we're going to the right direction. 👍

PR #615, opened 2021-11-24
feat(deployment): add restify adapter

jenseng:
copy-pasta/tweaked from express adapter

testing notes:
☑️  standard adapter tests, plus some more stream ones
☑️  generated a restify-remix app via create-remix and manually verified

lmk if there's anything I should add :)

PR #813, opened 2021-11-30
feat(create-remix): add support for custom server templates

jenseng:
Give create-remix a "Custom..." deployment option, and a corresponding `--server-type` flag. If you select "Custom...", you can then provide a package id or path containing your custom server type's templates. The `--server-type` flag accepts package ids, paths, and built-ins (e.g. "remix" or "express"), and will bypass questions accordingly.

This makes it easier to use custom deployments/adapters for things that don't make the cut for first class support within remix (e.g. restify).

Also:
* Enhance the package.json merging to consider both the server and lang-specific ones
* Add a section to the docs around custom adapters/servers (both for authoring and around how the templating works)
* Expand cli tests and (hopefully) simplify test authoring

Points of Discussion:
1. Should we also move the built-in server templates into their respective adapter packages?
   1. 👍 Helps scope server/adapter-specific stuff to one place
   2. 👍 Makes the layout the same as custom server templates. Combined with 1), this makes it trivial to promote or demote server types
   3. 👍 Simplifies create-remix
   4. 👎 create-remix will need to install built-in adapters (and their deps) on demand just like custom server types, even though we really only need their templates. Alternatively we could bake those templates into create-remix at build time, but this somewhat negates 2 and 3.
   5. 👎 Not all server types have their own adapter packages (e.g. fly uses remix-serve)
2. Should we go further with the tests in this PR? The new functionality is generally covered, and we now test some basic file copying (package.json). But we still don't truly test end-to-end afaict (i.e. generate each server/deployment type, start it up, test a few routes). Would be great for catching regressions and ensuring all supported adapters work. Feels out of scope for this PR, but I could look at doing that as a separate PR.

Additional context:
* @ryanflorence "🔥'd" this idea on discord [here](https://discord.com/channels/770287896669978684/770287896669978687/914926217243803669)
* Pull request #615 (restify adapter PR, will be abandoned in favor of this)

### DIFF
--- a/docs/other-api/adapter.md
+++ b/docs/other-api/adapter.md
@@ -5,7 +5,7 @@ order: 2
 
 # Server Adapters
 
-Idiomatic Remix apps can generally be deployed anywhere because Remix adapt's the server's request/response to the [Web Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). It does this through adapters. We maintain a few adapters:
+Idiomatic Remix apps can generally be deployed anywhere because Remix adapts the server's request/response to the [Web Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). It does this through adapters. You can [build your own adapter](#custom-adapters), but we provide several for common deployment scenarios:
 
 - `@remix-run/express`
 - `@remix-run/architect`
@@ -139,7 +139,7 @@ import * as build from "../build";
 const handleRequest = createRequestHandler({ build });
 
 const handleEvent = async (event: FetchEvent) => {
-  const response = await handleAsset(event, build);
+  let response = await handleAsset(event, build);
 
   if (!response) {
     response = await handleRequest(event);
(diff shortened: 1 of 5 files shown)


docs/other-api/adapter.md
```diff
@@ -139,7 +139,7 @@ import * as build from "../build";
 const handleRequest = createRequestHandler({ build });
 
 const handleEvent = async (event: FetchEvent) => {
-  const response = await handleAsset(event, build);
+  let response = await handleAsset(event, build);
```
sergiodxa, line 142: Kent changed this to const today

packages/create-remix/cli.ts
```diff
@@ -63,27 +69,48 @@ async function run() {
...
+          "Where do you want to deploy? Choose Remix if you're unsure, it's easy to change deployment targets.",
+        loop: false,
+        choices: [
+          { name: "Remix App Server", value: "remix" },
+          { name: "Express Server", value: "express" },
+          { name: "Architect (AWS Lambda)", value: "arc" },
+          { name: "Fly.io", value: "fly" },
+          { name: "Netlify", value: "netlify" },
+          { name: "Vercel", value: "vercel" },
+          { name: "Cloudflare Workers", value: "cloudflare-workers" }
```
sergiodxa, line 101: Maybe add an Other option, if the user choose it ask for the adapter to use.

docs/other-api/adapter.md
jenseng: This one should still be `let`, it gets reassigned on line 145

packages/create-remix/cli.ts
jenseng: I like that idea, seems more user-friendly than command-line flag

sergiodxa: I think it should have both, the flag is useful if you already know it exists, the option helps for discoverability

docs/other-api/adapter.md
sergiodxa: Ahh ok, ignore the comment then

packages/create-remix/cli.ts
jenseng: Done

jenseng:
I'm still wishy washy on the naming, idk if `--server-type` is the best name for the flag... maybe `--server-template` or even `--template` would be better?

ryanflorence:
Sorry so long before providing feedback here.

We've been trying to figure out the whole deployment target/adapter/template/cli dance and aren't sure where we're going to land yet. This is great stuff though. I don't know when we'll figure out exactly what we want to do yet with it but here are a few things we've been planning.

- "Remix Stacks" - these are like our current templates but provide way more batteries included (sessions, database, prisma, github actions for auto deployment, etc.). We'll be focusing most of our energy here.
- Init templates from any random github repo. It's really just a clone but from a maintained list of repos in like `templates.yaml` or something in create-remix for visibility.

Do you want to trim down this PR to just the explanation on building your own adapter and we'll merge it?

Then when we have a clue about what we're doing w/ the CLI and templates we'll rope you back into it!

jenseng:
Sure, happy to trim this down to just the doc changes.

Any interest in a separate PR strictly to improve the create-remix tests? The current test is a bit basic and brittle IMO and is even skipped on main with a TODO comment :)

machour:
@jenseng Remix Stacks are now live, could you give them a look and turn your PR into a nice documentation addition? 🙏🏼

PR #816, opened 2021-12-01
chore: add format GitHub action

mcansh:
example repo: https://github.com/mcansh/test-repo

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -9,6 +9,10 @@ jobs:
   prettier:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        node-version: [14.x]
+
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.9.1
@@ -18,25 +22,23 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v2
 
-      - name: Setup Node
+      - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: ${{ matrix.node-version }}
 
-      - name: Install Dependencies
-        uses: jaid/action-npm-install@v1.2.1
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
 
-      - name: Run format script in package.json
-        run: npm run format --if-present
+      - name: Format
+        run: yarn format --if-present
 
-      - name: Setup Git User
+      - name: Commit
         run: |
           git config --local user.email "bot@remix.run"
           git config --local user.name "Remix Run Bot"
           git remote add github "https://$GITHUB_ACTOR:$GITHUB_TOKEN@github.com/$GITHUB_REPOSITORY.git"
 
-      - name: Commit and push formatting changes
-        run: |
           git add .
           if [ -z "$(git status --porcelain)" ]; then
             echo "💿 no formatting changed"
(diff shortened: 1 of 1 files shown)


.github/workflows/format.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Format
+        run: yarn format --if-present
```
mcansh, line 34: looks like yarn doesn't have a `--if-present` flag, so going to switch this to run with npm

FStell01 (commented):
?

.github/workflows/format.yml
```diff
@@ -0,0 +1,49 @@
...
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ github.token }}
```
MichaelDeBoey, line 20: `styfle/cancel-workflow-action`'s `access_token` is already defaulting to `github.token`
https://github.com/styfle/cancel-workflow-action/blob/dee858fff096b2cbcd4f845ce21871c50c5af452/action.yml#L15-L17

```suggestion
```

.github/workflows/format.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
```
MichaelDeBoey, line 28: No need to put this in a matrix, as it's only running on one version
```suggestion
      - name: Use Node.js
        uses: actions/setup-node@v2
        with:
          node-version: 14
```

.github/workflows/format.yml
```diff
@@ -0,0 +1,49 @@
...
+    branches:
+      - main
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x]
```
MichaelDeBoey, line 14:
```suggestion
```

MichaelDeBoey (changes requested):

.github/workflows/format.yml
mcansh: oh cool! i just used the example in the readme 🤷

mjackson (approved):

.github/workflows/format.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Format
+        run: npm run format --if-present
+
+      - name: Commit
+        run: |
+          git config --local user.email "bot@remix.run"
```
mjackson, line 32:
```suggestion
          git config --local user.email "github-actions@remix.run"
```

.github/workflows/format.yml
mcansh: i actually had with out a matrix, but i have no idea why i switched it use one haha

MichaelDeBoey: Haha those things happen 😛

PR #817, opened 2021-12-01
Add `react-router-dom` dependency to gists fixture

chaance:
This *technically* shouldn't be necessary, but since we don't actually install Remix in the gists fixture we don't get its dependencies, which means React Router was missing and all of our tests were failing. Tragic!


kentcdodds:
Thanks Chance!

PR #821, opened 2021-12-01
Fix typos

Notr0M:


kentcdodds:
Hi @Notr0M,

The correct directory is `build` not `server/build`. Thanks though :)

Notr0M:
Hi @kentcdodds 

We do not have any `build` directory  at the root level by running `npm run build`.

Thank you for your time.

kentcdodds:
It's possible you're using a different starter template that has configured the `build` directory to go somewhere else. But in the tutorial we use the Remix App Server and that puts the build output in `build` :)

PR #824, opened 2021-12-01
Remove comment from tsconfig file

### LINKED ISSUE #823: [Bug]: Don't put comments in generated tsconfig.json

- [ ] `remix` (Remix core)
- [X] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/vercel`

latest

### Steps to Reproduce

1. Generate a new project with: `npx create-remix@latest`
2. Choose `TypeScript`
3. Inspect the generated `tsconfig.json` file, it contains the following line:
    ```
    // Remix takes care of building everything in `remix build`.
    ```

### Expected Behavior

Frustratingly, it is not valid to include comments in JSON files, according to the JSON spec. Some tools do support comments in JSON, but others don't. For me the ALE syntax engine that I use with VIM was not able to parse the comment, and therefore just silently ignored the tsconfig file. Unfortunately to achieve interoperability, comments should not be used in JSON files. The expected behaviour is therefore for the generated remix template to not include comments in the JSON files it generates.

### Actual Behavior

The generated remix template DOES include comments in the `tsconfig.json` file it generates.

jnicklas:
Closes #823

Comments are not valid in JSON files according to the spec, and while some parsers do support comments, others don't and so Remix should not add any comments in the JSON files it generates.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -59,6 +59,7 @@
 - jesseflorig
 - joaosamouco
 - johannesbraeunig
+- jnicklas
 - juhanakristian
 - kalch
 - KenanYusuf
--- a/packages/create-remix/templates/_shared_ts/tsconfig.json
+++ b/packages/create-remix/templates/_shared_ts/tsconfig.json
@@ -13,8 +13,6 @@
     "paths": {
       "~/*": ["./app/*"]
     },
-
-    // Remix takes care of building everything in `remix build`.
     "noEmit": true
   }
 }


ryanflorence:
Thanks @jnicklas, but as discussed in the issue, we aren't setting the precedent here, TypeScript is. `tsconfig.json` is expected to be parsed as jsonc. Any tool trying to read a ts config should be aware of that.

PR #826, opened 2021-12-01
Remove id attribute from RouteChangeAnnouncement component

matmilbury:
https://github.com/remix-run/remix/discussions/763

tldr; the id attribute is not referenced anywhere, it seems it has no purpose.


ryanflorence:
we no longer have that in the template, but thank you!

PR #833, opened 2021-12-01
Add a tutorial for using SVGs as components.

mikeybinnswebdesign:
I've played around with this concept a bunch and this is the best method I've found:

Pros:
- It reduces all SVG imports to a single line, so if you call multiple SVGs, your imports don't get messy
- It allows you to use SVGs as components, which lets you add props for customisation

Cons:
- No automatic conversion of SVGs to JSX, requires updating some SVG attributes manually.
- No automatic optimisation of SVGs, SVGs should be run through SVGO beforehand for performance reasons.

Things I've also considered:

- Having an SVGs folder and importing individual TSX files (makes importing really messy)
- Pasting SVG code directly (makes code messier, prevents reuse of SVG code)
- Adding a build step for automatic compilation similar to create-react-app (increases build complexity, requires all SVGs go through the same optimisation steps where different steps are needed)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -23,6 +23,7 @@
 - mcansh
 - medayz
 - meetbryce
+- mikeybinnswebdesign
 - michaeldeboey
 - michaelfriedman
 - morinokami
--- a/docs/guides/svgs.md
+++ b/docs/guides/svgs.md
@@ -1,0 +1,59 @@
+---
+title: SVGs
+---
+# SVGs
+
+Adding SVGs as components can be achieved using the following steps:
+
+Create a file in your `app` folder called `svgs.{jsx,tsx}`
+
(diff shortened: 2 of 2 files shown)


ryanflorence:
Thanks @mikeybinnswebdesign but there are so many use cases for SVGs I'm not sure we can recommend a single approach.

Your method here repeats the SVGs in the SSR'd content. Imagine a list of 100 items that all have 5 SVGs in them. That's 500 repeated blocks of svg content. The most optimal method from a performance standpoint (not DX ofc) is to have basically a single svg file with several icons in it and then reference them by id where you want to use them. (I've been meaning to write about this).

mikeybinnswebdesign:
Makes sense :)

PR #840, opened 2021-12-02
feat: re-work server runtime internals

jacob-ebey:
goals:
- simplify catch / error tracking loop
- reduce internal state in function
- simplify types to follow code easier
- more comments
- solidify reload logic more

byproducts:
- userLoaderData can now be used in boundaries, with the catch it could be undefined. If your route throws, it will be undefined, but if a child route threw and that bubbled up, that route's loader data will be around if it was previously rendering.

### DIFF
--- a/fixtures/gists-app/app/root.jsx
+++ b/fixtures/gists-app/app/root.jsx
@@ -27,7 +27,7 @@ export function links() {
 
 export function loader({ request }) {
   return {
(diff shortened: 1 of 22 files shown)


mjackson (changes requested):
Large changes like this are really hard to review. would you mind splitting up this PR into smaller, more focused changes? I can't quite tell what your goals are.

jacob-ebey:
These changes will be broken out across multiple PR's.

PR #746, opened 2021-11-28
Check if the specified port is in use

Mahdi-Esbati:
In this pull-request, a known issue (#639) is fixed with the Remix dev-server, where it replaces the previous app using the specified port instead of checking the port before launch.

The new behavior is exactly like create-react-app which offers you to launch the app on a different port:

https://user-images.githubusercontent.com/17250923/143735338-2fcb3d43-e5ba-48a5-9291-36a75b7544fe.mp4

PR #844, opened 2021-12-02
feat(dev-server): add auto change dev server port

chenc041:
## Change file or folder
1. .gitignore
2. remix-dev

## What is the current behavior?
1. The .idea folder will submit
2. An error is thrown when the default port is occupied,

## What is the new behavior?
1. Ad idea folder to .gitignore
2. I modified the startup logic to increment the port number automatically when the default port is occupied to avoid throwing an error

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ yarn-error.log
 /fixtures/my-remix-app
 
 .eslintcache
+/.idea/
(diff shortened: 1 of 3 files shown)


chenc041:
> This issue has already been addressed in #746 waiting to be merged.

😂

PR #850, opened 2021-12-02
chore: fix params demo

dynamite-bud:
The basic example had issue with switch case and in the render method had a `<p>` tag inside a `<p>` tag.


kentcdodds (commented):
Thanks! I had to do a couple git gymnastics to fix something I messed up...

But now that I'm actually look at this change, I'm not sure I understand why it's necessary 😅 What's this change doing?

PR #855, opened 2021-12-02
chore(docs): fix typo

francisudeji:
Kent tweeted [this](https://twitter.com/kentcdodds/status/1466409430694576128?t=DiZ4XoBq09mr1I2KDScNaw&s=19) in response to a tweet this evening and while reading I noticed a reference to `value.showBanner` rather than `cookie.showBanner` in the loader function of an example, so I fixed it!

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -37,3 +37,4 @@
 - VictorPeralta
 - zachdtaylor
 - zainfathoni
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Sweet! Thanks!

PR #860, opened 2021-12-02
feat: surface errors from fetch submissions

jacob-ebey:
when an error surfaces from the transition manager it's not thrown from the react render stack, rather set in the app state and surfaced through props


jacob-ebey:
Address part of https://github.com/remix-run/remix/issues/599

PR #861, opened 2021-12-02
chore: add node version to issue report

jacob-ebey:


kentcdodds (approved):
Yup, this is a good idea I think 👍

PR #870, opened 2021-12-03
Docs: Explain how to handle different request methods with resource routes

davecalnan:
This PR:
- Adds an explanation to the resource routes doc explaining to use `loader` for `GET` requests and `action` for `POST`, `PUT`, `PATCH` and `DELETE` requests.

I was a bit confused and asked in the Discord and ran my own test. Hopefully this will short-circuit that for future people.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -6,6 +6,7 @@
 - BasixKOR
 - chaance
 - coryhouse
+- davecalnan
 - eps1lon
 - evanwinter
 - francisudeji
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Thanks for this!

This applies to all routes, not just resource routes. But this is a pretty good place to talk about this anyway and hopefully anyone with this question will just use the algolia search and land here anyway. Thanks!

PR #871, opened 2021-12-03
feat(dev): get available port if not available

### LINKED ISSUE #863: [Feature]: Graceful port suggestion if 3000 is taken

Suggest a new port if 3000 is being used		

Currently it fails and leaks a fairly comment error to the end user

CleanShot 2021-12-02 at 16 24 42@2x
s

mcansh:
uses [get-port](https://github.com/sindresorhus/get-port) instead of incrementing by 1 like #844 and #746

also applies the port checking to the `devServerPort`

closes #863, #844, and #746

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -7,6 +7,7 @@ import WebSocket from "ws";
 import type { Server } from "http";
 import type * as Express from "express";
 import type { createApp as createAppType } from "@remix-run/serve";
+import getPort from "get-port";
 
 import { BuildMode, isBuildMode } from "../build";
 import * as compiler from "../compiler";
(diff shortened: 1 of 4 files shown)


tascord (approved):
Looks good 👍

ryanflorence:
Let's rebase this on current dev, the config changed a bit.

ryanflorence:
Also, can LiveReload do the same?

mcansh:
> Also, can LiveReload do the same?

yeah, but we'll need to tell LiveReload in root.tsx about the new port

sergiodxa:
Shouldn’t https://github.com/remix-run/remix/blob/1fd70960e4d88740df5bf407a6ba2cd2b9549459/packages/remix-serve/cli.ts#L6 do the same?

kentcdodds (commented):
This would be quite nice! Maybe you could add https://github.com/remix-run/remix/issues/1277 to this one? Otherwise the PR is only marginally useful.

kentcdodds (approved):
This looks good to me. Would love to get this combined with #1352

kentcdodds (approved):
LGTM 👍

PR #880, opened 2021-12-04
Fix some code highlights in blog tutorial

Graham42:
For the code highlight for `useActionData` the import statement is highlighted, but the highlight is missing where the new hook is used.

For the the code for the dynamic post route (`routes/post/$slug`), in the previous sample code for the `PostSlug` component, line 14 was `const slug = //...` and changes in this snippet to `const post = //...` but was missing the highlight.


ryanflorence:

PR #882, opened 2021-12-04
feat: add support for pnpm

iamsahebgiri:
Now when `pnpx create-remix@latest` is called `pnpm install` is shown instead of `npm install`.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -13,6 +13,7 @@
 - gon250
 - goncy
 - graham42
+- iamsahebgiri
 - ianduvall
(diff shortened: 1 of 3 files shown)


ryanflorence:
I have no experience with pnpm, anybody else want to verify this code? Also does pnpm work with Remix?

We also added that question "want me to run npm install?" so that you can say "no" and do whatever you want, like use pnpm.

alber70g:
Does this also make the postinstall work with typings for typescript?

PR #883, opened 2021-12-04
docs: fix example command in envvars.md

kt3k:
`npm add` -> `npm install`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -30,6 +30,7 @@
 - kentcdodds
 - kgregory
 - kimdontdoit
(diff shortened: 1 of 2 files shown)


kentcdodds (commented):
`npm add` actually does work (it's an alias for the `install` commend) 😄  so we'll keep this as-is :)

kentcdodds:
Thanks anyway!

kt3k:
Oh, I didn't know that alias.. Thanks for your review!

PR #890, opened 2021-12-04
docs: Fix some broken links

Graham42:
This started as fixing one link, and then I found several more. I've committed related bits separately to hopefully make review easier, but if you prefer to squash these all to a general "fix links" go ahead. 🙂


kentcdodds (approved):
Thank you!

PR #891, opened 2021-12-04
docs: move 'HtmlMetaDescriptor' to be under 'meta' instead of 'links'

Graham42:
I'm pretty sure the section about `HtmlMetaDescriptor`  is meant to be under `meta` instead of under `links`.


kentcdodds (approved):
Ha! Not sure how that happened! Thanks 😊

PR #892, opened 2021-12-04
Add `async` to action function example

derekr:
Just a small tweak to make copying/pasting the example work.


kentcdodds (approved):
Thanks!

PR #895, opened 2021-12-05
fix: jokes tutorial actions should return 400 status for errors

Graham42:
## Why make this change?

When a request is sent to the server and there's something invalid about
the request, the server should return a 4xx status to accurately
represent the response.

We could have all the error cases use the `json` function directly, but
then we loose the advantage of the helpful typechecking `ActionData`
provides.

```ts
  //...
  return json({ wrongActionDataErrorKey: "oops!" }, { status: 400 }));
```

For this reason, I've introduced the `badRequest` helper function so
that we can still get typechecking for `ActionFunction` returns, while
returning the accurate status in the Response.

```ts
const badRequest = (data: ActionData) => json(data, { status: 400 });

  //...
  // and then TypeScript will alert us
  return badRequest({ wrongActionDataErrorKey: "oops!" });
  //                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

## Notes for the reviewer

I'm pretty sure I've updated all the line highlights correctly, but if you have any way of rendering a preview of the docs, that would help to increase confidence.

Thanks for your time reviewing this! 🙏

### DIFF
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -2009,9 +2009,9 @@ But if there's an error, you can return an object with the error messages and th
 
 <summary>app/routes/jokes/new.tsx</summary>
 
-```tsx filename=app/routes/jokes/new.tsx lines=[2,5-9,11-15,17-27,31,39,42-45,47-49,56,67,69-77,80-88,94,96-104,107-115]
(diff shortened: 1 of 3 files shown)


kentcdodds (approved):
Super 👍 Thank you for this improvement 👏

PR #901, opened 2021-12-05
Use inline sourcemaps for server build

### LINKED ISSUE #367: Incorrect sourceMappingURL added to generated server bundle prevents debugging
Updated my gist with patch to use `sourcemap: 'inline'`. This solves the issue.

Here's the gist:
https://gist.github.com/kiliman/a9d7c874af03369a1d105a92560d89e9

zigomir:
Fixes #367

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -46,3 +46,4 @@
 - ascorbic
 - IAmLuisJ
 - weavdale
+- zigomir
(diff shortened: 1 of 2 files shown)


jacob-ebey:
I'd actually like to make this configurable. The issue comes when deploying places like CF that have a limit on script sizes. inline source maps will almost double the files size a lot of the time.

zigomir:
When you run `build` command, `sourcemap` is set to `false` so this shouldn't affect prod build script size.

ryanflorence:
Thanks, we actually started on this before seeing your PR 🥲 #1087

zigomir:
No problem :) Happy to see it landed and debugging will be nicer.

PR #585, opened 2021-11-24
chore: Fix blog tutorial example filename typo

KenanYusuf:
There are references to `90s-mixtape` as well as `90s-mix-cdr` which could cause confusion.

PR #909, opened 2021-12-05
fix(doc): typo 90s-mix-cdr

ahmadrasyidsalim:


chaance:
Fixed in https://github.com/remix-run/remix/pull/585

Issue #839, opened 2021-12-02
[Bug]: createCloudflareKVSessionStorage isn't setting `expires` correctly


kentcdodds:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [X] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

Every

### Steps to Reproduce

```ts
export const kvSessionStorage = createCloudflareKVSessionStorage({
  kv: KV_SESSION,
  cookie: {
    name: "__session",
    secure: NODE_ENV === "production",
    secrets: [SESSION_SECRET],
    sameSite: "lax",
    path: "/",
    maxAge: 60 * 60 * 24 * 7, // sessions expire after a week <-- I'm setting the maxAge
    httpOnly: true,
  },
});
```

### Expected Behavior

I expect the `maxAge` to be able to control the `expires` option in KV.

### Actual Behavior

I get the following error:

```
Error: KV PUT failed: 400 Invalid expiration of 632. Please specify integer greater than the current number of seconds since the UNIX epoch.
```

Right now, [the code](https://github.com/remix-run/remix/blob/a317b79132c8854f11e6ce334803e90fafbfc2e7/packages/remix-cloudflare-workers/sessions/cloudflareKVSessionStorage.ts#L49) sets the expires value based on what it gets in the `expires` parameter. So maybe the problem actually lies somewhere else. Not sure.

GregBrimble:
Two bugs in one! Currently, one [L49](https://github.com/remix-run/remix/blob/57949fe0d73f79811b207fb435657da93f5306d6/packages/remix-cloudflare-workers/sessions/cloudflareKVSessionStorage.ts#L49) and [L66](https://github.com/remix-run/remix/blob/57949fe0d73f79811b207fb435657da93f5306d6/packages/remix-cloudflare-workers/sessions/cloudflareKVSessionStorage.ts#L66), data is being written to KV like this:

```ts
await kv.put(id, JSON.stringify(data), {
  expiration: expires?.getUTCMilliseconds()
});
```

[...]

PR #910, opened 2021-12-05
Use seconds in KV session storage expiration

### LINKED ISSUE #414: Cloudflare KV namespaces expiration seems wrong
Actually, if we set a maxAge to our cloudflare session's cookie, it makes the KV namespace entry being valid for less than a second and that throws an error.

I think there's an issue here:
https://github.com/remix-run/remix/blob/a69a631cb5add72d5fb24211ab2a0be367b6f2fd/packages/remix-cloudflare-workers/sessions/cloudflareKVSessionStorage.ts#L49

Indeed, `.getUTCMilliseconds()` returns milliseconds portion of the time object's value (between 0 and 999) whereas Cloudflare expects a time since UNIX epoch in seconds.

So I think the right code would be something like `.getTime() / 1000`.

GregBrimble:
Fixes #414 and #839.

Description of fix here: https://github.com/remix-run/remix/issues/839#issuecomment-986295640.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -13,6 +13,7 @@
 - gon250
 - goncy
 - graham42
+- GregBrimble
 - ianduvall
 - jacob-ebey
 - jesseflorig
--- a/packages/remix-cloudflare-workers/sessions/cloudflareKVSessionStorage.ts
+++ b/packages/remix-cloudflare-workers/sessions/cloudflareKVSessionStorage.ts
@@ -46,7 +46,7 @@ export function createCloudflareKVSessionStorage({
         }
 
         await kv.put(id, JSON.stringify(data), {
-          expiration: expires?.getUTCMilliseconds()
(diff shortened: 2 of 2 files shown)


GregBrimble:
I've got other PRs open which add me as a contributor, so I won't add that here to avoid conflicts.

kentcdodds (approved):
LGTM. @jacob-ebey?

mcansh (approved):

kentcdodds:
I think we can get this one merged. Would you mind adding yourself as a contributor on this PR? We'll handle merge conflicts on the other PRs.

GregBrimble:
Duck landing

kentcdodds:
@chaance I mistakenly marked this one as a `feat`. Should be a `fix`.

kentcdodds:
Thank you @GregBrimble! We're looking at your other PRs as well 👍

PR #911, opened 2021-12-05
fix: add missing break statements for CatchBoundary

stevenspads:
Without adding the missing `break` statements, the `default` case is executed for all the cases in the switch statement. The 401 and 404 links on page `http://localhost:3000/demos/params` display the `message` of the `default` case rather than that of the `401` and `404` cases.


MichaelDeBoey:
Duplicate of #611

ryanflorence:

PR #917, opened 2021-12-06
Add config path to tailwind style documentation

csprle:
I didn't get tailwind to work with a custom config file until I added the path to the config within the scripts. 

Maybe I did something wrong because in the [tailwind docs](https://tailwindcss.com/docs/installation#customizing-your-configuration) it also says that a file named `tailwind.config.js` should be automatically be read.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -7,6 +7,7 @@
 - bruno-oliveira
 - chaance
 - coryhouse
+- csprle
 - davecalnan
 - derekr
 - dunglas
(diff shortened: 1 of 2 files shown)


kentcdodds:
I think maybe this should be a bug report in the tailwind project?

csprle:
I just tried to reproduce this by creating a CodeSandbox but everything works as expected. So the error will be somewhere in my own project.

PR #925, opened 2021-12-06
Fix deploying to fly.io from local terminal

p10ns11y:
When there is an existing `build` folder (generated during development)
`npm run build` (`remix build`) in the fly generated `Dockerfile`
 Ignores generating new folder and looses environment context (production).


p10ns11y:
Hope https://github.com/remix-run/remix/pull/930 already fixes the issue!

PR #928, opened 2021-12-06
feat(remix-react,remix-server-runtime): make the `MetaFunction` interface accept a generic

### LINKED ISSUE #374: Make the MetaFunction interface generic
I'm not sure how one is supposed to handle issues like this. I tried to submit a PR, but I don't have permission.

When writing a `meta` function, it would be nice if `data` were typed and not just `any`, like this, for example:

```ts
export const meta: MetaFunction<LoaderData> = ({ data }) => {
  return {
    title: 'Public Gists',
    description: `View the latest ${data.length} public gists`,
  };
};
```

If you changed `MetaFunction` to be generic and accepte an optional type for `data`, this would solve the problem as well as maintain backward compatibility.

```diff
From e8cf0eaae88d44b1771b1bb9faaa5002eabe6ce9 Mon Sep 17 00:00:00 2001
From: donavon <github@donavon.com>
Date: Tue, 9 Nov 2021 20:40:40 -0500
Subject: [PATCH] make MetaFunction generic

---
 packages/remix-react/routeModules.ts          | 4 ++--
 packages/remix-server-runtime/routeModules.ts | 4 ++--
 2 files changed, 4 insertions(+), 4 deletions(-)

diff --git a/packages/remix-react/routeModules.ts b/packages/remix-react/routeModules.ts
index 4bbd945..7fefc9e 100644
--- a/packages/remix-react/routeModules.ts
+++ b/packages/remix-react/routeModules.ts
@@ -45,9 +45,9 @@ export interface LinksFunction {
  * `<meta>` tags for a route. These tags will be merged with (and take
  * precedence over) tags from parent routes.
  */
-export interface MetaFunction {
+export interface MetaFunction<T = AppData> {
   (args: {
-    data: AppData;
+    data: T;
     parentsData: RouteData;
     params: Params;
     location: Location;
diff --git a/packages/remix-server-runtime/routeModules.ts b/packages/remix-server-runtime/routeModules.ts
index 847261c..5fa5ddd 100644
--- a/packages/remix-server-runtime/routeModules.ts
+++ b/packages/remix-server-runtime/routeModules.ts
@@ -76,9 +76,9 @@ export interface LoaderFunction {
  * `<meta>` tags for a route. These tags will be merged with (and take
  * precedence over) tags from parent routes.
  */
-export interface MetaFunction {
+export interface MetaFunction<T = AppData> {
   (args: {
-    data: AppData;
+    data: T;
     parentsData: RouteData;
     params: Params;
     location: Location;
-- 
2.30.1 (Apple Git-130)

```

sergiodxa:
# Why is this needed

Right now, if you export a meta function from a route with a loader, and you need to access the loader data inside the meta function, you will need to cast the `data` to be the type of your loader. Something like this:

```ts
export let meta: MetaFunction = ({ data }) => {
  let { title } = data as LoaderData
  return { title }
}
```

# What this change

This PR let MetaFunction accepts a generic called Data which defaults to AppData (current type of data), this way you can do `MetaFunction<LoaderData>` to type the value for the `data` property. And because `data` could be `undefined` in case the loader throw an error, this PR also make the `data` to be typed as `Data | undefined` forcing you to check data exists.

```ts
export let meta: MetaFunction<LoaderData> = ({ data }) => {
  if (!data) return { title: "Error!" }; // here data is undefined so there was an error
  // here data is the LoaderData type so you know it has a title
  return { title: data.title }
}
```

---

This closes #374

### DIFF
--- a/packages/remix-server-runtime/routeModules.ts
+++ b/packages/remix-server-runtime/routeModules.ts
@@ -76,7 +76,7 @@ export interface LoaderFunction {
  * `<meta>` tags for a route. These tags will be merged with (and take
  * precedence over) tags from parent routes.
  */
-export interface MetaFunction<Data = AppData> {
+export interface MetaFunction<Data extends AppData = AppData> {
   (args: {
     data: Data | undefined;
     parentsData: RouteData;
--- a/packages/remix-server-runtime/routeModules.ts
+++ b/packages/remix-server-runtime/routeModules.ts
@@ -76,9 +76,9 @@ export interface LoaderFunction {
  * `<meta>` tags for a route. These tags will be merged with (and take
  * precedence over) tags from parent routes.
  */
-export interface MetaFunction {
+export interface MetaFunction<Data = AppData> {
   (args: {
-    data: AppData;
+    data: Data | undefined;
     parentsData: RouteData;
     params: Params;
     location: Location;


packages/remix-server-runtime/routeModules.ts
```diff
@@ -76,9 +76,9 @@ export interface LoaderFunction {
  * `<meta>` tags for a route. These tags will be merged with (and take
  * precedence over) tags from parent routes.
  */
-export interface MetaFunction {
+export interface MetaFunction<Data = AppData> {
```
MichaelDeBoey, line 79: I think we should make sure that it's an extension of `AppData`

```suggestion
export interface MetaFunction<Data extends AppData = AppData> {
```

MichaelDeBoey (changes requested):

packages/remix-server-runtime/routeModules.ts
sergiodxa: Do you think it would be useful? The AppData type is an alias of `any`
https://github.com/remix-run/remix/blob/31c33109b2cb68a615106dd5cf33b17d33ae10e1/packages/remix-server-runtime/data.ts#L12-L15
Maybe to be future proof if it's changed to something else?

MichaelDeBoey: It's a good practice to extend the default type or some common thing imo.

Since it's `any` it doesn't really matter, but if it would ever be changed to something else (like an object or something else), it will make the `MetaFunction` generic a bit more type-safe.

MichaelDeBoey (approved):

ryanflorence:
Thanks @sergiodxa, this is useful.

Meta is kinda funny right now though because it doesn't work well with error and catch boundaries. It also gets `parentsData` iirc.

I'm thinking it needs to also receive `error` and `caught`, how would you type `meta` if its signature was `meta({ data, error, caught, parentsData })`?

sergiodxa:
> Thanks @sergiodxa, this is useful.
> 
> Meta is kinda funny right now though because it doesn't work well with error and catch boundaries. It also gets `parentsData` iirc.
> 
> I'm thinking it needs to also receive `error` and `caught`, how would you type `meta` if its signature was `meta({ data, error, caught, parentsData })`?

Probably something like this:

```ts
export interface MetaFunction<
  Data extends AppData = AppData,
  Caught extends ThrownResponse = ThrownResponse,
  Error extends global.Error = global.Error,
  ParentsData extends Array<AppData> = RouteData,
> {
  (args: {
    data: Data | undefined;
    parentsData: ParentData;
    error: Error | undefined;
    caught: Caught;
    params: Params;
    location: Location;
  }): HtmlMetaDescriptor;
}
```
And maybe play a little bit with the order of the generics so ensure it makes sense, for example, Data makes sense to be the first one since it's going to be used more than the rest.

sergiodxa:
Another option could be to pass another param to let the dev know the status.

```ts
type MetaBaseArgs = { params: Params; location: Location };

type MetaSuccessArgs<
  Data extends AppData = AppData,
  ParentsData extends Array<AppData> = RouteData
> = MetaBaseArgs & {
  status: "success";
  data: Data;
  parentsData: ParentsData;
  error?: never;
  caught?: never;
};

type MetaErrorArgs<
  Error extends global.Error = global.Error,
  ParentsData extends Array<AppData> = RouteData
> = MetaBaseArgs & {
  status: "error";
  data?: never;
  parentsData: ParentsData;
  error: Error;
  caught?: never;
};

type MetaCaughtArgs<
  Caugth extends ThrownResponse = ThrownResponse,
  ParentsData extends Array<AppData> = RouteData
> = MetaBaseArgs & {
  status: "caught";
  data?: never;
  parentsData: ParentsData;
  error?: never;
  caught: Caugth;
};
```

MichaelDeBoey:
@sergiodxa These changes should also be done in the `remix-react` package.

pcattori:
We've [added support for inferring the (JSON serialized) return type of `loader`](https://github.com/remix-run/remix/releases/tag/remix%401.6.5), so that should be used here as well:

```ts
const meta: MetaFunction<typeof loader> = ...
```

@sergiodxa : you can copy [how `useActionData` is typed](https://github.com/remix-run/remix/blob/main/packages/remix-react/components.tsx#L1379-L1381):

```ts
export function useActionData<T = AppData>():
  | UseDataFunctionReturn<T>
  | undefined { /* ... */ }
```

pcattori:
Superceded by #4022

oncet:
I'm really looking forward to have caught data inside the `MetaFunction`.

Just recently I submitted to the ideas section: https://github.com/remix-run/remix/discussions/4020

PR #930, opened 2021-12-06
fix(fly): use remote-only deploys

kentcdodds:
This is simpler for beginners

### DIFF
--- a/packages/create-remix/templates/fly/package.json
+++ b/packages/create-remix/templates/fly/package.json
@@ -3,7 +3,7 @@
     "postinstall": "remix setup node",
     "build": "remix build",
     "dev": "remix dev",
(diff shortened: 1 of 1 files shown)


mcansh (approved):

packages/create-remix/templates/fly/package.json
```diff
@@ -3,7 +3,7 @@
     "postinstall": "remix setup node",
     "build": "remix build",
     "dev": "remix dev",
-    "deploy": "flyctl deploy",
+    "deploy": "fly deploy --remote-only",
```
aganoza, line 6: Is the ```fly``` command interchangeable with ```flyctl```?

kentcdodds: Yes

Issue #927, opened 2021-12-06
[Bug]: useSubmitImpl() not respects incoming query search when using POST


jacargentina:
### What version of Remix are you using?

1.0.6

### What version of Node are you using? Minimum supported version is 14.

17.0.1

### Steps to Reproduce

- `routes/myform.tsx` which is navigated like this `http://localhost:3000/myform?returnUrl=/blog`
- myform.tsx calls `submit({values: ...}, {method: 'post'})`
- myform.tsx action simply returns null
- URL now is missing the `returnUrl=/blog` query string

### Expected Behavior

URL should remain unchanged and maintain its query string

### Actual Behavior

URL only points to the plain route, without any query string

machour:
Tracked in #931

mcansh:
🤖 Hello there,

We just published version `v1.6.8` which includes a fix for this. If you'd like to take it for a test run please try it out and let us know what you think!

Thanks!

PR #931, opened 2021-12-06
[Bug]: useSubmitImpl() not respects incoming query search when using POST

jacargentina:
My first contribution. This tests shows the bug explained on #927


ryanflorence:
Ah, the real problem here is that `useFormAction()` doesn't preserve the search params.

Your current solution is just always grabbing the current URL's search params, which is not what we want, it's the form that determines the action we're posting to, not the current page.

The mixup is that `<Form>` and `<Form action=".">` default to the current URL, and the bug is that it's not preserving the search params.

If you do `<Form>` or `<Form action=".">` at "/some/url?q=1" then the resulting HTML should be `<form action="/some/url?q=1">`.

That logic is in `useFormAction`. Want to take a stab at getting it fixed there?

ryanflorence:
Actually this seems to go deep. @chaance can you look into this? I think this is busted all the way back into React Router, `useResolvedPath` isn't preserving the search param

https://github.com/remix-run/react-router/blob/7aa62850ec86ddaf70b9ac7ec81dc76ce1254cf4/packages/react-router/index.tsx#L608-L620

chaance:
The behavior in React Router is as intended. I'll take a look at this tomorrow morning and make sure we get it fixed.

jacargentina:
@ryanflorence Ok.
@chaance May be I can refactor my PR to be only the test, and leave the fix at your charge? Here to help.

chaance:
@jacargentina That would actually be quite helpful, thanks! I'll be adding a few more tests as well.

nrako:
I made this little contraption to work around that problem for the time being:

```typescript
import { useFormAction, useSearchParams } from '@remix-run/react'

/**
 * Hook to work around https://github.com/remix-run/remix/pull/931
 */
export default function useFormActionWithSearchParams(to: string = '.') {
  const actionUrl = new URL(useFormAction(to), 'http://localhost')
  const [actionSearch] = useSearchParams()

  const combinedSearch = new URLSearchParams({
    ...Object.fromEntries(actionUrl.searchParams),
    ...Object.fromEntries(actionSearch),
  })

  return actionUrl.pathname + '?' + combinedSearch.toString()
}
```

```jsx
<Form action={useFormActionWithSearchParams()} >
```

mcansh:
🤖 Hello there,

We just published version `v1.6.8` which includes a fix for this. If you'd like to take it for a test run please try it out and let us know what you think!

Thanks!

PR #939, opened 2021-12-07
feat: Add <Outlet context> prop

mcansh:
closes [REM-576](https://linear.app/remix-run/issue/REM-576)

### DIFF
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -2236,6 +2236,48 @@ return json(data, {
 });
 ```
 
+### `<Outlet context>`
+
+This component is a wrapper around React Router's Outlet with the ability to pass it's data down to nested routes.
+
+```tsx filename=app/routes/users.$userId.tsx
+import { Outlet } from "remix";
+import type { LoaderFunction } from "remix";
+import type { User } from "~/models/user";
+
+export const loader: LoaderFunction = ({ params }) => {
+  let user = db.get(params.userId)
+  if (!user) {
+    throw new Response("User not found", {status: 404})
+  }
+  return user
+}
+
+export default function Parent() {
+  const data = useLoaderData()
+  return (
+    <Layout>
+      <Outlet context={data}>
+    </Layout>
+  );
+}
+```
+
+### `useOutletContext()`
+
+This hook returns the data from the Outlet context.
+
+```tsx filename=app/routes/users.$userId.edit.tsx
+import { useOutletContext } from "remix";
+import type { User } from "~/models/user";
+
+export default function Child() {
+  const data = useOutletContext<User>();
+
+  return <h1>{data.name}</h1>;
+}
+```
+
 ## Types
 
 ```ts
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -2236,13 +2236,139 @@ return json(data, {
 });
 ```
 
-### `<Outlet context>`
+### `<Outlet context />`
 
 This component is a wrapper around React Router's Outlet with the ability to pass UI state down to nested routes.
+
+<docs-warning>You can use this for loader data, but you don't need to. It's easier to access all loader data in any component via [`useLoaderData`](#useloaderdata) or [`useMatches`](#usematches).</docs-warning>
+
+Here's a practical example of when you may want to use this feature. Let's say you've got a list of companies that have invoices and you want to display those companies in an accordion. We'll render our outlet in that accordion, but we want the invoice sorting to be controlled by the parent (so changing companies preserves the invoice sorting). This is a perfect use case for `<Outlet context>`.
+
+```tsx filename=app/routes/companies.tsx lines=[6,27-30,35-41,50-54,65]
+import type { LoaderData } from "remix";
+import {
+  json,
+  useLoaderData,
+  useParams,
+  Outlet
+} from "remix";
+import {
+  Accordion,
+  AccordionItem,
+  AccordionButton,
+  AccordionPanel
+} from "@reach/accordion";
+import { getCompanies, Companies } from "~/utils/companies";
+
+type LoaderData = {
+  companies: Array<Companies>;
+};
+
+export const loader: LoaderFunction = async () => {
+  const data: LoaderData = {
+    companies: await getCompanies()
+  };
+  return json(data);
+};
+
+type Sort = "ASC" | "DESC";
+export type ContextType = {
+  invoiceSort: Sort;
+};
+
+export default function CompaniesRoute() {
+  const data = useLoaderData<LoaderData>();
+
+  const [invoiceSort, setInvoiceSort] =
+    React.useState<Sort>("ASC");
+  function changeInvoiceSort() {
+    setInvoiceSort(invoiceSort === "ASC" ? "DESC" : "ASC");
+  }
+  const context: ContextType = { sort };
+  const outlet = <Outlet context={context} />;
+
+  const params = useParams();
+  const selectedCompanyIndex = data.companies.findIndex(
+    company => company.id === params.companyId
+  );
+
+  return (
+    <div>
+      <button onClick={changeInvoiceSort}>
+        {invoiceSort === "ASC"
+          ? "Sort Descending"
+          : "Sort Ascending"}
+      </button>
+      <Accordion index={selectedCompanyIndex}>
+        {data.companies.map(company => (
+          <AccordionItem key={company.id}>
+            <AccordionButton as={Link} to={company.id}>
+              {company.name}
+            </AccordionButton>
+            {/* render the outlet by the
+            currently selected company */}
+            <AccordionPanel>
+              {params.companyId === company.id
+                ? outlet
+                : null}
+            </AccordionPanel>
+          </AccordionItem>
+        ))}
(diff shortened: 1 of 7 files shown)


packages/remix-react/components.tsx
```diff
@@ -1309,3 +1310,18 @@ function useComposedRefs<RefValueType = any>(
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, refs);
 }
+
+let OutletContext = React.createContext<AppData>(null);
+
+export function Outlet<Data = AppData>({ context }: { context?: Data }) {
+  return (
+    <OutletContext.Provider value={context}>
```
sergiodxa, line 1318: What if it's called data? The generic is also named that way
```suggestion
export function Outlet<Data = AppData>({ data }: { data?: Data }) {
  return (
    <OutletContext.Provider value={data}>
```

packages/remix-react/components.tsx
```diff
@@ -1309,3 +1310,18 @@ function useComposedRefs<RefValueType = any>(
...
+      <RouterOutlet />
+    </OutletContext.Provider>
+  );
+}
+
+export function useOutletContext<ContextData>() {
```
sergiodxa, line 1324:
```suggestion
export function useOutletData<ContextData>() {
```
useOutletData?

packages/remix-react/components.tsx
```diff
@@ -1309,3 +1310,18 @@ function useComposedRefs<RefValueType = any>(
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, refs);
 }
+
+let OutletContext = React.createContext<AppData>(null);
+
+export function Outlet<Data = AppData>({ context }: { context?: Data }) {
```
sergiodxa, line 1316:
```suggestion
export function Outlet<Data extends AppData = AppData>({ context }: { context?: Data }) {
```
I think this should extend AppData too, that way it will match the type the OutletContext expects always (I know AppData is any but it may not be any in the future)

fixtures/gists-app/app/routes/__layout.tsx
```diff
@@ -1,10 +1,22 @@
...
+  return data;
+};
+
 export default function LayoutTest() {
+  let data = useLoaderData<RouteData>();
   return (
     <div data-test-id="_layout">
       <h1>Layout Test</h1>
-      <Outlet />
+      <Outlet context={data} />
```
kentcdodds, line 19: We don't want people using the `context` prop to share loader data because they can already access that more easily via `useMatches`. Do you mind updating this test to use something else so people reading the tests don't get the wrong idea? I'm writing more docs on this too.

```suggestion
      <Outlet context={contextData} />
```

fixtures/gists-app/app/routes/__layout.tsx
```diff
@@ -1,10 +1,22 @@
...
+  nested: boolean;
+}
+
+export let loader: LoaderFunction = () => {
+  let data: RouteData = { nested: true };
+  return data;
+};
+
 export default function LayoutTest() {
+  let data = useLoaderData<RouteData>();
```
kentcdodds, line 15:
```suggestion
  let [contextData] = useState<ContextData>({ nested: true });
```

fixtures/gists-app/app/routes/__layout.tsx
```diff
@@ -1,10 +1,22 @@
...
 import { Outlet } from "remix";
 
+export interface RouteData {
+  nested: boolean;
+}
+
+export let loader: LoaderFunction = () => {
+  let data: RouteData = { nested: true };
+  return data;
+};
```
kentcdodds, line 12:
```suggestion
export interface ContextData {
  nested: boolean;
}
```

fixtures/gists-app/app/routes/__layout.tsx
```diff
@@ -1,10 +1,22 @@
+import type { LoaderFunction } from "remix";
+import { useLoaderData } from "remix";
```
kentcdodds, line 2:
```suggestion
```

fixtures/gists-app/app/routes/__layout/with-layout.tsx
```diff
@@ -1,3 +1,13 @@
+import { useOutletContext } from "remix";
+
+import type { RouteData } from "../__layout";
```
kentcdodds, line 3:
```suggestion
import type { ContextData } from "../__layout";
```

kentcdodds (commented):
I'm going to push some more docs, but here are a few updates to the tests to communicate the intended use of this feature.

kentcdodds:
Just pushed docs changes. Feedback welcome!

mcansh:
thanks @kentcdodds!

packages/remix-react/components.tsx
kentcdodds: I think I prefer sticking to the termonology of `context` rather than `data` because I'm really concerned that people will unnecessarily use this in place of what's arguably a better solution for accessing the parent's loader data (via `useMatches`).

packages/remix-react/components.tsx
```diff
@@ -1309,3 +1310,18 @@ function useComposedRefs<RefValueType = any>(
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, refs);
 }
+
+let OutletContext = React.createContext<AppData>(null);
+
+export function Outlet<Data extends AppData = AppData>({ context }: { context?: Data }) {
```
kentcdodds, line 1316: The context value should *not* be `AppData` because we don't want people using this for data they already have access to in `useMatches`.

kentcdodds (approved):
@mcansh do you mind reviewing my changes and docs? This looks good to be merged for me if you're good with it.

mcansh:
yup, giving them a read now! thanks again!

docs/api/remix.md
```diff
@@ -2236,6 +2236,140 @@ return json(data, {
...
+import {
+  Accordion,
+  AccordionItem,
+  AccordionButton,
+  AccordionPanel
+} from "@reach/accordion";
+import { getCompanies, Companies } from "~/utils/companies";
+
+type LoaderData = {
+  companies: Array<Companies>;
```
MichaelDeBoey, line 2264: As this is a simple type, you can just do

```suggestion
  companies: Companies[];
```

docs/api/remix.md
```diff
@@ -2236,6 +2236,140 @@ return json(data, {
...
+  invoiceSort: Sort;
+};
+
+export default function CompaniesRoute() {
+  const data = useLoaderData<LoaderData>();
+
+  const [invoiceSort, setInvoiceSort] =
+    React.useState<Sort>("ASC");
+  function changeInvoiceSort() {
+    setInvoiceSort(invoiceSort === "ASC" ? "DESC" : "ASC");
```
MichaelDeBoey, line 2285:
```suggestion
    setInvoiceSort(sort => sort === "ASC" ? "DESC" : "ASC");
```

fixtures/gists-app/app/routes/__layout.tsx
```diff
@@ -1,10 +1,16 @@
+import * as React from "react";
 import { Outlet } from "remix";
 
+export interface ContextData {
+  nested: boolean;
+}
+
 export default function LayoutTest() {
+  let [contextData] = React.useState<ContextData>({ nested: true });
```
MichaelDeBoey, line 9:
```suggestion
  const [contextData] = React.useState<ContextData>({ nested: true });
```

fixtures/gists-app/app/routes/__layout/with-layout.tsx
```diff
@@ -1,3 +1,13 @@
+import { useOutletContext } from "remix";
+
+import type { ContextData } from "../__layout";
+
 export default function WithLayout() {
-  return <h1>Page inside layout</h1>;
+  let outletData = useOutletContext<ContextData>();
```
MichaelDeBoey, line 6:
```suggestion
  const outletData = useOutletContext<ContextData>();
```

docs/api/remix.md
kentcdodds: Nope. [I don't like that syntax](https://www.briefs.fm/3-minutes-with-kent/222) 😅

docs/api/remix.md
kentcdodds: Whoops, yeah, should fix this...

fixtures/gists-app/app/routes/__layout.tsx
kentcdodds: We only `const`-ify externally-facing code. We `let` it out internally though :)

docs/api/remix.md
mcansh: glad i'm not alone in that 😆

packages/remix-react/components.tsx
ryanflorence: It's not necessarily data, it's just ... whatever. Could be a function or a Symbol!

docs/api/remix.md
ryanflorence: You two are `weirdo[]`. I love that syntax.

PR #942, opened 2021-12-07
remove unnecessary {" "} from doc codes

MohammadHosseinBagheri:


ryanflorence:
we actually want whitespace between those elements, but thanks!

PR #966, opened 2021-12-09
chore: remove all `React` default imports

MichaelDeBoey:
We're pushing `react-jsx` in our templates

https://github.com/remix-run/remix/blob/f5a551ee139484f8d821220986a0515f95a69ce2/packages/create-remix/templates/_shared_ts/tsconfig.json#L6

### DIFF
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -13,8 +13,7 @@ This package provides all the components, hooks, and [Web Fetch API](https://dev
 
 These components are to be used once inside of your root route (`root.tsx`). They include everything Remix figured out or built in order for your page to render properly.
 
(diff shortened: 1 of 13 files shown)


MichaelDeBoey:
@jacob-ebey Do you have any idea where the problem could be in the failing tests?

ryanflorence:
Note that our compiler (esbuild) doesn't actually use the react-jsx runtime, it still uses `React.createElement`, we just automatically include it in `.jsx|.tsx` files. Might be the reason things are failing?

Do they fail locally?

MichaelDeBoey:
@ryanflorence This should be fine now I think

PR #967, opened 2021-12-09
chore(tutorials/jokes): prefetch `new` page

MichaelDeBoey:
Fixes https://github.com/remix-run/remix/discussions/961

### DIFF
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -1288,7 +1288,11 @@ export default function JokesRoute() {
                 <Link to="some-joke-id">Hippo</Link>
               </li>
             </ul>
-            <Link to="new" className="button">
+            <Link
+              to="new"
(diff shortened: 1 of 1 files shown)


ryanflorence:
@kentcdodds you want this in the tutorial? If so go ahead and merge :)

kentcdodds (approved):
Yeah, this is probably a good idea 👍

kentcdodds:
Actually, in all cases, these code blocks come before we add JS to the page so it won't do anything. There's a section in the tutorial where I describe prefetching and I suppose we could mention you could add it to the `new` link and even add a code block there if you'd like: https://remix.run/docs/en/v1/tutorials/jokes#prefetching

ryanflorence:
Ah yeah, it introduces the concept already, so I think we're good.

PR #973, opened 2021-12-09
Added the needed cmd to run the app at localhost:3000

ahmdtalat:
The ```npm run dev``` cmd is not enough to run the dev server


ryanflorence:
Thanks for this PR! `remix dev` is enough if you use the Remix App Server as instructed in tutorial. If you pick anything else you need to check the README :)

PR #974, opened 2021-12-09
chore: move Outlet context to React Router

mcansh:
dependent on https://github.com/remix-run/react-router/pull/8461 and it's release in React Router

partially reverts 64e14e95aa5c4e6aaa6a789dddb69a797d2b0369


kentcdodds (commented):
Let's also update react-router in this PR (which we'll need to wait for).

kentcdodds (approved):
LGTM once the `yarn.lock` is fixed 👍

kentcdodds (changes requested):
Oh, actually, we probably want to remove the tests as well.

mcansh:
> Oh, actually, we probably want to remove the tests as well.

done!

PR #975, opened 2021-12-09
chore: fix import order

MichaelDeBoey:
@kentcdodds Already told a couple of times in his livestreams that this is the order of imports he wants in this repo

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,14 +22,20 @@ module.exports = {
   ],
   rules: {
     "@typescript-eslint/consistent-type-imports": "error",
+
+    "sort-imports": [
+      "error",
+      { ignoreCase: true, ignoreDeclarationSort: true }
+    ],
     "import/order": [
       "error",
       {
+        groups: ["builtin", "external", "internal", "parent", "sibling"],
         "newlines-between": "always",
-        groups: [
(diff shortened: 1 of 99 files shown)


kentcdodds (commented):
Thanks for taking the time! This is my preference, though I don't know that I'd care to enforce it via eslint 🤔 

I guess if this autofixes and only warns then it's fine. May be cool if @mcansh adds `eslint --fix` to the auto-prettier bot 🤔

mcansh:
> Thanks for taking the time! This is my preference, though I don't know that I'd care to enforce it via eslint 🤔
> 
> I guess if this autofixes and only warns then it's fine. May be cool if @mcansh adds `eslint --fix` to the auto-prettier bot 🤔

the bot runs any `format` script in package.json if it exists :)

kentcdodds (commented):
I don't think I should decide this. I think @mjackson set it up in the first place. I think he has opinions on sorting imports. My opinions are pretty fuzzy and I don't care all that much.

ryanflorence:
Thanks for your work here Micahel! We're going to be adding a lint config to everything soon, and then I'm going to run it on all the code after I get the PRs down to single digits to avoid a bunch of merge conflicts.

PR #976, opened 2021-12-09
feat: allow ESM output of server builds

jacob-ebey:
chore: updated package.json from "browser" to "module" field

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -70,11 +70,19 @@ export interface AppConfig {
    */
   devServerPort?: number;
   /**
-   * The delay before the dev server broadcasts a reload event
+   * The delay before the dev server broadcasts a reload event.
    */
   devServerBroadcastDelay?: number;
 
+  /**
+   * Additional MDX remark / rehype plugins.
+   */
   mdx?: RemixMdxConfig | RemixMdxConfigFunction;
+
+  /**
+   * The output format of the server build. Defaults to "cjs".
+   */
+  serverModuleFormat: "esm" | "cjs";
 }
 
 /**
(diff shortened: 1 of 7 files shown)


mcansh (approved):

packages/remix-dev/config.ts
```diff
@@ -141,7 +155,10 @@ export interface RemixConfig {
    */
   devServerBroadcastDelay: number;
 
-  mdx?: RemixMdxConfig | RemixMdxConfigFunction;
+  /**
+   * Configuration that effects how the bundles are generated.
+   */
+  buildOptions: RemixBuildOptions;
```
mjackson, line 161: How is `buildOptions` distinguished from other remix config options in your mind? Can we come up with a better name for what you're looking for? Or just leave the config flat without adding another key?

jacob-ebey: The distinguishing feature to me of these properties are that they effect the bundled output contents / format, rather than where it ends up.

jacob-ebey: Maybe compilerOptions?

packages/remix-node/package.json
```diff
@@ -3,6 +3,8 @@
   "description": "Node.js platform abstractions for Remix",
   "version": "1.0.6",
   "license": "MIT",
+  "main": "./index.js",
```
jacob-ebey, line 6: Revert changes to node module as it's not even runable in ESM because of the React package.

rollup.config.js
```diff
@@ -598,7 +638,18 @@ function remixServe() {
           exclude: /node_modules/,
           extensions: [".ts"]
         }),
-        nodeResolve({ extensions: [".ts"] })
+        nodeResolve({ extensions: [".ts"] }),
+        // Allow dynamic imports in CJS code to allow us to utilize
+        // ESM modules as part of the compiler.
+        {
```
jacob-ebey, line 644: Revert.

Issue #903, opened 2021-12-05
[Bug]: @remix-run/serve sets high `max-age` and `immutable` cache header on everything 


sheepsteak:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [X] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.0.4

### What version of Node are you using? Minimum supported version is 14.

16.13.0

### Steps to Reproduce

* Build and start an application in production that has some resources in the `public` (remix blog)
* See in dev tools that everything in `public` folder is served with a high `max-age` and `immutable`

![image](https://user-images.githubusercontent.com/1379888/144753439-bcb6ac73-335d-4782-a0eb-7dbcfe0c9b66.png)

### Expected Behavior

Only files in `public/build` should have long, immutable caching as they are fingerprinted. The `express` starter template separates the two (although it has issues with the ordering of `express.static` as shown in https://github.com/remix-run/remix/issues/677 and fixed in https://github.com/remix-run/remix/pull/678):

https://github.com/remix-run/remix/blob/57949fe0d73f79811b207fb435657da93f5306d6/packages/create-remix/templates/express/server/index.js#L13-L17

Files in the base `public` folder should have a low `max-age` of `1h` like elsewhere so that favicons, images, fonts, etc. can be updated as needed.

### Actual Behavior

Due to this line everything in `public` gets the same cache headers:

https://github.com/remix-run/remix/blob/57949fe0d73f79811b207fb435657da93f5306d6/packages/remix-serve/index.ts#L10

This is fine for the built application files in `public/build` as they are fingerprinted but not for resources like `favicon.ico`that might need to change.

mcansh:
hey @sheepsteak this was fixed in #2363 and released in v1.3.2

PR #977, opened 2021-12-09
Add MaxAge for remix serve #903

ch4nd4n:
Fixes issue raised #903 by introducing ENV variable REMIX_SERVE_MAX_AGE

```bash
REMIX_SERVE_MAX_AGE=5s npm run dev
```


ryanflorence:
Thanks for the catch! Let's remove the envvar and just hard code it to `1h`. If people want it to work differently they can use the express adapter.

ch4nd4n:
@ryanflorence done.

sheepsteak:
This doesn't quite fix #903 as all assets are still set as `immutable` like `public/favicon.ico` and `public/robots.txt` even though they're not fingerprinted. I've been using this in my Express template:

```typescript
if (mode === "production") {
  // Remix fingerprints its assets so we can cache forever
  app.use(
    "/build",
    express.static("public/build", { immutable: true, maxAge: "1y" })
  );
  // Any other assets such as favicons.
} else {
  // No caching needed
  app.use(express.static("public"));
}
```

In development there's then no caching used at all but in production the correct thing happens for the relevant type of asset. This could be put in `@remix-run/serve` and the Remix Express template as they both need to do this.

kiliman:
@ch4nd4n Can you update your PR to match what the Express adapter uses. @jacob-ebey said he'll merge it in. https://discord.com/channels/770287896669978684/770287896669978687/934209360589103185

```ts
// You may want to be more aggressive with this caching
app.use(express.static('public', { maxAge: '1h' }))

// Remix fingerprints its assets so we can cache forever
app.use(express.static('public/build', { immutable: true, maxAge: '1y' }))
```

sheepsteak:
> @ch4nd4n Can you update your PR to match what the Express adapter uses. @jacob-ebey said he'll merge it in. https://discord.com/channels/770287896669978684/770287896669978687/934209360589103185
> 
> ```ts
> // You may want to be more aggressive with this caching
> app.use(express.static('public', { maxAge: '1h' }))
> 
> // Remix fingerprints its assets so we can cache forever
> app.use(express.static('public/build', { immutable: true, maxAge: '1y' }))
> ```

Just wanted to state again that this still won't fix #903. The first `app.use(express.static('public', { maxAge: '1h' }))` will greedily take anything in `public` so the second `express.static('public/build', { immutable: true, maxAge: '1y' })` will never be used.

What I put in my comment above will fix this and also turn off caching for development altogether. The express template could do with this too.

ch4nd4n:
> What I put in my comment above will fix this and also turn off caching for development altogether. The express template could do with this too.

This makes sense to me. I have updated the PR. Please check.

ryanflorence:
Sorry, can you get it to match #1861? Apparently express will just use the first one defined here for `public/`.

mjackson:
@ryanflorence I think this will actually work as-is, and it's a much more elegant approach than is currently being used in #1861.

He's using `app.use('/build', express.static('public/build', ...))`, which tells express to only apply that middleware to `/build` requests. And since that comes first, other requests for public assets will fall through to the following `app.use(express.static('public', ...))`.

mjackson:
Actually just spoke w @ryanflorence about this. @ch4nd4n Can you please remove the `mode === "production"` check? We want to have the same caching behavior in dev that you get in production. If you don't want caching in dev, you can disable it in the browser devtools.

ch4nd4n:
> Actually just spoke w @ryanflorence about this...
> 
> @ch4nd4n Can you please remove the `mode === "production"` check? 

@mjackson please see if this helps. I haven't been active due to office work.

philip-weber:
@mjackson I am wondering if the caching simplification from PR #1994 in [line 23](https://github.com/remix-run/remix/blob/5ab4513e8859e72a5afcfb2a12efce7fb5b3853b/packages/create-remix/templates/express/server.js#L23) `app.use(express.static("public/build", ...))` is correct. Shouldn't it be `app.use(express.static('public', ...))` like in this PR?

MichaelDeBoey:
This was implemented by #2363

PR #983, opened 2021-12-10
feat: define process.env.NODE_ENV for server build

jacob-ebey:
This will help avoid descrepancies at runtime between it being defined for the browser and a different runtime value for the server. This has been a source of issues for many regarding surfacing `<LiveReload>` in production.

### DIFF
--- a/packages/remix-cloudflare-workers/globals.ts
+++ b/packages/remix-cloudflare-workers/globals.ts
@@ -1,9 +1,14 @@
 import { sign, unsign } from "./cookieSigning";
 
 declare global {
+  interface ProcessEnv {
+    NODE_ENV: "development" | "production" | "test";
+  }
+
   interface WorkerGlobalScope {
     sign: typeof sign;
     unsign: typeof unsign;
+    process: { env: ProcessEnv };
   }
 }
 
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -357,6 +357,9 @@ async function createServerBuild(
     // of CSS and other files.
     assetNames: "_assets/[name]-[hash]",
     publicPath: config.publicPath,
+    define: {
+      "process.env.NODE_ENV": JSON.stringify(options.mode)
+    },
     plugins: [
       mdxPlugin(config),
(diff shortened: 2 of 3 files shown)


ryanflorence:
our browser builds already set this, so it seems like we should do it for server builds too, the chance of mismatches from the REAL env var is higher right now than it is with this change I'd bet?

ryanflorence:
(side note: I wish bundlers never picked up on this envvar, it should have been a build arg from the beginning 🥲)

jacob-ebey:
> our browser builds already set this, so it seems like we should do it for server builds too, the chance of mismatches from the REAL env var is higher right now than it is with this change I'd bet?

Yeah, this should help avoid the issue more so than today 🤞

mcansh (approved):

mjackson:
I like leaving `process.env.NODE_ENV` in code that runs in node.js because that's a node-ism that is perfectly valid in node. The original idea behind not setting it in the server build was so that people could run their server build easily in development or production modes just by setting an env var instead of having to rebuild with a different env var.

What is the workflow that is causing people to have mismatched NODE_ENV on the server and browser builds? What exactly are people doing?

mjackson (approved):

salper:
Enforcing `NODE_ENV` type leads to this kind of typescript issue when defining different local values:
```
Subsequent property declarations must have the same type.  Property 'NODE_ENV' must be of type '"production" | "development" | "test"'
```
For the record, I'm deploying using `architect`, which sets `NODE_ENV` to `production | staging | testing`, so I have my own global types that now conflict with remix types.  It's probably possible to circumvent the issue using some local typescript config that I don't know of, as I'm no expert in the field. Do you have any tips on that?

salper:
FTR: I've discussed with arc maintainers (https://github.com/architect/functions/pull/512), and they'll rely only on `ARC_ENV` as a runtime env in the future. Like @ryanflorence, I also think it's a little bit weird to use a runtime variable like `NODE_ENV` as a build one. Out of curiosity, did you try to circumvent that with `REMIX_ENV' at the time? If that's the case, what happened since?

PR #984, opened 2021-12-10
fix: use document over window for browser check

jacob-ebey:
Runtimes such as Deno have a window defined making this check not work.

### DIFF
--- a/docs/other-api/constraints.md
+++ b/docs/other-api/constraints.md
@@ -248,13 +248,13 @@ The most common scenario is intitializing a third party API when your module is 
 
 #### Window Guard
 
-This ensures the library is only initialized if there is a `window`, meaning you're in the browser.
+This ensures the library is only initialized if there is a `document`, meaning you're in the browser. We recomend `document` over `window` because server runtimes like Deno have a global `window` available.
 
 ```js [3]
 import firebase from "firebase/app";
 
-if (typeof window !== "undefined") {
-  firebase.initializeApp(window.ENV.firebase);
+if (typeof document !== "undefined") {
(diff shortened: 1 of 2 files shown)


docs/other-api/constraints.md
```diff
@@ -248,13 +248,13 @@ The most common scenario is intitializing a third party API when your module is
 
 #### Window Guard
```
mjackson, line 249: Let's change this to say "Document Guard"

docs/other-api/constraints.md
```diff
@@ -248,13 +248,13 @@ The most common scenario is intitializing a third party API when your module is
...
 
-This ensures the library is only initialized if there is a `window`, meaning you're in the browser.
+This ensures the library is only initialized if there is a `document`, meaning you're in the browser. We recomend `document` over `window` because server runtimes like Deno have a global `window` available.
```
mjackson, line 251: s/server runtimes like Deno have/Deno has/

mjackson (changes requested):

mcansh (approved):

kentcdodds (approved):
Perfect

btargac:
Does remix run on Deno instead of Node runtime ?

girishk21:
> Does remix run on Deno instead of Node runtime ?

https://twitter.com/ebey_jacob/status/1471009211416932353?s=20 it does, and it runs really well 😄 . The official adapter should land soon I guess

kentcdodds:
It runs on both and more so we need code that runs equally well on all platforms we support.

PR #985, opened 2021-12-10
wip: initial cloudflare-pages package

jacob-ebey:
feat: added esm build for cloudflare-pages package

### DIFF
--- a/packages/create-remix/templates/cloudflare-pages/README.md
+++ b/packages/create-remix/templates/cloudflare-pages/README.md
@@ -1,0 +1,20 @@
+# Welcome to Remix!
+
+- [Remix Docs](https://remix.run/docs)
+
+## Development
+
+You will be utlizing Wrangler for local development to emulate the Cloudflare runtime. This is already wired up in your package.json as the `dev` script:
+
+```sh
+# start the remix dev server and wrangler
+$ npm run dev
+```
+
+Open up [http://127.0.0.1:8788](http://127.0.0.1:8788) and you should be ready to go!
+
+## Deployment
+
+Cloudflare Pages are currently only deployable through their Git provider integrations.
+
+If you don't already have an account, then [create a cloudflare account here](https://dash.cloudflare.com/sign-up) and after verifying your email address with Cloudflare, go to your dashboard and follow the [CF Pages deployment guide](https://developers.cloudflare.com/pages/framework-guides/deploy-anything).
--- a/packages/create-remix/templates/cloudflare-pages/functions/[[path]].js
+++ b/packages/create-remix/templates/cloudflare-pages/functions/[[path]].js
@@ -1,0 +1,16 @@
+import { createFetchHandler } from "@remix-run/cloudflare-pages";
+
+// @ts-ignore
+import * as build from "../build";
+
+const handleFetch = createFetchHandler({
+  build
+});
+
+export function onRequest(context) {
+  const request = new Request(context.request);
+  // https://github.com/cloudflare/wrangler2/issues/117
+  request.headers.delete("If-None-Match");
+
+  return handleFetch(context);
+}
--- a/packages/create-remix/templates/cloudflare-workers/package.json
+++ b/packages/create-remix/templates/cloudflare-workers/package.json
@@ -13,7 +13,7 @@
     "@remix-run/cloudflare-workers": "*"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^2.2.2",
+    "@cloudflare/workers-types": "^3.2.0",
     "esbuild": "0.13.14",
     "miniflare": "2.0.0-next.3"
   }
--- a/packages/remix-cloudflare-pages/magicExports/platform.ts
+++ b/packages/remix-cloudflare-pages/magicExports/platform.ts
@@ -1,0 +1,4 @@
+// This file lists all exports from this package that are available to `import
+// "remix"`.
+
+export { createCloudflareKVSessionStorage } from "@remix-run/cloudflare-pages";
--- a/packages/remix-cloudflare-pages/worker.ts
+++ b/packages/remix-cloudflare-pages/worker.ts
@@ -1,0 +1,84 @@
+import type { ServerBuild, AppLoadContext } from "@remix-run/server-runtime";
+import { createRequestHandler as createRemixRequestHandler } from "@remix-run/server-runtime";
+
+export interface CreateFetchHandlerParams<Env = any> {
+  build: ServerBuild;
+  getLoadContext?: (context: EventContext<Env, any, any>) => AppLoadContext;
+  mode?: string;
+}
+
+export function createRequestHandler<Env = any>({
+  build,
+  getLoadContext,
+  mode
+}: CreateFetchHandlerParams<Env>): PagesFunction<Env> {
+  let platform = {};
+  let handleRequest = createRemixRequestHandler(build, platform, mode);
+
+  return context => {
+    let loadContext =
+      typeof getLoadContext === "function"
+        ? getLoadContext(context)
+        : undefined;
+    return handleRequest(context.request, loadContext);
+  };
+}
+
+type EnvWithAssets = Record<string, unknown> & {
+  ASSETS: { fetch: typeof fetch };
+};
+
+async function handleAsset(request: Request, env: unknown) {
+  let envWithAssets = env as EnvWithAssets;
+
+  const response = await envWithAssets.ASSETS.fetch(request);
+  if (response.ok) return response;
+}
+
+declare const process: any;
+
+export function createFetchHandler<Env = any>({
+  build,
+  getLoadContext,
+  mode
+}: CreateFetchHandlerParams<Env>) {
+  const handleRequest = createRequestHandler<Env>({
+    build,
+    getLoadContext,
+    mode
+  });
+
+  const handleFetch = async (context: EventContext<Env, any, any>) => {
+    let response: Response | undefined;
+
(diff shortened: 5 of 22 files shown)


packages/remix-cloudflare-workers/package.json
```diff
@@ -16,9 +18,9 @@
     "@remix-run/server-runtime": "1.0.6"
   },
   "peerDependencies": {
-    "@cloudflare/workers-types": "^2.2.2"
+    "@cloudflare/workers-types": "^3.2.0"
```
MichaelDeBoey, line 21: This would be a breaking change for `@remix-run/cloudflare-workers`

packages/create-remix/templates/cloudflare-pages/wrangler.toml
```diff
@@ -0,0 +1,18 @@
+name = "remix-cloudflare-workers"
```
GregBrimble, line 1: There's no need for a `wrangler.toml` in a Pages project.

packages/create-remix/templates/cloudflare-pages/README.md
```diff
@@ -0,0 +1,34 @@
...
+$ npm start
+```
+
+Open up [http://127.0.0.1:8787](http://127.0.0.1:8787) and you should be ready to go!
+
+If you'd rather run everything in a single tab, you can look at [concurrently](https://npm.im/concurrently) or similar tools to run both processes in one tab.
+
+## Deployment
+
+Use [wrangler](https://developers.cloudflare.com/workers/cli-wrangler) to build and deploy your application to Cloudflare Workers. If you don't have it yet, follow [the installation guide](https://developers.cloudflare.com/workers/cli-wrangler/install-update) to get it setup. Be sure to [authenticate the CLI](https://developers.cloudflare.com/workers/cli-wrangler/authentication) as well.
```
GregBrimble, line 26: Deployment of Pages projects is currently only possible with our git provider integrations. We're looking at supporting uploads from wrangler soon, but it's not ready just yet.

packages/create-remix/templates/cloudflare-pages/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```sh
+# in one tab, start the remix dev server
+$ npm run dev
+
+# in another, start the miniflare server
+$ npm start
+```
+
+Open up [http://127.0.0.1:8787](http://127.0.0.1:8787) and you should be ready to go!
```
GregBrimble, line 20: `wrangler pages dev` runs on `127.0.0.1:8788` by default.

packages/remix-cloudflare-pages/magicExports/platform.ts
```diff
@@ -0,0 +1,4 @@
+// This file lists all exports from this package that are available to `import
+// "remix"`.
+
+export { createCloudflareKVSessionStorage } from "@remix-run/cloudflare-workers";
```
GregBrimble, line 4: Think this should be `@remix-run/cloudflare-pages` if you're copying that over to this package as well.

packages/remix-cloudflare-pages/worker.ts
```diff
@@ -0,0 +1,96 @@
+import type { ServerBuild, AppLoadContext } from "@remix-run/server-runtime";
+import { createRequestHandler as createRemixRequestHandler } from "@remix-run/server-runtime";
+
+export interface CreateRequestHandlerParams<Env = any> {
+  build: ServerBuild;
+  getLoadContext?: (context: EventContext<Env, any, any>) => AppLoadContext;
+  mode?: string;
+}
+
+function createInternalHandler<Env = any>({
```
GregBrimble, line 10: Y'all call this `createRequestHandler` in the `@remix-run/cloudflare-workers` package.

packages/remix-cloudflare-pages/worker.ts
```diff
@@ -0,0 +1,96 @@
...
+    );
+  }
+
+  const response = await envWithAssets.ASSETS.fetch(request);
+  if (response.ok) return response;
+}
+
+declare const process: any;
+
+export function createRequestHandler<Env = any>({
```
GregBrimble, line 45: I think `createFetchHandler` is a slightly more correct description for this function.

packages/remix-cloudflare-pages/worker.ts
```diff
@@ -0,0 +1,96 @@
...
+  };
+}
+
+type EnvWithAssets = Record<string, unknown> & {
+  ASSETS: { fetch: typeof fetch };
+};
+
+async function handleAsset(request: Request, env: unknown) {
+  let envWithAssets = env as EnvWithAssets;
+  if (!envWithAssets?.ASSETS?.fetch) {
```
GregBrimble, line 33: This check isn't required. We'll always populate `env.ASSETS.fetch`, and you're casting its type above anyway.

packages/remix-cloudflare-pages/worker.ts
```diff
@@ -0,0 +1,96 @@
...
+  if (response.ok) return response;
+}
+
+declare const process: any;
+
+export function createRequestHandler<Env = any>({
+  build,
+  getLoadContext,
+  mode,
+  disableAssets
```
GregBrimble, line 49: Why would someone ever want to `disableAssets`? Is that something new in Remix?

packages/remix-cloudflare-pages/worker.ts
```diff
@@ -0,0 +1,96 @@
...
+        : undefined;
+    return handleRequest(context.request, loadContext);
+  };
+}
+
+type EnvWithAssets = Record<string, unknown> & {
+  ASSETS: { fetch: typeof fetch };
+};
+
+async function handleAsset(request: Request, env: unknown) {
```
GregBrimble, line 31: The `@remix-run/cloudflare-workers` package exports this asset handler and the request handler function. The docs show a 'simplified' version which uses the primary export and a 'low-level' version where developers can customize the entrypoint more themselves. Are you moving away from that 'low-level' option by not exporting these same functions here?

packages/remix-cloudflare-workers/package.json
jacob-ebey: I didn't notice any breaking changes in the types, but I might have missed something. Seems like they are just additions.

packages/remix-cloudflare-pages/worker.ts
jacob-ebey: I wanted to support some sort of multi function Pages deployment but on second thought I don't think I need or want to do that. Exposing the underlying handler factory like the workers runtime seems to be a better approach.

packages/remix-cloudflare-workers/package.json
MichaelDeBoey: It doesn't matter if the API didn't change, people using `@remix-run/cloudflare-workers` would need to use a new major version of `@cloudflare/workers-types` so this will be a breaking change (as is every change to the `peerDependencies`)

packages/remix-cloudflare-pages/worker.ts
GregBrimble: Ah, sorry! I didn't realize you were using `functions/[[path]].ts`. `createRequestHandler` is totally fine! I was thinking you were using `public/_worker.js`.

packages/create-remix/templates/cloudflare-pages/functions/[[path]].js
```diff
@@ -0,0 +1,15 @@
...
+// @ts-ignore
+import * as build from "../build";
+
+const handleFetch = createRequestHandler({
+  build
+});
+
+export function onRequest(context) {
+  const request = new Request(context.request);
+  request.headers.delete("If-None-Match");
```
Comment (GregBrimble) [inline packages/create-remix/templates/cloudflare-pages/functions/[[path]].js:12]: We'll make sure this works before GA. Sorry!

mjackson (changes requested):
Looks good. A few small changes, but I'm fine to ship this w/out abstracting into a separate remix-cloudflare.

packages/create-remix/templates/cloudflare-pages/functions/[[path]].js
Comment (GregBrimble) [reply packages/create-remix/templates/cloudflare-pages/functions/[[path]].js:12]: Just made one [over here](https://github.com/cloudflare/wrangler2/issues/117). Pretty certain this is something on our side, but not exactly sure where, so we'll take a look and circle back once we know what's going on.

packages/create-remix/templates/cloudflare-workers/package.json
```diff
@@ -13,7 +13,7 @@
     "@remix-run/cloudflare-workers": "*"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^2.2.2",
+    "@cloudflare/workers-types": "^3.2.0",
```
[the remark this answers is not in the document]
GregBrimble: Feel free to grab [the PagesFunction declaration from our repo](https://github.com/cloudflare/workers-types/blob/master/manual-ts/pages.d.ts) in the meantime.

packages/remix-cloudflare-pages/worker.ts
jacob-ebey: There isn't a separate package required to serve assets anymore and it's a simple couple liner so I'm not going to export this. We will still document the lower level handler pattern.

packages/create-remix/templates/cloudflare-workers/package.json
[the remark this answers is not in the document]
jacob-ebey: We can use the latest for the pages package as it is totally separate from the workers package at the moment.

mjackson (approved):
Looks good to me!

PR #986, opened 2021-12-11
Docs: Fixed paths in Tailwind example.

andrioid:
Nothing too serious. I was following the Tailwind guide on the site and the paths are wrong in the `package.json` script.

The "review the CLA" link is broken btw. If this goes through smoothly, I'll be happy to fix that too.


kentcdodds (changes requested):
This is correct as is. Thanks though!

PR #989, opened 2021-12-11
feat(dev): `ignoredRouteFiles` config option

chaance:
This PR introduces a new configuration API for ignoring non-route files in `app/routes`. This would allow users to keep CSS or test files adjacent to their route modules. 

```js
// remix.config.js
module.exports = {
  // ...
  // new optional config option
  ignoredRouteFiles: ["*.css"]
};
```

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -85,10 +85,11 @@ export interface AppConfig {
   serverModuleFormat: "esm" | "cjs";
 
   /**
-   * A filename or a glob pattern to match files in the `app/routes` directory
-   * that Remix will ignore. Matching files will not be recognized as routes.
+   * A list of filenames or a glob patterns to match files in the `app/routes`
+   * directory that Remix will ignore. Matching files will not be recognized as
+   * routes.
    */
-  ignoredRouteFiles?: string;
+  ignoredRouteFiles?: string[];
 }
 
 /**
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -83,6 +83,12 @@ export interface AppConfig {
    * The output format of the server build. Defaults to "cjs".
    */
   serverModuleFormat: "esm" | "cjs";
+
+  /**
+   * A filename or a glob pattern to match files in the `app/routes` directory
+   * that Remix will ignore. Matching files will not be recognized as routes.
+   */
+  ignoredRouteFiles?: string;
 }
 
 /**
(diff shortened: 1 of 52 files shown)


mcansh (approved):

packages/remix-dev/config.ts
```diff
@@ -83,6 +83,12 @@ export interface AppConfig {
    * The output format of the server build. Defaults to "cjs".
    */
   serverModuleFormat: "esm" | "cjs";
+
+  /**
+   * A filename or a glob pattern to match files in the `app/routes` directory
+   * that Remix will ignore. Matching files will not be recognized as routes.
+   */
+  ignoredRouteFiles?: string;
```
kentcdodds, line 91: Could we make this an array of globs instead?

kentcdodds (commented):
This looks good code-wise with one change. Needs consensus on including the feature in the first place and could use tests

kentcdodds (approved):
Looks good to me. Once we decide we want this I can add some docs. Cc @mjackson and @ryanflorence 
I'm in favor of this change.

ryanflorence:
I've hesitated on this for months because I anticipated doing fancy stuff with `.css` or `.pdf`, etc. in the routes folder.

But now that we have resource routes, that seems covered.

My real question now is why have it be configuration at all? Just ignore everything except `{js,jsx,ts,tsx,md,mdx}`

ryanflorence:
oh right, test files ...

chaance:
Another thing I think that is nice about the config option is that if we do decide to do something fancy with certain file types in the future, it wouldn't break existing apps that have those file types ignored. They could simply opt-in by removing them from the ignore list.

kentcdodds:
We talked about this a bit and we're thinking that the dotfiles stuff we did earlier should actually be removed (before it's released) and people can ignore dotfiles through this. The templates could be updated to ignore dotfiles via this API as well. Can you do that @chaance?

chaance:
TBH I'm not sure how I feel about this, but I made the changes to all configs in the project to match the template defaults. Seems to add more boilerplate just to avoid a tiny bit of magic 🤷‍♂️

packages/remix-dev/config/routesConvention.ts
```diff
@@ -21,20 +22,24 @@ export function isRouteModuleFile(filename: string): boolean {
...
 
   // First, find all route modules in app/routes
   visitFiles(path.join(appDir, "routes"), file => {
-    let routeId = createRouteId(path.join("routes", file));
-
-    if (isRouteModuleFile(file)) {
-      files[routeId] = path.join("routes", file);
+    if (
+      ignoredFilePatterns &&
+      ignoredFilePatterns.some(pattern => minimatch(file, pattern))
```
MichaelDeBoey, line 35: No need for a check if you give `ignoredFilePatterns` a default value of `[]`

```suggestion
      ignoredFilePatterns.some(pattern => minimatch(file, pattern))
```

packages/remix-dev/config/routesConvention.ts
```diff
@@ -21,20 +22,24 @@ export function isRouteModuleFile(filename: string): boolean {
  * For example, a file named `app/routes/gists/$username.tsx` creates a route
  * with a path of `gists/:username`.
  */
-export function defineConventionalRoutes(appDir: string): RouteManifest {
+export function defineConventionalRoutes(
+  appDir: string,
+  ignoredFilePatterns?: string[]
```
MichaelDeBoey, line 27:
```suggestion
  ignoredFilePatterns?: string[] = []
```

MichaelDeBoey (changes requested):

kentcdodds (approved):
seal of approval

PR #997, opened 2021-12-12
docs: fix api/remix.md extra semicolon

dokeet:
Also added `Form` import to make copy & paste easier

Screen Shot 2021-12-11 at 23 06 43


kentcdodds (approved):
Thanks!

Issue #769, opened 2021-11-29
[Bug]: Layout routes on same directory level causes build failure


sjones6:
### Which Remix packages are impacted?

- [X] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.0.6

### Steps to Reproduce

See repro case here: https://github.com/sjones6/duplicate_template_route_repro

After running `npm i`, try `npm run dev` and you should get a build error: 

```
npm run dev

> remix-app-template@ dev /Users/spencer/Code/template_repro
> remix dev

Path undefined defined by route "routes/__foo" conflicts with route "routes/__bar"
npm ERR! code ELIFECYCLE
npm ERR! errno 1
```

### Expected Behavior

Application builds, and routes in nested template directories are registered.

### Actual Behavior

Build step errors and fails.

abn5x:
Can confirm!

Weyker:
fixed on #1347

sjones6:
Closing, as this has been fixed referencing a duplicate issue.

PR #1011, opened 2021-12-12
fix: Fix error when defining multiple layout routes

universse:
`uniqueRouteId` is an empty string instead of `undefined` for layout routes.
Fix https://github.com/remix-run/remix/issues/769

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -76,6 +76,7 @@
 - tjefferson08
 - twhitbeck
 - tylerbrostrom
+- universse
 - UsamaHameed
 - veritem
 - VictorPeralta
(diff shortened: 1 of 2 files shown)


ryanflorence:
@jacob-ebey I didn't dig in, but seems like even layout routes should have a unique ID (filename)? Why would it be undefined or empty string?

ryanflorence:
Oops, didn't mean to submit yet. @jacob-ebey If you can take a look at this and approve it then I'll merge it, thanks!

Weyker:
fixed on #1347

ryanflorence:
thanks for taking some time on this!

PR #1015, opened 2021-12-12
build(fly): build before deploy

CanRau:
includes the changes of KCD from https://github.com/remix-run/remix/pull/930

from: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type
from: https://www.conventionalcommits.org/en/v1.0.0/
from: https://cheatography.com/albelop/cheat-sheets/conventional-commits/

from: https://gist.github.com/lisawolderiksen/a7b99d94c92c6671181611be1641c733

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -5,6 +5,7 @@
 - ashocean
 - BasixKOR
 - bruno-oliveira
+- CanRau
 - chaance
 - christophgockel
 - coryhouse
--- a/packages/create-remix/templates/fly/package.json
+++ b/packages/create-remix/templates/fly/package.json
@@ -3,7 +3,7 @@
     "postinstall": "remix setup node",
     "build": "remix build",
     "dev": "remix dev",
-    "deploy": "flyctl deploy",
+    "deploy": "npm run build && fly deploy --remote-only",
     "start": "remix-serve build"
   },
   "dependencies": {


kentcdodds (commented):
This PR needs to be made to `dev` ratchet than the `main` branch. Not sure I think we should build before the deploy. What do we do in the other templates? Either way it should be the same with all the templates.

CanRau:
Yea sorry I was kinda distracted by making my first PR via gh-cli and forgot to add my reasoning

And for the branch, there isn't a way to move this right? So I close this one and make a new one on the other branch right?

kentcdodds:
That would be an easier way, yeah

CanRau:
Alright, before opening the new one I thought it might be better to check in, as we already have this discussion, sharing my thinking.

So, to me it was quite confusing, that running `npm run deploy` wouldn't update my website 😳 I know it's kinda my fault not properly checking what the npm scripts are doing, but there's also no mention of ever running `npm run build` in the Readme.

Therefore I propose either adding it to the readme or running it automatically before deploy.

Personally I prefer the latter as I don't have much reason to deploy without building, but there might be other preferences I guess.
As for what other templates are doing, the Workers one [builds already before deploying](https://github.com/remix-run/remix/blob/main/packages/create-remix/templates/cloudflare-workers/package.json#L10) and others don't require builds or deploys, as far as I can tell.

So closing this one, awaiting feedback on if and how to make a new PR to `dev`

kentcdodds:
Thanks for the feedback. I agree we should make a change. I just want to make sure we're the same in every template. Could you open an issue about it so it can be discussed and a decision can be made?

PR #1021, opened 2021-12-13
Add baseUrl to tsconfig.json for VSCode autocomplete

weavdale:
Add `"baseUrl": "./"` to tsconfig.json so VSCode will autocomplete `~` path alias.

### DIFF
--- a/packages/create-remix/templates/_shared_ts/tsconfig.json
+++ b/packages/create-remix/templates/_shared_ts/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "include": ["remix.env.d.ts", "**/*.ts", "**/*.tsx"],
   "compilerOptions": {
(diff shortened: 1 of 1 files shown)


packages/create-remix/templates/_shared_ts/tsconfig.json
```diff
@@ -11,7 +12,7 @@
     "strict": true,
     "baseUrl": ".",
```
kentcdodds, line 13: Won't this conflict? Which is correct?

weavdale: Ah my bad! I should have checked updated dev branch before submitting the PR. Didn't see `"baseUrl"` got added with PR #482. That works great also. I'll close my PR.

PR #1027, opened 2021-12-13
refactor(types): Allow meta to return undefined

donavon:
This is allowed (obviously) and works fine when using JavaScript. This allows us to do the same when using TypeScript.

Use this when you have a meta handler, but for some reason you conditionally set the meta information.


donavon:
I've already signed the CLA in #940

sergiodxa (approved):

PR #1039, opened 2021-12-14

MichaelDeBoey:
To make @kentcdodds happy

https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

---

Same as #536


kentcdodds (approved):
Sure. But don't feel compelled to make this change unless you're in the code anyway. It's not *that* important to me.

Thanks for taking the time!

kentcdodds:
Also, docs and examples changes can happen in `main` unless they depend on unreleased changes.

PR #1040, opened 2021-12-14
refactor: use optional chaining

MichaelDeBoey:


mcansh:
node 14 according to MDN! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

ryanflorence:
`onRebuildStart?.()` <-- that one freaks me out.

mcansh:
> `onRebuildStart?.()` <-- that one freaks me out.

same to a degree, I usually reach for 
```ts
if (typeof onRebuildStart === 'function') {
	onRebuildStart()
}
```
or similar in these cases

PR #1047, opened 2021-12-14
chore: remove husky and lint staged, rely on github action

mcansh:
REM-605


mcansh:
we can't commit to some one's repo, can we?

mjackson:
We aren't interested in making sure PRs are formatted correctly. If people have prettier installed and configured correctly, great. Otherwise when that work gets merged into main/dev it will be automatically formatted.

PR #905, opened 2021-12-05
fix: dev process can't exit after using prisma

yesmeck:
prisma is not compatible with signal-exit, because prisma installs a exit handler to `process` which [prevented](https://github.com/tapjs/signal-exit/blob/2d6a8a29bd8d4265ad769b772415bb85c6b5be8b/index.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L118-L119) signal-exit to emit exit signal

related issue https://github.com/prisma/prisma/issues/2917

fix #752 and #762

PR #1048, opened 2021-12-14
Mention in the Jokes tutorial how to stop the process

sergiodxa:
When using Prisma and you stop the Remix server, it left the port 8002 open and you had to manually close it, this is because of Prisma, adding this to the `utils/db.server.ts` file before the export close the process correctly.


yesmeck:
I submit another approach to fix this issue, see #905.

sergiodxa:
@yesmeck yours is better since it fix the problem directly instead of documenting it, I will close this then.

PR #1049, opened 2021-12-14
update remix to use react router 6.1.1

jeremyjfleming:
Update remix to use react router version 6.1.1 which includes the useOutletContext functionality.


MichaelDeBoey:
Code changes should go against `dev` branch

This is already included there btw.

PR #1067, opened 2021-12-15
fix: update pages handler to use middleware pattern for asset handling

jacob-ebey:
fix: add missing content types (known CF issue)
chore: update template; missing dep and config
doc: added a little MD file to document adding a package

### DIFF
--- a/packages/ADDING_A_PACKAGE.md
+++ b/packages/ADDING_A_PACKAGE.md
@@ -1,7 +1,7 @@
 # How to add a package
 
-[] add folder and make sure it's a unique "name" field in package.json
-[] update rollup config to build the package
-[] add root tsconfig.json entry and make sure pkg tsconfig is proper output dir
-[] add to publish scriptS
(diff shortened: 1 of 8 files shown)


packages/ADDING_A_PACKAGE.md
```diff
@@ -0,0 +1,7 @@
+# How to add a package
+
+[] add folder and make sure it's a unique "name" field in package.json
+[] update rollup config to build the package
+[] add root tsconfig.json entry and make sure pkg tsconfig is proper output dir
+[] add to publish scriptS
+[] update version script
```
MichaelDeBoey, line 7: Checkboxes need to be like

```suggestion
- [] add folder and make sure it's a unique "name" field in `package.json`
- [] update rollup config to build the package
- [] add root `tsconfig.json` entry and make sure pkg tsconfig is proper output dir
- [] add to publish scripts
- [] update version script
```

Issue #1056, opened 2021-12-15
[Bug]: express server can't start in production because `remix setup node` didn't run


smhg:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [X] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.0.6

### What version of Node are you using? Minimum supported version is 14.

16.13.1

### Steps to Reproduce

```bash
npx create-remix@latest
// choose express, javascript, run npm install
cd [app-name]
npm run postinstall
npm run build

// 'simulate' production environment:
rm -rf node_modules
npm i --production

npm run start
```

### Expected Behavior

It should be able to start the express server.

### Actual Behavior

```bash
Error: Cannot find module './client'
Require stack:
- my-remix-express-app/node_modules/remix/index.js
- my-remix-express-app/server/build/index.js
- my-remix-express-app/server/index.js
```

This can be solved with `npm install @remix-run/dev` and `remix setup node` (which was the case before `rm -rf node_modules` and `npm i --production` happened).

kiliman:
What Node environment are you running? I use Heroku for simple test projects and it works fine in "production" mode.

Here's my deployment log
```log
-----> Installing dependencies
       Installing node modules
       
       > remix-component-data@0.2.0 postinstall
       > remix setup node
       
       Successfully setup Remix for node.
       ...
-----> Build
       Running build
       
       > remix-component-data@0.2.0 build
       > remix build
       
       Building Remix app in production mode...
       Built in 406ms
       
-----> Caching build
       - node_modules
       
-----> Pruning devDependencies
       
       up to date, audited 96 packages in 1s
       
       found 0 vulnerabilities
       
-----> Build succeeded!
...
-----> Launching...
       Released v8
       https://remix-component-data.herokuapp.com/ deployed to Heroku
```
I'm not sure what exactly Heroku means by *Pruning devDependencies* but I don't think they are nuking node_modules and reinstalling as --production.

jacob-ebey:
[...]

PR #1068, opened 2021-12-16
fix: runnable node bundle without setup command

jacob-ebey:
This removes direct `import abc from "remix"` and delegates to the underlying package that owns the import. So instead of `let { json } = require("remix")` in the server bundle, it would be `let { json } = require("@remix-run/server-runtime")`.


jacob-ebey:
Fixes https://github.com/remix-run/remix/issues/1056

mcansh:
👻 #1026

ryanflorence:
superseded by #1026

PR #1071, opened 2021-12-16
feat: mainfield config based on serverModuleFormat

jacob-ebey:


kentcdodds (commented):
Seems sensible to me. Though it does feel odd to list "browser" in the possible mains for a server build... 🤔

jacob-ebey:
> Seems sensible to me. Though it does feel odd to list "browser" in the possible mains for a server build... 🤔

The thought here is that a ESM "browser" build is a more appropriate format for an ESM bundle than a CJS "main" field.

kentcdodds (approved):
Seems solid 👍 Thanks!

PR #1075, opened 2021-12-16
chore(create-remix): remove Netlify node version step

ascorbic:
Netlify's default Lambda runtime [is now Node 14](https://answers.netlify.com/t/functions-node-14-will-be-the-new-default-runtime/47224), so the step setting the version is no longer needed


mcansh:
🎉 thanks @ascorbic

PR #1085, opened 2021-12-16
chore: Clean up test output in gists app

chaance:
This should clean up our CI logs a bit and make it easier to spot real problems in tests!

### DIFF
--- a/fixtures/gists-app/server.js
+++ b/fixtures/gists-app/server.js
@@ -5,6 +5,32 @@ const { createRequestHandler } = require("@remix-run/express");
 
 const port = process.env.PORT || 3000;
 
+if (process.env.NODE_ENV === "test") {
+  class MockConsole {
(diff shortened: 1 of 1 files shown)


mcansh (approved):

kentcdodds:
When a test fails, will we still get the output we need to know why?

chaance:
> When a test fails, will we still get the output we need to know why?

Yes, we aren't mocking or overriding the console in the test env itself, just the app. In Jest the console should stay in tact.

kentcdodds:
Ok. I'm just concerned that if something breaks *within* the app there will be `console.error` logs that would help us diagnose the issue but we won't see them. This is why I typically assert that `console.error` wasn't called and if it was then I log what it was called with so we can use that to help use figure out why the assertion failed. Not sure how to do that with puppeteer though.

PR #1086, opened 2021-12-16
chore(ci): don't format dev branch as it's protected

mcansh:


mcansh:
we figured out a workaround https://github.com/remix-run/remix/pull/1455

PR #1092, opened 2021-12-16
feat: hide CF bug fix in adapter

jacob-ebey:
fix: update start command

### DIFF
--- a/packages/remix-cloudflare-pages/worker.ts
+++ b/packages/remix-cloudflare-pages/worker.ts
@@ -41,6 +41,11 @@ export function createFetchHandler<Env = any>({
   const handleFetch = async (context: EventContext<Env, any, any>) => {
     let response: Response | undefined;
 
+    const request = new Request(context.request);
+    // https://github.com/cloudflare/wrangler2/issues/117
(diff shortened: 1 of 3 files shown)


packages/remix-cloudflare-pages/worker.ts
```diff
@@ -41,6 +41,11 @@ export function createFetchHandler<Env = any>({
   const handleFetch = async (context: EventContext<Env, any, any>) => {
     let response: Response | undefined;
 
+    const request = new Request(context.request);
+    // https://github.com/cloudflare/wrangler2/issues/117
```
kentcdodds, line 45: This indicates that the issue is during dev only. Should we do an env check here?

kentcdodds (approved):
So we basically want this behavior for all CFP apps then? Sounds fine to me 👍

packages/remix-cloudflare-pages/worker.ts
jacob-ebey: It's a production issue as well. They are investigating and we should have this here for now regardless of local or prod.

PR #1095, opened 2021-12-17
Add more content types to extractData

donavon:
`extractData` currently only handles "application/json", otherwise it returns plain text. This PR adds support for "multipart/form-data", "text/event-stream", and "application/x-www-form-urlencoded".

Q: is there an easy way to share this function between client and server?


jacob-ebey:
This is actually not where you want this functionality. `extractData` is only used for extracting JSON loader data for serialization in the HTML response for hydration. It doesn't make sense to extract and serialize form data and other types.

PR #1097, opened 2021-12-17
refactor: use url.href vs url.toString()

donavon:
The code used `url.href` in places and `url.toString()` in others. Be consistant throughout and use the more concise `url.href`.


ryanflorence:
> yeah but `url.href` could be re-assigned and `url.toString()` can't!

PR #1102, opened 2021-12-17
refactor(types): remove superfluous union types

donavon:
Once you add "string" to a union type, any string literal values are ignored. 

Here you see autocomplete of type `Foo` with "bar" and "foo". Cool.

image

But once you add `| string`, we no longer get autocomplete.

image

I moved the literal values into a JSDoc comment as I felt they still had merit, but removed them from the union type itself.

### DIFF
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -26,23 +26,11 @@ export interface HtmlLinkDescriptor {
   crossOrigin?: "anonymous" | "use-credentials";
 
   /**
-   * Relationship between the document containing the hyperlink and the destination resource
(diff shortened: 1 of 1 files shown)


MichaelDeBoey:
Shouldn't we remove the `string` type and keep the individual ones?

This will cause more maintenance, but more type-safety

mcansh:
we can use `LiteralUnion` from https://github.com/sindresorhus/type-fest/blob/main/source/literal-union.d.ts to keep both as well :)

MichaelDeBoey:
@mcansh Seems even better imo!

PR #1105, opened 2021-12-17
docs: fix incorrect method name

Ehesp:
The `method` should be lowercase, currently it's uppercase so TS complains.

```ts
export declare type FormMethod = "get" | "post" | "put" | "patch" | "delete";
```


ryanflorence:

PR #1114, opened 2021-12-17
chore: change template to bare bones

ryanflorence:
Right now the starter template is a bit overwhelming and requires you to delete some folders and sort through layers of composition in root.tsx. For a newcomer it's unclear what is essential and what is just starter stuff (are `Document` and `Layout` magic Remix things or not?).

While all the demos are useful, the immediate feeling for most people unfamiliar with Remix is "holy crap that's a lot of code already, Remix seems hard." When really it only requires the two entries and a root route.

Our Quickstart and Deep Dive tutorials are a great way to introduce Remix already. The bare bones template will help folks not feel overwhelmed.

Of course, we know that they will want their own ErrorBoundary and their own CatchBoundary, etc. I made a previous commit that makes the defaults a bit better for a production setting, so even if they don't provide one, the default isn't as terrible. Folks will learn through the docs to add their own.

I still want a kitchen sink demo. But it needs to be a CLI option or just a separate repo. For now, let's keep it simple.


tannerlinsley:
This was a good call 💯

PR #1121, opened 2021-12-17
feat: Example updating the form with the correct initial value on route transition

Girish21:
Example to update the `form` with the correct initial values on route transition using the `key` prop

@kentcdodds have taken a stab at it (as mentioned here https://youtu.be/s0Oe5l3voHU?t=4528)


kentcdodds (changes requested):
Thanks for working on this! I'd really like to keep the examples minimal. So if we don't need a *real* database then we shouldn't bother including it. Even styles don't need to be anything fancy (in some cases you shouldn't even need styles at all).

Also I want to keep the examples consistent so let's not customize things like prettier or eslint please.

Would you mind simplifying this quite a bit?

girishk21:
Hey @kentcdodds have made the required changes
- removed Prisma and DB
- removed styles
- updated Remix packages to 1.1.1
Let me know if anything else is required, thanks!

Girish21:
Hey @kentcdodds have made the required changes
- removed Prisma and DB
- removed styles
- updated Remix packages to 1.1.1
Let me know if anything else is required, thanks!

examples/outlet-form-rerender/app/routes/users/$userId.tsx
```diff
@@ -0,0 +1,70 @@
...
+     * defaultValue, React will initialise the input field with the
+     * initial value and leave the subsequent updates uncontrolled
+     *
+     * To rerender the input field with the updated value on the route
+     * transition, we have to inform React to unmount and remount the component
+     * which resets the initial values
+     *
+     * https://reactjs.org/docs/uncontrolled-components.html
+     * */
+    <form method="post" key={user.id}>
```
kentcdodds, line 44: Let's use a remix form here.

Let's also make note in the comment that the key prop is what makes this work. 

Also, could you update the README to explain the example a bit and tell where the relevant bits of code are?

examples/outlet-form-rerender/app/routes/users.tsx
```diff
@@ -0,0 +1,34 @@
...
+interface LoaderData {
+  users: User[];
+}
+
+export const meta: MetaFunction = () => {
+  return { title: "Users" };
+};
+
+export const loader: LoaderFunction = async () => {
+  return { users } as LoaderData;
```
kentcdodds, line 16:
```suggestion
  return json<LoaderData>({ users });
```

And we'll need to import json above

examples/outlet-form-rerender/app/routes/users/$userId.tsx
```diff
@@ -0,0 +1,70 @@
...
+export const loader: LoaderFunction = async ({ params }) => {
+  const userId = params.userId;
+
+  const user = users.find(({ id }) => id === userId);
+
+  if (!user) {
+    throw json(null, { status: 404 });
+  }
+
+  return { user } as LoaderData;
```
kentcdodds, line 27:
```suggestion
  return json<LoaderData>({ user });
```

kentcdodds (changes requested):
Looks great! Just a few things

examples/outlet-form-rerender/app/routes/users/$userId.tsx
Girish21: hey @kentcdodds, have updated the README here https://github.com/remix-run/remix/pull/1121/commits/46823b189ca1a51c54b61a0ae4ffe70ebb02c52e. And have also added a note that the `key` prop does the magic of updating the Form with the new values here https://github.com/remix-run/remix/pull/1121/commits/a57543bb36fa7dd1f624b4d18e491462c84a74c2#diff-0db1e51405f843f23e57a234a98186c4009f1f66dd8b4af5543e0ac8cf508b76R44. Thanks!

kentcdodds (approved):
This is awesome, thank you!

PR #1122, opened 2021-12-17
chore: Bump react-router-dom dependency

chaance:
Also fixed some of the broken types as a result of the History changes that should improve type narrowing a bit.


jacob-ebey (approved):

mcansh (approved):

kumard3:
Need a help , is their any way i can disable filesystem routing and using react router dom ?

PR #1123, opened 2021-12-17
chore(react): update links rel and as unions to LiteralUnion for better IntelliSense

mcansh:
x-ref https://github.com/remix-run/remix/pull/1102#issuecomment-996850247


donavon:
Suggest changing the title to "chore(react): update links rel and as unions to LiteralUnion for better IntelliSense" because as we still allow `string`, it has nothing to do with type safety.

ryanflorence:
instead of another dependency, can we just do whatever type trick `LiteralUnion` does?

mcansh:
> instead of another dependency, can we just do whatever type trick `LiteralUnion` does?

done :)

PR #1124, opened 2021-12-17
chore: update netlify functions deps

jacob-ebey:


kentcdodds (approved):
Sounds good to me 👍

packages/remix-netlify/package.json
```diff
@@ -16,9 +16,9 @@
     "@remix-run/server-runtime": "1.0.6"
   },
   "peerDependencies": {
-    "@netlify/functions": "^0.7.2"
+    "@netlify/functions": "^0.10.0"
```
MichaelDeBoey, line 19: This is a breaking change, so this should have published `@remix-run/netlify@2.0.0`

kentcdodds: Good point. I'll bring this up with the team. We need to discuss how to handle stuff like this in the future...

MichaelDeBoey: The current CI only supports having **all** packages the same version.
This isn't a problem if we had done `^0.7.2 || ^0.10.0` in this case, but once we start using a new API we should at least update `@remix-run/netlify` to v2

PR #1133, opened 2021-12-18
fix: export UploadHandler & UploadHandlerArgs from node package

Ehesp:
The `UploadHandler` & `UploadHandlerArgs` types are not exported from the node package.

I'm not sure how to get them to export from the main `remix` package since the types seem generated. I locally patched that using `patch-package` which seems to work for now. Please let me know if anything else needs changing.


Vivalldi (approved):

ryanflorence:
@jacob-ebey you mind verifying if we should be exporting these types?

Ehesp:
@ryanflorence They're mentioned in the docs: https://remix.run/docs/en/v1/api/remix#custom-uploadhandler

```ts
import type { UploadHandler } from "remix";

export let action: ActionFunction = async ({ request }) => {
  const userId = getUserId(request);
  let uploadHandler: UploadHandler = async ({
    name,
    stream
  }) => {
```

Also if someone want's to build their own upload handler they'll most likely want to use the `UploadHandlerArgs ` type to provide a familiar API.

CanRau:
VSC just auto-imported for me `import { UploadHandler } from "@remix-run/node/formData";`
Though at least the docs would need to be updated

jacob-ebey (approved):

jacob-ebey:
@Ehesp I've pulled this off into a PR here https://github.com/remix-run/remix/pull/1913 as the "magic" portion had to be added and the tests are failing here for some reason.

PR #1135, opened 2021-12-18
const > let

donavon:
This is really just a shade PR for @ryanflorence so please, feel free to reject it out of hand. Have codemod, will travel.

image


chaance:
No 💜

PR #1147, opened 2021-12-19
docs: fix a typo in the example of the resource routes docs

manosim:
It looks like there is a misleading dot in the docs for resource routes. I realised that the example didn't work properly and upon reading the instructions (just above the example), I got the feeling that there is an extra dot there that breaks it.

Removing the dot fixed the route for me. Feel free to close this one if it's invalid. 😄  

...and most importantly, thank you for making Remix! I've been really impressed by how simple it is to use and understand. Liking it even more day by day! Thank you 👏


kentcdodds (approved):
Thanks 👍

PR #1151, opened 2021-12-19
(fix:docs) Changes to the highlighted code in the Jokes tutorial

jesse-deboer:
Added the right highlighting to the Jokes tutorial


ryanflorence:
I think this may have been fixed elsewhere, if not, please submit a new PR (but this time make it against `main` since that's where we add docs updates 😁)

Issue #1148, opened 2021-12-19
useFetcher load() shows confusing message when invoked improperly


arvigeus:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [X] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [X] `@remix-run/vercel`

### What version of Remix are you using?

1.1.1

### What version of Node are you using? Minimum supported version is 14.

14.18.2

### Steps to Reproduce

```tsx
// routes/index.tsx
import { useFetcher } from "remix";

export default function Index() {
  const fetcher = useFetcher();
  fetcher.load("/data");
  if (!fetcher.data) return null;
  return <div>{JSON.stringify(fetcher.data)}</div>
}
```

### Expected Behavior

Error that hints the incorrect usage of `fetcher.load()`.

This code is taken from the [example docs](https://remix.run/docs/en/v1/api/remix#fetcherload). Addressed in ##1152

### Actual Behavior

`ReferenceError: AbortController is not defined`

geocine:
I am actually calling `fetcher.load()` outside of a hook and it works well but I call it inside an event handler like so
```jsx
<button onClick={() => fetcher.load("/data")} />
```

zachdtaylor:
@geocine yeah, as long as `fetcher.load()` isn't called during the server render, it's fine. So calling it in a callback like you have there is the correct way to use it 🙂

kiliman:
```ts
const fetcher = useFetcher();
useEffect(() => {
  if (fetcher.type === 'idle') {
    fetcher.load("/data");
  } else if (fetcher.type === 'done') {
    console.log(fetcher.data)
  }
}, [fetcher]);
```
fetcher.load should be called in an event handler or useEffect().

machour:
This is now properly documented, thank you @zachdtaylor !

PR #1152, opened 2021-12-19
docs: fix misleading example usage of useFetcher load()

arvigeus:
This is made clear later in docs, but it's easy to confuse someone (me!) that `fetcher.load('/some/route')` can be called outside a hook, and the resulting error message is not helping at all to resolve the problem.

Related: #1148


docs/api/remix.md
```diff
@@ -757,7 +757,10 @@ Loads data from a route loader.
 
 ```tsx
 const fetcher = useFetcher();
-fetcher.load("/some/route");
+React.useEffect(() => {
+  if (fetcher.type === 'init')
+    fetcher.load('/some/route');
```
kentcdodds, line 762: We'll probably have fewer people freak out at us if we include braces 😅 Thanks!

```suggestion
  if (fetcher.type === 'init') {
    fetcher.load('/some/route');
  }
```

kentcdodds (approved):
Nice, thanks!

PR #1161, opened 2021-12-20
Docs: Fixes error in useSubmit example

craigglennie:
Was passing `e.target` to `useSubmit`, which is incorrect and won't type-check. The other example of `useSubmit` on this page uses `e.currentTarget`, as do the [docs for useSubmit](https://remix.run/docs/en/v1/api/remix#usesubmit)


kentcdodds (approved):
Thanks!

PR #1166, opened 2021-12-20
Specifying that the devServerPort is specifically for the websocket server

DavidHollins6:
I spent a lot of time wondering why I was getting a `426 Upgrade Required` response back from my local application when running `remix dev`. I had made the assumption that `devServerPort` was the port of my applications server, but it is in-fact the port of the websocket server. This PR specifies in the docs that for anyone in the future who might make the same assumption.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -128,3 +128,4 @@
 - yesmeck
 - zachdtaylor
 - zainfathoni
+- DavidHollins6
--- a/docs/api/conventions.md
+++ b/docs/api/conventions.md
@@ -79,7 +79,7 @@ The path to the server build, relative to remix.config.js. Defaults to "build". 
 
 ### devServerPort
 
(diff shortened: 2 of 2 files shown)


kentcdodds (approved):
Thanks! 👍

almamariali:
def fibonacci(n)
  return n if ( 0..1 ).include? n
  (fibonacci(n - 1) + fibonacci(n - 2)) #recursive calls
end

almamariali (commented):
def fibonacci(n)
  return n if ( 0..1 ).include? n
  (fibonacci(n - 1) + fibonacci(n - 2)) #recursive calls
end

almamariali:
packages/remix/package.json

almamariali:
packages/remix/package.json

almamariali:
docs/api/conventions.md

almamariali (commented):
Leave a comment

PR #1167, opened 2021-12-20
docs: replace remix dev with remix watch

Glup3:
I get the error `Could not locate @remix-run/serve. Please verify you have it installed to use the dev command.` when using `remix dev`. This should be replaced with `remix watch`.


ryanflorence:
It really all depends on if you're using the remix app server or not, so it's wrong for somebody no matter what we use so we'll leave it as it is. Thank you though!

PR #1182, opened 2021-12-21
chore(examples): add sass example

mcansh:


sergiodxa (approved):
Looks great!

examples/with-sass/README.md
```diff
@@ -0,0 +1,55 @@
+# Welcome to Remix!
```
kentcdodds, line 1: Could you update this to use the same README format as in the template?

examples/with-sass/app/routes/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+          <a
+            target="_blank"
+            href="https://remix.run/docs"
+            rel="noreferrer noopener"
+          >
+            Remix Docs
+          </a>
+        </li>
+      </ul>
+    </div>
```
kentcdodds, line 34: Could you remove the irrelevant stuff that can be removed (like this bit) and comment the primary relevant bits?

kentcdodds (changes requested):
Thanks for this! In addition to the comments here, let's remove the unnecessary `with-` prefix. That's what the Next.js repo does and it's just extra fluff that's not necessary.

examples/with-sass/README.md
mcansh: yep sorry, made this before we had a template 😅

kentcdodds (approved):
Thanks!

mcansh:
haha, just realized i accidentally removed the css import from root.tsx 🤦

kentcdodds:
Whoops 😅 that might be *too* minimal 🤣 I didn't notice. Could you get that fixed?

mcansh:
> Whoops 😅 that might be _too_ minimal 🤣 I didn't notice. Could you get that fixed?

🤣 - #1467

PR #1184, opened 2021-12-21
Fix header casing windows cloudflare pages

efkann:
On Windows, `If-None-Match` header causes an error, this PR changes the header to the lowercase version.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -20,6 +20,7 @@
 - dunglas
 - dwt47
 - edgesoft
+- efkann
 - eps1lon
 - evanwinter
 - fergusmeiklejohn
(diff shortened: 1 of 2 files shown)


GregBrimble:
Worked through this with @efkann .

The Turkish lowercase version of `I` is `ı` not `I`.

It looks like there's something going on with undici's header normalization (maybe https://github.com/nodejs/undici/issues/1119 ?). This was recently fixed, but until that reaches miniflare/wrangler, simply lowercasing headers first seems to work.

GregBrimble (approved):

kentcdodds (approved):
This is fine 👍

kentcdodds:
Thank you both for taking the time!

PR #1207, opened 2021-12-23
feat(remix-cloudflare-workers): support ES Module

### LINKED ISSUE #764: [Feature]: Cloudflare Workers ES Modules Support

[Cloudflare Workers has recently GA'd support for ES modules](https://blog.cloudflare.com/workers-javascript-modules/) and now uses this format as the default in templates.

ES modules brings three changes that affect Remix:

1. **The entry point syntax is different**

    Where before, you set up a Worker using the service workers syntax, you now export a fetch handler:

    ```js
    // Previous service worker syntax

    addEventListener('fetch', (event) => {
      event.respondWith(async () => {
        return new Response("Hello, world!")
      })
    })

    // New ES modules syntax

    export default {
      async fetch(request, env, context) {
        return new Response("Hello, world!")
      }
    }
    ```

1. **Bindings are no longer global**

    Previously, any bindings of your Worker were available in the global scope. This pollution was part of problem with the service worker syntax, since there was no 'neat' way to pass them to a handler. Now, these bindings are available on the `env` (environment) object.

    ```js
    // Previous service worker syntax

    addEventListener('fetch', (event) => {
      event.respondWith(async () => {
        const value = await MY_KV_NAMESPACE.get("key")
        return new Response("Value retrieved from KV: " + value)
      })
    })

    // New ES modules syntax
    
    export default {
      async fetch(request, env, context) {
        const value = await env.MY_KV_NAMESPACE.get("key")
        return new Response("Value retrieved from KV: " + value)
      }
    }
    ```

1. **Finally, Durable Objects are available when using ES modules syntax**

    [Durable Objects](https://developers.cloudflare.com/workers/runtime-apis/durable-objects) are powerful database primitives which I personally think are better suited to session storage than KV. KV is eventually consistent, so competing requests could easily squash values. Durable Objects, on the other hand, act as a point of synchronicity, as only one instance (for a given identifier) can ever be in existence. This controller can act as a gatekeeper, and removes the possibility for mistakes if you race data writes. Durable Objects are automatically created near to the user who requested them, and they can migrate as needed. They can also be restricted to a given jurisdiction, which is really useful (particularly for user data) in the context of GDPR etc.

- Durable Objects support would bring a great alternative to KV for `SessionStorage`.
- ES modules syntax is the default for Cloudflare Workers going forward. Adopting this would mean less friction for users who want to adapt their Worker, and also means documentation can be more easily understood between Cloudflare and Remix.
- Full-stack Cloudflare Pages uses ES modules syntax, which would offer Remix users another target to deploy to, which offers a variety of new functionality such as rollbacks, preview URLs and git provider integrations.

I have already written an adapter for ES modules, a Cloudflare Pages project, and Durable Objects-backed SessionStorage. We'd need to work out a couple of things to finalize how that gets integrated (e.g. how a developer could specify a jurisdiction for a Durable Object), but it's pretty much there.

ekosz:
Hi there,

This is my first contribution to the Remix project so please let me know what additions you would like to see from this PR 😄 . 

I've very interested in using features like [Durable Objects](https://developers.cloudflare.com/workers/learning/using-durable-objects) in my Cloudflare workers, but to use them we need the ability to publish ESM versions of our workers. Remix has already(?) shipped the ability to generate native ESM, but we still need to update the `@remix/cloudflare-workers` package. 

This change adds three commits to help us get there. Each commit explains what it does, but happy to further explain here as well.

Cheers!

P.S. This also fixes a bug where if the result of the handler is a `Promise<Response>` that rejects, it wouldn't have been caught by the try catch block due to the promise not being awaited.

Closes #764

### DIFF
--- a/packages/remix-cloudflare-workers/worker.ts
+++ b/packages/remix-cloudflare-workers/worker.ts
@@ -10,6 +10,12 @@ import {
 import type { AppLoadContext, ServerBuild } from "@remix-run/cloudflare";
 import { createRequestHandler as createRemixRequestHandler } from "@remix-run/cloudflare";
 
+export type Evt<Env = unknown> = {
+  request: Request
+  waitUntil: (promise: Promise<any>) => void
+  env: Env,
+}
+
 /**
  * A function that returns the value to use as `context` in route `loader` and
  * `action` functions.
@@ -17,7 +23,7 @@ import { createRequestHandler as createRemixRequestHandler } from "@remix-run/cl
  * You can think of this as an escape hatch that allows you to pass
  * environment/platform-specific values through to your loader/action.
  */
-export type GetLoadContextFunction = (event: FetchEvent) => AppLoadContext;
+export type GetLoadContextFunction = (event: FetchEvent | Evt) => AppLoadContext;
 
 export type RequestHandler = ReturnType<typeof createRequestHandler>;
 
@@ -36,7 +42,7 @@ export function createRequestHandler({
 }) {
   let handleRequest = createRemixRequestHandler(build, mode);
 
-  return (event: FetchEvent) => {
+  return (event: FetchEvent | Evt) => {
     let loadContext = getLoadContext?.(event);
 
     return handleRequest(event.request, loadContext);
@@ -44,7 +50,7 @@ export function createRequestHandler({
 }
 
 export async function handleAsset(
-  event: FetchEvent,
+  event: FetchEvent | Evt,
   build: ServerBuild,
   options?: Partial<KvAssetHandlerOptions>
 ) {
@@ -111,7 +117,7 @@ export function createEventHandler({
     mode,
   });
 
-  let handleEvent = async (event: FetchEvent) => {
+  let handleEvent = async (event: FetchEvent | Evt) => {
     let response = await handleAsset(event, build);
 
     if (!response) {
@@ -121,20 +127,28 @@ export function createEventHandler({
     return response;
   };
 
-  return (event: FetchEvent) => {
+  return async (event: FetchEvent | Evt) => {
+    const respondWith = (payload: Response) => {
+      if ('respondWith' in event && typeof event.respondWith === 'function') {
+        event.respondWith(payload);
+        return;
+      }
+      return payload;
+    };
+
     try {
-      event.respondWith(handleEvent(event));
+      const response = await handleEvent(event);
+      return respondWith(response);
     } catch (e: any) {
       if (process.env.NODE_ENV === "development") {
-        event.respondWith(
+        return respondWith(
           new Response(e.message || e.toString(), {
             status: 500,
           })
         );
-        return;
       }
 
-      event.respondWith(new Response("Internal Error", { status: 500 }));
+      return respondWith(new Response("Internal Error", { status: 500 }));
     }
   };
 }
--- a/templates/cloudflare-workers/README.md
+++ b/templates/cloudflare-workers/README.md
@@ -35,3 +35,29 @@ Once that's done, you should be able to deploy your app:
 ```sh
 npm run deploy
 ```
+
+## Using [ES Module workers](https://blog.cloudflare.com/workers-javascript-modules/)
+
+If you would like to use features like [Durable
+Objects](https://developers.cloudflare.com/workers/learning/using-durable-objects)
+you will need to slightly change the [`worker/index.js`](./worker/index.js)
+file. In ES Module workers we no longer have a `FetchEvent` object and will
+need to generate something similar ourselves instead.
+
+```js
+import { createEventHandler } from "@remix-run/cloudflare-workers";
+
+import * as build from "../build";
+
+const handler = createEventHandler({ build })
+
+export default {
+  fetch(request, env, context) {
+    const event = { request, env, waitUntil: context.waitUntil.bind(waitUntil) };
+    return handler(event);
+  }
+}
+```
+
+**Note**: This will now be the new object passed to you if you add
+a `getLoadContext` function to the `createEventHandler`.
--- a/.changeset/dull-kings-hide.md
+++ b/.changeset/dull-kings-hide.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/cloudflare-workers": minor
+---
+
+Allow for ESM based Cloudflare Workers
--- a/contributors.yml
+++ b/contributors.yml
@@ -114,6 +114,7 @@
 - edgesoft
 - edmundhung
 - efkann
+- ekosz
 - eldarshamukhamedov
 - emzoumpo
 - eps1lon
--- a/packages/remix-cloudflare-workers/package.json
+++ b/packages/remix-cloudflare-workers/package.json
@@ -15,7 +15,7 @@
   "typings": "dist/index.d.ts",
   "module": "dist/esm/index.js",
   "dependencies": {
-    "@cloudflare/kv-asset-handler": "^0.1.3",
+    "@cloudflare/kv-asset-handler": "^0.2.0",
     "@remix-run/cloudflare": "1.7.2"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -1845,6 +1845,13 @@
   dependencies:
     mime "^2.5.2"
 
+"@cloudflare/kv-asset-handler@^0.2.0":
+  version "0.2.0"
+  resolved "https://registry.yarnpkg.com/@cloudflare/kv-asset-handler/-/kv-asset-handler-0.2.0.tgz#c9959bbd7a1c40bd7c674adae98aa8c8d0e5ca68"
+  integrity sha512-MVbXLbTcAotOPUj0pAMhVtJ+3/kFkwJqc5qNOleOZTv6QkZZABDMS21dSrSlVswEHwrpWC03e4fWytjqKvuE2A==
+  dependencies:
+    mime "^3.0.0"
+
 "@cloudflare/workers-types@^3.4.0":
   version "3.4.0"
   resolved "https://registry.npmjs.org/@cloudflare/workers-types/-/workers-types-3.4.0.tgz"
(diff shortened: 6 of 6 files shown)


ekosz:
@ryanflorence Is there something I can do to improve this PR? Or is there a separate work stream happening to improve the Cloudflare story?

Just wondering if I should continue to open these types of PRs right now. Thanks!

ryanflorence:
> Is there something I can do to improve this PR? 

Sorry, just been getting the company organized and the framework off the ground! We're finally able to spend some time in the PRs/Issues.

I'm going to assign this @jacob-ebey, I'm not as familiar w/ cloudflare workers.

ekosz:
@MichaelDeBoey I've updated the PR (better 10 months later than never). This was my first time using changesets so I would love a review to make sure I did it correctly.

Thank you and excited to finally get this merged as this has been blocker for us for a long time now

MichaelDeBoey:
@ekosz It seems like @jacob-ebey has a solution that only involves changes to the template in #4198.

Maybe it's a good idea to join forces on this & suggest changes to `@remix-run/cloudflare-workers` if still needed? 🤔

ekosz:
@MichaelDeBoey That seems like a much better change than this one? I tried to keep this as contained as possible. Does it seem like that PR have a high likelihood of landing soon? If so it can supersede this one. Otherwise if it seems that PR may take a while to land may I suggest this be merged in the meantime? This is something people have wanted for a while.

packages/remix-cloudflare-workers/worker.ts
```diff
@@ -10,14 +10,20 @@ import {
...
+
 /**
  * A function that returns the value to use as `context` in route `loader` and
  * `action` functions.
  *
  * You can think of this as an escape hatch that allows you to pass
  * environment/platform-specific values through to your loader/action.
  */
-export type GetLoadContextFunction = (event: FetchEvent) => AppLoadContext;
+export type GetLoadContextFunction = (event: FetchEvent | Evt) => AppLoadContext;
```
huw, line 26: I think it would make sense to parameterise `<Environment>` for this & dependent functions, to allow the user to easily type it. So:

```ts
export type GetLoadContextFunction<Environment = unknown> = (event: FetchEvent | Evt<Environment>) => AppLoadContext;
```

templates/cloudflare-workers/README.md
```diff
@@ -35,3 +35,29 @@ Once that's done, you should be able to deploy your app:
...
+export default {
+  fetch(request, env, context) {
+    const event = { request, env, waitUntil: context.waitUntil.bind(waitUntil) };
+    return handler(event);
+  }
+}
+```
+
+**Note**: This will now be the new object passed to you if you add
+a `getLoadContext` function to the `createEventHandler`.
```
huw, line 63: I would probably just document this in the above example, it’s not clear what ‘this’ refers to in this sentence until you inspect the types in TypeScript. And then this + the changes should be added to the Remix docs.

templates/cloudflare-workers/README.md
```diff
@@ -35,3 +35,29 @@ Once that's done, you should be able to deploy your app:
...
+```js
+import { createEventHandler } from "@remix-run/cloudflare-workers";
+
+import * as build from "../build";
+
+const handler = createEventHandler({ build })
+
+export default {
+  fetch(request, env, context) {
+    const event = { request, env, waitUntil: context.waitUntil.bind(waitUntil) };
```
huw, line 56: `waitUntil` is undefined in this context. What did you mean?

packages/remix-cloudflare-workers/worker.ts
```diff
@@ -121,20 +127,28 @@ export function createEventHandler({
...
   };
 
-  return (event: FetchEvent) => {
+  return async (event: FetchEvent | Evt) => {
+    const respondWith = (payload: Response) => {
+      if ('respondWith' in event && typeof event.respondWith === 'function') {
+        event.respondWith(payload);
+        return;
+      }
+      return payload;
```
huw, line 136: The mixed return here is _tolerable_ (since Wrangler/Miniflare will run without typechecking), but it does mean that the fetch handler no longer conforms to `ExportedHandlerFetchHandler` from `@cloudflare/workers-types`. The code is similar enough that it would be annoying to split it into 2 separate functions, but I gave it a crack and couldn’t figure out how to use conditional types in a way that resolved it properly.

packages/remix-cloudflare-workers/worker.ts
```diff
@@ -36,15 +42,15 @@ export function createRequestHandler({
...
     return handleRequest(event.request, loadContext);
   };
 }
 
 export async function handleAsset(
-  event: FetchEvent,
+  event: FetchEvent | Evt,
   build: ServerBuild,
   options?: Partial<KvAssetHandlerOptions>
 ) {
```
huw, line 56: Right below this, you need to pass [`ASSET_NAMESPACE` and `ASSET_MANIFEST`](https://github.com/cloudflare/kv-asset-handler#asset_namespace-required-for-es-modules) to `getAssetFromKV`. Otherwise Cloudflare won’t be able to bind the KV to the worker.

huw (commented):
I gave this a crack on my machine—it’s still missing a few things before it will work. I think this is a preferable solution to the template change with a few tweaks. (I am not a Remix contributor, just a dev that also wants this 😌)

templates/cloudflare-workers/README.md
```diff
@@ -35,3 +35,29 @@ Once that's done, you should be able to deploy your app:
...
+If you would like to use features like [Durable
+Objects](https://developers.cloudflare.com/workers/learning/using-durable-objects)
+you will need to slightly change the [`worker/index.js`](./worker/index.js)
+file. In ES Module workers we no longer have a `FetchEvent` object and will
+need to generate something similar ourselves instead.
+
+```js
+import { createEventHandler } from "@remix-run/cloudflare-workers";
+
+import * as build from "../build";
```
dario-piotrowicz, line 50: shouldn't this be:
```suggestion
import * as build from "@remix-run/dev/server-build";
```
?

templates/cloudflare-workers/README.md
petebacondarwin:
```suggestion
    const event = { request, env, waitUntil: context.waitUntil.bind(context) };
```

ekosz:
With the new comments on this PR, does that mean @MichaelDeBoey that the Remix team wants to go this direction instead? I just don't want to address these comments just to have the PR closed

dario-piotrowicz:
> With the new comments on this PR, does that mean @MichaelDeBoey that the Remix team wants to go this direction instead? I just don't want to address these comments just to have the PR closed

Hi @ekosz sorry about that, me and @petebacondarwin were just trying to get Remix to work with Cloudflare Worker ESM mode as that would be a very valuable thing to have, we've stumbled across this PR without noticing that there was already  https://github.com/remix-run/remix/pull/4198 (and there seem to even be a third one https://github.com/remix-run/remix/pull/4401 😨).

I'll be checking out https://github.com/remix-run/remix/pull/4198 next to see if that solution work 🙂, as of this PR unfortunately it doesn't seem to work in its current state, because of the 2 comments we've added + another issue with the `ASSET_NAMESPACE` and `ASSET_MANIFEST` not being correctly passed to `getAssetFromKV` (but I totally understand you not fixing all the issue while being unsure of the state of this PR).

dario-piotrowicz:
> @MichaelDeBoey That seems like a much better change than this one? I tried to keep this as contained as possible. Does it seem like that PR have a high likelihood of landing soon? If so it can supersede this one. Otherwise if it seems that PR may take a while to land may I suggest this be merged in the meantime? This is something people have wanted for a while.

As you mentioned @ekosz this is something that people would have needed for a while and as time progresses more Cloudflare features assume and work better with the use of the (preferred) ESM mode, so it would be very good to have this issue fixed soon, @MichaelDeBoey please let me know if I can help in any way as I would be very happy to do so 🙏

templates/cloudflare-workers/README.md
dario-piotrowicz: I likely misunderstood the code here, my bad 😅

MichaelDeBoey:
@ekosz @huw @dario-piotrowicz I'm wondering if you could obtain the same result with the new config properties that were exposed by #4841 🤔

MichaelDeBoey:
@ekosz @huw @dario-piotrowicz I'm wondering if you could obtain the same result with the new config properties that were exposed by #4841 🤔

huw:
@MichaelDeBoey Nope, still need to add an additional handful of files ([see solution here](https://github.com/remix-run/remix/pull/4676#issuecomment-1423663413))—the new config options helped a tonne though and ultimately unblocked that work. The most idiomatic solution should be my one there, or the [update this morning](https://github.com/remix-run/remix/pull/4676#issuecomment-1529613908) from @xanderberkein.

To make progress on this we just need agreement on who should own the adapter package and those discussions never happened (but you & I don’t work for CF or Remix so our options are limited here). IMHO this PR should be closed in favour of having that discussion, if we can ever get it to happen.

MichaelDeBoey:
@huw We also still have @jacob-ebey's #4198 & #6229, which adds ESM-support to the CF Workers template without the need of changing the adapter

Could those be possible solutions as well you think?

huw:
Yep, I weighed in on that debate [here](https://github.com/remix-run/remix/pull/4198#issuecomment-1323611620) last time it came up (most of my points still apply, but the situation has improved a bit with the new config properties).

Given how much time has passed without much of a discussion between Remix & CF, it’s probably better to just merge #6229 when it’s done. I would _prefer_ we create a new package for the reasons in that comment, but given that that seems politically impossible right now (not enough interest from either side), updating the template would at least provide user value sooner. I’ll leave some suggestions on that PR :)

MichaelDeBoey:
@huw For what it's worth: I thought @jacob-ebey was also planning on only supporting ESM in v2 (or at least making it the default), so I guess the changes that are done in this PR could come in handy when implementing that 🤔

xanderberkein:
I didn't realize the debate on who should own the ES Modules adapter was still ongoing. I just released my own adapter yesterday: https://github.com/xanderberkein/remix-cloudflare-module-workers.

From a DX perspective, I agree with @huw: I'd prefer a separate adapter (or even better - an update to the current `cloudflare-workers` adapter) that abstracts away the boilerplate otherwise needed in the template. 

The service worker syntax does nothing you can't do with module workers, so there is no real need to keep the old service worker syntax around. As you can see from my adapter, the only change needed to convert the current `cloudflare-workers` adapter from a service worker to a module worker, is a few minor changes to the `createEventHandler` (https://github.com/xanderberkein/remix-cloudflare-module-workers/blob/main/src/worker.ts#L16), and some small adjustments to the template. 
The breaking change when switching to module workers would be that people would need to manually follow [the three steps described in the readme of my adapter](https://github.com/xanderberkein/remix-cloudflare-module-workers#usage).

If updating the current `cloudflare-workers` is not an option, I'd still vote for a separate `module-worker` adapter then.

brookslybrand:
I'm going to go ahead and close this PR since the original issue (#764) is closed by #6650

Thanks @ekosz for opening this up so long ago

PR #1216, opened 2021-12-23
fix: multiple Flat Layout Routes

### LINKED ISSUE #1210: [Bug]: Flat Layout Routes are not working as expected

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [X] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/vercel`

1.1.1

16.13.1

### Steps to Reproduce

If we try this [Example](https://remix.run/docs/en/v1/api/conventions#flat-layout-routes) the `routesConvention.js` trow this error 

> Path undefined defined by route "routes/__app" conflicts with route "routes/__marketing"

i see this is because the routes have the same `uniqueRouteId` = '' (empty string)

I want to have multiple layouts and nested flat layouts something like this:

```
__public/
 * * home.tsx
 * * __auth/
 * * * * login.tsx
 * * * * register.tsx
__private/
 * * admin.tsx
 __public.tsx
 __private.tsx
 ```
 
 if i do this 2 things happen:

   1. Not work because the problem described up  
   2. the __auth Flat Layout its not in the routes so `login` and `register` routes are not created

### Expected Behavior

work like the example and have multiple Flat Routes inside the Flat Routes

### Actual Behavior

Error on routes

Weyker:
this close #1210

fullPath is allways undefined when is a Flat Layout Routes and its not necessary to mark as unique route

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -87,5 +87,6 @@
 - VictorPeralta
 - yesmeck
 - weavdale
+- Weyker
 - zachdtaylor
 - zainfathoni
--- a/packages/remix-dev/config/routesConvention.ts
+++ b/packages/remix-dev/config/routesConvention.ts
@@ -70,7 +70,7 @@ export function defineConventionalRoutes(
       let fullPath = createRoutePath(routeId.slice("routes".length + 1));
       let uniqueRouteId = (fullPath || "") + (isIndexRoute ? "?index" : "");
 
-      if (typeof uniqueRouteId !== "undefined") {
+      if (typeof uniqueRouteId !== "undefined" && fullPath !== undefined) {
         if (uniqueRoutes.has(uniqueRouteId)) {
           throw new Error(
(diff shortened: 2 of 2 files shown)


ryanflorence:
Thanks @Weyker :)

@jacob-ebey does this look right to you?

ryanflorence:
We need a test for this, just add two sets of pathless layout routes in the gist app and load them up.

Weyker:
> We need a test for this, just add two sets of pathless layout routes in the gist app and load them up.

@ryanflorence it's that for me or @jacob-ebey?

Weyker:
fixed on #1347

PR #1226, opened 2021-12-24
docs: fix inconsistency in cookie options expires and maxAge

karimsan:
There is a misleading inconsistency in `createCookie` and `createCookieSessionStorage` documentation. It confuses seconds for `maxAge` and milliseconds for `expires` configuration by using the value 60 in both cases.
 - `maxAge` accepts value in seconds . Value `expires: 60` represents one minute.
 - `expires` accepts Date. Date calculation with Date.now() requires using milliseconds. To set the expiration value for one minute the could would look like this: `expires: new Date(Date.now() + 60_000)` instead of `new Date(Date.now() + 60)`.

Current version of documentation sets `maxAge` equal to 60 (one minute) but for `expires` value is sets 60 millisecond from now (instead of `60_000` milliseconds to make it one minute). Suggested PR fixes the inconsistency so that for both properties documentation example defines one minute expiration.

In the `remix-server-runtime` [source code](https://github.com/remix-run/remix/blob/1fd70960e4d88740df5bf407a6ba2cd2b9549459/packages/remix-server-runtime/cookies.ts#L84-L87) the calculation is correct and default value is calculated from maxAge by multiplying by 1000.


kentcdodds (approved):
Thank you!

PR #1230, opened 2021-12-24
Add how to solve Unsupported ESM URL Scheme Error in Cloudflare Pages deployment

supachaidev:
Cloudflare uses Node version 12 as default, so we will get an error from some dependencies at build time.

So, it is recommended to set NODE_VERSION to 14 to the Pages production environments for consistency.


ryanflorence:
Not sure I understand this, Cloudflare doesn't run Node.js. Is this a CI thing?

supachaidev:
> Not sure I understand this, Cloudflare doesn't run Node.js. Is this a CI thing?

Yeah, it is a CI thing when Pages run `npm run build`.

Pages Node default version is 12.18.0. So, we need to explicitly set the environment variable `NODE_VERSION` to a more recent version.

supachaidev:
I've changed the explanation to be concise. Take a look at [this commit](https://github.com/remix-run/remix/pull/1230/commits/6ded0e4e1debb65b4fb5d0eaab0be5189495f3fe?short_path=a1a6b30#diff-a1a6b30e92592d5427b0c38406cafd3a5405667dde336db5fa38c33462338464).

jacob-ebey:
There is a `.node-version` in the pages template that takes care of this already but maybe it's good to call it out?

ryanflorence:
Oh, I added a `.node-version` to your branch but we already have that on main (like Jacob said). I don't think we need to call anything out, it will only break if people delete the file :)

PR #1246, opened 2021-12-26
Show error when using useLoaderData but no loader is exported

nicastelo:
I'm probably overthinking it, but I feel loader could be easily mistyped and lead to many minutes of "WTF is this not working".

The message is probably not the best and maybe a `console.warn` works better in this scenario?

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -64,6 +64,7 @@
 - mskoroglu
 - msutkowski
 - nareshbhatia
+- nicastelo
 - niconiahi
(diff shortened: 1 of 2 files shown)


ryanflorence:
This is a pretty solid idea but I'm not 100% sure we want to do it. There may be times where an abstraction calls `useLoaderData` and decides to do one thing or another depending on getting `undefined` or not.

I've also not really heard of folks bumping into this?

PR #1260, opened 2021-12-28
docs(blog): fix diff highlights in tutorial for getPost function

roddds:
This bit was a little confusing to me (but also, granted, it was very late at night).


ryanflorence:
hey sorry for taking a while to get to this, you mind doing this again but branch from the `main` branch? Docs updates go to `main` because then they sync to the website. Thanks!

kentcdodds (approved):
You're right, that was weird. Let's just remove the highlight altogether. Doesn't make sense to highlight basically the whole block. https://youtu.be/GYmHYQPaHaw?t=46 😅

Thanks!

PR #1264, opened 2021-12-28
add: configurable live reload host

hedwiggggg:
LiveReload seems to be limited to localhost only, because the host is hardcoded. I created a hotfix by adding an additional host parameter to the LiveReload component.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1270,13 +1270,23 @@ export function useFetchers(): Fetcher[] {
   return [...fetchers.values()];
 }
 
-export function LiveReload({ port = 8002 }: { port?: number }) {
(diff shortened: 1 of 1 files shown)


hedwiggggg:
Hi @sairaj2119, sure! Right now I'm developing on a remote server by sshing into it with vscode. Everything works fine except LiveReload, because when I visit the site on http://server:3000 it tries to connect to ws://localhost:8002, but it should connect to ws://server:8002. Keeping it simple, I thought the most obvious change would be to add a host parameter.

sairaj2119:
Makes sense, Thanks for explanation

ryanflorence:
fixed by #1685, sorry for not getting to this sooner :)

PR #1272, opened 2021-12-29
docs: update cloudflare KV example

mtt87:
When using Cloudflare KV storage to store a session it's important to use file name `sessions.server.ts` with the `.server` keyword in order for Remix to only bundle this code on the server bundle.
Spent a couple of hours trying to figure out why a route that has a `loader` and a `default` export with the UI would throw an error although the session is only handled on the loader (server side) and this was the solution.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -88,6 +88,7 @@
 - morinokami
 - mskoroglu
 - msutkowski
+- mtt87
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Thanks!

PR #1281, opened 2021-12-30
feat(remix-react): add `useRouteData` hook

sergiodxa:
In Remix Utils I created this [useRouteData hook](https://github.com/sergiodxa/remix-utils/tree/v2.4.0#useroutedata) which uses the `useMatches` hook to get the data of any route in the current URL

The way it works is you pass the route path and you get the data back, but if the route is a layout the path may be repeated, in that case I had to ask the devs to define an `id` in the `handle` export to search using that.

This PR moves this useRouteData hook and taking advantage of being able to access the internal context of Remix you can with this implementation pass the route ID (the file name without the extension), which is easy to know for the developer.

---

**Why is this hook useful?**

A super common requirement is to access the data of the parent route inside the child route, most people put that in the `<Outlet context />` but it makes no sense when Remix already have the data of the loader inside a context, using useMatches is a way to solve it in user-land but this hook has a simpler implementation and could be used out of the box.

And, contrarian to the Outlet solution, you can also access a child route data from a parent one, so you can communicate data in both directions from top to bottom or from bottom to top.


kyleharper:
I built this hook yesterday. 

+1 for getting it into the Remix Utils 👍

sergiodxa:
> I built this hook yesterday.
> 
> +1 for getting it into the Remix Utils 👍

This is already in Remix Utils, this PR is to add it directly in Remix which is better because it can use the internal context.

kentcdodds:
Another reason `<Outlet context />` isn't suitable for loader data is because you could have a situation where route `A` puts something into Outlet context which route `C` depends upon. Then later someone adds `context` to route `B` (which is between the two) and surprisingly/unexpectedly breaks route `C`.

I'm in favor of this improvement.

zachdtaylor:
+1, this would be awesome to have as part of Remix itself.

ryanflorence:
We only ever intended `useMatches` for use cases where layouts needed child route data to build things like breadcrumbs.

I'm not sure why we're so bothered by child routes accessing parent data with it 🙃

`useMatches` as an access point for "all route data" is dang convenient though, and this hook makes it even *more* convenient. When you make something more convenient it gets used more!

This whole thing reminds me of when we created React Broadcast and the React team was like "yeah, that's nice but it makes using context too convenient", and then they implementing it directly into React as `React.createContext()` the next release 🤣

@mjackson I want to get your thumbs up on this, but I think we should add it.

I've already made my own `useRootData()` and `useParentData()` hooks around `useMatches` in my Remix apps. I know it's not how we intended `useMatches` to be used but in practice it's turning out to be a really great way to share route data across the hierarchy.

I'd change this to just use route ids to access the data. We already exposed route ids as public API in `parentsData` of the `meta` function.

```js
export function meta({ parentsData }) {
  let root = parentsData["root"];
  let whatever = parentsData["routes/whatever"];
  // ...
}
```

This hook is not any different:

```js
export function Comp() {
  let { user } = useRouteData("root");
  let whatever = useRouteData("routes/whatever");
  // ...
}
```

It's far more convenient than `<Outlet context />` because you don't have to keep drilling the values down (and augmenting with the new data) at each level. Additionally, if Remix wasn't the one doing the fetching, your app code would already have access to all this data. It's not like we're leaking implementation details, the app defined the routes and supplied the data, we would just be exposing those app-supplied values with app-defined ids.

@sergiodxa if @mjackson gives the thumbs up, I'd like to see the hook just accept route ids and not require the handle or url paths.

kentcdodds:
I've also put a lot of thought into this and agree we should have this feature with the route ID as mentioned. We just need to take care in the docs to explain the trade-offs so people don't use it unnecessarily.

sergiodxa:
> @sergiodxa if @mjackson gives the thumbs up, I'd like to see the hook just accept route ids and not require the handle or url paths.

The hook of this PR uses the route ID, the hook of Remix Utils doesn’t because it was not possible.

andresgutgon:
+1 niceee

docs/api/remix.md
```diff
@@ -351,6 +351,29 @@ See also:
...
+}
+
+// ...
+```
+
+```tsx filename=app/routes/invoices/$invoice.tsx lines=[1,4]
+import { useRouteData } from "remix";
+
+export default function Invoices() {
+  const invoices = useRouteData<Invoice[]>("routes/parent");
```
machour, line 370:
```suggestion
  const invoices = useRouteData<Invoice[]>("routes/invoices");
```

machour:
> Right now, one would have to use the loader hook to request the data twice:

You can already use `useMatches()` and avoid requesting the data again.

Mange:
> > Right now, one would have to use the loader hook to request the data twice [in a resource route]:
> 
> You can already use `useMatches()` and avoid requesting the data again.

`useMatches` can only be used in a React context, not inside a loader function. Resource routes are only a loader function.

> It would help when sending data to a resource route whose parent already has that data.

Would it? This is still a hook, isn't it?

sergiodxa:
> +1 for this feature. It would help when sending data to a resource route whose parent already has that data. Right now, one would have to use the loader hook to request the data twice: once for the parent route and once for the resource route.

This will not solve that, because this is a React hook, it'll make it easier to get another route data on React but server-side you still need to get the same data on each loader it's needed.

machour:
@sergiodxa Can you fix conflicts pretty please?

sergiodxa:
@machour done

docs/api/remix.md
```diff
@@ -467,6 +467,31 @@ See also:
 - [`action`][action]
 - [`useTransition`][usetransition]
 
+### `useRouteData`
+
+This hook returns the JSON parsed data from the route loader function of any matching route in the current URL.
+
+```tsx filename=app/routes/invoices.tsx
+export function loader() {
```
MichaelDeBoey, line 475: Since we're in a TS file

```suggestion
export const loader: LoaderFunction = () => {
```

docs/api/remix.md
```diff
@@ -467,6 +467,31 @@ See also:
...
+  return fakeDb.invoices.findAll();
```
MichaelDeBoey, line 476: `action` & `loader` need to return `Response` or use `json` helper

```suggestion
  return json(await fakeDb.invoices.findAll());
```

fixtures/gists-app/app/routes/route-data.tsx
```diff
@@ -0,0 +1,22 @@
+import type { LoaderFunction } from "remix";
+import { Outlet, useRouteData } from "remix";
```
MichaelDeBoey, line 2:
```suggestion
import { json, Outlet, useRouteData } from "remix";
```

fixtures/gists-app/app/routes/route-data.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export let loader: LoaderFunction = async () => {
+  return { title: "Parent Route Data" };
```
MichaelDeBoey, line 5:
```suggestion
  return json({ title: "Parent Route Data" });
```

fixtures/gists-app/app/routes/route-data/index.tsx
```diff
@@ -0,0 +1,17 @@
+import type { LoaderFunction } from "remix";
+import { useRouteData } from "remix";
```
MichaelDeBoey, line 2:
```suggestion
import { json, useRouteData } from "remix";
```

fixtures/gists-app/app/routes/route-data/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export let loader: LoaderFunction = async () => {
+  return { title: "Child Route Data" };
```
MichaelDeBoey, line 5:
```suggestion
  return json({ title: "Child Route Data" });
```

fixtures/gists-app/tests/route-data-test.ts
```diff
@@ -0,0 +1,90 @@
...
+  let browser: Browser;
+  let page: Page;
+  beforeEach(async () => {
+    browser = await puppeteer.launch();
+    page = await browser.newPage();
+  });
+
+  afterEach(() => browser.close());
+
+  describe("the parent should render the child title", () => {
```
MichaelDeBoey, line 19: As @kentcdodds also requested in https://github.com/remix-run/remix/pull/2411#discussion_r830625426

> I think we’re trying to use Ryan’s `createFixture` utility as much as possible instead of the gist app. Could you switch this test to use that instead?

MichaelDeBoey (changes requested):

fixtures/gists-app/tests/route-data-test.ts
mjackson: Agreed. We are trying to get rid of the gists app for testing stuff and use a separate `createFixture` test instead. Thanks!

docs/api/remix.md
mjackson: I'm not 100% sold on this requirement. I know we already merged some docs changes that said we have to do this, but I'd like to find a way to still allow people to return whatever data they want and we just encode/decode for them.

kiliman: I think a lot of the confusion about typing loader return value and useLoaderData have to do with Remix allowing you to return a naked object from the loader.

I strongly recommend deprecating this and having them return an actual Response. That way there is no magic that leads to misunderstanding. They will know that loaders always return a Response.

bluefire2121:
> > +1 for this feature. It would help when sending data to a resource route whose parent already has that data. Right now, one would have to use the loader hook to request the data twice: once for the parent route and once for the resource route.
> 
> This will not solve that, because this is a React hook, it'll make it easier to get another route data on React but server-side you still need to get the same data on each loader it's needed.

Thanks @sergiodxa and @Mange for the reminder. You all are both correct. Hooks have to top level.

chaance:
Closing this not because we don't want the feature, but because we'll be introducing it into React Router instead (as well as many other Remix APIs). Stay tuned, and thanks @sergiodxa for the work here!

PR #1286, opened 2021-12-30
fix(docs): add missing await for External APIs

ValentinH:
image


docs/guides/data-loading.md
```diff
@@ -105,7 +105,7 @@ Remix polyfills the `fetch` API on your server so it's very easy to fetch data f
 
 ```tsx filename=app/routes/gists.jsx lines=[2]
 export function loader() {
```
ghoshnirmalya, line 107: If you are using `await`, then the function should have the `async` keyword.

```suggestion
export async function loader() {
```

ValentinH: This is totally right!

PR #923, opened 2021-12-06
feat: change devServer config structure to allow more options

ikhattab:
# Background

following the discussion in https://github.com/remix-run/remix/pull/664 , this PR is to refactor options used in dev server to allow adding more options without adding root keys.

https://github.com/remix-run/remix/pull/664#discussion_r761500087

> Let's be careful about adding root keys to the config. Can we try to come up with a key name for development overlay configuration? We will probably need a few more keys in there in the future for surfacing uncaught errors and whatnot.

The proposed change in this PR is to pass dev server options as

```
devServer:  {
  broadcastDelay: 0,
  port: 8002,
}
```

instead of  using `devServerBroadcastDelay` and `devServerPort`

- [x] `devServerBroadcastDelay` or `devServerPort` will be used if passed.
- [x] `console.warn` with a deprecation note if `devServerBroadcastDelay` or `devServerPort`  are passed.

PR #1293, opened 2021-12-30
Setup new devServerAdditionalDirectories config option

jfranciscosousa:
As per this [discord interaction](https://discord.com/channels/770287896669978684/778004202282287104/926070217904386068), Jacob suggested me to add a new config option to Remix to allow extra directories to be watched by its file watcher.

The problem arose during some experiments with Turborepo. A company I'm working with wants to setup a monorepo with all of the frontend-based repos and while trying out their examples I noticed that updates to a package were not triggering any rebuild on Remix, which makes a subpar dev experience. Imagine you are working on a design system, you would have to reboot the dev-server with every change to the design system package.

This PR adds a `devServerAdditionalDirectories` option to the Remix config. You can add absolute paths to directories you want to watch for changes. You can add `node_modules` for example, or any folder outside the `app` that traditional Remix projects have.

I then updated the part of the docs where we talk about the config, and added an example to the `gists-app`.

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -69,10 +69,16 @@ export interface AppConfig {
    * The port number to use for the dev server. Defaults to 8002.
    */
   devServerPort?: number;
+
   /**
    * The delay before the dev server broadcasts a reload event.
    */
   devServerBroadcastDelay?: number;
+
+  /**
+   * A list of absolute paths that Remix will include on it's file watcher.
+   */
+  devServerAdditionalDirectories?: string[];
 
   /**
    * Additional MDX remark / rehype plugins.
@@ -162,6 +168,11 @@ export interface RemixConfig {
   devServerBroadcastDelay: number;
 
   /**
+   * A list of absolute paths that Remix will include on it's file watcher.
+   */
+  devServerAdditionalDirectories: string[];
+
+  /**
    * Additional MDX remark / rehype plugins.
    */
   mdx?: RemixMdxConfig | RemixMdxConfigFunction;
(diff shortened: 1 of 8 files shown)


packages/remix-dev/config.ts
```diff
@@ -161,6 +167,11 @@ export interface RemixConfig {
    */
   devServerBroadcastDelay: number;
 
+  /**
+   * A list of absolute paths that Remix will include on it's file watcher.
+   */
+  devServerAdditionalDirectories: string[];
```
c43721, line 173: Did you mean to duplicaet this?

c43721, line 173: Did you mean to duplicate this?

c43721 (changes requested):
Found a typo I think

packages/remix-dev/config.ts
jfranciscosousa: It's two different types, one for the app config type which matches `remix.config.js` and another that matches the outputted internal config that's used by Remix

leon:
I'm using mdx-bundler and the mdx files live in a folder outside of app.
This would be a great addition and be a great companion to tailwinds `content` globs.

🚀

MichaelDeBoey:
If this feature is wanted, this should be implemented after #923 imo

jfranciscosousa:
Thanks, I'll keep a watch on that PR.

One thing that made me a little inactive here is that this works for files outside `app` nicely, but my monorepo use case still didn't work at all and in the meantime, I scrapped that idea. If this is useful for more people I'll update it once #923 lands

ryanflorence:
Thanks for the PR!

I'm pretty sure @jacob-ebey has done some work here, I'm closing because of that and hopefully he can point us to where the use cases you're after are (or will be met) :)

PR #1297, opened 2021-12-31
feat(remix-react): add `useIsHydrated` hook & `ClientOnly` component

### LINKED ISSUE #139: Document `<ClientOnly>` strategies
```tsx
// global hydrating so future components don't do this dance, they just render
let hydrating = true;

// Component that renders `null` on the server and the initial browser render
// (while it's "hydrating")
function ClientOnly({ children }) {
  // set initial state to whatever the global hydrating is, so once the app is
  // hydrated, these components just render, no dance
  let [hydrated, setHydrated] = useState(() => !hydrating);

  // after hydration, rerender
  useEffect(() => {
    hydrating = false;
    setHydrated(true);
  }, []);

  // only render if hydrated
  return hydrated ? children : null;
}
```

### LINKED ISSUE #180: Add `<ClientOnly>`
Lots of third party libs don't work with server rendering and we keep posting the same code in the discord to help out. Should probably just put it into Remix itself.

```tsx
// global hydrating so future components don't do this dance, they just render
let hydrating = true;

// Component that renders `null` on the server and the initial browser render
// (while it's "hydrating")
function ClientOnly({ children }) {
  // set initial state to whatever the global hydrating is, so once the app is
  // hydrated, these components just render, no dance
  let [hydrated, setHydrated] = useState(() => !hydrating);

  // after hydration, rerender
  useEffect(() => {
    hydrating = false;
    setHydrated(true);
  }, []);

  // only render if hydrated
  return hydrated ? children : null;
}
```

sergiodxa:
Right now Remix has an internal boolean called isHydrated, it use it to know when JS has already loaded and change the result of rendering the Scripts component.

This PRs adds two things:

## useIsHydrated

This hook expose the isHydrated boolean without being able to change it, it also runs a one-time only effect to change the boolean to `true` and update the internal state.

The way it works is:

1. On SSR the result is always `false`
2. On CSR, the first render is always `false`, then it's updated to `true`
3. Subsequent CSR will always return `true` and it will not cause a second render.

This hook is useful to implement the second change and to be able to change prop values once JS load, e.g. if you have `<button type="button" />` you can add `disable={!isHydrated}` so it's disabled until JS load, this way the button will not be interactive without JS. Other use cases are to disable a `<fieldset>` with `<input type="checkbox" />` that only works if JS is enabled with useFetcher or useSubmit.

## ClientOnly

Closes #139
Closes #180
Closes #2852

This component uses the `useIsHydrated` hook internally, it receives two props a `children` to be render only client-side (after JS loaded) and an optional `placeholder` to be rendered server-side useful to render something that uses the same space of the client only component so it doesn't cause a layout shift or to render a fallback.

Examples:
1. If you have an interactive Chart component that can't be rendered server-side wrap it with ClientOnly and use a placeholder of an empty box with the same size of the chart or render a static SVG that can still show the data while JS loads or in case it fails.
2. If you have a rich editor and you need to support non-JS users, render a textarea as the placeholder, that way the form is still usable (without rich editing features) and once JS load the textarea can be replaced.
3. If you have a third-party component that needs to be rendered only client-side because it access browser only APIs in the render (like localStorage to initialize a state or window or element), re-export it from a `.client.ts|js` file and wrap it in ClientOnly, that way you won't get an error because server-side the imported component will be `undefined`.
4. If you need to render a React Portal you will need to pass the DOM element used to attach the portal, wrap the `React.createPortal()` call in ClientOnly and use it safely.

---

This PR also deprecated https://github.com/sergiodxa/remix-utils#clientonly and https://github.com/sergiodxa/remix-utils#usehydrated from Remix Utils.

---

Docs are pending

### DIFF
--- a/fixtures/gists-app/app/routes/client-only.tsx
+++ b/fixtures/gists-app/app/routes/client-only.tsx
@@ -1,0 +1,23 @@
+import { ClientOnly, useIsHydrated } from "remix";
+
+export default function ClientOnlyRoute() {
+  let isHydrated = useIsHydrated();
+
+  return (
+    <>
+      <ClientOnly
+        fallback={<h1 data-test-id="server-only-title">Server-Side</h1>}
+      >
+        <h1 data-test-id="client-only-title">Client Side</h1>
+      </ClientOnly>
+
+      <button
+        type="button"
+        disabled={!isHydrated}
+        data-test-id="client-only-button"
+      >
+        I only work client side
+      </button>
+    </>
+  );
+}
--- a/fixtures/gists-app/app/routes/client-only.tsx
+++ b/fixtures/gists-app/app/routes/client-only.tsx
@@ -6,7 +6,7 @@ export default function ClientOnlyRoute() {
   return (
     <>
       <ClientOnly
-        fallback={<h1 data-test-id="server-only-title">Server-Side</h1>}
+        placeholder={<h1 data-test-id="server-only-title">Server-Side</h1>}
       >
         <h1 data-test-id="client-only-title">Client Side</h1>
       </ClientOnly>
--- a/fixtures/gists-app/tests/client-only-test.ts
+++ b/fixtures/gists-app/tests/client-only-test.ts
@@ -1,0 +1,70 @@
+import type { Browser, Page } from "puppeteer";
+import puppeteer from "puppeteer";
+
+import { disableJavaScript, getHtml, reactIsHydrated } from "./utils";
+
+const testPort = 3000;
+const testServer = `http://localhost:${testPort}`;
+
+describe("ClientOnly and useIsHydrated", () => {
+  let browser: Browser;
+  let page: Page;
+  beforeEach(async () => {
+    browser = await puppeteer.launch();
+    page = await browser.newPage();
+  });
+
+  afterEach(() => browser.close());
+
+  describe("the h1 should change if JS loaded or not", () => {
+    it("with javascript enabled", async () => {
+      await page.goto(`${testServer}/client-only`);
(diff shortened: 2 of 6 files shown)


ryanflorence:
I think I want to get these in, but I'm also afraid that if `useHydrated()` and `<ClientOnly>` is super easy to do, people will use them in tons of places they don't need to. Kinda want to let Remix bake a bit longer? I don't know!

sergiodxa:
> I think I want to get these in, but I'm also afraid that if `useHydrated()` and `<ClientOnly>` is super easy to do, people will use them in tons of places they don't need to. Kinda want to let Remix bake a bit longer? I don't know!

I’m ok with keeping them on Remix Utils more time too, people is using them a lot mostly to support third party component using browser APIs in the render or in the module code combined with the .client extension.

The reason to send the PR is mostly because of that, people use them a lot already so they could probably be made official, there’s even an example I created using it from Remix Utils.

Also I think the hook is probably the one that is most useful to support progressive enhancement, so you can disable buttons needing JS to work before JS loads, and you can easily implement the component using the hook.

MichaelDeBoey:
@sergiodxa Could you please rebase your branch & fix conflicts?

MichaelDeBoey:
@sergiodxa It seems like we have some conflicts to resolve here

MichaelDeBoey:
@sergiodxa Could you please rebase your branch onto latest `dev` & resolve conflicts?

chaance:
@sergiodxa I'm going to close this for now, though we may reopen in the future if we decide to adopt these as Remix APIs. We're planning to promote your `remix-utils` package (alongside others) on a new docs page soon.

PR #1303, opened 2021-12-31
Remove duplicate useEffect in useScrollRestoration

zachdtaylor:
I noticed the same `useEffect` appeared twice in `useScrollRestoration`, so I removed the second one.

### DIFF
--- a/packages/remix-react/scroll-restoration.tsx
+++ b/packages/remix-react/scroll-restoration.tsx
@@ -116,10 +116,4 @@ function useScrollRestoration() {
       window.scrollTo(0, 0);
     }, [location]);
   }
-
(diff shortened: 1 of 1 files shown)


ryanflorence:
one is a layout effect, the other is a normal effect, note the difference in the boolean being set 😉 Whole thing will blow up w/o that effect!

zachdtaylor:
@ryanflorence I'm pretty sure they're both normal effects? 🤔 So just trying to understand, is the second copy of 
```javascript
React.useEffect(() => {
    if (transition.submission) {
      wasSubmissionRef.current = true;
    }
  }, [transition]);
```
necessary because it comes after the `useLayoutEffect`?

Issue #683, opened 2021-11-26
[Bug]: Yarn 3 support?


azat-io:
### Which Remix packages are impacted?

- `remix` (Remix core)
- `create-remix`

### What version of Remix are you using?

1.0.5

### Steps to Reproduce

Trying to install dependencies with Yarn berry. My current Yarn version is `v3.2.0-rc.5`.

Log file:

```
# This file contains the result of Yarn building a package (azat-io@workspace:.)
# Script name: postinstall

(node:67592) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
EROFS: read-only filesystem, open '/node_modules/remix/package.json'
```

Dependencies in my `package.json`:

```json
{
  "dependencies": {
    "@remix-run/node": "1.0.5",
    "@remix-run/react": "1.0.5",
    "@remix-run/serve": "1.0.5",
    "@remix-run/server-runtime": "1.0.5",
    "react": "17.0.2",
    "react-dom": "17.0.2",
    "remix": "1.0.5"
  },
  "devDependencies": {
    "@remix-run/dev": "1.0.5",
    "@types/react": "17.0.24",
    "@types/react-dom": "17.0.9",
    "typescript": "4.1.2"
  }
}
```

### Expected Behavior

Installation successful

### Actual Behavior

Installation failed

mxro:
I am able to install on yarn v `3.0.2` but when running `yarn build` using `remix@1.0.5` I get the following error:

```
Building Remix app in production mode...

 > error: [plugin: remix-mdx] Cannot find package 'xdm' imported from C:\Users\Max\repos\goldstack-mega\.yarn\cache\@remix-run-dev-npm-1.0.6-73f0b55172-2fca586246.zip\node_modules\@remix-run\dev\compiler\plugins\mdx.js
Did you mean to import xdm-npm-2.1.0-e825e62593-5a3f2434a5.zip/node_modules/xdm/index.js?


Build failed with 1 error:
error: Cannot find package 'xdm' imported from C:\Users\Max\repos\goldstack-mega\.yarn\cache\@remix-run-dev-npm-1.0.6-73f0b55172-2fca586246.zip\node_modules\@remix-run\dev\compiler\plugins\mdx.js
Did you mean to import xdm-npm-2.1.0-e825e62593-5a3f2434a5.zip/node_modules/xdm/index.js?
Built in 170ms
```

[...]

PR #1316, opened 2022-01-02
wip: Add Support for Yarn 3 PnP

cmd-johnson:
Yarn 3 has a feature called [Plug'n'Play (or PnP)](https://yarnpkg.com/features/pnp), which, among other things, gets rid of the `node_modules` directory.
Unfortunately, the way Remix modifies the `remix` package when running `remix setup node` is not compatible with PnP, because it tries to add/edit files in the `node_modules` directory.
PnP also enforces that all dependencies by a package are included in their respective `package.json` file, even when the package already indirectly depends on them. Additionally, `esbuild` doesn't work with PnP out of the box.

I have made a few changes that should enable Remix to work with projects that are using PnP, which I have outlined in #683.
Builds using yarn 3.2.0 (which is currently in the canary stage) now work without having to disable the PnP feature.

What _doesn't_ work though, is using the `remix` package. As far as I can tell, there's no easy way to change this.
For now, you can work around this by importing from the `@remix-run/node`, `@remix-run/server-runtime`,`@remix-run/react`, ... packages directly.

What _could_ work, is to include everything the `remix` package does into the project directory. This could also be automated by e.g. adding a flag to the `remix setup` command. That way, you would only have to change the `import {} from "remix";` imports to `import {} from "./remix";`. In TypeScript projects, the auto-generated files could also be added to the `paths` property in the `tsconfig.json`, which means you wouldn't have to change the imports at all.

I'm open for suggestions regarding the `remix` package issues with PnP and happy to implement them, if that will get PnP support added to Remix.

### DIFF
--- a/packages/create-remix/package.json
+++ b/packages/create-remix/package.json
@@ -26,6 +26,7 @@
   },
   "devDependencies": {
     "@types/chalk-animation": "^1.6.0",
-    "@types/inquirer": "^7.3.1"
+    "@types/inquirer": "^7.3.1",
+    "strip-ansi": "^6.0.0"
   }
 }
--- a/packages/create-remix/templates/yarn-pnp_js/app/root.jsx
+++ b/packages/create-remix/templates/yarn-pnp_js/app/root.jsx
@@ -1,0 +1,31 @@
+import {
+  Links,
+  LiveReload,
+  Meta,
+  Outlet,
+  Scripts,
+  ScrollRestoration
+} from "@remix-run/react";
+
+export function meta() {
+  return { title: "New Remix App" };
+}
+
+export default function App() {
+  return (
+    <html lang="en">
+      <head>
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
+      </body>
+    </html>
+  );
+}
--- a/packages/create-remix/templates/yarn-pnp_ts/app/root.tsx
+++ b/packages/create-remix/templates/yarn-pnp_ts/app/root.tsx
@@ -1,0 +1,32 @@
+import {
+  Links,
+  LiveReload,
+  Meta,
+  Outlet,
+  Scripts,
+  ScrollRestoration
+} from "@remix-run/react";
+import type { MetaFunction } from "@remix-run/server-runtime";
+
+export const meta: MetaFunction = () => {
+  return { title: "New Remix App" };
+};
+
+export default function App() {
+  return (
+    <html lang="en">
+      <head>
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
+      </body>
+    </html>
+  );
+}
--- a/packages/remix-architect/package.json
+++ b/packages/remix-architect/package.json
@@ -18,7 +18,9 @@
     "@remix-run/server-runtime": "1.2.2"
   },
   "peerDependencies": {
-    "@architect/architect": "^8.3.7"
+    "@architect/architect": "^8.3.7",
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
   },
   "devDependencies": {
     "@types/architect__functions": "^3.13.6",
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -29,6 +29,22 @@ import { writeFileSafe } from "./compiler/utils/fs";
 // as a source file when building the app.
 const reactShim = path.resolve(__dirname, "compiler/shims/react.ts");
 
+/**
+ * In case the remix project is using Yarn 3's PnP feature, we need to include
+ * the esbuild pnp plugin for the build to work. Otherwise, we just return a
+ * plugin that doesn't do anything.
+ */
+async function getYarnPnpPlugin(): Promise<esbuild.Plugin | null> {
+  try {
+    const esbuildPluginPnp = await import('@yarnpkg/esbuild-plugin-pnp');
+    if (esbuildPluginPnp) {
+      const { pnpPlugin } = esbuildPluginPnp;
+      return pnpPlugin();
+    }
+  } catch {}
+  return null;
+}
+
 interface BuildConfig {
   mode: BuildMode;
   target: BuildTarget;
@@ -341,6 +357,18 @@ async function createBrowserBuild(
       path.resolve(config.appDirectory, config.routes[id].file) + "?browser";
   }
 
+  const plugins: esbuild.Plugin[] = [
+    mdxPlugin(config),
+    browserRouteModulesPlugin(config, /\?browser$/),
+    emptyModulesPlugin(config, /\.server(\.[jt]sx?)?$/),
+    NodeModulesPolyfillPlugin()
+  ]
+
+  const yarnPnpPlugin = await getYarnPnpPlugin();
+  if (yarnPnpPlugin) {
+    plugins.push(yarnPnpPlugin);
+  }
+
   return esbuild.build({
     entryPoints,
     outdir: config.assetsBuildDirectory,
@@ -368,12 +396,7 @@ async function createBrowserBuild(
         config.devServerPort
       )
     },
-    plugins: [
-      mdxPlugin(config),
-      browserRouteModulesPlugin(config, /\?browser$/),
-      emptyModulesPlugin(config, /\.server(\.[jt]sx?)?$/),
-      NodeModulesPolyfillPlugin()
-    ]
+    plugins
   });
 }
 
@@ -408,6 +431,11 @@ async function createServerBuild(
 
   if (config.serverPlatform !== "node") {
     plugins.unshift(NodeModulesPolyfillPlugin());
+  }
+
+  const yarnPnpPlugin = await getYarnPnpPlugin();
+  if (yarnPnpPlugin) {
+    plugins.push(yarnPnpPlugin);
   }
 
   return esbuild
--- a/packages/remix-dev/compiler/assets.ts
+++ b/packages/remix-dev/compiler/assets.ts
@@ -73,7 +73,7 @@ export async function createAssetsManifest(
     if (!output.entryPoint) continue;
 
     let entryPointFile = path.resolve(
-      output.entryPoint.replace(/(^browser-route-module:|\?browser$)/g, "")
+      output.entryPoint.replace(/(^browser-route-module:|^pnp:|\?browser$)/g, "")
     );
     if (entryPointFile === entryClientFile) {
       entry = {
--- a/packages/remix-dev/package.json
+++ b/packages/remix-dev/package.json
@@ -37,6 +37,15 @@
     "@types/cacache": "^15.0.0",
     "@types/lodash.debounce": "^4.0.6",
     "@types/ws": "^7.4.1",
+    "@yarnpkg/esbuild-plugin-pnp": "^2.0.0",
     "semver": "^7.3.4"
+  },
+  "peerDependencies": {
+    "@yarnpkg/esbuild-plugin-pnp": "^2.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@yarnpkg/esbuild-plugin-pnp": {
+      "optional": true
+    }
   }
 }
--- a/packages/remix-dev/package.json
+++ b/packages/remix-dev/package.json
@@ -23,6 +23,7 @@
     "exit-hook": "2.2.1",
     "fs-extra": "^10.0.0",
     "get-port": "^5.1.1",
+    "jsesc": "^3.0.1",
     "lodash.debounce": "^4.0.8",
     "meow": "^7.1.1",
     "minimatch": "^3.0.4",
@@ -35,17 +36,23 @@
   },
   "devDependencies": {
     "@types/cacache": "^15.0.0",
+    "@types/jsesc": "^2.5.1",
     "@types/lodash.debounce": "^4.0.6",
     "@types/ws": "^7.4.1",
     "@yarnpkg/esbuild-plugin-pnp": "^2.0.0",
+    "rollup": "^2.36.1",
     "semver": "^7.3.4"
   },
   "peerDependencies": {
-    "@yarnpkg/esbuild-plugin-pnp": "^2.0.0"
+    "@remix-run/serve": "*",
+    "@yarnpkg/esbuild-plugin-pnp": "^2.0.0",
+    "express": "^4.17.1",
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
   },
   "peerDependenciesMeta": {
-    "@yarnpkg/esbuild-plugin-pnp": {
-      "optional": true
-    }
+    "@remix-run/serve": { "optional": true },
+    "@yarnpkg/esbuild-plugin-pnp": { "optional": true },
+    "express": { "optional": true }
   }
 }
--- a/contributors.yml
+++ b/contributors.yml
@@ -41,6 +41,7 @@
 - christianhg
 - christophgockel
 - clarkmitchell
+- cmd-johnson
 - codymjarrett
 - confix
 - coryhouse
--- a/packages/create-remix/cli.ts
+++ b/packages/create-remix/cli.ts
@@ -3,9 +3,8 @@ import chalkAnimation from "chalk-animation";
 import inquirer from "inquirer";
 import meow from "meow";
 
-import type { Lang, Server, Stack } from ".";
-import { appType } from ".";
-import { createApp } from ".";
+import type { Lang, PackageManager, Server, Stack } from ".";
+import { appType, createApp, getInstallCommand } from ".";
 
 const help = `
   Usage:
@@ -69,12 +68,14 @@ async function run() {
         stack?: never;
         server: Server;
         lang: Lang;
+        packageManager: PackageManager;
         install: boolean;
       }
     | {
         appType: "stack";
         stack: Stack;
         server?: never;
+        packageManager?: never;
         install: boolean;
       }
   >([
@@ -143,9 +144,30 @@ async function run() {
       ]
     },
     {
+      name: "packageManager",
+      type: "list",
+      message: "Which package manager?",
+      when(answers) {
+        return answers.appType !== appType.stack;
+      },
+      loop: false,
+      choices: [
+        { name: "npm", value: "npm" },
+        { name: "yarn (without PnP)", value: "yarn" },
+        { name: "yarn (with PnP enabled)", value: "yarn-pnp" }
+      ]
+    },
+    {
       name: "install",
       type: "confirm",
-      message: "Do you want me to run `npm install`?",
+      message: answers => {
+        const packageManager = typeof answers.packageManager === 'string'
+          ? answers.packageManager
+          : 'npm';
+
+        const command = getInstallCommand(packageManager);
+        return `Do you want me to run \`${command}\`?`;
+      },
       default: true
     }
   ]);
@@ -155,14 +177,15 @@ async function run() {
       projectDir,
       lang: "ts",
       stack: answers.stack,
-      install: answers.install
+      install: answers.install,
     });
   } else {
     await createApp({
       projectDir,
       lang: answers.lang,
       server: answers.server,
-      install: answers.install
+      install: answers.install,
+      packageManager: answers.packageManager,
     });
   }
 }
--- a/packages/create-remix/index.ts
+++ b/packages/create-remix/index.ts
@@ -27,6 +27,11 @@ export type AppType = typeof appType[keyof typeof appType];
 
 export type Lang = "ts" | "js";
 
+export type PackageManager =
+  | "npm"
+  | "yarn"
+  | "yarn-pnp";
+
 export type CreateAppArgs =
   | {
       projectDir: string;
@@ -35,6 +40,7 @@ export type CreateAppArgs =
       stack?: never;
       install: boolean;
       quiet?: boolean;
+      packageManager?: PackageManager;
     }
   | {
       projectDir: string;
@@ -43,13 +49,15 @@ export type CreateAppArgs =
       stack: Stack;
       install: boolean;
       quiet?: boolean;
+      packageManager?: PackageManager;
     };
 
-async function createApp({
+export async function createApp({
   projectDir,
   lang,
   install,
   quiet,
+  packageManager = 'npm',
   ...rest
 }: CreateAppArgs) {
   let server = rest.stack ? rest.stack : rest.server;
@@ -95,6 +103,16 @@ async function createApp({
     await fse.copy(serverLangTemplate, projectDir, { overwrite: true });
   }
 
+  // copy the package manager template
+  const packageManagerLangTemplate = path.resolve(
+    __dirname,
+    "templates",
+    `${packageManager}_${lang}`
+  );
+  if (fse.existsSync(packageManagerLangTemplate)) {
+    await fse.copy(packageManagerLangTemplate, projectDir, { overwrite: true });
+  }
+
   // rename dotfiles
   let dotfiles = ["gitignore", "github", "dockerignore", "env.example"];
   await Promise.all(
@@ -114,8 +132,13 @@ async function createApp({
   appPkg.dependencies = appPkg.dependencies || {};
   appPkg.devDependencies = appPkg.devDependencies || {};
   let serverPkg = require(path.join(serverTemplate, "package.json"));
+  let packageManagerPkgPath = path.join(packageManagerLangTemplate, "package.json");
+  let packageManagerPkg = fse.existsSync(packageManagerPkgPath)
+    ? require(packageManagerPkgPath)
+    : {};
+
   ["dependencies", "devDependencies", "scripts"].forEach(key => {
-    Object.assign(appPkg[key], serverPkg[key]);
+    Object.assign(appPkg[key], serverPkg[key], packageManagerPkg[key]);
   });
 
   appPkg.main = serverPkg.main;
@@ -133,7 +156,15 @@ async function createApp({
     }
   });
 
+
   appPkg = sortPackageJSON(appPkg);
+
+  if (packageManager === "yarn-pnp") {
+    // yarn PnP does not support the remix package's magic, so we remove it from
+    // the dependencies and delete the postinstall hook
+    delete appPkg.dependencies.remix;
+    delete appPkg.scripts.postinstall;
+  }
 
   // write package.json
   await fse.writeFile(
@@ -141,8 +172,14 @@ async function createApp({
     JSON.stringify(appPkg, null, 2)
   );
 
+  // enable yarn PnP if requested
+  if (packageManager === "yarn-pnp") {
+    execSync("yarn set version berry", { stdio: "inherit", cwd: projectDir });
+  }
+
   if (install) {
-    execSync("npm install", { stdio: "inherit", cwd: projectDir });
+    const command = getInstallCommand(packageManager);
+    execSync(command, { stdio: "inherit", cwd: projectDir });
   }
 
   let serverScript = path.resolve(serverTemplate, "scripts/init.js");
@@ -174,4 +211,12 @@ async function createApp({
   }
 }
 
-export { createApp };
+export function getInstallCommand(packageManager: PackageManager): string {
+  switch (packageManager) {
+    case "npm":
+      return "npm install";
+    case "yarn":
+    case "yarn-pnp":
+      return "yarn install";
+  }
+}
--- a/packages/create-remix/templates/fly/package.json
+++ b/packages/create-remix/templates/fly/package.json
@@ -9,5 +9,8 @@
   "dependencies": {
     "@remix-run/serve": "*",
     "cross-env": "^7.0.3"
+  },
+  "devDependencies": {
+    "express": "^4.17.1"
   }
 }
--- a/packages/create-remix/templates/remix/package.json
+++ b/packages/create-remix/templates/remix/package.json
@@ -8,5 +8,8 @@
   "dependencies": {
     "@remix-run/serve": "*",
     "cross-env": "^7.0.3"
+  },
+  "devDependencies": {
+    "express": "^4.17.1"
   }
 }
--- a/packages/create-remix/templates/vercel/package.json
+++ b/packages/create-remix/templates/vercel/package.json
@@ -9,6 +9,7 @@
   },
   "devDependencies": {
     "@remix-run/serve": "*",
-    "cross-env": "^7.0.3"
+    "cross-env": "^7.0.3",
+    "express": "^4.17.1"
   }
 }
--- a/packages/create-remix/templates/yarn-pnp/gitignore
+++ b/packages/create-remix/templates/yarn-pnp/gitignore
@@ -1,0 +1,14 @@
+node_modules
+
+/.cache
+/server/build
+/public/build
+.env
+
+.yarn/*
+!.yarn/cache
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
--- a/packages/create-remix/templates/yarn-pnp_js/app/entry.client.jsx
+++ b/packages/create-remix/templates/yarn-pnp_js/app/entry.client.jsx
@@ -1,0 +1,4 @@
+import { hydrate } from "react-dom";
+import { RemixBrowser } from "@remix-run/react";
+
+hydrate(<RemixBrowser />, document);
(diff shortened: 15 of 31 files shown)


cmd-johnson:
I chose to not further go down the path of getting the `remix` package to work with PnP.
PnP doesn't support the `paths` property in the `tsconfig.json`, and will report it as an undeclared dependency. (This also applies to imports from `~/`.)

While copying all those files into a sub-directory could still work, I think it's easier to just accept that you'll need to import most things directly from `@remix-run/react` instead of the `remix` package.

I'm working on extending the create-remix package with an option for selecting the package manager. The resulting projects will then include all required dependencies to make Remix work with PnP, as well as changing all the imports from the `remix` package to their respective `@remix-run/*` packages.

cmd-johnson:
Quick update:
These changes don't work with the current version of yarn (v3.1.1), because of https://github.com/yarnpkg/berry/issues/3687.
The issue was already solved by yarn, but is not released yet. It should find its way into the v3.2.0 release.

I'll continue working on this PR once the next yarn version is released.

badeAdebayo:
is there a chance these changes can get merged anyway. for people who are using yarn 3.2.0?

cmd-johnson:
Yarn 3.2.0 has just been released! :tada:
I'll try and get around to updating this PR tomorrow.

cmd-johnson:
From my tests with Yarn (with and without PnP enabled), almost everything appears to be working fine.
What doesn't work are the `dev` scripts created by `create-remix`. Those use `npm-run-all` to run commands in parallel, which fails on my machine. I'm guessing that's because of `npm-run-all` not being compatible with PnP.

chaance:
I'm going to close this for the time being. We may revisit this down the road but we have some other priorities we need to focus on before we overhaul our repo tooling, as Yarn 3 is a pretty big shift from how we're doing things at the moment.

cmd-johnson:
Thank you for your response!
I'm not 100% sure what you mean, because I'm not changing anything about Remix's repo tooling in this PR. These are just changes that allow users of Remix to utilize yarn PnP and doesn't influence how Remix itself is built.

It's mainly about declaring all direct and indirect (peer-)dependencies in all packages in order for yarn to stop complaining.

The only actual changes to Remix features are
- the addition of an (optional) peer dependency on `@yarnpkg/esbuild-plugin-pnp` that non-PnP users don't need to bother with. As long as the `esbuild` parameters cannot be customized by users, there unfortunately is no way around this if you want to allow PnP to work with Remix.
- adding an option to use yarn to `create-remix`

If you think that adding a yarn option to `create-remix` is too much, I'd be happy to just add a yarn example to the `example` directory instead.

badeAdebayo:
@cmd-johnson now that this is reopen do you plan to fix the conflicts and get this merged?

cmd-johnson:
They seem to change quite a few things around in the past couple of weeks. I just did that 4 days ago! :smile: 
I might get to it tomorrow.

Also I think I'd rather go the route of just adding a yarn example over adding an option for it to `create-remix`. That should make this PR easier to approve and merge.
…and also reduce the number of future merge conflicts. They appear to do a lot of work on `create-remix` atm.

chaance:
@cmd-johnson we made a pretty big impactful change by updating our prettier config. A simple way to reduce a lot of the merge conflicts might be to copy/paste the new config and run `yarn format` from the project root. https://github.com/remix-run/remix/pull/2085

packages/create-remix/templates/yarn-pnp_js/app/root.jsx
```diff
@@ -0,0 +1,31 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 27:
```suggestion
        <LiveReload />
```
See #1352

packages/create-remix/templates/yarn-pnp_ts/app/root.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 28:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

cmd-johnson:
There we go!
I've added an example that's using Yarn 3.2.0 and PnP and removed the "Yarn" and "Yarn PnP" options from the `create-remix` command again.
That should reduce the overall complexity of this PR.

badeAdebayo:
will this be merged soon?

machour:
@cmd-johnson would you mind fixing conflicts and applying @MichaelDeBoey suggestions? 🙏🏼 
I'll get some attention to this PR.

cmd-johnson:
I fixed the merge conflicts again.

I think the changes requested by @MichaelDeBoey are covered since I've removed the yarn templates and just added an example instead, which doesn't use the `process.env.NODE_ENV === "development"` check anymore.

What I'm kind of missing is a build step that automatically checks for undeclared dependencies without having to switch to Yarn 3 for the whole Remix project.
I'm a bit worried about undeclared dependencies creeping in during development, which would break Yarn support again.

`yarn dlx @yarnpkg/doctor ./packages/` is the closest thing I found (and is what I've used to make this PR), but it reports all kinds of false positives (e.g. direct references to the `node_modules` directory instead of using `require.resolve`, which is fine as long as you're using Yarn 1.* for Remix itself, as well as occurrences of `import type` that are covered by dependencies on `@types/*`).

mjackson:
Thank you for your work here, @cmd-johnson. We really appreciate it.

> What doesn't work though, is using the remix package. As far as I can tell, there's no easy way to change this.
For now, you can work around this by importing from the @remix-run/node, @remix-run/server-runtime,@remix-run/react, ... packages directly.

This is the approach that we're taking in Remix. Some time in the next few weeks we will be deprecating importing directly from the `remix` package and recommend importing from the individual server runtime packages instead. The approach will be to get all server stuff from one of three server runtimes: `@remix-run/node`, `@remix-run/cloudflare`, or `@remix-run/deno`. All hooks/components will come from `@remix-run/react`.

The short story is that we won't have any problems with yarn pnp or pnpm because all our packages will just be normal npm packages. We won't have any magic `remix` package anymore.

A bit of backstory: when we first started building Remix everything lived in a single package. Over time, we've been figuring out how to tease that out into separate packages, but we weren't sure exactly which ones we needed. Using the "magic" `remix` package was a nice stopgap, but kinda janky because it requires using `remix setup`. So we're also deprecating `remix setup` soon.

We have already begun this work in #2359, and will be merging more changes to our existing docs and templates over the coming weeks. I just wanted to give y'all the heads up here so you know the future direction we are headed in.

mjackson:
I should also clarify: the `remix` package (and `remix setup`) should continue to work through Remix 1.x.

That said, I think I'd prefer to just tell people to `import` directly from the `@remix-run/*` packages and avoid the problem altogether instead of merging changes to our compiler to support Yarn pnp.

cmd-johnson:
Thanks for the update, @mjackson!
I really like the direction where Remix is headed.
One thing though: PnP changes how modules are resolved and without the plugin, esbuild will not be able to find anything you try to import.

As far as I can tell, it's down to either one of
- adding esbuild-plugin-pnp to Remix's esbuild config (what this PR does)
- allow users to customize the esbuild config (which as I understand it isn't something you want to do as of now)
- not supporting PnP, period (which tbh is also fine. Things work perfectly well without it, even if it would be nice to have.)

SagnikPradhan:
Any updates on where this is headed right now?

MichaelDeBoey:
@cmd-johnson Since `v1.3.5` is released (which includes #2359, as mentioned above by @mjackson), does this still need to be fixed?

If you use the latest templates (which are using the `@remix-run/*` imports instead of the magicExports the `remix` package is having), I think all would be fine when using Yarn PnP?

SagnikPradhan:
@MichaelDeBoey If I am not wrong the [esbuild pnp plugin needs to be added conditionally too](https://github.com/remix-run/remix/blob/81830a9a35af7254aaa455334dadda311b0ea5cb/packages/remix-dev/compiler.ts#L373).

lensbart:
@MichaelDeBoey in my experience, `@remix-run/dev` imports from a `.bin` directory which doesn’t exist when using PnP (and is otherwise supposed to be inside `node_modules`). This still causes an error on install and when running the `dev` script:

> Error: @remix-run/dev tried to access .bin, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

cmd-johnson:
@MichaelDeBoey: @SagnikPradhan is correct - without `eslint-plugin-pnp`, `eslint` won't be able to find any installed dependencies, because PnP doesn't put them in the `node_modules` directory.

I also encountered the error reported by @lensbart, but this one should be an easy fix (instead of [this](https://github.com/remix-run/remix/blob/0d681e3bab8cfdc76e3296589534dcda7b66362a/packages/remix-dev/cli/migrate/jscodeshift.ts#L3), use `require.resolve("jscodeshift/bin/jscodeshift")`)

pcattori:
@cmd-johnson the `jscodeshift`/`.bin` issues should have been resolved by #3114 , which has been released as of v1.5.0.

Do we need anything else for PnP support? Do we still need to conditionally add the esbuild plugin for PnP?

cmd-johnson:
@pcattori Yep, `eslint-plugin-pnp` unfortunately is a must-have if you want to build anything that uses Yarn's PnP feature with esbuild.

SagnikPradhan:
@MichaelDeBoey @cmd-johnson Is there something that this PR needs to move forward?

packages/create-remix/package.json
```diff
@@ -16,6 +16,8 @@
     "create-remix": "cli.js"
   },
   "dependencies": {
-    "@remix-run/dev": "1.3.2"
+    "@remix-run/dev": "1.3.2",
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
```
pcattori, line 21: Why are `react` and `react-dom` included as deps of `create-remix`?

packages/remix-architect/package.json
```diff
@@ -17,6 +17,10 @@
     "@remix-run/node": "1.3.2",
     "@remix-run/server-runtime": "1.3.2"
   },
+  "peerDependencies": {
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
+  },
```
pcattori, line 23: None of our adapters (`remix-architect`, `remix-cloudflare-pages`, `remix-cloudflare-workers`, `remix-express`, `remix-netlify`) depend on `react` nor `react-dom`. Is it intentional do add them as `peerDependencies` to all of these packages?

cmd-johnson: Yarn complained about those, because react and react-dom are indirect peer dependencies of all of these packages, which means that those dependencies should be declared as peer dependencies of your package as well.
E.g. for `remix-architect` via `remix-server-runtime` → `remix-router`, which has a peer dependency on `react >= 16.8`.

cmd-johnson:
I'll try and find some time to rebase this PR onto the latest `dev` branch and fix the conflicts today or tomorrow.

packages/remix-architect/package.json
MichaelDeBoey: Currently these are indeed in `peerDependencies` of `react-router`, which is in `dependencies` of `@remix-run/server-runtime`.

However, we're not using any `react`-related things from `react-router`, so having it in `peerDependencies` isn't necessary.

In the (near) future, we'll have `@remix-run/router`, which will be a framework agnostic version of `react-router`.
Whenever we have this, we'll update `@remix-run/server-runtime`'s `dependencies` to include the new package.

We can remove them from all other packages for now.

cmd-johnson:
I've rebased the PR onto the current `dev` branch again and
- removed the peer dependencies that were not actually required after building `remix`.
- added a missing optional peer dependency on `@remix-run/serve` to the `@remix-run/dev` package as it's only used when running `remix dev`
- added an optional peer dependency on `@yarnpkg/esbuild-plugin-pnp` to `@remix-run/dev` which you only have to install when you're actually using yarn PnP
- rewrote three imports from `react-router` to use `react-router-dom` instead to circumvent having to add `react-router` as peer dependencies to `@remix-run/react` and `@remix-run/server-runtime` (those three were already marked by a TODO comment, so I just went ahead and did that)
- updated the yarn-pnp example to match the other examples again.

Looking forward to finally being able to use remix with Yarn PnP!

packages/remix-dev/compiler.ts
```diff
@@ -425,6 +441,11 @@ function createServerBuild(
     plugins.unshift(NodeModulesPolyfillPlugin());
   }
 
+  let yarnPnpPlugin = await getYarnPnpPlugin();
+  if (yarnPnpPlugin) {
+    plugins.push(yarnPnpPlugin);
+  }
```
pcattori, line 447: I'm trying to determine why we are adding the Yarn PnP plugin conditionally. Is it because the plugin wouldn't work with users using `npm` or `yarn`? Or is it for having `@yarnpkg/esbuild-plugin-pnp` in `peerDependencies` and cut down the number of dependencies for non-Yarn-PnP users?

SagnikPradhan: Yep, it's an optional peer dependency, so for most users there is no point in having it. Yarn PNP users would have to explicitly install it themselves.

packages/remix-react/routeMatching.ts
```diff
@@ -1,7 +1,7 @@
 // TODO: We eventually might not want to import anything directly from `history`
 // and leverage `react-router` here instead
 import type { Location } from "history";
-import type { Params, RouteObject } from "react-router"; // TODO: export/import from react-router-dom
+import type { Params, RouteObject } from "react-router-dom";
```
MichaelDeBoey, line 4: These changes aren't necessary in this PR and could be done in a separate PR, so let's revert them to keep this PR focussed

```suggestion
import type { Params, RouteObject } from "react-router"; // TODO: export/import from react-router-dom
```

packages/remix-react/routeModules.ts
```diff
@@ -2,7 +2,7 @@
 // and leverage `react-router` here instead
 import type { Location } from "history";
 import type { ComponentType } from "react";
-import type { Params } from "react-router"; // TODO: import/export from react-router-dom
+import type { Params } from "react-router-dom";
```
MichaelDeBoey, line 5:
```suggestion
import type { Params } from "react-router"; // TODO: import/export from react-router-dom
```

packages/remix-server-runtime/routeMatching.ts
```diff
@@ -1,4 +1,4 @@
-import type { Params, RouteObject } from "react-router"; // TODO: export/import from react-router-dom
+import type { Params, RouteObject } from "react-router-dom";
```
MichaelDeBoey, line 1:
```suggestion
import type { Params, RouteObject } from "react-router"; // TODO: export/import from react-router-dom
```

MichaelDeBoey (changes requested):

packages/remix-dev/compiler.ts
```diff
@@ -299,7 +310,7 @@ async function buildEverything(
     // it when loading the @remix-run/dev/assets-manifest virtual module.
     assetsManifestPromiseRef.current = assetsManifestPromise;
 
-    let serverBuildPromise = createServerBuild(
+    let serverBuildPromise = await createServerBuild(
```
MichaelDeBoey, line 313: We shouldn't await `createServerBuild` here, as it's already in the `Promise.all` we're `await`ing.

```suggestion
    let serverBuildPromise = createServerBuild(
```

MichaelDeBoey (changes requested):

packages/remix-dev/compiler.ts
cmd-johnson: Fixed :+1:

examples/yarn-pnp/.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
```
MichaelDeBoey, line 1: Let's keep the example as small as possible & remove the `.vscode` folder

packages/remix-dev/package.json
```diff
@@ -68,6 +68,19 @@
     "msw": "^0.39.2",
     "shelljs": "^0.8.5",
     "tiny-invariant": "^1.2.0",
-    "type-fest": "^2.12.2"
+    "type-fest": "^2.12.2",
+    "@yarnpkg/esbuild-plugin-pnp": "^2.0.0"
+  },
+  "peerDependencies": {
+    "@remix-run/serve": "1.5.1",
```
MichaelDeBoey, line 75: Since we now have a remix package in `peerDependencies`, we'll need to update `updateRemixVersion` script util as well to account for it.

https://github.com/remix-run/remix/blob/5daa20d7992a061dedb6e8a1883fd259ae424ca9/scripts/utils.js#L138-L145

MichaelDeBoey (changes requested):

examples/yarn-pnp/.vscode/extensions.json
cmd-johnson: will do

examples/yarn-pnp/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "@types/react-dom": "^17.0.9",
+    "@yarnpkg/esbuild-plugin-pnp": "^2.0.0",
+    "esbuild": ">=0.10.0",
+    "express": "^4.17.1",
+    "typescript": "^4.1.2"
+  },
+  "engines": {
+    "node": ">=14"
+  },
+  "packageManager": "yarn@3.2.0"
```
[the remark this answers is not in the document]
cmd-johnson: This is a field used by [Corepack](https://nodejs.org/api/corepack.html) to determine the used package manager, and gets set automatically if you initialize your project with Yarn ≥ 2.
When using Yarn ≥ 2, it copies itself into the `.yarn/releases` directory. That way, everyone who clones the repository uses the exact same version of yarn and as such, you'll want to pin it to an exact version here. VSCode actually complains if you write `yarn@^3.2.0`: "String does not match the pattern of `"(npm|pnpm|yarn)@\d+\.\d+\.\d+(-.+)?"`"

Although I just noticed that I actually forgot to add the (non-gitignored parts of the) `.yarn` directory to the example :sweat_smile:

packages/remix-dev/compiler.ts
```diff
@@ -355,6 +366,11 @@ async function createBrowserBuild(
     NodeModulesPolyfillPlugin(),
   ];
 
+  let yarnPnpPlugin = await getYarnPnpPlugin();
+  if (yarnPnpPlugin) {
+    plugins.push(yarnPnpPlugin);
+  }
```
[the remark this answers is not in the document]
cmd-johnson: I like this one, although TypeScript then complains when passing this array to `esbuild.build`, because it now is of type `(esbuild.Plugin | null)[]`.
Writing `let plugins = [/* ... */].filter((p): p is esbuild.Plugin => !!p);` works, although it's a bit more verbose.

packages/remix-dev/compiler.ts
```diff
@@ -67,6 +67,17 @@ export function formatBuildFailure(failure: BuildError) {
   console.error(failure?.message || "An unknown build error occurred");
 }
 
+async function getYarnPnpPlugin(): Promise<esbuild.Plugin | null> {
+  try {
+    let esbuildPluginPnp = await import("@yarnpkg/esbuild-plugin-pnp");
+    if (esbuildPluginPnp) {
+      let { pnpPlugin } = esbuildPluginPnp;
+      return pnpPlugin();
+    }
```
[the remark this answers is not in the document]
cmd-johnson: Huh, true. I'm not sure why I added that check.

packages/remix-dev/compiler.ts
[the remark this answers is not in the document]
lensbart: maybe if you add `?? undefined` after the nullable item it’ll satisfy the type checker

examples/yarn-pnp/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "@types/react-dom": "^17.0.9",
+    "@yarnpkg/esbuild-plugin-pnp": "^2.0.0",
+    "esbuild": ">=0.10.0",
+    "express": "^4.17.1",
+    "typescript": "^4.1.2"
+  },
+  "engines": {
+    "node": ">=14"
+  },
+  "packageManager": "yarn@3.2.1"
```
MichaelDeBoey, line 32: Let's remove the `packageManager` field & add `yarn` to the `engines` field instead, as Corepack is still experimental and we don't have a way to specify a range

```suggestion
  "engines": {
    "node": ">=14"
    "yarn": ">=3.0.0"
  }
```

MichaelDeBoey (changes requested):

packages/remix-dev/compiler.ts
```diff
@@ -353,7 +361,8 @@ async function createBrowserBuild(
     browserRouteModulesPlugin(config, /\?browser$/),
     emptyModulesPlugin(config, /\.server(\.[jt]sx?)?$/),
     NodeModulesPolyfillPlugin(),
-  ];
+    await getYarnPnpPlugin(),
+  ].filter((p): p is esbuild.Plugin => !!p);
```
MichaelDeBoey, line 365: Let's make it more obvious that we're returning a `boolean` here

```suggestion
  ].filter((p): p is esbuild.Plugin => Boolean(p));
```

packages/remix-dev/compiler.ts
```diff
@@ -419,7 +428,8 @@ function createServerBuild(
     serverEntryModulePlugin(config),
     serverAssetsManifestPlugin(assetsManifestPromiseRef),
     serverBareModulesPlugin(config, dependencies, options.onWarning),
-  ];
+    await getYarnPnpPlugin(),
+  ].filter((p): p is esbuild.Plugin => !!p);
```
MichaelDeBoey, line 432:
```suggestion
  ].filter((p): p is esbuild.Plugin => Boolean(p));
```

MichaelDeBoey (changes requested):

examples/yarn-pnp/package.json
cmd-johnson: I'm not sure. The recommended way to install yarn [_is_ using corepack](https://yarnpkg.com/getting-started/install), so I think it'd be better to keep the `"packageManager"` field as is.
At least on my Linux distribution, you only get yarn 1.x from the official repositories. To use yarn ≥ 2 (and PnP), you only really have two options:
- use corepack and the `"packageManager"` field
- run `yarn set version berry`, which copies the latest stable version of yarn into the project's `.yarn/releases` directory and put a reference to it in `.yarnrc.yml`. This local copy _must not be gitignored_ or you won't be able to run yarn in the project at all.

Doing the latter would work, but would add a `.cjs` file with ~2.1MiB, which I think is a bit much for such an (otherwise small) example project.

MichaelDeBoey (approved):

MichaelDeBoey:
@cmd-johnson Could you please rebase your branch onto latest `dev` & resolve conflicts?

cmd-johnson:
@MichaelDeBoey Done!

packages/remix-dev/package.json
```diff
@@ -68,9 +68,22 @@
...
+    "@yarnpkg/esbuild-plugin-pnp": "^2.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@remix-run/serve": {
+      "optional": true
+    },
+    "@yarnpkg/esbuild-plugin-pnp": {
+      "optional": true
+    }
```
pcattori, line 87: I'm not convinced yet that having `@yarnpkg/esbuild-plugin-pnp` as optional dependency is the way we want to go. I'm leaning towards "yes", but will discuss with the team. If we do go with this optional dependency, then we'd need some docs that explain to users that they need to install it if they are using yarn pnp.

SagnikPradhan:
We are done with this, no?

pcattori:
After talking with the team, here is what we envision:

- We'd like to support the major package managers, including Yarn PnP
- We don't mind adding dependencies to the compiler. We're internally discussing changes to how we want to bundle the compiler and/or CLI that should address any concerns about dependency bloat. So don't worry about it for this PR.
- We'd prefer that support for Yarn PnP work out-of-the-box without requiring users to install a dev dependency

I _think_ that means we can just add `@yarnpkg/esbuild-plugin-pnp` as a dependency of `@remix-run/dev` (not an optional peer dependency) and unconditionally add it to our list of esbuild plugins. Since the pnp plugin is added at the end of the list, it should not be a breaking change for apps/projects that want to use normal Node-resolution for their packages.

I also had a couple questions about the code, but I'll leave those as comments on the code rather than within this higher level feedback.

packages/remix-dev/compiler/assets.ts
```diff
@@ -73,7 +73,10 @@ export async function createAssetsManifest(
     if (!output.entryPoint) continue;
 
     let entryPointFile = path.resolve(
-      output.entryPoint.replace(/(^browser-route-module:|\?browser$)/g, "")
+      output.entryPoint.replace(
+        /(^browser-route-module:|^pnp:|\?browser$)/g,
```
pcattori, line 77: I didn't expect to need to handle the `pnp` namespace here. Can you explain why we need to consider it here?

examples/yarn-pnp/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "dev": "remix dev",
+    "start": "remix-serve build"
+  },
+  "dependencies": {
+    "@remix-run/node": "1.5.1",
+    "@remix-run/react": "1.5.1",
+    "@remix-run/serve": "1.5.1",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "~": "link:./app"
```
pcattori, line 18: What does this line do? I was expecting to see an `import {...} from "~/...";` somewhere in the example codebase that exercises this, but maybe I'm misunderstanding the purpose of `~` and `link:`

cmd-johnson: PnP doesn't know or care about `tsconfig.json` files. In order to use the `~` alias that the other examples all provide through their `tsconfig.json`, we can use the `link:` protocol that yarn supports to remedy this issue. (See also https://yarnpkg.com/features/protocols#why-is-the-link-protocol-recommended-over-aliases-for-path-mapping)

packages/remix-dev/compiler/assets.ts
cmd-johnson: Short version: without adding this, the PnP build fails with "Missing output for entry point"

I'm assuming that the plugin takes over all module resolutions, no matter if they use an absolute or relative path or if you try to import installed dependencies. It resolves everything into the `pnp` namespace, including the `entry.client.tsx` file.

cmd-johnson:
@pcattori 

 > Is there a good way to automatically detect if Yarn PnP is being used?

The way esbuild-plugin-pnp does it is by [checking if `require("module")` has a `findPnpApi` property](https://github.com/yarnpkg/berry/blob/master/packages/esbuild-plugin-pnp/sources/index.ts#L111-L113).
After a quick test, that property is only there if the PnP feature is used. As soon as I disable PnP and re-install all dependencies, the property is gone.

> Since the pnp plugin is added at the end of the list, it should not be a breaking change for apps/projects that want to use normal Node-resolution for their packages.

As far as `esbuild-plugin-pnp` is concerned, the plugin basically becomes a no-op when PnP isn't used because of the check I linked above.

MichaelDeBoey:
> the plugin basically becomes a no-op when PnP isn't used

Good news, as this means we can put `@yarnpkg/esbuild-plugin-pnp` in `dependencies` instead of having it in `peerDependencies`.

packages/remix-dev/compiler.ts
```diff
@@ -353,7 +354,8 @@ async function createBrowserBuild(
     browserRouteModulesPlugin(config, /\?browser$/),
     emptyModulesPlugin(config, /\.server(\.[jt]sx?)?$/),
     NodeModulesPolyfillPlugin(),
-  ];
+    yarnPnpPlugin(),
+  ].filter((p): p is esbuild.Plugin => Boolean(p));
```
MichaelDeBoey, line 358: No need to filter this array anymore now that we're always including `@yarnpkg/esbuild-plugin-pnp`

```suggestion
  ];
```

packages/remix-dev/compiler.ts
```diff
@@ -419,7 +421,8 @@ function createServerBuild(
     serverEntryModulePlugin(config),
     serverAssetsManifestPlugin(assetsManifestPromiseRef),
     serverBareModulesPlugin(config, dependencies, options.onWarning),
-  ];
+    yarnPnpPlugin(),
+  ].filter((p): p is esbuild.Plugin => Boolean(p));
```
MichaelDeBoey, line 425:
```suggestion
  ];
```

packages/remix-dev/compiler.ts
```diff
@@ -386,7 +388,7 @@ async function createBrowserBuild(
   });
 }
 
-function createServerBuild(
+async function createServerBuild(
```
MichaelDeBoey, line 391: Also no need to make this an `async` function now.

MichaelDeBoey (changes requested):

packages/remix-dev/compiler.ts
```diff
@@ -353,7 +354,8 @@ async function createBrowserBuild(
     browserRouteModulesPlugin(config, /\?browser$/),
     emptyModulesPlugin(config, /\.server(\.[jt]sx?)?$/),
     NodeModulesPolyfillPlugin(),
-  ];
+    yarnPnpPlugin(),
+  ]
```
MichaelDeBoey, line 358:
```suggestion
  ];
```

packages/remix-dev/compiler.ts
```diff
@@ -419,7 +421,8 @@ function createServerBuild(
     serverEntryModulePlugin(config),
     serverAssetsManifestPlugin(assetsManifestPromiseRef),
     serverBareModulesPlugin(config, dependencies, options.onWarning),
-  ];
+    yarnPnpPlugin(),
+  ]
```
MichaelDeBoey, line 425:
```suggestion
  ];
```

MichaelDeBoey (changes requested):

examples/yarn-pnp/app/root.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <html lang="en">
+      <head>
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        <LiveReload />
```
MichaelDeBoey, line 28: See @mcansh's #3411

```suggestion
        <LiveReload port={Number(process.env.REMIX_DEV_SERVER_WS_PORT)} />
```

MichaelDeBoey (changes requested):

examples/yarn-pnp/package.json
pcattori: Ok that makes sense, but I still can't find anywhere in the `examples/yarn-pnp` code where the `~` is actually used.

If its not used at all, could we remove this alias from the example to keep things minimal?

pcattori:
@cmd-johnson : changes to the Remix compiler look good! If you can address @MichaelDeBoey 's comments for the example, then I think we should be good to merge this after.

examples/yarn-pnp/package.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "devDependencies": {
+    "@remix-run/dev": "1.5.1",
+    "@types/react": "^17.0.24",
+    "@types/react-dom": "^17.0.9",
+    "esbuild": ">=0.10.0",
+    "express": "^4.17.1",
```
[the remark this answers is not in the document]
cmd-johnson: Oh, those are leftovers from when we used peer dependencies for the pnp plugin. I'll remove those as everything appears to be working fine without them now.

cmd-johnson:
@pcattori I've included the change requests and rebased on the `dev` branch again.

Re: integration tests: any pointers how I should go about implementing them?
Especially given that I would have to install dependencies of the test project using a different version of yarn than the tests themselves would be running with

examples/yarn-pnp/app/root.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <html lang="en">
+      <head>
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        <LiveReload port={Number(process.env.REMIX_DEV_SERVER_WS_PORT)} />
```
MichaelDeBoey, line 28: Let's revert this for now, as #3411 still has a discussion going on & we're not sure yet about the solution for the problem.

```suggestion
        <LiveReload />
```

examples/yarn-pnp/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "remix-example-yarn-pnp",
```
MichaelDeBoey, line 2: This field can also be removed

```suggestion
```

MichaelDeBoey (changes requested):

pcattori:
@cmd-johnson : good point, figuring out how to test this is tricky. I'll think more about this today and get back to you.

pcattori:
Going to merge this as-is without a test since its relatively simple / low-risk. If for whatever reason other changes break Yarn PnP support more than once or twice, we can revisit how to add a test.

Thanks for everyone who participated in this PR and special thanks to @cmd-johnson for the work (and patience!).

pcattori (approved):

PR #1331, opened 2022-01-03
chore(remix-react): make inline scripts more robust

valerie-makes:
For inline scripts, instead of directly injecting JavaScript using a template literal:

- Write a function with the desired code, using params for external values
- Convert this function to a string using [`Function.prototype.toString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString)
- Inject the resulting string, and call the function with the required values

This provides the following benefits for inline scripts:

- Problem checking with ESLint
- Type checking with TypeScript
- Consistent formatting with Prettier
- Consistent baseline indentation

I haven't added or changed any tests since this isn't strictly a bug fix or new feature.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -167,6 +167,7 @@
 - tylerbrostrom
 - unhackit
 - UsamaHameed
+- valerie-makes
 - veritem
 - VictorPeralta
 - weavdale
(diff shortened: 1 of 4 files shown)


valerie-makes:
@ryanflorence I've rebased with the new script changes 🙂

ryanflorence:
Heh, our tests here are a little funny, I'll fix them.

PR #1332, opened 2022-01-03
docs(jokes): add key type in types import

ozeas:


ryanflorence:
fixed elsewhere, thanks!

PR #1336, opened 2022-01-04
fix(create-remix): check if .gitignore exisits and prompt user

AbhinavRobinson:
I had noticed that in a preinitialised git repository, the create-remix app fails to run. After going through the `create-remix/cli.ts` I had not seen a check for if `.gitignore` exists. For this I have added a check as well as a user prompt to ask to replace the existing `.gitignore` with the one that remix ships with.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,4 +1,5 @@
 - abereghici
+- abhinavrobinson
 - ahbruns
 - ahmedeldessouki
 - Alarid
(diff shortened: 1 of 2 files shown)


ryanflorence:
I think I'd rather see it check if any file it's about to write exists already and prompt to replace it or not. Feel free to tackle that if you'd like :)

PR #1340, opened 2022-01-04
fix(docs): minor fix in pages/contributing#Setup

hollandThomas:


ryanflorence:
prettier will just put that right back, the new paragraph/code blocks under a bullet need to be indented five spaces I think

PR #1343, opened 2022-01-04
Propose alternative way to load dotenv

lkostrowski:
Hey

So I figured that conditional loading dotenv in development in entry might be better than preloading in package.json script, what do you think?

### DIFF
--- a/docs/guides/envvars.md
+++ b/docs/guides/envvars.md
@@ -43,13 +43,12 @@ Edit your `.env` file.
 SOME_SECRET=super-secret
 ```
 
-Then update your package.json dev script to this:
+Then update your entry.server.tsx and add these lines before your imports
 
-```json lines=[2] filename=package.json
-{
-  "dev": "node -r dotenv/config node_modules/.bin/remix dev",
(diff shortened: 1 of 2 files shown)


docs/guides/envvars.md
```diff
@@ -43,13 +43,12 @@ Edit your `.env` file.
...
-Then update your package.json dev script to this:
+Then update your entry.server.tsx and add these lines before your imports
 
-```json lines=[2] filename=package.json
-{
-  "dev": "node -r dotenv/config node_modules/.bin/remix dev",
-  "start": "remix-serve build"
-}
+```js filename=entry.server.tsx
+  if(process.env.NODE_ENV = 'development') {
```
sergiodxa, line 49: You are using a single equal instead of two or three to compare, this is assigning development to the env variable.

lkostrowski: Of course, thanks

ryanflorence:
thanks but we'd rather not change app code for something like environment variables, also this wouldn't work well for ESM server builds that we will be supporting very soon :)

PR #1347, opened 2022-01-04
fix: sibling __layout routes conflict

jacob-ebey:


packages/remix-dev/config/routesConvention.ts
```diff
@@ -70,7 +70,7 @@ export function defineConventionalRoutes(
       let fullPath = createRoutePath(routeId.slice("routes".length + 1));
       let uniqueRouteId = (fullPath || "") + (isIndexRoute ? "?index" : "");
 
-      if (typeof uniqueRouteId !== "undefined") {
+      if (uniqueRouteId) {
```
ryanflorence, line 73: Is there a reason this doesn't use the actual unique routeId? They're all unique because routes must be a file in `app/`.

If I understand this code, it's making sure a route definition in `remix.config` doesn't define the same path as another route.

```js
route("/foo", "foo.tsx")
route("/foo", "bar.tsx")
```

But route IDs also need to be unique, so you can't point two paths to the same file:

```js
route("/foo", "a.tsx");
route("/bar", "a.tsx");
```

It seems that really all we need is the route's id and you can ditch the `uniqueRouteId` variable and first `if` altogether

```
if (uniqueRoutes.has(routeId)) {
  // ...
}
```

PR #1348, opened 2022-01-04
feat: Disambiguate useMatches

jacob-ebey:


kentcdodds (approved):
Sweet! This will fix a big I think I've run into as well 👍

kentcdodds:
Do the types need updating as well?

jacob-ebey:
> Do the types need updating as well?

Nope, there are no types for this. The return type is defined by the function return value in this case.

PR #1349, opened 2022-01-04
(fix:docs) small typo fix in jokes app tutorial

schalkneethling:
I believe lines 4256-4258 should be as follows:

```jsx
<div className="error-container">
    {`Huh? What the heck is "${params.jokeId}"?`}
</div>
```

### DIFF
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -4254,7 +4254,7 @@ export function CatchBoundary() {
   if (caught.status === 404) {
     return (
       <div className="error-container">
-        Huh? What the heck is "{params.jokeId}"?
+        {`Huh? What the heck is "${params.jokeId}"?`}
       </div>
     );
   }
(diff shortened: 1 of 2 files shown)


docs/tutorials/jokes.md
```diff
@@ -4254,7 +4254,7 @@ export function CatchBoundary() {
   if (caught.status === 404) {
     return (
       <div className="error-container">
-        Huh? What the heck is "{params.jokeId}"?
+        {`Huh? What the heck is "${params.jokeId}"?`}
```
schalkneethling, line 4257: Another alternative option:

```suggestion
        Huh? What the heck is {params.jokeId}?
```

schalkneethling:
Out of curiosity, would one ever hit lines 4598-4603?

```jsx
if (!joke) {
    throw new Response(
      "Can't delete what does not exist",
      { status: 404 }
  );
}
```

Would the `CatchBoundary` not already have caught this when the `404` is thrown in the `LoaderFunction` on line 4580?

ryanflorence:
you don't need a string template for JSX :)

PR #1350, opened 2022-01-04
Fix filename in Tutorial

michaseel:
the root.tsx contains the html structure of the app.


ryanflorence:
It used to be part of a layout but I suppose it's not anymore with the new bare-bones starter template. Thanks!

PR #1351, opened 2022-01-04
chore: add pm app example

kentcdodds:
Thanks @chaance!

clapping


kentcdodds:
Would probably be cool to have this deployed somewhere at some point, but having this example is pretty sweet for now.

PR #1352, opened 2022-01-05
feat(LiveReload): automatically determines the port to use and DCE

### LINKED ISSUE #1277: [Feature]: Make it so LiveReload automatically detects the correct port

Currently, the [LiveReload](https://github.com/remix-run/remix/blob/1fd70960e4d88740df5bf407a6ba2cd2b9549459/packages/remix-react/components.tsx#L1273) component accepts a `port` prop that accepts the port for the websocket server. This is a good start, but that port number really shouldn't be something app devs should have to think about. And once we add https://github.com/remix-run/remix/pull/871 (which I think we should) then the port will be non-deterministic.

I suggest we add a mechanism for the LiveReload component to know which port to use (via an environment variable for example).

Without this feature, the app developer will have to always specify the port and ensure it's always available which is annoying and kinda makes https://github.com/remix-run/remix/pull/871 less useful.

kentcdodds:
Closes #1277

TL;DR: now devs won't have to ever think about which port the dev server is set to, the `<LiveReload />` component will always use the right one. Combine this with @mcansh's https://github.com/remix-run/remix/pull/871 PR and devs wont even have to worry about configuring ports at all, we'll just choose an available one which will make it much easier for people to run multiple Remix apps at once.

Another thing this PR does is it makes it so devs won't have to put the `process.env.NODE_ENV` guard around their usage of the component anymore. Thanks to the magic of dead code elimination, if we just restructure things a bit we can make it so the production build eliminates our LiveReload component completely in production and devs don't have to think about it one way or another. (I manually tested this is how it works in the gists-app and it worked 😄).

This PR also adds a couple handy tools for our testing setup. Here's the typeahead plugin in action:

[typeahead functionality](https://github.com/jest-community/jest-watch-typeahead)

And here's the project selection in action:

[project selection](https://github.com/jest-community/jest-watch-select-projects).

This makes it a lot easier to slim down what tests you're running in our codebase. I have more ideas beyond this but it's a nice start. Also, we should probably upgrade to jest v27 eventually. It didn't immediately work when I first tried so I punted on that for now.

### DIFF
--- a/docs/guides/styling.md
+++ b/docs/guides/styling.md
@@ -637,9 +637,7 @@ Here's some sample code to show how you might use Styled Components with Remix (
            <Outlet />
            <ScrollRestoration />
            <Scripts />
-           {process.env.NODE_ENV === "development" && (
-             <LiveReload />
-           )}
+           <LiveReload />
          </body>
        </html>
      );
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -204,9 +204,7 @@ export default function App() {
       </head>
       <body>
         Hello world
-        {process.env.NODE_ENV === "development" ? (
-          <LiveReload />
-        ) : null}
+        <LiveReload />
       </body>
     </html>
   );
@@ -290,9 +288,7 @@ export default function App() {
       </head>
       <body>
         <Outlet />
-        {process.env.NODE_ENV === "development" ? (
-          <LiveReload />
-        ) : null}
+        <LiveReload />
       </body>
     </html>
   );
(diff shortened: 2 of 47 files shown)


kentcdodds:
I'll look into the test failures tomorrow. That happened to me locally when I tried to upgrade to Jest 27. Not sure what's up since this is running Jest 26 still. Must not be related I guess. Would appreciate reviews anyway :)

mcansh (approved):

kentcdodds (commented):
This is ready to go 👍

chris-kruining:
@kentcdodds is there somewhere I can see when this patch will reach stable? I would very much like to use this

kentcdodds:
I believe there's a way to subscribe to releases on GitHub. Once something is merged into dev, it'll be in the next release. We don't have a regular cadence for releases though.

dpcesx (commented):
Excellent work!

PR #1353, opened 2022-01-05
Add missing mkdir step to create the directory

BogdanDevBst:
On some operating systems touching a file without having a directory folder will fail. This extra step of explicitly creating the styles directory will ensure that this works for everyone.


ryanflorence:
ah thanks, original starter template had that directory, thanks!

PR #1363, opened 2022-01-05
docs(examples): Remix x Quirrel example

m5r:


examples/quirrel/app/routes/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+					</a>
+				</li>
+				<li>
+					<a target="_blank" href="https://remix.run/docs" rel="noreferrer">
+						Remix Docs
+					</a>
+				</li>
+			</ul>
+		</div>
+	);
```
kentcdodds, line 40: This could be simplified. If there's nothing UI relevant in this example then you can just say so in one line:

```suggestion
	return <p>Check the code. Nothing relevant in the UI.</p>;
```

examples/quirrel/app/routes/queues/greetings.ts
```diff
@@ -0,0 +1,3 @@
+import greetingsQueue from "~/queues/greetings.server";
+
+export const action = greetingsQueue;
```
kentcdodds, line 3: Is this code ever used?

examples/quirrel/app/routes/index.tsx
```diff
@@ -0,0 +1,41 @@
+import type { LoaderFunction } from "remix";
+
+import greetingsQueue from "~/queues/greetings.server";
+
+export const loader: LoaderFunction = async () => {
+	await greetingsQueue.enqueue("Groot");
```
kentcdodds, line 6: Could you add some comments here to describe a bit what a real-world use case for doing this in a loader would be?

kentcdodds (changes requested):
Thanks for this! Looking good. Just a few things.

examples/quirrel/app/routes/queues/greetings.ts
m5r: Yes, this is the action called by the Quirrel server. The actual body of the action is the second parameter passed to the `Quirrel()` constructor in `app/queues/greetings.server.ts`.

We can't directly export the action as below because Quirrel is supposed to be used only from the server and esbuild wouldn't be able to tree-shake the Quirrel import from the client-side bundle.
```ts
import { Queue } from "quirrel/remix";

export const action = Queue("/queues/greetings", async (name: string) => {
	console.log(`Greetings, ${name}!`);
});
```
```ts

import { action as greetingsQueue } from "~/routes/queues/greetings";

	await greetingsQueue.enqueue("Groot"); // browser error

	return null;
};
```

kentcdodds (approved):
Looks good! Thank you 👏👏

examples/quirrel/app/routes/index.tsx
m5r: I would say the better real-world use case would be in an action. For instance, a Stripe webhook handler would queue a job to properly handle its payload and immediately return a 200.

kentcdodds: If you'd like to update this example to do something more like that I think that would be good. You can still do console.log but maybe describe what you would typically use this for with a code comment. Thanks for taking the time :)

PR #1366, opened 2022-01-05
docs(example): example app with `mantine`

Johnson444:
This example features how to use [mantine](https://github.com/mantinedev/mantine) as the component library within a Remix app.

We wrap the app in `mantin`'s Theme and ColorMode providers so the pages can have app-wide dark/light mode. We are also creating a route to toggle between dark/light mode.

Screen Shot 2022-01-05 at 23 27 44


kentcdodds (changes requested):
Looks good to me! Just one thing

kentcdodds (approved):
Super! Thanks 😊

PR #1368, opened 2022-01-06
docs(examples): Add Remix Auth with GitHubStrategy example

sergiodxa:
This adds an example using [Remix Auth](https://github.com/sergiodxa/remix-auth) with the [GitHubStrategy](https://github.com/sergiodxa/remix-auth-github).


examples/remix-auth/github/.env.example
```diff
@@ -0,0 +1,3 @@
+GITHUB_CLIENT_ID=
+GITHUB_CLIENT_SECRET=
+BASE_URL=http://localhost:3000/
```
kentcdodds, line 3: I'd prefer that this be derived from the request at runtime like so: https://github.com/kentcdodds/kentcdodds.com/blob/main/app/utils/misc.tsx#L229

examples/remix-auth/github/README.md
```diff
@@ -0,0 +1,22 @@
+# Remix Auth - GitHubStrategy
+
+Authentication using Remix Auth with the GitHubStrategy.
+
+## Preview
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+[![Open in codesandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/remix-auth/github)
```
kentcdodds, line 9: Let's make this flat after all

```suggestion
[![Open in codesandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/remix-auth-github)
```

kentcdodds (changes requested):
Super! Just a couple things

examples/remix-auth/github/.env.example
sergiodxa: That should need to be supported by the strategy code, this needs to be initialized outside the loader

sergiodxa:
@kentcdodds I moved the code from remix-auth/github to remix-auth-github, I also moved the code of remix-auth/form to remix-auth-form

kentcdodds (approved):
Thanks!

PR #1369, opened 2022-01-06
docs(examples): Add on-demand hydration example

sergiodxa:
This shows how using the useShouldHydrate hook you can disable or enable JS on a per route basis and using the loader data to do it on-demand.


kentcdodds (approved):
This is great. Thank you 👏

PR #1370, opened 2022-01-06
tailwind example

juwiragiye:


examples/with-tailwindcss/postcss.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
sergiodxa, line 1: I think it would be better to show how to use Tailwind with the CLI instead of PostCSS, this can be another example for PostCSS usage.

juwiragiye: okay I'll make the changes

PR #1371, opened 2022-01-06
docs(examples): Add NProgress example

sergiodxa:
This adds an example showing how to use NProgress as a global progress bar for all transitions.


kentcdodds (approved):
Perfect! Thanks 👍

PR #1372, opened 2022-01-06
docs(examples): adds form to notion db example

Matthew-Mallimo:
Adding an example showcasing how to post data from a form to a Notion DB


examples/form-to-notion-db/.env
```diff
@@ -0,0 +1,2 @@
+NOTION_TOKEN=<notion-integration-key>
```
kentcdodds, line 1: Let's rename this file to `.env.example` to keep it consistent with other examples

kentcdodds (changes requested):
This is great! Just one thing

examples/form-to-notion-db/.env
Matthew-Mallimo: done!

kentcdodds (approved):
Super 👍

kentcdodds:
Thanks!

PR #1376, opened 2022-01-06
feat(compiler): CSS modules support

chaance:
⚠️ **NOTE:** This is a draft WIP, not ready for review or merging ⚠️

In its current state, CSS modules are compiled into a stylesheet that gets loaded via `<Links />` before all other links. This is enabled with `unstable_cssModules` in the Remix config.

This is currently working in the gists app, but there is an issue in dev mode with the watch script so the asset gets lost in the manifest after a rebuild.

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -11,6 +11,7 @@ import type { RemixConfig } from "./config";
 import { readConfig } from "./config";
 import { warnOnce } from "./warnings";
 import type { AssetsManifest } from "./compiler/assets";
+import type { RootAssets } from "./compiler/assets";
 import { createAssetsManifest } from "./compiler/assets";
 import { getAppDependencies } from "./compiler/dependencies";
 import { loaders } from "./compiler/loaders";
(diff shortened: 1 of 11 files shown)


packages/remix-dev/compiler.ts
```diff
@@ -1,4 +1,4 @@
-import { promises as fsp } from "fs";
+import * as fse from "fs-extra";
```
chaance, line 1: Note: this change isn't necessary anymore since I moved the plugin into a separate file. Not using anything from `fs-extra` in this file.

sergiodxa:
> ⚠️ **NOTE:** This is a draft WIP, not ready for review or merging ⚠️

CB717D43-D284-4365-A670-525E1F71A08D

fixtures/gists-app/app/components/Counter.js
```diff
@@ -1,9 +1,14 @@
 import { useState } from "react";
+import styles from "./Counter.module.css";
```
sergiodxa, line 2: Is this decomposable? Could I do `import { button } from "./Counter.module.css"`

packages/remix-dev/package.json
```diff
@@ -17,12 +17,15 @@
   "dependencies": {
     "cacache": "^15.0.5",
     "chokidar": "^3.5.1",
+    "css-tree": "^2.0.4",
```
chaance, line 20: Note: we may not need this afterall, but I'm going to experiment with using the AST to update relative URLs used in CSS. If it's not too much complexity I think we should support that.

fixtures/gists-app/app/components/Counter.js
chaance: Yes. All I'm doing is resolving this import as a JSON module that maps the object key to the hashed classname, so it'll work just like any other object import.

chaance:
Closing this as we've made some significant changes that will be dealt with in a separate PR. Related discussion here: https://github.com/remix-run/remix/discussions/2214

PR #1381, opened 2022-01-06
Ra/multiple forms one route

royanger:
Created a simple example of using multiple forms in one route.


kentcdodds (changes requested):
Thanks for taking the time on this! I'm sorry I didn't see it earlier, but we do have a multiple forms on one route example. If you'd like to improve it in any way, please do: https://github.com/remix-run/remix/tree/main/examples/multiple-forms

Thank you!

PR #1383, opened 2022-01-06
docs(examples): remove `package-lock.json`

MichaelDeBoey:
All original examples had these removed **and** they could cause problems over time, so I think it's best to delete them.


kentcdodds (changes requested):
I don't anticipate the examples getting changed very frequently so I think it's actually better to keep the lockfiles. Thanks though@

PR #1386, opened 2022-01-06
styled component example added

kumard3:


examples/with-styledcomponent/app/entry.server.tsx
```diff
@@ -0,0 +1,39 @@
+import ReactDOMServer,{renderToString} from "react-dom/server";
```
kentcdodds, line 1: The README.md should reference this file because it's relevant.

Might be useful to put some code comments in this file as well to describe what's going on and why different bits are needed.

examples/with-styledcomponent/README.md
```diff
@@ -0,0 +1,53 @@
+# Welcome to Remix!
```
kentcdodds, line 1: Could you update this README.md file to look like the other examples READMEs?

contributors.yml
```diff
@@ -89,3 +89,4 @@
 - weavdale
 - zachdtaylor
 - zainfathoni
+-kumard3
```
kentcdodds, line 92: Need a space here:
```suggestion
- kumard3
```

kentcdodds (changes requested):
Thanks for taking the time to put this together! Could you move the directory from `examples/with-styledcomponent` to `examples/styled-components`?

Thanks!

contributors.yml
kumard3: > Thanks for taking the time to put this together! Could you move the directory from `examples/with-styledcomponent` to `examples/styled-components`?
> 
> Thanks!

Sure

examples/with-styledcomponent/app/entry.server.tsx
kumard3: sure will be doing that

kumard3: I have done that please check

mikeatoctoml:
Testing this out (great timing! I just converted my project to styled-components!) Looks like it takes care of the flashing string for the initial style render that occurs by following the existing docs, however now there is a console error:

`react-dom.development.js:67 Warning: Prop `className` did not match. Server: "sc-edERGn dwZgnQ" Client: "sc-bdvvtL izgbrR"`


I wonder if that's possible to fix?

kentcdodds:
I've just updated this example as well as the docs. We've got a better approach and I'm aware of the issue you're talking about and hopefully Styled Components will enable us to fix this soon: https://github.com/styled-components/styled-components/issues/3660

kentcdodds:
I made a mistake when merging this and it's not on the `main` branch yet. It's on the `dev` branch. You can find the latest version of this example here: https://github.com/remix-run/remix/tree/dev/examples/styled-components

gopeter:
@kentcdodds looking at the example code (and the Codesandbox), the `styles-context.tsx` isn't used anymore, right?

kentcdodds:
Correct, you shouldn't need the styles context because we just do a string replace.

gopeter:
Unfortunately this doesn't seem to work with a CatchBoundary. I've created a [CodeSandbox](https://codesandbox.io/s/serene-bogdan-6id9f?file=/app/root.tsx), where you can see that the 404 page should work with a styled component too 😞

kentcdodds:
Ah yeah, good point... Hmmm... I'm not sure how to handle that case to be honest. Maybe the context approach is the better one here. Have you tried that approach and found it working for you?

PR #1387, opened 2022-01-06
docs(examples): Client-side validation

AriGunawan:
Add example on how to client-side validate form input as per suggested by the Remix documentation.


examples/client-side-validation/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+[![Open in CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/client-side-validation)
+
+## Example
+
+This example shows how to use the built-in HTML5 validation features.
+
+Check [app/routes/example.tsx](app/routes/example.tsx) to see the client-side validation in action.
```
kentcdodds, line 17: Could we put the example in `app/root.tsx` instead?

examples/client-side-validation/app/routes/example.tsx
```diff
@@ -0,0 +1,158 @@
+import {
```
kentcdodds, line 1: Could we move this stuff to `app/root.tsx`?

examples/client-side-validation/app/routes/example.tsx
```diff
@@ -0,0 +1,158 @@
...
+      - Number with min max: ${form.get("number-with-min-max")}
+      - Text with minlength maxlength: ${form.get("text-with-minlength-maxlength")}
+      - Date with min max: ${form.get("date-with-min-max")}
+  `;
+  return json({ message }, { status: 200 });
+};
+
+export default function Example() {
+  const actionData = useActionData();
+  const date = new Date();
```
kentcdodds, line 33: This will have SSR hydration issues since the server and client can be in different timezones. It would probably be better if these values came from the loader to avoid that.

kentcdodds (changes requested):
Looks great! Just a few things.

AriGunawan:
Thanks for the feedback. I learn a lot when creating this example 🙂
Please review it again. Thanks!

kentcdodds (approved):
Super! Thanks :)

PR #1388, opened 2022-01-06
docs(examples): Add a Strapi example

hollandThomas:
This adds a fairly minimal [Strapi](https://strapi.io/) headless CMS example.

I decided to open up a `examples/cms/` subdirectory as I expect more CMS examples to follow as per [this discussion](https://github.com/remix-run/remix/discussions/1064).

This uses the default `Quickstart` installation of Strapi, which uses SQLite (more info in README.md).

The example could be even more minimalistic. However, I decided to do some very simple markdown parsing with the help of [marked](https://github.com/markedjs/marked). I believe that this little touch of "real world"-ness might be justifiable, especially in a CMS example.

I did not add a codesandbox repository since I'm not sure how/if that would work in this scenario.


kentcdodds (changes requested):
Thanks for taking the time to put this together! It is a bit bigger than I want the examples to be, but I think it's fine. We decided to go with flat file structure for examples. Could you move this from `examples/cms/strapi` to `examples/strapi`?

hollandThomas:
Done!

kentcdodds (approved):
Thanks!

PR #1391, opened 2022-01-06
fix: specify node version for create-remix on cloudflare-pages

jkup:
Hey folks!

Cloudflare Pages uses Node version 12.18.0 by default and new Remix Pages apps will no longer build without specifying a Node version. I was able to get this template working by adding a `.node-version` file but @GregBrimble thought it would be a nice thing to add to the template.


LANCENET (approved):
Good

kentcdodds (approved):
LGTM 👍

kentcdodds:
Thanks @jkup!

PR #1393, opened 2022-01-06
feat: added import map support for client bundles

jacob-ebey:
feat: import maps can be external or bundled


sergiodxa:
Is this only for manually creating import maps?

jacob-ebey:
> Is this only for manually creating import maps?

What do you mean?

docs/api/conventions.md
```diff
@@ -81,6 +81,30 @@ The path to the server build, relative to remix.config.js. Defaults to "build".
 
 The port number to use for the dev server. Defaults to 8002.
 
+### importMapFile
+
+A path to an [import map](https://github.com/WICG/import-maps), relative to remix.config.js. This is used at build time to resolve imports. If [bundleImportMap](#bundleimportmap) is "true", the import paths must be HTTPS URL's.
```
mjackson, line 86: Can we be more specific here about when this is actually used? Something like:

A path to an import map (relative to remix.config.js) that is used at build time to resolve imports in the browser build. Also used by the Deno server runtime to resolve imports.

Is that accurate? Too soon to talk about Deno support?

sergiodxa:
> > Is this only for manually creating import maps?
> 
> What do you mean?

From what I understood, this is used so I can tell Remix that certain imports point to some URL, right?

It’s not Remix generating the import map, right?

docs/api/conventions.md
jacob-ebey: This is nothing to do with deno. The server build doesn't include this. Deno will use an import map and re-map bare module imports that way totally out of the scope of remix.

jacob-ebey:
> > > Is this only for manually creating import maps?
> > 
> > 
> > What do you mean?
> 
> From what I understood, this is used so I can tell Remix that certain imports point to some URL, right?
> 
> It’s not Remix generating the import map, right?

Correct. There are two ways you can use it:

Bundle Mode:
The import map will essentially become your lock file. When you build (or dev), the http imports defined in the import map are downloaded once and cached offline. From there on out it's essentially like using node_modules. This is very similar to how deno works (and uses the same mechanism under the hood).

No Bundle Mode:|
The import map is used to directly replace import statements with the URL's in the client Bundle. So instead of `import "react"` becoming `import "/chunk-reac123345.js"`, it becomes `import "http://your-url-in-import-map/index.js"`.

The idea behind bundle mode is to not require you to npm install all the dependencies when developing for a platform that doesn't use node_modules (such as deno). The only reason to have `node` installed will be to run `remix build`  `remix watch`.

Hope this helps.

jfsiii:
@jacob-ebey I'd love to see this feature. Can you share if it's coming back later or anywhere we can follow the status?

jacob-ebey:
More thought needs to be put into the approach. I will be brining import_map support back in some capacity in a future PR, but for now I don't feel this PR is the correct solution.

PR #1394, opened 2022-01-06
docs(examples): Add CMS Collected Notes example

sergiodxa:
This PR adds an example on how to use the CMS [Collected Notes](https://collectednotes.com) as a headless CMS to get the list of notes/articles on the Remix app.


kentcdodds (approved):
Super!

PR #1395, opened 2022-01-06
docs(examples): Add client only components example

sergiodxa:
This adds an example of how to work with components only working client side, it shows three different scenarios inside.


kentcdodds (approved):
Stellar! Thanks :)

PR #1396, opened 2022-01-06
fix: stop sorting params on data fetch

ryanflorence:
The prefetch link tag and the actual data being fetched were not the same because the data fetching code sorted the params. I'm the dork that sorted it in the first place, it was a mistake.

Now the href's match up and prefetching is actually useful for link with search params:

```jsx
// given this anchor
<Link to="/page?same=thing" />

// prefetch link
<link rel="prefetch" href="/page?same=thing&_data={routeId}" as="fetch" />`

// data fetch
fetch(`/page?same=thing&_data={routeId}`)
```

Updated the prefetch search params to use `set` instead of `append` because `_data` is a remix-reserved param, apps don't get to use it.

### DIFF
--- a/packages/remix-react/data.ts
+++ b/packages/remix-react/data.ts
@@ -37,7 +37,6 @@ export async function fetchData(
   submission?: Submission
 ): Promise<Response | Error> {
   url.searchParams.set("_data", routeId);
-  url.searchParams.sort(); // Improves caching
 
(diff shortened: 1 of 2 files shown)


mjackson:
Shouldn’t we just sort in both places?

jacob-ebey (approved):

PR #1397, opened 2022-01-06
feat: add deno package

jacob-ebey:
This PR adds a server adapter for Deno. Currently the approach to building a Remix app on Deno will be the same "bundle everything" approach that we are taking with CF.

Deno has "legacy" support for bare module imports from node_modules, but this is not in line with the spirit of native URL imports Deno is great at, and not all Deno deployment targets support it. I will be re-visiting import_map support, both bundling and externalized. This will require more thought as it must support a good UX and DX for Deno *and* browsers.


mcansh:
i think we want/need to add this to the `workspaces` in the root package.json too

packages/remix-deno/server.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+}: {
+  build: ServerBuild;
+  mode?: string;
+  staticFiles?: {
+    cacheControl?: string | ((url: URL) => string);
+    publicDir?: string;
+  };
+}) {
+  let remixHandler = createRequestHandler(build, {}, mode);
```
mjackson, line 29: Let's split this up into 3 separate functions:

- `createRequestHandler`
- `createRequestHandlerWithStaticFiles`
- `serveStaticFiles`

packages/remix-deno/server.ts
```diff
@@ -0,0 +1,69 @@
+import { createRequestHandler } from "@remix-run/server-runtime";
+import type { ServerBuild } from "@remix-run/server-runtime";
+
+import mime from "mime";
+
+export function defaultCacheControl(url: URL) {
+  if (url.pathname.startsWith("/build/")) {
```
mjackson, line 7: The `/build/` should be an arg we get from the user.

packages/remix-deno/server.ts
```diff
@@ -0,0 +1,69 @@
...
+        headers.set("Cache-Control", cacheControl(url));
+      } else if (cacheControl) {
+        headers.set("Cache-Control", cacheControl);
+      } else {
+        headers.set("Cache-Control", defaultCacheControl(url));
+      }
+
+      // @ts-expect-error
+      const file = await Deno.readFile(
+        `${publicDir || "./public"}${url.pathname}`
```
mjackson, line 51: Let's use something like `path.join` here to resolve the file path.

mjackson (changes requested):

packages/remix-deno/server.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  return new Response(file, { headers });
+}
+
+export function createRequestHandlerWithStaticFiles<Context = unknown>({
+  build,
+  mode,
+  getLoadContext,
+  staticFiles = {
+    publicDir: "./public",
```
mjackson, line 80: Should we have a `cacheControl` option here as well to match the args of `serveStaticFiles`?

packages/remix-react/package.json
```diff
@@ -14,6 +14,7 @@
   "main": "index.js",
   "module": "esm/index.js",
   "dependencies": {
+    "history": "^5.2.0",
```
mjackson, line 17: Is this just something that was missing? I thought we didn't need an explicit dep on `history` now...

jest.config.js
```diff
@@ -11,6 +11,19 @@ module.exports = {
       testEnvironment: "node",
       testMatch: ["<rootDir>/packages/create-remix/**/*-test.[jt]s?(x)"]
     },
+    {
+      displayName: "deno",
+      testEnvironment: "jsdom",
+      testMatch: ["<rootDir>/packages/remix-deno/**/*-test.[jt]s?(x)"],
+      moduleNameMapper: {
+        "https://deno.land/std/path/mod.ts":
+          "<rootDir>/packages/remix-deno/__tests__/pathmock.js"
```
mjackson, line 20: Clever 👍

packages/remix-deno/server.ts
```diff
@@ -0,0 +1,106 @@
...
+    cacheControl,
+    publicDir = "./public",
+    assetsPublicPath = "/build/"
+  }: {
+    cacheControl?: string | ((url: URL) => string);
+    publicDir?: string;
+    assetsPublicPath?: string;
+  }
+) {
+  const url = new URL(request.url);
```
mjackson, line 53: nit: use `let` instead of `const`

packages/remix-react/package.json
jacob-ebey: Whenever we import from a package in code it should be listed in deps regardless of if it's a dep of another dep we have listed. It helps bundlers and ESM CDN's like esm.sh. Without it more config is required to get remote modules to load. This is currently the only thing stopping us from consuming @remix-run packages from esm.sh.

packages/remix-deno/server.ts
jacob-ebey: It defaults later on in the codepath in the serveStaticFiles if it's not provided :) No need to add it here.

packages/remix-react/package.json
mjackson: Yeah, I get that. I guess what I'm saying is let's remove whatever imports we're using from `history` in `remix-react` so we don't have to do this

mjackson (approved):

packages/remix-react/package.json
jacob-ebey: `parsePath` and `history` need to be re-exported before we can do this.

PR #1408, opened 2022-01-07
ci: prerelease test infrastructure

mcansh:
This PR adds automated testing of `prereleases` to Netlify, Vercel, Fly, Cloudflare Pages, Cloudflare Workers, and AWS (via architect).

first and foremost, `{target}-prerelease-deployment.test.yml` GitHub Action for each target that runs a node script to create a new project on the given target with the name `remix-{target}-{sha}` and runs a very basic cypress assertion on the dev server and the deployed url.


.github/workflows/arc-prerelease-deployment.test.yml
```diff
@@ -0,0 +1,46 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: ./examples/arc-prerelease-deployment-test
```
kentcdodds, line 21: I'll give this a deeper look later, but one thing I'd like is if we can use create-remix instead of just using the pre-built example directories. That way we get more confidence out of these tests that the generators work.

mcansh: agreed, just need to tweak `create-remix` a bit to allow for easier app creation

mcansh: alllllrighty @kentcdodds, just spent the last ~4 hours just to realize the default npm version in GitHub actions is 6... but... each app is created on the fly now

scripts/deployment-test/arc.mjs
```diff
@@ -0,0 +1,108 @@
+import path from "path";
```
kentcdodds, line 1: beautiful

kentcdodds (changes requested):
This is brilliant! Very well done @mcansh 👏

Just one thing I think either you missed or I misunderstand.

.github/workflows/cf-workers-prerelease-deployment.test.yml
```diff
@@ -0,0 +1,47 @@
+name: cloudflare workers prerelease deployment test
+
+on: push
+#   release:
+#     types: [prereleased]
```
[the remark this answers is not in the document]
mcansh: oh right! meant to ask you about that.. can you add a cloudflare secret

.github/workflows/cf-pages-prerelease-deployment.test.yml
```diff
@@ -0,0 +1,49 @@
+name: cloudflare pages prerelease deployment test
+
+on:
+  - push
+#   release:
+#     types: [prereleased]
```
kentcdodds, line 6:
```suggestion
  release:
    types: [prereleased]
```

scripts/deployment-test/fly-dockerfile
```diff
@@ -0,0 +1,44 @@
+# base node image
```
kentcdodds, line 1: Does the create-remix fly not generate this properly for us? I wouldn't expect us to need to have this file.

kentcdodds (changes requested):
Super! Just two quick things.

scripts/deployment-test/fly-dockerfile
mcansh: it probably does, iirc i added it when i tried to make the app on fly with the api - let me verify

mcansh: yup, it does! removing now :)

kentcdodds (approved):
This is awesome work. Great job Logan. Have a good weekend!

cmd-johnson:
Hi there!
I'm currently working on a PR for getting yarn PnP to work with Remix (#1316). Today I rebased my changes to the current `dev` branch and the `create-remix` command no longer works. The prompts all show, but after creating the project directory, the program just exits. Using `git bisect`, I pinned this down to changes made in this PR. Commit c8dc234fd8346e45a6fbfa33c4e39d85ba6b803f to be precise.

I'm not quite sure what's happening here.
The program runs fine until https://github.com/remix-run/remix/blob/c8dc234fd8346e45a6fbfa33c4e39d85ba6b803f/packages/create-remix/index.ts#L37 after which it just exits without giving any reason whatsoever.

cmd-johnson:
Oooooh, I see what's happening
There's an `await` missing here: https://github.com/remix-run/remix/blob/c8dc234fd8346e45a6fbfa33c4e39d85ba6b803f/packages/create-remix/cli.ts#L104

kentcdodds:
Thanks for letting us know! We've fixed that now :) Sorry for the trouble.

PR #1410, opened 2022-01-08
Update blog.md

jmasson:
I'm very new to the stack and Typescript, so perhaps there I missed something, but I needed to do this for it to work.


kentcdodds (approved):
Perfect. Thanks!

PR #1411, opened 2022-01-08
Docs (Example): codesandbox fix - added "@mantine/hooks" package

Johnson444:
codesandbox can't resolve the hooks package and throws an error; so we have to add it manually. I will check again when server container is out of beta.


kentcdodds (approved):
👍

PR #1413, opened 2022-01-08
Docs (Example) : With Ant-Design

kumard3:


examples/ant-design/app/entry.server.tsx
```diff
@@ -0,0 +1,23 @@
+import { renderToString } from "react-dom/server";
+import { RemixServer } from "remix";
+import type { EntryContext } from "remix";
+//imported the ant design styles
+import 'antd/dist/antd.css'
+import './styles/index.css'
```
buraksaraloglu, line 6: Hi, I saw that you've imported AntD css files two times both in the 
`client.server.tsx` and `root.tsx` 
I was curious if that was on purpose or did you forgot it?

kumard3: sorry i have done it by mistake

examples/ant-design/app/root.tsx
```diff
@@ -0,0 +1,36 @@
...
+  Meta,
+  Outlet,
+  Scripts,
+  ScrollRestoration
+} from "remix";
+import type { MetaFunction } from "remix";
+
+//imported the ant design styles
+import 'antd/dist/antd.css'
+import './styles/index.css'
```
kentcdodds, line 13: This won't get those styles on the page. You have to add a `links` export: https://remix.run/docs/en/v1/guides/styling#regular-stylesheets

kentcdodds (changes requested):
This looks great, just one thing.

examples/ant-design/app/root.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 32:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

examples/ant-design/app/root.tsx
MichaelDeBoey, line 32:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):
Also make sure to set `main` as the base branch

PR #1414, opened 2022-01-08
Update documentation about fetcher.load()

zachdtaylor:
`fetcher.load()` has to be called from within a `useEffect` or a callback. I added an explanation of this to the documentation.


ryanflorence:
thanks, I was wondering why heck so many folks were trying to call `load` as they rendered 😆 I already updated the doc elsewhere though, thanks for the error message PR though!

PR #1429, opened 2022-01-08
Remove unneeded while loop

gavriguy:
Removing  unneeded while loop in the Upstash example.

Thanks @donavon for letting me know.


donavon:
Thank YOU for this very useful example.

LANCENET (approved):

ryanflorence:
why was that there in the first place? wouldn't this just totally crash a server?

donavon:
@ryanflorence no b'cuz is returning inside of the loop. It's just unnecessary and confusing.

kentcdodds:
I have no idea how I missed this 😅 thanks for cleaning this up

donavon:
Not a problem. I saw it and went whoa!

PR #1433, opened 2022-01-09
Add Cloudflare Pages package to bug-report template

rppld:


ryanflorence:
thanks, but we don't have those checkboxes anymore :)

PR #1438, opened 2022-01-09
feat(example): Remix + toast messages

Girish21:


examples/toast-message/app/message.server.ts
```diff
@@ -0,0 +1,35 @@
...
+    path: "/",
+    httpOnly: true,
+    sameSite: "lax",
+    expires: new Date(Date.now() + ONE_YEAR),
+    secrets: ["SUPER_SECRET"],
+    secure: true
+  }
+});
+
+const setSuccessMessage = async (request: Request, message: string) => {
```
sergiodxa, line 17: I would make these functions receive the session object so you share a session created before calling them, which usually happens at the end of the action function.

Also, a little nitpicking, they could be created using the function keyword instead of an arrow 😛

examples/toast-message/app/root.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  return json<LoaderData>(
+    { message, type },
+    { headers: { "Set-Cookie": await commitSession(session) } }
+  );
+};
+
+export default function App() {
+  const data = useLoaderData<LoaderData>();
```
sergiodxa, line 46: I would restructure this so the effect below only depend on type and message.

Girish21: if I de-structure the loader data here

```js
const {message, type} = useLoaderData()

React.useEffect(() => {
  // effect is not triggered when the message and type
  // from previous fetch is the same
}, [message, type])
```

if the value between two fetch are the same, the effect did not get triggered (eg. fetch 1 -> error, fetch 2 -> error).

but since the object will change after every time the loader is called, it triggered the effect 😅

examples/toast-message/app/message.server.ts
```diff
@@ -0,0 +1,28 @@
...
+    sameSite: "lax",
+    expires: new Date(Date.now() + ONE_YEAR),
+    secrets: ["SUPER_SECRET"],
+    secure: true
+  }
+});
+
+const setSuccessMessage = (session: Session, message: string) => {
+  session.flash("message", message);
+  session.flash("type", "success");
```
kentcdodds, line 20: Thoughts on doing this instead?

```suggestion
  session.flash("toastMessage", {type: 'success', message});
```

examples/toast-message/app/root.tsx
kentcdodds: Let's a code comment explaining that (because a lot of people will probably bump into this.

kentcdodds (changes requested):
Love this! Just two things.

examples/toast-message/app/root.tsx
sergiodxa: If you do the change to have a single flash key with both type and message of the toast, then you could destructor it here

examples/toast-message/app/root.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <Toaster />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 79:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

girishk21:
Oops, @kentcdodds forgot about this. Have addressed the comments, let me know if anything else is required. Thanks!

Girish21:
Oops, @kentcdodds forgot about this. Have addressed the comments, let me know if anything else is required. Thanks!

kentcdodds (approved):
Love this 👏👏 Thank you for putting it together!

PR #1439, opened 2022-01-09
docs(example): add Remix Auth with SupabaseStrategy example

rphlmr:
This adds an example using [Remix Auth](https://github.com/sergiodxa/remix-auth) with the [SupabaseStrategy](https://github.com/mitchelvanbever/remix-auth-supabase-strategy).


kentcdodds (approved):
Stellar! Thank you 👏👏

PR #1441, opened 2022-01-09
Material UI example

Anishpras:


examples/material-ui/README.md
```diff
@@ -0,0 +1,53 @@
+# Welcome to Remix!
```
kentcdodds, line 1: Could you update this README to follow the format that's in the `examples/template/README.md`?

examples/material-ui/app/entry.server.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  styles.forEach(({ key, ids, css }) => {
+    const emotionKey = `${key} ${ids.join(' ')}`;
+    const newStyleTag = `<style data-emotion="${emotionKey}">${css}</style>`;
+    stylesHTML = `${stylesHTML}${newStyleTag}`;
+  });
+
+  // Add the emotion style tags after the insertion point meta tag
+  const markup = html.replace(
+    /<meta(\s)*name="emotion-insertion-point"(\s)*content="emotion-insertion-point"(\s)*\/>/,
```
kentcdodds, line 49: I think this could be much simpler if this were something like `__STYLES__` (like the styled-components example).

kentcdodds (changes requested):
Looks great to me 👍 Just a couple changes needed.

examples/material-ui/app/entry.server.tsx
Anishpras: @kentcdodds I think it is correct because in the styled-components example the __STYLES__ was replaced from the extracted CSS and in material-UI, we need to add the emotion style tags after the insertion point meta tags and this is a one-time thing as shown in the official material-UI examples.

examples/material-ui/README.md
```diff
@@ -0,0 +1,53 @@
+# Material-UI Example
```
sergiodxa, line 1: This README doesn’t follow the other examples style, for example it doesn’t have a button to try it in CodeSandbox

examples/material-ui/app/root.tsx
```diff
@@ -0,0 +1,37 @@
...
+        {title ? <title>{title}</title> : null}
+        <Meta />
+        <Links />
+        <meta name="emotion-insertion-point" content="emotion-insertion-point" />
+      </head>
+      <body>
+        {children}
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === 'development' && <LiveReload />}
```
MichaelDeBoey, line 23:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

kentcdodds:
I think you may need to apply the same strategy that's done in #1485 to avoid some pit-falls with CSS-in-JS. Can you give that a look?

PR #1442, opened 2022-01-09
docs(examples): resize images on the fly using sharp

leon:
This example shows how to build a stream pipeline using sharp.

The dogs in the example are creative commons images I found on google images, but maybe we should change them to some remix images?

Let me know what you think 🚀


examples/image-resize/README.md
```diff
@@ -0,0 +1,45 @@
...
+npm run dev
+```
+
+Visit
+http://localhost:3000/ to view the resized images
+
+The `Image` component is a simple wrapper which creates the resize url for our `/assets/resize/$.ts` route
+
+```
+http://localhost:3000/assets/resize/dog-1.jpg?w=600&h=600&fit=cover
```
kentcdodds, line 25: Could you make this work to point to external images as well?

```suggestion
http://localhost:3000/assets/resize?url=%2Fdog-1.jpg&w=600&h=600&fit=cover
```

Then you could optimize even hot-linked images 🔥

kentcdodds: We could argue that that's beyond the scope of this example, but maybe you could add a comment about this and say it's possible and leave it as an exercise for the reader.

examples/image-resize/app/root.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export default function App() {
+  return (
+    <html lang="en">
+      <head>
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        {/* only for demo purposes, do not use tailwind play cdn for your live website */}
+        <script src="https://cdn.tailwindcss.com"></script>
```
kentcdodds, line 15: We don't need the styling to be nice in these examples. Could you remove this (and tailwind classes)? (Keep in mind that most people will just look at the code anyway).

```suggestion
```

examples/image-resize/app/routes/assets/resize/$.ts
```diff
@@ -0,0 +1,129 @@
...
+ * The implementation of the demo uses a stream based approach where the image is never stored in memory.
+ * This means it's very good at handling images of any size, and is extremely performant.
+ * Further improvements could be done by implementing ETags, but that is out of scope for this demo.
+ */
+
+import { LoaderFunction } from 'remix'
+import sharp, { FitEnum } from 'sharp'
+import path from 'path'
+import { createReadStream, statSync } from 'fs'
+import { PassThrough } from 'stream'
```
kentcdodds, line 19: Could you mention in the README that this requires a Node.js environment?

examples/image-resize/app/routes/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { Image } from '~/components/image'
+
+export default function Index() {
+  return (
+    <div className="container">
+      <h1 className="mt-12 text-4xl">Cover</h1>
+      <div className="flex gap-4 items-center overflow-x-scroll">
```
kentcdodds, line 7: Could use inline styles for this example.

kentcdodds (changes requested):
This is sweet! Just a few suggestions.

kentcdodds:
Oh, and the dog images are adorable 👍

gavriguy:
This example is great. @leon I wonder if it's possible to create a dynamic route inside Remix that will resize the image by passing the image URL and size properties through query params (something similar to Cloudinary). This way we can use it on any React component that requires image `src`

examples/image-resize/README.md
leon: This isn't really a scalable demo, since usually you don't store your images in the source code, if it isn't your personal blog.

The code for creating a read stream to GCP isn't much

```
const {Storage} = require('@google-cloud/storage');

const storage = new Storage();

// create read stream directly to gcp bucket
const bucketName = 'my-gcp-bucket'
const bucketPath = '/dogs/cute/dog-1.jpg'
const readStream = storage.bucket(bucketName).file(bucketPath).createReadStream()

```

but since it requires that we have a gcp bucket and permissions to read from it, it would require some more work.
But I think it will be something people will ask for.

Should we add it as a secondary route, `/assets/google-storage-resize/$.ts`?

or maybe a totally different example?

examples/image-resize/app/root.tsx
leon: Sure, I can remove it.
it was only for testing purposes

examples/image-resize/app/routes/assets/resize/$.ts
leon: Will do

examples/image-resize/app/routes/index.tsx
leon: I'll fix this together with the tailwind removal

leon:
@gavriguy This is how it's implemented, I wrote about it in the README file.

https://github.com/remix-run/remix/blob/5a5044241096698eb6156bc526e2b4a709baeb29/examples/image-resize/README.md

leon:
@kentcdodds I will have a look at the improvements this evening

gavriguy:
> @gavriguy This is how it's implemented, I wrote about it in the README file.
> 
> https://github.com/remix-run/remix/blob/5a5044241096698eb6156bc526e2b4a709baeb29/examples/image-resize/README.md

@leon Thanks for clearing that out. 
This is a great contribution.

examples/image-resize/README.md
leon: I added a commented out section for both GCP and S3

kentcdodds (approved):
This is awesome. Love it. Thanks for putting this together 👍

PR #1447, opened 2022-01-10
docs(jokes): update loader return type to match other samples

mattmazzola:
### Loader return type
The other examples in Jokes guide use pattern of return a `data` object directly.
There is one instance where it returns `{ data }` instead of `data`

### Using loader data in component
The return type is shaped as `type LoaderData = { users: Array<User> }` which means the array is on the `users` object not the `data` object. 

However, the template shows `data.map(...)` but it should be `data.users.map(...)`


kentcdodds (approved):
Thank you!

PR #1458, opened 2022-01-10
docs(examples): add Prisma example

juwiragiye:


examples/prisma/app/root.tsx
```diff
@@ -0,0 +1,27 @@
+import {
+  Links,
+  LiveReload,
+  Meta,
+  Outlet,
+  Scripts,
+  ScrollRestoration
+} from "remix";
+
```
kentcdodds, line 9: Could you add a loader/action here that allows people to do some basic CRUD (minimal is better). I think it would be helpful for people to see what it takes to make things work together with Remix.

kentcdodds (changes requested):
This looks awesome! Just one thing.

examples/prisma/app/root.tsx
```diff
@@ -0,0 +1,27 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 23:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

PR #1460, opened 2022-01-11
Add dynamic websocket host

warmpigman:
Changes static localhost websocket host url to dynamic, fetches the host and applies that + port. #1264 Proposes a suggestion via specifying the host but that means static and you would have to type it in if you were to change/the first time.


ryanflorence:
fixed in #1685, sorry for not seeing this first :)

PR #1462, opened 2022-01-11
feat(cloudflare-workers): edge cache static assets

jacob-ebey:
Part 1 from PR https://github.com/remix-run/remix/pull/924 by @GregBrimble.

~~This was pulled into its own PR because I don't think we can do part 2 because unless I misunderstand CF's edge cache, they operate on the `max-age` header and not `s-maxage`. This would land you in cases where you want to `max-age=5` or some small value for a piece of user-specific content in just their browser and have it end up cached at the edge and served to other visitors.~~ I forgot about the `private` directive 🤦‍♀️

Still good to pull it into its own PR to give it a good test run and make sure CF isn't caching things we don't expect them to be.

### DIFF
--- a/packages/remix-cloudflare-workers/worker.ts
+++ b/packages/remix-cloudflare-workers/worker.ts
@@ -69,6 +69,8 @@ export async function handleAsset(
     let requestpath = url.pathname.split("/").slice(0, -1).join("/");
 
     if (requestpath.startsWith(assetpath)) {
+      // Assets are hashed by Remix so are safe to cache in the browser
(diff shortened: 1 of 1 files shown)


GregBrimble (approved):
You've dropped [a related comment from the original PR](https://github.com/remix-run/remix/pull/924/files#diff-a105e364b2439defa113dd21b9acb19f2addc20d0f7481bf18ba7841121d0fddR72-R73), but other than that minor nit, LGTM :+1:

PR #1465, opened 2022-01-12
docs(examples): fix `eslintrc`

MichaelDeBoey:


examples/.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   rules: {
-    "prefer-const": "warn",
+    "prefer-const": "error",
```
MichaelDeBoey, line 3: It's seen as `error` by ESLint because of the deleted file

kentcdodds (approved):
Thanks!

PR #1468, opened 2022-01-12
chore: charSet -> charset

web2033:


sergiodxa:
In React, the charset attribute is defined as charSet with the S capitalized as it’s currently being used.

See https://reactjs.org/docs/dom-elements.html

Issue #1244, opened 2021-12-26
[Bug]: error on vercel adapter when returning a stream


mbarto:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [X] `@remix-run/vercel`

### What version of Remix are you using?

1.1.1

### What version of Node are you using? Minimum supported version is 14.

16.3.0

### Steps to Reproduce

Use the following Resource Route to fetch a remote image and return it as a response:

```typescript
import { LoaderFunction } from "remix"

export const loader: LoaderFunction = async ({ params }) => {
    const imageRes = await fetch(<link to any remote PNG image>)

    return new Response(imageRes.body, {
        status: 200,
        headers: {
            "Content-Type": "image/png",
        },
    })
}
```

### Expected Behavior

When calling the Resource Route, I expect to get back the remote image in the browser. 
It works locally with Remix App Server

### Actual Behavior

When deployed on Vercel, I get an error:

```
[...]

PR #1470, opened 2022-01-12
fix(vercel): allow for streams to be returned as the response body

mbarto:
…e a stream passed as the response body

See #1244 for additional info


mcansh (approved):

mcansh:
thanks! we do this in the express adapter, but didn't update this one

PR #1477, opened 2022-01-13
docs(examples): Add ioredis example

vimutti77:


inakiAndres (approved):

sergiodxa:
Why name the example `ioredis` instead of just Redis?

inakiAndres:
> Why name the example `ioredis` instead of just Redis?

I think is because of the library

vimutti77:
> > Why name the example `ioredis` instead of just Redis?
> 
> I think is because of the library

Yes, It is the name of the library.

But if you thing that `redis` is make more sense. I can update the example name to it and may be change the library from [ioredis](https://github.com/luin/ioredis) to [redis](https://github.com/redis/node-redis).

examples/ioredis/app/root.tsx
```diff
@@ -0,0 +1,27 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 23:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

examples/ioredis/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Preview
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+[![Open in codesandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/ioredis)
+
+## Usages
+
+- [Install](https://redis.io/topics/quickstart) or use your existing Redis server.
+- Duplicate the local `.env.example` file to `.env` and change the REDIS_URL environment variable to your redis server URL.
```
kentcdodds, line 14: What do you say about including a docker-compose file that starts a redis server? I do that in my app: https://github.com/kentcdodds/kentcdodds.com/blob/2af8efc6cebf25052f1f95f2210317a8086d74ef/docker-compose.yml#L3-L11

That way people can get redis running without having to install redis and configure themselves.

kentcdodds (changes requested):
This is almost perfect! Thank you 👏👏 Just one idea.

examples/ioredis/README.md
vimutti77: I think we shouldn't add the docker-compose file to the example.

Because it only useful for people who are already using docker and don't have redis installed. 
And it may cause other people to get the file that are not irrelevant to them.

So I think we just add about it to README.md should be enough.

vimutti77, line 14:
```suggestion
## Usages

- Use your existing redis server or [install new redis server](https://redis.io/topics/quickstart) or [start redis server with docker](https://hub.docker.com/_/redis).
- Duplicate the local `.env.example` file to `.env` and change the REDIS_URL environment variable to your redis server URL.
```

@kentcdodds may be something like this.

PR #1481, opened 2022-01-13
Split Newsletter example in two files to avoid missleading the reader

gustavoguichard:
By reading the docs my colleague got confused about this bit of the docs so in a call we figured out it was talking about 2 different files.
I hope this change speaks for itself, let me know if I need to work further on this PR.


gustavoguichard:
I think this PR can be closed in favor of #1482
Cheers

PR #1485, opened 2022-01-13
docs(examples): Add Emotion example 

helderburato:


examples/emotion/app/root.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+        {styles}
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 33:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

examples/emotion/app/root.tsx
helderburato: Thanks! :D

danestves:
Hi, can you test if this implementation works with Cloudflare Worker? I try it, but it's appearing that emotion is calling some packages that don't work in a worker

kentcdodds (changes requested):
Thanks for taking the time here!

Could you add one or two routes to make sure that it works across those route boundaries? And double-check that `CatchBoundary` and `ErrorBoundary` on the root route will work because I'm not certain it will.

helderburato:
Hey @kentcdodds!

If I catch this well, I added two new routes `/jokes` and `/jojes/:id` and a `CatchBoundary` to the root`. Is that what you expect?

The styles still working through the routes.

kentcdodds:
My big concern is that if there's a CatchBoundary or an ErrorBoundary and those are rendered due to an error, all the styles will be removed because the `App` (where the styles are rendered) will be removed from the page. If you can verify that's not a problem then we can go forward with this. Otherwise we may need to come up with a different approach.

helderburato:
If I catch it well, I triggered a catch/error boundary into the `jokes/:jokeId` and `jokes/not-found`, like the following screenshots:
Screenshot 2022-02-15 at 09 29 43
Screenshot 2022-02-15 at 09 29 55

And the styles persist into the nested routes.

Is that what you expect @kentcdodds ?

kentcdodds:
What if you have a `CatchBoundary` and `ErrorBoundary` in the `root.tsx` file and those are rendered? My guess is the styles will all be removed in that case because the `App` (which renders the styles) will be replaced by the `CatchBoundary` and the `ErrorBoundary`.

kentcdodds:
Here are more details on the issue with CSS-in-JS: https://github.com/remix-run/remix/pull/1512#issuecomment-1042413526

helderburato:
Hi @kentcdodds, I updated the project at the commit https://github.com/remix-run/remix/pull/1485/commits/7894b95529414673990e414768a2154b2b8dec0c.

The behavior looks like this:

### Root Error
Screenshot 2022-02-17 at 09 08 44

### CSS Rendered
Screenshot 2022-02-17 at 09 08 54

Is that the expected?

kentcdodds:
If it's server rendered then it works fine. But it breaks for client transitions. Can you try doing the same thing I'm doing in that video?

ulentini:
I was able to reproduce the problem and it actually breaks the client, like with other CSS-in-JS libraries.
I was also able to fix it starting from @primos63 repo (where he fixed it for ChakraUI): https://github.com/primos63/chakra-remix-example

@helderburato, let me know if you want me to make a PR to your branch.

helderburato:
What do you think about @kentcdodds ?

kentcdodds (changes requested):
This is awesome! Could you have a section in the README that describes how the example works and why you need to interact with the cache? Also make reference to the relevant files in the example. Then this will be ready to merge. Thanks!

kentcdodds (approved):
Super 👍

This is very helpful. Thank you 👏👏

PR #1486, opened 2022-01-13
Fix typecheck

zigomir:
With `"noEmit": true` already in `tsconfig.json`,
`tsc` does all the type checking you need.

(didn't find these docs in `dev` branch)


ryanflorence:
you're saying `-b` is redundant because of the tsconfig?

PR #1488, opened 2022-01-13
docs(examples): fix expires calculation in redis session

karimsan:
Method `getSeconds()` returns the seconds of the specified date (from 0 to 60) and it leads to incorrect expiration time calculation.

For correct total time difference calculation should be used method `getTime()` - it returns milliseconds since the Unix Epoch.


kentcdodds (approved):
Thanks!

Issue #1426, opened 2022-01-08
[Bug]: Compiling Cloudflare Worker results in error from Wrangler


kpalmvik:
### What version of Remix are you using?

1.1.1

### What version of Node are you using? Minimum supported version is 14.

16.13.1

### Steps to Reproduce

Create a new Remix `npx create-remix@latest` and select "Cloudflare Pages" / "TypeScript" / "Run npm".

Run `npm run dev` to start the dev server.

### Expected Behavior

The dev server starts without any errors.

### Actual Behavior

An error occurs:

```
...
> remix watch

Watching Remix app in development mode...
Compiling worker to "/var/folders/5w/rj0_j25j70g2_qcfpnr8w6m40000gn/T/functionsWorker.js"...
✘ [ERROR] Expected ";" but found ":"

    build/assets.json:2:11:
      2 │   "version": "25ffdec3",
        │            ^
        ╵            ;

1 error(s) and 0 warning(s) when compiling Worker.
Build failed with 1 error:
build/assets.json:2:11: ERROR: Expected ";" but found ":"
ERROR: "dev:wrangler" exited with 1.
```

This might be related to changes to the [most recent version of Wrangler `0.0.7`](https://github.com/cloudflare/wrangler2/releases/tag/wrangler%400.0.7), since everything works if I pin the version to `0.0.6`

universse:
Same thing happens with Cloudflare Pages. Probably a wrangler issue.

Rocinante89:
Cloudflare Pages is throwing the below for me on `npm run dev`

```
Watching Remix app in development mode...
Compiling worker to "/tmp/functionsWorker.js"...
✘ [ERROR] Could not resolve "../build"

    functions/[[path]].js:4:23:
      4 │ import * as build from "../build";
        ╵                        ~~~~~~~~~~

1 error(s) and 0 warning(s) when compiling Worker.
Build failed with 1 error:
functions/[[path]].js:4:23: ERROR: Could not resolve "../build"
ERROR: "dev:wrangler" exited with 1.
```

markshenouda:
Same problem here

traviseden:
I came across issue doing with Cloudflare Pages and solved it by running `npm run build` first. Then it worked as expected for `npm run dev`.

Wow! Remix is even faster than I was expecting.

markshenouda:
Nope it didn't work for me :/

universse:
From Remix discord, can try `wrangler@beta` instead of `alpha`.

[...]

PR #1449, opened 2022-01-10
Use wrangler@beta in Cloudflare Pages template

GregBrimble:
We've managed to [automatically watch dependencies outside of just the `functions` folder](https://github.com/cloudflare/wrangler2/issues/49), and that has now been added to the `beta` tag as we gear up for an official release.

PR #1489, opened 2022-01-14
Cloudflare pages wrangler build fix

jerryharrison:
Fixes for #1426

I've noticed that the correct version for cloudflare pages builds is `beta` not `alpha`. I've been able to confirm this with two projects and with several successful builds.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -94,3 +94,4 @@
 - anishpras
 - kumard3
 - mattmazzola
+- jerryharrison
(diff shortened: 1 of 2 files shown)


msutkowski:
This looks to be a duplicate of https://github.com/remix-run/remix/pull/1449, which adds more context to the watch cmd. Would recommend closing in favor of that as it's from the CF team.

jerryharrison:
Yep they added the same ref change and the --watch attribute! I didn't see that PR from 5 days ago. :)

I'll close.

PR #1495, opened 2022-01-14
docs(examples): Add Stitches example

helderburato:


examples/stitches/app/root.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+        {styles}
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 33:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

examples/stitches/app/root.tsx
helderburato: Thanks! :D

danestves:
CleanShot 2022-01-29 at 20 12 20

I'm getting these errors if I want to use this example

kentcdodds (changes requested):
Thanks for this too! Could you address @danestves's concern there and also the same comments I left on [the emotion example](https://github.com/remix-run/remix/pull/1485)?

> Could you add one or two routes to make sure that it works across those route boundaries? And double-check that `CatchBoundary` and `ErrorBoundary` on the root route will work because I'm not certain it will.

Thanks a ton!

helderburato:
After some research into the Stitches issues, I found two issues related to the Remix, you can read more in the following links:
https://github.com/modulz/stitches/issues/886
https://github.com/modulz/stitches/issues/908

I think, for now, Stitches has this unsolved issue related to Remix.

I read some code trying to implement Stitches + Remix, and I found that case I submitted to review, it looks weird for me, but I didn't find a proper way to solve it.

What do you think about this implementation? cc @kentcdodds

kentcdodds:
I'm guessing this will suffer from the same issue: https://github.com/remix-run/remix/pull/1512#issuecomment-1042413526

kentcdodds:
Looks like we've found a solution in #1485. Can we do the same thing here?

danestves:
I doubt that will apply to Stitches @kentcdodds because Stitches doesn't depend on `@emotion`

kentcdodds:
I was more thinking about the approach. 🤷‍♂️

danestves:
Yeah, I get it 😄 the sad thing about it is that SSR with stitches is with a function but nothing more https://stitches.dev/docs/server-side-rendering, so I guess that is something that Stitches must add

helderburato:
I agree with you @danestves, the approach applied to Emotion will not work for this case.

danestves:
Hi @eldarshamukhamedov today was launched a new version of Stitches https://github.com/modulz/stitches/releases/tag/v1.2.7

And I think that `Fix for #908: toString in hydrated group sheet` could solve the  problem using the `getCssText()` on server and client

danestves:
> Hey @danestves, I'm not familiar with Stitches, so I think you probably mis-tagged me 😅

Yeah, sorry hahaha was for @helderburato

examples/stitches/app/root.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  return (
+    <html lang="en">
+      <head>
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        {title ? <title>{title}</title> : null}
+        <Meta />
+        <Links />
+        <style
```
derekr, line 49: @helderburato should this be something like:

```jsx
dangerouslySetInnerHTML={{
						__html: clientStyleData.sheet || serverStyleData,
					}}
```

I noticed with the setup as is I was getting unstyled SSR on initial load, but this seems to do the trick.

PR #1506, opened 2022-01-15
docs(blog): Modify `routes/index.tsx` instead of `root.tsx`

Awjin:


ryanflorence:
changed the wording instead, thanks!

PR #1507, opened 2022-01-15
docs(cookies): update form data name in action function to match HTML form element

mattmazzola:
### Problem:

The route is rendering a form with the markup
```
<input
    type="hidden"
    name="bannerVisibility"
    value="hidden"
/>
```
Notice name: "bannerVisibility"

but the `action` function for this route is using the name "banner"

```
if (bodyParams.get("banner") === "hidden") {
```

this would return `null` and the `if` statement would never be true.

### Change:

`bodyParams.get("banner")`
`bodyParams.get("bannerVisibility")`

to match the component markup

### DIFF
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -1567,7 +1567,7 @@ export async function action({ request }) {
     (await userPrefs.parse(cookieHeader)) || {};
   const bodyParams = await request.formData();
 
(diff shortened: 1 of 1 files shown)


ryanflorence:
this is fixed on main, remember doc PRs go to `main`, source PRs go to `dev`, thanks!

PR #1512, opened 2022-01-15
docs(examplex): add Next UI example

mzaien:
Creating a simple example with Nextui a ui library that is so promising ( the name suggests it is for next.js but no it can be used in any react framework)


examples/next-ui/README.md
```diff
@@ -0,0 +1,22 @@
...
+## Preview
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+
+[![Open in CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/next-ui)
+
+## Example
+
+Describe the example and how it demonstrates solving the problem. Reference any relevant files/dependencies if needed.
```
kentcdodds, line 14: Could you fill this in with some notes about what this example is demonstrating?

examples/next-ui/app/entry.server.tsx
```diff
@@ -0,0 +1,23 @@
...
+  request: Request,
+  responseStatusCode: number,
+  responseHeaders: Headers,
+  remixContext: EntryContext
+) {
+  const markup = renderToString(
+    <NextUIProvider>
+    <RemixServer context={remixContext} url={request.url} />
+    </NextUIProvider>
+  ).replace(/<\/head>/, `<style id="stitches">${getCssText()}</style></head>`);
```
kentcdodds, line 15: Why is the `id` here "stitches"? Does nextui depend on stitches?

kentcdodds (changes requested):
Thanks for this!

Could you add one or two routes to make sure that it works across those route boundaries? And double-check that `CatchBoundary` and `ErrorBoundary` on the root route will work because I'm not certain it will.

Thanks a ton :)

examples/next-ui/app/entry.server.tsx
mzaien: Yes it does

examples/next-ui/README.md
mzaien: Sure 👍🏻

examples/next-ui/app/root.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <title>{title}</title>
+        <Links />
+      </head>
+      <body>
+        {children}
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 33:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

examples/next-ui/app/root.tsx
mzaien: Nice 👍, thank you

kentcdodds:
I just added a commit that adds an error boundary to demonstrate the problem with a root error boundary and CSS-in-JS libraries that mutate the `style`. Here's a demo of the issue:

https://user-images.githubusercontent.com/1500684/154375743-d90d9b66-7d2a-4f22-80b9-84ffd36f057a.mov

I think you could work around this with a pathless layout route: https://remix.run/docs/en/v1/api/conventions#pathless-layout-routes

Kinda frustrating, I know, but when your library is mutating elements in the `head` and we're rendering the `head` with React, and that gets unmounted, this is what happens.

CSS-in-JS libraries really should allow you to store the CSS stuff outside of React so you can access the same style sheet object in the App and ErrorBoundary components.

I'll leave it up to you for how you want to approach solving this problem.

chaance:
Since this lib is built on top of Stitches, won't we run into some of the same hydration issues as https://github.com/remix-run/remix/pull/1495?

mzaien:
> Since this lib is built on top of Stitches, won't we run into some of the same hydration issues as #1495?

Unfortunately we are running into the same issues

kentcdodds:
We've found a solution in #1485 for emotion. Maybe you can use the same approach?

PR #1513, opened 2022-01-15
docs(tutorials): fix link in your-first-route step

cam31007:
The Developer Blog tutorial mentions to add a `<Link />`  inside `app/root.tsx` and that could also delete the entire content of that file. I believe that is incorrect since modifying that file incorrectly could "break" the app. My best guess is the intention probably was to add the `<Link />`  inside `app/routes/index.tsx` instead.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -24,6 +24,7 @@
 - christianhg
 - christophgockel
 - clarkmitchell
+- cam31007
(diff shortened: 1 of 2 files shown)


ryanflorence:
we removed the stuff that says you can delete things

PR #1515, opened 2022-01-16
docs(tutorial): Resolve TS errors before updating UI

Awjin:


ryanflorence:
Thanks for the PR. I don't write code in this order but for a tutorial with both audiences (TS and JS users) it's deliberate. Make it work and then add the types after to not distract the non-typescript users.

PR #1517, opened 2022-01-16
docs(example): Fix examples typos

sergiodxa:
This fix two typos on examples I previously sent

1. In the client-only-component example the button type was `buttonn` instead of `button`
2. The folder of the CMS Collected Notes example was `csm-collected-notes` instead of `cms-collected-notes`, I renamed it to be just `collected-notes` as with Strap and Sanity CMS examples.

Idea: Prefix all CMS examples with `cms-` so they are all together, I can do this change if desired.


kentcdodds:
I thought about prefixing things with the category (like `cms-`), but the problem is that some technologies could have multiple categories (I've been corrected when I referred to sanity as a "cms" 🤷‍♂️). People don't come to the examples to compare CMS-es I think. They normally have a CMS they've decided on and want to see how to integrate or some others they have in mind and want to compare them.

Thanks for these fixes!

PR #1519, opened 2022-01-16
docs(example): Add io-ts FormData decoding example

hollandThomas:


kentcdodds (approved):
This looks perfect. Thanks a bunch!

PR #1238, opened 2021-12-25
fix: blog tutorial post link

arrudaricardo:
The Blog tutorial is ask to add Link Component on app/root.tsx but it should be at app/routes/index.tsx

PR #1521, opened 2022-01-16
docs(tutorial): The first Link should be in `app/routes/index.tsx`

etienne-dldc:
Hi,

The tutorial is telling to add a link in `app/root.tsx` and says that "you might want to just delete everything that's there" but that does not feel right...
My guess is that it should be `app/routes/index.tsx` instead.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -42,6 +42,7 @@
 - edgesoft
 - emzoumpo
 - eps1lon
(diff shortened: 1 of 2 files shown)


etienne-dldc:
Whoops,

This is already fixed by at least two open PRs (#1513, #1238), closing this one.

PR #1527, opened 2022-01-17
feat(cli): exit with code 1 when remix build fails

gmaliar:
When builds fail remix does not exit with an error code which makes it hard to catch build failures in CI environments

Example:
```sh
❯ yarn create remix
❯ cd ./my-remix-app
❯ echo -e "import doesntExists from \"doesnt-exists\";\ndoesntExists();\n$(cat app/routes/index.tsx)" > app/routes/index.tsx
❯ npx remix build

npx remix build                                                                                                           
The path "doesnt-exists" is imported in app/routes/index.tsx but doesnt-exists is not listed in your package.json dependencies. Did you forget to install it?
 > node_modules/@web-std/file/tsconfig.json:6:24: warning: Non-relative path "packages/file/src/lib.js" is not allowed when "baseUrl" is not set (did you forget a leading "./"?)
    6 │       "@web-std/file": ["packages/file/src/lib.js"]
      ╵                         ~~~~~~~~~~~~~~~~~~~~~~~~~~


 > app/routes/index.tsx:1:25: error: Could not resolve "doesnt-exists" (mark it as external to exclude it from the bundle)
    1 │ import doesntExists from "doesnt-exists";
      ╵                          ~~~~~~~~~~~~~~~


Build failed with 1 error:
app/routes/index.tsx:1:25: error: Could not resolve "doesnt-exists" (mark it as external to exclude it from the bundle)
Built in 571ms

❯ echo $?
0
```

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,20 @@
 # yarn lockfile v1
 
 
+"@architect/functions@^4.1.1":
+  version "4.1.1"
+  resolved "https://registry.yarnpkg.com/@architect/functions/-/functions-4.1.1.tgz#4520f6070858f8731c1da403e40591f3909fcfec"
+  integrity sha512-x0+B/V9Jo5onksOce6iYdYeLh5F91dOp5fTnGJdkgOFCS+N0YN9zlI2asdPIs1Rb6xxCwoH1nuzs/jnNkupQWA==
+  dependencies:
+    aws-serverless-express "^3.4.0"
+    cookie "^0.4.1"
+    cookie-signature "^1.1.0"
+    csrf "^3.1.0"
+    node-webtokens "^1.0.4"
+    run-parallel "^1.2.0"
+    run-waterfall "^1.1.7"
+    uid-safe "^2.1.5"
+
 "@architect/parser@5.0.2":
   version "5.0.2"
   resolved "https://registry.npmjs.org/@architect/parser/-/parser-5.0.2.tgz"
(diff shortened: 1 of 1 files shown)


gmaliar:
@kentcdodds can I get some feedback on this one as well? 🙏🙏🙏

contributors.yml
```diff
@@ -159,3 +159,8 @@
 - yomeshgupta
 - zachdtaylor
 - zainfathoni
+- dima-takoy
+- anishpras
+- kumard3
+- mattmazzola
+- gmaliar
```
kentcdodds, line 166: Should all of these people be added? Seems like you should only add your own username.

packages/remix-dev/cli/commands.ts
```diff
@@ -60,7 +60,11 @@ export async function build(
 
   let start = Date.now();
   let config = await readConfig(remixRoot);
-  await compiler.build(config, { mode: mode, sourcemap });
+  try {
+    await compiler.build(config, { mode: mode, sourcemap });
+  } catch (error: unknown) {
```
kentcdodds, line 65: I was under the impression that a runtime error would result in a non-zero exit code.

kentcdodds (commented):
Just left some thoughts. I don't merge changes to the compiler. So we really should have @mjackson and/or @jacob-ebey look at this

packages/remix-dev/cli/commands.ts
gmaliar: They get swallowed here https://github.com/remix-run/remix/blob/197f570841e4813b2005d1ae007523c09df3f5c5/packages/remix-dev/compiler.ts#L307-L314

contributors.yml
gmaliar: You are correct, my bad!

gmaliar:
@kentcdodds I accidentally deleted this branch, I made the amendments and will wait for @mjackson and/or @jacob-ebey comments here https://github.com/remix-run/remix/pull/1822

Thanks for the quick response!

PR #1531, opened 2022-01-18
Optimistic UI guide: update example file extension

waleed-alfaifi:
There were some examples using Typescript but with file extension set to `.js`. 

- Updated `new.js` to `new.tsx` for all examples
- Imported missing type `ActionFunction ` in `new.tsx`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -33,6 +33,7 @@
 - DavidHollins6
 - derekr
 - developit
+- waleed-alfaifi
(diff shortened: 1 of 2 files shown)


ryanflorence:
Looks like we fixed this in a later PR, sorry about that 😔. Took us a bit to get back into the PRs at the start of the year. Thanks for taking the time to make the Remix docs great 😁

PR #1533, opened 2022-01-18
docs(tutorial): Sanitize parsing of markdown in blog tutorial to remove potential XSS vulnerability

wdoug:
Obviously, the blog tutorial is an introductory example of some of the capabilities of Remix and not a secure website. That said, I think we should try to reduce the risks of people having security vulnerabilities in their code by copying documentation. I believe that we should encourage people to pass that `sanitize` parameter by default. If someone fully trusts the markdown source and also wants HTML capabilities beyond what markdown offers directly, it is easy enough for them to remove, but at least that way they would be more inclined to understand the ramifications.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -147,6 +147,7 @@
 - UsamaHameed
 - veritem
 - VictorPeralta
+- wdoug
 - weavdale
 - yauri-io
 - yesmeck
--- a/docs/tutorials/blog.md
+++ b/docs/tutorials/blog.md
@@ -540,7 +540,7 @@ export async function getPost(slug: string) {
     isValidPostAttributes(attributes),
     `Post ${filepath} is missing attributes`
   );
-  const html = marked(body);
+  const html = marked(body, { sanitize: true });
   return { slug, html, title: attributes.title };
 }
 ```


hollandThomas:
I think that's a good idea!

However, the [official marked docs](https://marked.js.org/using_advanced#options) flag `sanitize` as deprecated:

> If true, sanitize the HTML passed into markdownString with the sanitizer function.
Warning: This feature is deprecated and it should NOT be used as it cannot be considered secure.
Instead use a sanitize library, like [DOMPurify](https://github.com/cure53/DOMPurify) (recommended), [sanitize-html](https://github.com/apostrophecms/sanitize-html) or [insane](https://github.com/bevacqua/insane) on the output HTML!

The docs should not recommend the usage of deprecated options either I think. It's probably still a good idea to either use or point to something like DOMPurify, even if that complicates the tutorial a bit.

Whatever we land on, it'd be great to update the [strapi example](https://github.com/remix-run/remix/blob/main/examples/strapi/app/routes/index.tsx#L37) accordingly, for consistency (it uses marked as well).

wdoug:
Ooh, good catch! Hmm

ryanflorence:
If you're gonna XSS yourself on your own blog you can just write the script you wanted :P

Agree we should help folks get exposed to these kinds of things. You want to make a different PR that says something in the tutorial like this:

> If we were accepting markdown from users (and not ourselves here) we'd want to sanitize this HTML with something like [insert library][link]. We don't have to worry about it here since we aren't accepting data from users, it's our own blog!

PR #1535, opened 2022-01-18
docs(examples): Add DataLoader example

hollandThomas:


sergiodxa:
Shouldn’t data loaders be created on every request? To avoid sharing cached data between users.

hollandThomas:
Yes, that's correct. It's entirely possible I'm missing something, conceptually 😅  But isn't that the case here?

sergiodxa:
> Yes, that's correct. It's entirely possible I'm missing something, conceptually 😅 But isn't that the case here?

In `examples/dataloader/app/loaders/userLoader.server.ts` you are creating the new DataLoader instance, so the instance is created once for the whole app, not on every request.

I think to use data loaders you need to create them on the server, and pass them through context, they way if a request run more than one loader (first request) it will use the same data loader, for client-side navigations since they are individual requests it will not share them which I think it's fine.

hollandThomas:
Hmm. Now I'm confused. I just moved the example to the Express version of Remix and put the loader on context:

```typescript
createRequestHandler({
  build,
  mode: MODE,
  getLoadContext: () => ({
    loaders: {
      usersById
    }
  })
})
```

Interestingly, the server _now_ caches across requests, calling `db.user.findMany` only once on the initial request:

```
Express server listening on port 3000
user#findMany
GET /users 200 - - 70.793 ms
GET /users 200 - - 19.482 ms
GET /users 200 - - 18.388 ms
GET /users 200 - - 15.159 ms
GET /users 200 - - 20.222 ms
GET /users 200 - - 16.582 ms
```

Whereas with the current implementation in this PR, the `db.user.findMany` is called on each request:

```
Remix App Server started at http://localhost:3000
user#findMany
GET /users 200 - - 51.325 ms
user#findMany
GET /users 200 - - 10.255 ms
user#findMany
GET /users 200 - - 8.496 ms
user#findMany
GET /users 200 - - 7.681 ms
user#findMany
GET /users 200 - - 6.759 ms
user#findMany
GET /users 200 - - 14.401 ms
```

sergiodxa-silverback:
@hollandThomas try in production mode, the Remix App Server is clearing the require cache on every request, so the files are required again making a new instance of the data loader.

sergiodxa:
@hollandThomas try in production mode, the Remix App Server is clearing the require cache on every request, so the files are required again making a new instance of the data loader.

hollandThomas:
Ah, my bad. It needs to be a factory that creates the loader of course

```typescript
export const createUsersByIdLoader = () =>
  new DataLoader(async (keys: Readonly<string[]>) =>
    db.user.findMany({
      where: {
        id: {
          in: keys
        }
      }
    })
  );
```

```typescript
createRequestHandler({
  build,
  mode: MODE,
  getLoadContext: () => ({
    loaders: {
      usersById: createUsersByIdLoader()
    }
  })
})
```

I will clean up, then push the updated version

examples/dataloader/app/root.tsx
```diff
@@ -0,0 +1,27 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 23:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

examples/dataloader/README.md
```diff
@@ -0,0 +1,26 @@
+# DataLoader
+
+This example shows how to avoid redundant roundtrips to data sources via batching and caching, implemented with [DataLoader](https://github.com/graphql/dataloader). This comes in handy when there are multiple nested routes that depend on the same data source.
+
+Note that in many cases, you can meet the same requirement with Remix's own [useMatches](https://remix.run/docs/en/v1/api/remix#usematches) which lets you use all of your parent-routes' data in child routes. If that suits your use case, using `useMatches` is preferable. There are cases where this might not be the best fit though, including:
+
+- You can't be sure if a parent did already request a particular piece of data
+- You don't want your data loading to be tightly coupled with your parent routes, for architectural reasons
```
kentcdodds, line 8:
```suggestion
- You don't want your data loading to be tightly coupled with your parent routes, for architectural reasons
- You need access to the data within the `loader`, not just the component
```

kentcdodds (approved):
Thanks everyone for your help on this. @hollandThomas you did a great job of making this practical and also focused. Well done. Thank you for your work on this!

tshddx:
To be clear, this only works for an SSR request to `/users`, right? If we did a client-side navigation to a URL that caused 2 new nested routes to load, the client will do 1 request for each nested route, and thus no batching will occur across those 2 requests. Is my understanding correct?

PR #936, opened 2021-12-07
fix: Numerical URL targets cause site hang

chenc041:
if location hash is start of numerical, it will cause useScrollRestoration hook throw an error and infinity loop 

change file  [package/remix-react/scroll-restoration.tsx](https://github.com/remix-run/remix/blob/main/packages/remix-react/scroll-restoration.tsx#L101)

issue num [#935](https://github.com/remix-run/remix/issues/935)

reference [Rule sets, declaration blocks, and selectors](https://www.w3.org/TR/CSS2/syndata.html#characters)

PR #1542, opened 2022-01-19
fix: DOMException in scroll restoration

N1ck:
## Issue

When trying to scroll to a hash, if the provided hash is not a valid selector it currently throws a DOMException.

I noticed this when viewing the blog https://remix.run/blog/remix-vs-next and clicking on a header with an emoji: "Remix ❤️ the Web"  as it has a `href` of `#remix-%EF%B8%8F-the-web`

https://user-images.githubusercontent.com/1018141/150040187-5ada8897-e37b-4b84-bc18-476ec3cc5e0d.mov

## Fix

I've attempted to fix this by wrapping the call to `document.querySelector` with a try/catch block to avoid the exception and log the error. If there is a different, preferred approach to resolve this let me know.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -64,6 +64,7 @@
 - morinokami
 - mskoroglu
 - msutkowski
+- n1ck
 - nareshbhatia
 - niconiahi
 - nobeeakon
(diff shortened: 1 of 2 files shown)


chenc041:
Has pr already [#936](https://github.com/remix-run/remix/pull/936)

ryanflorence:
fixed in #936 thanks!

Issue #1549, opened 2022-01-19
[Bug]: Pathless layout route causes invalid conflict


Ehesp:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [X] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.1.1

### What version of Node are you using? Minimum supported version is 14.

16.1.0

### Steps to Reproduce

Create a route structure like so:

```
routes/__foo/index.tsx
routes/__bar/bar.tsx
routes/__foo.tsx
routes/__bar.tsx
```

### Expected Behavior

That the`/` route goes via `__foo/index.tsx` & the `/bar` route goes via `__bar/bar.tsx`.

### Actual Behavior

The error is thrown when starting remix:

```
Path undefined defined by route "routes/__bar" conflicts with route "routes/__foo"
```

[You can see a reproduction here](https://codesandbox.io/s/reverent-kepler-89csf).

At the moment, if I edit the [check here](https://github.com/remix-run/remix/blob/main/packages/remix-dev/config/routesConvention.ts#L73) I can get it to work.

```diff
- if (typeof uniqueRouteId !== "undefined") {
+ if(uniqueRouteId)
```

The `uniqueRouteId` variable will never be undefined, since it's concatenated - this allows invalid checking of the unique route ids for the layout route.

```js
let uniqueRouteId = (fullPath || "") + (isIndexRoute ? "?index" : "");
```

elkevinwolf:
Came here for the same issue. Apparently it got fixed in `v1.1.2` [here](https://github.com/remix-run/remix/pull/1347)

Ehesp:
Looks like it - will close.

PR #1550, opened 2022-01-19
fix: fix invalid checking of layout route id

Ehesp:
Fixes https://github.com/remix-run/remix/issues/1549

Layout routes (e.g. `__foo.tsx`) are getting treated as a route because the typeof check will never be undefined (since strings are concatenated). This changes the logic to only check for path uniqueness if there is a route id with a string length greater than 0.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -153,3 +153,4 @@
 - yomeshgupta
 - zachdtaylor
 - zainfathoni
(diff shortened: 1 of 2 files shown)


Ehesp:
Fixed via https://github.com/remix-run/remix/pull/1347

Issue #1552, opened 2022-01-19
[Bug]: Parent headers are not loaded if child does not set the headers


dg1234uk:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [X] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.1.1

### What version of Node are you using? Minimum supported version is 14.

17.3.0

### Steps to Reproduce

Set the following on a parent route:

```ts
export function headers() {
  return {
    test: "value"
  };
}
```

Then without any headers set on a child route. Go to the child route and note that the `test: "value"` header is not set.

### Expected Behavior

I have taken this from the docs: [Remix.run Docs - headers](https://remix.run/docs/en/v1/api/conventions#headers)

> Because Remix has nested routes, there's a battle of the headers to be won when nested routes match. In this case, the deepest route wins.

I'm not sure if my interoperation is correct but I believe it is intended that if there are no headers set on the child route/loader/action then it will use the parents (if set).

### Actual Behavior

For the `test: "value"` header to be set on all child routes unless they define their own headers, then it is up to those child routes to merge the parent headers if they require.

dg1234uk:
#1553 is my attempt to resolve this

dg1234uk:
Also the remix-cla-bot did not prompt me to sign the CLA as per the [Contributing to Remix](https://remix.run/docs/en/v1/pages/contributing#contributor-license-agreement) docs mention.

jacob-ebey:
This is actually expected behavior. Headers within nested routes are not as simple as "just surface the parent's if I don't have my own".

Think of this case where your routes folder has this structure:

```
routes
  | - dashboard.tsx
  | - dashboard
    | - item.tsx
```

Visiting /dashboard/item will render the nested routes. If dashboard.tsx returns headers that say "cache me for 1 hour", but the item needs to *not* be cached at all, you have just shot yourself in the foot if parent headers surface automatically and have accidentally cached your item.tsx for 1 hour as well.

[...]

PR #1553, opened 2022-01-19
Fix load parent headers

dg1234uk:
Fixes: #1552


dg1234uk:
Deleted due to the git commit history, created new pull request #1556 with cleaner history

PR #1556, opened 2022-01-19
Fix load parent headers

### LINKED ISSUE #1552: [Bug]: Parent headers are not loaded if child does not set the headers

dg1234uk:
Fixes #1552

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -17,6 +17,7 @@
 - derekr
 - donavon
 - developit
+- dg1234uk
 - dunglas
 - dwt47
 - edgesoft
--- a/packages/remix-server-runtime/headers.ts
+++ b/packages/remix-server-runtime/headers.ts
@@ -21,7 +21,7 @@ export function getDocumentHeaders(
         ? typeof routeModule.headers === "function"
           ? routeModule.headers({ loaderHeaders, parentHeaders, actionHeaders })
           : routeModule.headers
-        : undefined
+        : parentHeaders
     );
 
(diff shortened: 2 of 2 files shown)


jacob-ebey:
Please see comment on #1552. Maybe we need better documentation around this, but this is intentional behavior and modifying it to automatically surface parent headers goes from a footgun to a footbazooka.

PR #1558, opened 2022-01-19
docs(tutorial): refer to the right file to add a link

saharbit:
I think the docs are somewhat misleading, because they tell you to go to `root.tsx` and then 

_"You can put it anywhere you like, you might want to just delete everything that's there."_

Obviously we don't want people removing everything in `root.tsx` (Scripts, LiveReload etc.) so it's better to edit `routes/index.tsx`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -127,6 +127,7 @@
 - RossMcMillan92
 - rphlmr
 - ryanflorence
+- saharbit
 - sean-roberts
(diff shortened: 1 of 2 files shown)


ryanflorence:
thanks, the tutorial changed a bit when we simplified the CLI template generator, I'm going to remove the line about removing everything instead. Thanks!

PR #1561, opened 2022-01-19
fix(ScrollRestoration): errors with invalid selector

Commandtechno:
With an invalid selector `document.querySelector` throws a `SyntaxError` which crashes the website
Example of an invalid selector that throws a `SyntaxError` is an ID that starts with a number (#1)

Stackoverflow Reference: https://stackoverflow.com/a/34777644
MDN Docs:
https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#parameters

To reproduce go to https://remix.run/ and then change the URL to https://remix.run/#1 for ScrollRestoration to activate
I'm guessing the remix website just trys to re-render on an error so it leads to an error loop in the console

image

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -11,6 +11,7 @@
 - bsharrow
 - chaance
 - christophgockel
+- Commandtechno
 - coryhouse
 - donavon
 - davecalnan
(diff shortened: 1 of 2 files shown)


chenc041:
Duplicate pr [#936](https://github.com/remix-run/remix/pull/936)

Commandtechno:
yeah that looks better because it allows for ones that start with numbers

PR #1572, opened 2022-01-20
Docs: Remove merge conflict artefact

mattstobbs:
I noticed there was some left overs from a merge conflict in the docs. I took a guess at which one was the correct one but let me know if I should use `HEAD` instead and I'll update the PR


mcansh:
thanks! I missed this PR, but noticed we've accumulated a few in a couple other places
fixed by https://github.com/remix-run/remix/commit/d99e2b22d2859876174505fd2433d9b014f348af

PR #1586, opened 2022-01-20
Enhancement: Implemented a node minimum version check

damiensedgwick:
### What is the feature?
This prevents users from running the create-remix command when they do not have the minimum version of Node required for Remix.

### What is the solution?
At the top of `createApp`, I added a  simple check to grab the users node version and if it is below v14, it will log to the console and prevent further execution of the script.

I have attached a screen shot of a failed attempt and a succeeding attempt using 2 different node versions.

Screenshot 2022-01-20 at 19 32 11

### DIFF
--- a/packages/create-remix/index.ts
+++ b/packages/create-remix/index.ts
@@ -24,6 +24,17 @@ interface CreateAppArgs {
 }
 
 async function createApp({ projectDir, lang, server, install }: CreateAppArgs) {
+  // Check the host's node version
+  const versions = process.versions;
+  if (versions && versions.node) {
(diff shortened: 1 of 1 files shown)


sergiodxa (approved):

damiensedgwick:
I've just noticed I have opened up the PR to main rather than dev. Please let me know if you would like this amended.

ryanflorence:
You mind fixing the conflicts and we'll get this merged (sorry, our fault for taking so long 😔)

damiensedgwick:
Should be all good now @ryanflorence

PR #1595, opened 2022-01-21
Update contributors.yml

sndrem:


ryanflorence:
I think we've already got you in the dev branch. thanks!

PR #1606, opened 2022-01-22
feat(remix-server-runtime): add `loadContext` to `handleDocumentRequest`'s signature

jacob-ebey:


mjackson:
Can you describe a little about the use case you have for this?

jacob-ebey:
General availability of the `({ request, context })` "context" value for access to things that are injected. Potentially a Redis client for loading things from a cache, etc.

Once use case I particularly have is around streaming and gaining access to the adapters response object. You can check it out here: https://stackblitz.com/edit/github-3vromp?file=app%2Fentry.server.tsx

jacob-ebey:
Use-case: Relay info from loaders back to entry.server.
Example: Calculate a header in the root loader that should be added to every response without having to export a headers function from every route:

```ts
// root.tsx
async function loader({ context }) {
  let response = await fetch("some-third-party");
  context.headers.append("X-Custom", response.headers.get("X-Custom"));
}
```

```ts
// entry.server.ts;
import { renderToString } from "react-dom/server";

  request: Request,
  responseStatusCode: number,
  responseHeaders: Headers,
  remixContext: EntryContext,
  loadContext: { headers: Headers }
) {
  let markup = renderToString(
  );

  responseHeaders.set("X-Custom", loadContext.headers.get("X-Custom"));

  return new Response("<!DOCTYPE html>" + markup, {
    status: responseStatusCode,
    headers: responseHeaders,
  });
}
```

PR #1611, opened 2022-01-22
chore: null check on remix-serve

elifnurkarakoc:
null check on remix-serve


packages/remix-serve/cli.ts
```diff
@@ -18,7 +18,7 @@ let buildPath = path.resolve(process.cwd(), buildPathArg);
 createApp(buildPath).listen(port, () => {
   let address = Object.values(os.networkInterfaces())
     .flat()
-    .find(ip => ip?.family == "IPv4" && !ip.internal)?.address;
```
ryanflorence, line 21: does the first `ip?.family` check not protect the second?

elifnurkarakoc: You're right. thanks 🙂

PR #1617, opened 2022-01-22
solve getPosts for vercel users

Adbib:
if you deploy old code to Vercel you cant see the index if posts with this update it can show index blog posts easily

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -33,6 +33,7 @@
 - DavidHollins6
 - derekr
 - developit
+- Adbib
 - dhargitai
 - dhmacs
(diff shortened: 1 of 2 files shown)


DennisKo:
I would think you have to adjust `getPost` as well

Adbib:
> I would think you have to adjust `getPost` as well

I don't think you have to
its work fine with the old one

ryanflorence:
Thanks for helping :) But the tutorial is clear that the file system is not the right approach for persistence in a production setting. The jokes tutorial uses persistence that works in production.

PR #1623, opened 2022-01-23
fix the link to the NavLinks component

theMosaad:


ryanflorence:
got this elsewhere, thanks

PR #1626, opened 2022-01-23
Apollo Client (example)

visormatt:
# Overview

This PR adds a relatively simple example using [Apollo Client](https://www.apollographql.com/docs/react/) and a readily available [GraphQL API](https://rickandmortyapi.com/graphql) working with Remix. Additionally I've opted to add in a codegen setup which automates the creation of types and react hooks. 

There are two routes being used, each making a request for some data. The `index` route makes use of Remix Loaders whereas the details view `$id` uses react hooks to achieve the same. 

Thanks to everyone involved in this fantastic project and LMK how I can improve on this PR.


ryanflorence:
I'm not sure what use cases Apollo client fills in the context of Remix.

- On the initial SSR, `getDataFromTree` wipes out all the benefits of Remix's SSR approach.
- You can't fetch all data in parallel on the initial SSR
- You can't fetch all data/modules/css in parallel on client side transitions
- You can't prefetch pages the user is about to navigate to
- It ships a lot more JS to the client
- You can't model mutations in the UI as simple forms

I'd be interested to see this example but remove Apollo client and just fetch from the graphql API from a loader and then compare the performance characteristics.

Perhaps this is most useful as an example of migrating to Remix from Apollo Client? I don't know, would love to hear your thoughts.

visormatt:
@ryanflorence Thanks for the initial feedback and questions!

Yes this example was absolutely meant as more of a `migration path` from Apollo Client to **Remix**. I'd imagine there are a lot of orgs that have tons of logic wrapped up in Apollo already and this would help to minimize the LOE to port things over to Remix. However I also understand that in doing so we're making other trade offs. 

So, hows about I take a crack at a version that's free of Apollo Client all together 💯. I've create a separate [PR here](https://github.com/remix-run/remix/pull/1754) which strips out all those bits noted and makes light use of the `codegen` to help automate the types which can then be used in our `useLoaderData` hook. 

I'd also love to iterate on this one as a clear `migration path` example and hopefully better callout the trade-offs if that's something you think _(with a little polish)_ would be useful.

sergiodxa:
If this PR is intended to be a migration step from React + Apollo to Remix, maybe adding a note in the README explaining that and suggesting next steps on the migration would be useful

ryanflorence:
Yeah, a migration path in the README with detailed instructions on how to iterate toward *not shipping the graphql client anymore* would be awesome :) And then it can link to the other example of just using graphql in loaders/actions. Perhaps even a mention of how much less JS is shipped and how the UX is improved with the initial render + prefetching? I don't know, getting carried away now 😋

The apollo "recursive render" and server-side render+fetch chains that creates is just not something I want anybody to be doing in a Remix app 🥲 So the README needs to be very clear that this is about migration 😇

examples/apollo-client/app/root.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 36:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

kentcdodds (changes requested):
Thanks for working on this! Could you address Ryan's feedback please?

visormatt:
@ryanflorence and @kentcdodds, thank you both for the feedback and sorry for the slow response. I was hoping to work through the smaller version (non Apollo) [version here](https://github.com/remix-run/remix/pull/1754) before tackling the updates in here. The code in that PR is potentially a last phase of this migration PR. Here we started with Apollo and by the end of it Apollo is stripped out in favor of Remix + GraphQL (existing server). 

Happy to close it out and re-open shortly if that makes life any easier on your side of things.

kentcdodds:
No rush. I think having an example of Remix working with Apollo (with the note about migration) is still valuable. Take your time :)

kentcdodds:
I will close this for now though. When you're ready to start working on it again we can definitely open it up again :)

marwan38:
> I'm not sure what use cases Apollo client fills in the context of Remix.
> 
> * On the initial SSR, `getDataFromTree` wipes out all the benefits of Remix's SSR approach.
> * You can't fetch all data in parallel on the initial SSR
> * You can't fetch all data/modules/css in parallel on client side transitions
> * You can't prefetch pages the user is about to navigate to
> * It ships a lot more JS to the client
> * You can't model mutations in the UI as simple forms
> 
> I'd be interested to see this example but remove Apollo client and just fetch from the graphql API from a loader and then compare the performance characteristics.
> 
> Perhaps this is most useful as an example of migrating to Remix from Apollo Client? I don't know, would love to hear your thoughts.

Hi Ryan, I feel like i'm stepping out of line by butting in on something that has nothing to do with me, but, I really wanted to understand more on some of the points you've mentioned here. At my company, we're about to adopt Remix for a big re-design of the application (for various reasons). We are heavily reliant on apollo and our mindset and paradigm are very different from what Remix is introducing, so, It will be a slow and potentially a not so straightforward path.

I'm having difficulties understanding the first 2 points you've mentioned, and both of which, I need to fully understand as we embark on to adopting Remix. Point 3 onwards, I think, is clear to why these are issues (it's pretty much the cons of client side logic). The first two points I can't wrap my head around because of my lack of knowledge here. 

1. How does `On the initial SSR, `getDataFromTree` wipes out all the benefits of Remix's SSR approach.`? What approach are you talking about and what are the benefits being wiped out, and why?
2. Why can't the data be fetched in parallel? Whether it is apollo or just a regular fetch being awaited, in the end it's just a promise waiting to be resolved, no? What am I missing here?

CC @kentcdodds since you've been assigned here. I hope my tags don't add any weight to your work and that this post (and hopefully the reply) will be beneficial to others who might be in a similar situation to us.

kentcdodds:
@marwan38, could you post this as a GitHub discussion instead? Old closed PRs are tough for the community to notice and help on and Ryan and I don't always see GitHub notifications.

PR #1629, opened 2022-01-23
docs: update installation command for tailwindcss guide

xstevenyung:
Since v3, Tailwind.css requires PostCSS to run the CLI and recommend `autoprefixer` as a `peerDependency`.

I updated the styling guide to include those update


ryanflorence:
thanks! but we followed the official instructions from tailwind: https://tailwindcss.com/docs/installation

PR #1634, opened 2022-01-24
docs: Adds example of Prisma generated type

guerra08:
This PR adds an example of **Prisma Client generated types** when using a **LoaderFunction**. This is similar to other samples found on the example projects and video guides on Remix YouTube channel.

image

### DIFF
--- a/docs/guides/data-loading.md
+++ b/docs/guides/data-loading.md
@@ -161,16 +161,17 @@ export default function ProductCategory() {
 }
 ```
 
-If you are using TypeScript, you can use Prisma Client generated types to get better type safety and intellisense when writing your code to load data.
+If you are using TypeScript, you can use type inference to use Primsa Client generated types on when calling `useLoaderData`. This allowes better type safety and intellisense when writing your code that uses the loaded data.
 
 ```tsx filename=tsx filename=app/routes/products/$productId.tsx
 import { useLoaderData } from "remix";
-import type { LoaderFunction } from "remix";
-import type { Product } from "@prisma/client";
+import { db } from "~/db.server";
 
-type LoaderData = Product;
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
You actually don't need these extra types

https://twitter.com/millsp_/status/1485931866586857479

guerra08:
> You actually don't need these extra types
> 
> https://twitter.com/millsp_/status/1485931866586857479

Oh, nice! I have update the PR with this suggestion.

ryanflorence:
@kentcdodds please review and merge when you're ready :)

docs/guides/data-loading.md
```diff
@@ -161,6 +161,30 @@ export default function ProductCategory() {
...
+
+```tsx filename=tsx filename=app/routes/products/$productId.tsx
+import { useLoaderData } from "remix";
+import { db } from "~/db.server";
+
+type LoaderData = Awaited<ReturnType<typeof loader>>
+
+export let loader = async ({ params }) => {
+  return db.product.findMany({});
+};
```
kentcdodds, line 174: The challenge here is that if you want to set headers you can no longer use the return value as it's a Response object instead. So how about this:

```suggestion
import { useLoaderData, json } from "remix";
import { db } from "~/db.server";

type LoaderData = Awaited<ReturnType<typeof getLoaderData>>

async function getLoaderData() {
  const products = await db.product.findMany({
    select: {
      id: true,
      name: true,
      imgSrc: true,
    }
  })
  return { products }
}

export let loader = async () => {
  return json<LoaderData>(await getLoaderData());
};
```

kentcdodds (approved):
Thanks for this! I'm going to make a small change here, but it's great 👍

guerra08:
Thanks guys, happy to contribute!

PR #1636, opened 2022-01-24
(chore) - create Preact example

JoviDeCroock:
This PR makes a version of the `examples/basic` app leveraging Preact. The only thing that stands out here is the "hack" we have to do in Preact to make the `hydrate(x, document)` work correctly.

There's an additional fix to be found in the `remix-dev` CLI, when missing a package like `React` (or in my case Preact... Yes I forgot to install Preact :joy:) it would print out that the dev cli was missing, which had me confused for quite a while. Now it will take the first line of the error message, which is the same it was before, and leverage that to indicate the missing module.


examples/basic-preact/app/entry.client.tsx
```diff
@@ -0,0 +1,12 @@
+import { hydrate } from "react-dom";
+import { RemixBrowser } from "remix";
+
+const documentElement = document.documentElement;
+const apply = (n: HTMLElement) => document.replaceChild(n, documentElement);
+// Temp fix
```
JoviDeCroock, line 6: Preact currently fails when passing in `document` as that has a few nodes that aren't reconcilable 😅

phiberber:
Does this also apply Preact for production? How are you checking for Preact in the website?

JoviDeCroock:
The npm deps replace React with Preact, verified production bundles :)

examples/basic-preact/app/root.tsx
```diff
@@ -0,0 +1,248 @@
...
+        {title ? <title>{title}</title> : null}
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        {children}
+        <RouteChangeAnnouncement />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" ? <LiveReload /> : null}
```
MichaelDeBoey, line 75:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

ryanflorence:
We've been working (sporadically, our fault) with @developit on this. I'm going to close this but welcome it being reopened when we announce official Preact support :)

ryanflorence:
To be clear, we will support Preact "natively". No compat layer so we can get the client bundle even smaller :)

PR #1504, opened 2022-01-15
feat: Allow for custom server to participate in bundling

jacob-ebey:
feat: introduces virtual module `@remix-run/assets-manifest` that contains the client asset manifest
We previously wrote out a json file in the server build directory outside of the bundler. this introduced the requirement of another external bundle step for any deployment platform that requires a single file.

feat: introduces virtual module `@remix-run/server-build` that is the module that is currently output as the server bundle
We previously piped this into esbuild stdin, but with the introduction of `customServer` being able to participate in bundling we needed a way to access the "finalized" remix server build before the build finishes 🤯

feat: introduces `serverBuildTarget` config option and deprecates `serverModuleFormat` and `serverPlatform`
`serverBuildTarget` now configures defaults in the compiler for you that look like this:

| serverBuildTarget  | serverModuleFormat | serverPlatform | node_modules | serverBuildPath | publicPath |
|--------------------|--------------------|----------------|--------------|--------------|--------------| 
| node-cjs (default)           | cjs                | node           | external     | build/index.js | /build/ |
| arc                | cjs                | node           | external     | server/index.js | /_static/build/ |
| netlify            | cjs                | node           | external     | netlify/functions/server/index.js | /build/ |
| vercel             | cjs                | node           | external     | api/_build/index.js | /build/ |
| cloudflare-pages   | esm                | neutral        | bundled      | functions/[[path]].js | /build/ |
| cloudflare-workers | esm                | neutral        | bundled      | build/index.js | /build/ |

If a `serverBuildTarget` is set besides `node-cjs` and a `customServer` is not provided, the output module for the server build will now be a "ready to go" file for that build target. If `serverBuildTarget` is not defined or `node-cjs` the output module is the same format as today.

feat: explicitly exclude `@remix-run/server-runtime`, `@remix-run/node`, `@remix-run/cloudflare-pages`, `@remix-run/cloudflare-workers`, `@remix-run/deno` from client bundles. This may introduce new runtime errors in existing projects that are currently bundling server modules in the client by accident. Using `xxx.server.js` pattern will fix these issues for people who run into issues.

PR #1643, opened 2022-01-25
getLoadContext param for arc's createRequestHandler optional

lpsinger:
The `getLoadContext` parameter for remix-architect's `createRequestHandler` function should be optional.

According to https://remix.run/docs/en/v1/api/conventions#loader-context:

> This API is an escape hatch, it’s uncommon to need it

And indeed other Remix hosting targets (`remix-cloudflare-pages`, `remix-cloudflare-workers`, `remix-express`, `remix-netlify`, `remix-vercel` declare this parameter as optional.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -92,6 +92,7 @@
 - leo
 - leon
 - levippaul
+- lpsinger
 - lswest
 - luispagarcia
 - luistorres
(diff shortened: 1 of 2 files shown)


mcansh:
thanks @lpsinger! Not sure how we've gone this long with it like that 😅! We actually _just_ fixed this on the dev branch via [#1504](https://github.com/remix-run/remix/pull/1504/commits/bac98a5b9e29ba5a5e2288c2b75e8b5385769845?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-e37d9e295b178aa51a4880acae4b8bc7f82759e6868583513a5fa52b16846408R44)

PR #1645, opened 2022-01-25
Bundle code before upload to AWS in arc template

lpsinger:
arc-plugin-esbuild bundles and tree-shakes the code before uploading it to a Lambda function, resulting in much faster Lambda startups.

Note that the build directory has been moved from `/server/build` to `/src/build` because arc-plugin-esbuild hardcodes the directory that it scans for entry points to `/src`. (With the build directory set to `/server/build`, arc-plugin-esbuild was attempting to overwrite the file due to this issue.)


ryanflorence:
This is great thanks! We've got an improved arc template coming that I'd like to get your eyes on, want to take a look at #1664? Maybe we'll add this to it.

@brianleroux do you have any opinions on `arc-plugin-esbuild` for Remix?

lpsinger:
I am thinking that #1504 might make this PR obsolete.

ryanflorence:
alright, gonna close then, let us know if not :)

lpsinger:
Is there a simple way to point a remix project to the dev branch? If so I can take #1504 for a spin to check that it addresses this.

PR #1661, opened 2022-01-26
docs(examples): add dark mode example

mattstobbs:
Created an example of adding dark mode to Remix, based on Kent C. Dodds' implementation for his site. 

One thing I wasn't sure about was whether to include the `.env` file or not. I chose not to, as an example of keeping secrets secret, but it means the example throws an error until the user realised to add it, so it might be better to include it (I think I saw other examples use an `.env.example` file). Let me know what you think and I'll update it with one if you think it would be best.


examples/dark-mode/app/root.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+        <NonFlashOfWrongThemeEls ssrTheme={Boolean(data.theme)} />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 52:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

examples/dark-mode/app/root.tsx
mattstobbs: Wow, that's so nice!

examples/dark-mode/app/entry.server.tsx
```diff
@@ -0,0 +1,23 @@
+import { renderToString } from "react-dom/server";
+import { RemixServer } from "remix";
+import type { EntryContext } from "remix";
+
+require("dotenv").config();
```
sergiodxa, line 5: This should not be needed, none example adds dotenv to load env variables

examples/dark-mode/app/root.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const themeSession = await getThemeSession(request);
+
+  const data: LoaderData = {
+    theme: themeSession.getTheme()
+  };
+
+  return data;
+};
+
+export function App() {
```
sergiodxa, line 34: Why is this being exported? Also the code is simple enough to combine it into a single component with the default export

examples/dark-mode/app/routes/action/set-theme.tsx
```diff
@@ -0,0 +1,31 @@
...
+  themeSession.setTheme(theme);
+  return json(
+    { success: true },
+    { headers: { "Set-Cookie": await themeSession.commit() } }
+  );
+};
+
+export const loader: LoaderFunction = () => redirect("/", { status: 404 });
+
+export default function SetThemeRoute() {
```
sergiodxa, line 29: If you don’t export a component it will work as a resource route

examples/dark-mode/app/utils/theme.server.ts
```diff
@@ -0,0 +1,34 @@
+import { createCookieSessionStorage } from "remix";
+
+import { Theme, isTheme } from "./theme-provider";
+
+// Make use to set the environment variable SESSION_SECRET before running the code
+const sessionSecret = process.env.SESSION_SECRET;
```
sergiodxa, line 6: You could set a fallback value for the example to work

examples/dark-mode/app/entry.server.tsx
mattstobbs: `dotenv` has been removed 👍

examples/dark-mode/app/root.tsx
mattstobbs: This was a typo, I've removed the export. I think it still needs to be a separate component though because `useTheme` is basically a wrapper around `useContext` so we need the provider in a parent component?

examples/dark-mode/app/routes/action/set-theme.tsx
mattstobbs: Awesome, removed 👍

examples/dark-mode/app/utils/theme-provider.tsx
```diff
@@ -0,0 +1,164 @@
...
+    if (typeof window !== "object") {
+      return null;
+    }
+
+    return getPreferredTheme();
+  });
+
+  const persistTheme = useFetcher();
+  // TODO: remove this when persistTheme is memoized properly
+  const persistThemeRef = useRef(persistTheme);
```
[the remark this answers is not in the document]
mattstobbs: Ah sorry, this was based on Kent's code here: https://github.com/kentcdodds/kentcdodds.com/blob/6f9378fec907322cffa64f609b6204cd6121e777/app/utils/theme-provider.tsx#L49-L53

I must have missed the `useMemo` when I went to check if it had been updated. I've removed the `ref` now though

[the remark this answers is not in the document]
mattstobbs: Also, how did you do that line preview in your comment? I've never seen that before

examples/dark-mode/app/utils/theme.server.ts
mattstobbs: I've added a default value, as you suggested

ryanflorence:
@kentcdodds please take a look at this and merge when you're ready :)

kentcdodds (approved):
This looks great to me!

Do you suppose we should include my [`Themed`](https://github.com/kentcdodds/kentcdodds.com/blob/2af8efc6cebf25052f1f95f2210317a8086d74ef/app/utils/theme-provider.tsx#L184-L215) component and the [`handleDarkAndLightModeEls`](https://github.com/kentcdodds/kentcdodds.com/blob/2af8efc6cebf25052f1f95f2210317a8086d74ef/app/utils/theme-provider.tsx#L125-L145) that needs to be called prior to hydration?

mattstobbs:
I guess it depends whether the examples are meant to be an MVP of what they're demoing or a more complete "here's all the tools you might need". I'm happy to include them though if you think they'll add value 🙂

kentcdodds:
Let's add them. I think it's a common enough need that people will want to have them in the example. Thanks!

kentcdodds (changes requested):
Just marking this as "Request changes" for my own process 😅

Whenever you get around to it! Thanks :)

mattstobbs:
Sorry for the delay, I thought it would take longer than it did to update so I was putting it off for a moment where I'd have more time. I've added the Themed component as you suggested (I'm definitely going to need to write a part 2 to my blog post to include it).

One other thing I noticed is that, after removing the `persistThemeRef` as suggested by @sergiodxa, the `useEffect` runs repeatedly. This doesn't happen with the ref though. I'm not sure if it's a bug in how `useFetch` is memoized or whether I'm doing something wrong in the code but for now I've added the ref back.

examples/dark-mode/app/utils/theme.server.ts
```diff
@@ -0,0 +1,31 @@
...
+    httpOnly: true
+  }
+});
+
+async function getThemeSession(request: Request) {
+  const session = await themeStorage.getSession(request.headers.get("Cookie"));
+  return {
+    getTheme: () => {
+      const themeValue = session.get("theme");
+      return isTheme(themeValue) ? themeValue : Theme.LIGHT;
```
kentcdodds, line 24: This should return the theme in the session or `null`. If we just default to `Theme.LIGHT` then we'll never take the user's preference into account when they arrive on the site.

```suggestion
      return isTheme(themeValue) ? themeValue : null;
```

I tried doing this locally and had some funny stuff happen where it looked like some things showed up twice. I'm not sure what was causing that. Hopefully you don't see the same thing and it was just a weird thing on my machine or something.

examples/dark-mode/app/bootstrap.client.tsx
```diff
@@ -0,0 +1,3 @@
+import { handleDarkAndLightModeEls } from "./utils/theme-provider";
+// fixup stuff before hydration
+handleDarkAndLightModeEls();
```
kentcdodds, line 3: When I was working on this locally, I noticed that this file doesn't get loaded in the browser and this function is not actually run. Can you confirm that for me @mattstobbs? @jacob-ebey, my guess is this would be because we treat all modules as pure and side-effect free?

The objective here is we need this function run as quickly as possible. We need it to happen before importing and evaluating the `react-dom` and `remix` modules in `entry.client.tsx`. Is that not possible or did I just have something weird on my machine?

kentcdodds (changes requested):
Thanks for this! I tried playing around with things locally and had some trouble, also had some feedback. I really appreciate you taking the time on this one!

examples/dark-mode/app/bootstrap.client.tsx
jacob-ebey: yeah, importing handleDarkAndLightModeEls into entry.client and calling it there will work. That is the only place side-effects are allowed.

jacob-ebey: Why would you want to import and evaluate before react-dom and remix? That makes no sense to me in ESM land. If this was CJS module loading, then sure, but not here. You'd actually need to dynamically import things if this is *really* what you are after, but I don't believe it is.

It also just reads nicer if you import that function and call it directly before they hydrate call showing that this is in fact blocking and executing before react can hydrate.

kentcdodds: This just needs to happen *very* quickly. Maybe in that case we should move it to an inline script tag instead 🤔 Could probably put it in the other one we've already got. What do you think @mattstobbs?

mattstobbs: I don't think we could put it in the current one because the body won't have loaded yet so the JavaScript wouldn't be able to find any elements. We could potentially put it on a script tag straight after the Outlet though so it's the first thing that runs?

kentcdodds: Oh yeah, that's why I didn't include it in the other script tag. Yeah, let's put it in a separate script tag. That will be even better because we don't want to wait for the JS to load.

Another thing to maybe add is: https://github.com/kentcdodds/kentcdodds.com/blob/fa62721066f849e086bc00431982e887f1f26922/styles/app.css#L114-L118

It's possibly something we could do like this:

```css
/* default light, but app-preference is "dark" */
html.dark {
   light-mode {
    display: none;
  }
}

/* default light, and no app-preference */
html:not(.dark) {
  dark-mode {
    display: none;
  }
}

@media (prefers-color-scheme: dark) {
  /* prefers dark, but app-preference is "light" */
  html.light {
    dark-mode {
      display: none;
    }
  }

  /* prefers dark, and app-preference is "dark" */
  html.dark,
  /* prefers dark and no app-preference */
  html:not(.light) {
     light-mode {
      display: none;
    }
  }
}
```

Might even think about inlining that in a style tag as well. Could make sense to have a `<ThemeHead />` and `<ThemeBody />` component.

Another idea is to have everything in a single component people put in the head and then just use a JS event to know when the document is ready to remove all the `<(dark|light)-mode>` elements.

Anyway, there are some ideas for you. Let me know if you get tired of this and I can take it across the finish line for you. Thanks for working on this!

mattstobbs: Sorry for the long delay, things have been really busy for me. I don't know if I'll get time to work on this this week so if you have time and want to get this over the line, feel free to take over! If not, I'll try and find some time next week!

kentcdodds (approved):
Super, thank you!

kentcdodds:
This was a lot of work and I really appreciate you taking the time @mattstobbs!

PR #1538, opened 2022-01-18
feat(remix-architect): add DynamoDB tables session storage

lpsinger:

PR #1664, opened 2022-01-26
feat: add Remix Arc Stack

mcansh:
going to need to some fine tuning on the readme


lpsinger:
See #1538 for session storage integrated with arc tables, to avoid storing sensitive data in cookies.

packages/create-remix/templates/arc-stack/app.arc
```diff
@@ -0,0 +1,24 @@
...
+
+@aws
+region us-east-1
+
+@tables
+people
+  pk *String
+
+notes
+  pk **String
```
shamsup, line 19: in DynamoDB-land, `pk` generally means partition or primary key, and `sk` is for sort key, but they're defined backwards here. This is more than likely going to confuse people, especially considering the `people` table is using `pk` as the partition key.

packages/create-remix/templates/arc-stack/app/routes/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+const loader: LoaderFunction = async ({ request }) => {
+  const session = await getSession(request);
+  const user = session.get("user");
+  if (!user) return redirect("/login");
+
+  const data = await arc.tables();
+  const result = await data.notes.query({
+    KeyConditionExpression: "sk = :sk",
+    ExpressionAttributeValues: { ":sk": `email#${user.email}` }
+  });
```
shamsup, line 26: This is what threw me for a loop when I noticed the `notes` table had reversed the `pk` and `sk` definitions. My gut reaction was "this is querying with only a sort key, this seems like a bug".

mcansh: thanks for taking a look at this 🙏

jkarsrud:
As [stated in the official Architect guides](https://arc.codes/docs/en/reference/configuration/local-preferences#local-preferences) `preferences.arc` is intended to be added to `.gitignore`. That's because environment variables from all your invironments are synced there, as well as it containing local preferences like autocreate, sandbox startup scripts etc.

packages/create-remix/templates/arc-stack/server/config.arc
```diff
@@ -0,0 +1,5 @@
+@aws
+# runtime nodejs14.x
+# memory 1152
+# timeout 30
+# concurrency 1
```
tbeseda, line 5: Technically, since these settings are commented out (with the leading `# `) this file doesn't do anything. It can safely be removed.

It does demonstrate some of Arc's [function configuration options](https://arc.codes/docs/en/reference/configuration/function-config) which take priority over the `@aws` entry in the project's `app.arc` file.

Up to you if you want to include it in the stack template, I just wanted to point out that it isn't needed.

packages/create-remix/templates/arc-stack/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "license": "",
+  "scripts": {
+    "postinstall": "remix setup node",
+    "build": "remix build",
+    "dev:remix": "remix watch",
+    "dev:arc": "arc sandbox",
+    "dev": "cross-env ARC_LOCAL=1 ARC_TABLES_PORT=5555 run-p dev:*"
+  },
+  "dependencies": {
+    "@architect/functions": "5.0.0-RC.1",
```
tbeseda, line 14: IMO it would be safer to use `4.1.2` of architect/functions since a user's global Architect installation (recommended in the Readme here) is v9.

The latest arc/functions@RCs are for the Arc 10 RC.

packages/create-remix/templates/arc-stack/app.arc
```diff
@@ -0,0 +1,24 @@
...
+
+@http
+/*
+  method any
+  src server
+
+@static
+
+@aws
+region us-east-1
```
tbeseda, line 12: This is totally optional, so it's up to the Remix crew, but [the `@aws` pragma](https://arc.codes/docs/en/reference/project-manifest/aws) can be omitted and Architect will default to `us-west-2`.

packages/create-remix/templates/arc-stack/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+- Make sure you have your `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` saved to your GitHub repo's secrets. You can re-use the ones you used for the AWS CLI or you can go to your AWS [security credentials][aws_access_key_id] and click on the "Access keys" tab, and then click "Create New Access Key", then you can copy those and add them to your repo's secrets.
+
+Along with your AWS credentials, you'll also need to give your CloudFormation a `SESSION_SECRET` variable of its own for both staging and production environments.
+
+```sh
+arc env staging SESSION_SECRET $(openssl rand -hex 32)
+arc env production SESSION_SECRET $(openssl rand -hex 32)
+```
+
```
tbeseda, line 55: This is a great, brief summary of deploying with the included GitHub workflow. I might have to use something similar in some content I'm working on 👍

packages/create-remix/templates/arc-stack/README.md
```diff
@@ -0,0 +1,72 @@
...
+## Where do I find my CloudFormation?
+
+You can find the CloudFormation template that Architect generated for you in the sam.yaml file.
+
+To find it on AWS, you can search for [CloudFormation][cloudformation] (make sure you're looking at the correct region!) and find the name of your stack that matches what's in `app.arc`
+
+To find your deployed api, you can search for [APIGateway][apigateway] (make sure you're looking at the correct region!)
+
+To find your deployed lambda functions, you can search for [Lambda][lambda] (make sure you're looking at the correct region!)
```
tbeseda, line 66: I might mention that the created CloudFormation "Stack" will be named `RemixAwsStackStaging` based on the `app.arc`'s `@app` entry. Also a user can access all resources from that stack in the CloudFormation "Resources" tab.
I use this to quickly navigate to Lambdas and then to logs since it automatically goes to the correct region.

packages/create-remix/templates/arc-stack/app.arc
tbeseda: I did notice that the Readme references this region setting specifically, which may be easier than mentioning Arc's default. Totally up to you :)

packages/create-remix/templates/arc-stack/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+This is a pretty simple note-taking app, but it's a good example of how you can build a full stack app with Architect and Remix. The main functionality is creating users, logging in and out, and creating and deleting notes.
+
+### Relevant code:
+
+- creating users, and logging in and out [./app/models/user.server.ts](./app/models/user.server.ts)
+- user sessions, and verifying them [./app/session.server.ts](./app/session.server.ts)
+- creating, and deleting notes [./app/models/note.server.ts](./app/models/note.server.ts)
+
+The database that comes with `arc sandbox` is an in memory database, so if you restart the server, you'll lose your data. Production environments won't behave this way.
```
tbeseda, line 37: This is a good caveat to mention about how sandbox emulates the db locally.

It might also be helpful to the user to mention that in staging and production on AWS, the database integration hooks up to DynamoDB so data will be persisted between Lambda executions and through new deployments.

tbeseda (commented):
I left a few comments/thoughts on Arc-specific things, but overall it looks good 👍 
(caveat: I didn't look closely at the session and db interactions since it's based on an Arc example.)

I mentioned in a review comment that we're currently working on Architect 10. Arc 10 has some extensive work being done on our plugin APIs. So I'm actively working on an Arc Remix plugin as a way to test those interfaces. This template is definitely helpful.

Since Arc 10 is currently in the RC stage, it's definitely best to move forward with your approach here. As we get close to a v10 release, maybe we can pair on how to best integrate Remix + Arc from both the Remix angle and the Arc angle for users.

Let me know if I can help with anything else. I'm excited for the crossover here.
Thanks again for reaching out, Logan.

mcansh:
> I left a few comments/thoughts on Arc-specific things, but overall it looks good 👍 (caveat: I didn't look closely at the session and db interactions since it's based on an Arc example.)
> 
> I mentioned in a review comment that we're currently working on Architect 10. Arc 10 has some extensive work being done on our plugin APIs. So I'm actively working on an Arc Remix plugin as a way to test those interfaces. This template is definitely helpful.
> 
> Since Arc 10 is currently in the RC stage, it's definitely best to move forward with your approach here. As we get close to a v10 release, maybe we can pair on how to best integrate Remix + Arc from both the Remix angle and the Arc angle for users.
> 
> Let me know if I can help with anything else. I'm excited for the crossover here. Thanks again for reaching out, Logan.

Awesome! thanks again for doing this @tbeseda, I would be more than happy to pair on the Remix plugin :)

PR #1666, opened 2022-01-26
chore: add `prettier-plugin-tailwindcss`

MichaelDeBoey:
This will order Tailwind classes according to the official recommended way

https://tailwindcss.com/blog/automatic-class-sorting-with-prettier


mcansh (approved):

ryanflorence:
Thanks, but we don't really need this in the source repo. Would be great in the tailwind example though :)

PR #1669, opened 2022-01-26
docs(example): add an example of using useMatches to access loader data across your application

andrelandgraf:
Creating a useful example for the examples collection. It showcases the usage of `useMatches` to access loader data across the Remix application.

**Motivation:** With Remix, you don't need global React contexts anymore to access your
React state. Instead, you can use the `useMatches` hook to access the loader
data across your application.

PS: Remix rocks!


andrelandgraf:
You can review this example in codesandbox: https://codesandbox.io/s/usematches-loader-data-2h798

examples/usematches-loader-data/app/root.tsx
```diff
@@ -0,0 +1,38 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 34:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

ryanflorence:
@kentcdodds please review this and merge if you like it or request changes :)

examples/usematches-loader-data/app/useLoaderStore.ts
```diff
@@ -0,0 +1,20 @@
+import { useMemo } from "react";
+import { useMatches } from "remix";
+
+/**
+ * This base hook is used in other hooks to quickly search for specific data
+ * across all loader data using useMatches.
+ * @param {string} key
+ * @returns {unknown} The value of the key based on your currently loaded routes.
+ */
+export default function useLoaderStore(key: string): unknown {
```
kentcdodds, line 10: Interesting approach.

I don't think I'm a fan, because you could have an unexpected behavior if you had two parent loaders that use the same key (for example, you have a `/users/$userId` where that loader returns a `user` for that route).

I'd rather have something like:

```suggestion
export default function useParentLoaderData(id: string): unknown {
```

Where the `id` is the ID of the route and it gives you *all* the loader data for that route.

examples/usematches-loader-data/app/useUser.ts
```diff
@@ -0,0 +1,28 @@
...
+
+function isUser(user: unknown): user is User {
+  return (
+    !!user &&
+    typeof (user as User).name === "string" &&
+    typeof (user as User).email === "string"
+  );
+}
+
+export default function useUser(): User | undefined {
```
kentcdodds, line 16: When something like this can return `undefined`, I like giving it clear distinction with `Optional`

```suggestion
export default function useOptionalUser(): User | undefined {
```

And then I have another hook (in this case `useUser`) which throws if the value is undefined. That way I can have a component use this hook and not worry about the `undefined` case. Do you think you could put that together?

examples/usematches-loader-data/README.md
```diff
@@ -0,0 +1,32 @@
+# useMatches example
+
+Use the `useMatches` hook to access loader data across your application.
+
+With Remix, you don't need global React contexts anymore to access your
+React state. Instead, you can use `useMatches` to access loader
+data across your application.
```
kentcdodds, line 7:
```suggestion
With Remix, you don't need global React contexts anymore to access your application state. Instead, you can use `useMatches` to access loader data across your application.

Note: if you have UI state (non-persisted state managed by React) then you'd probably want to look into [the `context` prop on `<Outlet />`](https://remix.run/docs/en/v1/api/remix#outlet-context-) combined with [the `useOutletContext` hook](https://remix.run/docs/en/v1/api/remix#useoutletcontext).
```

kentcdodds (changes requested):
This is quite helpful! Thanks for the work on it!

examples/usematches-loader-data/app/useLoaderStore.ts
andrelandgraf: Thank you for the review and the detailed change request! I will adapt the code and push a commit. 

Just one follow-up question @kentcdodds : The naming `useParentLoaderData` could be confusing as you can theoretically also access data of your child routes, right? Imagine a global nav component that displays a more detailed menu once the project details page returns the `project` object from its loader. So it could also use the `useParentLoaderData` hook to access optional data from its child routes. That might be bad practice though. Your call!

andrelandgraf: Maybe `useMatchesLoaderData`?

examples/usematches-loader-data/app/useUser.ts
andrelandgraf: Great point! Pushed the changes!

examples/usematches-loader-data/app/useMatchesData.ts
```diff
@@ -0,0 +1,21 @@
...
+import { useMatches } from "remix";
+
+/**
+ * This base hook is used in other hooks to quickly search for specific data
+ * across all loader data using useMatches.
+ * @param {string} id The route id
+ * @param {string} key The object key (optional)
+ * @returns {unknown} The value of the key based on your currently loaded routes.
+ */
+export default function useMatchesData(id: string, key?: string): unknown {
```
kentcdodds, line 11: I don't personally like the `key` thing. It's not saving much from a typing perspective and it's adding API that people have to learn and understand.

examples/usematches-loader-data/app/useOptionalUser.ts
```diff
@@ -13,8 +13,17 @@ function isUser(user: unknown): user is User {
...
+ * Build a custom hook for each data object
+ * of your loader data.
+ * Use useMatchesData to access loader data
+ * across your application.
+ * Use tiny-invariant and Typescript "is"
+ * to require data on runtime.
+ * Or return undefined if data is optional and not found.
+ */
+export default function useOptionalUser(): User | undefined {
+  const user = useMatchesData("root", "user");
```
kentcdodds, line 26: I would prefer:

```suggestion
  const { user } = useMatchesData("root");
```

kentcdodds (changes requested):
I'm liking this even better. Thanks for making the adjustment! Just one last thing.

examples/usematches-loader-data/app/useOptionalUser.ts
andrelandgraf: @kentcdodds Implemented! I had to change the return type of `useMatchesData` to safely destruct the user in `useOptionalUser`. Now the return value of `useMatchesData` cannot be undefined anymore but empty Json object. Lmk if that works!

examples/usematches-loader-data/app/useMatchesData.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @returns {JSON} The value of the key based on your currently loaded routes.
+ */
+export default function useMatchesData(id: string): Record<string, unknown> {
+  const matchingRoutes = useMatches();
+  const route = useMemo(() => matchingRoutes.find((route) => route.id === id), [
+    matchingRoutes,
+    id
+  ]);
+  return route?.data || {};
+}
```
kentcdodds, line 17:
```suggestion
export default function useMatchesData(id: string): Record<string, unknown> | undefined {
  const matchingRoutes = useMatches();
  const route = useMemo(() => matchingRoutes.find((route) => route.id === id), [
    matchingRoutes,
    id
  ]);
  return route?.data;
}
```

examples/usematches-loader-data/app/useOptionalUser.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { user } = useMatchesData("root");
+  /*
+   * You can make use of tiny-invariant here to throw runtime errors
+   * if the loader data is required and not optional
+   */
+  if (!isUser(user)) {
+    return undefined;
+  }
+  return user;
+}
```
kentcdodds, line 35: Let's call this file `user.ts` and have two named exports:

```suggestion
export function useOptionalUser(): User | undefined {
  const data = useMatchesData("root");
  /*
   * You can make use of tiny-invariant here to throw runtime errors
   * if the loader data is required and not optional
   */
  if (!data || !isUser(data.user)) {
    return undefined;
  }
  return data.user;
}

export function useUser(): User {
  const maybeUser = useUser();
  if (!maybeUser) {
    throw new Error("No user found in root loader, but user is required by useUser. If user is optional, try useOptionalUser instead.");
  }
  return maybeUser
}
```

kentcdodds (changes requested):
Thanks for the back-and-forth. I think we're getting close to a terrific solution 👍

examples/usematches-loader-data/app/useMatchesData.ts
andrelandgraf: > Thanks for the back-and-forth. I think we're getting close to a terrific solution 👍

Thank you for taking the time! `useMatches` is awesome but it's low-level, so it's great that we reiterate on this particular usage example! I really appreciate the feedback! :) 

Snap! I had issues because `unknown` was complaining but I already fixed it by using `Record<string, unknown>`, so there was really no need to avoid the undefined anymore! Thanks for the change request!

examples/usematches-loader-data/app/useOptionalUser.ts
andrelandgraf: @kentcdodds I like it! You called `useUser` inside `useUser` though :P I fixed it and pushed the changes!

kentcdodds (approved):
Super, thanks for all of this! Well done!

PR #1676, opened 2022-01-27
docs(faq): Embed multiple forms video in docs/faq.md

hollandThomas:
The Remix Single video series is very cool! It deserves to be in the spotlight more prominently.

This makes the YouTube video on multiple forms appear more prominent in the FAQs, which I think is both beneficial for the docs itself and the promotion of the YouTube channel in general.

If you like this PR, @ryanflorence should probably add some nice thumbnails on YouTube (which would be a nice touch regardless) before merging it, otherwise it renders as something like this:

Rendered preview of the proposed changes

### DIFF
--- a/docs/pages/faq.md
+++ b/docs/pages/faq.md
@@ -64,7 +64,9 @@ export async function loader({ request }) {
 
 ## How do I handle multiple forms in one route?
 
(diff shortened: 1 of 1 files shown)


ryanflorence:
thanks, we've got plans for embedding them right in the docs :)

PR #1679, opened 2022-01-27
docs(example): add an example for server-side rendering of markdown (fetched from anywhere)

andrelandgraf:
Remix supports MDX out of the box. This is great if you have your markdown/mdx files at compile time! If you are fetching your markdown files from a remote source, you have to work with unified or a wrapper such as react-remark.

Unfortunately, react-remark does not support cjs anymore but Remix.run will not work with esm just yet (without async imports). The older versions of react-remark do not implement useRemarkSync. Thus, right now there is no way to use react-remark in remix to render markdown server-side since it will always requires async code (and useEffect).

This example showcases how you can implement a small markdown container to render markdown strings server-side (without useEffect).


examples/render-md-server-side/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+A solution: Make use of old versions of remark, unified, and rehype directly to basically reimplement react-remark for remix. Sounds complicated but it's really not!
+
+We wrap everything up in one small component that we then can use to render markdown across our application! We can even apply custom plugins and use custom React components for the output of our markdown! And the best: everything renders nicely on the server! 🚀
+
+## Preview
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+[![Open in CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/render-md-sync-dgdze)
```
sergiodxa, line 15: This CodeSandbox url should be something like https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/client-only-components using the repo and not a custom slug, that way clicking the link will open a sandbox with the lastest version of this example, with a custom CodeSandbox url you will have to keep the sandbox code updated with the example here

examples/render-md-server-side/app/routes/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  // Fetch the markdown content server-side
+  const content = await fetchMarkdownFormSomewhere();
+  return { content };
+};
+
+export default function Index() {
+  // Access the markdown in your app
+  const { content } = useLoaderData<{ content: string }>();
+  // Render the markdown anywhere in your app (it also works server-side!)
+  return <MarkdownContainer source={content} />;
```
sergiodxa, line 39: Wouldn’t it be better to transform MD to HTML in the loader and return html directly? That way you don’t even need to ship MD parsing code to the browser

andrelandgraf: Good point! Would that also work with custom React components?

This implementation supports both server-side and client-side markdown strings. E.g. you can also use it to implement a preview for a textarea with markdown support. What do you think? I can definitely refactor this to render the HTML server side!

examples/render-md-server-side/README.md
andrelandgraf: Thanks for the explanation! I updated the url. It's not working now though, since the example is not present in the main branch. Just for review purposes, here is the link to the custom one: https://codesandbox.io/s/render-md-sync-dgdze

examples/render-md-server-side/app/root.tsx
```diff
@@ -0,0 +1,27 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 23:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

ryanflorence:
We're working on ESM for the server build, so remark's latest versions should work soon if you want to opt-in to ESM-only apps, I think maybe it'd be better to wait for that and then make an example that's far more straightforward?

andrelandgraf:
> We're working on ESM for the server build, so remark's latest versions should work soon if you want to opt-in to ESM-only apps, I think maybe it'd be better to wait for that and then make an example that's far more straightforward?

That's great news! Will close this PR then. Next example can just take advantage of `react-markdown` or a similar package.

PR #1685, opened 2022-01-28
Better websocket support

mehulmpt:
The current remix implementation hardcodes the hostname and protocol of websocket, making it unusable over reverse tunnelling proxies like ngrok.

I ran into this issue while trying to deploy a Remix playground on codedamn (https://codedamn.com/playgrounds)
It fails to connect because codedamn playground websocket is listening on the same host but port 8002.

(Next addition should be running the development WebSocket server on the same HTTP port over which remix serves the app to save resources)

In the screenshot below, Remix should ideally connect to `wss://dawn-gravity.webx.sh:8002`

Screenshot 2022-01-28 at 6 27 50 AM

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -116,6 +116,7 @@
 - mcansh
 - medayz
 - meetbryce
+- mehulmpt
 - michaeldeboey
 - michaelfriedman
 - mjackson
--- a/contributors.yml
+++ b/contributors.yml
@@ -75,6 +75,7 @@
 - IshanKBG
 - jacob-ebey
 - jakewtaylor
+- jamiebuilds
 - jaydiablo
 - jenseng
 - jeremyjfleming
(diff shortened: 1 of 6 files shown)


ryanflorence:
Please make non-documentation PRs against the `dev` branch :)

mehulmpt:
Done @ryanflorence

ryanflorence:
Haven't dug in here yet, but quick note:

> Next addition should be running the development WebSocket server on the same HTTP port over which remix serves the app to save resources

Remix isn't your app's dev server (unless you use `remix dev`), it can be AWS lambda emulators architect, vercel and netlify, your own express server, a firebase emulator, etc. So we can't run the socket for the asset server on the same port as the app server.

ryanflorence:
the commit history was off, follow in #1881

PR #1686, opened 2022-01-28
chore: Fix tests

mjackson:
This should fix the tests so we can run them locally again. The main thing left to do is to remove a bunch of code duplication in the gists-app tests. There are a few lines in each file that setup the server port and host that should probably be consolidated somewhere.


kentcdodds (commented):
So far so good from my perspective 👍

kentcdodds:
I'm guessing this is marked as a draft because it's incomplete?

PR #1688, opened 2022-01-28
docs: access Cloudflare KV namespace via env

justinnoel:
According to the Cloudflare docs and in practice, the KV namespace is only available on the `env` and can't be accessed directly.

https://developers.cloudflare.com/pages/platform/functions#kv-namespace-locally

### DIFF
--- a/docs/guides/data-loading.md
+++ b/docs/guides/data-loading.md
@@ -169,8 +169,8 @@ If you picked Cloudflare Workers as you environment, [Cloudflare Key Value][clou
 import { useLoaderData } from "remix";
 import type { LoaderFunction } from "remix";
 
-export let loader: LoaderFunction = async ({ params }) => {
-  return PRODUCTS_KV.get(`product-${params.productId}`, {
+export let loader: LoaderFunction = async ({ env, params }) => {
+  return env.PRODUCTS_KV.get(`product-${params.productId}`, {
     type: "json"
   });
 };
--- a/docs/guides/data-loading.md
+++ b/docs/guides/data-loading.md
@@ -169,8 +169,8 @@ If you picked Cloudflare Workers as you environment, [Cloudflare Key Value][clou
 import { useLoaderData } from "remix";
 import type { LoaderFunction } from "remix";
 
-export let loader: LoaderFunction = async ({ env, params }) => {
-  return env.PRODUCTS_KV.get(`product-${params.productId}`, {
+export let loader: LoaderFunction = async ({ context, params }) => {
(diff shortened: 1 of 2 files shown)


docs/guides/data-loading.md
```diff
@@ -169,8 +169,8 @@ If you picked Cloudflare Workers as you environment, [Cloudflare Key Value][clou
 import { useLoaderData } from "remix";
 import type { LoaderFunction } from "remix";
 
-export let loader: LoaderFunction = async ({ params }) => {
-  return PRODUCTS_KV.get(`product-${params.productId}`, {
+export let loader: LoaderFunction = async ({ env, params }) => {
+  return env.PRODUCTS_KV.get(`product-${params.productId}`, {
```
Ponjimon, line 173: Isn't the `env` property within the `context` object?

```suggestion
export let loader: LoaderFunction = async ({ context, params }) => {
  return context.env.PRODUCTS_KV.get(`product-${params.productId}`, {
```

justinnoel: I don't believe so. The `context` is the parameter of the loader function and `env`, `params`, etc are being destructured out of it.

`export let loader: LoaderFunction = async ({ env, params }) => {` 

is equivalent to

```
export let loader: LoaderFunction = async (context) => {
const { env, params } = context;
}
```

Ponjimon: But the `LoaderFunction` type has `DataFunctionArgs` as the type for the arguments:

```ts
export interface DataFunctionArgs {
    request: Request;
    context: AppLoadContext;
    params: Params;
}
```

Therefore: 
image

ryanflorence: yep, this needs to be on `context.env`

justinnoel: Great job catching my mistake Ponjimon and thanks for the clarification Ryan.

Thanks!

ryanflorence:
@jacob-ebey I'm not familiar enough with this, can you merge this or request changes? 🙏

jacob-ebey:
I'm not sure if this should be merged. In the latest CF Pages template going out in the next release we pass through the CF "env" as the Remix context for loaders / actions, so this isn't *quite* inline with what we will be defaulting to for that deployment target. It will actually be more like:
```js
function loader({ context }) { context.YOUR_KV_NAMESPACE.xyz() }
```

Maybe we do want to pass through the whole CF context instead of ".env" from it? What do you think @justinnoel ?

justinnoel:
@jacob-ebey Sorry for the horribly late reply.

I agree now, this is simply accessible via `context.NAME_OF_ENV_OR_KV_HERE.

PR #1693, opened 2022-01-28
Explain how to create your own hashed password in the jobs tutotial

andresgutgon:
# What?
While following the jobs tutorial I wanted to use a different password for my seed user. So I though it was helpful to show how to do this. Maybe is already explained and I missed. In that case sorry for the noise


ryanflorence:
Hey thanks for this but the tutorial is already quite large!

andresgutgon:
No problem :)

Issue #1618, opened 2022-01-23
[Bug]: TypeError: Cannot read properties of undefined (reading 'root')


adrianiskandar:
### Which Remix packages are impacted?

- [X] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

    "remix": "^1.1.3",

### What version of Node are you using? Minimum supported version is 14.

v16.4.2

### Steps to Reproduce

I was just trying to follow the first tutorial, and right after the part when you create the posts .md files everything was working ok. But after you add the code mentioned in the tutorial: 

```
import path from 'path';
import fs from 'fs/promises';
import parseFrontMatter from 'front-matter';
import invariant from 'tiny-invariant';

export type Post = {
  slug: string;
  title: string;
};

export type PostMarkdownAttributes = {
  title: string;
};

const postsPath = path.join(__dirname, '..', 'posts');

function isValidPostAttributes(attributes: any): attributes is PostMarkdownAttributes {
  return attributes?.title;
}

export async function getPosts() {
  const dir = await fs.readdir(postsPath);
  return Promise.all(
    dir.map(async (filename) => {
      const file = await fs.readFile(path.join(postsPath, filename));
      const { attributes } = parseFrontMatter(file.toString());
      invariant(isValidPostAttributes(attributes), `${filename} has bad meta data!`);
      return {
        slug: filename.replace(/\.md$/, ''),
        title: attributes.title,
      };
    })
  );
}
```

Then the application breaks and starts showing this error on localhost/posts
```
TypeError: Cannot read properties of undefined (reading 'root')
    at RemixRoute (http://10.0.0.229:3000/build/_shared/chunk-EW2HPZY2.js:3039:19)
    at renderWithHooks (http://10.0.0.229:3000/build/entry.client-X6Z7634E.js:11070:26)
    at mountIndeterminateComponent (http://10.0.0.229:3000/build/entry.client-X6Z7634E.js:13190:21)
    at beginWork (http://10.0.0.229:3000/build/entry.client-X6Z7634E.js:13977:22)
[...]

PR #1698, opened 2022-01-28
fix: avoid `export *` in magic files for shaking

jacob-ebey:
fix: add mainFields to client build to resolve ESM
fix: force sideEffects false for browser route modules

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -356,7 +356,7 @@ async function createBrowserBuild(
     sourcemap: options.sourcemap,
     metafile: true,
     incremental: options.incremental,
-    mainFields: ["module", "browser", "main"],
+    mainFields: ["browser", "module", "main"],
     treeShaking: true,
     minify: options.mode === BuildMode.Production,
     entryNames: "[dir]/[name]-[hash]",
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -356,6 +356,7 @@ async function createBrowserBuild(
     sourcemap: options.sourcemap,
     metafile: true,
     incremental: options.incremental,
+    mainFields: ["module", "browser", "main"],
     treeShaking: true,
     minify: options.mode === BuildMode.Production,
     entryNames: "[dir]/[name]-[hash]",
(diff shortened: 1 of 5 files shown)


jacob-ebey:
Fixes https://github.com/remix-run/remix/issues/1618

packages/remix-dev/compiler.ts
```diff
@@ -356,6 +356,7 @@ async function createBrowserBuild(
     sourcemap: options.sourcemap,
     metafile: true,
     incremental: options.incremental,
+    mainFields: ["module", "browser", "main"],
```
jacob-ebey, line 359: I think browser might want to come first on second thought.

packages/remix/index.ts
```diff
@@ -1,6 +1 @@
-// @ts-ignore
-export * from "./client";
-// @ts-ignore
-export * from "./server";
-// @ts-ignore
-export * from "./platform";
+throw new Error("Did you forget to run `remix setup` for your platform?");
```
kentcdodds, line 1: Oh nice! This will be very helpful :)

packages/remix-dev/compiler.ts
mjackson, line 359: This is the main line that makes me a little nervous about this PR. If we're specifically building for the browser, and someone specifies the browser field in package.json, it feels like that should take precedence. Can you say a little more about why this change is necessary?

packages/remix-dev/compiler/plugins/browserRouteModulesPlugin.ts
```diff
@@ -38,7 +38,11 @@ export function browserRouteModulesPlugin(
       );
 
       build.onResolve({ filter: suffixMatcher }, args => {
-        return { path: args.path, namespace: "browser-route-module" };
+        return {
+          path: args.path,
+          namespace: "browser-route-module",
+          sideEffects: false
```
mjackson, line 44: Why do we need this?

packages/remix-dev/compiler.ts
jacob-ebey: My thought was that the "browser" field just means "this can run in the browser" and doesn't mean "ESM format". Since we are bundling to ESM, might as well take the module if it exists. On second thought though I've reverted it to be browser first. We need some definition for mainFields here though otherwise esbuild falls back to it's default behavior of "if something was `required` anywhere, always take "main". Without this we break client bundle tree shaking.

packages/remix-dev/compiler/plugins/browserRouteModulesPlugin.ts
jacob-ebey: We don't, leftover code that I removed :)

PR #1708, opened 2022-01-29
docs: fixed wrong type import

ShafSpecs:
_docs: Fixed wrong import of `UploadHandler` type from `remix` to `@remix-run/node/formData`


ryanflorence:
we don't want deep imports, looks like the problem is we aren't exporting the types from the main package, would love a PR that does that :)

PR #1715, opened 2022-01-29
Fix link to NavLink component docs

geoffreyjball:

PR #1718, opened 2022-01-29
Change < and > in Acessibility guide to their html escaped versions

Runner-dev:
This is important because in the current [guide](https://remix.run/docs/en/v1.1.3/guides/accessibility#links) the navlink component is getting rendered as an empty html element rendering the text "Remix also provides the which behaves..." while it should have been "Remix also provides the <NavLink> which behaves...."

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -132,6 +132,7 @@
 - RossJHagan
 - RossMcMillan92
 - rphlmr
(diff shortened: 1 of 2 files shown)


Runner-dev:
Not sure but #1715 may also fix this issue in a different way or be complimentary to this fix.

PR #1731, opened 2022-01-30
docs(env-vars): update dotenv config

MichaelDeBoey:


ryanflorence:
Thanks, but I'd rather not have dotenv stuff in application code, feel free to open it if there's a reason to do it differently :)

PR #1734, opened 2022-01-30
docs(examples): rename `with-rust` to `rust`

MichaelDeBoey:
As @kentcdodds asked already in some PRs: there shouldn't be a `with-` prefix in the examples names


ryanflorence:
@kentcdodds what's the thinking here? ... is there ever an example that that isn't `with-`?

kentcdodds:
Yeah, the `with-` prefix is a thing that the Next.js examples do and it's completely redundant, so I decided to not have the `with-` prefix for our examples.

PR #1735, opened 2022-01-31
docs(examples): indicate in jokes tutorial that fly.io has now a free tier for hosting databases

mkrtchian:
Fly.io has now a free tier also for hosting databases (3GB). They have written a [blog post about that](https://fly.io/blog/free-postgres/).

I have followed the jokes tutorial yesterday and the volume for the database created automatically was 1GB instead of 10GB I guess before.
They are asking a credit card during the account creation though.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -123,6 +123,7 @@
 - meetbryce
 - michaeldeboey
 - michaelfriedman
+- mkrtchian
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
Thank you for this update!

PR #1744, opened 2022-01-31
fix: allow remix-serve to run while off network

BenMcH:
remix-run#1328 prevents us from running remix locally while off of a network (On a train or bus for example during a commute)

This PR should allow us to run apps locally and simply receive a warning otherwise.


BenMcH:
Closing in favor of #1743

PR #1746, opened 2022-01-31
docs(jokes): fix divergent login.tsx import example

monitaure:
In Jokes tutorial, fix a summary of `app/routes/login.tsx` which has divergent imports from previous summaries.


kentcdodds (approved):
Thank you!

PR #1748, opened 2022-01-31
fix: allow import of node builtins in routes

jacob-ebey:
this adds polyfills for native node modules for the browser. These don't end up being shipped in the bundle if you don't use them outside your loaders / actions, but it prevents nasty failures at runtime caused by esbuild not knowing it can tree shake out those imports.

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -370,6 +370,7 @@ async function createBrowserBuild(
       )
     },
     plugins: [
+      NodeModulesPolyfillPlugin(),
       mdxPlugin(config),
       browserRouteModulesPlugin(config, /\?browser$/),
       emptyModulesPlugin(config, /\.server(\.[jt]sx?)?$/)


jacob-ebey:
Prevents errors such as: https://github.com/remix-run/remix/issues/1618

mjackson:
Two things:
- I believe this will prevent people from using their own `path` shims in browser builds because instead of looking in `node_modules` for the shim this plugin will use an empty module instead
- I believe what we really want instead of ignoring builtin modules is to mark them as `sideEffects: false` so that esbuild is able to remove their imports

Also, are we already using this plugin on the server build? I couldn't figure out why this PR didn't include an import for the new plugin.

jacob-ebey:
> Two things:
> 
> * I believe this will prevent people from using their own `path` shims in browser builds because instead of looking in `node_modules` for the shim this plugin will use an empty module instead
> * I believe what we really want instead of ignoring builtin modules is to mark them as `sideEffects: false` so that esbuild is able to remove their imports
> 
> Also, are we already using this plugin on the server build? I couldn't figure out why this PR didn't include an import for the new plugin.

It's only for cloudflare. It's to better support that ecosystem as many packages will run there with these shims that can't currently. Looking at things the community is doing like using prisma on CF, they're having to add either this plugin or switching to rollup / webpack for the second build step to add this behavior. Since we are trying to eliminate that second build step we need a solution for utilizing packages like that out of the box.

I don't think we have a way for people to provide a `path` shim in the browser as of now do we? Even so, this could just provide it for them. It ultimately uses `rollup-plugin-node-polyfills` under the hood.

sergiodxa (approved):

jacob-ebey:
Closed in favor of https://github.com/remix-run/remix/pull/1764 that has tests.

PR #1754, opened 2022-02-01
docs(examples): add GraphQL API example

visormatt:
# Overview

This PR demonstrates making use of Remix resource routes to interact and fetch from a GraphQL API. Additionally I've opted to add in a `codegen` setup which automates the creation of types which can then be used in by the `useLoaderData` hooks for improved typings.

Thanks to everyone involved in this fantastic project and LMK how I can improve on this PR.


hollandThomas:
There's something to be said about "just using `fetch`" for GraphQL queries. That makes it pretty simple to follow.

However, it might be beneficial for the example to build upon prior art and use this library instead
https://github.com/thomasheyenbrock/remix-graphql

I think that including codegen for TS types is good. I'm always a fan of examples that are a bit closer to real-world scenarios than just the bare minimum. It just seems that Remix wants the examples to be very focused:

> This means you should avoid adding stuff that isn't absolutely necessary for the example. Start with bare bones and add only what you need.

examples/graphql-api/app/root.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 30:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

examples/graphql-api/app/utils/index.ts
```diff
@@ -0,0 +1,27 @@
+import { GRAPHQL_API } from "~/config";
```
ruisaraiva19, line 1: Maybe `GRAPHQL_API` should come from the environment variables.

examples/graphql-api/app/utils/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * @name fetchFromGraphQL
+ * @external https://css-tricks.com/raw-graphql-querying
+ * @description This function is used to fetch data from the GraphQL API.
+ * Check out the link above for more information.
+ */
+ export const fetchFromGraphQL = async (
+    operation: 'mutation' | 'query',
```
ruisaraiva19, line 10: The GraphQL query/mutation/subscription should always be provided via the `query` body property.

```suggestion
```

examples/graphql-api/app/utils/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    payload: any,
+    variables?: any
+  ) => {
+    const body = {
+      [operation]: payload
```
ruisaraiva19, line 15:
```suggestion
      query: payload
```

examples/graphql-api/app/utils/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * @name fetchFromGraphQL
+ * @external https://css-tricks.com/raw-graphql-querying
+ * @description This function is used to fetch data from the GraphQL API.
+ * Check out the link above for more information.
+ */
+ export const fetchFromGraphQL = async (
+    operation: 'mutation' | 'query',
+    payload: any,
```
ruisaraiva19, line 11:
```suggestion
    payload: string,
```

examples/graphql-api/app/utils/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    variables?: any
```
ruisaraiva19, line 12:
```suggestion
    variables?: { [key: string]: any }
```

examples/graphql-api/app/utils/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    };
+
+    if (variables) body.variables = variables;
+
+    return fetch(GRAPHQL_API, {
+      body: JSON.stringify(body),
+      headers: {
+        'Content-Type': 'application/json'
+      },
+      method: 'post',
```
ruisaraiva19, line 25:
```suggestion
      method: 'POST',
```

examples/graphql-api/app/utils/index.ts
```diff
@@ -0,0 +1,27 @@
+import { GRAPHQL_API } from "~/config";
+
```
ruisaraiva19, line 2: /**
 * Clean a GraphQL document to send it via a GET query
 *
 * @param {string} str GraphQL query
 * @returns {string} Cleaned query
 */
const queryCleanner = (str: string): string => str.replace(/([\s,]|#[^\n\r]+)+/g, ' ').trim()

examples/graphql-api/.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules
+
+/.cache
+/build
+/public/build
```
ruisaraiva19, line 5:
```suggestion
/public/build
/app/generated/types.ts
```

examples/graphql-api/app/utils/index.ts
visormatt: Doh, thanks 👍

examples/graphql-api/.gitignore
visormatt: So I went with a `/app/generated/*` as the codegen supports multiple outputs, seemed like it would grow better

visormatt:
@hollandThomas thanks for the initial feedback! I've taken a look at the [thomasheyenbrock/remix-graphql](https://github.com/thomasheyenbrock/remix-graphql) but that feels like I'd be moving backwards and adding more than is necessary, thoughts?

examples/graphql-api/app/utils/index.ts
visormatt: Sure thing! I'm not sure how that will work with the Code Sandbox link, any idea there or would the user have to duplicate the file in there?

examples/graphql-api/app/root.tsx
visormatt: Thanks @MichaelDeBoey

sergiodxa:
What do you think of using a resource route to expose a GraphQL ending in Remix? That way the example could show both consume an endpoint and be the endpoint, ideally also have a route showing how to run the query against the schema and resolvers without doing a request.

kentcdodds:
Whoops! Accidentally clicked the close button 😅

Was just going to say that I like the resource route graphql endpoint idea as well 👍 Then you could show people that they could have a mobile client that hits the same endpoint which is a big reason people like graphql :)

kcoyner:
I cloned this example and while it worked great when running "npm run dev", it did not work for me when I ran "npm run build" and  "npm start".  It would throw an Application Error "Error: GRAPHQL_API is required" even though the environment variable is properly set in .env and as I said, works fine for "npm run dev".  Would love to get this working in production, not just dev.  Thank you and nice work!

examples/graphql-api/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "private": true,
+  "name": "remix-app",
+  "description": "",
+  "license": "",
+  "scripts": {
+    "build": "node -r dotenv/config node_modules/.bin/remix build",
+    "dev": "node -r dotenv/config node_modules/.bin/remix dev",
```
sergiodxa, line 8: I think the example shouldn't come with dotenv already installed, other examples have the .env.example file but let the developer see how to use them

examples/graphql-api/app/styles/index.css
```diff
@@ -0,0 +1,74 @@
+a,
```
sergiodxa, line 1: Do you need this, and other inline, styles? So far examples are really simple and to the point, I think just using a `ul` and `li` for the list of characters with the browser default styles would be enough to show the point of the example

examples/graphql-api/app/components/CharacterList.tsx
```diff
@@ -0,0 +1,36 @@
...
+export interface CharacterListProps {
+  data: Characters['results'];
+}
+
+/**
+ * @name CharacterList
+ * @description This component loops over an array of characters and renders
+ * a link to each characters detail page.
+ */
+export const CharacterList: React.FC<CharacterListProps> = (props) => {
```
sergiodxa, line 14: Similar questions as styles, this component is only used in the route, why not keep it there? It's also simple enough to directly inline the JSX in the route where it's being used and it will make the example simpler to read by using less files

examples/graphql-api/app/components/CharacterDetail.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface CharacterDetailProps {
+  data: Character;
+}
+
+/**
+ * @name CharacterDetail
+ * @description Detail view for a single Character.
+ */
+export const CharacterDetail: React.FC<CharacterDetailProps> = (props) => {
```
sergiodxa, line 12: Same as with CharacterList

examples/graphql-api/app/styles/index.css
visormatt: Sure we can trim all that down here and the other bits as well.

visormatt:
@kcoyner have a [look here](https://github.com/remix-run/remix/pull/1754/commits/7b771fb00fa625f9f37de5b07240556c17de2f5c)

examples/graphql-api/package.json
visormatt: I took a look at some of the [other examples](https://github.com/remix-run/remix/search?l=JSON&q=%22dotenv%22%3A&type=code) in here which have included the package. Happy to remove it though, do you have an "example" that I could use as reference?

visormatt:
@sergiodxa thanks for the review and feedback, I've taken a crack at the general updates but wanted to better understand the `resource route` comment a bit better before doing anything there. From what I understand I could easily create an `api/character.ts` route with no component effectively giving me an API which any client can consume, example below

<details>
<summary>Simple API endpoint</summary>

```tsx
import type { ApolloError } from "apollo-server-errors";

import { fetchFromGraphQL } from "~/utils/index";
import type { Character } from "~/generated/types";

type LoaderData = {
  data: { character: Character };
  errors?: ApolloError[];
};

export const loader: LoaderFunction = async (args): Promise<LoaderData> => {
  const { params } = args;

  const getCharacterQuery = `
      fragment CharacterFields on Character {
        gender
        id
        image
        name
        origin {
          dimension
          name
          type
        }
        species
        status
        type
      }

      query getCharacter($id: ID!) {
        character(id: $id) {
          ...CharacterFields
        }
      }
    `;

  const res = await fetchFromGraphQL(getCharacterQuery, { id: params.id });
  return res.json();
};
```
</details>

Then I can query that endpoint in the loader, but I'm not fully understanding how to accomplish the bit here ` ideally also have a route showing how to run the query against the schema and resolvers without doing a request.` specifically `without doing a request` 🤔

examples/graphql-api/package.json
sergiodxa: This one https://github.com/remix-run/remix/tree/main/examples/collected-notes has the .env.example file but it didn’t add dotenv as dependency for example.

visormatt: Thanks Sergio I've removed it and updated the setup with a link to the [Remix docs](https://github.com/remix-run/remix/pull/1754/files#diff-67c5b8c01700db6527b0438aaee80688f741b0f1bb67b2b43936230014b029f8R8), user can choose their own implemenation path that way 👍

sergiodxa:
> @sergiodxa thanks for the review and feedback, I've taken a crack at the general updates but wanted to better understand the `resource route` comment a bit better before doing anything there. From what I understand I could easily create an `api/character.ts` route with no component effectively giving me an API which any client can consume, example below
> 
> Simple API endpoint `/api/character`
> Then I can query that endpoint in the loader, but I'm not fully understanding how to accomplish the bit here ` ideally also have a route showing how to run the query against the schema and resolvers without doing a request.` specifically `without doing a request` 🤔

What I meant is that the example could have three parts

1. A `routes/graphql.ts` resource route with a GraphQL endpoint, there you can expose any mock data, the idea is to show that you can use Remix resource routes to build a GraphQL API
2. One or more UI routes fetching data from a GraphQL endpoint, it could be the same or another public one, this shows how to fetch GraphQL from a Remix loader/action
3. One or more UI routes using the same GraphQL schema and resolvers as the API endpoint but calling them directly with the query, this shows how to avoid doing the request while still using GraphQL so you can call your own API without doing an HTTP request

visormatt:
@sergiodxa alrighty I think I've checked all the boxes above but have a look and let know if I've missed anything. Thanks for that follow up comment, not sure why I didn't think of that initially. 

1. I've opted for two resource routes in an `api` directory
2. The `index` route which demonstrates calling the resource directly
3. The `$id` route fetches from the API (like a mobile client might) 
4. Lastly the `error` route uses a raw fetch within the loader, this wasn't listed but I wanted to show the GraphQL error scenario

examples/graphql-api/README.md
```diff
@@ -0,0 +1,30 @@
...
+  - See the [Remix docs](/docs/guides/envvars) for some suggested patterns
+- Run `$ npm install`
+
+## Preview
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+[![Open in CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/graphql-api)
+
+## Example
```
kentcdodds, line 17: Could this bit explain what's demonstrated in this example and what files to play close attention to?

kentcdodds (changes requested):
Looks pretty good to me! Just one thing on my end.

examples/graphql-api/README.md
visormatt: Thanks Kent, I've taken a stab at updating the README a bit, fresh eyes are appreciated though!

kentcdodds (approved):
Stellar! Thank you :)

PR #1764, opened 2022-02-01
feat: update esbuild

jacob-ebey:
fix: move node polyfills to last plugin for non node platforms

This resolves the issue of not being able to import node built-ins in a route module.


sergiodxa:
Will this allow libraries using Node built-in packages like path or buffer work client-side too?

jacob-ebey:
> Will this allow libraries using Node built-in packages like path or buffer work client-side too?

It will but I discourage this practice in general. I will not be advertising this, but some packages that did not work before may now.

sergiodxa:
> > Will this allow libraries using Node built-in packages like path or buffer work client-side too?
> 
> It will but I discourage this practice in general. I will not be advertising this, but some packages that did not work before may now.

I also don't like packages doing that, but multiple people already had issues with that not working and it's hard to convince them to open an issue on the library repo to ask them to don't expect webpack like behavior where that works.

mcansh (approved):

mjackson (approved):

Issue #972, opened 2021-12-09
[Bug]: `Request.formData()` never settles on larger payloads


Trve:
### What version of Remix are you using?

1.0.6

### What version of Node are you using? Minimum supported version is 14.

14.17.5

### Steps to Reproduce

On the client side:
```
const submit = useSubmit();
const data = new FormData();
const data= 'abc'; // some string with 105kB length
data.append('someKey', data);
submit(data, {
	method: 'post',
	action: '/action',
});
```
On the server side:
```
export let action: ActionFunction = async ({ request }) => {
	let formData = await request.formData(); // never settles
	// this code is never reached
	return json('saved!');
};
```

### Expected Behavior

The `request.formData()` should settle with the correctly parsed form data. This should happen within a reasonable amount of time (probably some milliseconds).
This should work with any payload size which is allowed by the specs.

### Actual Behavior

The Promise returned by the `request.formData()` method never settles if the sent payload is (in my case) 105kB or larger. In my case it worked with a payload of ~3kB. I did not test any other payload sizes. But it must be something in between 3 and 105 kilobytes.

mmarkell:
+1, I have the same issue

mmarkell:
`request.formData()` is hanging for me with a 45KB payload

mmarkell:
I'm not sure what it will break in the process, but it seems like a workaround for this issue is the edit @remix-run/node/fetch.js

```
   async formData() {
-    let body = await this.clone().text();
+    const body = await this.text()
+    // Michael -- I don't know what this will break but...it works for now
+    // let body = await this.clone().text();
     return new formData.RemixFormData(body);
   }
```

jraoult:
Pretty sure this PR https://github.com/remix-run/remix/pull/1120 fixes this issue @mjackson. you might want to close it then.

jraoult:
Updated patch for 1.1.3:

```
diff --git a/fetch.js b/fetch.js
index 5ea85d9e968806a942eea27d1e0ce706d4cb7af6..c0b1173e6c1e66eb2b48779e3a44b357fc8650dd 100644
--- a/fetch.js
+++ b/fetch.js
@@ -104,7 +104,7 @@ class NodeRequest extends nodeFetch.Request {
   }
 
   clone() {
-    return new NodeRequest(this);
+    return this;
   }
 
 }
```

[...]

Issue #1164, opened 2021-12-20
[Bug]: Larger files are not readable via UploadHandler


Ehesp:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [X] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.1.1

### What version of Node are you using? Minimum supported version is 14.

16.1.0

### Steps to Reproduce

Upload a file using the `Form` component.

Add a custom `UploadHandler`:

```ts
export const action: ActionFunction = async ({ request }) => {
  const handler: UploadHandler = async ({ name, stream }) => {
    if (name === 'foo') {
      const buff = await new Promise<Buffer>(async (resolve, reject) => {
        const chunks: any = [];

        stream.on("readable", () => {
          let chunk;
          while (null !== (chunk = stream.read())) {
            console.log('chunk...', chunk);
            chunks.push(chunk);
          }
        });

        stream.on("end", () => {
          console.log('end');
          return resolve(Buffer.concat(chunks));
        });
      });
    }
  }

  const form = await unstable_parseMultipartFormData(request, handler);
  // ..
};
```

### Expected Behavior

The Readable stream should trigger the `end` event once consumed.

### Actual Behavior

For small files (e.g. 10kb) the stream is handled correctly, and the `end` event is called. However, for larger files (90kb fails), a single chunk is logged, but then nothing else happens. The `end` event is never called.

Am I doing something wrong? I just cannot figure this out. The handling of the stream example is even taken from the nodejs docs.

ralrom:
I'll add the comment here because it's more relevant to this issue:

I'm having issues uploading larger images. After searching on the Discord server, I found this comment from cmwoodall:

[...]

PR #1766, opened 2022-02-01
fix: do not clone request for actions

jacob-ebey:
This is to avoid a stream halting bug in node-fetch


mcansh (approved):

jraoult:
Would probably fix https://github.com/remix-run/remix/issues/972 ?

kentcdodds (approved):
This makes sense to me, but I'd prefer @mjackson to merge since he may know more of the implications here.

mjackson:
This makes sense to me too, since we the main reason to clone a request is so that the body can be used in multiple handlers. But we only ever call a single handler on action requests, so we should be ok to not clone it here.

I think @jraoult is right, this should fix #972 as well. But I'd love to see some tests around handling large file uploads to make sure that this indeed fixes the problem.

kentcdodds:
Agreed! A test would be great 👍

integration/file-uploads-test.ts
```diff
@@ -0,0 +1,122 @@
...
+        "app/routes/file-upload.jsx": js`
+          import { Form, unstable_parseMultipartFormData as parseMultipartFormData, useActionData } from "remix";
+          import { uploadHandler } from "~/fileUploadHandler";
+
+          export let action = async ({ request }) => {
+            try {
+              let formData = await parseMultipartFormData(request, uploadHandler);
+
+              let file = formData.get("file");
+              console.log({ file });
```
kentcdodds, line 37:
```suggestion
```

kentcdodds (approved):
Killer 👍

arnaudambro:
related to #1164

Issue #1753, opened 2022-02-01
[Feature]: Return arbitrary Response object directly from loader


colinclerk:
### What is the new or updated feature that you are suggesting?

We would like to return an arbitrary Response object directly from a loader.

### Why should this feature be included?

Our use case is that we're building an authentication library and we have special handling for the 401 Unauthorized case. We'd like to provide our handling directly through the loader, instead of asking developers to also modify the CatchBoundary.

Beyond our use case, it was a little confusing to learn that `redirect()` or `json()` Responses can be returned directly from a loader, but arbitrary Response objects cannot. We'd love a convention like returned Response objects are passed directly to the browser, while thrown Response objects go through the CatchBoundary.

sergiodxa:
You can 100% return a new Response object with any status code and body from a loader.

If the route has a default export the data will be JSON.parse-d so your data should be compatible with that, for example don’t return a Buffer.

Also the json helper function can receive a status code or ResponseInit as second argument and there you can make the response created by it to be an unauthorized response.

colinclerk:
Hi @sergiodxa - thanks for your help!

Our route has a default export and we want to return an HTML response directly from the loader under certain conditions.  We haven't been able to make this work - perhaps because of the JSON.parse restriction you mention?  If the capability already exists we'd love to use it!

Here's failing loader:
```js
export const loader: LoaderFunction = async () => {
  return new Response("<!DOCTYPE html><html><body>Test</body></html>", {
    status: 401,
  });
};
```

The error we receive is somewhat cryptic (it results in a 500 in the browser):
```
TypeError: Cannot read properties of undefined (reading 'toString')
    at Index (/Users/colinsidoti/my-remix-app/build/index.js:443:14)
    at processChild (/Users/colinsidoti/my-remix-app/node_modules/react-dom/cjs/react-dom-server.node.development.js:3353:14)
    at resolve (/Users/colinsidoti/my-remix-app/node_modules/react-dom/cjs/react-dom-server.node.development.js:3270:5)
    at ReactDOMServerRenderer.render (/Users/colinsidoti/my-remix-app/node_modules/react-dom/cjs/react-dom-server.node.development.js:3753:22)
    at ReactDOMServerRenderer.read (/Users/colinsidoti/my-remix-app/node_modules/react-dom/cjs/react-dom-server.node.development.js:3690:29)
[...]

PR #1782, opened 2022-02-02
feat: Add support for returning arbitrary Responses directly from loader

colinclerk:
Adds support for https://github.com/remix-run/remix/issues/1753

This PR allows developers to return an arbitrary Response object directly from the loader during a "document" request. Developers can opt-in to the functionality by adding a `X-Remix-Skip-Render` header to their Response object, which was chosen to mimic the special behavior of the `X-Remix-Catch` header.

The special header can be added to any returned Response object (which would skip the default export render), or any thrown Response object (which would skip the boundary render).

This feature is purely additive.  If developers try to **return** a Response object from a loader today, Remix will error out unless that Response object is a redirect.  If developers try to **throw** a Response object from a loader today, the CatchBoundary will be rendered and the Response's body will be ignored.  This feature provides developers with a mechanism to skip the default or boundary render and return a Response object directly.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -25,8 +25,9 @@
 - christophgockel
 - clarkmitchell
 - codymjarrett
-- coryhouse
+- colinclerk
 - confix
+- coryhouse
 - craigglennie
 - crismali
 - danielweinmann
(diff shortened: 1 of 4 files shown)


ryanflorence:
Thanks for taking the time to improve Remix! Before you spend time implementing a feature, we'd like to discuss it first 🙂 I opened a discussion from the referenced issue.

https://github.com/remix-run/remix/discussions/1880

PR #1786, opened 2022-02-03
Test submit button with SVG element inside

jorupp:
PR with test case for bug report

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -36,27 +36,56 @@ beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/index.jsx": js`
-        import { json, useLoaderData, Link } from "remix";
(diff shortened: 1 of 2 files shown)


ryanflorence:
Awesome, thank you. The problem is right here:

https://github.com/remix-run/remix/blob/dev/packages/remix-react/components.tsx#L878

It's unclear to me why we have that check in the first place, probably a typescript thing, I think we can probably use `instanceof Element` instead?

If you (or anybody else) wants to tackle this, I'd add it to `integration/form-test.ts`.

ryanflorence:
also let's not have the `spawn` changes in a PR to fix this bug, for that I want it to just log when there's an error, not always.

edmundhung:
There is actually a fix on https://github.com/remix-run/remix/pull/1240/
But I guess the tests might be outdated and need to be updated 😅

ryanflorence:
thanks, I just dropped an svg into the new form tests in @edmundhung's PR to (indirectly) test this!

Issue #1691, opened 2022-01-28
[Bug]: useFetcher with method post is not updating its state when there a redirection in the action handler


rperon:
### Which Remix packages are impacted?

- [X] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [X] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.1.3

### What version of Node are you using? Minimum supported version is 14.

v16.13.0

### Steps to Reproduce

When creating a form with useFetcher, the state is not getting if the action is a redirect.

1. npx create-remix@latest 
2. create a form with useFetcher

Sample code :
```js
import { ActionFunction,  Form,  LoaderFunction,  redirect,  useFetcher,  useTransition} from "remix";

function timeout(ms: number) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}

export const action: ActionFunction = async ({ request }) => {
  await timeout(300);
  return redirect("/");
};

export const loader: LoaderFunction = async ({ request }) => {
  await timeout(300);
  return {};
};

export default function Index() {
  const fetcher = useFetcher();
  const transition = useTransition();

  const formIsSubmitting =
    transition && transition.submission?.formData.get("source") === "form";

  return (
    <div style={{ fontFamily: "system-ui, sans-serif", lineHeight: "1.4" }}>
      {formIsSubmitting ? transition.state : "idle"}
      <Form method="post">
        <input type="hidden" name="source" value="form" />
        <button type="submit" name="_action" value="add">
          Submit
        </button>
      </Form>
      {fetcher.state}
      <fetcher.Form method="post">
        <input type="hidden" name="source" value="fetcher" />
        <button type="submit" name="_action" value="add">
          Submit
        </button>
      </fetcher.Form>
    </div>
  );
}
```

### Expected Behavior

In this example a added a timeout to watch when the state are getting updated.

The first button use a `Form` and has the expected Behavior: idle => submitting => loading => idle .

I expect the form created with `useFetcher` to behave the same way.

### Actual Behavior

The form created with useFetcher stay stuck at the submitting state.

[...]

PR #1794, opened 2022-02-03
Add failing test for useFetcher statue issue.

rperon:
PR with test case for bug report #1691


ryanflorence:
@mcansh I know you're goofing around in the transition code, you want to get this test into it's proper place and then fix the bug?

mcansh:
thanks for the test case @rperon! https://github.com/remix-run/remix/pull/1875

rperon:
Thanks for your work !

PR #1801, opened 2022-02-04
chore: Create `release` script

chaance:
This PR introduces a `release` script to simplify the release process based on our branching strategy. I will need to stress test a few of the processes during the next release process, but I first want to make sure the flow is correct and accounts for the various types of releases we work with.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -8,6 +8,8 @@ const {
   prompt
 } = require("./utils");
 
+const releaseTypes = ["patch", "minor", "major"];
+
 run(process.argv.slice(2)).then(
   () => {
     process.exit(0);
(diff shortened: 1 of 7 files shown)


scripts/release.js
```diff
@@ -0,0 +1,357 @@
...
+    throw Error("Releases should be created from the dev branch.");
+  }
+  return branch;
+}
+
+/**
+ * @param {string} branch
+ */
+function ensureReleaseBranch(branch) {
+  let version = getVersionFromReleaseBranch(branch);
```
ryanflorence, line 326: we'll want to make sure this is *the latest release branch* semver has some helpers that can do that I think.

PR #1806, opened 2022-02-05
transpileModules config

cj:
This will allow transpiled modules just like https://www.npmjs.com/package/next-transpile-modules you can now just add transpileModules to your remix.config.js.

```javascript
/**
 * @type {import('@remix-run/dev/config').AppConfig}
 */
module.exports = {
  appDirectory: 'app',
  browserBuildDirectory: './public/build',
  publicPath: '/build/',
  serverBuildDirectory: './server/build',
  devServerPort: 8002,
  transpileModules: ['ui', 'lib', 'env'],
}
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -25,6 +25,7 @@
 - chenc041
 - christianhg
 - christophgockel
+- cj
 - clarkmitchell
 - codymjarrett
 - coryhouse
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -296,7 +296,7 @@ async function createBrowserBuild(
   };
   for (let id of Object.keys(config.routes)) {
     // All route entry points are virtual modules that will be loaded by the
-    // browserEntryPointsPlugin. This allows us to tree-shake server-only code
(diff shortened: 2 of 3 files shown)


cj:
@ryanflorence this was not intended for node_modules, but if you are using a monorepo (workspaces, lerna, turbo etc...) and needed to include your other packages with your deployment i.e. your UI components.  It looks like someone else opened up a pull request with the same approach but better documentation https://github.com/remix-run/remix/pull/1839 😄

liho00:
> Thanks for taking the time to improve Remix!
> 
> Unfortunately we'd rather not transpile modules from node_modules. The module ecosystem has gotten incredibly dependent on webpack-isms (like automatically polyfilling node built-ins like "http", which is 250kb of code that sneaks into your bundle!) and folks expecting a bundler to figure out how to import a css file.
> 
> We don't want to enable the same incompatibility issues by enabling people to ship "remix-isms" or "esbuild-isms" in their node_modules.
> 
> Thanks again though :)

Please merge this pr to allow people have the option to choose to transpile the modules

Issue #389, opened 2021-11-17
WebSocket API


sergiodxa:
From Discord discussion:

````md
sergiodxa — hoy a las 11:17
actually I would probably prefer this
```ts
// use this to get the initial data
export async function loader({ request }) {
  let data = await getData(request)
  return json(data)
}

// use this to mutate and broadcast data to WS subscribers
export async function action({ request, broadcast }) {
  let formData = await request.formData();
  let something = await doSomething(formData);
  await broadcast(something); // this should pass data from action to the socket
  return redirect("path")
}

// use this to get the data from the action and send it to subscriber
export async function socket({ data }) { // this data came from the action broadcast
  let actualData = await getActualData(data) // here you can use the action data to get more things if needed
  return actualData; // and return what you want to send to users
}

export default function Route() {
  let initialData = useLoaderData();
  let { data } = useSocketData(); // the data here should probably be an array, every return from socket should append new items
  // ...
}
```

Alex 🚀 — hoy a las 11:23
Something worth considering is whether you need subscription filters too. Route-by-route subscriptions help, since you could automatically be unsubscribed from a specific route's publishes when you leave that route. 

But then there might be circumstances where only specific clients want to get the subscription publishes based on parameters they provide to the server somehow.
Also, actions from one route need the ability to trigger broadcasts on another route

sergiodxa — hoy a las 11:26
what I was thinking is the broadcast from an action only trigger the socket function on the same file/route, so let's say you have a route /chat/:chatId/messages/new, there you have an action to create new chat messages for a specific chat, you can do useSocketData("/chat/1/messages/new") and subscribe to the data broadcasted from the action on that route, so you subscribe to new messages on the chat 1 basically 
if you don't pass anything to useSocketData it could try to subscribe to the WS on the same route of the component
this way, you could have routes with only socket and action if you want, the action do something and the socket let users subscribe to know when that something happens
another option is that broadcast could receive the data to broadcast and a path, then the socket function on that path can receive the data
so you can have socket and action in different routes

garand — hoy a las 11:29
Something to consider if the data from useSocketData is an array... I have an app using websockets and it's sending a JSON object every 200ms, and I have the app open for 8 hours a day. Storing that would fill up memory pretty fast I think.

sergiodxa — hoy a las 11:29
and a final idea, the socket function may not be required at all, if it's not defined whatever you pass to broadcast from action could be what subscribers will receive, if you define one you could use it to get more data or to reduce the data you sent

sergiodxa — hoy a las 11:30
[...]

PR #1811, opened 2022-02-05
docs(examples): add Firebase Auth + Firestore example

ccssmnn:
This PR adds a Firebase example with Auth-flow and minimal to-do functionality based on Firestore.

- `app/server/firebase.server.ts` initializes the client & server SDKs
- `app/server/auth.server.ts` implements auth utilities to sign-in users or to protect routes and actions
- `app/server/db.server.ts` shows how to utilize FirestoreDataConverters to have TypeScript support when dealing with Firestore


examples/firebase-auth-firestore/app/server/auth.server.ts
```diff
@@ -0,0 +1,56 @@
...
+  const idToken = await user.getIdToken();
+  const expiresIn = 1000 * 60 * 60 * 24 * 7; // 1 week
+  const sessionCookie = await auth.server.createSessionCookie(idToken, {
+    expiresIn,
+  });
+  return sessionCookie;
+};
+
+export const signUp = async (name: string, email: string, password: string) => {
+  const { user } = await createUserWithEmailAndPassword(
```
Ehesp, line 49: This could still happen on the client, and provide a id token. Not sure what happens on the Firebase JS SDK if someone is already signed in (since it would store this state on the server) and then another sign in is attempted.

ccssmnn: They are [recommending this approach](https://github.com/firebase/firebase-admin-node/issues/1141#issuecomment-765627139) for server-side sign in. I assume it's fine to create the user on the server as well.

ccssmnn: I've changed the signUp method to use `firebase-admin`.

dpcesx:
This is awesome! Thanks, @ccssmnn I saw your answer to a firebase auth question, but this is exactly what I needed.

https://github.com/remix-run/remix/discussions/900

examples/firebase-auth-firestore/app/root.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <meta charSet="utf-8" />
+        <meta name="viewport" content="width=device-width,initial-scale=1" />
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === "development" && <LiveReload />}
```
MichaelDeBoey, line 28:
```suggestion
        <LiveReload />
```
See #1352

MichaelDeBoey (changes requested):

Ehesp:
@davideast Might be a good one for you to review :)

examples/firebase-auth-firestore/app/routes/logout.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export default function Logout() {
+  return (
+    <div>
+      <h1>Logout</h1>
+      <p>Press the button below to log out.</p>
+      <Form method="post">
+        <button type="submit">Logout</button>
+      </Form>
+    </div>
```
dpcesx, line 19: Is this Logout component required? I get an application error without it in my Firebase implementation, but I'm trying to follow the jokes tutorial as best practice.

https://remix.run/docs/en/v1/tutorials/jokes#build-logout-action

ccssmnn: I put the form in the same route for clarity. It should also be possible to have the form in a different route, just like in the jokes example. Can you share your implementation?

I thought about building the jokes tutorial with Firebase as an example, but the [contributing guidelines for examples](https://github.com/remix-run/remix/tree/main/examples#contributing) recommend minimalistic examples. The jokes example should show how to use Remix in a real-world application, while this example should only show how to use Firebase with Remix.

Ehesp:
I already mentioned that in another comment here. I personally don't think it's the best approach to use the client sdk on the server since state is stored within the sdk so I don't know if it'll cause any adverse side effects.

kentcdodds:
Just so long as you're not unnecessarily using the sdk on the client. I haven't had a chance to review this yet, but I've successfully gotten firebase working with Remix with no client-side code required and that's what I'd want for the example if possible.

examples/firebase-auth-firestore/app/routes/logout.tsx
dpcesx: You did a stellar job. I finally traced my issues down to [Module Constraints](https://remix.run/docs/en/v1/guides/constraints). The conventions you have followed using `.server.ts` and `app/sessions.ts` make perfect sense and fit the remix recommendations. 

I was able to successfully merge the Jokes example with this example. The hardest part was that before this conversation I was chasing seemly cryptic application error messages. I can post the combined version if you want to see it.

Thanks for the help!

dpcesx:
@kentcdodds I'm still playing with my implementation of firebase/remix. I've seen [one of your youtube videos on firebase](https://youtu.be/bC4ZFe2CY00), but if you have others I can try and test some other approaches as well.

ccssmnn:
There is more to Firebase where things might need to be done differently when doing them the Remix way:

- Storage
  - Uploads ([MultipartFormData](https://remix.run/docs/en/v1/api/remix#unstable_parsemultipartformdata-node))
  - Downloads (e.g. Images, pass them through Remix vs. storing them under a secure but public path)
- Database listeners / real-time updates (#389 & #1996)
- sign-in with Google, Apple, ... (sprinkle in Firebase Client SDK?)
- Cloud Functions / reacting to database events (handle them entirely in Cloud Functions or use Cloud Functions to trigger APIs implemented in Remix?)
- Hosting (Firebase Hosting + Cloud Run seems to be the best way right now)

I think creating one example for each of these tasks will bloat the example directory. One well-structured example for all things Remix+Firebase would be wonderful. We could enhance this example, once nice approaches for these features are found.

dpcesx:
> There is more to Firebase where things might need to be done differently when doing them the Remix way:
> 
> * Storage
>   
>   * Uploads ([MultipartFormData](https://remix.run/docs/en/v1/api/remix#unstable_parsemultipartformdata-node))
>   * Downloads (e.g. Images, pass them through Remix vs. storing them under a secure but public path)
> * Database listeners / real-time updates ([WebSocket API #389](https://github.com/remix-run/remix/issues/389) & [Refresh Route Data API #1996](https://github.com/remix-run/remix/discussions/1996))
> * sign-in with Google, Apple, ... (sprinkle in Firebase Client SDK?)
> * Cloud Functions / reacting to database events (handle them entirely in Cloud Functions or use Cloud Functions to trigger APIs implemented in Remix?)
> * Hosting (Firebase Hosting + Cloud Run seems to be the best way right now)
> 
> I think creating one example for each of these tasks will bloat the example directory. One well-structured example for all things Remix+Firebase would be wonderful. We could enhance this example, once nice approaches for these features are found.

I need to work on a few of those items over the next month.

- Storage
  - Uploads ([MultipartFormData](https://remix.run/docs/en/v1/api/remix#unstable_parsemultipartformdata-node))
  - Downloads (e.g. Images, pass them through Remix vs. storing them under a secure but public path)
- Cloud Functions (need more info on what you have in mind for this example repo)

If you have a way you would like to coordinate the effort please let me know. Otherwise, I will keep checking here and update you with any progress.

kentcdodds:
I'm fine having one sizeable example that's the one-stop example for all things firebase 👍

examples/firebase-auth-firestore/app/server/auth.server.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+  return auth.server.getUser(uid);
+};
+
+export const signIn = async (email: string, password: string) => {
+  const { user } = await signInWithEmailAndPassword(
+    auth.client,
+    email,
+    password
+  );
```
jadlr, line 34: I don't think it's safe to use the firebase client sdk on server-side. It is a stateful lib that has the concept of a "current user".

ccssmnn: From [the Firebase Docs](https://firebase.google.com/docs/web/setup) on the client SDK:

> Follow this guide to use the Firebase JavaScript SDK in your web app or as a client for end-user access, for example, in a Node.js desktop or IoT application.

Of course, you can mess up your server implementation. I don't yet see why receiving the user directly from the sign-in or verify-token calls should be unsafe. See [this comment](https://github.com/firebase/firebase-admin-node/issues/1141#issuecomment-765627139).

cjbland: FWIW I've been dealing with this recently myself. I'm not sure if what I did is correct, but I decided to handle the authentication purely on the client, then I POST the token to the server where it is verified and then set the session cookie. I previously did the auth on the server like this example, but then I could not use the federated logins (like Apple, Facebook, Google, etc). Handling it on the client side was the only way I could get the federated logins to work.

ccssmnn: It's totally fine to use Firebase on the client and send the JWT through a cookie for the following requests. This example is to show how to use Firebase in your Remix app without using Firebase on the client. From there on you can enhance with client side js

ccssmnn:
This PR has the needs-response label. What is expected is not clear to me, though. I think, that a full Firebase example needs more Firebase features covered as described [here](https://github.com/remix-run/remix/pull/1811#issuecomment-1046773128) This can still be a useful example for people interested in using Firebase and Remix together.

kentcdodds:
Sounds good to me 👍 Thanks for following up!

mocon:
This was SUPER helpful. Thanks everyone! One small change I had to make was to destructure the imports in `firebase.server.ts`:

```ts
import {
  getApps as getServerApps,
  initializeApp as initializeServerApp,
  cert as serverCert,
} from 'firebase-admin/app'
import { getApps as getClientApps, initializeApp as initializeClientApp } from 'firebase/app'
import { getAuth as getServerAuth } from 'firebase-admin/auth'
import { getAuth as getClientAuth } from 'firebase/auth'

if (getClientApps().length === 0) {
  initializeClientApp(JSON.parse(process.env.CLIENT_CONFIG as string))
}

if (getServerApps().length === 0) {
  initializeServerApp({
    credential: serverCert(JSON.parse(process.env.SERVICE_ACCOUNT as string)),
  })
}

export const auth = {
  server: getServerAuth(),
  client: getClientAuth(),
}
```

examples/firebase-auth-firestore/app/server/auth.server.ts
penx: > Follow this guide to use the Firebase JavaScript SDK in your web app or as a client for end-user access, for example, in a Node.js desktop or IoT application.


> for end-user access

i.e. not for a server handling requests for multiple users

> in a Node.js desktop

i.e. where there would only be a single user at a time

> IoT application

Similarly, I think they are inferring where you would have a single user at a time.

As @jadlr mentioned, the client SDK is stateful. When you sign in on the server, this user remains signed in for all requests until the next sign in attempt. Sure, it is likely possible to avoid revealing this security hole to end users by limiting access to the API, but I don't think it's a good idea to put this out there as a recommended approach.


> See [this comment](https://github.com/firebase/firebase-admin-node/issues/1141#issuecomment-765627139).

This is a comment saying that they use it in integration tests. I'm not convinced this means it has been built and tested for the purpose of a multi user enviroment.

penx: I'm in the process of integrating Firebase auth in to a remix app via the REST API. I'll try report back on how I get on. I have this working via the Firebase emulator at the moment but not in production. I saw [issues with rate limiting](https://twitter.com/1fiws/status/1521426483037839362) being mentioned on Twitter which I think would apply to both the REST and client SDK approaches.

johnpolacek:
@mocon I was not able to get the Remix Firebase Auth example to work until I applied your update to `firebase.server.ts`. You should PR that sir :)

examples/firebase-auth-firestore/app/server/auth.server.ts
ccssmnn: @penx thank you for investigating this further. Indeed the firebase sdk's dont seem to be intended to be used this way which may or may not cause risks using this in production. I didn't have the time to investigate the REST APIs. I'm really looking forward to see the results of your investigations!

penx: I have auth working with the Firebase REST API now. I plan to work on some Firebase PRs for Remix in the coming weeks, including updating the example to use the REST API.

I think I've also figured a way to get around the rate limiting when used server side so will include that if it works, as far as I'm aware the rate limiting effects both the REST API and the client side SDK when used from the same IP.

My plan is to authenticate on the client via the REST API if the JavaScript has loaded, and fall back to doing it in the action JavaScript has not loaded or failed. Then when the action received a request, it would either:

- contain username/password => fallback to server side login via rest API
- contain response from Firebase after a client side login => server side validate the session before saving in to Remix session (as validation would not be subject to rate limiting)

PR #1816, opened 2022-02-06
feat(remix-server-runtime): log time taken in `action`s/`loader`s

ryankshaw:
This both logs the amount of time taken for each loader to the console
as well as sets a server timing header so you can see it in the browser

I just took a rough shot at a first draft of doing this. I don’t know
if this is something you guys would want to do. We’ll want to make a
way for people to opt out but I didn’t know what would be the best
way to check for that (I talk about it in my screencast)

To test, 
* go to a page that has a few loaders. 
* Go to the network panel
* hover over the waterfall for a request served by remix
* it should show the amount of time it spent in each loader

I made a little screencast to explain it more:
https://drive.google.com/file/d/1PSlg9kAstP42vGyIggNJ_TCBji0HbC5k/view

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -137,6 +137,7 @@
 - RossMcMillan92
 - rphlmr
 - ryanflorence
+- ryankshaw
 - sdavids
 - sean-roberts
 - sergiodxa
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -1,5 +1,6 @@
 import type { RouteMatch } from "./routeMatching";
 import type { ServerRoute } from "./routes";
+import type { instrument } from "./server";
 import { json, isResponse, isRedirectResponse } from "./responses";
 
 /**
@@ -16,11 +17,13 @@ export type AppData = any;
 export async function callRouteAction({
   loadContext,
   match,
-  request
+  request,
+  instrument
 }: {
   loadContext: unknown;
   match: RouteMatch<ServerRoute>;
   request: Request;
+  instrument: instrument;
 }) {
   let action = match.route.module.action;
 
@@ -34,11 +37,16 @@ export async function callRouteAction({
 
   let result;
   try {
-    result = await action({
-      request: stripDataParam(stripIndexParam(request.clone())),
-      context: loadContext,
-      params: match.params
-    });
+    result = await instrument(
+      "action",
+      match.route.id,
+      () =>
+        action && action({
+          request: stripDataParam(stripIndexParam(request.clone())),
+          context: loadContext,
+          params: match.params
+        })
+    );
   } catch (error: unknown) {
     if (!isResponse(error)) {
       throw error;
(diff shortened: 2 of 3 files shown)


mikeybinnswebdesign:
This may not be what you're looking for, but I just implemented server timings into one of my loaders using [Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance) and [Server Timings API](https://web.dev/custom-metrics/?utm_source=devtools#server-timing-api)
This allows me to not only get the time from start to finish for the loader, but also steps in-between, and then it displays in devtools.
WARNING: Be careful exposing server timing to end users. This can be used by attackers to break into your system. My function specifically filters this out in production, and yours should too if you write your own.

Here's my code:
```ts
export let loader: LoaderFunction = async () => {
	performance.mark("start");
	// Do local database request to get "parks" here
	performance.mark("parks_from_db");
	if (parks.length === 0) {
		performance.mark("cache_miss");
		// Do external database fetch to get "parks" here
	}
	performance.mark("parks_from_sf");
	performance.measure("get_parks_from_db_cache", "start", "parks_from_db");
	performance.measure("get_parks_from_sf", "parks_from_db", "parks_from_sf");
	performance.measure("total", "start", "parks_from_sf");
	const cache_miss = performance
		.getEntriesByType("mark")
		.find((mark) => mark.name === "cache_miss");
	const perf_entries = performance.getEntriesByType("measure");

	let server_timing: string;
	if (cache_miss) {
		server_timing = "cache_miss";
	} else {
		server_timing = "cache_hit";
	}
	for (let entry of perf_entries) {
		server_timing = `${server_timing}, ${entry.name};dur=${entry.duration}`;
	}
	performance.clearMeasures();
	return json(parks, {
		headers: {
			"Server-Timing": `${server_timing}`,
		},
	});
};

export let headers: HeadersFunction = ({
	loaderHeaders,
	actionHeaders,
	parentHeaders,
}) => {
	const server_timings = getAllServerTimingInfo({
		loaderHeaders,
		actionHeaders,
		parentHeaders,
	});
	// This function is from the Remix docs if you're interested
	const cache_control = getLeastAggressiveCacheControl({
		loaderHeaders,
		parentHeaders,
	});
	return new Headers({
		...Object.fromEntries(server_timings),
		...Object.fromEntries(cache_control),
	});
};
export function getAllServerTimingInfo({
	loaderHeaders,
	actionHeaders,
	parentHeaders,
}: {
	loaderHeaders: Headers;
	actionHeaders: Headers;
	parentHeaders: Headers;
}) {
	const headers = new Headers();
	if (process.env.NODE_ENV === "production") {
		// Return no timings info for security reasons.
		return headers;
	}
	const loader_server_timing = loaderHeaders.get("Server-Timing");
	if (loader_server_timing) {
		headers.append("Server-Timing", loader_server_timing);
	}
	const action_server_timing = actionHeaders.get("Server-Timing");
	if (action_server_timing) {
		headers.append("Server-Timing", action_server_timing);
	}
	const parent_server_timing = parentHeaders.get("Server-Timing");
	if (parent_server_timing) {
		headers.append("Server-Timing", parent_server_timing);
	}
	return headers;
}
```

PR #1821, opened 2022-02-06
docs(examples): link labels with their fields in examples, to enhance accessibility

mkrtchian:
I have found 3 examples in which `<label>` elements were not linked to their fields, and linked them correctly.

I have also removed a `pattern` attribute on an `<input type="number">` as it is not supported on that type (cf. [the MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern)).

As for the modifications in `examples/client-side-validation/package-lock.json`, it's simply the result of running `npm install` in that example folder.


ryanflorence:
🙏

PR #1822, opened 2022-02-06
feat(cli): exit with code 1 when remix build fails

gmaliar:
When builds fail remix does not exit with an error code which makes it hard to catch build failures in CI environments

https://github.com/remix-run/remix/blob/197f570841e4813b2005d1ae007523c09df3f5c5/packages/remix-dev/compiler.ts#L307-L314

Example:
```sh
❯ yarn create remix
❯ cd ./my-remix-app
❯ echo -e "import doesntExists from \"doesnt-exists\";\ndoesntExists();\n$(cat app/routes/index.tsx)" > app/routes/index.tsx
❯ npx remix build

npx remix build
The path "doesnt-exists" is imported in app/routes/index.tsx but doesnt-exists is not listed in your package.json dependencies. Did you forget to install it?
 > node_modules/@web-std/file/tsconfig.json:6:24: warning: Non-relative path "packages/file/src/lib.js" is not allowed when "baseUrl" is not set (did you forget a leading "./"?)
    6 │       "@web-std/file": ["packages/file/src/lib.js"]
      ╵                         ~~~~~~~~~~~~~~~~~~~~~~~~~~

 > app/routes/index.tsx:1:25: error: Could not resolve "doesnt-exists" (mark it as external to exclude it from the bundle)
    1 │ import doesntExists from "doesnt-exists";
      ╵                          ~~~~~~~~~~~~~~~

Build failed with 1 error:
app/routes/index.tsx:1:25: error: Could not resolve "doesnt-exists" (mark it as external to exclude it from the bundle)
Built in 571ms

❯ echo $?
0
```


gmaliar:
accidentally closed this by force pushing to my remix dev branch https://github.com/remix-run/remix/pull/1527

packages/remix-dev/cli/commands.ts
```diff
@@ -61,7 +61,11 @@ export async function build(
 
   let start = Date.now();
   let config = await readConfig(remixRoot);
-  await compiler.build(config, { mode: mode, sourcemap });
+  try {
+    await compiler.build(config, { mode: mode, sourcemap });
```
ryanflorence, line 65: looks like we intended to have this code pass `onBuildFailure`, and then we can `process.exit(1)` there instead of a try catch

ryanflorence: ah, also we use a defaultOnBuildFailure handler, I think the process.exit(1) needs to go there.

@mjackson can you help us understand the intent here? Right now it just does `console.error` in the default build failure handler, seems like that's also the spot we'd want to `process.exit(1)` as well.

gmaliar: I think that the default on build error is a great place to exit with an error code.
Would you like me to change?

gmaliar:
@ryanflorence we actually got it wrong.
@jacob-ebey already added error handling for the cli here:
https://github.com/remix-run/remix/blob/main/packages/remix-dev/cli.ts#L66

The only problem with it, is that the `defaultBuildFailureHandler` doesn't throw anything so the error just get swallowed and it doesn't exit for some use-cases.

Now there's another thing we got wrong, the default build failure handler is also used in watch, if it would `process.exit(1)` on watch, it's counterproductive to the watch intent, so it's not the correct spot to implement the exit procedure.

Instead, I've created a different `onBuidlFailure` that throw an error but still formats the failure as the default handler.
I propose an internal change in the `compiler.ts` file that extracts the formatting into a function `formatBuildFailure` that the `defaultOnBuildFailure` will use to format errors and also the different `onBuildFailure` can use when is invoked from the cli.

gmaliar:
@ryanflorence @mjackson can we get this merged?

gmaliar:
@mjackson @kentcdodds @ryanflorence this has been ready for a while, can we get this merged?

kentcdodds (approved):
This looks good to me. I'll ping Michael and Ryan about it.

PR #1835, opened 2022-02-08
Add Netlify-specific context to Remix server actions

anmonteiro:
### Motivation

At Netlify, we're gearing up to launch the Netlify Graph, which allows easy access to third-party APIs via the [Onegraph](https://www.netlify.com/blog/2021/11/17/netlify-acquires-onegraph-a-powerful-graphql-platform-for-connecting-apis-and-services/) acquisition.

### What does this PR do?

This PR adds a custom `getLoadContext` function to the Netlify server template which gets at most 2 pieces of information from the function invocation:

1. `netlifyGraphToken`, which is passed to handlers that fetch remote third-party data (check our docs [here](https://github.com/netlify/labs/blob/main/features/graph/documentation/api-authentication.md)).
2. `clientNetlifyGraphAccessToken`, which is extracted from the `Authorization` request header, if present. This might be in JWT form, containing signed information about the client. This short [video](https://www.youtube.com/watch?v=g8pyzqXAMjE) goes into great detail about the workflows this can allow.


ryanflorence:
@anmonteiro thanks so much for the PRs :)

We've anticipated each host will have unique features like this and we'll never be able to keep track of them all ourselves. Do ya'll want to own the netlify adapter? We'll keep the CLI option but we'll use the netlify adapter from your published package.

ryanflorence:
@anmonteiro can you rebase this on `dev` instead of `main` and we'll get it merged? Thanks!

anmonteiro:
@ryanflorence thx for the review. I just rebased this on dev.


I think we'd be happy to move this to the Netlify org. Tagging @sgrove and @localden here for visibility.

PR #1836, opened 2022-02-08
Add cache-control header to cloudflare pages template

schpet:
this PR adds the _header file described here:
https://developers.cloudflare.com/pages/platform/headers

and sets the same Cache-Control value as the netlify template:

https://github.com/remix-run/remix/blob/712ca2b5b08eb3200b5cb1bbf29e8faa77f556d1/packages/create-remix/templates/netlify/netlify.toml#L18

Todo

- [x] figure out how to run this template to test it out 🤔

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -25,8 +25,8 @@
 - christophgockel
 - clarkmitchell
 - codymjarrett
-- coryhouse
 - confix
+- coryhouse
 - craigglennie
 - crismali
 - danielweinmann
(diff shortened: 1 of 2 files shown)


ryanflorence:
@jacob-ebey this seems good? If you agree go ahead and merge.

sdavids:
https://developers.cloudflare.com/pages/platform/serving-pages#caching-and-performance

> Therefore, you should avoid setting Page Rules or custom caching on your site.

schpet:
@sdavids that reads to me like it applies to content that changes. my understanding is that the stuff in `/build` uses hashed filenames – is there any risk to caching these things longer than cloudflare's 4 hour default?

PR #1837, opened 2022-02-08
fix(remix-dev): omit Resource Routes from client build

### LINKED ISSUE #1828: [Bug]: Resource routes don't get server-side pruning, compiled browser code contains server-side code

1.1.3

16.13.2

### Steps to Reproduce

If you write a [resource route](https://remix.run/docs/en/v1/guides/resource-routes) (a route with server-side methods like `loader` and `action` but no default export and no React components returned), then [server code pruning](https://remix.run/docs/en/v1/guides/constraints#server-code-pruning) is not performed: the browser build contains the server-side code.

```sh-session
$ npx create-remix@latest

R E M I X

💿 Welcome to Remix! Let's get you set up with a new project.

? Where would you like to create your app? ./my-remix-app
? Where do you want to deploy? Choose Remix if you're unsure, it's easy to chang
e deployment targets. Remix App Server
? TypeScript or JavaScript? TypeScript
? Do you want me to run `npm install`? Yes

> postinstall
> remix setup node

Successfully setup Remix for node.

added 439 packages, and audited 440 packages in 25s

143 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
💿 That's it! `cd` into "my-remix-app" and check the README for development and deploy instructions!
$ cd my-remix-app/
$ cat > app/routes/foobar.ts <<EOF
> import { LoaderFunction } from "remix";
> export const loader: LoaderFunction = () => ({foo: 'bar'})
> EOF
$ cat public/build/routes/foobar-ITNQTQOS.js 
import{f as o}from"/build/_shared/chunk-AKSB5QXU.js";o();o();var e=()=>({foo:"bar"});export{e as loader};
```

### Expected Behavior

There should be no browser build output at all for resource routes, or at least the browser build output should not include the server-side `loader` and `action` methods.

### Actual Behavior

See example above.

rossipedia:
If resource routes don't have any client-safe exports, the entire thing gets included as a browser entry point (`export * from ${file}`). This can expose sensitive information.

Fix is to only include routes as entry points that have at least one browser safe export.

Fixes #1828

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -14,7 +14,10 @@ import type { AssetsManifest } from "./compiler/assets";
 import { createAssetsManifest } from "./compiler/assets";
 import { getAppDependencies } from "./compiler/dependencies";
 import { loaders } from "./compiler/loaders";
-import { browserRouteModulesPlugin } from "./compiler/plugins/browserRouteModulesPlugin";
+import {
+  browserRouteModulesPlugin,
+  isResourceOnlyRoute
+} from "./compiler/plugins/browserRouteModulesPlugin";
 import { emptyModulesPlugin } from "./compiler/plugins/emptyModulesPlugin";
 import { mdxPlugin } from "./compiler/plugins/mdx";
 import type { AssetsManifestPromiseRef } from "./compiler/plugins/serverAssetsManifestPlugin";
@@ -191,7 +194,7 @@ export async function watch(
     // should probably blow as it's not really recoverable.
     let browserBuildPromise = browserBuild.rebuild();
     let assetsManifestPromise = browserBuildPromise.then(build =>
-      generateAssetsManifest(config, build.metafile!)
+      generateAssetsManifest(config, options, build.metafile!)
     );
 
     // Assign the assetsManifestPromise to a ref so the server build can await
@@ -286,7 +289,7 @@ async function buildEverything(
   try {
     let browserBuildPromise = createBrowserBuild(config, options);
     let assetsManifestPromise = browserBuildPromise.then(build =>
-      generateAssetsManifest(config, build.metafile!)
+      generateAssetsManifest(config, options, build.metafile!)
     );
 
     // Assign the assetsManifestPromise to a ref so the server build can await
@@ -337,8 +340,14 @@ async function createBrowserBuild(
     // All route entry points are virtual modules that will be loaded by the
     // browserEntryPointsPlugin. This allows us to tree-shake server-only code
     // that we don't want to run in the browser (i.e. action & loader).
-    entryPoints[id] =
-      path.resolve(config.appDirectory, config.routes[id].file) + "?browser";
+
+    // If any route entry point has *zero* browser safe exports (ie: only action/loader etc)
+    // then omit it entirely from the browser build
+    if (!(await isResourceOnlyRoute(config, id))) {
+      // We can include it as an entry point
+      entryPoints[id] =
+        path.resolve(config.appDirectory, config.routes[id].file) + "?browser";
+    }
   }
 
   return esbuild.build({
@@ -457,16 +466,21 @@ async function createServerBuild(
 
 async function generateAssetsManifest(
   config: RemixConfig,
+  options: Required<BuildOptions>,
   metafile: esbuild.Metafile
 ): Promise<AssetsManifest> {
-  let assetsManifest = await createAssetsManifest(config, metafile);
+  let assetsManifest = await createAssetsManifest(config, options, metafile);
   let filename = `manifest-${assetsManifest.version.toUpperCase()}.js`;
 
   assetsManifest.url = config.publicPath + filename;
 
   await writeFileSafe(
     path.join(config.assetsBuildDirectory, filename),
-    `window.__remixManifest=${JSON.stringify(assetsManifest)};`
+    `window.__remixManifest=${JSON.stringify(
+      assetsManifest,
+      undefined,
+      options.mode !== BuildMode.Production ? 2 : undefined
+    )};`
   );
 
   return assetsManifest;
--- a/contributors.yml
+++ b/contributors.yml
@@ -146,6 +146,7 @@
 - pyr0gan
 - real34
 - reggie3
+- rossipedia
 - RossJHagan
 - RossMcMillan92
 - rphlmr
--- a/integration/compiler-test.ts
+++ b/integration/compiler-test.ts
@@ -5,6 +5,8 @@ describe("compiler", () => {
   let fixture: Fixture;
   let app: AppFixture;
 
+  const RESOURCE_ROUTE = "routes/resource/route";
+
   beforeAll(async () => {
     fixture = await createFixture({
       files: {
@@ -19,6 +21,7 @@ describe("compiler", () => {
             return <div id="index">{Object.keys(fake).length}</div>
           }
         `,
+
         "app/routes/built-ins.jsx": js`
           import { useLoaderData } from "remix";
           import * as path from "path";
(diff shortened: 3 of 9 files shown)


packages/remix-dev/compiler.ts
```diff
@@ -337,8 +340,15 @@ async function createBrowserBuild(
     // All route entry points are virtual modules that will be loaded by the
     // browserEntryPointsPlugin. This allows us to tree-shake server-only code
     // that we don't want to run in the browser (i.e. action & loader).
-    entryPoints[id] =
-      path.resolve(config.appDirectory, config.routes[id].file) + "?browser";
+
+    // If any route entry point has *zero* browser safe exports (ie: only action/loader etc)
+    // then omit it entirely from the browser build
```
rossipedia, line 345: I feel this comment would be clearer if inverted to match the if condition:

> Only include routes that export at least one browser-safe name (if not, it's a resource route and should only run on the server). 

Or something like that

ryanflorence:
This is great, thank you! Let's add a test into `integration/compiler.ts` for this and get it merged.

In the test you have access to the internal `build.assets` and could inspect them to ensure the resource route is not included.

I'll add if you're not up for it :) Just lmk.

ryanflorence:
Ah I forgot about a case here. Sometimes you want to add a route that just redirects, like a "/calendar" route. Maybe you want that to just redirect to "today"

```js
export function loader() {
  return redirect(`/calendar/${getTodayParam()}`);
}
```

By removing the route completely Remix in the browser can't know it should be calling a loader that will then redirect. If it's a normal anchor then it all works fine because the code is called on the server and the redirect is sent.

For typical Resource Routes, none of this is a problem because you don't `<Link>` to them (except like a PDF but we already require apps to use `<Link reloadDocument>` for those).

So we have a few choices here:

1. Require apps to `export default function() { return null }` to signify that this *is not a resource route*. If you `<Link>` to a route, it's part of the UI so you must export a UI component to gets included in the client bundles.
   - This is nice because it's explicit. You link to UI routes, so you export components for UI routes. The definition for a resource route is "not part of the UI" so it lines up nicely
   - `<Link prefetch>` will work
   - This is annoying because you have to kinda know more about Remix than I'd like people to have to know.
2. When a `<Link>` is clicked to a route that isn't in the route manifest in the browser, do a normal browser navigation instead of a Remix navigation.
  - This is nice because I think it'll generally Just Work™.
  - It's a bummer that client state will be lost (scroll positions, focus, react state, etc)
  - We'd probably need to do some more defensive stuff with `<Link prefetch>`, today I think it explodes (we should probably do this work anyway though)
3. Include resource routes in the manifest but exclude them from the bundles. This will enable a link click to call the loader and get the redirect.
   - This is nice because it will generally Just Work™
   - You won't lose client state
   - The problem here is what do we do when the loader *doesn't* redirect, but returns a PDF or something and we called `fetch` on it expecting JSON.

Again, this isn't a general resource route concern, the only case we're concerned about is linking to routes that exist only to redirect elsewhere--which is actually quite common.

I think I want to do (1), but with some nice error messaging. When a link is clicked that's not in the manifest, Remix throws a 404 because nothing matched. We can add a warning to the console about the nuance of "Resource Routes" vs. "UI Routes that just Redirect". In development we could even send a resource route manifest to give an even better message about how the resource route exists so add a default export if it's there to redirect, or add `<Link reloadDocument>` if it sends a PDF or whatever, but that `<Link>` on its own doesn't know what to do w/ that route.

ryanflorence:
To summarize, there are two valid times to `<Link>` to a resource route:

1. **Actual Resource Route**: Link to a downloadable resource like a PDF. In these cases, use a normal `<a href>` or `<Link reloadDocument>`
2. **UI Route Redirect**: Linking to a route that redirects to some other route dynamically. In this case, you are required to `export default function() { return null }` to make it part of your UI.

From there, it's a task for Remix to help steer you the right direction when you `<Link>` to a resource route or don't export a default component for the UI redirect route.

ryanflorence:
Actually, in development we could warn the app developer to tell Remix the intent of the link:

```
You linked to a "Resource Route" from your UI, please add `<Link reloadDocument>` for the browser to be able to view or download this resource.

If it's not a resource route, please add a default export to your route so that it gets included in the UI. If it's just there to redirect, use `export default function() { return null }`.
```

But in production we can do (2) and make it (usually) Just Work™. When we match the routes and find nothing, we do a `window.reload`. If it's a resource route, the browser will view/download it. If it's a redirect, they'll get redirected. If it doesn't exist, they'll get a real 404.

This way Remix steers app developers to the optimal UX before they deploy (don't lose client state, don't try to `fetch` PDFs in a client transition expecting JSON), but if the developer misses the messaging, the app will probably do the right thing.

rossipedia:
TIL about the subtleties of the route manifest :) I agree that 1) seems like the most pragmatic approach. Let me take a stab at adding that warning to `<Link />` and see about getting that test to pass.

ryanflorence:
I think the old test is wrong and should warn instead. I'm still not totally sure what we want here, but I'm thinking:

1. For development builds, add resource routes to the manifest.
2. In `Link`, when `process.env.NODE_ENV === "development"` we can `matchRoutes` and see if it matches anything.
   - If it matches nothing, `console.warn` that the route is not found, go add it to `app/routes/`
   - If it matches a resource route, `console.warn` that they're linking to a resource route and should use `reloadDocument` or add a default export (the error message from my earlier comment)
3. In production and dev, when we `matchRoutes` before calling the client side transition, if the route is not found, do a `window.reload` at the new URL and then they'll either get:
   - a 404
   - a redirect from a route w/o a component that redirects
   - the resource they linked to (like a PDF or whatever)

This way we steer the developer to providing the right semantics to Remix, but fallback to the Right Thing™.

We can only do the Right Thing™ automatically at the cost of losing client state and the benefits of client transitions. So we do the warnings in dev to get the developer to provide the right semantics so that the app doesn't have to make that tradeoff in production.

ryanflorence:
@rossipedia if you're going to take on this larger scope, we can ditch the tests in the gists-app/tests and do them all in `integration/resource-routes-test.ts`

rossipedia:
Would be happy to! Just need to ramp up a bit on how `<Link />` works. Will try to get to it this evening.

rossipedia:
@ryanflorence I'm looking into this, and I'm assuming that for this bit:

> In production and dev, when we matchRoutes before calling the client side transition,

You're referring to [here][1] and [here][2]? Just wanna make sure I'm grokking where it needs to go.

[1]: https://github.com/remix-run/remix/blob/dev/packages/remix-react/transition.ts#L453
[2]: https://github.com/remix-run/remix/blob/dev/packages/remix-react/transition.ts#L507

ryanflorence:
First one yes, but `case "fetcher"` doesn't matter, that code path isn't involved with links or resource routes. Thank you for digging in to this!

rossipedia:
@ryanflorence if I'm understanding things correctly, we're basically altering the default "not found" behavior, right? In essence, we're removing the client-side 404 behavior by deferring it to be strictly server handled?

Instead of `window.reload` after the transition finishes, couldn't we achieve the same by doing a `location.href = createPath(location);` instead of calling into `handleNotFoundNavigation()`? 

Just wanna make sure my logic is sound before I go ripping into a bunch of tests that have to be updated :)

rossipedia:
@ryanflorence ok probably not worth running the workflow since I didn't change anything in the tests, I wanted to mainly just push up what I had that I _think_ is what you outlined above and make sure that the approach is solid before I start fixing up the tests.

The changes were a bit more involved since adding "resource routes" in dev mode meant changing types in a handful of places, but the meat of the new changes are in [`components.tsx`](https://github.com/remix-run/remix/pull/1837/files#diff-594d4b6ed5bd4abff7e9319566b5505b6a85753e665ff4d2113e888384a98377) and [`transition.ts`](https://github.com/remix-run/remix/pull/1837/files#diff-6dda36fa0981ac6fdbdb64f280958e54decf21c27d91a610e3359f02d06bc10f).

Lemme know if you want me to split this up into more granular commits 👍

rossipedia:
> Lemme know if you want me to split this up into more granular commits 

Ok I went ahead and did this anyways :)

MichaelDeBoey:
@rossipedia Could you please rebase your branch onto latest `dev` & resolve conflicts?

PR #1839, opened 2022-02-08
feat: add serverDependenciesToBundle to remix config

jacob-ebey:
Goal here is to enable you to synchronically consume ESM modules in CJS server builds by allowing you to specify what modules are transpiled and included in the server bundle. You can just bundle specific modules via string matching, or get specific with regexes, or bundle everything with `/.*/`.


packages/remix-dev/config.ts
```diff
@@ -128,6 +128,13 @@ export interface AppConfig {
    * routes.
    */
   ignoredRouteFiles?: string[];
+
+  /**
+   * A list of module names or regular expressions that determined if a module
+   * is transpiled and included in the server bundle. This can be useful when
+   * consuming ESM only packages in a CJS build.
+   */
+  transpileModules?: Array<string | RegExp>;
```
kentcdodds, line 137: Is there a different name to give this property to make it more clear that:

1. This only applies to the server build
2. This only applies to node_modules

jacob-ebey: I'm down to call it "transpileServerBareModules" or similar, but the thing is that you're never(?) going to be given this option for a client build as that all needs to be bundled regardless.

mjackson: How about `serverDependenciesToBundle`?

Things I like about it:

- it starts with `server`, just like all our other server-only config vars
- it specifically talks about dependencies (node_modules)

jacob-ebey: What about `serverDependenciesToTranspile`? I like the "transpile" over "bundle" for a few reasons:
- it's more than just bundling, it does in-fact transpile ESM->CJS
- prior IP refers to bundling / transforming dependencies as transpile and might be easier for people to find / use https://www.npmjs.com/package/next-transpile-modules

kentcdodds: I prefer `serverDependenciesToBundle`:

1. It's doing more than "transpiling"
2. "Bundle" suggests also compiling in some way
3. "Transpile" is a made-up word that I never liked 😆

kentcdodds: In any case though, I wouldn't stop the ship over this. I'm fine with either.

integration/compiler-test.ts
```diff
@@ -36,8 +36,30 @@ describe("compiler", () => {
...
+        "app/routes/esm-only-pkg.jsx": js`
+          import esmOnlyPkg from "esm-only-pkg";
+
+          export default function EsmOnlyPkg() {
+            return <div id="esm-only-pkg">{esmOnlyPkg}</div>;
+          }
+        `,
+        "remix.config.js": js`
+          module.exports = {
+            ignoredRouteFiles: [".*"],
```
kentcdodds, line 51:
```suggestion
```

Better to leave unrelated things out to avoid people thinking it's a relevant part of the test.

packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts
```diff
@@ -82,6 +82,17 @@ export function serverBareModulesPlugin(
             return undefined;
         }
 
+        for (let transpileMatcher of remixConfig.serverDependenciesToBundle) {
```
kentcdodds, line 85:
```suggestion
        for (let matcher of remixConfig.serverDependenciesToBundle) {
```

I'm on mobile otherwise I would rename all the variables below as well. This is more of a nitpick so feel free to ignore.

packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts
```diff
@@ -82,6 +82,17 @@ export function serverBareModulesPlugin(
...
+          // bundle it if the path matches the transpile matcher
```
kentcdodds, line 86:
```suggestion
          // bundle it if the path matches the matcher
```

kentcdodds (changes requested):
Looks great to me. I just have a few small things. This will be so nice 👍

kentcdodds (approved):
Looks great to me 👏

Issue #1823, opened 2022-02-06
esbuild externals not accessible


Industrial:
### Which Remix packages are impacted?

- `remix` (Remix core)

### What version of Remix are you using?

1.1.3

### What version of Node are you using? Minimum supported version is 14.

17

### Steps to Reproduce

When using MikroORM, i get build errors from esbuild:

```
Build failed with 8 errors:
node_modules/.pnpm/registry.npmjs.org+@mikro-orm+core@4.5.10_4c746656b7f7a21ae58d5630207e3edf/node_modules/@mikro-orm/core/utils/Configuration.js:237:61: error: Could not resolve "@mikro-orm/mongodb" (mark it as external to exclude it from the bundle, or surround it with try/catch to handle the failure at run-time)
node_modules/.pnpm/registry.npmjs.org+@mikro-orm+core@4.5.10_4c746656b7f7a21ae58d5630207e3edf/node_modules/@mikro-orm/core/utils/Configuration.js:238:61: error: Could not resolve "@mikro-orm/mysql" (mark it as external to exclude it from the bundle, or surround it with try/catch to handle the failure at run-time)
node_modules/.pnpm/registry.npmjs.org+@mikro-orm+core@4.5.10_4c746656b7f7a21ae58d5630207e3edf/node_modules/@mikro-orm/core/utils/Configuration.js:239:65: error: Could not resolve "@mikro-orm/mariadb" (mark it as external to exclude it from the bundle, or surround it with try/catch to handle the failure at run-time)
node_modules/.pnpm/registry.npmjs.org+@mikro-orm+core@4.5.10_4c746656b7f7a21ae58d5630207e3edf/node_modules/@mikro-orm/core/utils/Configuration.js:241:63: error: Could not resolve "@mikro-orm/sqlite" (mark it as external to exclude it from the bundle, or surround it with try/catch to handle the failure at run-time)
node_modules/.pnpm/registry.npmjs.org+@mikro-orm+core@4.5.10_e93952d69f1164f307a4f6806ae7c0b3/node_modules/@mikro-orm/core/utils/Configuration.js:237:61: error: Could not resolve "@mikro-orm/mongodb" (mark it as external to exclude it from the bundle, or surround it with try/catch to handle the failure at run-time)
...
💿 Built in 667ms
```

I found that here: https://github.com/remix-run/remix/blob/c746c6f75fb8a74da92df91113df5af7bf87adea/packages/remix-dev/compiler.ts#L341

[...]

PR #1841, opened 2022-02-08
Added `externals` option for esbuild

Industrial:
I am using MikroORM and esbuild does static analysis of the source files and throws errors on the pakages of the connectors of the databases I am not using. For this reason, I added the option. This would fix https://github.com/remix-run/remix/issues/1823

Example: https://github.com/Industrial/test-remix-vercel/blob/main/remix.config.js#L10

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -123,7 +123,8 @@ export async function watch(
     },
     onFileDeleted(file) {
       log(`File deleted: ${path.relative(process.cwd(), file)}`);
-    }
+    },
+    externals: config.externals,
   });
 
   console.log(`💿 Built in ${prettyMs(Date.now() - start)}`);
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -33,6 +33,7 @@ interface BuildConfig {
   mode: BuildMode;
   target: BuildTarget;
   sourcemap: boolean;
+  externals: string[];
 }
 
 function defaultWarningHandler(message: string, key: string) {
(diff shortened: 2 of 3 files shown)


fixtures/gists-app/pnpm-lock.yaml
```diff
@@ -0,0 +1,4853 @@
+lockfileVersion: 5.3
```
chaance: @Industrial I'm going to nudge some other folks on the team to review, but in the mean time do you mind removing this file from the PR? I added `pnpm` lockfiles to `gitignore` on `dev` so this shouldn't be an issue if you merge.

pajasry:
Any progress or at least time when we can expect a merge?  I can't wait to use a mikroorm

andrewbrey:
This is an issue that impacted a little toy project I made (discussed in project notes here: https://github.com/andrewbrey/remix-curl#notes) and I just wanted to highlight that `patch-package` can be used pretty effectively as a workaround for this issue until this PR is merged (or some other solution is in place to handle end user's telling `esbuild` about externals) :+1:

dvargas92495:
Anything we could do to push over the finish line? I'm hitting this with a different package that I'd like to add to esbuild externals

chaance:
@dvargas92495 We still aren't really sure about exposing esbuild through our config. We consider esbuild an implementation detail rather than API. I've raised the issue up the flag pole internally, but if you need this I'd recommend `patch-package` for now.

dvargas92495:
Yea I'm already using a patch-package-esque workaround for now, so this is mostly me asking to just reduce the amount of workarounds in my build process as possible.

That's a fair point on not exposing the entire esbuild API. However `externals` is a concept you'd find across different bundlers (webpack, etc) so it wouldn't be too far removed to imagine it on remix's build api too

MichaelDeBoey:
@Industrial Could you please rebase your branch onto latest `dev` & resolve conflicts?

Issue #1140, opened 2021-12-18
[Bug]: Can no longer access header values sent through JSON helper via `loaderHeaders.get`


rjgux:
### Which Remix packages are impacted?

- [X] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.1.1

### What version of Node are you using? Minimum supported version is 14.

16.14.0

### Steps to Reproduce

If I try to access headers sent via the JSON HTTP helper within my exported `headers` function, the values are always null -

```
import { json } from 'remix'

export let loader: LoaderFunction = async () => {
  let data = await getData()

  return json(data, {
    headers: {
      'Cache-Control': 'max-age=300, s-maxage=300, stale-while-revalidate',
      'Some-Random-Header': 'some-random-value',
    },
  })
}

export function headers({ loaderHeaders }: { loaderHeaders: Headers }) {
  console.log(loaderHeaders.get('Cache-Control')) // null
  console.log(loaderHeaders.get('Some-Random-header')) // null
  return {
    'Cache-Control': loaderHeaders.get('Cache-Control'),
  }
}
```

### Expected Behavior

`loaderHeaders.get('Cache-Control')` value is `'max-age=300, s-maxage=300, stale-while-revalidate'`
`loaderHeaders.get('Some-Random-header')` value is `'some-random-value'`

### Actual Behavior

`loaderHeaders.get('Cache-Control')` value is `null`
`loaderHeaders.get('Some-Random-header')` value is `null`

Xiphe:
having the same issue in a regular route. 

(Same code as above but with a default component export)

marbiano:
Can confirm that this is happening to me too.

ryanflorence:
I'm not getting anything here, can ya'll make a reproduction on `https://remix.run`?

```js
import { json } from "remix";
import type { HeadersFunction } from "remix";

export function loader() {
  return json(null, { headers: { "X-Test": "SUCCESS!" } });
}

export let headers: HeadersFunction = ({ loaderHeaders }) => {
  return {
[...]

PR #1847, opened 2022-02-08
fix(remix-server-runtime): fix headers function access to loaderHeaders when parent has no loader

xstevenyung:
This should fixes issue: https://github.com/remix-run/remix/issues/1140

The issue came from `routeLoaderResponses` not taking into account that some parent route might not have loaders hence no response. Because `getDocumentHeaders` matched the response to the `index` position in `routeLoaderResponses`, this could provoke a mismatch.

I replaced `routeLoaderResponses` from an array to a `Record<string, Response>` (matching the route's id instead of the index) which would solve the issue.

I also created a test but I think it might not be in the right place, feel free to move it if needed.

Thanks for your awesome work, I love Remix !

### DIFF
--- a/packages/remix-server-runtime/__tests__/server-test.ts
+++ b/packages/remix-server-runtime/__tests__/server-test.ts
@@ -1679,37 +1679,4 @@ describe("shared server runtime", () => {
       expect(spy.console.mock.calls.length).toBe(1);
     });
   });
-
-  describe("header requests", () => {
-    test("can access loader headers if parent doesn't have a loader itself", async () => {
-      let indexLoader = jest.fn(() => {
-        return new Response(null, { headers: { "X-Test": "SUCCESS" } });
-      });
-      let indexHeaders = jest.fn(({ loaderHeaders }) => {
-        return {
-          "X-Test": loaderHeaders.get("X-Test") || "FAILED"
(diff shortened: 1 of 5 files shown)


packages/remix-server-runtime/__tests__/server-test.ts
```diff
@@ -1679,4 +1679,37 @@ describe("shared server runtime", () => {
       expect(spy.console.mock.calls.length).toBe(1);
     });
   });
+
+  describe("header requests", () => {
+    test.only("can access loader headers if parent doesn't have a loader itself", async () => {
```
mkrtchian, line 1684: I think you forgot to remove this `only`.

xstevenyung: Nice catch, fixed. Thanks very much for commenting 👍

ryanflorence:
Awesome! you mind removing your test and then simply removing the `skip` on this one? (and double check my work from last week that this is testing the same thing?)

https://github.com/remix-run/remix/blob/40216cdf83593b53e1f8d98b936e4662238787b6/integration/headers-test.ts#L83

ryanflorence:
nvm I got it :)

PR #1854, opened 2022-02-09
fix: `onSubmit`: Get submission values from form attributes instead of props

chaance:
> Named inputs are added to their owner form instance as properties and can overwrite native properties. ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement))

With a standard form, the form submission will still use the values set by the form attributes (or the default value if none is set), so this PR changes our behavior for `useSubmit` (and subsequently `<Form>`) to match the browser default.


ryanflorence:
You want to add a test that shows the form actions resolve correctly? I think that's why we use the `form.action` to get the automatic browser url resolution that we then resolve relative to the route in context (just like relative `<Link to>`). My memory is that we needed it, but I'm not so sure anymore, seems like it would be easier to do relative resolution to the route-in-context with the attribute--the whole point is to ignore the URL and just use the route match hierarchy the very same way `<Link to>` works.

So with routes like:

Put some forms with various actions (none, relative, absolute, etc.) in a parent route and then assert the form's html by throwing various routes at the fixture.

In fact, you can look at the react router tests to see how `to` resolves given different route hierarchies and current URLs (current URL shouldn't matter, relative URLs are all about route hierarchy). Form action should work the same (except link's don't have a default like form).

Might want to add those tests as a commit *before* your changes here so we can see any regressions more easily.

packages/remix-react/components.tsx
```diff
@@ -1090,11 +1096,20 @@ export function useSubmitImpl(key?: string): SubmitFunction {
...
-          options.method || target.getAttribute("formmethod") || form.method;
+          options.method ||
+          target.getAttribute("formmethod") ||
+          form.getAttribute("method") ||
+          defaultMethod;
         action =
-          options.action || target.getAttribute("formaction") || form.action;
+          options.action ||
+          target.getAttribute("formaction") ||
+          form.getAttribute("action") ||
```
ryanflorence, line 1106: I think this needs to be resolved now relative to the route hierarchy like Link (not the URL segments the way the browser does it)

packages/remix-react/components.tsx
```diff
@@ -1068,9 +1070,13 @@ export function useSubmitImpl(key?: string): SubmitFunction {
           options as any
         ).submissionTrigger;
 
-        method = options.method || target.method;
-        action = options.action || target.action;
-        encType = options.encType || target.enctype;
+        method =
+          options.method || target.getAttribute("method") || defaultMethod;
+        action =
+          options.action || target.getAttribute("action") || defaultAction;
```
ryanflorence, line 1076: I think this needs to be resolved now relative to the route hierarchy like Link (not the URL segments the way the browser does it)

packages/remix-react/components.tsx
ryanflorence: We're not in a hook here so it's gonna be a bit tricky, will have to copy/pasta some stuff from here: https://github.com/remix-run/react-router/blob/main/packages/react-router/index.tsx#L611-L623

@mjackson why do the matches go through JSON.stringify and then JSON.parse in useResolvedPath?

ryanflorence (changes requested):
Gotta make sure `<Form action>` and `<button formAction>` resolve with React Router's route hierarchy resolution and not the browser's URL segment hierarchy.

packages/remix-react/components.tsx
chaance: See other notes, I think this is fine!

packages/remix-react/components.tsx
```diff
@@ -1054,6 +1054,8 @@ export function useSubmit(): SubmitFunction {
 export function useSubmitImpl(key?: string): SubmitFunction {
   let navigate = useNavigate();
   let defaultAction = useFormAction();
```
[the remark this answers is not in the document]
chaance: See other notes, I think this is fine!

integration/helpers/create-fixture.tsx
```diff
@@ -399,6 +408,18 @@ export async function getHtml(page: Page, selector?: string) {
...
 }
 
+export function getAttribute(
+  html: string,
+  selector: string,
+  attributeName: string
+) {
+  let el = cheerio(selector, html);
+  if (!el.length) {
+    throw new Error("Invalid HTML");
```
mjackson, line 418: Seems like instead of saying "invalid HTML" we could say something like "no element matches selector X" or something.

integration/helpers/create-fixture.tsx
```diff
@@ -307,6 +307,15 @@ export async function createAppFixture(fixture: Fixture) {
        */
       getHtml: (selector?: string) => getHtml(page, selector),
 
+      /**
+       * Get the attribute from an HTML element given a selector.
+       *
+       * @param selector CSS Selector for the element's HTML you want
+       * @param attribute Name of the attribute you want
+       */
+      getAttribute: async (selector: string, attribute: string) =>
```
mjackson, line 316: `app.getAttribute` is a confusing interface. Is it getting the attribute of the app? It'd be clearer if we had an interface like:

```
let html = app.getHTML();
let el = getElement(html, selector);
let attr = getAttribute(el, attrName);
```

Can we make this more clear?

integration/helpers/create-fixture.tsx
chaance: Updated

integration/helpers/create-fixture.tsx
chaance: Updated, though I omitted `getAttribute` since `getElement` returns a cheerio instance we can use directly via `element.attr("#whatever")`

dpcesx (commented):
Cool

mjackson (approved):

Issue #688, opened 2021-11-26
[Feature]: Dark mode toggle example


arvigeus:
### What is the new or updated feature that you are suggesting?

An example of a toggle button that switches theme on demand.

### Why should this feature be included?

Showcases conditionally loading stylesheets

sergiodxa:
Check this https://ebey.me/blog/dynamic-css-with-remix-resource-routes

arvigeus:
Thank you, @sergiodxa! Although this looks like way too much work for something as simple as that. I think something like your example should be included in docs.

abereghici:
@arvigeus I created an abstraction for themes here : https://github.com/abereghici/remix-themes

arvigeus:
I finally implemented it in my project: <https://github.com/arvigeus/format-code>

Basically using cookies for css files, as @sergiodxa suggested. Also the [toggleTheme action](https://github.com/arvigeus/format-code/blob/master/app/routes/toggleTheme.ts) returns set of css variables, so the change is applied on the fly, without reloading. Caveat: the next Link clicked MUST reload the page, or the new page will still display the old theme.

I'll close this issue for now because this repo is already polluted enough.

PR #1857, opened 2022-02-09
docs(example): Add an example to persist theme with MUI and HTTP cookie

arpitdalal:
Using Remix to create a persistent MUI theme switcher with HTTP cookie.


kentcdodds:
Hi! We do have a Dark Mode example in the works and it's pretty extensive: #688 

If you want to make a simple MUI example that would be fine. I just think including all the dark mode/cookie stuff is duplicate work here. Would you like to simplify this?

kentcdodds (changes requested):
If you'd like to simplify this, feel free :) Thanks!

I should mention we also have an emotion example as well which maybe you'd be interested in helping with :) #1485 Thank you!

arpitdalal:
Yes, that makes sense, I wasn't aware of the dark mode example being in the works so I hacked this one together. You guys are doing an awesome job, just love Remix, haven't used web standards with React before lol. It's new and exciting!!

kentcdodds:
Thank you for understanding! We're so happy that folks like you are willing to contribute. It makes the whole community better. Thanks for taking part!

PR #1861, opened 2022-02-09
fix(create-remix/express): update caching to be dynamic based on path

mcansh:
otherwise it's always 1 hour https://canary.discord.com/channels/770287896669978684/777863745321304104/941061633000763444

### DIFF
--- a/packages/create-remix/templates/express/server.js
+++ b/packages/create-remix/templates/express/server.js
@@ -1,3 +1,4 @@
+import path from "path";
 import express from "express";
 import compression from "compression";
 import morgan from "morgan";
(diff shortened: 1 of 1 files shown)


packages/create-remix/templates/express/server.js
```diff
@@ -1,21 +1,35 @@
+import path from "path";
 import express from "express";
 import compression from "compression";
 import morgan from "morgan";
 import { createRequestHandler } from "@remix-run/express";
 
 import * as build from "@remix-run/dev/server-build";
 
+const PUBLIC_DIR = path.join(process.cwd(), "public");
+const BROWSER_BUILD_DIR = "/build/";
```
girishk21, line 10: won't this break the ability to configure [`publicPath`](https://remix.run/docs/en/v1/api/conventions#publicpath)? 🤔

PR #1444, opened 2022-01-09
fix: allow build/start as esm module [ready for review]

dima-takoy:
ATM, if you set `{"type": "module"}` in `package.json` you lost ability to build remix app because `require()` is not in module context.

With this change we will use `import()` statement that works with CommonJS and ESM. [More info in v14.x docs](https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_interoperability_with_commonjs)

Other changes:
— remix-dev createServer now async function, it can break code that relies on @remix-run/dev
— appConfig.serverModuleFormat inherits value from package.json type field. if type === undefined, then 'cjs' else 'esm'.

ATM serverModuleFormat is undocumented feature. I am not native speaker and i cannot update documentation :-(

Commonjs `remix.config.js`
```JS
module.exports = {
	appDirectory: "app",
	assetsBuildDirectory: "public/build",
	publicPath: "/build/",
	serverBuildDirectory: "build",
	devServerPort: 8002,
	ignoredRouteFiles: [".*"],
	serverModuleFormat: "esm"
};
```

ESM `remix.config.js`
```JS
export default {
	appDirectory: "app",
	assetsBuildDirectory: "public/build",
	publicPath: "/build/",
	serverBuildDirectory: "build",
	devServerPort: 8002,
	ignoredRouteFiles: [".*"],
	serverModuleFormat: "esm"
};
```
Related issue and PR:
https://github.com/remix-run/remix/issues/1283
https://github.com/remix-run/remix/pull/976
https://github.com/remix-run/remix/issues/1279

**P.S.**: Also looks like all build tests are skipped now =\, but trust me, it works! :D
https://github.com/remix-run/remix/blob/dev/packages/remix-dev/__tests__/build-test.ts#L20

PR #1871, opened 2022-02-10
wip: feat: add support for node-esm serverBuildTarget

jacob-ebey:
feat: add remix-esm template

### DIFF
--- a/integration/node-esm-test.ts
+++ b/integration/node-esm-test.ts
@@ -1,0 +1,36 @@
+import { createFixture, createAppFixture, js } from "./helpers/create-fixture";
+import type { Fixture, AppFixture } from "./helpers/create-fixture";
+
+describe("node-esm", () => {
+  let fixture: Fixture;
+  let app: AppFixture;
+
+  beforeAll(async () => {
+    fixture = await createFixture({
+      server: "remix-esm",
+      files: {
+        "app/routes/index.jsx": js`
+          export default function Index() {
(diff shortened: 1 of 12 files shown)


jacob-ebey:
https://github.com/remix-run/remix/pull/1444

jacob-ebey:
OK, we have issues with ESM believe it or not! Mainly the "require cache" equivalent for ESM is not exposed or mutable, this means no auto reload without a memory leak the size of the bundled server module: https://ar.al/2021/02/22/cache-busting-in-node.js-dynamic-esm-imports/

I guess we have a few options here:
1.) do not attempt to make `remix dev` work for node-esm and require you to have a custom server with nodemon and use `remix watch`
2.) manage a separate process for the app instead of running it as part of the `remix dev` process. This has the main drawback of loosing any and all global state / variables you use in your app. Things like DB connections would be reset and reconnect every change in dev. This may not be a big deal but does not seem ideal to me.
3.) just send it and have a memory leak in dev. We have one today that I have not yet investigated, how much worse could it get?

TomerAberbach:
For (2), if you wanted to avoid managing a separate process, then you could use a worker like what's mentioned in https://github.com/nodejs/modules/issues/307#issuecomment-858729422 (and kill the worker each time you want to "clear the cache")

MichaelDeBoey:
We already did a lot of ESM-related things the last weeks, so do we still need changes from this one?

PR #1876, opened 2022-02-10
types(remix-server-runtime): make `AppLoadContext` an empty interface instead of `any`

DanielFGray:
this allows users to override the `context` with their own definitions using ambient modules.

as an example:
```js
declare module "remix" {
  interface AppLoadContext {
    graphql: <Result, Variables>(operation: {
      query: TypedDocumentNode<Result, Variables>
      variables: Variables
    }) => Promise<ExecutionResult<Result>>
  }
}
```

### DIFF
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -6,7 +6,9 @@ import { json, isResponse, isRedirectResponse } from "./responses";
  * An object of arbitrary for route loaders and actions provided by the
  * server's `getLoadContext()` function.
  */
-export interface AppLoadContext {};
+export interface AppLoadContext {
+  [key: string]: unknown;
+};
 
 /**
  * Data for a route that was returned from a `loader()`.
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -3,7 +3,7 @@ import type { ServerRoute } from "./routes";
 import { json, isResponse, isRedirectResponse } from "./responses";
 
 /**
- * An object of arbitrary for route loaders and actions provided by the
+ * An object of unknown type for route loaders and actions provided by the
  * server's `getLoadContext()` function.
  */
 export interface AppLoadContext {
@@ -20,7 +20,7 @@ export async function callRouteAction({
   match,
   request,
 }: {
-  loadContext: unknown;
+  loadContext: AppLoadContext;
   match: RouteMatch<ServerRoute>;
   request: Request;
 }) {
@@ -67,7 +67,7 @@ export async function callRouteLoader({
 }: {
   request: Request;
   match: RouteMatch<ServerRoute>;
-  loadContext: unknown;
+  loadContext: AppLoadContext;
 }) {
   let loader = match.route.module.loader;
 
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -6,7 +6,7 @@ import { json, isResponse, isRedirectResponse } from "./responses";
  * An object of arbitrary for route loaders and actions provided by the
  * server's `getLoadContext()` function.
  */
-export type AppLoadContext = any;
+export interface AppLoadContext {};
 
 /**
  * Data for a route that was returned from a `loader()`.
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -8,7 +8,7 @@ import { json, isResponse, isRedirectResponse } from "./responses";
  */
 export interface AppLoadContext {
   [key: string]: unknown;
-};
+}
 
 /**
  * Data for a route that was returned from a `loader()`.
@@ -20,7 +20,7 @@ export async function callRouteAction({
   match,
   request,
 }: {
-  loadContext: AppLoadContext;
+  loadContext?: AppLoadContext;
   match: RouteMatch<ServerRoute>;
   request: Request;
 }) {
@@ -67,7 +67,7 @@ export async function callRouteLoader({
 }: {
   request: Request;
   match: RouteMatch<ServerRoute>;
-  loadContext: AppLoadContext;
+  loadContext?: AppLoadContext;
 }) {
   let loader = match.route.module.loader;
 
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -16,7 +16,7 @@ import { createServerHandoffString } from "./serverHandoff";
 
 export type RequestHandler = (
   request: Request,
-  loadContext: AppLoadContext
+  loadContext?: AppLoadContext
 ) => Promise<Response>;
 
 export type CreateRequestHandlerFunction = (
@@ -82,7 +82,7 @@ async function handleDataRequest({
   serverMode,
 }: {
   handleDataRequest?: HandleDataRequestFunction;
-  loadContext: AppLoadContext;
+  loadContext?: AppLoadContext;
   matches: RouteMatch<ServerRoute>[];
   request: Request;
   serverMode: ServerMode;
@@ -180,7 +180,7 @@ async function handleDocumentRequest({
   serverMode,
 }: {
   build: ServerBuild;
-  loadContext: AppLoadContext;
+  loadContext?: AppLoadContext;
   matches: RouteMatch<ServerRoute>[] | null;
   request: Request;
   routes: ServerRoute[];
@@ -516,7 +516,7 @@ async function handleResourceRequest({
   serverMode,
 }: {
   request: Request;
-  loadContext: AppLoadContext;
+  loadContext?: AppLoadContext;
   matches: RouteMatch<ServerRoute>[];
   serverMode: ServerMode;
 }): Promise<Response> {
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -16,7 +16,7 @@ import { createServerHandoffString } from "./serverHandoff";
 
 export type RequestHandler = (
   request: Request,
-  loadContext?: AppLoadContext
+  loadContext: AppLoadContext
 ) => Promise<Response>;
 
 export type CreateRequestHandlerFunction = (
@@ -82,7 +82,7 @@ async function handleDataRequest({
   serverMode,
 }: {
   handleDataRequest?: HandleDataRequestFunction;
-  loadContext: unknown;
+  loadContext: AppLoadContext;
   matches: RouteMatch<ServerRoute>[];
   request: Request;
   serverMode: ServerMode;
@@ -180,7 +180,7 @@ async function handleDocumentRequest({
   serverMode,
 }: {
   build: ServerBuild;
-  loadContext: unknown;
+  loadContext: AppLoadContext;
   matches: RouteMatch<ServerRoute>[] | null;
   request: Request;
   routes: ServerRoute[];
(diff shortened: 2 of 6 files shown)


DanielFGray:
> What does this do if you don't add that `declare`? Does it treat it as `any` or are you required to `declare` it?

do you mean the `declare` in my example? it's my understanding that it's not a valid ambient module declaration without it

itsMapleLeaf:
> What does this do if you don't add that `declare`? Does it treat it as `any` or are you required to `declare` it?

If you _don't_ add the `declare`, the type of context will be `{}`. This PR allows you to add a `declare` to change the context type to whatever you want. This is similar to declaring a type to extend the express request: https://stackoverflow.com/a/40762463/1332403

rootblack45:
This is how I currently override the type for `context` it's quite hard btw. I'm agreed with this PR since typescript will merge the interface with the same name within a module.
https://github.com/rootblack45/rootblack45.me/blob/484988376882c900beb0b08663afdeca85142d0d/app/types.d.ts#L3-L16

ryanflorence:
Sorry, I'm not as well-versed in TypeScript as you might expect, I just know enough to get by.

I understand this PR enables you to do this:

```ts
declare module "remix" {
  interface AppLoadContext {
    graphql: <Result, Variables>(operation: {
      query: TypedDocumentNode<Result, Variables>
      variables: Variables
    }) => Promise<ExecutionResult<Result>>
  }
}
```

What if in my app I don't declare anything and I add load context and try to access it, what will the type be?

```ts
export let loader: LoaderFunction = async ({ context }) => {
  // what is context? Any? Can I access stuff on it w/o typescript getting mad?
  console.log("is TS mad?", context.foo);
}
```

I'd like TypeScript to treat it as any.

rootblack45:
> What if in my app I don't declare anything and I add load context and try to access it, what will the type be?

That should be `undefined` since this just allows us to work with types easier. It's like javascript if the value doesn't exist in the object so that value should be `undefined`

airjp73:
> I don't know why this question is so hard for me to ask and get an answer that makes sense to me 😅
> 
> With this PR, do I get red squiggly lines under `context.foo` if I don't declare an app load context type?

I think the answer is yes. [Typescript playground example](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCoQM5mQbwL4BQBA9McgK4gID2AtrROMgEYQA21A7smNclBFrUAbigEATZBgCO5YAHN5bTEVLJxEBGzgDkoSLEQp0WXCTLJLyOBnEwAXFLBRQ888kIEaIU5CyOTbABeXDxrDGsQAE8AbiI-MAA6GzsYoA)

ryanflorence:
Do you have to have an empty interface to be able to do this or can you do it if it's typed as any?

```ts
declare module "remix" {
  interface AppLoadContext {
    /* type here */
  }
}
```

airjp73:
There are other things like namespaces that support [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html), but `type` does not. You get a "duplicate identifier" warning ([typescript playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEBbA9sArhBByA7gSwBcALAWjwDsCQYAzKMEAZy3gG8AoeL+EADwAdkMAvApVa9BAEF+-AMLJKfEawC+7de1CRYCFOkzxchUmOp0GzNp259Bw0UokN4M+Yqq8VN7txrJkAC54RgIYCgBzAG4feHVNbWg4JFQMbHxiEgIAT34QFg5fOyERHLzXWQUlL3gAXngocmyYhPAkvVTDY0yy-OsigRL4Xor3apF6tQ0gA)).

rootblack45:
> Do you have to have an empty interface to be able to do this or can you do it if it's typed as any?

You need to do that only if you want to add new properties to an interface. After that Typescript will merge the properties after you declare.

rootblack45:
This PR is just for better DX since esbuild doesn't care what you declare in the type system 😂

ryanflorence:
@itsMapleLeaf ah, okay, I can maybe live with that? Don't like that it has to be an object now, but seems weird to be anything else.

@mjackson you have an opinion here?

3x071c:
@ryanflorence Just came here after almost having opened a similar PR myself. This is the signature for `AppLoadContext` I'd recommend instead of an `any` type, which doesn't allow for declaration merging and therefore can't be strongly typed:
```
export declare interface AppLoadContext {
	[key: string]: unknown
};
```
This will continue allowing accessing any properties on `AppLoadContext` by default, but prevents unsafe property access by requiring type assertions at runtime. For those that prefer strict typing and know the structure of their `context` in advance, they can merge this generic interface declaration with their own.

MichaelDeBoey:
@DanielFGray Code/type changes go against `dev`.
Could you please rebase your branch onto latest `dev` & set base branch to `dev` too?

DanielFGray:
I have updated this to use `unknown`, per @3x071c's suggestion

@MichaelDeBoey I think I rebased this properly

marwan38:
Looks like it includes changes that weren't in this PR. Might want to reflog and or rebase again. Looking forwards to this being merged in as we have a similar use case.

MichaelDeBoey:
@DanielFGray Like @marwan38 said: the rebase included changes that doesn't belong in this PR.
Please remove these from your branch/PR.

MichaelDeBoey:
@DanielFGray Could you please rebase your branch onto latest `dev` & resolve conflicts?

marwan38:
@MichaelDeBoey I've cherry-picked the authors commits and added a few of mine own. https://github.com/remix-run/remix/pull/3422

Not to discourage the author, I just want to get this into the repo :).

MichaelDeBoey:
@marwan38 Please put remarks/suggestions in this PR instead of creating a separate PR.

marwan38:
> @marwan38 Please put remarks/suggestions in this PR instead of creating a separate PR.

I found an issue while trying out the changes introduced in this PR. The type references to the `AppLoadContext` were erroring out due to the other references being types as `unknown`.

```ts
- type AppLoadContext = any;
+ interface AppLoadContext {
+    [key: string]: unknown;
+ }

// Other functions that rely on the context
const someLoaderThing = (loadContext: unknown) => { // ... } // type error, type unknown conflicts with the new interface
```

I made a few changes that fixed this issue, they can be seen here: https://github.com/remix-run/remix/pull/3422/files

justinwaite:
@MichaelDeBoey just out of curiosity, at what point could we consider this stale or abandoned and give someone else a chance to get it across the finish line? I think this is a pretty large impact item for our company since our apps will all rely on this context. Let me know, I'm definitely willing to help!

packages/remix-server-runtime/data.ts
```diff
@@ -18,7 +20,7 @@ export async function callRouteAction({
   match,
   request,
 }: {
-  loadContext: unknown;
+  loadContext: AppLoadContext;
```
justinwaite, line 23:
```suggestion
  loadContext?: AppLoadContext;
```

packages/remix-server-runtime/data.ts
```diff
@@ -65,7 +67,7 @@ export async function callRouteLoader({
 }: {
   request: Request;
   match: RouteMatch<ServerRoute>;
-  loadContext: unknown;
+  loadContext: AppLoadContext;
```
justinwaite, line 70:
```suggestion
  loadContext?: AppLoadContext;
```

packages/remix-server-runtime/server.ts
```diff
@@ -82,7 +82,7 @@ async function handleDataRequest({
   serverMode,
 }: {
   handleDataRequest?: HandleDataRequestFunction;
-  loadContext: unknown;
+  loadContext: AppLoadContext;
```
justinwaite, line 85:
```suggestion
  loadContext?: AppLoadContext;
```

packages/remix-server-runtime/server.ts
```diff
@@ -180,7 +180,7 @@ async function handleDocumentRequest({
   serverMode,
 }: {
   build: ServerBuild;
-  loadContext: unknown;
+  loadContext: AppLoadContext;
```
justinwaite, line 183:
```suggestion
  loadContext?: AppLoadContext;
```

packages/remix-server-runtime/server.ts
```diff
@@ -519,7 +519,7 @@ async function handleResourceRequest({
   serverMode,
 }: {
   request: Request;
-  loadContext: unknown;
+  loadContext: AppLoadContext;
```
justinwaite, line 522:
```suggestion
  loadContext?: AppLoadContext;
```

MichaelDeBoey:
@justinwaite If you're willing to actively working on this, you can always open a new PR, since @DanielFGray isn't responding for over a month now.

DanielFGray:
> @justinwaite If you're willing to actively working on this, you can always open a new PR, since @DanielFGray isn't responding for over a month now.

@MichaelDeBoey I literally just pushed a new commit yesterday. I'm still working out some of the kinks since a lot of things changed since I first filed this, and I'm also having trouble running the test suite locally which I've been asking for help with in the Discord #contributions channel.

I'm very interested in having this merged and actively working fixing on the remaining failing tests.

MichaelDeBoey:
@DanielFGray Oh sorry, I looked over that commit, I only saw @justinwaite's message en then the Changeset message 🙈

@justinwaite Let's continue with this PR instead of creating a new one.

marwan38:
> 

Out of curiosity, what do you need the tests for? Running a type check should be all you need since all that's being change is the type definitions.

pcattori:
Hey all! Coming to this conversation a little late, but I've read through the approach here and it looks good 💯

We're putting some extra effort into type safety recently and this is the next thing on the docket. @DanielFGray let me know if there's any way I can help or support the awesome work you're doing! Happy to collaborate or support in whatever way.

huw:
I don’t think the documentation for this is newbie-friendly. It’s not clear where this declaration is supposed to reside, or whether it’s supposed to change for adapters—are you supposed to override `@remix-run/server-runtime` or `@remix-run/{adapter}`?

huw:
If anyone is visiting this issue after banging your head against the wall, if you choose to keep the module augmentation in a separate file, you also need to import `@remix-run/server-runtime`. The following is in a `global.d.ts` file:

```ts
import "@remix-run/server-runtime";

declare module "@remix-run/server-runtime" {
  interface AppLoadContext {
   key: "value "
  }
}
```

I can then run the following in an `index.tsx`:

```ts
import type { LoaderFunction } from "@remix-run/cloudflare"
```

And everything is typed correctly. (Well, this change also makes `context` optional, which wasn’t necessarily the case in the past—does this mean I should be checking for its existence, or is it safe to assert that it’s present?)

justinwaite:
> If anyone is visiting this issue after banging your head against the wall, if you choose to keep the module augmentation in a separate file, you also need to import `@remix-run/server-runtime`. The following is in a `global.d.ts` file:
> 
> 
> 
> ```ts
> 
> import "@remix-run/server-runtime";
> 
> 
> 
> declare module "@remix-run/server-runtime" {
> 
>   interface AppLoadContext {
> 
>    key: "value"
> 
>   }
> 
> }
> 
> ```
> 
> 
> 
> (You may also replace `@remix-run/server-runtime` with `@remix-run/{adapter}` if you import your loader function type declaration from the adapter package). I can then run the following in an `index.tsx`:
> 
> 
> 
> ```ts
> 
> import type { LoaderFunction } from "@remix-run/cloudflare"
> 
> ```
> 
> 
> 
> And everything is typed correctly. (Well, this change also makes `context` optional, which wasn’t necessarily the case in the past—does this mean I should be checking for its existence, or is it safe to assert that it’s present?)

I'm trying to address the optional issue here: https://github.com/remix-run/remix/pull/3989

PR #1878, opened 2022-02-10
fix(remix-server-runtime): keep memory storage data across cache clears

### LINKED ISSUE #1832: [Bug]: createMemorySessionStorage doesn't actually work in development

- `remix` (Remix core)
- `@remix-run/dev`

1.1.3

16.13.0

### Steps to Reproduce

use createMemorySessionStorage

### Expected Behavior

sessions are preserved for the duration of the development session `remix dev` (or at least, preserved between changes to the underlying source). 

### Actual Behavior

in-memory session storage wiped after each request, even if no source code has changed.

### Elaboration

I expected this API to be usable like:

```ts
export default process.env.NODE_ENV === "production"
    ? createCustomDBStore(cookie)
    : createMemorySessionStore(cookie)
```

Perhaps if this is not the intent and it is only intended for the case of "deployed development" environments, more clearly document that fact.

sheepsteak:
Fixes #1832 

Using an object stored on global the memory session storage will now keep data when the require cache is purged.

I started adding tests for this but I wasn't sure if they were useful. I can add them if needed.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -182,6 +182,7 @@
 - sbernheim4
 - sean-roberts
 - sergiodxa
+- sheepsteak
 - shumuu
 - sidkh
 - silvenon
--- a/packages/remix-server-runtime/sessions/memoryStorage.ts
+++ b/packages/remix-server-runtime/sessions/memoryStorage.ts
@@ -5,6 +5,21 @@ import type {
 } from "../sessions";
 import { createSessionStorage } from "../sessions";
 
+let memoryStorageData: {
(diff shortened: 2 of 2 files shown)


ryanflorence:
@mjackson I'm cool with this if you are. I know our require cache purging leaking into the implementation details of memory storage is unfortunate, but memory storage is really only useful for development anyway, so I think makes sense.

Give me a thumbs up and I'll merge it.

sheepsteak:
Closing in favour of #2768 as I'd deleted my fork and was unable to edit this PR again after.

PR #1883, opened 2022-02-11
feat(create-remix): run init script during project scaffolding

mcansh:
updates `create-remix` to run a "scripts/init.js" setup script if it's in the serverTemplate directory instead of having the end user manually run that script

this script is really only needed for fly as they require globally unique app names, but there's a few extra goodies included in there like automagically creating a `.env` file with the docker `DATABASE_URL` and generating a `SESSION_SECRET`


kentcdodds (approved):
Stellar, thank you for doing this 👍

PR #1888, opened 2022-02-11
chore(useMatches): memoize matches

manzano78:
Memoize the matches returned by `useMatches()` in order to not have a new reference of the array (and of all its items) at every re-rendering. As these matches are made to be looped over, I think it's more performant to memoize them.


packages/remix-react/components.tsx
```diff
@@ -1232,18 +1232,22 @@ export function useBeforeUnload(callback: () => any): void {
  */
 export function useMatches() {
   let { matches, routeData, routeModules } = useRemixEntryContext();
```
ryanflorence, line 1234: are all of these values also memoized?

PR #1893, opened 2022-02-11
chore: fix 'No Response' workflow

MichaelDeBoey:
All workflows failed due to invalid inputs

https://github.com/remix-run/remix/actions/workflows/no-response.yml

CC/ @ryanflorence


ryanflorence:
would it kill them to put an option on their example? I obviously have no idea how github workflows work 😆

ryanflorence:
do we need this on the `dev` branch too for it to work for PRs against dev?

MichaelDeBoey:
I think this should work for all created PRs 🤔

PR #1899, opened 2022-02-11
docs: fix typo in /api/remix

CanRau:
I'm on `v1.1.3` and it's still `unstable_parseMultipartFormData` and not (yet) `parseMultipartFormData`
unless I'm missing something?


CanRau:
Also changed

```tsx
<Form method="post">
```

to

```tsx
<Form method="post" encType="multipart/form-data">
```

as it doesn't seem to work without

PR #1915, opened 2022-02-12
update to actual console log

arange:
The log should only have 'id' and 'username' according to the code in session.server.ts
Updating the log to match the real result


kentcdodds (approved):
Thanks!

PR #1916, opened 2022-02-12
docs(jokes): reduce duplicate code

arange:
We can utilize the helper function created before to get the session instead of implementing it again.


kentcdodds (approved):
Thanks!

Issue #372, opened 2021-11-09
Built-in testing?


kentcdodds:
Might be helpful. My idea is basically a `@remix-run/test` package with all the utilities.

- Built on top of `@testing-library/react` and re-exports everything + a custom `render` that includes all providers our route components have access to when they're run + any utilities to set the loader/action data both on initial render as well as re-rendering with those values.
- `@testing-library/user-event` should be built-in as well
- As @sergiodxa suggested, Jest assertions for Responses
- Potentially some helpers for E2E tests with Cypress and MSW 🤔

sergiodxa:
Aside of a testing setup, it would be helpful to have Jest assertions for Responses so you could test losers and responses easily

andrewgadziksonos:
Would it be possible to spin up a checklist of items that are needed to implement this?  It seems like the community is eager to get this in and with a checklist we could help out!

> so you could test losers and responses easily

I giggled at this @sergiodxa 😂 (assuming you mean `loaders` here)

kentcdodds:
> spin up a checklist of items that are needed

This is probably something I'll need to work on myself to get the package off the ground I think. But I did put a couple high-level ideas. I'm sure the package will evolve.

wladiston:
I wonder if this render could also be used for the storybook rendering system. It would be really nice tweek things like transition and see things changing.

Also, since remix ins't going to be react only (as far I have seen), we'll probably have to split up the package

kentcdodds:
That's a good point. It's possible that we'll want multiple packages. I've had to put this work on hold for a little bit while I work on higher-priority things, but I'll be thinking more on this. I do want to be the one who does this work though so thank you all for your patience.

sergiodxa:
I think at least two libraries for now will make sense, one for server-side testing helpers (like the Jest assertions for response objects or helpers to test loaders/actions that throw responses), and another one with the React helpers (the render function).

That way in a future it could be more packages for more the UI part.

wladiston:
I don't know how MSW have been doing for you guys but for me it has made my tests freaking slow. Any request to the server is taking ~1.2 seconds... On those cases where I need to POST a form to a /reaally/long/route/like/this any of subsequents GETs take more than 2s where normally it takes 20ms.

We'll probably need a way to narrow down the route where the MSW we'll be listening.

kentcdodds:
[...]

PR #1939, opened 2022-02-13
docs(example): add component testing example

kentcdodds:
This is *very* much a work in progress. I'm still evaluating whether mocking is what we want to recommend to people. I'm not sure it is. I'm also experimenting with vitest which I mostly like. We'll see where we land.

Making this PR so folks can reference it and provide feedback.


examples/component-testing/app/routes/users/$userId.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+import UserScreen from "./$userId";
+
+vi.mock("remix", () => {
+  // mock whatever remix APIs you're using
+  return {
+    Form: (props: React.RefAttributes<HTMLFormElement>) => <form {...props} />,
+    useLoaderData: vi.fn(),
+    useLocation: vi.fn(() => ({ key: Math.random().toString(32).slice(2) }))
+  };
+});
```
MichaelDeBoey, line 15: Tbh, I'm also not sure if we should mock it this way 🤔

I think a custom `render` function that takes care of the contexts and has overridable `action`/`links`/`loader`/`meta` would give the team more control on the whole flow and on how things will be tested/mocked

kentcdodds: I agree, it's just a huge challenge to mock the remix context with something sensible. But I do plan on exploring that approach.

MichaelDeBoey:
Just like #2133, I think we can close this one (for now), now that we have #2481 & #372 is closed?

kentcdodds:
Thanks!

PR #1942, opened 2022-02-13
feat(create-remix): add use Tailwind CSS prompt

supachaidev:
Manually installing Tailwind CSS is tedious. This tailwind template can save time.

When users answer "y" to the "Do you want to use Tailwind CSS?" prompt, the CLI setups Tailwind CSS according to the [Install Tailwind CSS with Remix](https://tailwindcss.com/docs/guides/remix) guide.


MichaelDeBoey:
We already have the [Tailwind CSS example](https://github.com/remix-run/remix/tree/main/examples/tailwindcss) which can be looked at together with [Tailwind CSS styling docs](https://remix.run/docs/en/v1/guides/styling#tailwind), which is already enough imo

supachaidev:
In my opinion, it is tedious to manually install Tailwind CSS every time we scaffold a new Remix app. I'm thinking of creating a CLI to automate the task, but I think including a prompt into `create-remix` CLI might be easier for users.

sergiodxa:
I think the team wants to eventually add support for automatically detecting Tailwind and do the setup behind the scenes.

MichaelDeBoey:
This PR can indeed be a precedent for other PRs that include other styling solutions, which is not a preferred solution I think.

As @sergiodxa pointed out
> why would the CLI ask for Tailwind and not those?

It's indeed a bit tedious, but after all it's just running
```sh
npm i -D tailwindcss && npx tailwindcss init
```
and you're good to go and use Tailwind CSS 🤷‍♂️

ryanflorence:
we're still figuring out how to better support the CSS ecosystem right now, but the CLI isn't where we want to do it, but thank you!

ryanflorence:
Also, as a reminder, it's usually best to open a discussion before doing the work in a PR so you don't waste your time writing great code that we eventually don't use :)

PR #1948, opened 2022-02-14

MichaelDeBoey:
Just like the last couple of times

CC/ @kentcdodds

### DIFF
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -1693,7 +1693,7 @@ export default function Home() {
 
   return (
     <div>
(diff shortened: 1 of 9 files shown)


kentcdodds (approved):
We'll probably fight this forever to be honest, but I'll go ahead and merge anyway 😆 thanks!

MichaelDeBoey:
Whenever we there's a PR for https://github.com/yannickcr/eslint-plugin-react/issues/2073, we can add the rule and never do this manually again 😅🙈

PR #1927, opened 2022-02-12
docs(example): Bullmq Task Queue Example

BenMcH:
Adds an example of creating a task queue with bullmq and processing events in the background.

PR #1949, opened 2022-02-14
chore: fix ESLint errors

MichaelDeBoey:
Formatting workflow is stuck again.
I **really** think we should start including formatting in PRs, so we don't have these type of things anymore 🤔

https://github.com/remix-run/remix/runs/5170162309?check_suite_focus=true#step:6:1163

CC/ @ryanflorence

### DIFF
--- a/examples/bullmq-task-queue/app/queues/notifier.server.ts
+++ b/examples/bullmq-task-queue/app/queues/notifier.server.ts
@@ -1,49 +1,52 @@
-import {Queue, Worker, QueueScheduler} from 'bullmq';
-import { redis } from '~/utils/redis.server';
+import { Queue, Worker, QueueScheduler } from "bullmq";
+
+import { redis } from "~/utils/redis.server";
 
 declare global {
-	var __notifierQueue: Queue | undefined;
-	var __notifierWorker: Worker | undefined;
-	var __notifierScheduler: QueueScheduler | undefined;
+  var __notifierQueue: Queue | undefined;
+  var __notifierWorker: Worker | undefined;
+  var __notifierScheduler: QueueScheduler | undefined;
 }
 
-let queue: Queue;
-let worker: Worker;
-let scheduler: QueueScheduler;
-
-const QUEUE_NAME = 'notifier';
+const QUEUE_NAME = "notifier";
 
 type QueueData = {
-	emailAddress: string;
-}
+  emailAddress: string;
+};
 
 // Bullmq queues are the storage container managing jobs.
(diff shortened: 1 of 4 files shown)


examples/bullmq-task-queue/app/queues/notifier.server.ts
```diff
@@ -1,49 +1,52 @@
...
-	console.log(`Sending email to ${job.data.emailAddress}`);
-
-	// Delay 1 second to simulate sending an email, be it for user registration, a newsletter, etc.
-	await new Promise((resolve) => setTimeout(resolve, 1000));
-
-	console.log(`Email sent to ${job.data.emailAddress}`);
-}, {
-	connection: redis,
-}));
+const worker: Worker =
```
MichaelDeBoey, line 28: @BenMcH Do we need this variable?

Only `queue` is exported/used.

Asking you because you created this example in #1927

examples/bullmq-task-queue/app/queues/notifier.server.ts
```diff
@@ -1,49 +1,52 @@
...
 // Jobs may be queued in a delayed or waiting state, but the scheduler's
 // job is to eventually move them to an active state.
-scheduler = global.__notifierScheduler || (global.__notifierScheduler = new QueueScheduler(QUEUE_NAME, {
-	connection: redis,
-}));
-
-export {
-	queue
-}
+const scheduler: QueueScheduler =
```
MichaelDeBoey, line 48: @BenMcH Do we need this variable?

Only `queue` is exported/used.

Asking you because you created this example in #1927

kentcdodds (approved):
Thanks! Still interested in the answers to the questions.

May be handy to have the linting run on PRs. Formatting would probably be an issue though.

BenMcH:
@MichaelDeBoey you raise some good points. I don't think these variables need to be defined in our queue files necessarily. The worker and scheduler aren't super useful to have reference to in my experience, but they're important to the functionality. Perhaps these could be stored exclusively in the `global` object and reworked into something like this:


```js
  global.__notifierWorker = global.__notifierWorker || new Worker(...)
```

The same could be done with the scheduler. What are your thoughts around this approach? cc @kentcdodds since you were also interested in these questions.

I agree that running tests and/or linting on PR would be fantastic for infrequent contributors. I would've gladly taken the time to fix these!

MichaelDeBoey:
@BenMcH another approach would be to export these variables, so they're available whenever someone needs them.

I would just remove them tbh, as we try to have our examples as minimal as possible and they currently don't have any added value

LANCENET (commented):
Done

PR #1969, opened 2022-02-15
fix(create-remix/stacks): dont return user password from functions

mcansh:
- [x] Fly
- [x] Arc


packages/create-remix/templates/arc-stack/app/routes/login.tsx
```diff
@@ -48,7 +48,9 @@ const action: ActionFunction = async ({ request }) => {
     );
   }
 
-  return createUserSession(request, user, "/");
+  console.log({ user });
+
```
kentcdodds, line 52:
```suggestion
```

packages/create-remix/templates/arc-stack/README.md
```diff
@@ -57,6 +58,7 @@ You can find the CloudFormation template that Architect generated for you in the
 
 To find it on AWS, you can search for [CloudFormation][cloudformation] (make sure you're looking at the correct region!) and find the name of your stack (the name is a PascalCased version of what you have in `app.arc`, so by default it's RemixAwsStackStaging) that matches what's in `app.arc`, you can find all of your app's resources under the "Resources" tab.
 
+[aws_cli]: https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
```
kentcdodds, line 61: We should have all of these links inline.

In the docs we don't want to do that because it's easier to do translations if we don't.

But here, most people will be looking at the source in their editor and we don't want them to have to scroll down to the bottom to find the link. We're not going to do translations of this either.

packages/create-remix/templates/fly-stack/prisma/schema.prisma
```diff
@@ -11,13 +11,29 @@ model User {
...
   password String
-  notes    Note[]
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  userData UserData?
+}
+
+model UserData {
```
kentcdodds, line 21: Let's call this one "User" and call the other table "Auth"

@ryanflorence, is there a more common convention for this? I'm not a fan of "UserData" as a table name in a database 😅

kentcdodds (changes requested):
Super! Got a couple notes.

kentcdodds:
These are great improvements 👍

packages/create-remix/templates/fly-stack/prisma/schema.prisma
mcansh: same really, almost went with `Account` but i couldn't decide which one was the account and what was the user haha

jensmeindertsma: What about just storing the data on the user model directly and then omitting the password from the returned object by createUser?

packages/create-remix/templates/arc-stack/app/models/user.ts
```diff
@@ -1,6 +1,11 @@
 import arc from "@architect/functions";
 import bcrypt from "bcryptjs";
 
+function cleanUser(user: any) {
```
kentcdodds, line 4: Do we still need this for arc? Can we take the same approach here that we do for fly?

mcansh: won't be needed after i do that

packages/create-remix/templates/fly-stack/prisma/schema.prisma
```diff
@@ -8,16 +8,30 @@ generator client {
...
+  updatedAt DateTime @updatedAt
+
+  password Password?
   notes    Note[]
 }
 
+model Password {
+  user     User   @relation(fields: [userId], references: [id])
+  userId   String @unique
+  password String
```
kentcdodds, line 24: Calling this `hash` instead of `password` allows the code elsewhere to read: `user.password.hash` instead of `user.password.password` which I think is a bit nicer :)

```suggestion
  hash         String
```

packages/create-remix/templates/fly-stack/prisma/schema.prisma
```diff
@@ -8,16 +8,30 @@ generator client {
...
-  id       String @id @default(cuid())
-  email    String @unique
-  password String
+  id    String @id @default(cuid())
+  email String @unique
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  password Password?
```
kentcdodds, line 17: This should not be optional right?

```suggestion
  password Password
```

packages/create-remix/templates/fly-stack/app/routes/join.tsx
```diff
@@ -1,76 +1,138 @@
...
           <span>Password</span>
-          <input name="password" type="password" autoComplete="new-password" />
+          <input
+            name="password"
+            type="password"
+            autoComplete="new-password"
+            aria-invalid={actionData?.errors?.password ? true : undefined}
+            aria-errormessage={
+              actionData?.errors.password ? "password-error" : undefined
+            }
```
kentcdodds, line 111: @ryanflorence and @chaance, do we need this logic or could it simply be

```suggestion
            aria-errormessage="password-error"
```

I've done the ternary before but I'm not sure whether it's necessary. If there's no `password-error` element is that ok to have it here anyway?

packages/create-remix/templates/fly-stack/app/routes/join.tsx
```diff
@@ -1,76 +1,138 @@
...
+    );
+  }
 
   const user = await createUser(email, password);
 
   if (!user) {
-    return redirect("/login");
+    return json<ActionData>(
+      { errors: { email: "Invalid email or password" } },
+      { status: 400 }
```
kentcdodds, line 54: I think these should be [`422`s](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) actually... @mjackson?

kentcdodds (changes requested):
This is very much improved! Thanks for iterating so much on it! I've got some ideas and comments.

packages/create-remix/templates/fly-stack/app/routes/join.tsx
mcansh: "The aria-errormessage should only be used when the value of an object is not valid;" https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage

packages/create-remix/templates/fly-stack/app/routes/join.tsx
mcansh: the description of a [400](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) fits better imo as it's "malformed request syntax"

packages/create-remix/templates/fly-stack/prisma/schema.prisma
mcansh: yes! but, Prisma won't allow that 

> The relation field `password` on Model `User` is required. This is no longer valid because it's not possible to enforce this constraint on the database level. Please change the field type from `Password` to `Password?` to fix this.

mcansh: we might be able to make a [composite id](https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#composite-ids), but i think that requires at least 2 properties to be on a model, we just have "hash" on it

PR #1973, opened 2022-02-15
fix(remix-react): improve `useMatches`' type safety

gustavoguichard:
### The Problem

When using `useMatches` in combination with `handle` we currently have to work with the `any` type.

I'd like a way to explicitly type my `handle` functions and reuse that type on my `export let handle: T` signature, having 1 type that can be used in both of these places.

Let's say I'm doing a `<Breadcrumbs />` component
```tsx
function Breadcrumbs() {
  let matches = useMatches()
  let breadcrumbs = matches.filter((match) => match.handle.breadcrumb)
  return (
    <ol>
      {breadcrumbs.map((match, index) => (
        <li key={index}>
          {match.handle.breadcrumb(match)}
        </li>
      ))}
    </ol>
  )
}
```

Right now, the type of `match` on the third line is inferred from Remix's code as:
```ts
interface InferredType {
  id: string;
  pathname: string;
  params: Params<string>;
  data: RouteData;
  handle: any;
}
```

So on the component code above - line 3 - Typescript does not know that either the `match.handle` and the `match.handle.breadcrumb` properties can be `undefined` because `handle: any`. This is preventing TS from saving me by suggesting optional chaining:
```tsx
let breadcrumbs = matches.filter((match) => match.handle?.breadcrumb)
```

I also can't type the breadcrumb function.

### Proposed solution

By enabling a type param to be sent to `useMatches` and exporting the `Match` type from Remix's source we'll able to fully type our handle functions in a type safe way like so:
```tsx
import type { Match } from '@remix-run/react/components'

type HandleWithBreadcrumb = {
  breadcrumb?: (match: Match<any>, isCurrent: boolean) => React.ReactNode
}

function Breadcrumbs() {
  let matches = useMatches<HandleWithBreadcrumb>()
  let location = useLocation()
  let breadcrumbs = matches.filter((match) => match.handle?.breadcrumb)
  return (
    <ol>
      {breadcrumbs.map((match, index) => (
        <li key={index}>
          {match.handle.breadcrumb(match, location.pathname === math.pathname)}
        </li>
      ))}
    </ol>
  )
}
```

And we will also be able to use that Type in our `handle` export:
```tsx
export let handle: HandleWithBreadcrumb = {
  breadcrumb: (match, isCurrent) => (
    <BreadcrumbItem isCurrent={isCurrent} to={match.pathname}>
      Settings
    </BreadcrumbItem>
  ),
}
```

And if no type param is given to `useMatches` it will default to `any` thus preserving the current API.

### Some doubts I have
I couldn't figure out what is the current pattern for exporting interfaces so I'm exporting it right after the declaration of `useMatches` function.

I'm also not sure it should be exported from `@remix-run/react/components`. What do you folks think?


gustavoguichard:
To illustrate and explain my motivation, after I installed remix from my branch and used the new types I ended up with the following diff:

Captura de Tela 2022-02-15 às 16 17 56

MichaelDeBoey:
@gustavoguichard Could you please rebase your branch onto latest `dev` & resolve conflicts?

gustavoguichard:
Sure @MichaelDeBoey is there motivation to merge this PR tho? It'd be a pleasure

MichaelDeBoey:
@gustavoguichard I'm only making sure this PR is up to date & the team will decide whether they want to merge/decline it whenever they have time to look into it.

gustavoguichard:
I'll be sending the PR again soon instead of dealing with all the conflicts ;)

Issue #896, opened 2021-12-05
[Bug]: "Prop `href` did not match". when using tailwindcss


philldev:
### Which Remix packages are impacted?

- [X] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [X] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [X] `@remix-run/vercel`

### What version of Remix are you using?

1.0.6

### What version of Node are you using? Minimum supported version is 14.

16.13.1

### Steps to Reproduce

1. create remix app.
2. add tailwindcss from this guide https://remix.run/docs/en/v1/guides/styling#tailwind.
3. add tailwind.css to app/root.tsx.
```
import styles from './tailwind.css'

export function links() {
	return [{ rel: 'stylesheet', href: styles }]
}
```
4. start dev server

### Expected Behavior

No warning message appears on browser console

### Actual Behavior

This warning message appears on console
```
Warning: Prop `href` did not match. Server: "/build/_assets/tailwind-DGHXG3YF.css" Client: "/build/_assets/tailwind-JUD5S62U.css"
    at link
    at Links (http://localhost:3000/build/_shared/chunk-KOXENSZ3.js:3042:7)
    at head
    at html
    at Document (http://localhost:3000/build/root-WYXGDNDR.js:55:3)
    at App
    at RemixCatchBoundary (http://localhost:3000/build/_shared/chunk-KOXENSZ3.js:1476:10)
    at RemixErrorBoundary (http://localhost:3000/build/_shared/chunk-KOXENSZ3.js:1415:5)
    at RemixRoute (http://localhost:3000/build/_shared/chunk-KOXENSZ3.js:2905:3)
    at Routes (http://localhost:3000/build/_shared/chunk-KOXENSZ3.js:2888:7)
    at Router2 (http://localhost:3000/build/_shared/chunk-KOXENSZ3.js:457:21)
    at RemixCatchBoundary (http://localhost:3000/build/_shared/chunk-KOXENSZ3.js:1476:10)
    at RemixErrorBoundary (http://localhost:3000/build/_shared/chunk-KOXENSZ3.js:1415:5)
[...]

PR #1975, opened 2022-02-15
chore: revert esbuild version

jacob-ebey:
https://github.com/evanw/esbuild/issues/2019


kentcdodds (approved):
Worked for me 👍

kiliman:
Ugh. Too bad you had to revert. Because 0.13 still has that bug with the file hash not updating when the content changes, which causes a headache with generated assets like Tailwind. #896

kentcdodds:
Hopefully we get on with our regular Monday releases so it won't be long before we get this :)

PR #1977, opened 2022-02-15
feat(remix-dev): support worker package exports for Cloudflare runtime

### LINKED ISSUE #2299: [React 18] react-dom/server resolution on worker runtimes

1.2.3

### Steps to Reproduce

- Install `react-dom@rc react@rc` on project using a worker runtime (such as Cloudflare Workers). I tested this on Cloudflare Pages.
- Use `renderToReadableStream` on `entry.server.js`. Mine looks like this:
```typescript
  // ... something ...
  const stream = await renderToReadableStream(
  );

  return new Response(stream, {
    status: responseStatusCode,
    headers: responseHeaders,
  });
```
- Fails on Miniflare because Node.js globals aren't found.

### Expected Behavior

It should load `server.browser.js` as it's specified on the `worker` exports field of `react-dom`.

Ok, I know this API isn't even released yet and is not meant to be used right now but I'm guessing Remix would likely adopt React 18 so I thought this is worth investigating.

This happens because esbuild has [no target setting for Workers yet](https://esbuild.github.io/api/#platform) so Remix currently uses [`neutral`](https://github.com/remix-run/remix/blob/5b8a0ce0aa0201aa2402fc41405ffbe89605963b/packages/remix-dev/config.ts#L292) which loads default export.

However, `react-dom` exports `server.node.js` by default which uses some Node.js-specific APIs such as Buffer, which isn't available on Cloudflare nor Miniflare. I'm not quite sure why this has worked before (maybe Node.js API calls were DCE'd because it isn't used for `renderToString`?), but when it is used with `renderToReadableStream` it doesn't work as intended.

This may need to be addressed on esbuild and thus may be considered as an upstream issue.

##### References
- https://github.com/facebook/react/pull/23252
- https://github.com/evanw/esbuild/issues/1189

### Actual Behavior

```
[pages:err] ReferenceError: Buffer is not defined
    at stringToPrecomputedChunk (C:\Users\basix\Projects\basixpage\node_modules\react-dom\cjs\react-dom-server.node.development.js:117:3)
    at C:\Users\basix\Projects\basixpage\node_modules\react-dom\cjs\react-dom-server.node.development.js:1674:25
    at node_modules/react-dom/cjs/react-dom-server.node.development.js (C:\Users\basix\AppData\Local\Temp\functionsWorker.js:16223:9)
    at __require (C:\Users\basix\AppData\Local\Temp\functionsWorker.js:42:50)
    at node_modules/react-dom/server.node.js (C:\Users\basix\Projects\basixpage\node_modules\react-dom\server.node.js:9:7)
    at __require (C:\Users\basix\AppData\Local\Temp\functionsWorker.js:42:50)
    at C:\Users\basix\Projects\basixpage\app\entry.server.tsx:1:40
    at SourceTextModule.evaluate (node:internal/vm/module:224:23)
    at VMScriptRunner.runAsModule (C:\Users\basix\Projects\basixpage\node_modules\@miniflare\runner-vm\src\index.ts:38:18)
    at VMScriptRunner.run (C:\Users\basix\Projects\basixpage\node_modules\@miniflare\runner-vm\src\index.ts:82:17)
Could not start Miniflare.
```

Your message may vary but it would likely fail on `react-dom-server.node.something.js`.

edmundhung:
There are some packages define custom package exports dedicated for worker, e.g. https://github.com/maraisr/diary/blob/main/package.json#L21-L24

It would be really helpful if we can support this by setting up conditions specifically for `cloudlfare-workers` and `cloudflare-pages` serverBuildTarget.

Closes #2299


packages/remix-dev/compiler.ts
```diff
@@ -417,15 +421,11 @@ async function createServerBuild(
       entryPoints,
       outfile: config.serverBuildPath,
       write: false,
+      conditions: isCloudflareWorkersOrPages ? ['worker'] : [],
```
jacob-ebey, line 424: I believe we need the other "standard" fields listed as well, along with the default being "undefined" otherwise they won't ever be resolved.

jacob-ebey (changes requested):

packages/remix-dev/compiler.ts
edmundhung: Thanks for reviewing 👍🏼  I marked the default to be `undefined` now. But I am not sure what the "standard" fields to be put here.

On esbuild documentation about [conditions](https://esbuild.github.io/api/#how-conditions-work), it mentions `there are five conditions with special behavior that are built in to esbuild, and cannot be disabled`. While on the [commit](https://github.com/evanw/esbuild/commit/1a458e0bf581737e4962e8d33f0af01672b708e4#diff-ea8af26844b5be484f625fbe35bfe3be845e32d4a87d01b8dd9f8b3f3b3e6a37R740) that implement custom conditions for "exports" seems to be doing an append.

Do you think if `['worker']` could work without any standard fields?

jacob-ebey: Yep, looks like are good to go with this. I gave it a run-through locally but the part I don't like is that it's based on the order of the published package. This means that if the package.json for the module looks like this:
```json
{
  "name": "fake-module",
  "version": "1.0.0",
  "main": "./require.js",
  "module": "./module.js",
  "exports": {
    "require": "./require.js",
    "import": "./import.js",
    "worker": "./worker.js",
    "default": "./default.js"
  }
}
```
it will still resolve the "import" version as it's based on the order of the entries in the "exports" object. (this is just how the ecosystem is though 😮‍💨)
It's up the the package author to make sure their exports look more like:
```json
{
  "name": "fake-module",
  "version": "1.0.0",
  "main": "./require.js",
  "module": "./module.js",
  "exports": {
    "worker": "./worker.js",
    "require": "./require.js",
    "import": "./import.js",
    "default": "./default.js"
  }
}
```

jacob-ebey (approved):

jacob-ebey:
I'm going to add a test here for this and we'll get it merged.

packages/remix-dev/compiler.ts
```diff
@@ -397,6 +397,11 @@ async function createServerBuild(
     };
   }
 
+  let isCloudflareWorkersOrPages =
+    !!config.serverBuildTarget &&
+    ["cloudflare-workers", "cloudflare-pages"].includes(
+      config.serverBuildTarget
+    );
```
MichaelDeBoey, line 404:
```suggestion
  let isCloudflareRuntime =
    ["cloudflare-pages", "cloudflare-workers"].includes(
      config.serverBuildTarget ?? ""
    );
```

packages/remix-dev/compiler.ts
```diff
@@ -417,15 +422,12 @@ async function createServerBuild(
       entryPoints,
       outfile: config.serverBuildPath,
       write: false,
+      conditions: isCloudflareWorkersOrPages ? ["worker"] : undefined,
```
MichaelDeBoey, line 425:
```suggestion
      conditions: isCloudflareRuntime ? ["worker"] : undefined,
```

packages/remix-dev/compiler.ts
```diff
@@ -417,15 +422,12 @@ async function createServerBuild(
...
       platform: config.serverPlatform,
       format: config.serverModuleFormat,
       treeShaking: true,
       minify:
-        options.mode === BuildMode.Production &&
-        !!config.serverBuildTarget &&
-        ["cloudflare-workers", "cloudflare-pages"].includes(
-          config.serverBuildTarget
-        ),
+        options.mode === BuildMode.Production && isCloudflareWorkersOrPages,
```
MichaelDeBoey, line 430:
```suggestion
        options.mode === BuildMode.Production && isCloudflareRuntime,
```

packages/remix-dev/compiler.ts
```diff
@@ -417,15 +422,12 @@ async function createServerBuild(
...
       mainFields:
         config.serverModuleFormat === "esm"
           ? ["module", "main"]
```
MichaelDeBoey, line 433: #2076 also changed `mainFields`. Do we want to do that in this PR too?

```suggestion
      mainFields: isCloudflareRuntime
        ? ["browser", "module", "main"]
        : config.serverModuleFormat === "esm"
          ? ["module", "main"]
```

packages/remix-node/package.json
```diff
@@ -15,7 +15,7 @@
     "@remix-run/server-runtime": "1.1.3",
     "@types/busboy": "^0.3.1",
     "@types/node-fetch": "^2.5.12",
-    "@web-std/file": "^3.0.0",
+    "@web-std/file": "^3.0.2",
```
MichaelDeBoey, line 18: Is this update necessary for this PR?
If not, please revert changes to `package.json` & `yarn.lock`

MichaelDeBoey (changes requested):
Closed #2076 in favour of this one.

Do we also want to change `mainFields`, like @aiji42 did there?

MichaelDeBoey:
@edmundhung Can you please rebase your branch onto latest `dev` & fix conflicts?

packages/remix-node/package.json
edmundhung: It was added by @jacob-ebey when he is verifying the changes for me. I will double check if it is still needed with the latest setup.

integration/helpers/create-fixture.tsx
```diff
@@ -362,6 +363,7 @@ export async function createFixtureProject(init: FixtureInit): Promise<string> {
     init.template ? init.template : "remix"
   );
   let projectDir = path.join(TMP_DIR, Math.random().toString(32).slice(2));
+  let isCloudflareRuntime = init.template === 'cloudflare-pages' || init.template === 'cloudflare-workers';
```
MichaelDeBoey, line 366:
```suggestion
  let isCloudflareRuntime = ["cloudflare-pages", "cloudflare-workers"].includes(init.template);
```

MichaelDeBoey (changes requested):

packages/remix-dev/compiler.ts
edmundhung: @jacob-ebey Do you wanna get this in as well? I think it will be quite helpful but I observed an unexpected behaviour with `esbuild` after expanding the current cf-compiler test suite:

1. If the 3rd party package supports cjs only, i.e.
```json
{
    "name": "browser-pkg",
    "version": "1.0.0",
    "main": "./node-cjs.js",
    "browser": "./browser-cjs.js"
}
```

it bundles `./browser-cjs.js` correctly

2. if the 3rd party package is esm only, i.e.

```json
{
    "name": "browser-pkg",
    "version": "1.0.0",
    "type": "module",
    "main": "./node-esm.js",
    "browser": "./browser-esm.js"
}
```

it bundles `./browser-esm.js`  correctly

3. if the 3rd party package support both cjs and esm, i.e.
```json
{
    "name": "browser-pkg",
    "version": "1.0.0",
    "main": "./node-cjs.js",
    "module": "./node-esm.mjs",
    "browser": {
        "./node-cjs.js": "./browser-cjs.js",
        "./node-esm.mjs": "./browser-esm.mjs"
    }
}
```
it bundles `./node-esm.mjs` instead of `./browser-esm.mjs` this time and changing the platform to `browser` could fix the problem (which was set to `neutral` for cloudflare & deno).

edmundhung:
@MichaelDeBoey Since the mainFields change need a bit more thought as I mentioned earlier. We can consider merging this first as this should already help unblocking people using React 18 on Cloudflare Workers and Pages. I can create another PR after this with test cases explaining the problem and concerns there.

MichaelDeBoey (approved):

packages/remix-dev/compiler.ts
jacob-ebey: @edmundhung, let's change the platform to browser for CF and get this in as well. Looking at the wrangler v2 setup they use the browser: https://github.com/cloudflare/wrangler2/blob/main/packages/wrangler/pages/functions/buildWorker.ts.

jacob-ebey (changes requested):

packages/remix-dev/compiler.ts
edmundhung: @jacob-ebey I was trying to reproduce the error when platform is setill set to `netural` but it looks like the tests pass anyway 😓 Do we still wanna update it to `browser`?

omarkhatibco:
Hi @jacob-ebey is it included with v1.3.4 ?

PR #1979, opened 2022-02-16
bug: NavLink applies custom styles and class names (specified with a function) to all NavLinks regardless of the current route

ariofrio:
This is a bug report with an integration test as requested [in the docs](https://remix.run/docs/en/v1/pages/contributing#think-you-found-a-bug).

This seems to affect every usage of `NavLink` that requires a custom class name or styles. This is particularly a problem when using Tailwind, which is [recommended by the docs](https://remix.run/docs/en/v1/guides/styling#tailwind) as the most popular solution. The most straightforward way of using Tailwind is like this:

```
<NavLink to="/somewhere" className={(isActive) => isActive ? "underline" : ""}>Somewhere</NavLink>
```

But this bug causes all navigation links to appear active when this is done.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -188,3 +188,4 @@
 - guerra08
 - wladiston
 - supachaidev
+- ariofrio
--- a/integration/navlink-test.ts
+++ b/integration/navlink-test.ts
@@ -1,0 +1,122 @@
+import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
+import type { Fixture, AppFixture } from "./helpers/create-fixture";
+
+describe("NavLink", () => {
+  let fixture: Fixture;
+  let app: AppFixture;
+
+  beforeAll(async () => {
+    fixture = await createFixture({
+      files: {
+        "app/root.jsx": js`
+          import {
+            NavLink,
+            Outlet,
+            Scripts
+          } from "remix";
+
+          export default function App() {
+            return (
+              <html>
+                <head />
+                <body>
+                  <header>
+                    <nav>
+                      <NavLink
+                        to="/"
+                        end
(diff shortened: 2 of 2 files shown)


nktpro:
@ariofrio The parameter of the function is an object with an `isActive` `boolean` property, not a `boolean` itself, i.e your example would be:

```jsx
<NavLink to="/somewhere" className={({isActive}) => isActive ? "underline" : ""}>Somewhere</NavLink>
```

I spent 15 minutes pulling some hairs out tripping over this, in the exact same way you did, then I realized the same mistake. Granted, the API is counter-intuitive IMO; not sure why it isn't just a `boolean` instead.

ariofrio:
Wow, thanks for the tip @nktpro. I definitely spent a lot more than 15 minutes pulling my own hairs over this! I could have sworn that when I hovered over the variable in Visual Studio Code it told me that it was a boolean, but sure enough when I tried it today the type of `isActive` was `{isActive: boolean}` instead.

Looks like `typescript-eslint` has a [couple](https://typescript-eslint.io/rules/strict-boolean-expressions) of [rules](https://typescript-eslint.io/rules/no-unnecessary-condition) that could help with this, but the TypeScript compiler itself won't catch it. I particularly like `no-unnecessary-condition` in this case, but neither is included in the recommended set so they need to be enabled manually.

I agree that the API is counter-intuitive. I imagine that `isActive` is an object so that in the future other options can be passed down, but this doesn't seem like a case where that is likely. This informed my intuition so that I originally didn't even consider the possibility that the parameter is actually an object. Presumably the authors have some ideas for future expansion.

ariofrio:
So this is not a bug—at worst, it's a counter-intuitive API. Closing.

PR #1983, opened 2022-02-16
chore: make stacks a hidden option

mcansh:
they're not quite ready, but we can keep them around while we continue to cook them up


kentcdodds (approved):
Thanks! Whoops 😅

PR #1985, opened 2022-02-16
chore(create-remix/stacks): bring fly changes to arc

mcansh:
for real this time, we should be good


kentcdodds (approved):
Solid 👍

PR #1990, opened 2022-02-16
Add support for JSON payload in fetcher.submit()

kiliman:
This commit updates the target parameter to accept a JSON object or string. It adds an optional `json` property to indicate payload will be JSON if true. This will set `content-type: application/json` header.

Also includes test and updated documentation.

See Discussion https://github.com/remix-run/remix/discussions/1959


packages/remix-react/transition.ts
```diff
@@ -100,7 +100,7 @@ export interface TransitionManagerInit {
 export interface Submission {
   action: string;
   method: string;
-  formData: FormData;
+  formData: FormData | string;
```
jacob-ebey, line 103: I think I'd rather have a "json" property instead of re-using formData. FormData is VERY specific in what it is and is not at all JSON.

jacob-ebey (changes requested):

packages/remix-react/transition.ts
kiliman: I get that. I was just trying to be minimal in my PR. I'll make the change you requested then.

kiliman: I've updated the Submission interface.

Issue #1993, opened 2022-02-17
Deno template errors on dynamic require()


mjackson:
### What version of Remix are you using?

1.2.1

### Steps to Reproduce

`npx create-remix@latest` and choose Deno as deploy target. Then `npm run dev` and you'll see the error.

### Expected Behavior

The dev server boots.

### Actual Behavior

```
error: Uncaught Error: Dynamic require of "https://deno.land/std/node/stream/mod.ts" is not supported
  throw new Error('Dynamic require of "' + x + '" is not supported');
        ^
    at file:///Users/michael/Projects/my-remix-app/build/index.js:30:9
    at file:///Users/michael/Projects/my-remix-app/build/index.js:2207:22
    at node_modules/react-dom/cjs/react-dom-server.node.development.js (file:///Users/michael/Projects/my-remix-app/build/index.js:5510:9)
    at __require2 (file:///Users/michael/Projects/my-remix-app/build/index.js:48:44)
    at node_modules/react-dom/server.node.js (file:///Users/michael/Projects/my-remix-app/build/index.js:5523:24)
    at __require2 (file:///Users/michael/Projects/my-remix-app/build/index.js:48:44)
    at node_modules/react-dom/server.js (file:///Users/michael/Projects/my-remix-app/build/index.js:5533:22)
    at __require2 (file:///Users/michael/Projects/my-remix-app/build/index.js:48:44)
    at file:///Users/michael/Projects/my-remix-app/build/index.js:7021:33
```

machour:
Fixed by #1997

PR #1997, opened 2022-02-17
fix: use browser polyfills for deno node builtins

jacob-ebey:
Esbuild can't hoist `require("xyz")` calls to `import "xyz"`, this leads to an issue every time you try to re-map a require'd version in external code to a URL.

Closes: #1993

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -407,7 +407,7 @@ async function createServerBuild(
   ];
 
   if (config.serverPlatform !== "node") {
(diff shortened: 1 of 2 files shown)


mjackson:
Is there any way we can test the server bundle output here? Like running a build and then scan the server build to assert the import comes from the right place?

Eventually when we have a Deno stack I'd love to add Deno Deploy to our prerelease tests and actually do a deploy to verify everything works. But seems like we could at least have some tests around this part for now.

PR #1998, opened 2022-02-17
chore: update esbuild to latest

jacob-ebey:
The "issue" that was raised isn't a valid issue with esbuild and should be addressed in user-land (i.e, using libraries as intended by the author). Essentially a little bit of hand holding Esbuild did here has been removed to fall inline with other bundlers in the ecosystem.

Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/remix-dev/package.json
+++ b/packages/remix-dev/package.json
@@ -19,7 +19,7 @@
     "@remix-run/server-runtime": "1.2.1",
     "cacache": "^15.0.5",
     "chokidar": "^3.5.1",
-    "esbuild": "0.13.14",
+    "esbuild": "0.14.22",
     "exit-hook": "2.2.1",
     "fs-extra": "^10.0.0",
     "get-port": "^5.1.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4797,113 +4797,125 @@ es-to-primitive@^1.2.1:
     is-date-object "^1.0.1"
     is-symbol "^1.0.2"
 
-esbuild-android-arm64@0.13.14:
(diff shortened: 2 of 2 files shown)


kentcdodds (commented):
What are we doing to do about people for whom this will be a breaking change? It'll be a lot of people. This really should be a v2 thing I think. Cc @mjackson

jacob-ebey:
I don't agree with that. It's improper use of the libraries, it should be broken already. I.e attempting to import "default" when no "default" is exported.

kentcdodds:
I can be convinced. Would be good to find a codemod anyway 👍

kiliman:
Yeah, I think it's still early in adoption that something like this isn't really a "breaking" change, especially when they were technically not doing it correctly to begin with. I don't think Remix should perpetuate incorrect practices.

Besides, the actual fix is pretty simple. They would literally need to add `* as` to their imports for the broken ones. Hopefully the error message is clear.

This should be called out in the Release Notes.

kentcdodds:
Yeah, a codemod is probably unnecessary. I'd be surprised to find a Remix codebase larger than my own this early after the official release and it won't take me much time to address this issue. I'm convinced 👍

kentcdodds (approved):
Super 👍 Thanks for keeping us current 👏

kentcdodds:
cc @chaance, you'll want to make note of this in the release notes 👆

jacob-ebey:
Evan reference: https://github.com/evanw/esbuild/issues/2019#issuecomment-1041980825

PR #2013, opened 2022-02-17
fix(node): add source-map-support to base node pkg

jacob-ebey:
Since we are now inlining source-maps we can rely on the source-map-support library instead of having a custom implementation that has to be handed manually in our shared runtime.

image
image


mjackson:
Love it. Thank you!

PR #2018, opened 2022-02-18
chore(remix-react): use `document` instead of `AbortController` to check for bad `fetcher.load` call

zachdtaylor:
This is a fix for a PR I made a while back, #1416. I created that PR while on node 14, which doesn't have `AbortController` defined as a global object. In later versions of node, however, `AbortController` is defined and thus the check I created doesn't work. This PR replaces `AbortController` with `window`, which is a more reliable way of determining if we are in node or in the browser.

### DIFF
--- a/packages/remix-react/transition.ts
+++ b/packages/remix-react/transition.ts
@@ -904,7 +904,7 @@ export function createTransitionManager(init: TransitionManagerInit) {
     key: string,
     match: ClientMatch
   ) {
-    if (typeof AbortController === "undefined") {
+    if (typeof window === "undefined") {
       throw new Error(
         "handleLoaderFetch was called during the server render, but it shouldn't be. " +
           "You are likely calling useFetcher.load() in the body of your component. " +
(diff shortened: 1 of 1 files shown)


packages/remix-react/transition.ts
```diff
@@ -790,7 +790,7 @@ export function createTransitionManager(init: TransitionManagerInit) {
     key: string,
     match: ClientMatch
   ) {
-    if (typeof AbortController === "undefined") {
+    if (typeof window === "undefined") {
```
jacob-ebey, line 793: Please use document instead of window. Deno is a runtime we are trying to support and it has window defined in the server but not document.

jacob-ebey (changes requested):

machour (approved):

chaance:
This code path has since moved over to React Router since v6.4, and we do a more robust check there to be sure. https://github.com/remix-run/react-router/blob/main/packages/router/router.ts#L1487

PR #2019, opened 2022-02-18
fix: strip ".client" files from server bundle

jacob-ebey:
The regex is wrong? Not sure how this got changed or stopped working between release 🤷‍♀️

Closes: #

- [ ] Docs
- [x] Tests


kentcdodds (approved):
👍

PR #2021, opened 2022-02-18
docs: update blog.md

idanen:
npm is not yarn 😉

Closes: #

- [x] Docs
- [ ] Tests

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -99,6 +99,7 @@
 - hzhu
 - IAmLuisJ
 - ianduvall
+- idanen
 - imzshh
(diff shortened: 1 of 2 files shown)


joelbowen:
`npm add` is valid as an alias for `npm install` https://docs.npmjs.com/cli/v7/commands/npm-install – though if it's probably not super common.

idanen:
> `npm add` is valid as an alias for `npm install` https://docs.npmjs.com/cli/v7/commands/npm-install – though if it's probably not super common.

TIL. Thanks

PR #2027, opened 2022-02-18
Relativize route modules to make builds deterministic

### LINKED ISSUE #2024: `-[hash]` changes based on the absolute file path

1.2.1

### Steps to Reproduce

1. Create two remix projects (via `npx create-remix@latest`)
2. `npm run build` them both
3. Compare the contents of `public/build`

### Expected Behavior

The hashes and contents of all files should be the same.

For context, I need to run my server and browser builds in two different places because Reasons™. If they are not deterministic, the server build will expect different filenames than what the browser build creates, which results in 404s and a broken app.

### Actual Behavior

1. The `public/build/routes/index-[hash].js` files are identical but have different hashes.
2. The `public/build/root-[hash].js` files are identical but have different hashes.
3. The `public/build/manifest-[hash].js` files have different contents and different hashes (due to 1 and 2)

I've confirmed the behavior is deterministic based on the project directory. I.e. you can `mv` your projects around to swap their locations; if you build again you'll get the hashes/files originally seen in the other project.

In a larger remix app, I've also observed minification differences (different shortened variable names), likely due to the same issue.

While this may be an esbuild bug, it could also be caused by remix' virtual modules including the absolute path ([more context](https://github.com/evanw/esbuild/issues/1530#issuecomment-900801686) from a similar bug report).

I'll take a stab at fixing this 😄 🤞

jenseng:
Closes #2024

- [x] Tests

If virtual modules have non-deterministic paths or content (e.g. due to importing from other absolute paths), the input is technically different, and deterministic build output is not guaranteed.

Depending on how you build/deploy (e.g. if you need to build and deploy your server separately from your browser build), this can result in a broken app, since the server and browser manifests may differ (i.e. due to different fingerprints).

Possibly worth pointing out that this also affects file path comments in the server build, e.g. you'll now see stuff like:
```js
// app/root.tsx
```
instead of:
```js
// /absolute/path/on/the/build/machine/to/app/root.tsx
```

Testing notes:
 1. Added integration test
 2. Verified manually, i.e.
    1. Create two remix projects (via npx create-remix@latest)
    2. `npm run build` them both
    3. `diff -r project1/build project2/build` has no differences
    4. `diff -r project1/public/build project2/public/build` has no differences
    5. `dev` and `start` still work as per usual


integration/deterministic-build-output-test.ts
```diff
@@ -0,0 +1,49 @@
...
+          <div>
+            {data}
+          </div>
+        )
+      }
+    `,
+    },
+  };
+  let dir1 = await createFixtureProject(appInit);
+  let dir2 = await createFixtureProject(appInit);
```
kentcdodds, line 31: It looks like we don't actually need any custom files for this fixture. Would you mind changing `createFixtureProject` to allow `files` to be optional and then the entire `init` argument to be optional so then we can do this:

```suggestion
  let dir1 = await createFixtureProject();
  let dir2 = await createFixtureProject();
```

kentcdodds (changes requested):
I'm surprised that test case fails without your changes (I'm assuming you ran the test first to ensure it does fail without your changes). This is beyond where I'm comfortable merging personally, but I did leave a bit of feedback if you'd like :) Thanks!

integration/deterministic-build-output-test.ts
jenseng: > I'm surprised that test case fails without your changes

Indeed, the underlying bug is similarly surprising :) ... I verified it failed on `dev` 2 months ago, and again after rebasing yesterday

jenseng: > we don't actually need any custom files

I don't know if that's true; for the bug to manifest we need some routes and virtual modules. I can take a closer look though, perhaps I can simplify the files used

jenseng: Upon further testing, w/o any files you always still have root.tsx, so the cascading indeterminate hash problem still persists 👍

jenseng: elaborating slightly on why the test fails on `dev`, the virtual modules end up having different contents and thus different hashes, and so that cascades down resulting in different filenames/etc across what should be identical builds:
```
    Error: expect(received).toEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 2

      Array [
        "build/index.js",
        "public/build/entry.client-JPJ6IQB3.js",
    -   "public/build/manifest-F783E782.js",
    -   "public/build/root-QVMFUD3S.js",
    +   "public/build/manifest-545DEB0B.js",
    +   "public/build/root-DZTWF5EQ.js",
        "public/build/_shared/chunk-RSZTKXZZ.js",
      ]
```

jenseng: Ok done 👍

kentcdodds (approved):
Thanks for doing that!

I'll leave the rest of the review to @mjackson

jenseng:
Seems like an unrelated CLI test failed, is there a way to retrigger? Or does it not matter?

jenseng:
@mjackson let me know if you have any questions/suggestions 😄

integration/deterministic-build-output-test.ts
```diff
@@ -0,0 +1,28 @@
+import { test, expect } from "@playwright/test";
+import globby from "globby";
+import fs from "fs";
+import path from "path";
+
+import { createFixtureProject, js } from "./helpers/create-fixture";
```
machour, line 6: 'js' is defined but never used
```suggestion
import { createFixtureProject } from "./helpers/create-fixture";
```

ryanflorence:
Let's fix the conflicts and then merge this bugger.

ryanflorence:
Jon has local changes but Yarn PnP doesn't play nicely with relative paths, so going to wait for a fix there before we do this.

jenseng:
Alright, I [opened a PR](https://github.com/yarnpkg/berry/pull/4569) to fix the esbuild-plugin-pnp bug, but if we're comfortable shipping [this hack](https://github.com/remix-run/remix/commit/f8f6bc540501807fb599757a210b63dd552c5161#diff-fd95498eaad39398e83a12f7e537312245f41759db9edaa2f335dab8511f0b41) we don't technically need to wait for it 😅

jenseng:
The esbuild-plugin-pnp bugfix has been merged, so once they release a new version I'll unwind my hack and bump the dependency 🥳

jenseng:
This PR should be good to go, but it looks like something is generally broken around the caching done by `bahmutov\npm-install` step in GH actions... if a build can't use the cache (i.e. yarn.lock has changed), the windows builds will fails. Here's a simple repro I set up that fails the same way: https://github.com/remix-run/remix/pull/3584

jenseng:
Alright, the PR now uses `@yarnpkg/esbuild-plugin-pnp/3.0.0-rc.10` which lets us remove the plugin hack/wrapper... in addition to the tests still all passing, `examples/yarn-pnp` also still builds correctly with this change. 

`@yarnpkg/esbuild-plugin-pnp/3.x` **does** require `node >=14.15.0` (the [first 14.x LTS version](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V14.md)), but that's not new for remix since `@npmcli/package-json` (another dependency) does too.

That said, this all may be moot as I just noticed [this PR](https://github.com/remix-run/remix/pull/3594) which swaps out this plugin entirely 😆😭 ... in any event, here's the diff from 2.0.2 (what people would have been installing prior to this PR). Although the version we're using is an RC, IMO it's safe to use since the only change it brings is my bugfix:

```diff
diff --git a/packages/esbuild-plugin-pnp/package.json b/packages/esbuild-plugin-pnp/package.json
index dbb346c74..8f22e8fb6 100644
--- a/packages/esbuild-plugin-pnp/package.json
+++ b/packages/esbuild-plugin-pnp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yarnpkg/esbuild-plugin-pnp",
-  "version": "2.0.2",
+  "version": "3.0.0-rc.10",
   "license": "BSD-2-Clause",
   "main": "./sources/index.ts",
   "dependencies": {
@@ -31,6 +31,7 @@
     "directory": "packages/esbuild-plugin-pnp"
   },
   "engines": {
-    "node": ">=12 <14 || 14.2 - 14.9 || >14.10.0"
-  }
+    "node": ">=14.15.0"
+  },
+  "stableVersion": "2.0.1"
 }
diff --git a/packages/esbuild-plugin-pnp/sources/index.ts b/packages/esbuild-plugin-pnp/sources/index.ts
index 379b64959..64fff1588 100644
--- a/packages/esbuild-plugin-pnp/sources/index.ts
+++ b/packages/esbuild-plugin-pnp/sources/index.ts
@@ -139,9 +139,11 @@ export function pnpPlugin({
           conditions = conditionsRequire;

         // The entry point resolution uses an empty string
-        const effectiveImporter = args.importer
-          ? args.importer
-          : `${baseDir}/`;
+        const effectiveImporter = args.resolveDir
+          ? `${args.resolveDir}/`
+          : args.importer
+            ? args.importer
+            : `${baseDir}/`;

         const pnpApi = findPnpApi(effectiveImporter) as PnpApi | null;
         if (!pnpApi)
```

MichaelDeBoey:
As mentioned in https://github.com/remix-run/remix/pull/3594#discussion_r909444900, I personally would be in favor of solving the problems from #3594 in `@yarnpkg/esbuild-plugin-pnp` (like we did for this one) & update the dependency instead of removing the plugin & having extra maintenance on our side tbh.

@jenseng Maybe you could point @lensbart into the right direction for a fix in `@yarnpkg/esbuild-plugin-pnp`?
I think you creating a fix yourself could also work though.

MichaelDeBoey:
@jenseng  Due to merging #3633, we have some conflicts here. Could you please resolve them?

PR #2034, opened 2022-02-19
fix: button value not submitted if form is inside a container that calls `stopPropagation` on the click event

airjp73:
I ran into a bug while fixing [this issue](https://github.com/airjp73/remix-validated-form/issues/64) in `remix-validated form`.

In the following situation:
* The `Form` is inside an element where `stopPropagation` is being called on a click event
* The submit button of the form has a `name` and `value`

The value on the button will not be submitted. Here's an example:

```tsx
<div onClick={e => e.stopPropagation()}>
  <Form method="post">
    <button type="submit" name="buttonValue" value="Hello">Submit</button>
  </Form>
</div>
```

This situation commonly arises when putting a form inside a headlessui Dialog.

# The fix

By catching the click event in the capture phase, we can catch it before anything has had a chance to stop propagation.

### Possible alternate fix

By using `event.nativeEvent.submitter` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter))([caniuse](https://caniuse.com/?search=submitter)), we can reliably identify the submit button without having to catch it with a click event. Not sure if the browser support is sufficient to use it though?

### DIFF
--- a/integration/button-value-test.ts
+++ b/integration/button-value-test.ts
@@ -1,0 +1,53 @@
+import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
+import type { Fixture, AppFixture } from "./helpers/create-fixture";
+
+let fixture: Fixture;
+let app: AppFixture;
+
+beforeAll(async () => {
+  fixture = await createFixture({
+    files: {
+      "app/routes/index.jsx": js`
+        import { Form } from 'remix';
+
+        export default function Index() {
+          return (
+            <div onClick={e => e.stopPropagation()}>
+              <Form method="post" action="/action">
+                <button type="submit" name="buttonValue" value="Hello">Submit</button>
+              </Form>
+            </div>
+          )
+        }
+      `,
+
+      "app/routes/action.jsx": js`
+        import { json, useActionData } from "remix";
+
+        export async function action({ request }) {
+          const formData = await request.formData();
+          return json({ buttonValue: formData.get('buttonValue') });
+        }
+
+        export default function Index() {
+          const data = useActionData();
+          return (
+            <p>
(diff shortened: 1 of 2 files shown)


zolrath:
I encountered this issue while attempting to put a button inside a [Headless UI Dialog](https://headlessui.dev/react/dialog). Looking forward to this getting merged!

edwin177:
For now, using `useSubmit` seems to get around the issue:

```
<div onClick={e => e.stopPropagation()}>
  <Form method="post">
    // works
    <button
      type="button"
      name="buttonValue"
      value="Hello"
      onClick={(e): void => submit(e.currentTarget)}
    >
      Submit
    </button>

    // does not work
    <button
      type="submit"
      name="buttonValue"
      value="Bye"
    >
      Submit
    </button>
  </Form>
</div>
```

emilbryggare:
The workaround that @edwin177 provided worked for me, but I had to add e.preventDefault() to avoid double form submits where one failed due to the missing name/value.
```
onClick={(e) => {
    e.preventDefault();
    submit(e.currentTarget);
}}
```

MichaelDeBoey:
Duplicate of #2309

airjp73:
@MichaelDeBoey, I notice that PR doesn't actually fix the bug, but this one does. Do you need a PR to fix?

airjp73:
Oh, I see. It's already fixed via using `submitter`.

mcansh:
> Oh, I see. It's already fixed via using `submitter`.

thanks for the PR! unfortunately didn't see it before we went and did it ourselves 😅. browser support seems fine and safari already includes the clicked button without needing event.submitter

PR #2035, opened 2022-02-19
docs(examples): add stripe payment integration example

marvinwu:
- [x ] Docs


examples/stripe-integration/app/routes/buy.tsx
```diff
@@ -0,0 +1,53 @@
+import { loadStripe } from "@stripe/stripe-js";
+import { Form, useActionData, useLoaderData } from "remix";
+import type { ActionFunction } from "remix";
+import { getStripeSession } from "~/utils/stripe.server";
+
+type returnData = {
```
kentcdodds, line 6: This is very unusual to have your loader and your action return the same data type. Could you have a `LoaderData` type and an `ActionData` type? In this example they don't technically even return the same data, so it doesn't make sense to use the same data type.

examples/stripe-integration/app/utils/stripe.server.tsx
```diff
@@ -0,0 +1,20 @@
...
+      price: priceId,
+      quantity: 1
+    }
+  ];
+  const session = await stripe.checkout.sessions.create({
+    mode: "payment",
+    payment_method_types: ["card"],
+    line_items: lineItems,
+    success_url: "http://localhost:3000/payment/success",
+    cancel_url: "http://localhost:3000/payment/cancelled"
```
kentcdodds, line 16: In the real world, you wouldn't use `http://localhost:3000`. Could you instead get the domain from the `request`? Similar to what I do here: https://github.com/kentcdodds/kentcdodds.com/blob/ebb36d82009685e14da3d4b5d0ce4d577ed09c63/app/utils/misc.tsx#L229-L237

kentcdodds (changes requested):
Thanks for putting this together! I've just got a couple of comments.

examples/stripe-integration/app/routes/buy.tsx
marvinwu: Thanks much Kent for reviewing, fixed

examples/stripe-integration/app/utils/stripe.server.tsx
marvinwu: that's good idea, just stole it 😁

examples/stripe-integration/app/routes/buy.tsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+};
+export const action: ActionFunction = async ({
+  request
+}): Promise<actionData> => {
+  const formData = await request.formData();
+  const id = await getStripeSession(
+    formData.get("price_id") as string,
+    getDomainUrl(request)
+  );
```
kentcdodds, line 31: I'm confused why this needs to be in the action function. Why can't we put this in the loader and then just have a button with an `onClick` that calls `stripe.redirectToCheckout`. Or even better, if Stripe has a way to get that URL so *we* can do the redirect we could still have a form, but we can do the redirect server side.

examples/stripe-integration/remix.config.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @type {import('@remix-run/dev').AppConfig}
+ */
+module.exports = {
+  appDirectory: "app",
+  browserBuildDirectory: "public/build",
+  publicPath: "/build/",
+  serverBuildDirectory: "build",
+  devServerPort: 8002
```
kentcdodds, line 9:
```suggestion
```

kentcdodds (changes requested):
Thanks! Could we iterate on the buy route some more please?

examples/stripe-integration/app/routes/buy.tsx
marvinwu: Sorry Kent missed your comment somehow - Yes you are very correct! We can indeed get the stripe redirect url after getting the session. I have refractor it so now it handles all stripe functions purely on the server side.

examples/stripe-integration/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+A demo of integrate stripe payment with remix, it handles stripe function purely on the server side( getting sessions, redirect etc) and no stripe client needed on the client.
+
+Relevant files:
+
+```
+└── app
+    ├── routes
+    ├── api
+    │   └── stripe-web-hook.tsx // api route to get stripe webhook postback
```
MichaelDeBoey, line 11:
```suggestion
    │   ├── stripe-web-hook.tsx // api route to get stripe webhook postback
```

examples/stripe-integration/app/routes/buy.tsx
```diff
@@ -0,0 +1,37 @@
...
+  ENV: {
+    PRICE_ID: string;
+  };
+};
+export const loader = async (): Promise<loaderData> => {
+  return {
+    ENV: {
+      PRICE_ID: process.env.PRICE_ID as string
+    }
+  };
```
MichaelDeBoey, line 15: Be sure to return using the `json` function if you're not returning a `Response`

```suggestion
  return json({
    ENV: {
      PRICE_ID: process.env.PRICE_ID as string
    }
  });
```

examples/stripe-integration/app/routes/buy.tsx
```diff
@@ -0,0 +1,37 @@
+import { Form, redirect, useLoaderData } from "remix";
```
MichaelDeBoey, line 1:
```suggestion
import { Form, json, redirect, useLoaderData } from "remix";
```

MichaelDeBoey (changes requested):

kentcdodds (approved):
Thanks for iterating on this!

Issue #1091, opened 2021-12-16
[Bug]: link prefetch only works for routes with a loader


kentcdodds:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [X] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.0.6

### What version of Node are you using? Minimum supported version is 14.

16

### Steps to Reproduce

1. Create a route that does not export a loader
2. Add `<Link to="that-route" prefetch="intent">Go to the route</Link>` to another route
3. Hover the link and notice that the JS/CSS is not prefetched

### Expected Behavior

I expect that JS modules and things in the `links` export to be prefetched.

### Actual Behavior

Nothing is prefetched.

Here's the part in the code that limits this functionality to routes with loaders only: https://github.com/remix-run/remix/blob/1da0e00c5a3e4293bab524a351d796d9ce32d6a7/packages/remix-react/links.ts#L324

I *think* we could just remove the `hasLoader` check and be fine, but I'm not sure what other impacts that would have.

ch4nd4n:
@kentcdodds how about updating the docs and leaving it as is? Do you think removing the loader will have bad UX?

brophdawg11:
Fixed in #2096

PR #2038, opened 2022-02-19
fix(components): fix prefetch for routes without a loader (#1091)

na2hiro:
Closes: #1091

Fixed a bug that prefetch doesn't happen if the route doesn't have a loader.

Current behavior:
* ✅ prefetch links to routes with loader: Remix adds `<link>`s for bundle file and loader data
* 🐛 prefetch links to routes without loader: **Remix doesn't add `<link>` tag**

Expected behavior:
* ✅ prefetch links to routes with loader: Remix adds `<link>`s for bundle file and loader data
* ✅ prefetch links to routes without loader: **Remix adds `<link>` for bundle file**

I made a fix which is suggested by @kentcdodds in the issue. Also I added and improved unit tests around elements when prefetch happens.

- [x] Tests
  - [x] I added test cases which reproduces the issue. The unit tests passed.
  - [x] However, entire testing with `yarn test` results in bunch of failures even for `dev` branch in my local, including flaky tests with timeouts, and it's difficult to see if my change affects other tests. ~I'd like to run branch testing and revisit if something is broken~
    - Update: I ran the branch testing in my forked repo, and I confirmed it passed all the test cases. https://github.com/na2hiro/remix/runs/5257542886?check_suite_focus=true

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -186,6 +186,7 @@
 - mskoroglu
 - msutkowski
 - mtt87
+- na2hiro
 - nareshbhatia
 - navid-kalaei
 - niconiahi
--- a/packages/remix-react/__tests__/components-test.tsx
+++ b/packages/remix-react/__tests__/components-test.tsx
@@ -80,13 +80,13 @@ function itPrefetchesPageLinks<
       jest.useFakeTimers();
     });
 
-    function withContext(stuff: JSX.Element) {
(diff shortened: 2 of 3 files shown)


brophdawg11:
Thank you for looking into this and opening a PR @na2hiro!  We actually had this on the radar internally as well and fixed in #2096, so I'm going to close this out

PR #2040, opened 2022-02-19
docs(examples): remove duplicte main html elements in blog tutorial

mkrtchian:
There is already a `<main>` element added in `app/root.tsx`, so all the other `<main>` elements are additional ones in the same HTML document. In this PR I replace them by `<section>`.

[According to the HTML specification](https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element):
> A document must not have more than one main element that does not have the hidden attribute specified.

This is especially useful for assistive technologies that need to know what is the main content.

### DIFF
--- a/docs/tutorials/blog.md
+++ b/docs/tutorials/blog.md
@@ -611,7 +611,7 @@ export default function Admin() {
           ))}
         </ul>
       </nav>
-      <section>...</section>
+      <main>...</main>
     </div>
   );
 }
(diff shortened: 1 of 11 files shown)


chaance:
I actually think it probably makes more sense to change the main in `root.tsx` to a div and leave the others alone. Probably a bit nitpicky here but each route will generally know its semantic boundaries a bit better than the root!

mkrtchian:
@chaance ok, I removed the `main` element from `root.tsx`, and added it in the pages where it was missing.

PR #2044, opened 2022-02-19
docs(example): Setup eslint-prettier with Remix

Girish21:


louislefebvre:
Hey, thanks for this example, that is something that I would have been delighted to find before starting my remix project!
I was wondering if adding `api/` to the file `.eslintignore` would be relevant? On my side, if I do not have it, eslint takes ages to lint through and the output is unreadable.

chaance:
I'm going to close this for now, as we just introduced a new pre-configured ESLint config and are working on a new feature that will change how we approach third-party tools in general. Stay tuned for details!

PR #2046, opened 2022-02-19
feat(remix-dev): only purge require cache if updated

### LINKED ISSUE #1818: [Feature]: In dev mode, only purge the require cache on file changes

When running `remix dev`, instead of purging the `require` cache on every single request, only do so when any of the source files were changed.

Right now, the `require` cache is cleared [on every single request](https://github.com/remix-run/remix/blob/dev/packages/remix-dev/cli/commands.ts#L169-L174) when running `remix dev`.

This is very un-intuitive for me, and also different from the [documented behavior](https://remix.run/docs/en/v1/other-api/serve).

`createMemorySessionStorage` is intended to be used only during development, but doesn't actually work with `remix dev` because of this (or at least not without additional code to e.g. store it in the `global` object, which isn't documented either).

I can see that this change was made in #431, but without any comments as to _why_.

If purging the cache on every request is the actually intended behavior, it would be great if this could be documented somewhere, including some instructions on how to get `createMemorySessionStorage` working in dev mode.

warmpigman:
Fixes #1818  Check version of assets.json, if it has changed, then purge require cache.


MichaelDeBoey:
@warmpigman Could you please rebase your branch onto latest `dev` & resolve conflicts?

warmpigman:
> @warmpigman Could you please rebase your branch onto latest `dev` & resolve conflicts?

@MichaelDeBoey I think I did so, could you check?

packages/remix-dev/cli/commands.ts
```diff
@@ -302,9 +302,11 @@ export async function dev(remixRoot: string, modeArg?: string) {
 }
 
 function purgeAppRequireCache(buildPath: string) {
-  for (let key in require.cache) {
-    if (key.startsWith(buildPath)) {
-      delete require.cache[key];
+  if (require.cache[buildPath + "/assets.json"]?.exports.version !== JSON.parse(require('fs').readFileSync(BUILD_DIR + '/assets.json')).version) {
```
MichaelDeBoey, line 305: Could we make this an early return please?

This will create a positive condition (which has a lower cognitive load) + will make sure the code is less nested.

MichaelDeBoey (changes requested):

dmarkow:
I'm excited to see this get fixed, but does this approach still work? As far as I can tell, Remix stopped creating that assets.json file as of v1.2.0.

balzdur:
What about [this approach](https://github.com/remix-run/remix/pull/4294) ?

pcattori:
Superceded by #5133

Issue #258, opened 2021-08-25
Support for .env files


sergiodxa:
Right now to use a `.env` file I need to install the `dotenv` package and import it and run it inside the `entry.server` file, which is a simple thing to do but it would be nicer to have it already built-in by Remix (only server-side of course).

So if a project has a `.env` file it can use it automatically.

This should probably be added to the adapters, if using Vercel it already loads the .env (because `vercel dev` does it), for Express it could be added here https://github.com/remix-run/remix/blob/master/packages/remix-express/server.ts or in the Remix Serve package instead https://github.com/remix-run/remix/blob/master/packages/remix-serve/index.ts (which may be a better place).

---

Some extra step that may or may not be useful here could be to support `.env`, `.env.development`, `.env.production` and `.env.local` with this behavior:

- First load from `.env`, these are env shared everywhere
- Then load from `.env.development` or `.env.production` based on the NODE_ENV value (maybe also `.env.test`), these are env specific (as the `.env` should be)
- Then load from `.env.local` and these let you overwrite the previous one, useful to trying things

Each one should overwrite the previous one.

brophdawg11:
Added in #2063 👍

PR #2063, opened 2022-02-21
feat: load environment variables via dotenv for remix dev

brophdawg11:
Loads environment variables via `dotenv` when running via `remix dev`.  Attempts to load various files via the following order of precedence:

* `.env.local`, which overrides
* `.env.development`, which overrides
* `.env`

Maintains default behavior of `dotenv` and will not overwrite any existing ENV values.

Closes: #258

- [x] Docs
- [ ] Tests
  - Unsure if existing test harnesses can test this behavior - will check


packages/remix-dev/cli/commands.ts
```diff
@@ -158,6 +180,12 @@ export async function dev(remixRoot: string, modeArg?: string) {
 
   let config = await readConfig(remixRoot);
   let mode = isBuildMode(modeArg) ? modeArg : BuildMode.Development;
+
+  // Load .env files if present.  Precedence: .env.local -> .env.{mode} -> .env
+  await loadEnv(config.rootDirectory, "local");
+  await loadEnv(config.rootDirectory, mode);
```
brophdawg11, line 186: `npm run dev` sets `NODE_ENV=development` but it does seem possible to use `NODE_ENV=production npx remix dev` so this currently handles both `.env.development` and `.env.production`.  If the intention is not to use `remix dev` with `production` then we can simplify this to only handle `.env.development`.

packages/remix-dev/cli/commands.ts
```diff
@@ -142,8 +143,29 @@ export async function watch(
...
+
+  console.log(`Loading environment variables from: ${envFile}`);
+  const result = require("dotenv").config({ path: envPath });
+  if (result.error) {
+    throw result.error;
+  }
+}
+
 export async function dev(remixRoot: string, modeArg?: string) {
-  // TODO: Warn about the need to install @remix-run/serve if it isn't there?
```
brophdawg11, line 146: Stale TODO?  This seems to be handled in the lines below

docs/api/conventions.md
```diff
@@ -1310,3 +1310,15 @@ export default function Page() {
 [link tag]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link
 [minimatch]: https://www.npmjs.com/package/minimatch
 [handledatarequest]: #entryservertsx
+
+## Environment Variables
+
+Hopefully, your application has at least 3 deployment locations (local, test, and prod). In order to decouple your application from it's eventual deployment location, environment-specific application configuration values are best stored in Environment Variables. How you configure these will vary based on your deployment target in higher environments, so Remix does not do anything special during production runs in your deployment target or via `remix serve`.
```
brophdawg11, line 1316: Any use in linking out to something here about env var "best practices"?   Potentially https://12factor.net/config or something similar.

packages/remix-dev/cli/commands.ts
brophdawg11: The more I think about this, I'm not convinced all 3 are relevant for `remix dev` alone.  The original github issue was suggesting all 3 I think with the intention that they'd be used by adapters as well.  Since this is only for `remix dev`, I feel like only supporting `.env` will be sufficient?  I think if folks are doing advanced enough stuff to warrant `.production` and/or overrides in `.local`, it's not too much to leave that up to the user?

I can maybe see a use case for `.env.local` as an entry point for users to symlink to any number of specific local override files they care to maintain while not having to dup values from `.env` in them...

docs/api/conventions.md
brophdawg11: 🤦‍♂️ Just found the `envvars.md` guide - moving this over there

packages/remix-dev/cli/commands.ts
kentcdodds: I say let's start with only supporting `.env` and then we can add support for the different modes in the future if there's a need there.

kentcdodds (commented):
Looks pretty good to me. Just one thought.

mjackson (approved):
Complete with docs updates! 👏 

Thanks, @brophdawg11!

PR #2033, opened 2022-02-19
build: fix run-p glob matching for yarn > 2 ..

CanRau:
Just ran into [`No matches found: "dev:*"`](https://github.com/mysticatea/npm-run-all/issues/200) using Yarn 3

Quoting it `"dev": "run-p 'dev:*'",` like mentioned in the linked issue fixed it for me with Yarn 3 and also tested Yarn 1.22.10 and npm 8.1.0, just to make sure.

So I figured as it shouldn't break anything why not always quote it?

Please correct me if I'm mistaken 🙏

PR #2064, opened 2022-02-21
[fix] bash expansion typo in package.json

benallfree:
I am running zsh and found that I need to enclose `dev:*` in quotes. This may affect all templates.

Sorry this is my first PR, the PR template didn't really cover this case. Apologies if I'm doing it wrong :)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -25,6 +25,7 @@
 - ashocean
 - BasixKOR
 - BenMcH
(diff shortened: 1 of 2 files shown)


machour:
@benallfree Thank you so much for taking the time to open this PR (you did it right), however there was already one covering this issue: #2033.

Closing this PR as duplicate.

PR #2066, opened 2022-02-21
feat: Refactor `@remix-run/eslint-config` with added support for Testing Library

chaance:
This PR does a few things needed before we integrate our ESLint config with the starter templates:

- Updates dependencies for all test-related tooling
- Integrates some useful rules for Testing Library. This will make migration easier for users of Next or Create React App which both ship Testing Library ESLint rules by default.
  - Because this change will also change the lint results for users already using our config, this would be a breaking change if introduced directly in our `jest` module. This PR also deprecates that module and introduces a new one named `@remix-run/eslint-config/jest-testing-library` which is also more explicit about what the package includes.
- Breaks up rulesets into individual modules. This makes it simple to extend a specific ruleset with a different set of conditions. For example, a user who uses Cypress might want to use the same Testing Library rules for their Cypress and Jest tests, but they might not need other Jest rules for those tests. Those users can import the `testing-library` rules as a module and extend them directly via their config:
  ```js
  // eslint.config.js
  const testingLibRules = require("@remix-run/eslint-config/rules/testing-library");
  module.exports = {
    extends: ["@remix-run/eslint-config", "@remix-run/eslint-config/jest-testing-library"],
    overrides: [
      {
        files: ["cypress/**/*.{js,jsx,ts,tsx}"],
        rules: {
          ...testingLibRules,
          // override these because they don't make sense in Cypress:
          "testing-library/prefer-screen-queries": "off",
          "testing-library/await-async-query": "off",
        },
      },
    ],
  };
  ```
- Updates our internal ESLint config to use the new rules and fixes some internal files and examples that are flagged as a result
- Reduces the severity of more opinionated rules in the `jest` ruleset to a warning instead of an error
- Adds a few more useful rules for Jest:
  - `"jest/no-deprecated-functions"`
  - `"jest/no-disabled-tests"`
  - `"jest/no-export"`
  - `"jest/no-focused-tests"`


docs/pages/faq.md
```diff
@@ -31,7 +31,7 @@ export async function requireUserSession(request) {
   // put in the session when the user authenticated
   if (!session.has("userId")) {
     // if there is no user session, redirect to login
-    throw new redirect("/login");
+    throw redirect("/login");
```
chaance, line 34: Flagged by `yarn run lint`

docs/tutorials/jokes.md
```diff
@@ -5435,7 +5435,7 @@ function escapeCdata(s: string) {
 
 function escapeHtml(s: string) {
   return s
-    .replace(/\&/g, "&amp;")
+    .replace(/&/g, "&amp;")
```
chaance, line 5438: Flagged by `yarn run lint`

examples/image-resize/app/components/image.tsx
```diff
@@ -1,26 +1,34 @@
+import { forwardRef } from "react";
```
chaance, line 1: A few things in this example were flagged by `yarn lint` but I made some additional useful improvements that I think will be helpful for most users who will copy/paste this example.

packages/remix-server-runtime/__tests__/data-test.ts
```diff
@@ -165,13 +165,21 @@ describe("loaders", () => {
       }
     } as unknown as RouteMatch<ServerRoute>;
 
+    let possibleError: any;
```
chaance, line 168: Flagged by `yarn run lint`

mjackson (approved):

kentcdodds:
I'mma fan 👍

PR #2071, opened 2022-02-22
test: deno fly.io deployment test

jacob-ebey:
Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/create-remix/templates/deno/Dockerfile
+++ b/packages/create-remix/templates/deno/Dockerfile
@@ -1,0 +1,39 @@
+# install all node_modules, including dev
+FROM node:16-alpine as deps
+
+RUN mkdir /app/
+WORKDIR /app/
+
+ADD package.json package-lock.json ./
+RUN npm install --production=false
+
+# build app
+FROM node:16-alpine as build
+
+ENV NODE_ENV=production
+
+RUN mkdir /app/
+WORKDIR /app/
+
+COPY --from=deps /app/node_modules /app/node_modules
+
+# app code changes all the time
+ADD . .
+RUN npm run build
+
+# build smaller image for running
+FROM denoland/deno:1.16.4
+
+ENV PORT=8080
+
+RUN mkdir /app/
+WORKDIR /app/
+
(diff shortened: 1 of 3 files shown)


packages/create-remix/templates/deno/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /app/
+
+COPY --from=build /app/build /app/build
+COPY --from=build /app/public /app/public
+
+RUN deno cache ./build/index.js
+
+EXPOSE 8080
+
+CMD ["run", "--allow-net", "--allow-read", "--allow-env", "./build/index.js"]
```
mcansh, line 39: should this be `deno` prefixed?

```suggestion
CMD ["deno", "run", "--allow-net", "--allow-read", "--allow-env", "./build/index.js"]
```

scripts/deployment-test/fly-deno.mjs
```diff
@@ -0,0 +1,104 @@
+import path from "path";
+import { spawnSync } from "child_process";
+import fse from "fs-extra";
+import toml from "@iarna/toml";
+
+import { sha, getSpawnOpts, runCypress, addCypress } from "./_shared.mjs";
+import { createApp } from "../../build/node_modules/create-remix/index.js";
+
+let APP_NAME = `remix-fly-${sha}`;
```
mcansh, line 9:
```suggestion
let APP_NAME = `remix-fly-deno-${sha}`;
```

mcansh (changes requested):
looks great besides two teeny tiny things

packages/create-remix/templates/deno/Dockerfile
jacob-ebey: The default command is "deno" for this image so it's not needed :)

mcansh: ah beautiful

jacob-ebey:
We have a deno deployment test to deno.deploy

PR #2072, opened 2022-02-22
fix: PNPM does some weird hardlinking so attempting to write to index.d.ts fails

jacob-ebey:
fix: PNPM does some weird hardlinking so attempting to write to index.d.ts of the remix package causes completely unrelated packages to break?

Anyways, this stops shipping type definitions with the remix package as we generate them based on the setup command.

Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/remix/tsconfig.json
+++ b/packages/remix/tsconfig.json
@@ -9,8 +9,7 @@
     "allowSyntheticDefaultImports": true,
     "strict": true,
 
-    "declaration": true,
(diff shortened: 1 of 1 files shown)


MichaelDeBoey:
Do we still need this now that we're recommending to not use the global `remix` package anymore? 🤔

MichaelDeBoey:
@jacob-ebey Do we still need this now that we're recommending to not use the global `remix` package anymore? 🤔

Issue #2075, opened 2022-02-22
Build fails when using supabase-js


aiji42:
### What version of Remix are you using?

1.2.2

### Steps to Reproduce

1. Select Cloudflare workers (or pages) as runtime.
2. Install `@supabase/supabase-js`. 
3. Call createClient in loader or .server.ts
4. Start up a dev server (`yarn dev`)

### Expected Behavior

The server comes up without error.

### Actual Behavior

Error occurs. `TypeError: globalThis.XMLHttpRequest is not a constructor`
![スクリーンショット 2022-02-22 午後8 21 26](https://user-images.githubusercontent.com/6711766/155122543-5754d1fe-5797-454b-8a1d-39e75a9710a6.png)

aiji42:
You can work around it by adapting such a patch.
If the runtime is Cloudflare Workers, the browser module should be the preferred solution.

```
diff --git a/node_modules/@remix-run/dev/compiler.js b/node_modules/@remix-run/dev/compiler.js
index 0705b93..72ed487 100644
--- a/node_modules/user/node_modules/@remix-run/dev/compiler.js
+++ b/node_modules/user/node_modules/@remix-run/dev/compiler.js
@@ -364,7 +364,7 @@ async function createServerBuild(config, options, assetsManifestPromiseRef) {
     format: config.serverModuleFormat,
     treeShaking: true,
     minify: options.mode === build$1.BuildMode.Production && !!config.serverBuildTarget && ["cloudflare-workers", "cloudflare-pages"].includes(config.serverBuildTarget),
-    mainFields: config.serverModuleFormat === "esm" ? ["module", "main"] : ["main", "module"],
+    mainFields: ["browser", "module", "main"],
     target: options.target,
     inject: [reactShim],
     loader: loaders.loaders,
```

and3rsonsousa:
It also happens with graphql-request.

Pitasi:
Yeah it's happening to me too when using graphql-request. The small patch proposed by @aiji42 didn't help: it moved the problem from build-time to run-time (I get an application error in the browser: `ReferenceError: XMLHttpRequest is not defined`).

Pitasi:
Yes, I'm using v1.2.2, and yes I updated createServerBuild. I think that graphql-request tries to determine if fetch is available or not in some way but it doesn't work.

Anyway see my prev comment that I edited, I got it to work.

Thanks for helping!

simjes:
Same error goes for `@notionhq/client`

[...]

PR #2076, opened 2022-02-22
fix(remix-dev): fix compiler for Cloudflare runtime

aiji42:
Closes: #2075

If the runtime is Cloudflare Worlers (or pages), then the browser module must be resolved first.
Currently I have found the problem in `@supabase/supabase-js`, but similar problems must be occurring in other packages. An update to this pull request should resolve them.

I used this esbuild configuration as a reference.
https://github.com/gmencz/cloudflare-workers-typescript-esbuild-esm/blob/main/build.mjs

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -440,7 +440,12 @@ function createServerBuild(
           config.serverBuildTarget
         ),
       mainFields:
-        config.serverModuleFormat === "esm"
+        !!config.serverBuildTarget &&
+        ["cloudflare-workers", "cloudflare-pages"].includes(
+          config.serverBuildTarget
+        )
+          ? ["browser", "module", "main"]
+          : config.serverModuleFormat === "esm"
           ? ["module", "main"]
           : ["main", "module"],
       target: options.target,
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -423,6 +423,11 @@ function createServerBuild(
     plugins.unshift(NodeModulesPolyfillPlugin());
   }
 
+  const isCloudflareRuntime = [
+    "cloudflare-workers",
+    "cloudflare-pages"
+  ].includes(config.serverBuildTarget ?? "");
+
   return esbuild
     .build({
       absWorkingDir: config.rootDirectory,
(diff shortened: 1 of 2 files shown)


aiji42:
When will this pull request be merged?
Development is complete, but is there anything else I need to do?

Repositories containing fetch-dependent libraries (such as supabase-js and graphql-request) will not be able to run on cloududflare workers until this fix is released.

jacob-ebey:
@mjackson @ryanflorence I'm in favor of this. CF is closer to a browser runtime than node CJS or ESM. Even react 18 will need to use the "browser" version to support streaming mode via WebStreams. The CF workers + webpack setups I've seen do the same thing. I believe this would resolve issues with Prisma and Fauna packages as well.

packages/remix-dev/compiler.ts
```diff
@@ -427,7 +427,12 @@ async function createServerBuild(
...
         ),
       mainFields:
-        config.serverModuleFormat === "esm"
+        !!config.serverBuildTarget &&
+        ["cloudflare-workers", "cloudflare-pages"].includes(
+          config.serverBuildTarget
+        )
+          ? ["browser", "module", "main"]
+          : config.serverModuleFormat === "esm"
           ? ["module", "main"]
```
cliffordfajardo, line 436: Maybe the conditional could be encapsulated into a descriptive variable name to give history & context for the to the logic

For example:
```ts
const isCloudFlareWorkerRuntime =  ["cloudflare-workers", "cloudflarepages"].includes(config?.serverBuildTarget)


mainFields: isCloudFlareWorkerRuntime  ? ...
```

cliffordfajardo (commented):
.

packages/remix-dev/compiler.ts
aiji42: I think you're right and I'll fix it.

aiji42: Fixed

aiji42:
The code pointed out in the review has already been fixed. Please review it again or give me instructions.
I'm eager for this MR to be released as soon as possible since there are many libraries affected.

MichaelDeBoey:
Closing this in favor of #1977 as that seems to fix the same issue.
Please open a new PR if it doesn't.

PR #2077, opened 2022-02-22
fix: Fix Express caching

jensmeindertsma:
This pull request fixes a regression to the Express caching logic introduced in #1994. The use of `"public/build"` where `"public"` should have been used meant that files inside `public` but outside of `public/build` weren't being served.


machour:
Fixed in https://github.com/remix-run/remix/commit/b1ce044f812a9bd47a1cffbc1e598bfb2a2e5868

PR #2080, opened 2022-02-22
ci: move deployment test dependencies out of main remix deps

mcansh:
This PR does the following:
- installs create-remix from npm instead of using the built version from `build`
	- doing so, allows us to not have to rebuild remix during each deployment job
- moves dependencies that are only used by the deployment tests to `scripts/deployment-test/package.json`
	- allows us to not have to install unused deps when working on remix itself
- explicitly adds the various CLIs used to dependencies
- [x] updates 'checkUp' function

with all of that said, we cut out a nice chunk of time for each deployment Group 2

Closes: #

- [ ] Docs
- [x] Tests


.github/workflows/deployment-test.yml
```diff
@@ -14,32 +17,28 @@ jobs:
       - run: echo "::set-output name=version::$(cat .nvmrc)"
         id: nvmrc
 
-      - name: Use Node.js
-        uses: actions/setup-node@v2
+      - name: Setup node
+        uses: actions/setup-node@v1
```
MichaelDeBoey, line 21:
```suggestion
        uses: actions/setup-node@v2
```

MichaelDeBoey (changes requested):

PR #2103, opened 2022-02-23
chore: remove `dotenv` references

MichaelDeBoey:
As a follow-up of @brophdawg11's #2063

### DIFF
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -2982,7 +2982,7 @@ export const action: ActionFunction = async ({
 
 </details>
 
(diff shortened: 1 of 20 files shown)


brophdawg11:
Thanks @MichaelDeBoey!   I'll take a look at this shortly 👍

brophdawg11:
@MichaelDeBoey I think we can also remove `dotenv` references from `package-lock.json` in those examples where we removed from `package.json`?

MichaelDeBoey:
@brophdawg11 We could indeed, but these are already out of date (see `remix` version).

That's why I proposed to just delete all of these `package-lock.json` & `yarn.lock` files in #1383

brophdawg11:
@MichaelDeBoey I agree with @kentcdodds in #1383 that it's safer to keep the lock files in the long run for `examples/`.  Removing the lock file opens up the possibility for a breaking change to a package to slip in next time someone installs and break the example.  The lock files should ideally eliminate that scenario, or at the very least reduce the frequency.

As for the `remix` version - I asked around and it turns out there's a script that auto-updates the `remix` version in examples `package.json` files automatically on new releases, but looks like that script is not updating `package-lock.json` - so we're opening an issue internally to get that fixed.

For this PR - do you mind just doing a selective removal of `dotenv` in `package-lock.json`?  At a quick glance I think that remix was the only thing depending on it - and it wasn't a nested dependency of any other packages for the examples I checked.

Thanks!

MichaelDeBoey:
@brophdawg11 I've rebase the branch & update the lock files of the impacted examples

PR #2107, opened 2022-02-23
docs(examples): add infinite scrolling example

adicuco:
Minimal example of infinite scrolling with `useFetcher` and `URL Search Params`

I tried to keep it very basic with the styling, but I can make it look nicer if it brings more value.


examples/infinite-scrolling/app/routes/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const handleLoadMore = () => {
+    if (fetcher.state === "idle" && canLoadMore) {
+      setCanLoadMore(false);
+      fetcher.load(`/?page=${page + 1}`);
+    }
+  };
+
+  useEffect(() => {
+    if (fetcher.data && fetcher.data.results.length > 0) {
+      setJokes((prev) => [...prev, ...fetcher.data.results]);
```
kentcdodds, line 45: The problem with this approach is that you may run out of memory if the user scrolls too far. I have an implementation of infinite scroll that I like, I just never got around to finishing it up. Do you mind taking a look at that and helping it across the finish line? https://github.com/remix-run/remix/tree/kent/infinite-scroll/examples/infinite-scroll

kentcdodds (changes requested):
Thanks for working on this!

examples/infinite-scrolling/app/routes/index.tsx
adicuco: Thank you for the feedback and suggestion 😁  Next week, I'll come up with a new solution based on your approach.

kentcdodds: Thank you! And thanks for the follow-up 👍

adicuco: I refactored the demo to use `react-virtual` and the fake items from your examples which I also added along with 3 more that use `page` instead of the offset. The "alternative" demo is based on the example of infinite scrolling from `react-virtual`'s docs. 

I am not entirely satisfied with my final solution (especially `page/advanced`) and I'm sure it can be improved, but I am a bit stuck in terms of what I should focus on optimising. Any feedback is greatly appreciated and I will gladly try to implement it if I can.

kentcdodds (approved):
Pulled this down and I think it's great. Well done. Thank you for taking the time to do this! There's probably a library hidden in this code somewhere waiting for someone to extract and release it :)

Cheers!

PR #2109, opened 2022-02-24
feat: add generics to `Cookie` interface

htunnicliff:
This PR adds an optional generic to the `Cookie` interface and related functions. By having an optional generic available when a cookie is created, stronger types can be expected.

Here is an example:

```ts
import { createCookie } from "remix";

type Data = {
  userId: string;
  favoriteColor: string;
};

export const sessionCookie = createCookie<Data>("session");

// Enforce types when serializing
await sessionCookie.serialize({
  userId: "user-123",
  favoriteColor: false // <-- Error: Type 'boolean' is not assignable to type 'string'
});

// Receive types when parsing
const data = await sessionCookie.parse("... Cookie headers");

data.userId // <-- string
data.favoriteColor // <-- string

### DIFF
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -34,7 +34,7 @@ export type CookieOptions = CookieParseOptions &
  *
  * @see https://remix.run/api/remix#cookie-api
  */
-export interface Cookie {
+export interface Cookie<Value = any> {
   /**
    * The name of the cookie, used in the `Cookie` and `Set-Cookie` headers.
    */
@@ -60,13 +60,13 @@ export interface Cookie {
   parse(
     cookieHeader: string | null,
     options?: CookieParseOptions
-  ): Promise<any>;
+  ): Promise<Value | null | "">;
 
   /**
    * Serializes the given value to a string and returns the `Set-Cookie`
    * header.
    */
-  serialize(value: any, options?: CookieSerializeOptions): Promise<string>;
+  serialize(value: Value, options?: CookieSerializeOptions): Promise<string>;
 }
 
 /**
@@ -74,10 +74,10 @@ export interface Cookie {
  *
  * @see https://remix.run/api/remix#createcookie
  */
-export function createCookie(
+export function createCookie<Value extends any = any>(
   name: string,
   cookieOptions: CookieOptions = {}
-): Cookie {
+): Cookie<Value> {
   let { secrets, ...options } = {
     secrets: [],
     path: "/",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/remix-server-runtime/cookies.ts [hunk 1]
```diff
@@ -60,24 +60,24 @@ export interface Cookie {
   parse(
     cookieHeader: string | null,
     options?: CookieParseOptions
-  ): Promise<any>;
+  ): Promise<Value | null | "">;
```
htunnicliff, line 63: Adding a literal type for `""` is necessary because the way `parse()` is written means that a cookie may be set but have an empty string value. 

If this is accurate, these types should be correct. 

If this is _not_ the case, it might be a good idea to remove the `return "";` from the `parse` function and let it only return `Value | null`.

packages/remix-server-runtime/cookies.ts
```diff
@@ -99,12 +99,17 @@ export function createCookie(
...
-          : await decodeCookieValue(cookies[name], secrets)
-        : null;
+      const cookies = parse(cookieHeader, { ...options, ...parseOptions });
+
+      if (name in cookies) {
+        if (cookies[name] === "") {
+          return "";
+        } else {
+          return decodeCookieValue<Value>(cookies[name], secrets);
+        }
```
cliffordfajardo, line 109: is the omission of `await` in front of `decodeCookieValue` intentional?

cliffordfajardo (commented):
.

packages/remix-server-runtime/cookies.ts
cliffordfajardo: What are your thoughts on usage of `unknown` vs `any`?
- https://dmitripavlutin.com/typescript-unknown-vs-any/#:~:text=Conclusion,to%20perform%20operations%20on%20unknown%20

cliffordfajardo (commented):
.

packages/remix-server-runtime/cookies.ts
htunnicliff: I removed `await` since it wasn't necessary; in either case, the function will return a promise.

htunnicliff:
@cliffordfajardo is there anything else I can do to help merge this in?

cliffordfajardo:
@htunnicliff - someone from the Remix team needs to approve 💿 I’m not a core maintainer of this repo

htunnicliff:
> @cliffordfajardo is there anything else I can do to help merge this in?
>> Hi @htunnicliff - someone from the Remix team needs to approve - like the top 4 contributors here
>> 
>>     * https://github.com/remix-run/remix/graphs/contributors

Hey @ryanflorence 👋🏻 – is there anything else you would like to see in this PR before it can be merged in?

htunnicliff:
Since I didn't hear anything, I'll go ahead and close this PR. 

I'd be happy to assist with adding in TypeScript Generics in the future, if that would be of interest.

cliffordfajardo:
Hi @chaance this PR is small & seems useful 💿

PR #2110, opened 2022-02-24
bug: <Form> has incorrect referrer header

itsMapleLeaf:
- [ ] Docs
- [x] Tests


itsMapleLeaf:
oops

integration/bug-report-test.ts
```diff
@@ -36,27 +36,36 @@ beforeAll(async () => {
...
-          return <div>cheeseburger</div>;
+      "app/routes/action.jsx": js`
+        import { redirect } from 'remix'
+        export async function action({ request }) {
+          const data = await request.formData()
+          const redirects = data.get("redirects")
+          const referrer = new URL(request.headers.get("referer"))
+          referrer.searchParams.set("redirects", Number(redirects) + 1)
+          return redirect(referrer.href)
         }
```
machour, line 69:
```suggestion
        }
        
        export default function Render() {
            return <span />;
        }
```

integration/bug-report-test.ts
```diff
@@ -85,16 +97,19 @@ test.afterAll(async () => appFixture.close());
 // add a good description for what you expect Remix to do 👇🏽
 ////////////////////////////////////////////////////////////////////////////////
 
-test("[description of what you expect it to do]", async ({ page }) => {
+it("should redirect via referrer header", async ({page}) => {
```
machour, line 100:
```suggestion
test("should redirect via referrer header", async ({page}) => {
```

mcansh:
interesting that `fetcher.Form` doesn't have this issue, but `Form` does as they both call the same underlying code at the end of the day...

ryanflorence:
Just FYI: This happens because React Router does a `history.pushState` and then Remix kicks of data fetching in response to that. By the time the POST fetch is made, the browser URL has already changed, so the `fetch` sends the new URL (because that's the current URL now).

In React Router 6.4 we don't do fetching in response to `history.pushState` anymore, but we call it after data is done loading. The referrer will be correct when we get that work back over to Remix (which we're working on right now).

ryanflorence:
Except when the back button is clicked actually, and I'm not sure we can do anything about it 🤔

It seems odd for loaders to care about the referrer, it's usually just actions, so I think we'll be alright and just wait a decade for the [navigation API](https://github.com/WICG/navigation-api) to show up?

kevinvugts:
> This happens because React Router does a `history.pushState` and then Remix kicks of data fetching in response to that. By the time the POST fetch is made, the browser URL has already changed, so the `fetch` sends the new URL (because that's the current URL now).
> 
> In React Router 6.4 we don't do fetching in response to `history.pushState` anymore. Instead we do the data fetching first, and then call `history.pushState` when it's done. The referrer will naturally be correct when we get that work back over to Remix.

This still seems to be the case even after upgrading to remix 1.9.0 which has the latest version of react-router-dom. 

After clicking a NavLink on page with url /home and a destination path of /configurator the http referrer in object request inside the loader() function of Remix has the http referrer set to /configurator instead of the expected /home http referrer. 

Any idea what goes wrong?

PR #2111, opened 2022-02-24
ci: use cloudflare api to determine if deployment is done

mcansh:
This PR changes how we wait for deployments for Cloudflare Pages but pinging the api instead of a setTimeout, it also adds support for "fast builds" once they enable it to be configured via the API - if they don't, it'll be enabled by default on April 1st.

Closes: #

- [ ] Docs
- [x] Tests

### DIFF
--- a/scripts/deployment-test/cf-pages.mjs
+++ b/scripts/deployment-test/cf-pages.mjs
@@ -57,6 +57,7 @@ async function createCloudflareProject() {
           build_command: "npm run build",
           destination_dir: "public",
           root_dir: "",
+          fast_builds: true,
         },
       }),
     }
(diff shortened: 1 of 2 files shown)


scripts/deployment-test/cf-pages.mjs
```diff
@@ -86,8 +88,46 @@ async function createCloudflareDeployment() {
...
+      let response = await fetch(url, {
+        headers: {
+          "X-Auth-Email": process.env.CF_EMAIL,
+          "X-Auth-Key": process.env.CF_GLOBAL_API_KEY,
+        },
+      });
+
+      if (response.status >= 200 && response.status < 400) {
+        let data = await response.json();
+        let deployment = data.result[0];
```
mcansh, line 115: each run creates a new GitHub repo and Pages project, so this is _fine_ to do, we could also return the deployment id from `createCloudflareDeployment` and pass it to this function

mjackson (approved):

PR #2112, opened 2022-02-24
docs: fix 404 on conventions page

marr:
Fix incorrect link on conventions page. Also rename link title _Route Module Constraints_ => _Module Constraints_ to match the destination of the link.


kentcdodds:
This has been fixed already. Sorry about that!

PR #2116, opened 2022-02-24
docs(examples): remove lock files

MichaelDeBoey:
Re-submission of #1383

When working on #2103, I [noticed](https://github.com/remix-run/remix/pull/2103#issuecomment-1050255699) that these lock files are already out of date with their respective `package.json`, so I think it's better to just delete all of these

https://github.com/remix-run/remix/blob/b8e194e7e3e3b200aef6fa8980782746e5d2033a/examples/bullmq-task-queue/package.json#L22
https://github.com/remix-run/remix/blob/b8e194e7e3e3b200aef6fa8980782746e5d2033a/examples/bullmq-task-queue/package-lock.json#L101-L102

CC/ @brophdawg11 @kentcdodds


brophdawg11:
The mismatched `remix` versions turns out to be a miss in the script that auto-updates the remix version for `examples/` on new releases.  That script should also be updating the `package-lock` files, so we're going to open a ticket to fix that.  See https://github.com/remix-run/remix/pull/2103#issuecomment-1050311257 for the reasoning behind keeping the lock files though 👍

PR #2121, opened 2022-02-25
Fix outlet examples in docs/api/remix.md

uhoh-itsmaciek:
Hello! I think there's an inconsistent variable name in the docs here: the type is declare to use `invoiceSort`, but `sort` is used in a couple of places instead.


kentcdodds (approved):
Thanks!

PR #2122, opened 2022-02-25
feat(remix-server-runtime): add loader context generic types

rwu823:
Because the context does not change too often, users can easily customize it without having to use the type to correspond to it every time.

for example:

```ts
import { LoaderFunction } from 'remix';

type MyContext  = {
  user: string
}

type LoaderFunctionWithContext = LoaderFunction<MyContext>

export const loader: LoaderFunctionWithContext = () => {}
```

### DIFF
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -3,6 +3,7 @@ import type { ServerRoute } from "./routes";
 import { json, isResponse, isRedirectResponse } from "./responses";
 
 /**
+ * @deprecated Please switch to use Generics type
  * An object of arbitrary for route loaders and actions provided by the
  * server's `getLoadContext()` function.
  */
--- a/packages/remix-server-runtime/routeModules.ts
+++ b/packages/remix-server-runtime/routeModules.ts
@@ -2,7 +2,7 @@ import type { Location } from "history";
 import type { ComponentType } from "react";
 import type { Params } from "react-router-dom";
 
-import type { AppLoadContext, AppData } from "./data";
+import type { AppData } from "./data";
 import type { LinkDescriptor } from "./links";
 import type { RouteData } from "./routeData";
 
@@ -13,9 +13,9 @@ export interface RouteModules<RouteModule> {
 /**
  * The arguments passed to ActionFunction and LoaderFunction.
  */
-export interface DataFunctionArgs {
+export interface DataFunctionArgs<Context = unknown> {
   request: Request;
-  context: AppLoadContext;
+  context: Context;
   params: Params;
 }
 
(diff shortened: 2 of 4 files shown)


machour:
Hi @rwu823 and sorry for the delays.
Can you rebase your PR to target the dev branch?

rwu823:
@machour done.

MichaelDeBoey:
@rwu823 Could you please rebase your branch on latest `dev` & resolve conflicts?

TillaTheHun0:
Just came across this today. +1

My use case is that we would like to use the context for dependency injection w.r.t Actions and Loaders, so it would be nice to type the `context` and get autocomplete without having to typecast in each loader.

mattrigg9:
Just checking in, any way we can get this merged in? This becomes especially helpful now that the [meta function is typed](https://github.com/remix-run/remix/releases/tag/remix%401.7.0) off of the loader. Without this change, meta will always consider data to be `any` if the loader function is typed.

packages/remix-server-runtime/data.ts
```diff
@@ -3,6 +3,7 @@ import type { ServerRoute } from "./routes";
 import { json, isResponse, isRedirectResponse } from "./responses";
 
 /**
+ * @deprecated Please switch to use Generics type
```
chaance, line 6:
```suggestion
 * @deprecated This type is an alias of `any` and will be removed in the future
```

packages/remix-server-runtime/routeModules.ts
```diff
@@ -13,9 +13,9 @@ export interface RouteModules<RouteModule> {
 /**
  * The arguments passed to ActionFunction and LoaderFunction.
  */
-export interface DataFunctionArgs {
+export interface DataFunctionArgs<Context = unknown> {
```
chaance, line 16:
```suggestion
export interface DataFunctionArgs<Context = any> {
```

While `unknown` is better, it would be a breaking change. Let's change this to `any` for the time being.

packages/remix-server-runtime/routeModules.ts
```diff
@@ -63,8 +63,8 @@ export interface LinksFunction {
 /**
  * A function that loads data for a route.
  */
-export interface LoaderFunction {
-  (args: DataFunctionArgs):
+export interface LoaderFunction<Context = any> {
+  (args: DataFunctionArgs<Context>):
```
chaance, line 67: `DataFunctionArgs` isn't currently a generic. Did you mean to update this? We'd also need to update `LoaderArgs` and `ActionArgs`.

packages/remix-server-runtime/server.ts
```diff
@@ -14,9 +13,9 @@ import { createRoutes } from "./routes";
 import { json, isRedirectResponse, isCatchResponse } from "./responses";
 import { createServerHandoffString } from "./serverHandoff";
 
-export type RequestHandler = (
+export type RequestHandler<Context = unknown> = (
```
chaance, line 16:
```suggestion
export type RequestHandler<Context = any> = (
```

chaance (changes requested):

PR #2123, opened 2022-02-25
fix: pin eslint-plugin-react version

Colt45s:
Seems to be causing a bug now at eslint-plugin-react@7.29.0.
fix it to the latest working version.

https://github.com/yannickcr/eslint-plugin-react/issues/3215


kentcdodds:
Thanks. Looks like this has been resolved by eslint-plugin-react.

PR #2126, opened 2022-02-25
docs(example): Route transition animation using Framer Motion

Girish21:


kentcdodds (approved):
Perfect! Thank you for explaining what's going on in the README. It's so simple!

PR #645, opened 2021-11-25
feat(cloudflare+cloudflare-pages+cloudflare-workers+templates): add support for `@cloudflare/workers-types` v3

edmundhung:
Cloudflare released an [update](https://github.com/cloudflare/workers-types/releases/tag/v3.0.0) for `@cloudflare/workers-types` last month. It is now automatically generated from the source as announced [here](https://blog.cloudflare.com/automatically-generated-types/) and this ensures typing to be always up-to-dated with the latest runtime.

However, the current CF adapter has a peerDependency of v2 only. It would be nice if we could add support of v3 as well.
This will also makes experimenting with new ESM worker easier (as all new types are only defined on v3) before official support is ready. 

Thanks for all the great work!

PR #2128, opened 2022-02-25
feat(remix-cloudflare-workers): add support for `@cloudflare/workers-types` v3

jfsiii:
## Symptoms
I'm getting TS errors like `Property 'request' does not exist on type 'FetchEvent'.ts(2339)` when writing an event listener like

```
addEventListener('fetch', event => {
  console.log(event, event.request, event.eventPhase)
})
```
Screen Shot 2022-02-25 at 10 47 21 AM

## Root cause
The [type specified](https://github.com/cloudflare/workers-types/blob/v2.2.0/index.d.ts#L1-L3) is from version `2.2.2` but the current version is `3.4.0`.

The [v3 types](https://github.com/cloudflare/workers-types/blob/74c94f8f96fa427353acd60ebc00d910f8e7cdfe/index.d.ts#L570-L575
) have `request` and all the other properties that logging shows are available

## Proposed solution
Change package.json to allow users to use the current version. I wasn't sure if it was preferable to a) replace the v2 range with the v3 range b) extend the range to allow v2 or v3 

I chose the more permissive option b, but do not have a strong opinion so long as it's possible to use the current major version

- [ ] Docs
- [ ] Tests

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -113,6 +113,7 @@
 - jeremyjfleming
 - jesse-deboer
 - jesseflorig
+- jfsiii
 - jgarrow
 - jkup
 - jmasson
--- a/packages/remix-cloudflare-workers/package.json
+++ b/packages/remix-cloudflare-workers/package.json
@@ -18,9 +18,9 @@
     "@remix-run/server-runtime": "1.2.2"
   },
   "peerDependencies": {
-    "@cloudflare/workers-types": "^2.2.2"
+    "@cloudflare/workers-types": "^2.2.2 || ^3.0.0"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^2.2.2"
+    "@cloudflare/workers-types": "^2.2.2 || ^3.0.0"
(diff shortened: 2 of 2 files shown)


jacob-ebey:
I like this option. I've been putting off upgrading to V3 as it's a "breaking change". This wouldn't be. @ryanflorence let's get this merged and I can add in the CF ESM handler without a major bump.

jacob-ebey:
@jfsiii please re-target this PR to the dev branch. Code changes always go there as per: https://github.com/remix-run/remix/blob/main/docs/pages/contributing.md#making-a-pull-request

jfsiii:
@jacob-ebey Thanks for the 👀 and tip. Apologies for any git change noise, but I think this ready now.

machour:
Already addressed in #645 (but I think it was doing too much)

MichaelDeBoey:
Closing in favour of #2530, as that one's addressing all Cloudflare packages/templates.

PR #2129, opened 2022-02-25
chore: DRY up shared logic between links/transitions

brophdawg11:
DRY up route matching/filtering logic between links/transitions

- [x] Docs - n/a
- [x] Tests - internal refactor should leverage current tests

### DIFF
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -291,30 +291,28 @@ export function getNewMatchesForLinks(
 ): ClientMatch[] {
   let path = parsePathPatch(page);
 
-  let prevUrl = new URL(
-    location.pathname + location.search + location.hash,
-    window.origin
-  );
-  let url = new URL(page, window.origin);
-  let filterByRouteProps = filterByRoutePropsFactory(
-    currentMatches,
-    prevUrl,
-    url
-  );
-
   // NOTE: keep this mostly up-to-date w/ the transition data diff, but this
   // version doesn't care about submissions
-  let newMatches =
-    mode === "data" && location.search !== path.search
-      ? nextMatches.filter(filterByRouteProps)
-      : nextMatches.filter((match, index) => {
-          return (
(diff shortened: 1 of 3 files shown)


packages/remix-react/links.ts
```diff
@@ -282,60 +285,31 @@ export async function getStylesheetPrefetchLinks(
...
-  let matchPathChanged = (match: RouteMatch<ClientRoute>, index: number) => {
-    return (
-      // param change, /users/123 -> /users/456
-      currentMatches[index].pathname !== match.pathname ||
-      // splat param changed, which is not present in match.path
-      // e.g. /files/images/avatar.jpg -> files/finances.xls
-      (currentMatches[index].route.path?.endsWith("*") &&
-        currentMatches[index].params["*"] !== match.params["*"])
-    );
-  };
```
brophdawg11, line 306: These are private functions inside `routeMatching.ts` now, composed and exposed in a single `shouldReloadForMatch(currentMatches, match, index)` method

packages/remix-react/links.ts
```diff
@@ -282,60 +285,31 @@ export async function getStylesheetPrefetchLinks(
...
-              params: match.params,
-              prevUrl: new URL(
-                location.pathname + location.search + location.hash,
-                window.origin
-              ),
-              url: new URL(page, window.origin),
-            });
-          }
-          return true;
-        })
```
brophdawg11, line 334: The filter function here is identical between both usages, requiring `currentMatches`, `prevUrl`, `url`, and optionally `submission`.  Went with a factory to put them in a closure for now, but this could also be done without the factory and with a few inline lambdas:

```javascript
nextMatches.filter((match, index) => filterByRouteProps(
    currentMatches, 
    prevUrl, 
    url,
    undefined,
    match, 
    index
));
```

brophdawg11:
Closing to revisit once in-progress transition manager updates are completed

PR #2130, opened 2022-02-25
docs(examples): Fix expire bug in redis-upstash-session example

gavriguy:
The previous implementation didn't really added correctly the EXPIRATION_DURATION_IN_SECONDS


kentcdodds (approved):
Thank you 👍

PR #2133, opened 2022-02-25
testing progress

kentcdodds:
Here's another approach. I'm thinking maybe we could generate the manifest and things at runtime instead of hard code it... Maybe?

Closes: #372

- [ ] Docs
- [ ] Tests


MichaelDeBoey:
@kentcdodds Do we still need this one now that we have #2481 & #372 is closed?

kentcdodds:
You're right. Thanks!

PR #2137, opened 2022-02-26
docs(examples): remove not allowed and unneeded div in blog example

mkrtchian:
This `<div>` is not allowed as child of `<label>` (cf. [div spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element)).
And it's not needed here even for visual purposes because the `<input>` below is already set to `display: block; width: 100%;` [in the CSS](https://github.com/remix-run/remix/blob/9ea7d4ae37c59082cff9d8011f609268b42e15dd/examples/blog-tutorial/app/styles/global.css#L66).


machour (approved):

MichaelDeBoey:
Maybe we should have replaced it with a `<span>` instead? 🤔

PR #2141, opened 2022-02-26
docs(example): Refactor bullmq task queue example

m5r:
This change refactors the bullmq queue example to make it easier to register a new queue and allow for better types inference when queuing a task.


kentcdodds (approved):
Yes, this is an improvement. Thank you!

PR #2142, opened 2022-02-26
docs(examples): remove environment check when using `LiveReload`

MichaelDeBoey:
Because of @kentcdodds' #1352


kentcdodds (approved):
Thanks!

PR #2143, opened 2022-02-27
docs(styling): update Tailwind CSS docs

MichaelDeBoey:
I noticed the `package.json` was outdated & all templates/stacks are suggesting to use `npm-run-all`


MichaelDeBoey:
@kentcdodds This will bring the docs in line with the example again

kentcdodds (approved):
Thanks!

PR #2144, opened 2022-02-27
docs(examples): update Tailwind CSS example

MichaelDeBoey:
Bringing the Tailwind CSS example in line with the new templates + changes I made in #2143

### DIFF
--- a/examples/tailwindcss/package.json
+++ b/examples/tailwindcss/package.json
@@ -1,35 +1,40 @@
 {
+  "name": "remix-template-tailwindcss",
   "private": true,
-  "name": "remix-example-tailwindcss",
   "description": "",
   "license": "",
+  "sideEffects": false,
   "scripts": {
-    "build": "npm run build:css && remix build",
-    "build:css": "tailwindcss -o ./app/tailwind.css",
-    "dev": "concurrently \"npm run dev:css\" \"remix dev\"",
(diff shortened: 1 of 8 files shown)


examples/tailwindcss/package.json
```diff
@@ -1,35 +1,38 @@
...
+    "dev:remix": "cross-env NODE_ENV=development remix dev",
+    "generate:css": "npx tailwindcss -o ./app/tailwind.css",
     "postinstall": "remix setup node",
-    "start": "remix-serve build"
+    "start": "cross-env NODE_ENV=production remix-serve build"
   },
   "dependencies": {
     "@remix-run/react": "1.2.2",
     "@remix-run/serve": "1.2.2",
+    "cross-env": "^7.0.3",
```
Hamatek, line 21: maybe we could move it to devDependencies

MichaelDeBoey: Just copied over the base template, but should indeed be in `devDependencies` 👍

Fixed the template in #2157

jacob-ebey: `cross-env` is used in the start script, it is in fact a runtime dependency and therefore belongs in `dependencies` not `devDependencies`.

MichaelDeBoey: @jacob-ebey Shouldn't it be a dependency of `remix` itself then? 🤔

PR #2147, opened 2022-02-27
feat: switch deno template to use HTTP imports

jacob-ebey:
feat: added bundling of HTTP imports for browser build

Closes: #

- [ ] Docs
- [ ] Tests


packages/create-remix/templates/deno/package.json
```diff
@@ -7,9 +7,6 @@
     "dev": "cross-env NODE_ENV=development remix build && run-p dev:*",
     "start": "cross-env NODE_ENV=production deno run --allow-net --allow-read --allow-env ./build/index.js"
   },
-  "dependencies": {
-    "@remix-run/deno": "*"
```
pcattori, line 11: `npm install` on a project created from the `deno` template fails for me with `Cannot find module '@remix-run/deno/package.json'`. I added `@remix-run/deno` as a dependency of my project as a workaround.

pcattori: Alternatively, if we want to forgo `@remix-run/deno`, we'd need to remove or alter the `postinstall` setup step in the project's `package.json`.

pcattori:
Superceded by #2243

PR #2148, opened 2022-02-27
refactor(remix-dev): remove unused code + simplify some things

ftonato:
Hi, thanks for this amazing project 🚀 

## Description

I was browsing through the code, trying to figure out if there was anything "simple" where I could start contributing, and I discovered very simple details that can be considered the first step (to me), so I decided to make some changes to help the project (_if the changes add any value_ 🤷).

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -317,7 +317,7 @@ async function buildEverything(
 
 async function createBrowserBuild(
   config: RemixConfig,
-  options: BuildOptions & { incremental?: boolean }
+  options: Required<BuildOptions> & { incremental?: boolean }
 ): Promise<esbuild.BuildResult> {
   // For the browser build, exclude node built-ins that don't have a
   // browser-safe alternative installed in node_modules. Nothing should
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -428,8 +428,9 @@ function addTrailingSlash(path: string): string {
   return path.endsWith("/") ? path : path + "/";
 }
 
+const entryExts = [".js", ".jsx", ".ts", ".tsx"];
+
 function findEntry(dir: string, basename: string): string | undefined {
-  let entryExts = [".js", ".jsx", ".ts", ".tsx"];
   for (let ext of entryExts) {
     let file = path.resolve(dir, basename + ext);
     if (fse.existsSync(file)) return path.relative(dir, file);
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -377,7 +377,7 @@ export async function readConfig(
   let routes: RouteManifest = {
     root: { path: "", id: "root", file: rootRouteFile },
   };
-  function addRoutes(data) {
+  function addRoutes(data: RouteManifest) {
     for (let key of Object.keys(data)) {
       let route = data[key];
       routes[route.id] = { ...route, parentId: route.parentId || "root" };
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -260,12 +260,12 @@ export async function readConfig(
   remixRoot?: string,
   serverMode = ServerMode.Production
 ): Promise<RemixConfig> {
-  if (!remixRoot) {
-    remixRoot = process.env.REMIX_ROOT || process.cwd();
-  }
-
   if (!isValidServerMode(serverMode)) {
     throw new Error(`Invalid server mode "${serverMode}"`);
+  }
+
+  if (!remixRoot) {
+    remixRoot = process.env.REMIX_ROOT || process.cwd();
   }
 
   let rootDirectory = path.resolve(remixRoot);
@@ -428,9 +428,8 @@ function addTrailingSlash(path: string): string {
   return path.endsWith("/") ? path : path + "/";
 }
 
-const entryExts = [".js", ".jsx", ".ts", ".tsx"];
-
 function findEntry(dir: string, basename: string): string | undefined {
+  let entryExts = [".js", ".jsx", ".ts", ".tsx"];
   for (let ext of entryExts) {
     let file = path.resolve(dir, basename + ext);
     if (fse.existsSync(file)) return path.relative(dir, file);
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -377,22 +377,22 @@ export async function readConfig(
   let routes: RouteManifest = {
     root: { path: "", id: "root", file: rootRouteFile },
   };
+  function addRoutes(data) {
+    for (let key of Object.keys(data)) {
+      let route = data[key];
+      routes[route.id] = { ...route, parentId: route.parentId || "root" };
+    }
+  }
   if (fse.existsSync(path.resolve(appDirectory, "routes"))) {
     let conventionalRoutes = defineConventionalRoutes(
       appDirectory,
       appConfig.ignoredRouteFiles
     );
-    for (let key of Object.keys(conventionalRoutes)) {
-      let route = conventionalRoutes[key];
-      routes[route.id] = { ...route, parentId: route.parentId || "root" };
-    }
+    addRoutes(conventionalRoutes);
   }
   if (appConfig.routes) {
     let manualRoutes = await appConfig.routes(defineRoutes);
-    for (let key of Object.keys(manualRoutes)) {
-      let route = manualRoutes[key];
-      routes[route.id] = { ...route, parentId: route.parentId || "root" };
-    }
+    addRoutes(manualRoutes);
   }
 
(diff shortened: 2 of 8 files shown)


packages/remix-dev/config.ts
```diff
@@ -422,9 +422,8 @@ function addTrailingSlash(path: string): string {
   return path.endsWith("/") ? path : path + "/";
 }
 
-const entryExts = [".js", ".jsx", ".ts", ".tsx"];
-
 function findEntry(dir: string, basename: string): string | undefined {
+  const entryExts = [".js", ".jsx", ".ts", ".tsx"];
```
cliffordfajardo, line 426: This like an improvement as `extryExt` wasn't used anywhere else

packages/remix-dev/config.ts
```diff
@@ -260,21 +260,21 @@ export async function readConfig(
...
-  }
-
   if (!isValidServerMode(serverMode)) {
     throw new Error(`Invalid server mode "${serverMode}"`);
   }
 
+  if (!remixRoot) {
+    remixRoot = process.env.REMIX_ROOT || process.cwd();
+  }
+
```
cliffordfajardo, line 270: @ftonato - out of curiosity, what's the benefit of moving this after the `if (!isValidServerMode(serverMode)) {` conditional block?

packages/remix-dev/compiler.ts
```diff
@@ -311,16 +311,25 @@ async function buildEverything(
...
+  let externals = []
+  let fakeBuiltins = [];
+  for (let i = 0; i < nodeBuiltins.length; i++) {
+    let mod = nodeBuiltins[i];
+    if (dependencies.includes(mod)) {
+      fakeBuiltins.push(mod);
+    } else {
+      externals.push(mod);
+    }
+  }
```
cliffordfajardo, line 332: Doing this in 1 loop might speed things up
Maybe we could use a "for of" loop here?

```ts
const externals = [];
const fakeBuiltins = [];
for(const mod of nodBuiltins) {  
    if (dependencies.includes(mod)) {
      fakeBuiltins.push(mod);
    } else {
      externals.push(mod);
    }
}
```

cliffordfajardo (commented):
.

packages/remix-dev/config.ts
ftonato: @cliffordfajardo you should consider that if you have code like this:

``` js
const fn = x => {
    if (!x) throw new Error('Invalid parameter')
    
    let a = x // assign something to a variable
    console.log(a)
}

fn() /* if we don't have the parameter it will throw the error
        before "assigning" something that might not be used for anything */
```

packages/remix-dev/compiler.ts
ftonato: @cliffordfajardo probably using the traditional `for` would be faster than the `for..of`. I'll change it, thanks for the warning (I didn't even notice) o/

MichaelDeBoey: Although it might be "faster" (which is neglectable in this case imo, it's much less readable than having the built in `filter` methods, so I would rather revert this again.

packages/remix-dev/config.ts
MichaelDeBoey: I disagree.

Although this const is only used in the `findEntry`, it will be re-created on each method call.  
Having it outside of the method call only creates the const once, which is a cleaner way to do imo.

MichaelDeBoey (changes requested):

packages/remix-dev/config.ts
cliffordfajardo: > it will be re-created on each method call.
True, i haven't dug into this area of the code & don't know how often this gets called. If this gets called only once during dev or build time, then the `const` being defined inside just makes sure its get's garbage collected more quickly, immediately after the function is called

I do see your point though

packages/remix-dev/compiler.ts
cliffordfajardo: TBH - I do like the `filter `methods more & think they are more readable
When it comes to JS perf, switching from filter to `for` in terms of performance might be negligible (not saying its not important) its just that  given V8 is always changing internally, I don't often think about things like `for` vs `forEach` when the number of item's we're iterating over is low + what get's transpiled to production code javascript code, usually isn't written in the same way we write JS in the codebase & lets not get started on different JS runtimes lol 😅 , which all have different performance charasteristics

Appreciate the thought @MichaelDeBoey

packages/remix-dev/config.ts
ftonato: I definitely didn't notice the fact that the function was called in more places (currently 4 calls). Probably the "change" doesn't impact as much in terms of performance, but it makes more sense to keep it that way, since we don't know if the number of calls will increase or not.

MichaelDeBoey:
@ftonato Could you please rebase your branch onto latest `dev` branch & fix tests?

ftonato:
@MichaelDeBoey I will do it right now =)

ftonato:
@MichaelDeBoey done 🚀

MichaelDeBoey:
@ftonato Don't forget to fix tests too please

ftonato:
@MichaelDeBoey it's failing due to lack of typing, but I'll adjust that and see if the tests progress o/

packages/remix-dev/compiler.ts
```diff
@@ -311,7 +311,7 @@ async function buildEverything(
 
 async function createBrowserBuild(
   config: RemixConfig,
-  options: BuildOptions & { incremental?: boolean }
+  options: Required<BuildOptions> & { incremental?: boolean }
```
MichaelDeBoey, line 314: Is there a reason why you've put `Required` around this?

This would technically be a breaking change.

ftonato: Well, taking a look at the code we can see that the method
`createBrowserBuild` is invoked by `buildEverything` (reference: image below), for this method the `option` parameter is _required_.

So, we have two methods: **A** that invoke **B**.
1. Method A treats parameter X as required;
2. Method A sends the same parameter received directly to method B;
2.1. So method B should receive something in the same format, as "required" (right?);
3. Therefore, if method A does not receive something that is `required` it should throw an error and will never get to invoke method B.


image


@MichaelDeBoey Does this make sense?

MichaelDeBoey: It doesn't matter that the parent function (`buildEverything`) is more strict than the child function (`createBrowserBuild`). If it would be the other way around, that would indeed be a problem.

Since `createBrowserBuild` is only used in `buildEverything`, it's fine to make `createBrowserBuild` more strict too. 👍

packages/remix-dev/config.ts
```diff
@@ -371,22 +371,22 @@ export async function readConfig(
   let routes: RouteManifest = {
     root: { path: "", id: "root", file: rootRouteFile },
   };
+  function addRoutes(data: RouteManifest) {
+    for (let key of Object.keys(data)) {
+      let route = data[key];
+      routes[route.id] = { ...route, parentId: route.parentId || "root" };
+    }
+  }
```
MichaelDeBoey, line 379: While we're at it, let's make this

```suggestion
  let addRoutes = (routes: RouteManifest) =>
    Object.values(routes).forEach((route) => {
      routes[route.id] = { ...route, parentId: route.parentId || "root" };
    });
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

packages/remix-dev/config.ts
```diff
@@ -371,22 +371,20 @@ export async function readConfig(
   let routes: RouteManifest = {
     root: { path: "", id: "root", file: rootRouteFile },
   };
+  let addRoutes = (routes: RouteManifest) =>
+    Object.values(routes).forEach((route) => {
+      routes[route.id] = { ...route, parentId: route.parentId || "root" };
+    });
```
MichaelDeBoey, line 377: @ftonato These changes are causing tests to fail, so let's revert them (for now)

```suggestion
  function addRoutes(data: RouteManifest) {
    for (let key of Object.keys(data)) {
      let route = data[key];
      routes[route.id] = { ...route, parentId: route.parentId || "root" };
    }
  }
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

MichaelDeBoey:
@ftonato Could you please rebase your branch onto latest `dev` & resolve conflicts?

ftonato:
@MichaelDeBoey updated as requested!

MichaelDeBoey:
@ftonato Seems like something went wrong & you added `fixtures/gists-app` while rebasing.

ftonato:
@MichaelDeBoey just saw it, I'm going to remove those files from this PR and notify you as soon as finished!

pcattori:
Superceded by https://github.com/remix-run/remix/pull/5220

Issue #1917, opened 2022-02-12
Docs: `─`  character is too narrow in Source Code Pro font


Aprillion:
### What version of Remix are you using?

N/A

### Steps to Reproduce

* navigate to https://remix.run/docs/en/v1/api/conventions#ignoring-search-params on Windows

### Expected Behavior

* all characters in a monospace font have the same width

### Actual Behavior

* character `─` is too narrow
  * screenshot from Windows 10 (same on Firefox 98.0b3 and Chrome 98.0):
    <img src="https://user-images.githubusercontent.com/1087670/153714585-00507dc8-ecd8-40ad-a0a1-0e94dc3667a2.png" width="25%">
  * kinda better on Mac (or uncanny-valley-worse?):
    <img src="https://user-images.githubusercontent.com/1087670/153714924-ae4e8bfc-c36c-4c67-a98e-1d327aa51cab.png" width="25%">

BasixKOR:
Isn't this a problem with the font itself?

Aprillion:
it probably is, but I don't care whether it's a wrong character, wrong font, or a bug in build steps, blaming some external cause won't improve the user experience for me ¯\\\_(ツ)_/¯

sobrinho:
Created a PR using ascii chars to avoid this, it is currently broken on my macOS too.

machour:
Tracked in #2150

Aprillion:
Still looks 💩 on Windows:
![image](https://user-images.githubusercontent.com/1087670/162574696-83316afe-dbef-44ff-8044-d862dd3d98bb.png)

PR #2150, opened 2022-02-27
docs: use ASCII characters

sobrinho:
Closes: #1917

- [x] Docs
- [ ] Tests

Screen Shot 2022-02-27 at 18 52 42

### DIFF
--- a/docs/api/conventions.md
+++ b/docs/api/conventions.md
@@ -1310,18 +1310,18 @@ Consider these routes:
 And lets say the UI looks something like this:
 
 ```
-┌──────────────────────────────┐
-│    Project: Design Revamp    │
-├────────┬─────────┬───────────┤
-│  Tasks │ Collabs │ >ACTIVITY │
-├────────┴─────────┴───────────┤
-│  Search: _____________       │
-│                              │
-│  - Ryan added an image       │
-│                              │
-│  - Michael commented         │
-│                              │
-└──────────────────────────────┘
++------------------------------+
+|    Project: Design Revamp    |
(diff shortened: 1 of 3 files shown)


sobrinho:
I believe I replaced every char to a ascii equivalent.

If there are more, let me know!

The script I used is at the first commit message.

docs/api/conventions.md
```diff
@@ -187,8 +187,8 @@ Please note that you can use either `.jsx` or `.tsx` file extensions depending o
 <!-- prettier-ignore -->
 ```markdown [3]
 app/
-├── routes/
-└── root.tsx
```
kentcdodds, line 191: Does this look bad on non-macs? I think these should be fine as-is... It's just the table thing we had that should be changed I think.

kiliman: Windows uses the same characters, so should be fine.
```
Volume serial number is EE88-C117
C:\PROJECTS\SCRATCH\MY-REMIX-APP\APP
│   entry.client.tsx
│   entry.server.tsx
│   root.tsx
│
└───routes
        index.tsx
```

MichaelDeBoey:
@sobrinho As @kentcdodds & @kiliman pointed out, you should only change the tables, not the file trees.

Please rebase your branch on latest `main` & revert changes to anything other than the tables.

sobrinho:
Done but that way the scripts won't be automatic if you try again since they will change all the chars as I wrote it.

kentcdodds (approved):
Thanks for taking the time to do this 👍

PR #2155, opened 2022-02-28
feat(remix-server-runtime): throw error when cookie limit is reached

matt-l-w:
Closes: N/A

- [ ] Docs
- [x] Tests

Session storage inside a cookie is great for smaller things and
normally "just works". For newer developers (and older forgetful ones
like myself), the storage limit can be a source of errors that is
easily missed.

This makes the commit action for cookie session storage throw an
error describing the problem if data is attempted to be persisted that
will exceed the browser max of 4096 bytes.

### DIFF
--- a/packages/remix-server-runtime/sessions/cookieStorage.ts
+++ b/packages/remix-server-runtime/sessions/cookieStorage.ts
@@ -44,7 +44,10 @@ export const createCookieSessionStorageFactory =
     async commitSession(session, options) {
       const serializedCookie = await cookie.serialize(session.data, options);
       if (serializedCookie.length > 4096) {
-        throw new Error('Cookie length will exceed browser maximum. Length: ' + serializedCookie.length)
+        throw new Error(
+          "Cookie length will exceed browser maximum. Length: " +
+            serializedCookie.length
+        );
       }
       return serializedCookie;
     },
--- a/packages/remix-server-runtime/sessions/cookieStorage.ts
+++ b/packages/remix-server-runtime/sessions/cookieStorage.ts
@@ -58,3 +58,4 @@ export const createCookieSessionStorageFactory =
       });
     },
   };
+};
--- a/packages/remix-server-runtime/sessions/cookieStorage.ts
+++ b/packages/remix-server-runtime/sessions/cookieStorage.ts
@@ -43,9 +43,8 @@ export const createCookieSessionStorageFactory =
     },
     async commitSession(session, options) {
       const serializedCookie = await cookie.serialize(session.data, options);
-      const byteLength = Buffer.byteLength(serializedCookie);
-      if (byteLength > 4096) {
-        throw new Error('Cookie length will exceed browser maximum. Length: ' + byteLength)
+      if (serializedCookie.length > 4096) {
+        throw new Error('Cookie length will exceed browser maximum. Length: ' + serializedCookie.length)
       }
       return serializedCookie;
     },
--- a/packages/remix-server-runtime/sessions/cookieStorage.ts
+++ b/packages/remix-server-runtime/sessions/cookieStorage.ts
@@ -35,20 +35,24 @@ export const createCookieSessionStorageFactory =
 
     warnOnceAboutSigningSessionCookie(cookie);
 
-    return {
-      async getSession(cookieHeader, options) {
-        return createSession(
-          (cookieHeader && (await cookie.parse(cookieHeader, options))) || {}
-        );
-      },
-      async commitSession(session, options) {
-        return cookie.serialize(session.data, options);
-      },
-      async destroySession(_session, options) {
-        return cookie.serialize("", {
-          ...options,
-          expires: new Date(0),
-        });
-      },
-    };
+  return {
+    async getSession(cookieHeader, options) {
+      return createSession(
+        (cookieHeader && (await cookie.parse(cookieHeader, options))) || {}
+      );
+    },
+    async commitSession(session, options) {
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -9,14 +9,13 @@
   dependencies:
     "@jridgewell/trace-mapping" "^0.3.0"
 
-"@architect/functions@^4.1.1":
```
MichaelDeBoey, line 12: Please revert these changes, as you didn't update any `package.json`

MichaelDeBoey (changes requested):

yarn.lock
matt-l-w: These is from a totally fresh install, @MichaelDeBoey. Any changes will be because of some dependency in the various workspaces sepcified it or something else depends on it. For example, this one:

```sh
$ yarn why @architect/functions
yarn why v1.22.15
[1/4] 🤔  Why do we have the module "@architect/functions"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@architect/functions@5.0.3"
info Reasons this module exists
   - "_project_#@remix-run#architect" depends on it
   - Hoisted from "_project_#@remix-run#architect#@architect#functions"
info Disk size without dependencies: "256KB"
info Disk size with unique dependencies: "452KB"
info Disk size with transitive dependencies: "560KB"
info Number of shared dependencies: 7
✨  Done in 0.32s.
```

Are there any that you're specifically concerned about?

MichaelDeBoey: If we want to update these, they should be done in a separate PR imo.

sergiodxa (approved):
I like this change, it'll be really useful to avoid this error and probably know about it

yarn.lock
matt-l-w: Sure thing - I've opened it at #2163 and reverted the changes here.

matt-l-w: I'll leave to you to resolve, @MichaelDeBoey

MichaelDeBoey (approved):

kentcdodds (approved):
LGTM, but I'm still not 100% comfortable merging stuff into remix core yet, so I'll ping the team about this one.

packages/remix-server-runtime/sessions/cookieStorage.ts
```diff
@@ -37,7 +37,12 @@ export function createCookieSessionStorage({
       );
     },
     async commitSession(session, options) {
-      return cookie.serialize(session.data, options);
+      const serializedCookie = await cookie.serialize(session.data, options);
+      const byteLength = Buffer.byteLength(serializedCookie);
```
mjackson, line 41: We can't use node primitives like `Buffer` in `@remix-run/server-runtime`, because this package is supposed to be platform-agnostic.  Unless we care about multi-byte characters here, it should be acceptable to just check `serializedCookie.length` instead.

mjackson (changes requested):

packages/remix-server-runtime/sessions/cookieStorage.ts
mjackson: Also, please reuse existing code conventions and use `let` instead of `const` for local variables.

matt-l-w: Resolved both of these. Thanks for pointing out @mjackson .

On the let vs. const - this convention is totally opaque to newcomers and seems like it would be a hassle for maintainers to repeat. Perhaps an [eslint plugin](https://www.npmjs.com/package/eslint-plugin-prefer-let) would be an easier way to enforce things like this.

MichaelDeBoey:
@matt-l-w Could you please rebase on latest `dev` & resolve conflicts?

mcansh (approved):

mjackson (approved):

matt-l-w:
> @matt-l-w Could you please rebase on latest `dev` & resolve conflicts?

Look good to go for me now 👍

machour (approved):

yarn.lock
```diff
@@ -9,13 +9,14 @@
   dependencies:
     "@jridgewell/trace-mapping" "^0.3.0"
 
-"@architect/functions@^5.0.2":
```
MichaelDeBoey, line 12: Please revert these changes, as you didn't update any `package.json`

MichaelDeBoey (changes requested):

yarn.lock
MichaelDeBoey: They somehow got back, so please remove the changes to `yarn.lock` again.

The easiest way to do this is to run
```sh
git checkout dev ./yarn.lock
```
in the root of the project.

matt-l-w: This is now the _opposite_ of your original complaint though. Originally, I updated it to 5.0.2 because that's what I got with a fresh install. You requested that was reverted and dealt with in a different issue which is what we went for.

If I revert this _again_, we'll be back to square one. There must be something going wrong here.

MichaelDeBoey: I don't want it to be on a specific version, I just want to keep whatever we currently have in `dev` branch.
So probably this was changes in `dev` branch and that's why it looks like we would be back at square one 🤷‍♂️

But I just want to see **no** changes to `yarn.lock`

matt-l-w: I've plucked it out of the commit history. No diffs for `yarn.lock` now.

MichaelDeBoey (approved):

kentcdodds (approved):
Stellar, thank you!

PR #2156, opened 2022-02-28
docs(example): Remix + MSW example

Girish21:


MichaelDeBoey:
CC/ @kettanaito

examples/msw/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Preview
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+[![Open in CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/msw)
+
+## Example
+
+If the external APIs are metered and charged for the number of API calls made, it's pretty easy to burst through the API quota during development. Also, we need an active network during development. For some, this may be an issue. So instead, we can mock the external API using MSW, which intercepts the API calls from the server and returns a mocked response.
```
kettanaito, line 13: Those are some great points! 

When I was researching the applications of API mocking, having the ability to develop against a non-existing backend was one of the most common ones. This also applies to situations when you have a backend but it's not in the state that you need (you can augment responses with mocking as well). 

API mocking is also powerful for debugging, as you can reliably reproduce any HTTP call by mocking it. 

Perhaps, consider some of these points. You can read more about them [here](https://mswjs.io/docs/#when-to-mock-api).

examples/msw/README.md
```diff
@@ -0,0 +1,25 @@
+# Remix + MSW
+
+This example demonstrates [MSW's][msw] usage with Remix to mock external APIs called from the Remix server during development.
```
kettanaito, line 3: It's a nitpick but the word "external" may make an impression that MSW is only for third-party API mocking. I think you're safer saying just "to mock any HTTP calls" (I'd even omit API as it can mean a bunch of things depending on the context).

examples/msw/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+[![Open in CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/msw)
+
+## Example
+
+If the external APIs are metered and charged for the number of API calls made, it's pretty easy to burst through the API quota during development. Also, we need an active network during development. For some, this may be an issue. So instead, we can mock the external API using MSW, which intercepts the API calls from the server and returns a mocked response.
+
+## Relevant files
+
+- [mocks](./mocks/index.js)
```
kettanaito, line 17: This is neat! I like having sections like this one for people to quickly see where the relevant setup is located. One thing I may suggest is to also add a brief description of what a particular file achieves (i.e. "describes how to intercept requests", or "registers the service worker").

examples/msw/mocks/index.js
```diff
@@ -0,0 +1,15 @@
...
+    return res(ctx.status(200), ctx.json({ message: "from msw" }));
+  }),
+];
+
+const server = setupServer(...handlers);
+server.listen({ onUnhandledRequest: "warn" });
+console.info("MSW initialised");
+
+process.once("SIGINT", () => server.close());
+process.once("SIGTERM", () => server.close());
```
kettanaito, line 15: To be fair, if the process is terminated there's no way the `server` would keep doing anything. Keep in mind that despite the name, there are no actual HTTP servers establishes. `setupServer` is a plain function that listens to intercepted events and handles them.

Perhaps you had some other scenarios in mind for these process events?

kettanaito, line 15: I don't think this module is used anywhere yet, but we need to take Live Reload into account. We need to ensure that `setupServer` is not called repeatedly, as that will lead to event emitter leaks.

kettanaito:
This is definitely a good start!

I think the example is yet to add a browser integration (`setupWorker`) and figure out the right places to establish them in Remix. I'd suspect that adding a Node.js integration to `root.jsx` may be the case (watch out for Live Reload), not sure where the worker registration would go.

examples/msw/mocks/index.js
rodrigofuentes: @kettanaito @Girish21 It seems that headers aren't being read properly during unhandled requests. 


```sh
[MSW] Expected response resolver to return a mocked response Object, but got undefined. The original response is going to be used instead.

...

Response {
  size: 0,
  timeout: 0,
  [Symbol(Body internals)]: {
    body: PassThrough {
      _readableState: [ReadableState],
      _events: [Object: null prototype],
      _eventsCount: 5,
      _maxListeners: undefined,
      _writableState: [WritableState],
      allowHalfOpen: true,
      [Symbol(kCapture)]: false,
      [Symbol(kCallback)]: null
    },
    disturbed: false,
    error: null
  },
  [Symbol(Response internals)]: {
    url: 'https://randomuser.me/api',
    status: 200,
    statusText: 'OK',
    headers: Headers { [Symbol(map)]: [Object: null prototype] },
    counter: 0
  }
}
```

in interceptors: `ClientRequest/utils/getIncomingMessageBody.ts` and `ClientRequest/NodeClientRequest.ts`

rodrigofuentes:
@kettanaito @Girish21 It seems that headers aren't being read properly during unhandled requests. 


```sh
[MSW] Expected response resolver to return a mocked response Object, but got undefined. The original response is going to be used instead.

...

Response {
  size: 0,
  timeout: 0,
  [Symbol(Body internals)]: {
    body: PassThrough {
      _readableState: [ReadableState],
      _events: [Object: null prototype],
      _eventsCount: 5,
      _maxListeners: undefined,
      _writableState: [WritableState],
      allowHalfOpen: true,
      [Symbol(kCapture)]: false,
      [Symbol(kCallback)]: null
    },
    disturbed: false,
    error: null
  },
  [Symbol(Response internals)]: {
    url: 'https://randomuser.me/api',
    status: 200,
    statusText: 'OK',
    headers: Headers { [Symbol(map)]: [Object: null prototype] },
    counter: 0
  }
}
```

in interceptors: `ClientRequest/utils/getIncomingMessageBody.ts` and `ClientRequest/NodeClientRequest.ts`

rodrigofuentes:
My other comment is irrelevant, Node version issue. 

https://github.com/mswjs/msw/issues/1125#issuecomment-1054868745

edmundhung:
It might be worth mentioning that MSW current does not support intercepting server requests on Cloudflare Workers / Pages which uses `undici` instead of typical http based implementation.

Ref: https://github.com/mswjs/interceptors/issues/159

MichaelDeBoey:
`msw@0.39.0` is released.

https://github.com/mswjs/msw/releases/tag/v0.39.0

kentcdodds:
I say we upgrade, but it may be good to add a note that v0.39.0 requires Node v16. AWS currently does not support Node v16 (😭) so if you want to use the latest MSW and you deploy to AWS, you'll be running a different version of Node locally from the one that runs in production 😬

Girish21:
Wait. MSW in production? what would be the use case for that? 😅

examples/msw/mocks/index.js
Girish21: @kentcdodds, what do you think about adding the listeners?

kentcdodds: I've never done it and I don't know what situation would necessitate this. I think it's unnecessary

kentcdodds: I can confirm that it will not be because we're using `-r` to require this file 👍

kentcdodds:
> Wait. MSW in production? what would be the use case for that? 😅

That's not what I meant. What I meant is that if you are deploying to AWS but you also want to use the latest version of MSW then locally you're using Node 16 but in production you're using Node 14. This is not advisable. So if you're deploying to AWS then you can't use the latest version of MSW.

> Regarding adding a browser integration

In my experience, it doesn't make much sense to use MSW in the browser with Remix because all your requests happen on the server via loaders/actions. You don't want to get in the business of mocking out the requests that Remix makes to loaders and actions. Instead it's much easier to mock out the third party services your server is calling. So I wouldn't bother with browser integration here. MSW + Remix is really quite simple :)

examples/msw/mocks/index.js
kettanaito: This setup makes sense if `server` spawned a child process, or any other side-effect detached from the current process' thread. That is not the case, so if the process exists/terminates it will also destroy the execution context for `server`. 

You can safely remove this.

kettanaito:
@rodrigofuentes, I think there's something wrong with the mocked response in that headers scenario (the error message above suggests that—nothing was returned from the handler). The headers appear to be in order in your code snippet. Note that `Headers` never store values and instead use prototype getters/setters to read/write them respectively. That's why you only see what appears an empty object when logging it out.

kentcdodds (changes requested):
Sweet! Just one last thing and then we can get this merged. Thanks a ton for iterating on this so much! It's great 👏👏

kentcdodds (approved):
Stellar 👏👏 Thanks a lot!

wingy3181:
Stupid question about this example or maybe something to add to the README is how to toggle MSW?

I got it working by `import `../mocks` in the `root.tsx` file but is this the proper technique?

Was comparing it to kent's setup on his website (https://github.com/kentcdodds/kentcdodds.com/blob/bb3a2839c236bd11702e9fc11bd696cd98ac7bc3/package.json#L35) and what he has talked about in his talks....... but it's a different setup...

Not sure if i should be asking this on discussions or not....or this closed PR
but will try here first

jmaldon1:
@wingy3181 This is how I ended up doing it. (Not sure why this isn't in the docs anywhere. Maybe this is wrong?)

```tsx
// app/entry.server.tsx
...
if (process.env.NODE_ENV === "development") {
  const server = require("../mocks");

  server.listen({ onUnhandledRequest: "warn" });
  console.info("MSW initialised");
}
...
```

```ts
// mocks/index.ts
import { setupServer } from "msw/node";
import { handlers } from "./handlers";

// This configures a request mocking server with the given request handlers.
const server = setupServer(...handlers);

module.exports = server;
```

PR #2157, opened 2022-02-28
chore(create-remix): put `cross-env` in `devDependencies`

MichaelDeBoey:


jacob-ebey:
`cross-env` is used in the start script, it is therefor a `dependency` and not `devDependency`.

MichaelDeBoey:
My mistake, was reading your first message wrongly 🙈

Thanks for the explanation on why it's set in the template.
Indeed logical to have it in `dependencies` 👍

PR #2158, opened 2022-02-28
docs(examples): update template example

MichaelDeBoey:
Bringing it in line with the new templates


examples/template/remix.config.js
```diff
@@ -2,8 +2,10 @@
...
-  appDirectory: "app",
-  browserBuildDirectory: "public/build",
-  publicPath: "/build/",
-  serverBuildDirectory: "build",
+  ignoredRouteFiles: [".*"],
+  // appDirectory: "app",
+  // assetsBuildDirectory: "public/build",
+  // serverBuildPath: "build/index.js",
+  // publicPath: "/build/",
+  // devServerPort: 8002
```
kentcdodds, line 10:
```suggestion
```

kentcdodds (approved):
Looks good to me 👍 Just one thing I'll fix.

kentcdodds:
Thanks!

examples/template/remix.config.js
kentcdodds: Yeah, that's automatically determined and can change to a random port if it's not available anyway. The only reason anyone would need to specify this is if they're running their app inside a docker container during development (which is extremely rare).

MichaelDeBoey: PR incoming then 😉

PR #2162, opened 2022-02-28
fix(remix-react): fix `Form` submit when having a button with `formMethod` attribute

mdoury:
I found an issue when submitting a `Form` which has a child submit button with the `formMethod` attribute set.
It seems the `formMethod` attribute is simply ignored, and the request method fallbacks to the default "get" method.
This is inconvenient as the form submission triggers the loader instead of the action.

I started investigating this issue and I have a feeling it is occurring in the implementation of the `Form` component (i.e. `FormImpl`). The `submit` function receives the form method as an option, which is then used to submit the form in spite of the `formMethod` attribute : 

packages/remix-react/components.tsx:948
```ts
submit(clickedButtonRef.current || event.currentTarget, {
    method,
    replace,
});
```

packages/remix-react/components.tsx:1108
```ts
method =
          options.method ||
          target.getAttribute("formmethod") ||
          form.getAttribute("method") ||
          defaultMethod;
```

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -1,9 +1,6 @@
 import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
 import type { Fixture, AppFixture } from "./helpers/create-fixture";
 
-let fixture: Fixture;
-let app: AppFixture;
-
 ////////////////////////////////////////////////////////////////////////////////
 // 💿 👋 Hola! It's me, Dora the Remix Disc, I'm here to help you write a great
 // bug report pull request. You don't need to fix the bug, this is just to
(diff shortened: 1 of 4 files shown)


integration/bug-report-test.ts
```diff
@@ -29,65 +26,87 @@ let app: AppFixture;
...
-          return json("pizza");
-        }
+  beforeAll(async () => {
+    fixture = await createFixture({
+      ////////////////////////////////////////////////////////////////////////////
+      // 💿 Next, add files to this object, just like files in a real app,
+      // `createFixture` will make an app and run your tests against it.
+      ////////////////////////////////////////////////////////////////////////////
+      files: {
+        "app/routes/working.jsx": js`
```
MichaelDeBoey, line 40:
```suggestion
        "app/routes/form-method.jsx": js`
```

integration/bug-report-test.ts
```diff
@@ -29,65 +26,87 @@ let app: AppFixture;
...
-  // You can test any request your app might get using `fixture`.
-  let response = await fixture.requestDocument("/");
-  expect(await response.text()).toMatch("pizza");
-
-  // If you need to test interactivity use the `app`
-  await app.goto("/");
-  await app.clickLink("/burgers");
-  expect(await app.getHtml()).toMatch("cheeseburger");
+  test('should use "post" method when clicking a submit button inside a form with attribute method="post"', async () => {
+    await app.goto("/working");
```
MichaelDeBoey, line 99:
```suggestion
    await app.goto("/form-method");
```

integration/bug-report-test.ts
```diff
@@ -29,65 +26,87 @@ let app: AppFixture;
...
-          let data = useLoaderData();
-          return (
-            <div>
-              {data}
-              <Link to="/burgers">Other Route</Link>
-            </div>
-          )
-        }
-      `,
+      "app/routes/not-working.jsx": js`
```
MichaelDeBoey, line 61:
```suggestion
      "app/routes/button-form-method.jsx": js`
```

integration/bug-report-test.ts
```diff
@@ -29,65 +26,87 @@ let app: AppFixture;
...
+    await app.goto("/working");
+    await app.clickElement("button");
+    expect(await app.getHtml("pre")).toMatch("POST");
+  });
 
-  // If you're not sure what's going on, you can "poke" the app, it'll
-  // automatically open up in your browser for 20 seconds, so be quick!
-  // await app.poke(20);
+  test('should make a "post" request when clicking a submit button with formMethod="post" inside a form with attribute method not set', async () => {
+    await app.goto("/not-working");
```
MichaelDeBoey, line 105:
```suggestion
    await app.goto("/button-form-method");
```

yarn.lock
```diff
@@ -9,14 +9,13 @@
   dependencies:
     "@jridgewell/trace-mapping" "^0.3.0"
 
-"@architect/functions@^4.1.1":
```
MichaelDeBoey, line 12: Please revert these changes, as you didn't update any `package.json`

MichaelDeBoey (changes requested):
I think it would be a good idea to rename the routes, as they could otherwise be confusing when looking at these tests in the future

MichaelDeBoey (approved):

packages/remix-react/components.tsx
```diff
@@ -945,7 +945,6 @@ let FormImpl = React.forwardRef<HTMLFormElement, FormImplProps>(
                 event.preventDefault();
 
                 submit(clickedButtonRef.current || event.currentTarget, {
-                  method,
```
machour, line 948: Shouldn't this be `method: formMethod` instead, which is sanitized above?

mdoury: I don't think so, the `method` props of the `FormImpl` component is the `method` props of the `Form` element. 
Sanitization won't solve the issue here, in both cases it will pass `'get'` to the `submit` function `options`, which will then be used as the default value for the form submission despite the `formmethod` attribute. 

If I understand it well, the `method` in the submit function `options` is not intended to be inferred from the DOM elements and their attributes (this is handled in the definition of the `useSubmitImpl` hook), but rather to overwrite the method in case of imperative submissions (e.g. using the `useSubmit` hook).

MichaelDeBoey:
@mdoury Could you please rebase your branch onto latest `dev` & resolve conflicts?

nrako:
@mdoury – any way I could help? I'd be happy to resolve the conflicts and help push this forward if I'm granted the necessary accesses. I can also start over with a new PR if that's preferable.  Please let me know.

nrako:
I hope this will not offend anyone but I went ahead and opened a new PR #3613 which supersede this one and I've credited @mdoury as a "co-author" on the fix commit.

MichaelDeBoey:
Closing this one in favor of #3613

PR #2172, opened 2022-03-01
chore: Ignore `jest-haste-map` issues in test output

chaance:
re: REM-721


chaance:
Closing, likely to be solved with test overhauls in coming PRs

PR #2176, opened 2022-03-01
feat(create-remix): add `.nvmrc` file to templates

MichaelDeBoey:
People using `nvm` might have their shell setup to [automatically call `nvm use` in a directory with a `.nvmrc` file](https://github.com/nvm-sh/nvm#calling-nvm-use-automatically-in-a-directory-with-a-nvmrc-file).

Having the `.nvmrc` file already in the templates will make sure these people won't see the error implemented by @damiensedgwick in #1586, as they will always have the correct Node version in use.


MichaelDeBoey:
@damiensedgwick I'm sorry you feel to be called out, as that's totally not my intention!

What I meant was "Thanks to @damiensedgwick, users who are using a Node version below 14 are now seeing an error and that's a good thing. This PR creates an even better UX (imo) for people who have `nvm` installed".

So again, sorry if my wording was wrong and you feel bad about it, as I think your contribution was a good one and I just wanted to built on top of it.

damiensedgwick:
> @damiensedgwick I'm sorry you feel to be called out, as that's totally not my intention!
> 
> What I meant was "Thanks to @damiensedgwick, users who are using a Node version below 14 are now seeing an error (the `console.log` followed by the exit of the script) and that's a good thing. This PR creates an even better UX (imo) for people who have `nvm` installed".
> 
> So again, sorry if my wording was wrong and you feel bad about it, as I think your contribution was a good one and I just wanted to built on top of it.

Thank you for taking the time to reply to me @MichaelDeBoey - It was very early this morning when I replied and I could have been on the defensive / sensitive side slightly so my apologies.

I actually use nvm so I am interested to see what this improvement does as I didn't understand it initially from the brief look I took.

Am I right in thinking it will detect an incorrect node version and then install the correct / minimum version required?

3imed-jaberi (commented):
It's would better to see the latest LTS version (`lts/*`)!

MichaelDeBoey:
@3imed-jaberi Since Remix requires Node v14+, I think it's better to set that version as the default

kentcdodds (changes requested):
Hmmm... Not sure I'm into this. Not everyone uses `nvm` (I personally prefer `n`) and also several of these can/should be v16 anyway. I think I'd prefer to have people handle this themselves if needed.

MichaelDeBoey:
I still think it's valuable to add the `.nvmrc` file, as `nvm` is widely used across the community.

We even have an `.nvmrc` file in the root of this repo, just like a lot of other big OSS projects have.
It doesn't hurt to have it in there, but it helps a lot of people imo.

PR #2179, opened 2022-03-02
Update jokes.md

mountainash:
`app/styles` is no longer created by `create-remix` (was removed in 14ab1bf89e800e648b227b2c6fcf701ed2f8194a)

Making newbies look for something that doesn't exist, makes them question if the install worked correctly.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -195,6 +195,7 @@
 - mskoroglu
 - msutkowski
 - mtt87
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (approved):

machour:
Thank you for this PR, there's a better fix here that will get merged soon: #2329

PR #2181, opened 2022-03-02
fix(create-remix): express template docs and move npm-run-all to devDependencies

eastlondoner:
The `start:dev` script no longer exists in the express template

the `npm-run-all` dependency is only required for the `dev` script so can safely be made into a devDependency.


Wellers0n:
cool

kentcdodds (approved):
Thanks!

Issue #2031, opened 2022-02-19
When using pnpm, `remix setup node` writes to another babel dependency


itsMapleLeaf:
### What version of Remix are you using?

v1.2.1

### Steps to Reproduce

Repo: https://github.com/itsMapleLeaf/remix-setup-error

```bash
git clone https://github.com/itsMapleLeaf/remix-setup-error
cd remix-setup-error
pnpm install
cat node_modules/@babel/types/lib/ast-types/generated/index.js
```

### Expected Behavior

It should only write to remix-related packages

### Actual Behavior

It writes remix exports to a file at `node_modules/@babel/types/lib/ast-types/generated/index.js`

For context: some tools like Jest and [GraphQL Code Generator](https://www.graphql-code-generator.com/) have babel packages as direct or transitive dependencies, and this behavior can cause errors in those tools:

![error message shows when running graphql codegen after remix setup](https://user-images.githubusercontent.com/19603573/154781478-1cf5b315-dea5-4ac2-92a1-82b483a5ebf6.png)

My workaround was to write [a script to clear that file](https://github.com/itsMapleLeaf/anigreen/blob/35bfa1747d53fd5c46c9d96ab1ecca9e6f970e95/scripts/post.mjs). I'm not sure what that file is or what it's for, but it's supposed to be empty :shrug: 

I couldn't reproduce this with yarn or npm, only pnpm. Seems like an issue with package symlinking?

convcha:
This issue was reproduced in the [official Remix example of the mui](https://github.com/mui/material-ui/tree/master/examples/remix-with-typescript).

OS: Windows 10
Shell: PowerShell
Node: v17.5.0
pnpm: 6.31.0 (Corepack)
Remix: 1.2.1

**Note**: I could not reproduce this issue at `node:17` Docker container.

**Steps to Reproduce**
Repro: https://github.com/mui/material-ui/tree/master/examples/remix-with-typescript

```
git clone https://github.com/mui/material-ui.git
cd .\material-ui\examples\remix-with-typescript\
pnpm install
pnpm dev
```

Console output after `pnpm install`.
```
PS C:\Users\convcha\data\work\remix-with-typescript> pnpm install
 WARN  deprecated rollup-plugin-inject@3.0.2: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-inject.
Packages: +506
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
[...]

PR #2182, opened 2022-03-02
fix(remix): export a type so that remix/index.d.ts is not empty

eastlondoner:
This shows one way to fairly straightforwardly fix this issue with pnpm https://github.com/remix-run/remix/issues/2031

Closes: #2031 

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/remix/index.ts
+++ b/packages/remix/index.ts
@@ -1,1 +1,8 @@
-throw new Error("Did you forget to run `remix setup` for your platform?");
+// This class exists to prevent https://github.com/remix-run/remix/issues/2031 from occuring
+export class RemixNotSetupError extends Error {
+    constructor() {
+        super("Did you forget to run `remix setup` for your platform?")
+    }
+}
+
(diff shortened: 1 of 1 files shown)


eastlondoner:
This is one way that the linked issue could be solved.

Another possible option is to change `packages/remix/tsconfig.json` to have 
```
    "declaration": false,
    "emitDeclarationOnly": false,
```

changing tsconfig.json in that way would however risk upsetting some IDEs because they would consider it to be "missing type definitions" - however that would be fixed by running `remix setup`.

Either way there should probably be a corresponding test to check that `build/node_modules/remix/index.d.ts` is not an empty file (or is just not present at all if changing tsconfig).

eastlondoner:
great idea. Thanks :-D

kentcdodds:
Thank you for understanding and for your help! We have to prioritize things as a team so we really appreciate contributors like yourself who help us get these sorts of things done even when we have other priorities. ❤️

PR #2185, opened 2022-03-02
chore(CI): update workflow action versions

MichaelDeBoey:
If wanted, I can create a @dependabot file, so PRs are created automatically for these.


kentcdodds (approved):
I'm not a fan of dependabot. I typically only update deps on an as-needed basis. This PR is appreciated as it is! Thank you!

PR #2187, opened 2022-03-02
chore: allow override of rollup output directory for local development

brophdawg11:
For local development usage, allow specification of a different output directory for easier use in testing remix changes on a local app.  This avoids the need to try to symlink the `build/` directory (which always has issues) or manually `cp -r` the built files over.  Unfortunately, `remix dev` doesn't currently pick up on changes automatically, so you just have to Ctrl-C and re-run `remix dev` at the moment.  There's room for improvement there in the future.

```bash
# Tab 1 - create a playground app next to the remix repository
npx create-remix
cd my-remix-app
npm run dev

# Tab 2 - remix repository
REMIX_LOCAL_DEV_OUTPUT_DIRECTORY=../my-remix-app yarn watch
```

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,7 +1,32 @@
+import fs from "fs";
 import path from "path";
 import babel from "@rollup/plugin-babel";
 import nodeResolve from "@rollup/plugin-node-resolve";
 import copy from "rollup-plugin-copy";
+
+let outputDir = "build";
+if (process.env.REMIX_LOCAL_DEV_OUTPUT_DIRECTORY) {
+  const appDir = path.join(
+    process.cwd(),
+    process.env.REMIX_LOCAL_DEV_OUTPUT_DIRECTORY
+  );
+  try {
+    fs.readdirSync(path.join(appDir, "node_modules"));
+  } catch (e) {
+    console.error(
+      "Oops! You pointed REMIX_LOCAL_DEV_OUTPUT_DIRECTORY to a directory that " +
+        "does not have a node_modules/ folder. Please `npm install` in that " +
+        "directory and try again."
+    );
+    process.exit(1);
+  }
+  console.log("Writing rollup output to", appDir);
+  outputDir = appDir;
+}
+
+function getOutputDir(pkg) {
+  return path.join(outputDir, "node_modules", pkg);
+}
 
 function isBareModuleId(id) {
   return !id.startsWith(".") && !path.isAbsolute(id);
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
This seems fine to me

mjackson:
Seems fine. @brophdawg11 do you think this is better as an env var or as a cli arg? Like `yarn build to/this/path`?

brophdawg11:
@mjackson I tried a CLI arg first with something like `yarn watch --remix-output-dir ../path` but `rollup` shoots out one of these warnings per build (it still runs, but with just a ton of verbose output):

```
(!) You have passed an unrecognized option
Unknown CLI flags: remix-output-dir. Allowed options: acorn, acornInjectPlugins, amd, assetFileNames, banner, c, cache, chunkFileNames, compact, config, configPlugin, context, d, dir, dynamicImportFunction, e, entryFileNames, environment, esModule, experimentalCacheExpiry, exports, extend, external, externalLiveBindings, f, failAfterWarnings, file, footer, format, freeze, g, generatedCode, globals, h, hoistTransitiveImports, i, indent, inlineDynamicImports, input, interop, intro, m, makeAbsoluteExternalsRelative, manualChunks, maxParallelFileReads, minifyInternalExports, moduleContext, n, name, namespaceToStringTag, noConflict, o, onwarn, outro, p, paths, perf, plugin, plugins, preferConst, preserveEntrySignatures, preserveModules, preserveModulesRoot, preserveSymlinks, sanitizeFileName, shimMissingExports, silent, sourcemap, sourcemapExcludeSources, sourcemapFile, stdin, strict, strictDeprecations, systemNullSetters, treeshake, v, validate, w, waitForBundleInput, watch
```

It could also be tricky to get the argument into the right spot for `yarn build` since `rollup` runs first then `tsc`, so maybe we would reorder them?  It looks like `rollup` doesn't care if you just pass the directory as you have in your example, where `yarn watch ../path` results in a `process.argv` of:

```
[
  '/Users/matt/lib/n/bin/node',
  '/Users/matt/MyStuff/Work/remix/repos/remix/node_modules/.bin/rollup',
  '-c',
  '-w',
  '../path'
]
```

I think my main concern there would be the implicit nature of "the last thing is the output dir", but we could document in the code I suppose.

Let me know if you prefer the raw CLI arg and reordering of `yarn build` to put `rollup` last and I can swap if so 👍

kentcdodds:
Looks like this has merge conflicts

mjackson:
> I tried a CLI arg first with something like `yarn watch --remix-output-dir ../path` but `rollup` shoots out one of these warnings per build

Is it because you're using `yarn watch` instead of `yarn run watch`? `watch` isn't a yarn command, it's our own custom script in our package.json. yarn supports running our package.json scripts with the `yarn X` style, but I don't think it supports passing additional args to those scripts when we use the shorthand. If I were you I'd try running `yarn run watch --output-dir X` instead.

mjackson:
Oh wait, I misunderstood here. I think an env var will probably be better the way we have things setup now. Sorry for the confusion.

PR #2188, opened 2022-03-02
fix: keep fetcher in loading state during action submission redirect

brophdawg11:
### ~~Dependent on #2194~~

- [x] Docs - [preview](https://github.com/remix-run/remix/blob/matt/fetcher-redirect/DEVELOPMENT.md#transition-manager-flows)
- [x] Tests

### DIFF
--- a/packages/remix-react/transition.ts
+++ b/packages/remix-react/transition.ts
@@ -221,6 +221,12 @@ type FetcherStates<TData = any> = {
     submission: ActionSubmission;
     data: TData;
   };
+  LoadingSubmissionActionRedirect: {
+    state: "loading";
+    type: "actionSubmissionRedirect";
+    submission: ActionSubmission;
+    data: undefined;
+  };
   Loading: {
     state: "loading";
     type: "normalLoad";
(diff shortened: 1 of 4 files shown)


packages/remix-react/transition.ts
```diff
@@ -221,6 +221,12 @@ type FetcherStates<TData = any> = {
     submission: ActionSubmission;
     data: TData;
   };
+  LoadingSubmissionActionRedirect: {
```
ryanflorence, line 224: Let's rename this type to `LoadingActionRedirect`

packages/remix-react/transition.ts
```diff
@@ -221,6 +221,12 @@ type FetcherStates<TData = any> = {
...
+    state: "loading";
+    type: "actionSubmissionRedirect";
```
ryanflorence, line 226: Add call this one `actionRedirect`

So that `fetcher.type` lines up with `transition.type` for the same states.

brophdawg11: 👍  [27e6b23](https://github.com/remix-run/remix/pull/2188/commits/27e6b23fddd3cccd68bb2066b0c412ac75efd3f1)

packages/remix-react/transition.ts
brophdawg11: 👍 [27e6b23](https://github.com/remix-run/remix/pull/2188/commits/27e6b23fddd3cccd68bb2066b0c412ac75efd3f1)

PR #2189, opened 2022-03-02
feat(create-remix): add support for remote repos

mcansh:
this PR adds the ability to pass create-remix a `--template` flag to point to _"anything"_ whether that be a local directory or tarball on disk a `username/repo` a github url or any remote tarball.

I think the only bit i need to (re)implement is the remote tarballs - adding tests for the rest today, but we can go over it all in our 1:1

Closes: #

- [ ] Docs
- [x] Tests


kentcdodds (commented):
Sweet! Did you manually verify that this works well? I want to make sure the output looks like it's supposed to. I don't know if the templates have a `format` script (when we extract them maybe we should add some basic prettier support so they can).

mcansh:
> Sweet! Did you manually verify that this works well? I want to make sure the output looks like it's supposed to. I don't know if the templates have a `format` script (when we extract them maybe we should add some basic prettier support so they can).

yup, I've been running it a few times to make sure paths and everything are right - verifying functionality of the apps now

kentcdodds:
Since we're going to have this option for every stack whether or not they have a format script (even end user stacks) maybe we should reformat things with prettier automatically if they don't have a format script? In that case our basic templates can stay basic.

mcansh:
> Sweet! Did you manually verify that this works well? I want to make sure the output looks like it's supposed to. I don't know if the templates have a `format` script (when we extract them maybe we should add some basic prettier support so they can).

yikes, this gave me require statements :upside\_down\_face:

kentcdodds:
Yeah, yikes 😆 I'm pretty sure we can configure babel differently to only transform the things we want and no more 😅

kentcdodds:
I'm pretty confident we can still make this work. It'll probably just be including a few individual babel plugins rather than the typescript preset.

kentcdodds:
Yup, https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=JYWwDg9gTgLgBDAnmApnA3gFQBbAHYDmAvnAGZQQhwDkUKIwAHtQFAsDGEeAzvAIYAuOHgCuIAEYoocALxwATG1Ii87GMC5kIEABTihvKPgIBKDCzhw6MEVDxxxAOhgQAqmFRQAwn24odZr7CYpJQLERAA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.17.6&externalPlugins=%40babel%2Fplugin-transform-typescript%407.16.8&assumptions=%7B%7D

All we need to include is `@babel/plugin-transform-typescript` and it should remove all the types and nothing else.

mcansh:
> Yup, https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=JYWwDg9gTgLgBDAnmApnA3gFQBbAHYDmAvnAGZQQhwDkUKIwAHtQFAsDGEeAzvAIYAuOHgCuIAEYoocALxwATG1Ii87GMC5kIEABTihvKPgIBKDCzhw6MEVDxxxAOhgQAqmFRQAwn24odZr7CYpJQLERAA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=&prettier=false&targets=&version=7.17.6&externalPlugins=%40babel%2Fplugin-transform-typescript%407.16.8&assumptions=%7B%7D
> 
> 
> 
> All we need to include is `@babel/plugin-transform-typescript` and it should remove all the types and nothing else.

nice! i'll look into this again once i have a few other things ironed all the way out.

mcansh:
both failing test cases work locally when running `node build/node_modules/create-remix/cli.js` 🙃 - I'll dig into it after dinner

mjackson:
Let's get this merged before we explore how to automatically remove TS from the template with Babel. We can do that in a separate PR.

kentcdodds:
Wouldn't this 'create' command mean people would run "npx create-remix create proj". Seems a bit odd.

packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -23,36 +24,86 @@ describe("remix cli", () => {
       let { stdout } = await execFile("node", [remix, "--help"]);
```
mjackson: Let's run it with `--no-color` so we don't get the ANSI escape sequences in the output.

packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -72,4 +123,215 @@ describe("remix cli", () => {
...
+
+    // TODO: enable once this is live
+    it.skip("works for templates in the remix org", async () => {
+      let projectDir = path.join(tmpDir, "template");
+      let { stdout } = await execFile("node", [
+        remix,
+        "create",
+        projectDir,
+        "--template",
+        "arc-stack",
```
mjackson, line 164: FWIW I think this one is going to be called the `grunge-stack`

mjackson (commented):
Love those tests! 😍

packages/remix-dev/__tests__/cli-test.ts
mcansh: 😬 [`a76fd5b` (#2189)](https://github.com/remix-run/remix/pull/2189/commits/a76fd5b78dacdcda3f0f4a0949356915206405f3)

I tried --no-colors, but it threw a fit, let me try again

packages/remix-dev/__tests__/cli-test.ts
mcansh: I'll update :) - it does get redirected though 🤷‍♂️

packages/remix-dev/__tests__/cli-test.ts
mcansh: 🤨 `Command failed: node /Users/loganmcansh/Developer/github.com/remix-run/remix/build/node_modules/@remix-run/dev/cli.js --help --no-colors`

mjackson:
> Wouldn't this create command mean people would run npx create-remix create proj

@kentcdodds You can think of `npx create-remix` as an alias for `npx @remix-run/dev create`. Eventually I'd like to re-purpose the `remix` package to just be the Remix dev tools (i.e. rename `@remix-run/dev` to `remix`), so in the future this will be `npx remix create`. We can keep the `create-remix` script around at that point, but it'll still just be an alias and won't really have any special features.

kentcdodds:
Good deal 👍

kentcdodds:
Sorry about the netlify template merge conflicts. That was my bad merge on another PR.

mcansh:
> Sorry about the netlify template merge conflicts. That was my bad merge on another PR.

no sweat, just had to copy the changes to the js template, `git merge dev` surprisingly followed the code path to new one

mjackson (approved):

ridwansameer:
I noticed this removed some prompts like "where do you want to deploy"

So with this PR if a template is not provided it would automatically use default template. Is this intended?

I feel like the interactive steps were a nicer way to bring a beginner onboard, and most tutorials online are still referencing that flow.

PR #2192, opened 2022-03-03
fix(remix-server-runtime): add `X-Remix-Catch` to immutable headers

rkusa:
I am running Remix on Cloudflare Workers where I am receiving:

```
Can't modify immutable headers.
```

when throwing an error response in a loader. I've tracked the error down to the code setting a `X-Remix-Catch` header. This PR updates the code to create a new `Headers` and `Response` instance instead of trying to modify the existing `headers` instance which seems to be immutable on Cloudflare.

### DIFF
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -43,7 +43,13 @@ export async function callRouteAction({
     }
 
     if (!isRedirectResponse(error)) {
-      error.headers.set("X-Remix-Catch", "yes");
+      let headers = new Headers(error.headers);
+      headers.set("X-Remix-Catch", "yes");
+      return new Response(error.body, {
+        status: error.status,
+        statusText: error.statusText,
+        headers,
+      });
     }
     result = error;
   }
@@ -90,7 +96,13 @@ export async function callRouteLoader({
     }
 
     if (!isRedirectResponse(error)) {
-      error.headers.set("X-Remix-Catch", "yes");
+      let headers = new Headers(error.headers);
+      headers.set("X-Remix-Catch", "yes");
+      return new Response(error.body, {
+        status: error.status,
+        statusText: error.statusText,
+        headers,
+      });
     }
     result = error;
   }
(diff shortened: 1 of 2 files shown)


packages/remix-server-runtime/data.ts
```diff
@@ -45,9 +45,16 @@ export async function callRouteAction({
...
+      headers.set("X-Remix-Catch", "yes");
+      result = new Response(error.body, {
+        headers,
+        status: error.status,
+        statusText: error.statusText,
+      });
+    } else {
+      result = error;
     }
-    result = error;
```
rkusa, line 50: Could probably also be shortened to:

```ts
error = new Response(error.body, error);
error.headers.set("X-Remix-Catch", "yes");
```

Let me know if this variant is preferred.

MichaelDeBoey:
@rkusa Could you please rebase your branch onto latest `dev` & resolve conflicts?

packages/remix-server-runtime/data.ts
rkusa: Early return could also be replaced with the following if preferred:

```diff
 if (!isRedirectResponse(error)) {
   let headers = new Headers(error.headers);
   headers.set("X-Remix-Catch", "yes");
-  return new Response(error.body, {
+  result = new Response(error.body, {
     status: error.status,
     statusText: error.statusText,
     headers,
   });
-}
-result = error;
+} else {
+  result = error;
+}
```

rkusa:
@MichaelDeBoey Done 👍

jmdalmasso:
Any reason this was closed? I'm running into the same error:

```
TypeError: Can't modify immutable headers.
error.headers.set("X-Remix-Catch", "yes");
```

rkusa:
@jmdalmasso I just deleted my fork and forgot that there was an open PR attached to it. I still encounter this error on a regular basis, too, but work around it by not returning responses in a loader directly, instead I return `json(await res.json())`. Not great, but works around the issue.

So basically:

```diff
export const loader: LoaderFunction = async ({ context, params }) => {
  // ...
- return res;
+ return json(await res.json());
};
```

PR #2194, opened 2022-03-03
chore: add local debugging logs for transition manager

brophdawg11:
This adds a bunch of helpful `console.debug` statements to the transition manager to help debug the heavily async nature of it.  To avoid cluttering the console in normal application usage, I added https://babeljs.io/docs/en/babel-plugin-transform-remove-console/ to strip all `console.debug` statements during the build.  For local development, setting `REMIX_DEBUG=true` will leave them in for debugging.

Screen Shot 2022-03-03 at 9 31 53 AM

- [x] Docs - [preview](https://github.com/remix-run/remix/blob/matt/transition-debug/DEVELOPMENT.md#local-development-tips-and-tricks)
- [x] Tests - n/a

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -114,7 +114,7 @@ REMIX_DEBUG=true yarn watch
 
 **`REMIX_LOCAL_DEV_OUTPUT_DIRECTORY`**
 
-When developing Remix locally, you often need to go beyond unit/integration tests and test your changes in a local Remix application. The easiest way to do this is to run your local Remix build and use this environment variable to direct `rollup` to write the output files directly into the local remix application's `node_modules` folder. Then you just need to restart your local Remix application serve to pick up the changes.
+When developing Remix locally, you often need to go beyond unit/integration tests and test your changes in a local Remix application. The easiest way to do this is to run your local Remix build and use this environment variable to direct `rollup` to write the output files directly into the local Remix application's `node_modules` folder. Then you just need to restart your local Remix application server to pick up the changes.
 
 ```sh
 # Tab 1 - create an run a local remix application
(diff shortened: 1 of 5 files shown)


DEVELOPMENT.md
```diff
@@ -97,3 +97,33 @@ git commit -m "fix: squashed a super gnarly bug"
...
+
+**`REMIX_DEBUG`**
+
+By default, the Remix `rollup` build will strip any `console.debug` calls to avoid cluttering up the console during application usage. These `console.debug` statements can be preserved by setting `REMIX_DEBUG=true` during your local build.
+
+```sh
+REMIX_DEBUG=true yarn watch
+```
+
+**`REMIX_LOCAL_DEV_OUTPUT_DIRECTORY`**
```
brophdawg11, line 115: This is introduced in https://github.com/remix-run/remix/pull/2187 but documented here to avoid conflicts

babel.config.js
```diff
@@ -14,5 +14,14 @@ module.exports = {
   plugins: [
     "@babel/plugin-proposal-export-namespace-from",
     "@babel/plugin-proposal-optional-chaining",
+    // Strip console.debug calls unless REMIX_DEBUG=true
+    ...(process.env.REMIX_DEBUG === "true"
+      ? []
+      : [
+          [
+            "transform-remove-console",
+            { exclude: ["error", "warn", "log", "info"] },
```
brophdawg11, line 23: Only strip `console.debug` by default

packages/remix-react/transition.ts
```diff
@@ -1215,10 +1282,18 @@ export function createTransitionManager(init: TransitionManagerInit) {
...
 
   function abortNormalNavigation() {
-    pendingNavigationController?.abort();
+    if (pendingNavigationController) {
+      console.debug(`[transition] aborting pending navigation`);
+      pendingNavigationController.abort();
+      // TODO: Should this be cleared now so it doesn't stick around for
+      // subsequent navigations and make logs look weird?
+      // Or just remove this log?
+      // pendingNavigationController = undefined;
```
brophdawg11, line 1291: @ryanflorence What do you think here?  I noticed that after our _first_ navigation this is non-null forever, so we seemed to always see this `aborting pending navigation` log - even for non-interruption-navigations.  I'm unsure if this is exactly the right place to reset though - or if it should be reset elsewhere?  It's not breaking anything really since calling abort subsequently shouldn't really do anything

ryanflorence: should probably be cleaned up after the navigation completes

brophdawg11: 👍 Added in e475f236bf88070e204434d79452361199704230

PR #2200, opened 2022-03-04
docs(examples): bring in line with latest templates

MichaelDeBoey:
As discussed with @kentcdodds on Discord, I've made one big PR to update all examples at once:

- Bring examples in line with latest version of templates
- Add `sandbox.config.json` to each example
- Update dependencies to latest minor

Tailwind CSS example is already updated in #2144, so that one's not included in this PR.


examples/jokes/.env.example
```diff
@@ -0,0 +1,2 @@
+SESSION_SECRET="remixrulz"
```
MichaelDeBoey, line 1: Renamed this one to `.env.example` to be consistent with all other examples.

kentcdodds (approved):
Holy smokes 🤯 Thanks for this!

PR #2203, opened 2022-03-04
Add the version supporting env vars

hariombalhara:
Motivation: https://discordapp.com/channels/770287896669978684/949190278080770118

- [x] Docs


machour:
Thank you for taking the time to PR this, but I wouldn't want us to create a precedent where we mention what version changed what behavior in the documentation, it would be really hard to maintain 🙏🏼 

Releases Changelogs are the way to document this kind of little details.

PR #2212, opened 2022-03-04
docs(examples): simplify ESLint config for sanity example

MichaelDeBoey:
@kentcdodds Simplify what you did in cabcb02


kentcdodds (approved):
Sure 👍

PR #2213, opened 2022-03-04
fix(templates/arc): update `README` + dependencies

ryanblock:
Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/create-remix/templates/arc/README.md
+++ b/packages/create-remix/templates/arc/README.md
@@ -2,37 +2,40 @@
 
 - [Remix Docs](https://remix.run/docs)
 
-## Architect Setup
-
-When deploying to AWS Lambda with Architect, you'll need:
-
-- Architect (`arc`) CLI
-- AWS SDK
-
-Architect recommends installing these globally:
-
-```sh
-$ npm i -g @architect/architect aws-sdk
-```
 
 ## Development
 
-You will be running two processes during development when using Architect as your server.
+Both Remix and Architect have their own local servers, each of which is responsible for unique tasks related to local development.
 
-- Your Architect server sandbox in one
-- The Remix development server in another
+To start t both alongside one another, run:
 
 ```sh
-# in one tab
-$ arc sandbox
-
-# in another
 $ npm run dev
 ```
 
 Open up [http://localhost:3333](http://localhost:3333) and you should be ready to go!
 
 If you'd rather run everything in a single tab, you can look at [concurrently](https://npm.im/concurrently) or similar tools to run both processes in one tab.
+
+
+### TypeScript
+
+To develop in TypeScript, install Architect's official TS plugin:
+
+```sh
+$ npm i @architect/plugin-typescript
+```
+
+Then add the following to the `app.arc` file:
+
+```arc
+@plugins
+architect/plugin-typescript
+
+@aws
+runtime typescript
+```
+
 
 ## Deploying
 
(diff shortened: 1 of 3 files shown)


packages/create-remix/templates/arc/package.json
```diff
@@ -8,6 +8,7 @@
     "start": "cross-env NODE_ENV=production arc sandbox"
```
ryanblock, line 8: @mcansh: `NODE_ENV=development` is not required for any Architect processes, so in theory we can remove that from `start` + `dev:arc` if that doesn't cause any trouble for y'all!

packages/create-remix/templates/arc/package.json
```diff
@@ -8,6 +8,7 @@
...
   },
   "dependencies": {
+    "@architect/architect": "^10.0.5",
```
ryanblock, line 11: How can we help ensure this is relatively up to date?

packages/create-remix/templates/arc/README.md
```diff
@@ -2,37 +2,40 @@
...
-$ npm i -g @architect/architect aws-sdk
+$ npm run dev
 ```
 
-## Development
+Open up [http://localhost:3333](http://localhost:3333) and you should be ready to go!
+
 
-You will be running two processes during development when using Architect as your server.
+### TypeScript
```
ryanblock, line 19: Does all this make sense in the context of Remix? I know a lot of Remix + Arc folks have been trying to use TS, but have been a bit confused as to where to start, so I was thinking this might help shortcut them. Up to y'all!

packages/create-remix/templates/arc/package.json
mcansh: not 100% sure when we added that to the templates, but i'm 95% sure we set NODE_ENV during `remix dev` and `remix build` (and `remix-serve`)

packages/create-remix/templates/arc/package.json
mcansh: couldn't we use `^10.x` to signify any version of `v10`? we don't include lock files in the templates so once they're installed they _should_ use the latest version available, no?

mcansh: otherwise, we can keep it `^10.0.5` and wait for PRs to bump the version

packages/create-remix/templates/arc/package.json
ryanblock: Ok, so do you want to leave it? Remove it? Arc is good either way, but imo always good to have less stuff if possible.

packages/create-remix/templates/arc/package.json
ryanblock: Yeah we can def safely set to 10.x

MichaelDeBoey: `^10.0.0` should indeed just work

packages/create-remix/templates/arc/README.md
mcansh: unfortunately using `@architect/plugin-typescript` for all lambdas causes a "lambda not found error" for the remix app, perhaps we can mention the config.arc approach instead?

packages/create-remix/templates/arc/package.json
mcansh: let's remove it

packages/create-remix/templates/arc/README.md
mcansh: lol wut, today it works

mcansh: i think instead of mentioning the installation of the typescript plugin directly, we can add a `arc-ts` directory and add the dependency and full app.arc there

mcansh:
@ryanblock can you make these adjustments?

MichaelDeBoey:
@ryanblock Also make sure to rebase your branch onto latest `main`, as we've moved some things around in the meantime.

MichaelDeBoey:
@ryanblock Could you please rebase your branch onto latest `main` & resolve conflicts?

Issue #1633, opened 2022-01-24
[Feature]: allow to define custom JSX factories


jraoult:
### What is the new or updated feature that you are suggesting?

To be able to support [`@emtion/react`'s `css` prop](https://emotion.sh/docs/css-prop) on React elements, _esbuild's_ config needs to be[ slightly tweaked](https://github.com/evanw/esbuild/issues/832#issuecomment-951498187.).

I managed to patch Remix's https://github.com/remix-run/remix/blob/main/packages/remix-dev/compiler.ts and https://github.com/remix-run/remix/blob/main/packages/remix-dev/compiler/shims/react.ts to make it work in dev but ideally it we would be able to configure `esbuild` from `remix.config.js` (or similar).

It could be a jsx factory specific thing:

```js
module.exports = {
  [...],
  jsxFactory: "jsx",
  reactShim: "path/to/shim"
}
```

 or just allowing the whole `esbuild` configuration to be extended:
```js
module.exports = {
  [...],
  esbuildConf: {
    jsxFactory: "jsx",
    inject: ["path/to/shim"]
  }
}
```

### Why should this feature be included?

To be able to support custom JSX factories like emotion's one.

jraoult:
Here's the patch I created:

```
diff --git a/compiler/shims/react.ts b/compiler/shims/react.ts
index eb0f102ecb7ea96274f00ad964393c86981dbdf9..a65ef8521b1dedf754adc1013f5c32799b6d5136 100644
--- a/compiler/shims/react.ts
+++ b/compiler/shims/react.ts
@@ -1,2 +1,4 @@
+import { jsx } from "@emotion/react";
 import * as React from "react";
-export { React };
+
+export { jsx, React };
diff --git a/compiler.js b/compiler.js
index 1a6cabc2ae7102055cba88cb57084ba7ec8642b4..3cf652f33445a4b198b5f10b2915300654181e88 100644
--- a/compiler.js
+++ b/compiler.js
@@ -297,6 +297,7 @@ async function createBrowserBuild(config, options) {
     platform: "browser",
     format: "esm",
     external: externals,
+    jsxFactory: "jsx",
     inject: [reactShim],
     loader: loaders.loaders,
     bundle: true,
@@ -329,6 +330,7 @@ async function createServerBuild(config, options) {
     format: config.serverModuleFormat,
     mainFields: config.serverModuleFormat === "esm" ? ["module", "main"] : ["main", "module"],
     target: options.target,
[...]

PR #2216, opened 2022-03-05
feat(remix-dev): add `jsxImportSource` config option

jolle:
Libraries such as emotion (via `@emotion/react`) are commonly used with the tsconfig option `jsxImportSource`, which changes the default `React.createElement` calls to the specified source automatically. Because esbuild does not support this out-of-the-box (see https://github.com/evanw/esbuild/issues/718), by using a dynamically created shim file it is possible to emulate this behavior. Although an inline pragma works, it is required to add this to every file instead of being automatically detected.

As esbuild does not support inline code injection (see https://github.com/evanw/esbuild/issues/1169), this PR creates a temporary directory with a temporary file for the custom shim. While this is not very optimal, it is the best option I've found (didn't seem like esbuild plugins could do this?). In this PR, the tsconfig.json file is read (if one exists), and a shim is created based on `jsxImportSource`, if one is specified. A new config option, `customJSXShimPath`, is exposed as well if a custom path to the JSX factory is required for the shim.

This pull request also adds tests for this new behavior, and refactors the tests of `build-test.ts` that were being skipped (because they didn't work). Due to ESM-related issues and conflicts with the Jest environment, a new, simpler app needed to be created without any ESM-only modules, and some modules needed to be mocked during build-time.  The JSX factory tests run the built output within the test in a "sandbox" where the calls to the JSX factory can be monitored.

This is a quite big new feature – apologies for not creating a discussion first!

Closes: #1633. Related to: #2209.


machour:
@jolle Can you please update this PR to fix conflicts? 🙏🏼

jolle:
Merge conflicts have been fixed. I noticed there has been changes to the testing env that causes the new tests to fail (related to fixtures trying to import the remix package). I'll try to fix the tests soon, though the solution isn't immediately obvious to me – I'll have to go through the changes and find what has broken the fixtures!

MichaelDeBoey:
@jolle Please rebase your branch onto latest `dev` & fix conflicts + tests.
That way the team has a clean plate to starte reviewing if they want to have this merged into the library or not, thanks! 🙏

MichaelDeBoey:
@jolle Please rebase your branch onto latest `dev` & fix conflicts + tests.
That way the team has a clean plate to start reviewing if they want to have this merged into the library or not, thanks! 🙏

packages/remix-dev/__tests__/build-test.ts
```diff
@@ -1,24 +1,102 @@
...
+jest.mock("tsconfig-paths", () => ({
+  __esModule: true,
+  default: {
+    loadConfig() {
+      return {
+        resultType: "failed",
+      };
+    },
+  },
+}));
```
jolle, line 28: This is not optimal but regardless of what I did to Babel or Jest these modules seemed to break Jest one way or another.

jolle:
I've resolved the merge conflicts and fixed the tests. In addition to the caveat above, I downgraded `strip-json-comments` as v4 was released only to make it an ESM (which is breaking the Jest env; v3 does not have this issue).

jolle:
I have removed the temporary file system injection for adding the custom JSX factory; instead, the import source is loaded by adding a `@jsx` pragma header to every file (which is supported by esbuild) using a plugin.

luishdz1010:
This now works natively on esbuild, adding:

```
"overrides": {
    "@remix-run/dev": {
      "esbuild": "0.15.2"
    }
  }
```

To package.json (and using npm) works without any configs.

jolle:
Support for automatic `jsxImportSource` detection in esbuild was added in v0.14.51 (https://github.com/evanw/esbuild/pull/2349), so this is no longer necessary.

PR #2220, opened 2022-03-06
docs(examples): add supabase realtime subscription example 

marvinwu:
- [ x ] Docs

This is an example to implement a simple like button using supabase realtime subscription to keep the number of likes in sync across all clients


kentcdodds (approved):
Nice! This looks good to me. Thanks 👍

PR #2224, opened 2022-03-06
docs(guides/constraints): link to Reach UI's `useIsomorphicLayoutEffect`

Aprillion:
Best I could do right now since the utils are not documented on https://reach.tech/ yet.

But feel free to close this PR if you decide it will be better to keep the TODO until remix supports
`import {useIsometricLayoutEffect} from 'remix'` 😅

### DIFF
--- a/docs/guides/constraints.md
+++ b/docs/guides/constraints.md
@@ -362,7 +362,20 @@ The point is to perform the effect at the same time as the browser paint so that
 
 It is **not** good for setting state that is rendered inside of elements. Just make sure you aren't using the state set in a `useLayoutEffect` in your elements and you can ignore React's warning.
 
-TODO: Link to Reach UI `useIsomorphicLayoutEffect`
+If you know you're calling `useLayoutEffect` correctly and just want to silence the warning, a popular solution in libraries is to create your own hook that doesn't call anything on the server. `useLayoutEffect` only runs in the browser anyway, so this should do the trick. **Please use this carefully, because the warning is there for a good reason!**
+```js
+import * as React from "react";
+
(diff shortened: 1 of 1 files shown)


docs/guides/constraints.md
```diff
@@ -354,7 +354,10 @@ The point is to perform the effect at the same time as the browser paint so that
 
 It is **not** good for setting state that is rendered inside of elements. Just make sure you aren't using the state set in a `useLayoutEffect` in your elements and you can ignore React's warning.
 
-TODO: Link to Reach UI `useIsomorphicLayoutEffect`
+See also Reach UI's [useIsomorphicLayoutEffect](https://github.com/reach/reach-ui/blob/main/packages/utils/src/use-isomorphic-layout-effect.ts), installed by `npm i @reach/utils`:
```
chaance, line 357:
```suggestion
If you know you're calling `useLayoutEffect` correctly and just want to silence the warning, a popular solution in libraries is to create your own hook that doesn't call anything on the server. `useLayoutEffect` only runs in the browser anyway, so this should do the trick. **Please use this carefully, because the warning is there for a good reason!**
```

chaance (changes requested):
Thanks for this! I'd much prefer if we inline this code instead of linking out to Reach since that's an internal util. It would also help to explain what we're doing and why. Made some suggested changes for you.

docs/guides/constraints.md
Aprillion: Thanks for review, now rebased to include your suggestions.

chaance (approved):

PR #2226, opened 2022-03-06
chore: cleanup `.eslintrc` & fix ESLint warnings

MichaelDeBoey:
This makes the `.eslintrc` file clearer + removes unnecessary rules (as they were already fixed)

---

This PR is for `main` branch


MichaelDeBoey:

PR #2228, opened 2022-03-06
chore: enable `prefer-const` rule for all code blocks in `.md` files

MichaelDeBoey:
Because we want all public facing code to use `const`.

---

This PR is for `main` branch

This PR is dependent on #2226


MichaelDeBoey:

PR #2229, opened 2022-03-06
chore: enable `prefer-const` rule for all code blocks in `.md` files

MichaelDeBoey:
Just as in #2228, because we want all public facing code to use `const`.

---

This PR is for `dev` branch

This PR is dependent on #2227


MichaelDeBoey:

PR #2242, opened 2022-03-07
chore(scripts/publish-private): publish eslint config to remix registry

mcansh:
This PR adds the `@remix-run/eslint-config` package to `scripts/publish-private.js`, some ~users~ supporters still have a `.npmrc` file pointed to our registry which causes them to not be able to install dependencies.

https://canary.discord.com/channels/770287896669978684/771068344320786452/950398837460189234

Closes: #

- [ ] Docs
- [ ] Tests


mcansh:
yeah definitely not necessary, and is indeed a good way to _make_ them upgrade to use the npm registry for the day when we officially deprecate and stop publishing to the private registry

PR #2243, opened 2022-03-07
Run Remix on Deno

pcattori:
I removed `remix-deno` package (see commit message for details). Idea is to take the `remix-deno` package currently inlined in our new Deno template (see https://github.com/pcattori/remix-deno-template for a prototype of that) and make that importable in Deno. Note that this would mean treating `remix-deno` as a Deno package, not a node package.

I don't think this should be considered a breaking change since (1) Deno support is experimental (2) previous Deno template was immediately erroring out.

Closes: #

- [ ] Docs
- [ ] Tests


templates/deno/package.json
```diff
@@ -1,32 +1,20 @@
...
+    "sideEffects": false,
+    "scripts": {
+      "build": "cross-env NODE_ENV=production remix build",
+      "start": "cross-env NODE_ENV=production deno run --allow-net --allow-read --allow-env ./build/index.js",
+      "dev": "cross-env NODE_ENV=development remix build && run-p dev:*",
+      "dev:remix": "cross-env NODE_ENV=development remix watch",
+      "dev:deno": "cross-env NODE_ENV=development deno run --watch --allow-net --allow-read --allow-env ./build/index.js"
+    },
+    "devDependencies": {
+      "@remix-run/dev": "^1.2.2",
```
pcattori, line 12: TODO: We need to update this to be the version for the release of `@remix-run/dev` that includes these changes. @chaance do we know if this will be 1.2.3 or 1.3.0?

packages/remix-dev/tsconfig.json
```diff
@@ -13,6 +13,7 @@
 
     "declaration": true,
     "emitDeclarationOnly": true,
+    "skipLibCheck": true,
```
pcattori, line 16: @jacob-ebey : what was the reason for enabling `skipLibCheck`? Is it because `esbuild-plugin-cache` pulls in `deno-cache` and `deno-importmap` as dependencies and type checking fails for those?

templates/deno/package.json
pcattori: I guess we can say `^1.2.4` and it will work even if we go straight from `1.2.3` to `1.30`. Feels a little weird to reference a non-existing version but 🤷 it would work.

mjackson: I think this should be `*` for now. The install script finds all remix deps in package.json and updates them with the right version when it uses the template.

templates/deno/app/routes/index.tsx
```diff
@@ -1,9 +1,13 @@
-import * as React from "react";
+import React from "../deps/react.ts";
 
 export default function Index() {
+  const [state, setState] = React.useState("first");
```
mjackson, line 4: Is this some debugging code that should be removed?

templates/deno/app/entry.server.tsx
```diff
@@ -1,15 +1,15 @@
-import * as React from "react";
-import { renderToString } from "react-dom/server";
-import { RemixServer } from "remix";
-import type { EntryContext } from "remix";
+import  React from "./deps/react.ts";
+import  ReactDOMServer from "./deps/react-dom-server.ts";
```
mjackson, line 2: Looks like you have an extra space on these two lines after the `import` keyword

templates/deno/app/root.tsx
```diff
@@ -1,13 +1,13 @@
...
   Links,
   LiveReload,
   Meta,
   Outlet,
   Scripts,
   ScrollRestoration,
-} from "remix";
-import type { MetaFunction } from "remix";
+} from "./deps/@remix-run/react.ts";
+import type { MetaFunction } from "./deps/@remix-run/server-runtime.ts";
```
mjackson, line 10: This should import from `@remix-run/deno`. remix-deno should re-export anything it needs from remix-server-runtime. remix-server-runtime is just a common dep of remix-node, remix-cloudflare (future package) and remix-deno.

mjackson (changes requested):
Looking good, @pcattori! Almost there.

templates/deno/package.json
MichaelDeBoey: @mcansh changed this to actual values in #2189

templates/deno/app/root.tsx
pcattori: I've got a separate task to make the `@remix-run/*` packages present a complete server runtime interface, so I can address this for `remix-deno` then.

mjackson (approved):

templates/deno-ts/deno.json
```diff
@@ -0,0 +1,6 @@
+{
+  "compilerOptions": {
+    "lib": ["dom", "dom.iterable", "dom.asynciterable", "deno.ns"]
```
F3n67u, line 3: Hi @pcattori could you explain why we should customize lib? this line cause the error reported in https://github.com/remix-run/remix/issues/2402

pcattori: I copied it from some of the prior work we had done on Deno, but it might be a vestige from when we were trying to use standard Node tooling + Typescript compiler to produce a Deno bundle. We've since abandoned that approach in favor of writing Deno source code (eventually with Deno tooling like `deno fmt`, `deno lint`, etc... in CI).

Could very well be the case that we should remove the `lib` option.

pcattori: Let's continue the discussion in #2402

PR #2247, opened 2022-03-07
chore: Add internal shareable ESLint config

chaance:
This PR introduces an internal ESLint config we can share across our org's packages. This config is not meant for public use, nor is it considered public API.

It also adds a new rule per `REM-784` and fixes issues in the Deno templates that would be flagged by the updated rules. Deno.

All other changes are a result of running `eslint --fix` after updating the config.


MichaelDeBoey (changes requested):
Awesome work to make the internal ESLint rules a bit more defined. 💪👊

I was also already doing this in #2226, #2227 #2228 & #2229, so I added the changes I did there as a remark on this PR, together with some other remarks/questions I had.

.eslintrc.js
```diff
@@ -35,36 +27,19 @@ module.exports = {
...
-    },
-    {
-      files: ["examples/**/*.ts", "examples/**/*.tsx"],
+      files: [
+        "packages/remix-server-runtime/**/*.js?(x)",
+        "packages/remix-server-runtime/**/*.ts?(x)",
+      ],
       rules: {
-        "@typescript-eslint/no-unused-vars": "off",
+        "node/no-unsupported-features/node-builtins": [
```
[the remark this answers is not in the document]
mjackson: Yeah, we should definitely eliminate all node-specific code in remix-server-runtime. But not in remix-node, obviously. Is that not what this does?

docs/api/remix.md
```diff
@@ -98,10 +98,10 @@ import { NavLink } from "remix";
 function NavList() {
   // This styling will be applied to a <NavLink> when the
   // route that it links to is currently selected.
-  let activeStyle = {
+  const activeStyle = {
```
mjackson, line 101: This should be a `let`. Let's remove `prefer-const` from the docs dir and just use one consistent style throughout the whole repo.

mjackson (changes requested):
Thanks for taking this on, @chaance. I'm going to reneg on what I said last week and suggest we use `let` for anything that is an "intermediate" value. In other words, module-scope variables should probably be `const`. Everything else should be `let`, including in our docs.

.eslintrc.js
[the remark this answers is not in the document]
MichaelDeBoey: This rule is targeting the following files:

```ts
      files: [
        "packages/remix-server-runtime/**/*.js?(x)",
        "packages/remix-server-runtime/**/*.ts?(x)",
      ],
```

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,103 @@
...
+
+const OFF = 0;
+const WARN = 1;
+const ERROR = 2;
+
+module.exports = {
+  root: true,
+  extends: [
+    require.resolve("./index.js"),
+    require.resolve("./jest-testing-library.js"),
```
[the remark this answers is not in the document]
chaance: I'm keeping the markdown rules/plugin out of the shared config for now because we'll be using this in outside Remix repos that won't use them. Might move them in later.

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,103 @@
...
+        "no-unreachable": OFF,
+        "no-unused-expressions": OFF,
+        "no-unused-labels": OFF,
+        "no-unused-vars": OFF,
+        "jsx-a11y/alt-text": OFF,
+        "jsx-a11y/anchor-has-content": OFF,
+        "prefer-let/prefer-let": OFF,
+        "react/jsx-no-comment-textnodes": OFF,
+        "react/jsx-no-undef": OFF,
+      },
```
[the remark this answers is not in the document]
chaance: For now I think we'll leave this as-is. Reason is that we'll use this config outside of the Remix repo (React Router repo, for example) that may include JSX/React in non-jsx file extensions.

.eslintrc.js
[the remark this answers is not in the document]
chaance: I updated this and just added an override config in the `remix-server-runtime` project for now. I also added this override to the Deno adapter.

docs/api/remix.md
chaance: This should be reverted. Reminder that we'll handle all the stuff in the docs ahead of the next release when we merge everything from `main`.

packages/remix-eslint-config/internal.js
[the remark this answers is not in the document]
MichaelDeBoey: This is only about MD files though, not about the actual code

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+        "@typescript-eslint/no-unused-vars": OFF,
+      },
+    },
+    {
+      files: [
+        // All examples and docs, including code blocks in .md files
+        "examples/**/*.js?(x)",
+        "examples/**/*.ts?(x)",
+        "docs/**/*.js?(x)",
+        "docs/**/*.ts?(x)",
```
MichaelDeBoey, line 77: I don't think adding rules to the docs folder would help, as they only contain MD files, which are already covered above
```suggestion
```

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+        "no-unused-labels": OFF,
+        "no-unused-vars": OFF,
+        "prefer-let/prefer-let": OFF,
+      },
+    },
+    {
+      files: [
+        // All examples and docs, including code blocks in .md files (TypeScript only)
+        "examples/**/*.ts?(x)",
+        "docs/**/*.ts?(x)",
```
MichaelDeBoey, line 91: > I don't think adding rules to the docs folder would help, as they only contain MD files, which are already covered above

```suggestion
```

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+        "examples/**/*.js?(x)",
+        "examples/**/*.ts?(x)",
+        "docs/**/*.js?(x)",
+        "docs/**/*.ts?(x)",
+      ],
+      rules: {
+        "import/order": OFF,
+        "no-unused-expressions": OFF,
+        "no-unused-labels": OFF,
+        "no-unused-vars": OFF,
```
MichaelDeBoey, line 83: Do we really want to enable unused expressions/labels/vars in examples?  
These will only clutter the examples unnecessary imo.

```suggestion
```

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+    },
+    {
+      files: [
+        // All examples and docs, including code blocks in .md files (TypeScript only)
+        "examples/**/*.ts?(x)",
+        "docs/**/*.ts?(x)",
+      ],
+      rules: {
+        "@typescript-eslint/no-unused-expressions": OFF,
+        "@typescript-eslint/no-unused-vars": OFF,
```
MichaelDeBoey, line 95: > Do we really want to enable unused expressions/labels/vars in examples?  
> These will only clutter the examples unnecessary imo.

```suggestion
```

MichaelDeBoey (changes requested):

packages/create-remix/templates/.eslintrc.js
```diff
@@ -1,8 +1,7 @@
 module.exports = {
   rules: {
-    "prefer-const": "warn",
-
     // because in our templates we shouldn't care
```
mjackson, line 3: Why not?

package.json
```diff
@@ -13,7 +13,7 @@
     "test": "jest",
     "version": "node ./scripts/version.js",
     "watch": "rollup -c -w",
-    "lint": "eslint --cache --ext .tsx,.ts,.js,.jsx,.md . --fix",
+    "lint": "eslint --cache --ext .tsx,.ts,.js,.jsx,.md .",
```
mjackson, line 16: Why did you remove the `--fix`? Is this the part you were telling me about earlier where you said it won't be active yet?

packages/remix-deno/globals.ts
```diff
@@ -10,8 +10,10 @@ declare global {
 
   // This is here because we need it in our node setup,
   // but not in an actual deno project.
+  /* eslint-disable prefer-let/prefer-let */
```
mjackson, line 13: Seems like we should open an issue on the `prefer-let` code to explain this use case to them and ask them for the ability to opt-out in global TS declarations?

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,91 @@
...
+      },
+    },
+    {
+      files: [
+        // All examples and docs, including code blocks in .md files (TypeScript only)
+        "examples/**/*.ts?(x)",
+      ],
+      rules: {
+      },
+    },
```
mjackson, line 89:
```suggestion
```

mjackson (changes requested):
Looks pretty good, but I was confused about why this PR creates files in the deno template. Shouldn't those be created in a separate PR?

package.json
chaance: Ah, oversight. I removed it temporarily for dev purposes. Adding back.

packages/create-remix/templates/.eslintrc.js
MichaelDeBoey: It's user-facing code, so I think we _should_ care tbh.

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+      rules: {
+        "@typescript-eslint/no-unused-expressions": OFF,
+        "@typescript-eslint/no-unused-vars": OFF,
+      },
+    },
+    {
+      files: [
+        // All examples and docs, including code blocks in .md files
+        "docs/**/*.js?(x)",
+        "docs/**/*.ts?(x)",
```
MichaelDeBoey, line 75: ff9db8210390b0067510a9c428d49a2b0cfe5f08

> I don't think adding rules to the docs folder would help, as they only contain MD files, which are already covered above
> 
> ```suggestion
> ```

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+        "no-unused-expressions": OFF,
+        "no-unused-labels": OFF,
+        "no-unused-vars": OFF,
+        "prefer-let/prefer-let": OFF,
+      },
+    },
+    {
+      files: [
+        // All TS examples and docs, including code blocks in .md files
+        "docs/**/*.ts?(x)",
```
MichaelDeBoey, line 90: ff9db8210390b0067510a9c428d49a2b0cfe5f08

> > I don't think adding rules to the docs folder would help, as they only contain MD files, which are already covered above
> 
> ```suggestion
> ```

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+        "examples/**/*.ts?(x)",
+      ],
+      rules: {
+        "@typescript-eslint/no-unused-expressions": OFF,
+        "@typescript-eslint/no-unused-vars": OFF,
```
MichaelDeBoey, line 95: ff9db8210390b0067510a9c428d49a2b0cfe5f08

> > Do we really want to enable unused expressions/labels/vars in examples?  
> > These will only clutter the examples unnecessary imo.
> 
> ```suggestion
> ```

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+      },
+    },
```
MichaelDeBoey, line 97: d6fbe5a3f2fbfa63d75b80c52232ea38c8600e22

```suggestion
```

MichaelDeBoey (changes requested):
@chaance It seems like you accidentally reverted ff9db8210390b0067510a9c428d49a2b0cfe5f08 & d6fbe5a3f2fbfa63d75b80c52232ea38c8600e22 when merging branches

packages/remix-eslint-config/internal.js
```diff
@@ -0,0 +1,99 @@
...
+        "docs/**/*.js?(x)",
+        "docs/**/*.ts?(x)",
+        "examples/**/*.js?(x)",
+        "examples/**/*.ts?(x)",
+      ],
+      rules: {
+        "import/order": OFF,
+        "no-unused-expressions": OFF,
+        "no-unused-labels": OFF,
+        "no-unused-vars": OFF,
```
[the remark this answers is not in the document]
chaance: I think so. Examples may show how an object is initialized or where it's imported from without necessarily using every single reference. We'll just use our best judgment here on a case-by-case basis.

docs/api/remix.md
MichaelDeBoey: @mjackson Does this mean we don't want any user-facing code (so templates + docs) to use `const` anymore (like suggested/introduced by @kentcdodds a couple of weeks/months(?) ago) and _always_ use `let`?

package.json
```diff
@@ -76,6 +76,7 @@
     "eslint-plugin-jsx-a11y": "^6.5.1",
     "eslint-plugin-markdown": "^2.2.1",
     "eslint-plugin-node": "^11.1.0",
```
MichaelDeBoey, line 78: Shouldn't this be a dependency of `@remix-run/eslint-config` as we're now having extending `@remix-run/eslint-config/node` in the configs?

packages/remix-eslint-config/node.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @see https://github.com/eslint/eslint/issues/3458
+ * @see https://www.npmjs.com/package/@rushstack/eslint-patch
+ */
+require("@rushstack/eslint-patch/modern-module-resolution");
+
+module.exports = {
+  plugins: ["node"],
```
MichaelDeBoey, line 8: I think it doesn't make sense to have `eslint-plugin-node` in the `plugins` array if we're enabling any rules.

So I think we should either enable some basic rules **or** we should remove it from the `plugins` array

MichaelDeBoey (changes requested):

package.json
chaance: It already is. https://github.com/remix-run/remix/blob/main/packages/remix-eslint-config/package.json#L28

mjackson (commented):
This all looks pretty good to me but I'd like to hold off on merging until we can get Logan and Pedro's stuff merged from this cycle. Hopefully we can merge this on Monday after re-running it over the entire repo.

PR #2250, opened 2022-03-08
chore(docs): Explicitly return `Response` in loader/action examples

chaance:
This PR updates our docs and some examples to explicitly return a response for all loaders/actions. (internal team: see REM-779 in Linear for context)

PR #2252, opened 2022-03-08
deprecated: Warn if returning non-`Response` from loader/action

chaance:
This PR will result in warnings for users if they return a value other than a `Response` from either a loader or action function. This will deprecate the behavior and set us up to remove support for it in a later release.

Docs and examples were updated in https://github.com/remix-run/remix/pull/2250 and merged into `main` without explicitly mentioning deprecation at this point. I will make a notice in the docs about this once this is merged and again PR that change  to `main`.

### DIFF
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -57,7 +57,7 @@ export async function callRouteAction({
 
   if (!isResponse(result)) {
     console.warn(
-      `You returned a values other than a \`Response\` object from the action for route "${match.route.id}". This is deprecated and will result in a server error in a future version of Remix.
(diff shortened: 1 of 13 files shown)


packages/remix-server-runtime/data.ts
```diff
@@ -104,7 +135,38 @@ export async function callRouteLoader({
     );
   }
 
-  return isResponse(result) ? result : json(result);
```
machour, line 107: I think it would be better to create a `warnIfNotResponse()` function and use it in both places.

machour (commented):
A couple of typo fix + a small refactor suggestion

MichaelDeBoey (approved):

chaance:
Closing for now as we haven't completely decided to move forward with this. Will reopen if that changes.

PR #1290, opened 2021-12-30
fix(remix-server-runtime): fix invalid character error in cookie serialize

nimaa77:
### Description

you save a cookie that contains utf-8 characters as it's value, you don't get the right content after reading the cookie.
this is because atob and btoa function that used inside cookie module and these functions don't support utf-8.

[as MDN suggested in their docs,](https://developer.mozilla.org/en-US/docs/Web/API/btoa#unicode_strings) you can convert the utf8 content into binary string and then pass it to `btoa` function and vice versa for ...

this is what this Pull Request is all about, fixing the issue with saving a cookie with utf8 text content

### How to reproduce

https://codesandbox.io/s/dazzling-monad-0wt38?file=/app/routes/index.jsx

closes: #1282

PR #2257, opened 2022-03-08
fix(remix-server-runtime): fix invalid character error in cookie serialize

### LINKED ISSUE #1282: [Bug]: Cookies unable to encode UTF-8 strings

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [X] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

any

14+

### Steps to Reproduce

from https://github.com/sergiodxa/remix-auth/issues/115 (@thangngoc89)

Try to encode a cookie  with a UTF-16 string - ie `"Nguyễn Văn A"`

Essentially `btoa` expects a string to be binary, so encountering characters that are UTF-16 encoded with more than 8 bits will throw an error. This is well-documented by [Mozilla](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem). In JSON and JavaScript, strings are UTF-16, so representing the characters is not an issue. However, this means that `JSON.stringify`ing a string does not make it safe for `btoa`.

### Expected Behavior

Cookie properly set :sunglasses:

### Actual Behavior

Cookie is not set _and_ error is swallowed

### LINKED ISSUE #2255: Invalid character error in cookie serialize

1.2.3

### Steps to Reproduce

 Serialize data containing Non-ASCII characters with `.serialize` the cookie object generated by `createCookie`.

```js
const cookie = createCookie('foo')
cookie.serialize({ data: '日本語' })
```

### Expected Behavior

Serialization can be done without any errors.

### Actual Behavior

Invalid character error occurs.
```
InvalidCharacterError: Invalid character
    at btoa (/Users/uejima.aiji/git/user/node_modules/@miniflare/core/src/standards/encoding.ts:37:13)
    at encodeData (/Users/uejima.aiji/git/user/node_modules/@remix-run/server-runtime/esm/cookies.js:94:10)
    at encodeCookieValue (/Users/uejima.aiji/git/user/node_modules/@remix-run/server-runtime/esm/cookies.js:68:17)
    at Object.serialize (/Users/uejima.aiji/git/user/node_modules/@remix-run/server-runtime/esm/cookies.js:50:56)
    at loader (/Users/uejima.aiji/git/user/build/route:/Users/uejima.aiji/git/user/app/root.tsx:24:16)
    at callRouteLoader (/Users/uejima.aiji/git/user/node_modules/@remix-run/server-runtime/esm/data.js:69:20)
    at /Users/uejima.aiji/git/user/node_modules/@remix-run/server-runtime/esm/server.js:266:108
    at Array.map (<anonymous>)
    at renderDocumentRequest (/Users/uejima.aiji/git/user/node_modules/@remix-run/server-runtime/esm/server.js:266:67)
    at requestHandler (/Users/uejima.aiji/git/user/node_modules/@remix-run/server-runtime/esm/server.js:51:26)
```

aiji42:
Fixed a bug that caused an error when cookie serializing data containing non-ascii characters.

- [x] Tests

Closes #1282
Closes #2255


mkrtchian:
A similar PR exists as #1290, related to issue #1282

aiji42:
@mkrtchian Thanks for letting me know. I did not know that.
This PR takes a different approach (smaller size of the encoded result) than that one. I will leave it to the development team and community to decide which one to adopt.

mjackson:
One thing we need to be careful about: I'm not sure if all the JS runtimes we support will have `escape` and `unescape` because they are currently in Annex B of ECMA-262 which states that we "should not use or assume the existence of these features and behaviors when writing new ECMAScript code".

The main other runtimes I'm concerned about at this point are Cloudflare and Deno, both of which we have support for. Polyfilling the environment is totally something we can do, I just want to make sure we have test coverage for both before merging something like this.

aiji42:
@mjackson Thanks for the follow up.

> Polyfilling the environment is totally something we can do, I just want to make sure we have test coverage for both before merging something like this.

Does this mean specifically that I should use the alternative code (see core-js) instead of using `escape` and `unescape`? Or are you asking me to create samples that runs on both Deno and Cloudflare, test it, and present it?
If you give me specific instructions, I will start working immediately.

aiji42:
I haven't received a reply yet, but I polyfilled `escape` and `unescape` with reference to the core-js code. If the policy is different, please tell me. Withdraw the last commit.

mcansh (approved):

Issue #2195, opened 2022-03-03
`TypeError: unusable` when reading form data in actionfunction using miniflare / Cloudflare


stepri:
### What version of Remix are you using?

1.2.3

### Steps to Reproduce

- create a new project using Cloudflare Pages
- create a form with a POST method
- create an action and try to read formdata
- run project using `npm run dev`

This error also happens on Cloudflare Workers

See example repo: https://github.com/stepri/formdata-remix-cloudflare/blob/main/app/routes/test.tsx

### Expected Behavior

No error

### Actual Behavior

500 Internal Server Error:

```
TypeError: unusable
    at Request.clone (/myproject/node_modules/undici/lib/fetch/request.js:725:13)
    at Request.clone (/myproject/node_modules/@miniflare/core/src/standards/http.ts:423:38)
    at renderDocumentRequest (/myproject/node_modules/@remix-run/server-runtime/esm/server.js:414:50)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at requestHandler (/myproject/node_modules/@remix-run/server-runtime/esm/server.js:51:20)
    at handleFetch (/myproject/node_modules/@remix-run/cloudflare-pages/esm/worker.js:47:18)
    at /myproject/node_modules/@remix-run/cloudflare-pages/esm/worker.js:55:14
    at next (/myproject/node_modules/wrangler/pages/functions/template-worker.ts:108:26)
    at EventTarget.[kDispatchFetch] (/myproject/node_modules/@miniflare/core/src/standards/event.ts:359:13)
    at Server.<anonymous> (/myproject/node_modules/@miniflare/http-server/src/index.ts:168:20)
```

AntanasMisiunas:
I'm experiencing the same issue. However this happens only if I disable Javascript on the site.

brookslybrand:
Thank you for opening this issue, and our apologies we haven't gotten around to it yet!  

With the release of [React Router v7](https://remix.run/blog/react-router-v7) we are sun-setting continued development/maintenance on Remix v2.  If you have not already [upgraded to React Router v7](https://reactrouter.com/upgrading/remix), we recommend you do so. We've tried to make the upgrade process as smooth as possible with our [Future Flags](https://remix.run/docs/en/main/start/future-flags). We are now in the process of cleaning up outdated issues and pull requests to improve the overall hygiene of our repositories.  

We plan to continue to address 2 types of issues in Remix v2:

* Bugs that pose security concerns
[...]

PR #2265, opened 2022-03-08
fix(remix-cloudflare-pages): don't clone request

stepri:
Closes: #2195

#- [X] Docs
#- [X] Tests


GregBrimble:
Not quite sure this will work. Instead, I think we should add `.clone()` [here](https://github.com/remix-run/remix/blob/63c5eff12661e1982a37cef8e1bc919523f93c78/packages/remix-server-runtime/server.ts#L115) and [here](https://github.com/remix-run/remix/blob/63c5eff12661e1982a37cef8e1bc919523f93c78/packages/remix-server-runtime/server.ts#L230). It looks like it was just missed off by accident in those couple of places. The other handlers in that file look to correctly clone the request.

GregBrimble:
Looks like it was [intentionally removed by @jacob-ebey quite recently](https://github.com/remix-run/remix/commit/5d0a0ac89cacd2d30a2d05618140c80d2e07d9c9) to fix some other bug that I don't know enough about.

MichaelDeBoey:
@GregBrimble You mentioned the other PR. Does that also fix the problem you wanted to solve with this PR?
If so, please close this PR & #2195.

GregBrimble:
@MichaelDeBoey , no. @jacob-ebey's commit introduced the bug #2195. But I don't know enough about why that work was done to know how to fix it.

Happy to put up a PR to re-add the `.clone()`, but I imagine y'all probably want some other solution?

3x071c:
Hi, thankfully just found this issue/PR after having thought this was an error on my side. May I ask if there are any updates on this? I'm forced to use javascript-only `Form`s in my Remix+CF Pages project as a workaround now, which isn't ideal really.

MichaelDeBoey:
@stepri Is this PR ready for review?

packages/remix-cloudflare-pages/worker.ts
```diff
@@ -58,7 +58,7 @@ export function createPagesFunctionHandler<Env = any>({
     try {
```
jacob-ebey, line 58: Let's wrap the attempt at serving static files in a `if (["GET", "HEAD"].includes(request.method)) {` to prevent post request bodies from being locked.

chaance:
@stepri Mind addressing @jacob-ebey's feedback? We can get this merged after that's done.

Issue #125, opened 2020-12-08
@remix-run/firebase


ryanflorence:
Probably just grab David East's template and go.

ccssmnn:
What features should this template include?

Probably Auth & Firestore.

I would love to help.

Ehesp:
The template is probably reference to https://github.com/davideast/remix-starter-firebase (Remix adapter for running on Firebase Functions)

ccssmnn:
Alright, so this Issue is about firebase as a deployment target, not about an example on how to use remix and firebase together.

Ehesp:
I'd assume so yeah given the title :) 

Although a Remix + Firebase example would probably be worthwhile too.

chaance:
Probably not going to add an adapter for this anytime soon, though something in the `examples` directory would be great. Also a great candidate for someone to build a stack on top of. Closing for now since there's nothing actionable for us.

PR #2266, opened 2022-03-08
feat: add `@remix-run/google-cloud-functions` package

eastlondoner:
This is an adapter for google cloud functions. We believe that it works with gen1, gen2 and firebase http functions on google cloud.

The implementation is extremely similar to the express adapter. The main differences are that this uses types from `@google-cloud/functions-framework` and in `createRemixRequest` the `rawBody` parameter is used:

```
  if (req.method !== "GET" && req.method !== "HEAD") {
    init.body = req.rawBody;
  }
```

The latter is needed because GCF includes some pre-processing that already parses POST, PUT etc and sets `req.body` into an object - but Remix isn't expecting that to have happened and ends up blocking indefinitely inside busboy or something like that. Using `rawBody` provides a URL encoded buffer - which is one of the body types that Remix is expecting.


packages/remix-gcloud-functions-gen2/package.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "dependencies": {
+    "@remix-run/node": "1.5.1"
+  },
+  "peerDependencies": {
+    "express": "^4.17.1"
+  },
+  "devDependencies": {
+    "@google-cloud/functions-framework": "^3.1.1",
+    "@remix-run/server-runtime": "1.5.1"
```
MichaelDeBoey, line 22: You should use `@remix-run/node` instead

```suggestion
```

packages/remix-gcloud-functions-gen2/server.ts
```diff
@@ -0,0 +1,164 @@
+import type { NextFunction } from "express";
+import type { AppLoadContext, ServerBuild } from "@remix-run/server-runtime";
```
MichaelDeBoey, line 2:
```suggestion
import type { AppLoadContext, ServerBuild } from "@remix-run/node";
```

packages/remix-gcloud-functions-gen2/index.ts
```diff
@@ -0,0 +1,4 @@
+import "./globals";
+
+// export type { GetLoadContextFunction, RequestHandler } from "./server";
+export { createRequestHandler, createRemixHeaders, createRemixRequest } from "./server";
```
MichaelDeBoey, line 4: There's no need to export `createRemixHeaders` & `createRemixRequest` from the adapter I think?

```suggestion
export { createRequestHandler } from "./server";
```

build.sh
```diff
@@ -0,0 +1,5 @@
+rm -rf build/
```
MichaelDeBoey, line 1: I guess we can remove this file?

link.sh
```diff
@@ -0,0 +1,3 @@
+
```
MichaelDeBoey, line 1: I guess we can remove this file?

publish.sh
```diff
@@ -0,0 +1,2 @@
+./build.sh
```
MichaelDeBoey, line 1: I guess we can remove this file?

MichaelDeBoey (changes requested):

packages/remix-gcloud-functions-gen2/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "remix-adapter-gcf-gen2",
+  "description": "Unofficial Google Cloud Functions (gen2) request handler for Remix",
+  "version": "0.0.8",
```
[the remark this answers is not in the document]
penx: I would recommend naming this `@remix-run/google-cloud-functions`:

- remove "-gen2" as the adapter will work on both v1 and v2 of Google Cloud functions. I have a similar adapter deployed to Firebase which I have tested for both versions
- gcloud => google-cloud as no other adapters or packages use abbreviations

packages/remix-gcloud-functions-gen2/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "url": "https://github.com/eastlondoner/remix/issues"
+  },
+  "dependencies": {
+    "@remix-run/node": "1.5.1"
+  },
+  "peerDependencies": {
+    "express": "^4.17.1"
+  },
+  "devDependencies": {
+    "@google-cloud/functions-framework": "^3.1.1",
```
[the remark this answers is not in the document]
penx: as far as I'm aware it's only used for types, but I think that still means it should be in dependencies

packages/remix-gcloud-functions-gen2/server.ts
```diff
@@ -0,0 +1,164 @@
...
+import {
+  createRequestHandler as createRemixRequestHandler,
+  Headers as NodeHeaders,
+  Request as NodeRequest,
+  writeReadableStreamToWritable
+} from "@remix-run/node";
+import {
+  // This has been added as a global in node 15+
+  AbortController,
+} from "@remix-run/node";
```
penx, line 20: The "@remix-run/node" imports can be merged

packages/remix-gcloud-functions-gen2/server.ts
```diff
@@ -0,0 +1,164 @@
...
+ * `action` functions.
+ *
+ * You can think of this as an escape hatch that allows you to pass
+ * environment/platform-specific values through to your loader/action.
+ */
+export interface GetLoadContextFunction {
+  (req: GcfRequest, res: GcfResponse): AppLoadContext;
+}
+
+//export type RequestHandler = ReturnType<typeof createRequestHandler>;
```
penx, line 34:
```suggestion
```

packages/remix-gcloud-functions-gen2/server.ts
```diff
@@ -0,0 +1,164 @@
...
+      let response = (await handleRequest(
+        request,
+        loadContext
+      )) as NodeResponse;
+
+      await sendRemixResponse(res, response);
+    } catch (error) {
+      // Express doesn't support async functions, so we have to pass along the
+      // error manually using next().
+      next(error);
```
penx, line 72: Pretty sure this will fail as Google Cloud functions don't provide a next function.

https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/759aac18ad297363c4651efcf19ab2bf8fa9625a/src/functions.ts#L46-L48

I was working on an adapter independently, and throw an error at this point in my version. Likely needs a test though https://github.com/remix-run/remix/pull/3397

packages/remix-gcloud-functions-gen2/server.ts
```diff
@@ -0,0 +1,164 @@
...
+  let headers = new NodeHeaders();
+
+  for (let [key, values] of Object.entries(requestHeaders)) {
+    if (values) {
+      if (Array.isArray(values)) {
+        for (let value of values) {
+          headers.append(key, value);
+        }
+      } else {
+        headers.set(key, values as any);
```
penx, line 89: don't think you need to cast to `any` here

packages/remix-gcloud-functions-gen2/package.json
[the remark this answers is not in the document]
MichaelDeBoey: @penx Good idea! 👍
Updated my suggestion accordingly.

packages/remix-gcloud-functions-gen2/server.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  let init: NodeRequestInit = {
+    method: req.method,
+    headers: createRemixHeaders(req.headers),
+    signal: controller.signal,
+  };
+
+  if (hasBody(req)) {
+    init.body = createRemixBody(req);
+  }
```
penx, line 117: rawBody is available on `req` so we don't need to create it again.

I'd followed @ryanflorence 's [recommendation on Discord](https://discord.com/channels/770287896669978684/771068344320786452/905168966706675782) and [gist](https://gist.github.com/ryanflorence/8a12cef023765bd0520a04cf25d07cf7#file-remix-firebase-js-L49-L51):

```suggestion
  if (req.method !== "GET" && req.method !== "HEAD") {
    init.body = req.rawBody;
  }
```

packages/remix-gcloud-functions-gen2/server.ts
MichaelDeBoey: It should be exported instead

```suggestion
export type RequestHandler = ReturnType<typeof createRequestHandler>;
```

packages/remix-gcloud-functions-gen2/server.ts
MichaelDeBoey:
```suggestion
        headers.set(key, values);
```

packages/remix-gcloud-functions-gen2/server.ts
MichaelDeBoey, line 20: See https://github.com/remix-run/remix/pull/2266#discussion_r890304493

```suggestion
import {
  // This has been added as a global in node 15+
  AbortController,
  createRequestHandler as createRemixRequestHandler,
  Headers as NodeHeaders,
  Request as NodeRequest,
  writeReadableStreamToWritable
} from "@remix-run/node";
```

packages/remix-gcloud-functions-gen2/server.ts
```diff
@@ -0,0 +1,164 @@
...
+ * something that Remix is expecting. In this case a Buffer with URLSearchParams encoded data works.
+ *
+ * @param req the request passed in from the Google Cloud Function
+ */
+function createRemixBody(req: GcfRequest) {
+  let s = new Readable()
+  s.push((new URLSearchParams(req.body).toString()).toString());
+  s.push(null);
+  return s;
+}
```
MichaelDeBoey, line 153: See https://github.com/remix-run/remix/pull/2266#discussion_r890318763

```suggestion
```

packages/remix-gcloud-functions-gen2/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "target": "ES2019",
+
+    "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
+    "strict": true,
+
+    "declaration": true,
+    "emitDeclarationOnly": true,
+
+    "outDir": "../../build/node_modules/@remix-run/gcloud-functions-gen2",
```
MichaelDeBoey, line 14:
```suggestion
    "outDir": "../../build/node_modules/@remix-run/google-cloud-functions",
```

tsconfig.json
```diff
@@ -10,6 +10,7 @@
     { "path": "packages/remix-cloudflare-workers" },
     { "path": "packages/remix-dev" },
     { "path": "packages/remix-express" },
+    { "path": "packages/remix-gcloud-functions-gen2" },
```
MichaelDeBoey, line 13:
```suggestion
    { "path": "packages/remix-google-cloud-functions" },
```

MichaelDeBoey (changes requested):

MichaelDeBoey (changes requested):
@eastlondoner Code changes should go against `dev` branch.
Could you please rebase your branch onto latest `dev` & set base branch to `dev` as well?

Please also make sure to rename the `remix-gcloud-functions-gen2` folder to `remix-google-cloud-functions`

eastlondoner:
Hi everyone!

Thanks for all the feedback, this was left without any feedback for a long time so it's caught me a bit off-guard.

I will try and take a look at this and update it as soon as I can

penx:
I've opened https://github.com/eastlondoner/remix/pull/1 which adds the unit tests from the now-closed #3397

packages/remix-gcloud-functions-gen2/server.ts
penx: it's exported on the next line already (35)

build.sh
eastlondoner: yeah this is for my own deployment so I could actually use this

packages/remix-gcloud-functions-gen2/server.ts
MichaelDeBoey: `@remix-run/server-runtime` is the implementation detail of `@remix-run/{cloudflare,deno,node}`, which includes common functionality.

Specific runtimes either export the functionality directly _or_ create a custom version of it, so we should always use a specific runtime as a user.

packages/remix-google-cloud-functions/__tests__/server-test.ts
```diff
@@ -0,0 +1,271 @@
+import supertest from "supertest";
+import { createRequest } from "node-mocks-http";
+import {
+  createRequestHandler as createRemixRequestHandler,
+  Response as NodeResponse,
+} from "@remix-run/node";
+import { Readable } from "stream";
+import { http } from "@google-cloud/functions-framework";
+import { getTestServer } from "@google-cloud/functions-framework/build/src/testing.js";
```
eastlondoner, line 9: @penx how do you feel about using this import? 

It's not pretty but it works without blocking this on the upgrade to jest 28

eastlondoner:
@MichaelDeBoey I think this is ready for review again when you have time :-)

packages/remix-google-cloud-functions/__tests__/server-test.ts
MichaelDeBoey: I think we can keep this like this for now & change in the Jest 28 PR

MichaelDeBoey (changes requested):

packages/remix-google-cloud-functions/__tests__/server-test.ts
MichaelDeBoey: CI seems to be green, so let's keep the import as-is then. 👍

penx:
Approved from me other than

- https://github.com/remix-run/remix/pull/2266#discussion_r892156729
- https://github.com/remix-run/remix/pull/2266#discussion_r892158040

It would be nice to have a unit test that covers an error being thrown as per my comment at

- https://github.com/remix-run/remix/pull/2266#discussion_r890312851

...however off the top of my head I don't have a suggestion on how to create this test 😄

penx (approved):

eastlondoner:
@MichaelDeBoey do you have the ability to re-run the failing Github Action ? It's failing with a timeout while installing packages which I am pretty sure is a build system problem not a problem this PR has introduced. But I cannot see a place where I can trigger the CI to re-run

eastlondoner:
Actually moving back to draft status - we did a manual test with real deployment google cloud functions and we're not sure that this is working - hold on while we check ourselves

eastlondoner:
Update from our testing (thanks @danny-yu):

## Deploying with regular google cloud functions

If we deploy a remix app using this adapter to a google cloud function called 'remixApp' like this:
```
gcloud functions deploy remixApp --runtime nodejs16 --trigger-http --source . --allow-unauthenticated
```
then we get a function that we can visit at a url something like `https://us-central1-<my project id>.cloudfunctions.net/remixApp`

notice that google cloud functions adds a base path `/remixApp`.

The remix app doesn't know about the base path and all `<Form>` and similar actions in the remix app are immediately broken because they don't include it. A simple `Form` tries to post to `https://us-central1-<my project id>.cloudfunctions.net/` which gets back a google cloud permissions error.

The google cloud function behaviour is documented here: https://cloud.google.com/functions/docs/calling/http

this will probably throw a lot of people who try to use this adapter off because this is the most straightforward way to check an app after deployment but what they will experience is something that looks OK but breaks unexpectedly on any form submission - not a good experience

I know that there are other ways to use this

 - using google cloud functions 'gen2' doesn't have this problem, the deployed function has a unique url without an extra path - we are currently checking that this adapter works in that situation.
 - deploying using firebase functions  - I don't know what happens if you use firebase functions, @penx might know ?  Otherwise @danny-yu could investigate.
 - I don't know if there are deployment options to combine gen 1 functions with other google cloud infrastructure like load balancers that might fix this - but it would affect users trying out the simplest use case.

penx:
I have a Remix application deployed to Firebase functions v1 and v2 using an adapter almost identical to this, which I published to npm at https://www.npmjs.com/package/remix-google-cloud-functions

I have this in a draft PR against the Remix Firebase example at https://github.com/remix-run/remix/pull/3398

The things that may be of help:

1. All incoming http requests go to Firebase hosting which has a rewrite rule in firebase.json to pass all requests other than /public/* to the function:

```json
    "rewrites": [{
       "source": "**",
       "function": "remix"
     }]
```

2. from what I remember, when I initially deployed the function I had to set some permissions on it so that it will run for unauthenticated users, but after I'd done this once it didn't come up again. When I next get some time, I'll run through the process again and add add it to the readme for #3398.

3. When deploying via the Firebase CLI, it said that v2 functions need to be lowercase, I'm not sure if that's a Firebase specific thing though.

eastlondoner:
great update @penx

I am very convinced this works for google cloud functions *provided the correct configuration / deployment* - it's just slightly sad that the most basic of google cloud function deployments will see limited functionality.

Some thoughts:

 - is there a good place to document google cloud deployment configurations that do work and those that don't work?
 - what about inspecting the host header of incoming requests and logging a warning in the adapter if we see `*.cloudfunctions.net`? 
 
 Promoting this back to a PR though - I'm in favour or merging as-is

packages/remix-google-cloud-functions/server.ts
```diff
@@ -0,0 +1,133 @@
...
+  (req: GcfRequest, res: GcfResponse): AppLoadContext;
+}
+
+export type RequestHandler = (
+  req: GcfRequest,
+  res: GcfResponse,
+) => Promise<void>;
+
+/**
+ * Returns a request handler for Express that serves the response using Remix.
```
MichaelDeBoey, line 36:
```suggestion
 * Returns a request handler for Google Cloud Functions that serves the response using Remix.
```

packages/remix-google-cloud-functions/server.ts
```diff
@@ -0,0 +1,133 @@
...
+      let response = (await handleRequest(
+        request,
+        loadContext
+      )) as NodeResponse;
+
+      await sendRemixResponse(res, response);
+    } catch (error) {
+        console.error(error);
+        await sendRemixResponse(res ,new NodeResponse("Internal Error", { status: 500 }));
+    }
```
MichaelDeBoey, line 66: Other adapters don't have a `try`/`catch` block here, so let's not do it either.

```suggestion
    let request = createRemixRequest(req);
    let loadContext =
      typeof getLoadContext === "function"
        ? getLoadContext(req, res)
        : undefined;

    let response = (await handleRequest(
      request,
      loadContext
    )) as NodeResponse;

    return sendRemixResponse(res, response);
```

packages/remix-google-cloud-functions/__tests__/server-test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+import {
+  createRemixHeaders,
+  createRemixRequest,
+  createRequestHandler,
+} from "../server";
+
+// We don't want to test that the remix server works here (that's what the
+// puppetteer tests do), we just want to test the express adapter
```
MichaelDeBoey, line 18:
```suggestion
// puppetteer tests do), we just want to test the Google Cloud Functions adapter
```

packages/remix-google-cloud-functions/__tests__/server-test.ts
```diff
@@ -0,0 +1,271 @@
...
+        "first=one; Expires=0; Path=/; HttpOnly; Secure; SameSite=Lax",
+        "second=two; MaxAge=1209600; Path=/; HttpOnly; Secure; SameSite=Lax",
+        "third=three; Expires=Wed, 21 Oct 2015 07:28:00 GMT; Path=/; HttpOnly; Secure; SameSite=Lax",
+      ]);
+    });
+  });
+});
+
+describe("google-cloud-functions createRemixHeaders", () => {
+  describe("creates fetch headers from express headers", () => {
```
MichaelDeBoey, line 135:
```suggestion
  describe("creates fetch headers from Google Cloud Functions headers", () => {
```

packages/remix-google-cloud-functions/__tests__/server-test.ts
```diff
@@ -0,0 +1,271 @@
...
+          Symbol(context): null,
+        }
+      `);
+    });
+  });
+});
+
+describe("google-cloud-functions createRemixRequest", () => {
+  it("creates a request with the correct headers", async () => {
+    let expressRequest = createRequest({
```
MichaelDeBoey, line 225:
```suggestion
    let request = createRequest({
```

packages/remix-google-cloud-functions/__tests__/server-test.ts
```diff
@@ -0,0 +1,271 @@
...
+      method: "GET",
+      protocol: "http",
+      hostname: "localhost",
+      headers: {
+        "Cache-Control": "max-age=300, s-maxage=3600",
+        Host: "localhost:3000",
+      },
+    });
+
+    expect(createRemixRequest(expressRequest)).toMatchInlineSnapshot(`
```
MichaelDeBoey, line 236:
```suggestion
    expect(createRemixRequest(request)).toMatchInlineSnapshot(`
```

MichaelDeBoey (changes requested):

packages/remix-google-cloud-functions/server.ts
eastlondoner: also the express adapter has `next(error)` which serves a similar purpose but as previously discussed on this PR google cloud functions doesn't have an equivalent

MichaelDeBoey: `@remix-run/deno` is a runtime, not an adapter.

`@remix-run/express` indeed has a `try`/`catch` in `createRequestHandler`, but that's because `express` doesn't support async functions, so we have to pass along the error manually using `next()`.

penx:
@MichaelDeBoey think this is ready for another review, in case you missed it 🙂

packages/remix-google-cloud-functions/server.ts
penx: > Without this errors go unrecorded and the client http request will just hang

although this is a fair point, I'm not sure  what the best approach would be in that case, @MichaelDeBoey ?

penx: in the adapter I was writing, I was originally using the firebase-functions library and throwing a specific error type, but this didn't seem to exist in the google cloud functions library, sec I'll go dig it out

penx: Here's the `firebase-functions` documentation on throwing `functions.https.HttpsError`, though as above I couldn't find an equivalent for `@google-cloud/functions-framework`


https://firebase.google.com/docs/functions/callable#handle_errors

penx: I think we need a unit test here that shows the hanging or handling of the error depending on approach

eastlondoner:
@MichaelDeBoey this is updated ready for review - I've removed the error handling. 

There doesn't seem to be definitive documentation from google regarding error handling so we carried out quick tests and it seems that thrown errors and rejected promises are handled by google cloud functions gen 1 and gen 2 both returning a 500 error in good time when the result of `createRequestHandler` is awaited so my claims about hanging requests must have been based on cases where I wasn't correctly awaiting the response, sorry.

our tests if you are interested:
https://gist.github.com/danny-yu/65e52467123991e09cab03bf44ec5925
https://gist.github.com/danny-yu/de77a8508781137f544bb4b1de623055

packages/remix-google-cloud-functions/server.ts
```diff
@@ -0,0 +1,129 @@
...
+  mode?: string;
+}): RequestHandler {
+  let handleRequest = createRemixRequestHandler(build, mode);
+
+  return async (req: GcfRequest, res: GcfResponse) => {
+    let request = createRemixRequest(req);
+    let loadContext =
+      typeof getLoadContext === "function"
+        ? getLoadContext(req, res)
+        : undefined;
```
MichaelDeBoey, line 55: See #3442

```suggestion
    let loadContext = getLoadContext?.(req, res);
```

MichaelDeBoey (changes requested):

eastlondoner:
did we make it gif

eastlondoner:
did we make it gif

packages/remix-google-cloud-functions/__tests__/server-test.ts
```diff
@@ -0,0 +1,271 @@
+import supertest from "supertest";
+import { createRequest } from "node-mocks-http";
```
MichaelDeBoey, line 2: I think these packages should be in `devDependencies` as well?

packages/remix-google-cloud-functions/__tests__/server-test.ts
```diff
@@ -0,0 +1,271 @@
...
+import {
+  createRequestHandler as createRemixRequestHandler,
+  Response as NodeResponse,
+} from "@remix-run/node";
+import { Readable } from "stream";
```
MichaelDeBoey, line 7: This one as well

packages/remix-google-cloud-functions/__tests__/server-test.ts
eastlondoner: @danny-yu 's added these

packages/remix-google-cloud-functions/__tests__/server-test.ts
eastlondoner: I'm not sure what you mean here? We've added `"@remix-run/node"` to devDependencies, although it's already a regular dependency. `stream` is part of node core library and it doesn't make sense to add it to dependencies

eastlondoner:
@MichaelDeBoey I think we made the last changes that you requested, is there anything else ?

MichaelDeBoey:
@eastlondoner Could you please rebase your branch onto latest `dev` & resolve conflicts?

eastlondoner:
@MichaelDeBoey rebased and green 😀

eastlondoner:
hi @MichaelDeBoey is there anything else needed on this PR ?

MichaelDeBoey:
@eastlondoner The team will look into this PR whenever they have the time for it, but new adapters are not high on the priority list for the moment.

eastlondoner:
thanks for the update :-)

chaance:
We aren't looking to add new adapter packages for the time being, though we encourage hosts to maintain and promote them! If anything changes on that front we will reopen this.

ronnylt:
> We aren't looking to add new adapter packages for the time being, though we encourage hosts to maintain and promote them! If anything changes on that front we will reopen this.

This seems to be a wise move. Can you please make this "official", for example by adding a note in the the [contributing guidelines](https://github.com/remix-run/remix/blob/main/docs/pages/contributing.md) or in the documentation.

machour:
Could be documented in https://github.com/remix-run/remix/blob/main/docs/other-api/adapter.md as well 👌🏼

penx:
FWIW this doubles as a Firebase adapter, and there was a ticket for this, though now closed #125.

I have published the adapter I made in #3398 to npm, from my branch at https://github.com/penx/remix/tree/firebase-adapter:

- https://www.npmjs.com/package/remix-google-cloud-functions
- https://www.npmjs.com/package/remix-firebase

Though these haven't seen much activity as I was hoping this PR would go in.

I'll revisit these and release them as v1 with their own repository at some point.

machour:
@penx ping us on discord when it's done, we'll add links in the documentation

PR #2270, opened 2022-03-09
Warn only when adding fake builtins

mzaatar:
Why: 
There are a lot of issues and discussions to open the REMIX `esbuild` process to allow inject modules that might be removed by the tree shaking process. 
One example is `Buffer` node npm package which is required by many projects but cannot be added to Remix projects as this error will stop the Remix building process.

There are many ~solutions~ hacks now:
* by either hijacking the esbuild config
* using `patch-package` to patch Remix compiler
* using node packages directly into the `root.tsx`

Solution:   
This small change will allow projects to add the required packages and enable this flag to warn us that we know what we do.

This is to 
- [ ] Docs
- [ ] Tests


packages/remix-dev/compiler.ts
```diff
@@ -329,11 +329,19 @@ async function createBrowserBuild(
...
+          ", "
+        )} before continuing.`
+      );
+    } else {
+      throw new Error(
+        `It appears you're using a module that is built in to node, but you installed it as a dependency which could cause problems. Please remove ${fakeBuiltins.join(
+          ", "
+        )} before continuing.`
+      );
+    }
```
MichaelDeBoey, line 344:
```suggestion
    let message = `It appears you're using a module that is built in to node, but you installed it as a dependency which could cause problems. Please remove ${fakeBuiltins.join(
      ", "
    )} before continuing.`;

    if (config.warnOnFakeBuiltins) {
      console.warn(message);
    } else {
      throw new Error(message);
    }
```

packages/remix-dev/config.ts
```diff
@@ -401,6 +412,8 @@ export async function readConfig(
 
   let serverDependenciesToBundle = appConfig.serverDependenciesToBundle || [];
 
+  let warnOnFakeBuiltins = appConfig.warnOnFakeBuiltins? appConfig.warnOnFakeBuiltins: false
```
MichaelDeBoey, line 415: This can be simplified to

```suggestion
  let warnOnFakeBuiltins = appConfig.warnOnFakeBuiltins ?? false;
```

MichaelDeBoey (changes requested):

MichaelDeBoey:
Hi @mzaatar!

mzaatar:
Hey @MichaelDeBoey , I just pushed it now. Please check and let me know if it is ok to merge it?

Issue #1291, opened 2021-12-30
`useFetcher().submit` client side validation is wrong


ryanflorence:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [ ] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [X] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

1.1.1

### What version of Node are you using? Minimum supported version is 14.

16

### Steps to Reproduce

```js
// in routes/some-layout
// URL at /some-layout/some-param

let fetcher = useFetcher()
let action = useFormAction()
console.log(action)
fetcher.submit({}, { method: "post", action })
```

### Expected Behavior

It should post to the layout route.

### Actual Behavior

It doesn't ever submit to the server and logs this error in the browser:

> transition.js:858 Uncaught (in promise) Error: Route "routes/some-layout/index" does not have an action, but you are trying to submit to it. To fix this, please add an action function to the route

What's really strange is that the `action` string is correct, it's `"/some-layout"`. Looks like calling `fetcher.submit()` does some validation in the client against the route manifest and incorrectly validates the an action on "routes/some-layout/index" instead of "routes/some-layout".

## Two Bugs To Fix 

1. The error boundary should have caught this, but didn't, fix that first.
2. (my guess is) the validation is just wrong, have it look for the correct route, or just skip validation in the client and let the server handle it.

ryanflorence:
@jacob-ebey I think we missed a spot with our `?index` action disambiguation in 0.19 (0.20?)

anonrig:
Hi @ryanflorence 

I'm seeing a familiar problem where the route has an action function but using fetcher.submit with URLSearchParams throws the same error. I'm attaching the logs.

[remix-error.log](https://github.com/remix-run/remix/files/7849821/remix-error.log)

LANCENET:
What is it

chaance:
Closed in https://github.com/remix-run/remix/pull/2271

PR #2271, opened 2022-03-09
fix: disambiguate layout vs index route fetcher submissions

jacob-ebey:
Closes: https://github.com/remix-run/remix/issues/1291

- [ ] Docs
- [x] Tests

### DIFF
--- a/integration/fetcher-layout-test.ts
+++ b/integration/fetcher-layout-test.ts
@@ -24,7 +24,7 @@ beforeAll(async () => {
             <div>
               <h1>Layout</h1>
               <button onClick={invokeFetcher}>Invoke Fetcher</button>
-              {!!fetcher.data && <p id="fetcher-data">{fetcher.data}</p>}
+              {!!fetcher.data && <p id="layout-fetcher-data">{fetcher.data}</p>}
               <Outlet />
             </div>
           );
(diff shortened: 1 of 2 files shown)


packages/remix-react/transition.ts
```diff
@@ -596,6 +600,28 @@ export function createTransitionManager(init: TransitionManagerInit) {
...
+  function isIndexRequestUrl(url: URL) {
+    let indexRequest = false;
+
+    for (let param of url.searchParams.getAll("index")) {
+      if (!param) {
+        indexRequest = true;
+      }
+    }
+
+    return indexRequest;
```
mjackson, line 612:
```suggestion
    for (let param of url.searchParams.getAll("index")) {
      if (!param) return true;
    }

    return false;
```

mjackson (commented):
I'm not super familiar with this code so I'll defer to @ryanflorence

packages/remix-react/transition.ts
ryanflorence: Okay, this is weird I know ... but folks may need to use `?index=123` for bookmark pagination (or really anything), so the way we can do this without clobbering the `index` search param namespace is *only a bare index param means the index route*.

```
✅ /foo?index
✅ /foo?index&index=123
✅ /foo?index=123&index
❌ /foo?index=123
```

This code does exactly that, but it's not super clear since it's relying on "falsey" empty string and the fact that the loop happened at all. I'd go with:


```js
for (let param of url.searchParams.getAll("index")) {
  if (param === "") return true;
}
```

ryanflorence, line 612:
```suggestion
  function isIndexRequestUrl(url: URL) {
    for (let param of url.searchParams.getAll("index")) {
      // only use bare `?index` params without a value
      // ✅ /foo?index
      // ✅ /foo?index&index=123
      // ✅ /foo?index=123&index
      // ❌ /foo?index=123
      if (param === "") return true;
    }
    
    return false;
```

ryanflorence: Seems worth calling out for "future us".

integration/fetcher-layout-test.ts
```diff
@@ -0,0 +1,81 @@
...
+});
+
+it("fetcher calls layout route action when at paramaterized route", async () => {
+  await app.goto("/layout/foo");
+  await app.clickElement("button");
+  let dataElement = await app.getElement("#fetcher-data");
+  expect(dataElement.text()).toBe("layout action data");
+  dataElement = await app.getElement("#child-data");
+  expect(dataElement.text()).toBe("foo");
+});
```
ryanflorence, line 81: Can we get a test that tests `fetcher.load("/path?index")` and `fetcher.submit("/path?index")` here too? Or did I miss it?

ryanflorence (changes requested):
Would like tests for `path?index` and a bit clearer intent in the loop :)

integration/fetcher-layout-test.ts
jacob-ebey: Loaders are in another ticket I'll do after this merges. There are changes that need to be made there. I've added a test for submitting to index loaders though.

packages/remix-react/transition.ts
jacob-ebey: Good call.

PR #2280, opened 2022-03-09
docs(examples): bring `msw` example in line with templates

MichaelDeBoey:
Like I did in #2200

Follow-up of #2156


kentcdodds (approved):
Super 👍

PR #2285, opened 2022-03-10
docs(examples): make Stitches example render properly the SSR pass

matthova:
This PR patches the the [stitches](https://github.com/remix-run/remix/tree/main/examples/stitches) example to properly set server-rendered styles. Previously, the styles (other than the theme) were not set on initial render. I'm not the biggest fan of the regex I wrote, please advise if a better replacement comes to mind.


examples/stitches/app/entry.server.tsx
```diff
@@ -11,15 +11,16 @@ export default function handleRequest(
...
+  let sheet = getCssText();
 
   let markup = renderToString(
     <ServerStyleContext.Provider value={sheet}>
       <RemixServer context={remixContext} url={request.url} />
     </ServerStyleContext.Provider>
   );
 
-  markup = markup.replace("__STYLES__", sheet);
+  sheet = getCssText();
```
LexSwed, line 22: The change of `const sheet = getCssText();` is unnecessary, repeating twice `getCssText()`

matthova: Or 
```ts
  const sheet = getCssText();

  let markup = renderToString(
    <ServerStyleContext.Provider value={sheet}>
    </ServerStyleContext.Provider>
  );

  const updatedSheet = getCssText();
  markup = markup.replace(/<style id=\"stitches\">.*\<\/style>/g, `<style id="stitches">${updatedSheet}</style>`);
```

LexSwed: The later looks good, it was not immediately clear to me why calling it twice 👀 And still not clear though 😅  If `getCssText` is only returning a string, why passing it to the initial markup render? It's not gonna be used during first render anyway, right?
I mean changing this to:
```js
  let markup = renderToString(
  );

  const cssText = getCssText();
  markup = markup.replace(/<style id=\"stitches\">.*\<\/style>/g, `<style id="stitches">${cssText}</style>`);
```
will work for the resulting markup anyway, right?

matthova: Good eye. This pattern had been copy/pasted from the `styled-components` example, where it appears the server context isn't necessary at all. I've updated it to be a bit cleaner.

LexSwed (approved):
Looks sweet to me 🍰

kentcdodds (approved):
Looks like this is fixing some copy/paste mistake. Thanks!

coreybruyere:
@matthova I'm getting FOUC when following your example? Care to take a look here? https://github.com/coreybruyere/california-custom-tile/pull/2

daphnesmit:
Quick dumb question; where would you typically but the : 
```
  globalStyles();
```

of stitches?

PR #2286, opened 2022-03-10
docs: add remix stacks docs

kentcdodds:
- [x] Docs
~~- [ ] Tests~~ N/A

### DIFF
--- a/docs/pages/stacks.md
+++ b/docs/pages/stacks.md
@@ -1,0 +1,84 @@
+---
+title: Remix Stacks
+description: The quickest way to get rocking and rolling with Remix
+order: 3
+---
+
+# Remix Stacks
+
+Remix Stacks is a feature of the Remix CLI that allows you to generate a Remix project quickly and easily. There are several built-in and official stacks that are full blown applications. You can also make your own (read more below).
+
+The built-in official stacks come ready with common things you need for a production application including:
+
+- Database
+- Automatic deployment pipelines
+- Authentication
+- Testing
+- Linting/Formatting/TypeScript
+
+What you're left with is everything completely set up for you to just get to work building whatever amazing web experience you want to build with Remix. Here are the built-in official stacks:
+
+- [The Blues Stack](https://github.com/remix-run/blues-stack): Deployed to the edge (distributed) with a long-running Node.js server and PostgreSQL database. Intended for large and fast production-grade applications serving millions of users.
+- [The Indie Stack](https://github.com/remix-run/indie-stack): Deployed to a long-running Node.js server with a persistent SQLite database. Intended to get you going quickly with little complexity for prototypes and proof of concepts that can be later updated to the Blues stack without much trouble.
+- [The Grunge Stack](https://github.com/remix-run/grunge-stack): Deployed to a serverless function running Node.js with DynamoDB for persistance. Intended for folks who want to deploy a production-grade application on AWS infrastructure serving millions of users.
+
+Yes, these are named after music genres. 🤘 Rock on.
+
+There will be more stacks available in the future. And you can make your own (and we strongly encourage it)!
+
+## Custom Stacks
+
+The Remix CLI will help you get started with one of these built-in stacks, but if you want, you can create your own stack and the Remix CLI will help you get started with that stack. There are several ways to do this, but the most straightforward is to create a GitHub repo:
+
+```
+npx create-remix@latest --template my-username/my-repo
+```
+
+Custom stacks give an enormous amount of power and flexibility and we hope you create your own that suites the preferences of you and your organization (feel free to fork ours!).
+
+### `--template`
+
+The template option can be any of the following values:
+
+- The name of a stack in the remix-run GH org (e.g. `blues-stack`)
+- A GH username/repo combo (e.g. `mcansh/snkrs`)
+- A file path to a directory on disk (e.g. `/my/remix-stack`)
+- A path to a tarball on disk (e.g. `/my/remix-stack.tar.gz`)
+- A URL to a tarball (e.g. `https://example.com/remix-stack.tar.gz`)
+- A file URL (e.g. `file:///Users/michael/remix-stack.tar.gz`)
+
+Additionally, if your stack is in a private GitHub repo, you can set a GitHub token in the `PRIVATE_GITHUB_TOKEN` environment variable:
+
+```
+PRIVATE_GITHUB_TOKEN=yourtoken npx create-remix@latest --template your-private/repo
+```
+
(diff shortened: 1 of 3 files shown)


mcansh (commented):
this is great!

docs/pages/stacks.md
```diff
@@ -0,0 +1,84 @@
...
+The template option can be any of the following values:
+
+- The name of a stack in the remix-run GH org (e.g. `blues-stack`)
+- A GH username/repo combo (e.g. `mcansh/snkrs`)
+- A file path to a directory on disk (e.g. `/my/remix-stack`)
+- A path to a tarball on disk (e.g. `/my/remix-stack.tar.gz`)
+- A URL to a tarball (e.g. `https://example.com/remix-stack.tar.gz`)
+- A file URL (e.g. `file:///Users/michael/remix-stack.tar.gz`)
+
+Additionally, if your stack is in a private GitHub repo, you can set a GitHub token in the `PRIVATE_GITHUB_TOKEN` environment variable:
```
mcansh, line 50: i currently have this as just `GITHUB_TOKEN`, but i have no preference on name

docs/pages/stacks.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+PRIVATE_GITHUB_TOKEN=yourtoken npx create-remix@latest --template your-private/repo
+```
+
+The [token just needs `repo` access][repo access token].
```
kentcdodds, line 56: Is this correct @mcansh?

docs/pages/stacks.md
```diff
@@ -0,0 +1,84 @@
...
+#### Remix deps
+
+If you set the any remix dependencies in package.json to `*`, the Remix CLI will change it to a semver carrot of the latest released version:
+
+```diff
+-   "remix": "*",
++   "remix": "^1.2.3",
+```
+
+This allows you to not have to regularly update your template to the latest version of Remix. Of course you do not have to put `*` if you'd prefer to manually manage that version.
```
kentcdodds, line 69: Is this correct? I assumed this is how it works, but wasn't sure.

kentcdodds (commented):
@mcansh could you double check the bits I'm asking about. I just want to be sure :)

docs/pages/stacks.md
kentcdodds: I looked in the commit where you added it and it said `PRIVATE_GITHUB_TOKEN`, but I think `GITHUB_TOKEN` is better so I'll change this 👍

docs/pages/stacks.md
```diff
@@ -0,0 +1,84 @@
...
+
+#### `remix.init.js`
+
+If the template has a `remix.init.js` file at the root then that file will be executed after the project has been generated but before dependencies have been installed. This gives you a chance to do anything you'd like as part of the initialization of your template. For example, in the blues stack, the `app` property has to be globally unique so we use the `remix.init.js` file to change it to the name of the directory that was created for the project + a couple random characters.
```
kentcdodds, line 73: > will be executed after the project has been generated but before dependencies have been installed.

Is this correct @mcansh?

kentcdodds (commented):
One other question

docs/pages/stacks.md
mcansh: 100% correct :) https://github.com/remix-run/remix/blob/040fba47395c16b236655f5a6a1d889086b93e65/packages/create-remix/index.ts#L133-L143

docs/pages/stacks.md
mcansh: it has to be after, just in case they have dependencies in their init script - also! Michael and I decided to remove the `remix.init.js` and only support `remix.init/index.js`

docs/pages/stacks.md
mcansh: besides the name, that is correct

docs/pages/stacks.md
kentcdodds: Oh nice, so it's not just remix packages, but any package?

docs/pages/stacks.md
kentcdodds: That sounds good 👍

kentcdodds:
Thanks for the feedback @mcansh!

docs/pages/stacks.md
```diff
@@ -0,0 +1,84 @@
...
+
+The Remix CLI will help you get started with one of these built-in stacks, but if you want, you can create your own stack and the Remix CLI will help you get started with that stack. There are several ways to do this, but the most straightforward is to create a GitHub repo:
+
+```
+npx create-remix@latest --template my-username/my-repo
+```
+
+Custom stacks give an enormous amount of power and flexibility and we hope you create your own that suites the preferences of you and your organization (feel free to fork ours!).
+
+### `--template`
```
MichaelDeBoey, line 39: I think the `--template` flag can be confusing, as we already have basic templates.

Since we're not supporting to choose the basic templates with this flag, it's maybe a good idea to either rename it to `--stack` **or** support choosing basic templates as well? 🤔

mcansh: the current templates will behave exactly the same for the time being where not passing a `--template` flag brings you to the same questionnaire it does today (suppose we could've mentioned that), you can also use `--template arc` to get the current Architect template

MichaelDeBoey: @mcansh That's awesome!

I would indeed mention that in the docs, as it seemed confusing to me (and I think it will for other people as well)

kentcdodds: I'm not sure I understand the confusion. Could you elaborate (or make a PR)? :)

docs/pages/stacks.md
```diff
@@ -0,0 +1,84 @@
...
+```diff
+-   "remix": "*",
++   "remix": "^1.2.3",
+```
+
+This allows you to not have to regularly update your template to the latest version of that specific package. Of course you do not have to put `*` if you'd prefer to manually manage the version for that package.
+
+#### `remix.init/index.js`
+
+If the template has a `remix.init.js` file at the root then that file will be executed after the project has been generated and dependencies have been installed. This gives you a chance to do anything you'd like as part of the initialization of your template. For example, in the blues stack, the `app` property has to be globally unique so we use the `remix.init/index.js` file to change it to the name of the directory that was created for the project + a couple random characters.
```
mjackson, line 73: We aren't supporting `remix.init.js`. Just `remix.init/index.js`. This will keep init script dependencies out of the main app's `package.json`.

kentcdodds: Will fix 👍

Issue #2215, opened 2022-03-05
npm package eslint-config does not exist under owner remix-run


TomVance:
### What version of Remix are you using?

1.2.3

### Steps to Reproduce

- Run `npm install @remix-run/eslint-config` or `yarn add @remix-run/eslint-config` in any remix based project

OR

- Create a new remix project with `npx create-remix@latest` and run `npm install` 

### Expected Behavior

To be able to create a new rmix project with the default eslint config enabled

### Actual Behavior

Dependency installs fail with the following error:

NPM:
```
npm install @remix-run/eslint-config
npm ERR! code ENOVERSIONS
npm ERR! No versions available for undefined

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/thomasvance/.npm/_logs/2022-03-05T07_11_16_882Z-debug.log
```

YARN:
```
yarn add @remix-run/eslint-config
yarn add v1.22.10
info No lockfile found.
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
warning @remix-run/dev > @esbuild-plugins/node-modules-polyfill > rollup-plugin-node-polyfills > rollup-plugin-inject@3.0.2: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-inject.
error An unexpected error occurred: "https://npm.remix.run/@remix-run%2feslint-config: npm package \"eslint-config\" does not exist under owner \"remix-run\"".
info If you think this is a bug, please open a bug report with the information provided in "/Users/thomasvance/Developer/tech-assesments/true-layer-merchant-x/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command
```

TomVance:
This appears to have been connected to me still using the private npm registry from the initial remix private days.

Resolved by removing the settings in my root `.npmrc` file

Issue #2202, opened 2022-03-04
Unable to create Remix project using Vercel as provider


ajay-paramlogic:
### What version of Remix are you using?

1.2.3

### Steps to Reproduce

 npx create-remix@latest
 Choose Vercel
 Choose TypeScript
 Choose Yes

### Expected Behavior

The project should get created and all the dependencies should be installed

### Actual Behavior

```sh
➜  cmca npx create-remix@latest


R E M I X - v1.2.3

💿 Welcome to Remix! Let's get you set up with a new project.

? Where would you like to create your app? ./my-remix-app
? Where do you want to deploy? Choose Remix if you're unsure, it's easy to change deployment targets. Vercel
? TypeScript or JavaScript? TypeScript
? Do you want me to run `npm install`? Yes
npm ERR! code ENOVERSIONS
npm ERR! No versions available for undefined

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/mystical/.npm/_logs/2022-03-04T05_34_00_052Z-debug.log
Error: Command failed: npm install
    at checkExecSyncError (node:child_process:826:11)
    at Object.execSync (node:child_process:900:15)
    at Object.createApp (/Users/mystical/.npm/_npx/5164864a48bff686/node_modules/create-remix/index.js:135:19)
    at async run (/Users/mystical/.npm/_npx/5164864a48bff686/node_modules/create-remix/cli.js:199:5) {
  status: 1,
  signal: null,
  output: [ null, null, null ],
  pid: 45737,
  stdout: null,
  stderr: null
}
```

girishk21:
> To the maintainers

It's a problem with the `"@remix-run/eslint-config"` package. It throws the error:

```sh
npm ERR! code ENOVERSIONS
npm ERR! No versions available for undefined
```

@ajay-paramlogic can you remove `"@remix-run/eslint-config"` from your `package.json` and as well as the entry from `.eslintrc.js`, and run

```sh
npm i
```

it should work as expected in the mean time the team fixes the eslint package issue

ajay-paramlogic:
> > To the maintainers
> 
> It's a problem with the `"@remix-run/eslint-config"` package. It throws the error:
> 
> ```shell
> npm ERR! code ENOVERSIONS
> npm ERR! No versions available for undefined
> ```
> 
> @ajay-paramlogic can you remove `"@remix-run/eslint-config"` from your `package.json` and as well as the entry from `.eslintrc.js`, and run
> 
> ```shell
[...]

PR #2290, opened 2022-03-11
fix: check for private registry config

ascorbic:
This PR adds a check for the old private registry config. If it's still configured, it causes errors such as those reported in #2202 and #2215 and others I've seen in Discord.

I can't find any docs on migrating from the private beta, but there could be a link to these if they exist.


MichaelDeBoey (approved):

mcansh (approved):
love it, thanks! gonna wait for @mjackson, @ryanflorence, or @kentcdodds  to give the final say

kentcdodds:
Thanks!

PR #2291, opened 2022-03-11
fix(netlify): use Netlify internal functions dir

ascorbic:
This PR changes the Netlify template, and the default `serverBuildPath` to use the internal functions directory. This is preferred for frameworks because it is in a directory that is already gitignored, and does not interfere with the user's own functions. The redirects do not need changing, because the URL is unchanged.


kentcdodds (approved):
LGTM! Thanks 👍

PR #2043, opened 2022-02-19
* add a nonce prop to <LiveReload/> and <ScrollRestoration/>

axel-habermaier:
The `<Scripts/>` component already supports a `nonce` prop for nonce-based CSPs, however, `<LiveReload/>` and `<ScrollRestoration/>` do not. This PR adds the missing prop to both components.

I'm not sure if this needs any tests. If so, let me know.

PR #2292, opened 2022-03-11
add nonce prop to LiveReload and ScrollRestoration

dusty:
This is a duplicate of https://github.com/remix-run/remix/pull/2043, which had merge conflicts.

Adds the nonce prop to LiveReload and ScrollRestoration so one can use a Content Security Policy

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -62,6 +62,7 @@
 - donavon
 - Dueen
 - dunglas
+- dusty
 - dwt47
 - eastlondoner
 - edgesoft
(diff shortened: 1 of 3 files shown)


dusty:
What do you guys think about this? Any possibility of getting this added?

machour:
Hi @dusty, usually when a PR go stale, its considered a good practice to nudge its author instead of opening a duplicate 😬 
The original PR have been updated, and will get some attention from the team shortly.

dusty:
Thanks!

PR #2297, opened 2022-03-11
docs(examples): bring `stripe-integration` example in line with templates

MichaelDeBoey:
Like I did in #2200 & #2280

Follow-up of #2035


kentcdodds (approved):
Nice! Thanks 👍

PR #2309, opened 2022-03-13
test: add test where `form` is descendant of element with stop propagation on click

penspinner:
**Bug**

Added a failing test where clicking a clicking a submit button as a descendant of an element that stops propagation on click does not send the clicked submit button's `name` and `value` properties to the action request payload.

I think this bug was introduced from https://github.com/remix-run/remix/pull/1781 (released in [v1.2.0](https://github.com/remix-run/remix/releases/tag/v1.2.0)) because this used to work until I updated dependencies at some point.

---

**How I encountered this**

I encountered this bug through my own app. I had a modal with a form in it, and it suddenly stopped sending `{ _action: add_note }` to my action's request payload. My code looks something like this:

```tsx
import { Dialog } from '@headlessui/react'
import { Form } from 'remix'

const Page = () => {
  return (
    <Dialog>
      <Form method="post">
        <button name="_action" value="add_note">
          Add note
        </button
      </Form
    </Dialog>
  )
}

export default Page
```

And I noticed Headless UI's Dialog has an `event.stopPropagation()` on the div's `onClick`: https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/dialog/dialog.tsx#L303.


Penspinner:
Related Discord thread with @ryanflorence: https://discord.com/channels/770287896669978684/950943922966900756.

integration/bug-report-test.ts
```diff
@@ -37,28 +37,25 @@ beforeAll(async () => {
...
-          let data = useLoaderData();
+          let actionData = useActionData();
           return (
-            <div>
-              {data}
-              <Link to="/burgers">Other Route</Link>
+            <div onClick={(event) => event.stopPropagation()}>
+              <pre>{JSON.stringify(actionData)}</pre>
+              <Form action="/?index" method="post">
+                <button type="submit" name="_action" value="add">Add</button>
```
MichaelDeBoey, line 53: `name` can just be `action` instead

See f47a5114bb1a208f8316d1cc86eaee25699e7457 & #2164

```suggestion
                <button type="submit" name="action" value="add">Add</button>
```

integration/bug-report-test.ts
```diff
@@ -73,15 +70,15 @@ afterAll(async () => app.close());
...
-  expect(await response.text()).toMatch("pizza");
+  // let response = await fixture.requestDocument("/");
+  // expect(await response.text()).toMatch("pizza");
 
   // If you need to test interactivity use the `app`
   await app.goto("/");
-  await app.clickLink("/burgers");
-  expect(await app.getHtml()).toMatch("cheeseburger");
+  await app.clickSubmitButton("/?index");
+  expect(await app.getHtml()).toMatch('{"_action":"add"}');
```
MichaelDeBoey, line 81:
```suggestion
  expect(await app.getHtml()).toMatch('{"action":"add"}');
```

MichaelDeBoey (changes requested):

mcansh (approved):

penspinner:
@mcansh Thanks! Another duplicated PR for this bug: https://github.com/remix-run/remix/pull/2034.

PR #2311, opened 2022-03-14
feat(remix-server-runtime): export argument types for functions that can be exported from routes

mbrowne:
This PR adds and exports some types from `@remix-run/server-runtime` and makes them available in the `remix` package as well.

Many developers and teams (myself included) prefer to use `function` syntax rather than `const` syntax for top-level functions, e.g.:

```ts
export async function loader({ request }: DataFunctionArgs) {
```

Exporting argument types from the `remix` package makes for a friendlier development experience for those who prefer this coding style.

---

- [ ] Docs
- [x] Tests

Note: I wasn't able to verify this locally, despite adding the following in the `package.json` of a demo app:

```
    "@remix-run/react": "file:/usr/local/www/remix/build/node_modules/@remix-run/react",
    "@remix-run/serve": "file:/usr/local/www/remix/build/node_modules/@remix-run/serve",
    "remix": "file:/usr/local/www/remix/build/node_modules/remix",
```

...so please let me know if `magicExports/remix.ts` isn't the right place to add these exports.


MichaelDeBoey:
@mbrowne We recently changed a lot of internal packages.
Could you please rebase your branch onto latest `dev`?

mbrowne:
Sure. Should be up-to-date now.

packages/remix-server-runtime/magicExports/remix.ts
```diff
@@ -19,13 +19,17 @@ export type {
   HtmlLinkDescriptor,
   PageLinkDescriptor,
   ErrorBoundaryComponent,
+  ErrorBoundaryProps,
```
MichaelDeBoey, line 22: @mbrowne I suppose you want these to be available from `@remix-run/cloudflare` & `@remix-run/node` too?
We're moving away from using `magicExports` & `@remix-run/server-runtime` isn't seen as a public API (see #2359)

MichaelDeBoey (changes requested):

packages/remix-server-runtime/magicExports/remix.ts
mbrowne: No problem and thanks for the link. I added the exports to `@remix-run/cloudflare` and `@remix-run/node`. Let me know if I should also remove them from `remix-server-runtime/magicExports/remix.ts` since those magic exports will be going away at some point.

MichaelDeBoey: We currently keep the `magicExports` to be backwards compatible.
We'll remove them in v2 though.

packages/remix-cloudflare/index.ts
```diff
@@ -31,16 +31,19 @@ export type {
   DataFunctionArgs,
   EntryContext,
   ErrorBoundaryComponent,
+  ErrorBoundaryProps,
```
MichaelDeBoey, line 34: @mbrowne I think we should add these to the `replace-remix-imports` migration too.

https://github.com/remix-run/remix/blob/15bf240cddf871494522e4c917fa17400b89dda5/packages/remix-dev/cli/migrate/run/transforms/replace-remix-imports/jscodeshift-transform/map-normalized-imports/package-exports.ts#L27

Make sure to rebase on latest `dev` to be able to access this file

MichaelDeBoey (changes requested):

packages/remix-cloudflare/index.ts
mbrowne: I'm happy to add this, but I just wanted to double-check that it's necessary? It says `runtime` exports and these are just types that aren't needed at run-time. Please confirm and then I can add them if needed.

mbrowne: Just checking in on this...should I go ahead and add these to the `replace-remix-imports` migration? ...given that these are types used only at compile-time

machour:
Hey @mbrowne and thank you so much for this contribution.

Things changed a bit since you've opened this PR, especially in [1.6.5](https://github.com/remix-run/remix/releases/tag/remix%401.6.5) where we added type inference.

I'm going to close this PR, please submit a discussion/proposal if you still think something is missing 🙏🏼

PR #2312, opened 2022-03-14
chore: fix ESLint errors

MichaelDeBoey:
https://github.com/remix-run/remix/runs/5545560283?check_suite_focus=true#step:6:1441

---

Follow-up of #2189

CC/ @mcansh


kentcdodds (approved):
Thanks!

kentcdodds:
I'm thinking we shouldn't use ESLint on deno code. They have their own linter and things are a bit different in that runtime.

PR #2196, opened 2022-03-03
docs(examples): remove `devServerPort` from `remix.config.js`

MichaelDeBoey:
As suggested by @kentcdodds in https://github.com/remix-run/remix/pull/2158#discussion_r818948749

PR #2314, opened 2022-03-14
fix(templates): bring in line with each other

MichaelDeBoey:
- Add `.env` to `.gitignore`
- Remove `devServerPort` from `remix.config.js` (like I did in #2196)
- Move `cross-env` to `dependencies` whenever it's used in the `start` script
- Move `npm-run-all` to `devDependencies` whenever it's only used in `dev` script

---

Follow-up of #2189

CC/ @mcansh


mcansh (approved):
thanks 🙏

PR #2317, opened 2022-03-15
feat(remix-dev): add option to remove typescript types during project creation

mcansh:
Closes: REM-830

- [ ] Docs
- [x] Tests


packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -236,12 +236,23 @@ describe("remix cli", () => {
...
+      expect(fs.existsSync(path.join(projectDir, "app/root.tsx"))).toBeFalsy();
+      expect(
+        fs.existsSync(path.join(projectDir, "jsconfig.json"))
+      ).toBeTruthy();
+      expect(fs.existsSync(path.join(projectDir, "tsconfig.json"))).toBeFalsy();
+      let pkgJSON = JSON.parse(
+        fs.readFileSync(path.join(projectDir, "package.json"), "utf8")
+      );
+      expect(pkgJSON.scripts.typecheck).toBeUndefined();
+      expect(pkgJSON.devDependencies).not.toContain("typescript");
```
mjackson, line 255: I believe `toContain` only works on arrays, and possibly strings. I think you'll want to use `Object.keys(pkgJSON.devDependencies)` here.

mcansh: yup, you're right https://jestjs.io/docs/expect#tocontainitem

mcansh:
> @mcansh It seems like something went wrong with the `examples` 🤔

I tried to just copy the latest dev ones over, I'll reset and cherry-pick over the commits that actually matter in a sec

templates/arc/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "remix-template-arc",
+  "name": "remix-template-arc-ts",
```
MichaelDeBoey, line 2: It's probably a good idea to run the `patchup-example-names` script?

```suggestion
  "name": "remix-template-arc",
```

MichaelDeBoey (changes requested):

packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -134,6 +135,11 @@ describe("remix cli", () => {
   });
 
   describe("the create command", () => {
+    beforeAll(async () => {
+      await fsp.rm(TEMP_DIR, { force: true, recursive: true });
+      await fsp.mkdir(TEMP_DIR, { recursive: true });
```
mjackson, line 140:
```suggestion
      await fse.emptyDir(TEMP_DIR);
```

For stuff like this I love using [fs-extra](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/emptyDir.md). We should probably just default to using that everywhere instead of fs and fs/promises.

packages/remix-dev/cli/create.ts
```diff
@@ -451,3 +452,80 @@ async function detectTemplateType(
   }
   return "remoteTarball";
 }
+
+function untype(filename: string, source: string, projectDir: string): string {
```
mjackson, line 456: Let's call this function `convertToJavaScript`

packages/remix-dev/cli/create.ts
```diff
@@ -451,3 +452,80 @@ async function detectTemplateType(
...
+
+  if (!result || !result.code) {
+    throw new Error("Could not parse typescript");
+  }
+
+  /*
+    Babel's `compact` and `retainLines` options are both bad at formatting code.
+    Use Prettier for nicer formatting.
+  */
+  return prettier.format(result.code, { parser: "babel-ts" });
```
mjackson, line 474: We shouldn't need `babel-ts` here now that it's JavaScript and not TypeScript, right?

integration/helpers/global-setup.ts
```diff
@@ -13,5 +13,5 @@ export default async function setup(globalConfig: any) {
     force: true,
     recursive: true,
   });
-  await fs.mkdir(TMP_DIR);
+  await fs.mkdir(TMP_DIR, { recursive: true });
```
mjackson, line 16: Let's go ahead and use `fs-extra` here too.

mjackson (approved):

PR #2322, opened 2022-03-15
[@remix-run/remix-eslint-config] Include jest-testing-library.js

mennopruijssers:
Right now `jest-testing-library.js` isn't included when publishing to NPM while this should be the case.
This PR fixes it by adding it to the `files` array in `package.json`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -182,6 +182,7 @@
 - medayz
 - meetbryce
 - mehulmpt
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
CC/ @chaance

mennopruijssers:
I just rebased on latest `main`!

PR #2330, opened 2022-03-15
fix(create-remix): prevent npx querying to install remix dev

kentcdodds:


mcansh (approved):

mcansh:
i was looking for this haha, thanks!

mjackson:
@kentcdodds Which versions of npm did you test? I'd like to make sure we test at least 6, 7 and 8 since we currently support all the way down to node 14. IIRC I had to do the `--yes` flag differently in npm 6.

kentcdodds:
> IIRC I had to do the --yes flag differently in npm 6.

Ah, yeah, hmmm... We'll have to do a version check before providing the flag and users of old npm can just get the question. I'll work on that 👍 We can get this in next week.

kentcdodds:
Just fixed this for old versions of npx:

image

Turns out npx v6 doesn't ask anyway so it's fine :)

kentcdodds:
Oh, and here it is with a modern version of npx:

image

Issue #1717, opened 2022-01-29
[Feature] Add fb:app_id and fb:page_id to Meta component 


gixxerblade:
### What is the new or updated feature that you are suggesting?

Request add Facebook's open graph markup `fb:page_id` and `fb:app_id` to the Meta component.

### Why should this feature be included?

 When a link is shared to Facebook, the crawler indexes the fb:page_id and fb:app_id if they are present. According to a [Guide to Sharing for Webmasters](https://developers.facebook.com/docs/sharing/webmasters#markup) on Facebook's site, they recommend the use of `fb:app_id`

fb:app_id | In order to use Facebook Insights you must add the app ID to your page. Insights lets you view analytics for traffic to your site from Facebook. Find the app ID in your App Dashboard.
-- | --

kiliman:
It doesn't look like there's consistency in how `<meta>` tags are used. The HTML standard is to use `<meta name="" content=""/>`. But apparently there's an RDFa standard that allows `<meta property="" content=""/>`.

The `og:*` meta properties, use the `property` attribute, same with `fb:*`. I thought that a good heuristic would be to look for `:` in the property name and if present use `property`, otherwise use `name`. Unfortunately that doesn't work, as `twitter:*` meta tags use `name`. 😮‍💨

Currently the Remix `<Meta>` component is hard-coded to check for `og:` prefix. Every time a new meta property came out, they would need to update this.

It may be better to allow you to specify this in the `meta` export.

```ts
export function meta() {
  return {
    title: 'My title',
    description: 'My description',
    'og:image': 'path/to/image',    // keep for backward compatibility
    'fb:app_id': { property: 'fb:app_id', content: '1234' },  // use object for non-name version
    '@fb:page_id': 'page-id',       // or use a prefix like @ to denote this is 'property'
  }
}
```

Ethaan:
So i guess for now this is what we need to do right?

```
<head>
  <meta charSet="utf8" />
  <meta name="viewport" content="width=device-width,initial-scale=1" />
  <meta property="fb:app_id" content="xxxx" />
  <Meta />
  <Links />
</head>
```

kiliman:
Yes, for now you would need to manually render those <meta> elements yourself.

14850842:
This however does not allow you to specify dynamic values such as your deep link:

```
 <meta property="al:ios:url" content="example://applinks/:id" />
 ```

machour:
@14850842 a PR which should allow that to work was recently merged:
[...]

PR #2331, opened 2022-03-15
feat(remix-react): add support for `http-equiv` in `meta` function

### LINKED ISSUE #941: [Feature]: Support meta http-equiv tag

I would for like Remix to allow me to add a [`<meta http-equiv>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-http-equiv) tag to my page via the meta export.

I would like for the meta export to support `http-equiv` because there doesn't appear to be any other way to add a `<meta http-equiv>` tag a single route.

If this change sounds good, I'm happy to put together a PR adding a special case for http-equiv. I could add [`charset`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-charset) too for good measure if that's desired.

MichaelDeBoey:
Took inspiration of the OG tags: if people want to use the `http-equiv` tag, they'll have to do

```ts
export const meta: MetaFunction = () => ({
  'http-equiv:refresh': '3;url=https://www.mozilla.org',
});
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-http-equiv

@kentcdodds Not sure how this would best fit in the docs, so I'll leave that up to you if you don't mind.

---

Closes #941

### DIFF
--- a/integration/meta-test.ts
+++ b/integration/meta-test.ts
@@ -10,20 +10,24 @@ describe("meta", () => {
       files: {
         "app/root.jsx": js`
           import { json, Meta, Links, Outlet, Scripts } from "remix";
-          
+
           export const loader = async () =>
             json({
               description: "This is a meta page",
               title: "Meta Page",
             });
-          
+
           export const meta = ({ data }) => ({
+            charset: "utf-8",
             description: data.description,
             "og:image": "https://picsum.photos/200/200",
             "og:type": data.contentType, // undefined
+            refresh: {
+              httpEquiv: "refresh",
+              content: "3;url=https://www.mozilla.org",
+            },
             title: data.title,
           });
-
 
           export default function Root() {
             return (
@@ -52,8 +56,26 @@ describe("meta", () => {
     app = await createAppFixture(fixture);
   });
 
-  afterAll(async () => {
-    await app.close();
+  afterAll(async () => await app.close());
+
+  test("empty meta does not render a tag", async () => {
+    let enableJavaScript = await app.disableJavaScript();
+
+    await app.goto("/");
+
+    await expect(app.getHtml('meta[property="og:type"]')).rejects.toThrowError(
+      'No element matches selector "meta[property="og:type"]"'
+    );
+    await enableJavaScript();
+  });
+
+  test("meta { charset } adds a <meta charset='utf-8' />", async () => {
+    let enableJavaScript = await app.disableJavaScript();
+
+    await app.goto("/");
+
+    expect(await app.getHtml('meta[charset="utf-8"]')).toBeTruthy();
+    await enableJavaScript();
   });
 
   test("meta { title } adds a <title />", async () => {
@@ -61,20 +83,7 @@ describe("meta", () => {
 
     await app.goto("/");
 
-    expect(await app.getHtml("title")).toEqual(
-      expect.stringMatching(/<title>/s)
-    );
-    await enableJavaScript();
-  });
-
-  test("meta { description } adds a <meta name='description' />", async () => {
-    let enableJavaScript = await app.disableJavaScript();
-
-    await app.goto("/");
-
-    expect(await app.getHtml("meta")).toEqual(
-      expect.stringMatching(/<meta\s+name="description"/s)
-    );
+    expect(await app.getHtml("title")).toBeTruthy();
     await enableJavaScript();
   });
 
@@ -83,20 +92,29 @@ describe("meta", () => {
 
     await app.goto("/");
 
-    expect(await app.getHtml("meta")).toEqual(
-      expect.stringMatching(/<meta\s+property="og:*/s)
-    );
+    expect(await app.getHtml('meta[property="og:image"]')).toBeTruthy();
     await enableJavaScript();
   });
 
-  test("empty meta does not render a tag", async () => {
+  test("meta { description } adds a <meta name='description' />", async () => {
     let enableJavaScript = await app.disableJavaScript();
 
     await app.goto("/");
 
-    expect(await app.getHtml("meta")).not.toEqual(
-      expect.stringMatching(/<meta\s+property="og:type/s)
-    );
+    expect(await app.getHtml('meta[name="description"]')).toBeTruthy();
+    await enableJavaScript();
+  });
+
+  test("meta { refresh } adds a <meta http-equiv='refresh' content='3;url=https://www.mozilla.org' />", async () => {
+    let enableJavaScript = await app.disableJavaScript();
+
+    await app.goto("/");
+
+    expect(
+      await app.getHtml(
+        'meta[http-equiv="refresh"][content="3;url=https://www.mozilla.org"]'
+      )
+    ).toBeTruthy();
     await enableJavaScript();
   });
 });
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -663,28 +663,38 @@ export function Meta() {
   return (
     <>
       {Object.entries(meta).map(([name, value]) => {
-        if (!value) return null;
+        if (!value) {
+          return null;
+        }
+
+        if (["charset", "charSet"].includes(name)) {
+          return <meta key="charset" charSet={value as string} />;
+        }
+
+        if (name === "title") {
+          return <title key="title">{value}</title>;
+        }
 
         // Open Graph tags use the `property` attribute, while other meta tags
         // use `name`. See https://ogp.me/
         let isOpenGraphTag = name.startsWith("og:");
-        return name === "title" ? (
-          <title key="title">{value}</title>
-        ) : ["charset", "charSet"].includes(name) ? (
-          <meta key="charset" charSet={value as string} />
-        ) : Array.isArray(value) ? (
-          value.map((content) =>
-            isOpenGraphTag ? (
-              <meta key={name + content} property={name} content={content} />
-            ) : (
-              <meta key={name + content} name={name} content={content} />
-            )
-          )
-        ) : isOpenGraphTag ? (
-          <meta key={name} property={name} content={value} />
-        ) : (
-          <meta key={name} name={name} content={value} />
-        );
+        return [value].flatMap((content) => {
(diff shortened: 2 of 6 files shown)


kentcdodds:
I wonder... Are there a lot of unique tags like this? Should we maybe allow people to have a different way to define these for more control?

MichaelDeBoey:
It seems like those are the only ones

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

kentcdodds (approved):
Looks good to me 👍

kiliman:
Remix currently special-cases `og:*` prefix to set the `property` attribute vs `name`.

However, `fb:*` prefix also uses `property`, but `twitter:*` uses `name`. Who needs consistency?! 🤦‍♂️

I discuss in this issue. https://github.com/remix-run/remix/issues/1717

machour:
@kentcdodds it's not so meaningless, as it allows to override a specific item in the merged metas:

// parent.tsx
```
export const meta: MetaFunction = () => {
  return {
    author1: {name: "author", content: "Author One"},
    author2: {name: "author", content: "Author Two"},
  };
};
```

// parent/child.tsx

```
export const meta: MetaFunction = () => {
  return {
    author2: {name: "author", content: "Another Author"},
  };
};
```

in the child route, this will be rendered: 

```
<meta name="author" content="Author One" />
<meta name="author" content="Another Author" /> <-- took the place of Author Two
```

(the specification allow some meta to appear more than once)

kentcdodds:
Oh nice! Yeah, I'm a fan 👍

kentcdodds (approved):
Love it 👍

Would you mind adding some docs for this?

packages/remix-react/components.tsx
```diff
@@ -679,9 +680,11 @@ export function Meta() {
             )
           )
         ) : isOpenGraphTag ? (
-          <meta key={name} property={name} content={value} />
+          <meta property={name} content={value as string} />
+        ) : typeof value === "string" ? (
+          <meta name={name} content={value} />
         ) : (
-          <meta key={name} name={name} content={value} />
+          <meta {...value} />
```
machour, line 687: We still need keys on all elements, as we're in a loop.

Just a question on code readability: wouldn't it be better to break that giant ternary operation into something like:

```js
if (["charset", "charSet"].includes(name)) {
  return <meta key="charset" charSet={value} />
}

if (name === "title") { 
 // ...
```

kentcdodds: I think if I were going to write this code from scratch I would do the if/else myself.

packages/remix-server-runtime/routeModules.ts
```diff
@@ -92,7 +92,7 @@ export interface MetaFunction {
  * `name` attribute.
  */
 export interface HtmlMetaDescriptor {
-  [name: string]: string | string[];
+  [name: string]: object | string | string[];
```
machour, line 95:
```suggestion
  [name: string]: Record<string, string> | string | string[];
```

packages/remix-react/components.tsx
MichaelDeBoey: > We still need keys on all elements, as we're in a loop.

@machour This part isn't in the loop, so we can omit the `key` prop

> I think if I were going to write this code from scratch I would do the if/else myself.

@kentcdodds I'll refactor it to use the separate `if`-statements

machour (approved):

machour:
Still needs documentation update (and ideally tests)

packages/remix-react/components.tsx
```diff
@@ -666,22 +666,31 @@ export function Meta() {
...
+
+        if (["charset", "charSet"].includes(name)) {
+          return <meta key="charset" charSet={value as string} />;
+        }
+
+        if (name === "title") {
+          return <title key="title">{value}</title>;
+        }
+
+        if (Array.isArray(value)) {
```
kentcdodds, line 678: I think the array case should probably get into recursion for this logic rather than duplicating just the `isOpenGraphTag` logic. Maybe we should extract this logic into a function to make that possible.

packages/remix-react/components.tsx
```diff
@@ -666,22 +666,31 @@ export function Meta() {
...
-        ) : (
+          );
+        }
+
+        return isOpenGraphTag ? (
+          <meta key={name} property={name} content={value as string} />
+        ) : typeof value === "string" ? (
           <meta key={name} name={name} content={value} />
+        ) : (
+          <meta key={name} {...value} />
```
kentcdodds, line 693: Let's be consistent and keep the whole thing `if/else`

kentcdodds (changes requested):
This is looking great. I just have some thoughts.

packages/remix-react/components.tsx
kentcdodds: Based on the code I think it was already and I seem to recall this is documented somewhere.

MichaelDeBoey: I only enabled objects, not array of objects.

MichaelDeBoey:
@machour I think it's a problem on `dev`, as this only happened after rebasing & `dev` fails too

https://github.com/remix-run/remix/runs/5580333691

MichaelDeBoey:
@machour It seems like https://github.com/remix-run/remix/commit/32335011e2fc0b77d9ec26471f14f9a458814323 is passing, but a rebase on `dev` (https://github.com/remix-run/remix/commit/e9ef8abf8b02851425ab7486ea3cd597fe5b584f) seems to be failing 🤔

MichaelDeBoey:
> I suspect that the problems is with the flat map part

@machour Even with the `flatMap` call, it passed a couple of commits before current `dev` HEAD

> I don't see why we're expecting an eventual array there anyways

The `value` is either `string` or `object` **or** an array of `string`s/`object`s.
To not have to do the same thing twice (once in the loop and once outside of it, like we did before), I first wrapped the `value` in an array. Since we can now have nested arrays (if the `value` was already an array), we flatten the array.
This way we're **always** having an array of `string`s/`object`s and can do the `if`/`else` only once.

MichaelDeBoey:
@machour I first only added support for non-array objects, but @kentcdodds asked me to add support for object arrays too, so it's more consistent.

machour:
Gotcha. I wasn't aware that arrays were supported before this PR. Thanks for baring with me 🙏

MichaelDeBoey:
@kentcdodds @machour I also added some tests now.
I also made the types a bit stricter for `charset`, `charSet` & `title` as those were only supported anyways.

kentcdodds:
I think we should be able to get this merged as soon as we get CI happy 👍

MichaelDeBoey:
> get CI happy

That's where I could use some help I think 😅
I don't really see the problem in my code changes... 🤔😕

integration/meta-test.ts
```diff
@@ -52,29 +56,34 @@ describe("meta", () => {
...
+  test("empty meta does not render a tag", async () => {
     let enableJavaScript = await app.disableJavaScript();
 
     await app.goto("/");
 
-    expect(await app.getHtml("title")).toEqual(
-      expect.stringMatching(/<title>/s)
+    expect(await app.getHtml("meta")).not.toEqual(
+      expect.stringMatching(/<meta\s+property="og:type/s)
     );
```
mcansh, line 68: this works, but seems overly verbose

```suggestion
    await expect(app.getHtml('meta[property="og:type"]')).rejects.toThrowError(
      "No element matches selector \"meta[property=\"og:type\"]\""
    );
```

mcansh (approved):

kentcdodds (approved):
This is a great change. Thank you for this PR!

packages/remix-react/routeModules.ts
```diff
@@ -71,7 +71,16 @@ export interface MetaFunction {
  * `name` attribute.
  */
 export interface HtmlMetaDescriptor {
-  [name: string]: string | string[] | null | undefined;
+  charset?: "utf-8";
+  charSet?: "utf-8";
```
machour, line 75: @MichaelDeBoey A little late to the party here, but unless I'm wrong, we shouldn't force `utf-8` here.
`windows-1252` is valid for example.

machour: 👍🏼  (I was looking at an older version of the HTML spec.)

kiliman:
@kentcdodds Any chance we could look at making `meta` allow returning an array like `links`. If so, it would simplify meta handling and we wouldn't need this weird object syntax.

See PR #2497 where we discuss this. Still need to create a PR for implementing array version, but if you're open to it, I'd gladly work on it.

I think it would be nice to have similar API for both `meta` and `links`. If this is supported, even deprecate the current object syntax in favor of array of objects.
```ts
export const meta = ({data}) => [
  { title: 'my title' },
  { charset: "utf-8" },
  { property: "og:xxx", content: "url" },
  { property: "article:published_time", content: "..." },
  { name: "description", content: "..." },
  { name: "twitter:card", content: "..." },
  { property: "fb:app_id", content: "..." },
  { httpEquiv: "refresh", content: "..." },
  ...
]
```

donavon:
I also would like to see:

*  #2497 merged
*  Array support for meta (with the current object syntax)
*  This PR #2331 reverted

I don't think we should deprecate object syntax as it's simpler. Array syntax would be your escape hatch if you need complete control.

kentcdodds:
The benefit of the object syntax here is when we merge all the routes it can be used to override the meta values.

I'm fine with adding array syntax. I'm not sure why we need to revert #2331 though.

kentcdodds:
As noted, the name is not meaningless, which is why I agreed and merged this PR: https://github.com/remix-run/remix/pull/2331#issuecomment-1069420261

My bad on merging it before getting Michael and Ryan's approval (I thought I had, but obvious I hadn't).

But I think it's important to consider the override case which is why I'd say we should not deprecate the object version and *only* do this with arrays in the future. I think both make sense.

kiliman:
I was waiting for the other PR to land before adding the Array variant. I should have a PR ready soon.

PR #2333, opened 2022-03-15
chore: update example and template package.json names to resolve 'jest-haste-map' warning

mcansh:
Closes: #

- [ ] Docs
- [ ] Tests


MichaelDeBoey:
I should have written this script when I was doing #2200, It would have helped me a lot 😅

mcansh:
> I should have written this script when I was doing #2200, It would have helped me a lot 😅

that PR 🤯, it sure woulda helped haha

templates/deno-ts/package.json
```diff
@@ -1,20 +1,20 @@
...
+  },
+  "devDependencies": {
+    "@remix-run/dev": "*",
+    "cross-env": "^7.0.3",
+    "npm-run-all": "^4.1.5"
+  },
+  "engines": {
+    "node": ">=14"
+  },
+  "name": "remix-template-deno-ts"
```
MichaelDeBoey, line 19: @mcansh I think running `npx sort-package-json` on this file would be a good idea.
This way the file in our repo is the same as the file you get when running `npx create-remix@latest`

MichaelDeBoey (changes requested):

templates/deno-ts/package.json
mcansh: we do this during create-remix fwiw, but might not be a bad idea to get them all cleaned up before that

mjackson:
This PR is great. But is there any way we can just tell Jest to ignore the examples and templates directories?

mcansh:
> This PR is great. But is there any way we can just tell Jest to ignore the examples and templates directories?

you're supposed to be able to ignore them via `modulePathIgnorePatterns` but that doesnt appear to work facebook/jest#8114

mcansh:
in theory we could just disable the warning 😅, but im not sure that's the **right** fix
```js
haste: {
	throwOnModuleCollision: false,
}
```

mjackson:
It's ok, if Jest isn't going to allow us to ignore examples and templates then I guess we can just script our way around it.

scripts/patchup-example-names.mjs
```diff
@@ -1,17 +1,28 @@
...
-let dirPath = path.join(process.cwd(), "examples");
-let dir = await fs.readdir(dirPath);
+let examplePath = path.join(process.cwd(), "examples");
+let templatePath = path.join(process.cwd(), "templates");
+let examples = await fs.readdir(examplePath);
+let templates = await fs.readdir(templatePath);
 
-dir.forEach(async (name) => {
-  let fullPath = path.join(dirPath, name);
+async function update(dir, name) {
```
mjackson, line 9: Let's make `base` an arg to this function and pass it explicitly instead of deriving it from the parent dir name.

scripts/patchup-example-names.mjs
```diff
@@ -1,17 +1,28 @@
...
     let pkg = await fs.readFile(pkgPath);
     let obj = JSON.parse(pkg);
-    obj.name = `remix-example-${name}`;
+    obj.name = `remix-${base}-${name}`;
     await fs.writeFile(pkgPath, JSON.stringify(obj, null, 2) + "\n");
   }
+}
+
+examples.forEach(async (name) => {
+  await update(examplePath, name);
```
mjackson, line 23:
```suggestion
  await update(examplePath, name, "example");
```

scripts/patchup-example-names.mjs
```diff
@@ -1,17 +1,28 @@
...
+});
+
+templates.forEach(async (name) => {
+  await update(templatePath, name);
```
mjackson, line 27:
```suggestion
  await update(templatePath, name, "template");
```

jest.config.js
```diff
@@ -1,5 +1,9 @@
 module.exports = {
-  modulePathIgnorePatterns: ["<rootDir>/examples", "<rootDir>/.tmp"],
+  modulePathIgnorePatterns: [
```
mjackson, line 2: [This guy](https://github.com/facebook/jest/issues/8114#issuecomment-475068766) seems to think that using `modulePathIgnorePatterns` will work. It doesn't in our case? It's just a shame because if it did we could omit the rename script and job altogether.

scripts/patchup-example-names.mjs
```diff
@@ -1,17 +1,28 @@
...
+  let base = path.basename(dir).slice(0, -1);
+  let fullPath = path.join(dir, name);
   let stat = await fs.stat(fullPath);
   if (stat.isDirectory()) {
     let pkgPath = path.join(fullPath, "package.json");
     let pkg = await fs.readFile(pkgPath);
     let obj = JSON.parse(pkg);
-    obj.name = `remix-example-${name}`;
+    obj.name = `remix-${base}-${name}`;
     await fs.writeFile(pkgPath, JSON.stringify(obj, null, 2) + "\n");
```
mjackson, line 18: We may as well use `sortPackageJson(obj)` here instead of `JSON.stringify`.

See https://www.npmjs.com/package/sort-package-json

mjackson (changes requested):

jest.config.js
mcansh: well we have that config in place already 😅, so if it did we wouldn't be here 😭

mjackson (approved):

mcansh:
update: looks like `testPathIgnorePatterns` makes it work - tested by having every example/template use the same name

mjackson:
Sounds good. Let's use `testPathIgnorePatterns` @mcansh but leave everything else in place as well. I like having different names for everything, but I think it'll probably speed up Jest a bit to not have to scan all those other dirs for test files.

mcansh:
> Sounds good. Let's use `testPathIgnorePatterns` @mcansh but leave everything else in place as well. I like having different names for everything, but I think it'll probably speed up Jest a bit to not have to scan all those other dirs for test files.

interestingly enough it seems about the same 😞

mjackson (approved):

scripts/patchup-example-names.mjs
```diff
@@ -1,17 +1,29 @@
...
+  let fullPath = path.join(dir, name);
   let stat = await fs.stat(fullPath);
   if (stat.isDirectory()) {
     let pkgPath = path.join(fullPath, "package.json");
     let pkg = await fs.readFile(pkgPath);
     let obj = JSON.parse(pkg);
-    obj.name = `remix-example-${name}`;
-    await fs.writeFile(pkgPath, JSON.stringify(obj, null, 2) + "\n");
+    obj.name = `remix-${base}-${name}`;
+    let sorted = sort(obj);
```
MichaelDeBoey, line 18: @mcansh Not sure if it's a good idea to add running `sort-package-json` in this PR tbh as I think it could create some conflicts with #2200 (which is merged in `main`) 🤔

mcansh: yeah i think we can make a follow up PR after we get everything tightened up and squared away for re-adding that to this

MichaelDeBoey:
Seems like the `jest-haste-map` warning is back again 😢

https://github.com/remix-run/remix/runs/5580333691

PR #2334, opened 2022-03-15
fix: empty gists app remix node modules before copying new ones

mcansh:
we changed platform.ts to remix.ts but if you didnt nuke your node_modules we would combine them in remix setup

Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/fixtures/gists-app/scripts/install_remix.js
+++ b/fixtures/gists-app/scripts/install_remix.js
@@ -9,6 +9,9 @@ let installDir = path.resolve(__dirname, "../node_modules");
 async function run() {
   // Install all remix packages
   await fs.ensureDir(installDir);
+  await fs.emptyDir(path.resolve(installDir, "@remix-run"));
+  await fs.emptyDir(path.resolve(installDir, "remix"));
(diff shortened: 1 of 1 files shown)


fixtures/gists-app/scripts/install_remix.js
```diff
@@ -9,6 +9,9 @@ let installDir = path.resolve(__dirname, "../node_modules");
 async function run() {
   // Install all remix packages
   await fs.ensureDir(installDir);
+  await fs.emptyDir(path.resolve(installDir, "@remix-run"));
+  await fs.emptyDir(path.resolve(installDir, "remix"));
+  await fs.emptyDir(path.resolve(installDir, "create-remix"));
```
mjackson, line 14:
```suggestion
  await fs.emptyDir(installDir);
```

All we need is to change this line from `fs.ensureDir(installDir)` to `fs.emptyDir(installDir)`. The `emptyDir` call with [both ensure the dir exists and that it's empty](https://github.com/jprichardson/node-fs-extra/blob/master/docs/emptyDir.md#emptydirdir-callback).

mjackson (changes requested):

PR #2339, opened 2022-03-16
ci(deployment-test): switch to @remix-run/dev for deployment tests

mcansh:
I missed this when we changed create-remix to a proxy of `@remix-run/dev`

so even though it wasn't public api, we might want to mention it?

Closes: #

- [ ] Docs
- [ ] Tests


scripts/deployment-test/package.json
```diff
@@ -9,9 +9,9 @@
     "@cloudflare/wrangler": "latest",
     "@iarna/toml": "^2.2.5",
     "@octokit/rest": "^18.12.0",
+    "@remix-run/dev": "1.3.0-pre.1",
```
mjackson, line 12: Can you please make a note to come back in and update this once we publish 1.3? Thanks!

mcansh: will do, but it should be updated by the `version` script :)

Issue #111, opened 2021-05-04
pages respond with 500 status if missing action


Graham42:
For the home page of my site I only want to support `GET` requests, but if a bot comes along and does a drive-by `POST` request, then I end up with a `500` status response being returned and a stacktrace in my logs. This adds noise and false positives to monitoring which makes it harder to detect real issues. The `500` status code should be reserved for unexpected errors, and in the case of a `POST` request to `/` I intend to not support that method. 

For the first approach I tried adding an action function the returned a simple response with a `405` (Method Not Allowed) status
```js
export let action: ActionFunction = async () => {
  return new Response(undefined, {
    status: 405,
    headers: {
      Allow: "GET"
    }
  })
}
```
But when I make the `POST` request, the result is a `500` internal server error and the log error is:
```
Error! Failed to complete request to /: Error: socket hang up
Unhandled rejection: Error: You made a POST request to http://localhost:3000/ but did not return a redirect. Please `return redirect(newUrl)` from your `action` function to avoid reposts when users click the back button.
```
This makes lots of sense for a form, but for a page where I expect to never recieve a `POST`, `PUT`, etc I don't think it applies.

For my next approach I tried what feels like a bit of a 🐒 🔧 monkey patch as it uses remix internals. In the `server/index.js` file I've set up a handler to wrap the remix handler which checks if there's a route match (I think I did this wrong) and if there is, checks if there's an action for the route. If there's no action it then sends a `405` response. Otherwise it falls back to the remix handler. ⚠️ This code ends up always returning `405`s though, and not `404`s because it doesn't detect route matching correctly.

`server/index.js`:

```js
const { createRequestHandler } = require("@remix-run/vercel");
const { matchServerRoutes } = require("@remix-run/node/routeMatching");
const { createRoutes } = require("@remix-run/node/routes");

const build = require("./build");

const remixHandler = createRequestHandler({
  build: build,
});

const routes = createRoutes(build.routes);

/** @type {import("@remix-run/vercel").RequestHandler} */
const handler = async (req, res) => {
  let method = req.method.toUpperCase();
  // Everything except GET
  if (["HEAD", "OPTIONS", "POST", "PUT", "PATCH", "DELETE"].includes(method)) {
    let matches = matchServerRoutes(routes, req.url);
    // If there's no match for this route, the remix handler will return 404,
    // which is good. This is just to hand the case where the route matches, but
    // there's no action handler defined.
    if (matches || matches.length < 1) {
[...]

PR #2342, opened 2022-03-16
feat: send 405 not 500 for missing action and surface in CatchBoundary

jacob-ebey:
This really boils down to we need to treat caught action submissions as "successful" and still load data for the new location, filtering out routes at and below the boundary, passing through the boundary ID's for retention in the state after the loads complete.

I've updated the action submission navigation codepath first to verify, but I also need to update the fetcher codepath if this pans out.

Closes: https://github.com/remix-run/remix/issues/111

- [ ] Docs
- [x] Tests


packages/remix-react/routes.tsx
```diff
@@ -175,10 +175,15 @@ function createLoader(route: EntryRoute, routeModules: RouteModules) {
...
-
+function createAction(route: EntryRoute, routeModules: RouteModules) {
   let action: ClientRoute["action"] = async ({ url, signal, submission }) => {
+    if (!route.hasAction) {
+      console.error(
+        `Route "${route.id}" does not have an action, but you are trying ` +
+          `to submit to it. To fix this, please add an \`action\` function to the route`
+      );
+    }
+
```
ryanflorence, line 186: Do we have a way to do this in dev only? I don't want bots POSTing all over the place making a mess of our logs. I think I'd rather just not even have this I think if we don't have a way to branch in dev vs. prod.

integration/catch-boundary-test.ts
```diff
@@ -8,10 +8,12 @@ describe("CatchBoundary", () => {
...
-  let HAS_BOUNDARY_LOADER = "/yes/loader";
-  let HAS_BOUNDARY_ACTION = "/yes/action";
-  let NO_BOUNDARY_ACTION = "/no/action";
-  let NO_BOUNDARY_LOADER = "/no/loader";
+  const HAS_BOUNDARY_LOADER = "/yes/loader";
+  const HAS_BOUNDARY_ACTION = "/yes/action";
+  const HAS_BOUNDARY_NO_LOADER_OR_ACTION = "/yes/no-loader-or-action";
+  const NO_BOUNDARY_ACTION = "/no/action";
+  const NO_BOUNDARY_LOADER = "/no/loader";
+  const NO_BOUNDARY_NO_LOADER_OR_ACTION = "/no/no-loader-or-action";
```
machour, line 16: Could you revert those ?

packages/remix-react/routes.tsx
jacob-ebey: We could also surface this message in the catch boundary data, but it would be the first case where the framework lands you there with data of its own. Would also mean you'd have to handle a data case that you did not throw yourself and I'm not a fan of having to have that knowledge.

PR #2351, opened 2022-03-16
fix(templates): remove `cross-env` references

MichaelDeBoey:
As discussed with @chaance, all `cross-env` references should be removed when calling Remix scripts

---

Follow-up of 9e3f164e

### DIFF
--- a/templates/arc-ts/package.json
+++ b/templates/arc-ts/package.json
@@ -5,11 +5,11 @@
   "license": "",
   "sideEffects": false,
   "scripts": {
-    "postinstall": "remix setup node",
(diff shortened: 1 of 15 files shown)


kentcdodds (approved):
Yup 👍 Thanks! I think we may need to do this in the indie stack as well actually.

jensmeindertsma:
What is the reasoning behind this? 😊

chaance:
@jensmeindertsma Remix now automatically injects values from `.env` + the `NODE_ENV` into the build, so `cross-env` is really only needed in cases where those values are needed for adapters running parallel scripts.

PR #2352, opened 2022-03-17
fix(templates): fix version of remix packages

MichaelDeBoey:
Version isn't updated anymore since `templates` were moved to root of the project

---

Follow-up of #2189

CC/ @mcansh

### DIFF
--- a/templates/arc-ts/package.json
+++ b/templates/arc-ts/package.json
@@ -13,16 +13,16 @@
     "start": "cross-env NODE_ENV=production arc sandbox"
   },
   "dependencies": {
-    "@remix-run/architect": "^1.2.3",
-    "@remix-run/react": "^1.2.3",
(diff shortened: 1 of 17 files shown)


mcansh:
rats! thanks @MichaelDeBoey these should actually be "*" so they can get replaced during `create-remix`

MichaelDeBoey:
@mcansh I'll update the `package.json`s & revert the `version` script changes then 👍

mcansh:
> @mcansh I'll update the `package.json`s & revert the `version` script changes then 👍

perfect, thanks!

mcansh:
aside, but the best part of these no longer being bundled is there doesn't need to be a release!

MichaelDeBoey:
Less work for us/our scripts 🥳🎉

PR #2354, opened 2022-03-17
chore(templates): fix ESLint errors in `deno-ts`

MichaelDeBoey:
https://github.com/remix-run/remix/runs/5578734010?check_suite_focus=true#step:6:1546


MichaelDeBoey:
#2312 is reverted

PR #2359, opened 2022-03-17
Refactor runtime packages to conform to the Remix server runtime interface

pcattori:
Currently, we use "magic" imports to expose the entire Remix API through the `remix` package.

`import { anything } from 'remix'`

To achieve this, we rely on the `remix setup` command to modify `node_modules/remix` post-installation. This is non-standard for an `npm` package and causes problems with tools like [`pnpm`](https://pnpm.io/) and [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp).

Instead, we'd like to refactor to only use standard JS package conventions via four PRs:

1. Refactoring for `packages/*` (and also including `remix-deno`) to provide a common API for server runtimes (`node`, `cloudflare`, `deno`) and a common API for adapters (`express`, `netlify`, `cloudflare-workers`, etc...). **No breaking changes** (this PR)
2. Updates to templates, docs, examples mainly replacing `remix` "magic" imports with server runtime imports. **No breaking changes**
3. Deprecation of `remix` "magic" imports (e.g. console.warn when people use remix imports). **No breaking changes**
4. In a separate release: removal of the all deprecated behaviors from the previous three PRs. 🚨 **Breaking changes** 🚨


packages/remix-server-runtime/interface/cookies.ts
```diff
@@ -0,0 +1,65 @@
...
+  ): Promise<any>;
+
+  /**
+   * Serializes the given value to a string and returns the `Set-Cookie`
+   * header.
+   */
+  serialize(value: any, options?: CookieSerializeOptions): Promise<string>;
+}
+
+export type CreateCookie = (name: string, cookieOptions?: CookieOptions) => Cookie
```
mjackson, line 64: Just curious, any reason you prefer to use `type` over `interface` here? I typically prefer `interface` if I can get away with it, and use `type` when I need to create a type alias for a primitive or type union.

So e.g. I would've written this like this:

```ts
export interface CreateCookie {
  (name: string, options?: CookieOptions): Cookie;
}
```

packages/remix-server-runtime/interface/responses.ts
```diff
@@ -0,0 +1,2 @@
+export type Json = <Data>(data: Data, init?: number | ResponseInit) => Response
```
mjackson, line 1: Likewise, I probably would've preferred an `interface` here, like:

```ts
export interface JsonFunction<Data> {
  (data: Data, init?: number | ResponseInit): Response;
}
```

I think we should also name our interfaces like `XFunction` to indicate they are functions. How do you feel about that?

packages/remix-server-runtime/interface/README.md
```diff
@@ -0,0 +1,16 @@
...
+- `isSession`
+- `createSessionStorage`
+- `createCookieSessionStorage`
+- `createMemorySessionStorage`
+- `json`
+- `redirect`
+
+Corresponding types for each of these exports are defined in this directory.
+
+Additionally, each Remix Server Runtime must re-export the type definitions in `./types.ts`.
```
mjackson, line 16: I think you meant to say `./index.ts`

mjackson (approved):
This looks solid so far 👍

packages/remix-server-runtime/interface/cookies.ts
pcattori: My rule-of-thumb is to use `interface` to define object shapes, and `type` for everything else. Also, since function types can only be applied to arrow functions (and not to `function`s), I like that the syntax mirrors its usage:

```ts
type CreateCookie = (name: string, cookieOptions?: CookieOptions) => Cookie 
const createCookie: CreateCookie = (name, cookieOptions) => { ... }
```

I'd use `interface` to define a function if I needed to attach additional properties to the function, but I don't think that's ever happened for me.

packages/remix-server-runtime/interface/responses.ts
pcattori: I'm fine with that, especially since its an already-established convention in Remix codebase. 👍

packages/remix-server-runtime/interface/README.md
pcattori: Oops, I haven't added the `./types.ts` file to the draft yet. Coming soon...

packages/remix-server-runtime/reexport.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type { EntryContext } from "./entry";
+
+export type {
+  HtmlLinkDescriptor,
+  LinkDescriptor,
+  PageLinkDescriptor,
+} from "./links";
+
+export type { ServerPlatform } from "./platform";
```
mjackson, line 26: Pretty sure we don't need this anymore with this work. See my other comment about the platform arg to createRequestHandler.

packages/remix-cloudflare/globals.ts
```diff
@@ -16,3 +16,5 @@ export function installGlobals() {
   self.sign = sign;
   self.unsign = unsign;
 }
+
+installGlobals();
```
mjackson, line 20: We shouldn't need to install any globals on Cloudflare.

packages/remix-cloudflare-workers/index.ts
```diff
@@ -1,12 +1,6 @@
-import { installGlobals } from "./globals";
-
-export { createCloudflareKVSessionStorage } from "./sessions/cloudflareKVSessionStorage";
```
mjackson, line 3: Although I agree this doesn't belong here, this will probably be a breaking change for anyone running remix on cloudflare. For now, we should probably just deprecate importing `createCloudflareKVSessionStorage` from here and tell them to get it from `@remix-run/cloudflare` instead.

packages/remix-cloudflare-pages/index.ts
```diff
@@ -1,8 +1,2 @@
-import { installGlobals } from "./globals";
-
-export { createCloudflareKVSessionStorage } from "./sessions/cloudflareKVSessionStorage";
```
mjackson, line 3: Although I agree this doesn't belong here, this will probably be a breaking change for anyone running remix on cloudflare. For now, we should probably just deprecate importing `createCloudflareKVSessionStorage` from here and tell them to get it from `@remix-run/cloudflare` instead.

mjackson (changes requested):
Very excited about this!

packages/remix-cloudflare/globals.ts
```diff
@@ -0,0 +1,20 @@
...
+    process: { env: ProcessEnv };
+  }
+}
+
+export function installGlobals() {
+  self.sign = sign;
+  self.unsign = unsign;
+}
+
+installGlobals();
```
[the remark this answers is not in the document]
pcattori: Yes! Just haven't gotten around to doing the crypto cookie helpers in `@remix-run/server-runtime` yet, but it'll be in here

packages/remix-cloudflare-workers/index.ts
pcattori: `import { createCloudflareKVSessionStorage } from "remix"` should still work for existing apps since we still have `createCloudflareKVSessionStorage` in the magic exports (until we decide to do a breaking change to remove the magic exports).

It would only break if users did `import { createCloudflareKVSessionStorage } from "@remix-run/cloudflare-workers"`, but I think we agreed that changing the API `@remix-run/cloudflare-workers` isn't considered a breaking change since no one should be importing directly from `@remix-run/cloudflare-workers`.

packages/remix-server-runtime/server.ts
```diff
@@ -24,14 +24,20 @@ export interface RequestHandler {
   (request: Request, loadContext?: AppLoadContext): Promise<Response>;
 }
 
-/**
- * Creates a function that serves HTTP requests.
- */
-export function createRequestHandler(
+export type CreateRequestHandlerFunction = (
   build: ServerBuild,
   platform: ServerPlatform,
```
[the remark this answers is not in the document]
pcattori: > I don't think we're really using this for anything. 

If by "this" you mean the `platform` arg, then sure, I can remove it.

packages/remix-architect/__tests__/server-test.ts
```diff
@@ -79,7 +85,7 @@ describe("architect createRequestHandler", () => {
 
     it("handles requests", async () => {
       mockedCreateRequestHandler.mockImplementation(() => async (req) => {
-        return new Response(`URL: ${new URL(req.url).pathname}`);
+        return Promise.resolve(new Response(`URL: ${new URL(req.url).pathname}`));
```
pcattori, line 88: note: the `Promise.resolve` is not necessary. Will remove

packages/remix-architect/__tests__/server-test.ts
```diff
@@ -102,50 +108,50 @@ describe("architect createRequestHandler", () => {
...
+    //     .expectResolve((res) => {
+    //       expect(res.statusCode).toBe(200);
+    //       expect(res.headers["x-time-of-year"]).toBe("most wonderful");
+    //       expect(res.cookies).toEqual([
+    //         "first=one; Expires=0; Path=/; HttpOnly; Secure; SameSite=Lax",
+    //         "second=two; MaxAge=1209600; Path=/; HttpOnly; Secure; SameSite=Lax",
+    //         "third=three; Expires=Wed, 21 Oct 2015 07:28:00 GMT; Path=/; HttpOnly; Secure; SameSite=Lax",
+    //       ]);
+    //     });
+    // });
```
pcattori, line 154: note: forgot to uncomment theses. Will uncomment before merging.

packages/remix-cloudflare/globals.ts
pcattori: Once we remove the `sign`/`unsign` globals, will we still need to declare types for `process`/`NODE_ENV` for Cloudflare?

packages/remix-cloudflare/globals.ts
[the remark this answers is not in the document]
pcattori: Latest commits implement `createCookie` without globals

packages/remix-cloudflare-workers/index.ts
mjackson: Ah yes, of course. You're right. Thanks for the reminder. I was just reminding you of this yesterday, haha ;)

packages/remix-cloudflare/globals.ts
mjackson: We do currently encourage the use of `process.env.NODE_ENV` in user code to distinguish environment, so yeah. We still need that one.

mjackson: Note: Someday we should support a `REMIX_BUILD_MODE` global or something like that to replace `process.env.NODE_ENV`, so it feels more natural on non-node environments.

packages/remix-server-runtime/server.ts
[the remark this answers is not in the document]
mjackson: Yep, that's what I meant. I think we can refactor it out of the codebase with this work. It's ok if the function signature changes here since the remix-server-runtime interface isn't yet considered public or documented anywhere.

templates/arc-ts/app/entry.client.tsx
```diff
@@ -1,4 +1,4 @@
 import { hydrate } from "react-dom";
-import { RemixBrowser } from "remix";
+import { RemixBrowser } from "@remix-run/react";
```
MichaelDeBoey, line 2: Is this really what we want?

One of the advertised/suggested things was
> just import everything from the `remix` package

Now it seems like this isn't possible anymore?

CC/ @mjackson @ryanflorence

mjackson (commented):
Looking good. I think the only thing that remains to be done is the examples, which we can do in an other PR so this one doesn't get too big.

Would you agree?

Edit: I totally spaced it on docs changes. I think those should be done in this PR as well.

templates/arc-ts/app/entry.client.tsx
mjackson: Yes, @MichaelDeBoey. The purpose of this PR is to better define the server runtime packages and move away from importing everything from `remix`.

The `remix` package has a few downsides, not the least of which is that it requires a `remix setup`. It's also non-standard behavior to change things in `node_modules` after they are installed. This will improve compatibility with the rest of the node ecosystem, including alternative package managers like yarn plug'n'play and pnpm.

pcattori:
> Looking good. I think the only thing that remains to be done is the examples, which we can do in an other PR so this one doesn't get too big.
> 
> Would you agree?
> 
> Edit: I totally spaced it on docs changes. I think those should be done in this PR as well.

I think a good way to split this work up into PRs is:
1. PR with all refactoring for `packages/` (including `remix-deno`). This let's us test backward-compatibility.
2. PR with changes to templates, docs, examples mainly replacing `remix` magic imports.
3. PR that deprecates `remix` magic imports (e.g. `console.warn` when people use `remix` imports).

packages/remix-server-runtime/magicExports/remix.ts
```diff
@@ -34,13 +34,9 @@ export type {
 } from "@remix-run/server-runtime";
 
 export {
-  createCookie,
   isCookie,
   createSession,
   isSession,
-  createSessionStorage,
-  createCookieSessionStorage,
-  createMemorySessionStorage,
```
pcattori, line 43: TODO: for backwards compatibility, we need to add `createCookie`, `createSessionStorate`, `createCookieSessionStorage`, and `createMemorySessionStorage` to the magic exports for each server runtime package.

packages/remix-architect/__tests__/server-test.ts
```diff
@@ -2,10 +2,10 @@ import fsp from "fs/promises";
 import path from "path";
 import lambdaTester from "lambda-tester";
 import type { APIGatewayProxyEventV2 } from "aws-lambda";
-import { createRequestHandler as createRemixRequestHandler } from "@remix-run/server-runtime";
 import {
   // This has been added as a global in node 15+
   AbortController,
+  createRequestHandler as createRemixRequestHandler,
```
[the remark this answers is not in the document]
pcattori: I've thought about this some more...

The `createRequestHandler` function is defined in `@remix-run/server-runtime` as a runtime-agnostic helper function. `@remix-run/node` simply re-exports this function, not adding any runtime-specific logic.

I think `createRemixRequestHandler` might be the best name (like the alias we use in the `import`), but `createRequestHandler` might be ok too.

packages/remix-cloudflare/globals.ts
pcattori: Since we no longer `import "./globals"`, do we need to do anything to get the globals.d.ts file in  remix-cloudflare to be used by Remix apps that depend on remix-cloudflare? Do we need to point the app's tsconfig to that globals.d.ts file?

mjackson (approved):

pcattori:
TODO:
- [ ] Figure out best way for apps to consume the `globals.d.ts` file from `remix-cloudflare`
- [ ] Audit usage of `cloudflare-workers` and `cloudflare-pages` in as a server platform in `@remix-run/dev`.

PR #2362, opened 2022-03-17
feat(scripts): sort updated files in `patchup-example-names` script

MichaelDeBoey:
Follow-up of @mcansh's #2333


mcansh (approved):
perfect, thanks :)

PR #2370, opened 2022-03-18
feat: revalidate on redirect if cookies were set

jacob-ebey:
Consider this use case:

- Root loader reads the cookie to flash messages in one location
- User signs up, but need to verify their email address before they can create new projects
- User is logged in, and visits /projects/new but have not yet verified their email
- The projects/new loader redirects them to /account and sets a cookie with a flash message why they were redirected `session.flash("You need to add an email address and verify it")
- The root loader will not get called and therefore no flash message happens because Remix's optimizations didn't call the root loader.

Today we revalidate for three reasons:

- After an action (POST/PUT/PATCH/DELETE) because loaders usually read from the database and this will capture it
- If the url search params change because any loader can use them to return data
- If the exact same URL was clicked because the user probably wants fresh data (and that's what the browser does)

We're missing a fourth:

- If a redirect sets a cookie.

Just like (2) with url search params, any loader can use cookies to return data.

Additionally, this workflow works today for document transitions, but not script transitions. As a Remix principle, script and document transitions should result in the same UI.

Closes: #

- [ ] Docs
- [x] Tests

### DIFF
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -148,7 +148,7 @@ async function handleDataRequest({
       headers.set("X-Remix-Redirect", headers.get("Location")!);
       headers.delete("Location");
       if (response.headers.get("Set-Cookie") !== null) {
-        headers.set("X-Remix-Set-Cookie", "yes");
+        headers.set("X-Remix-Revalidate", "yes");
       }
 
       return new Response(null, {
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -147,6 +147,9 @@ async function handleDataRequest({
       let headers = new Headers(response.headers);
       headers.set("X-Remix-Redirect", headers.get("Location")!);
       headers.delete("Location");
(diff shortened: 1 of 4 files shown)


packages/remix-server-runtime/server.ts
```diff
@@ -147,6 +147,9 @@ async function handleDataRequest({
       let headers = new Headers(response.headers);
       headers.set("X-Remix-Redirect", headers.get("Location")!);
       headers.delete("Location");
+      if (response.headers.get("Set-Cookie") !== null) {
```
jacob-ebey, line 150: In the browser, you can not access `response.headers.get("Set-Cookie")` to determine if a cookie was set in the response when `credentials: "same-origin"` is set. To avoid having to set it to `credentials: "include"` and adding additional CORS headers, this is my workaround.

ryanflorence: Seems fine to me, maybe X-Remix-Revalidate instead incase we discover other reasons to revalidate we can use the same mechanism?

PR #2374, opened 2022-03-18
feat: add getDependenciesToBundle to @remix-run/dev

jacob-ebey:
chore: made getAppDependencies and getPackageDependencies synchronous
chore: removed `read-package-json-fast` dep in @remix-run/dev

Closes: #

- [ ] Docs
- [ ] Tests


packages/remix-dev/compiler/dependencies.ts
```diff
@@ -1,21 +1,54 @@
...
+}
+
+function getPackageDependenciesRecursive(
+  pkg: string,
+  aggregatedDeps: Set<string>,
+  visitedPackages: Set<string>
+): void {
+  visitedPackages.add(pkg);
+
+  let pkgJson = require.resolve(`${pkg}/package.json`);
```
mjackson, line 39: Should probably use a `path.join` instead of a `/` here.

packages/remix-dev/compiler/dependencies.ts
```diff
@@ -1,21 +1,54 @@
...
+  if (!fs.existsSync(pkgJson)) {
+    console.log(pkgJson, `does not exist`);
+    return;
+  }
+
+  let dependencies = getPackageDependencies(pkgJson);
+  console.log(dependencies);
```
mjackson, line 46: Looks like this should be removed.

mjackson (changes requested):

packages/remix-dev/compiler/dependencies.ts
jacob-ebey: Path.join will do a "\" on windows, and since this is actually require syntax, not a path, that would break.

PR #2375, opened 2022-03-18
chore: Cli cleanup

mjackson:
No need to use a shell to execute remix-dev. Instead, export the CLI interface from remix-dev and use it directly in create-remix.


pcattori (approved):

kentcdodds:
Looks like there's a merge conflict.

chaance (approved):

PR #2376, opened 2022-03-18
test: use shared puppeteer instance for tests

mcansh:
use shared puppeteer instance for tests

https://jestjs.io/docs/puppeteer#custom-example-without-jest-puppeteer-preset

note using [`--forceExit`](https://jestjs.io/docs/cli#--forceexit) here, which i don't love, but `--detectOpenHandles` doesn't print anything due to never exiting 😅 

Closes: #

- [ ] Docs
- [ ] Tests


packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -133,6 +134,21 @@ describe("remix cli", () => {
...
+        let pkg = await fsp.readFile(pkgPath);
+        let obj = JSON.parse(pkg.toString());
+        obj.name = path.basename(dir);
+        await fsp.writeFile(pkgPath, JSON.stringify(obj, null, 2) + "\n");
+      }
+
+      let dirs = fs.readdirSync(path.join(process.cwd(), ".tmp"));
+      for (let dir of dirs) {
+        renamePkgJsonApp(path.join(process.cwd(), ".tmp", dir));
+      }
```
pcattori, line 149: Why do we need to change the `name` in `package.json` for each of the created apps?

mcansh: jest-haste https://github.com/remix-run/remix/runs/5579325183?check_suite_focus=true#step:7:6

pcattori: Does the naming collision happen if we just omit the `name` field from the apps?

Since these are apps and not libraries, the `name` field is misleading anyway.

mcansh: let me see, that'd be great if we could omit them :)

we'd just need to remove them from the examples/templates directories as well

pcattori (approved):
LGTM

I think we _should_ omit the `name` field from `package.json` for any Remix apps we create. But that can be done as follow-on work later.

packages/remix-dev/__tests__/cli-test.ts
mcansh: so far it seems fine, gonna run it a few times to pile up the amount of apps in .tmp and open a PR to remove all our shenanigans

Issue #167, opened 2021-05-15
Docs issue: delete is not is not a valid formMethod


nickdima:
[The example](https://github.com/remix-run/remix/blame/master/docs/api/remix.md#L219) for `useFormAction` is not correct as `delete` is not a valid value for the `<button>`'s `formMethod` attribute. Only `get` and `post` are browser supported values.

machour:
This is the updated link : https://github.com/remix-run/remix/blob/main/docs/api/remix.md#useformaction

PR #2377, opened 2022-03-18
docs(useFormAction): delete is not a valid form method

machour:
Closes: #167

- [x] Docs
- [ ] Tests

### DIFF
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -476,7 +476,7 @@ function SomeComponent() {
   return (
     <button
       formAction={useFormAction("destroy")}
-      formMethod="DELETE"
+      formMethod="POST"
     >
       Delete
     </button>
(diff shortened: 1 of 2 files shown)


docs/api/remix.md
```diff
@@ -476,7 +476,7 @@ function SomeComponent() {
   return (
     <button
       formAction={useFormAction("destroy")}
-      formMethod="DELETE"
+      formMethod="POST"
```
MichaelDeBoey, line 479: [The docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Button#attr-formmethod) only talk about `get` or `post` (lowercase), so maybe we should use lowercase too?

```suggestion
      formMethod="post"
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

PR #2378, opened 2022-03-18
Default <Outlet /> loader not called after post

alexinfurs:
Hi, this is a bug I'm facing now
The defalut `<Outlet />` loader is not called after posting with `<Form />`, causing `UseLoaderData()` returning undefined and that breaks the page.

I think it's a bug because all loaders should be called after an action


jacob-ebey:
This seems to have been addressed in a recent release. I'm going to close this for now but if you run into this issue again please re-open this issue :)

PR #2388, opened 2022-03-18
feat(server-runtime): default cookies to `sameSite: "lax"`

### LINKED ISSUE #200: default cookies to `SameSite: Lax`
Chrome recently started defaulting to `SameSite: Lax`, and if you don't do it in your remix cookies/sessions, you get this error in the devtools:

image

We should probably add this as the default in Remix to get rid of the error and makes sites more secure by default for browsers that aren't doing this yet.

chaance:
Resolves #200

### DIFF
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -81,6 +81,7 @@ export function createCookie(
   let { secrets, ...options } = {
     secrets: [],
     path: "/",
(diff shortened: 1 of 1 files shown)


mcansh (approved):

mcansh:
LGTM, might be more of a "feature" than a "fix" though 😉

MichaelDeBoey (approved):

PR #2392, opened 2022-03-18
docs(examples): bring `supabase-subscription` in line with template

MichaelDeBoey:
Follow-up of #2220


kentcdodds (approved):
Thanks for doing this cleanup

PR #2394, opened 2022-03-19
tutorial blog fix: remove non functional code

keshavkaul:
removed non imported remix `json` function returning post slug when creating new post.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -292,3 +292,4 @@
 - youngvform
 - zachdtaylor
 - zainfathoni
+- keshavkaul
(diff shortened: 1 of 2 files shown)


machour:
Hi and thank you for taking the time to propose this PR 🙏🏼  

The call you removed is returning the whole new Post object, not its slug only. 

It's a good practice to return the newly created entity from there, to be able to do things like optimistic UI for example.

We are in the process of retiring the blog tutorial to use remix stacks, so we will leave the code as it's for now. The new tutorial will be up soon and we will be more than glad to have your feedback then.

PR #2396, opened 2022-03-19
docs(examples/framer-motion): Fix wrong codesandbox link on README.md

DavOnGit:
The link now is pointing to the basic example, with this change it will open the 'framer-motion' example.
_(This is a docs change)_


machour:
Nice catch @DavOnGit, thank you!

machour (approved):

PR #2398, opened 2022-03-19
docs: typo in data-writes

pacexy:


machour:
Nice catch, thank you @pacexy 🙌

PR #2401, opened 2022-03-19
chore(issues-template): Clarify instructions when the user have a fix 

machour:
So that we can properly see the failure, and then the fix.


machour:
We might want to advise the user to open the PR against the `dev` branch, since its code related.
Not sure if this is done elsewhere tho.

PR #2403, opened 2022-03-19
fix(remix-react): add proper error-message if server-side code spills to client

DavOnGit:
Closes: #1618 

- [x] Tests


mjackson:
This PR is just treating a symptom of the problem. Can we dig a little deeper and figure out why the code is making it into the client bundles in the first place?

kiliman:
It's primarily a newbie issue. They don't always understand the distinction between server and client modules and side-effects, etc. Even that of us that are experienced get caught up when dealing with a new package.

All this PR does is just give the user a hint at what may be the problem. Unfortunately it's not something that can be caught at compile-time. So at least getting a more descriptive error than *'root' is undefined* is helpful.

machour (approved):

jacob-ebey:
Thanks for the contribution @DavOnGit.

Franrro2:
It's primarily a newbie issue. They don't always understand the distinction between server and client modules and side-effects, etc. Even that of us that are experienced get caught up when dealing with a new package.

All this PR does is just give the user a hint at what may be the problem. Unfortunately it's not something that can be caught at compile-time. So at least getting a more descriptive error than 'root' is undefined is helpful

Franrro2:
Fix(remix:react):add proper error_massage if server-side code spills to client

ryanflorence:
> Can we dig a little deeper and figure out why the code is making it into the client…

Developers can import from any package they want and try to use it in a component, blowing up the client. There’s nothing we can do or know about it.

We can keep getting better at preventing many cases but I can’t image we’d ever be able to prevent all of them.

I think this is useful even as we get better at splitting client/server code. I’ve wanted to add this warning for a while now.

mjackson:
I'm not opposed to adding a warning for server code that makes it into the client bundles. What I meant was: this is the wrong place for this code. We now have two invariants in `RemixRoute` that we will likely never be able to remove, or we will be scared to remove them at some point in the future when our compiler is better because they seem useful.

At the very least these invariants should be in `useRemixEntryContext` or perhaps in the root provider itself since anyone who needs these properties from the root context is going to be broken if they aren't present.

PR #2411, opened 2022-03-20
fix: Keep empty meta from rendering a tag

chaance:
Currently, if a meta key is `null` or `undefined`, we will render an empty tag. Our types also don't support an empty meta key, which is a bit annoying if you're working with optional data.

```js
export function meta({ data }) {
  return {
    title: data.title, // always there, cool
    description: data.description // optional, don't bother with a tag but please don't yell at me!
  };
}
```

### DIFF
--- a/fixtures/gists-app/app/routes/meta.tsx
+++ b/fixtures/gists-app/app/routes/meta.tsx
@@ -1,0 +1,28 @@
+import { LoaderFunction, MetaFunction } from "remix";
+import { json, useLoaderData } from "remix";
+
+export let loader: LoaderFunction = async () => {
+  return json({
+    title: "Meta Page",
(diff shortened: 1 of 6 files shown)


machour (approved):

fixtures/gists-app/app/routes/meta.tsx
```diff
@@ -0,0 +1,28 @@
+import { LoaderFunction, MetaFunction } from "remix";
```
kentcdodds, line 1: I think we’re trying to use Ryan’s `createFixture` utility as much as possible instead of the gist app. Could you switch this test to use that instead?

kentcdodds (changes requested):
I’ma superfan of this change 👍

Issue #1831, opened 2022-02-07
[Feature]: Allow dev server to listen on localhost


RazerM:
### What is the new or updated feature that you are suggesting?

I want to be able to have the dev server listen on localhost. I think https://github.com/remix-run/remix/pull/1328 should be reconsidered (e.g. by making it configurable).

### Why should this feature be included?

I don't want my development app to be available on the whole network. It seems a bit of an unusual default compared to other web frameworks I'm familiar with — Django and Flask's dev servers both listen on 127.0.0.1, and I assume it's similar for most others. Typically I'd only change this for demo purposes, but in any case I doubt any default suits everyone so an environment variable or CLI option would be ideal.

kiliman:
They don't include the reason for this change, so I'm assuming it's to handle something internal. Probably due to testing?

Anyway, if you need more control over things, then I would switch to the Express adapter instead of using Remix App Server. This will allow you to configure the server however you want. *NOTE: RAS uses Express under the hood.*

nmaves:
I agree the change in [#132](https://github.com/remix-run/remix/pull/1328) should be reconsidered.  It causes secure cookies to fail for local development.

nmaves:
Maybe we could have the best of both worlds and display something like this....

  Local:            http://localhost:3000
  On Your Network:  http://192.168.1.87:3000

kettanaito:
The change to always use local IP is disruptive. Most of the dev tools and integrations will rely on `localhost`, which is fixed, while local IP is variable. For example, I cannot work with an OAuth in my app because its callback is set to `localhost:300/auth`, so I end up using `localhost` anyway. 

👍 to making this configurable. How to establish the dev server is the developer's decision.

mcansh:
I'm not opposed either way, but just letting you know that the server is still accessible via localhost, we just log the local ip address

mcansh:
opened #2301, thanks for patience every one :pray:

MichaelDeBoey:
Since #2301 is merged, I think this one can be closed now?

mcansh:
> Since #2301 is merged, I think this one can be closed now?

yup thanks!

RazerM:
@mcansh that PR does not address the issue, which is about which IP address remix _listens on_. I'm aware that I can access remix via localhost when it is listening on the IP address of my machine. I think the description is clear (?)

[...]

PR #2415, opened 2022-03-20
feat(remix-dev,remix-serve): Allow specifying HOST environment variable

rossipedia:
This will allow users to bind `remix-serve` to a specific interface, if they want to keep it private for example.


mcansh:
can we bring this same change over to `remix dev`? https://github.com/remix-run/remix/blob/5acfa8519f93b13d166749e0a5fb9cbda4cd76fe/packages/remix-dev/cli/commands.ts#L269-L281

rossipedia:
@mcansh yeah just added :+1:

mcansh:
> @mcansh yeah just added 👍

perfect! i think this also covers the last bit from https://github.com/remix-run/remix/issues/1831#issuecomment-1075425553 so we can restrict to localhost only, right? `HOST=localhost remix dev`

rossipedia:
The `express` types are particular about overloads. While it will work if you call `.listen(port, undefined, onListen)`, the typings complain because the overload that takes those 3 arguments requires that the second is a `string`, not a `string | undefined`. It also wants the first arg to be a `number`, instead of `string | number`, hence the `parseInt` stuff.

rossipedia:
@mcansh gentle ping :)

mcansh:
one more thing, can you switch this to point to dev?

PR #2416, opened 2022-03-20
fix: wrong server build path in express template

dogukanakkaya:
Express server build is placed into the build directory in the root of the project, not to the server directory.


dogukanakkaya:
Opened different pull request for dev branch.

PR #2430, opened 2022-03-21
feat(remix-dev): add `codemod` command

MichaelDeBoey:
Now that @pcattori's work on #2359 is merged, I think it would be a good idea to have a codemod that can transform all `remix` imports to specific `@remix-run/*` imports.

This package is heavily inspired by how things are done in [`react-codemod`](https://github.com/reactjs/react-codemod) & [`@next/codemod`](https://github.com/vercel/next.js/tree/canary/packages/next-codemod).


packages/remix-codemod/bin/cli/check-git-status.ts
```diff
@@ -0,0 +1,39 @@
+import isGitClean from "is-git-clean";
```
pcattori, line 1: I _think_ this file is copied from [`react-codemod`](https://github.com/reactjs/react-codemod/blob/master/bin/cli.js#L22).

Could you include a file-level comment with a link to the original and stating if this purely copied or if there were modifications from the original?

^I'd request the same for any other files in `cli/` that were taken from `react-codemod`, just to make reviewing this a bit easier 😃

packages/remix-codemod/bin/cli/expend-file-paths-if-needed.ts
```diff
@@ -0,0 +1,11 @@
+import { globbySync } from "globby";
```
pcattori, line 1: nit: typo in filename `expend` -> `expand`

pcattori:
@MichaelDeBoey I think we want the codemod to be a subcommand of the existing CLI in `@remix-run/dev`, rather than its own `remix-codemod` CLI. E.g.

```sh
npx @remix-run/dev codemod
```

For context, our (tentative) plan is to revive the `remix` package as the CLI, after all four PRs described in #2359 are in.

So a possible future would be:
```sh
remix create # instead of create-remix
remix codemode
```

MichaelDeBoey:
@pcattori I updated the transform to work properly now and used some of your code of #2471 doing so.

You can see it working at https://astexplorer.net/#/gist/f4dc5a2ef4867524526893e90a160e3d/latest.

I also kept remaining `remix` imports to not break people's code (this should normally never happen though 🤷‍♂️)

Let me know what you think of it.

Still need to

- [ ] Remove `remix` dependency from `package.json`
- [ ] Remove `remix setup` call from `postinstall` script (+ delete if it's empty now)

packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -58,11 +59,16 @@ describe("remix cli", () => {
...
+            --dry               Dry run (no changes are made to files)
+            --force             Bypass Git safety checks and forcibly run codemods
 
           Values:
             - projectDir        The Remix project directory
             - template          The project template to use
-            - remixPlatform     node or cloudflare
+            - remixPlatform     \`node\` or \`cloudflare\`
+            - transform         One of the choices from https://github.com/remix-run/remix/tree/main/packages/remix-codemod
+            - path              Files or directory to transform. Can be a glob like app/**.ts(x)
```
pcattori, line 71: @MichaelDeBoey what do you think about reusing `projectDir` instead of `path`?

packages/remix-dev/cli/run.ts
```diff
@@ -37,11 +39,20 @@ ${colors.heading("Options")}:
...
+  --force             Bypass Git safety checks and forcibly run codemods
 
 ${colors.heading("Values")}:
   - ${colors.arg("projectDir")}        The Remix project directory
   - ${colors.arg("template")}          The project template to use
-  - ${colors.arg("remixPlatform")}     node or cloudflare
+  - ${colors.arg("remixPlatform")}     \`node\` or \`cloudflare\`
+  - ${colors.arg(
+    "transform"
+  )}         One of the choices from https://github.com/remix-run/remix/tree/main/packages/remix-codemod
```
pcattori, line 52:
```suggestion
  )}         One of the choices from https://github.com/remix-run/remix/tree/main/packages/remix-dev/cli/codemod/transforms
```

packages/remix-dev/cli/run.ts
```diff
@@ -303,6 +316,22 @@ export async function run(argv: string[] = process.argv.slice(2)) {
...
+      }
+
+      let answers = await commands.codemod.questions({
+        input: {
+          files: input[2],
+          transform: input[1],
+        },
+        showHelp,
+      });
+
```
pcattori, line 331: If we ask for a `projectDir` (like I mentioned in my other comment), we can use that here to find the `package.json` file and do all the updates to it that we need. We can also auto-detect the server runtime and server adapters from the dependencies listed in `package.json`.

packages/remix-dev/cli/run.ts
```diff
@@ -303,6 +313,22 @@ export async function run(argv: string[] = process.argv.slice(2)) {
     case "setup":
       await commands.setup(input[1]);
       break;
+    case "codemod": {
+      if (!flags.dry) {
+        checkGitStatus(flags.force);
```
pcattori, line 318: I think it would be better to check the `git status` of the actual project directory where we are trying to run the transform. Right now, its checking the `git status` of the current working directory before even considering the project directory.

pcattori (changes requested):

packages/remix-dev/cli/run.ts
MichaelDeBoey: That means we should move it to the `validateAnswers` function?

pcattori (approved):

PR #2440, opened 2022-03-22
fix(remix-dev): use logical or instead of nullish coalescing

mcansh:
it landed in [node 14](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator#browser_compatibility), but I've seen it come up in discord https://canary.discord.com/channels/770287896669978684/771068344320786452/95567838904860263 and https://canary.discord.com/channels/770287896669978684/954545525984722954/956296309256646707

ah!!! the Node version check doesn't run because it explodes trying to parse before it even gets there 🤦 

Closes: #

- [ ] Docs
- [ ] Tests


mcansh:
turns out latest [wrangler-action](https://github.com/cloudflare/wrangler-action/releases/tag/1.3.0) still uses node 12, but they published a `next` version with node 16, 14 hours ago (of posting this), so there's hope

sergiodxa (approved):

MichaelDeBoey:
@mcansh Shouldn't the fix be to make the CLI not explode earlier then instead of what's in this PR?

jenseng (commented):
see inline comment... stepping back a bit, is this change even the right way to fix it? would a better approach be to fix it earlier as @MichaelDeBoey suggests?

e.g. maybe add this to package.json
```json
"engines": {
  "node": ">=14"
}
```

mcansh:
> 
> ```json
> 
> "engines": {
> 
>   "node": ">=14"
> 
> }
> 
> ```

that doesn't appear to prevent you from running it though, which is the reason for the runtime check

jenseng:
Ah yeah, I think yarn might prevent the install, but not npx 🤔

MichaelDeBoey:
@mcansh Once https://github.com/remix-run/remix/pull/2556 is merged (with the Node version check moved), I don't think we need this anymore? 🤔

mcansh:
> @mcansh Once https://github.com/remix-run/remix/pull/2556 is merged (with the Node version check moved), I don't think we need this anymore? 🤔

yup, you're right. i also updated some things to help aid in node < 14 handling :)

PR #2447, opened 2022-03-22
ci: re-use arc app

mcansh:
AWS has a 100 s3 bucket per account limit and we've quickly reached that, this PR removes the oldest deployment prior to making a new one

Closes: REM-901

- [ ] Docs
- [ ] Tests


kentcdodds (approved):
Nice trick 👍

PR #2450, opened 2022-03-22
chore(remix-server-runtime): standardize on uppercase method

### LINKED ISSUE #145: Normalize fetch methods to uppercase
Apparently `PATCH` is case-sensitive, so we should just normalize all HTTP methods to uppercase when submitting forms via `fetch`.

donavon:
According to the [Fetch spec](https://fetch.spec.whatwg.org/#methods), method names are case insensitive _except_ for standard names `DELETE`, `GET`, `HEAD`, `OPTIONS`, `POST`, and `PUT` which are all normalized to "byte uppercase".

The only exception is `PATCH` which was never made the spec. For `PATCH` you will need to upper case it before you compare.

The PR brings consistency with the spec as well as saves the need to `toLowerCase` on methods that are guaranteed to be upper case.

---

Closes #145


packages/remix-server-runtime/server.ts
```diff
@@ -544,24 +544,19 @@ async function handleResourceRequest({
...
   return (
-    method === "post" ||
-    method === "put" ||
-    method === "patch" ||
-    method === "delete"
+    method === "POST" ||
+    method === "PUT" ||
+    method === "DELETE" ||
+    method.toUpperCase() === "PATCH"
   );
```
MichaelDeBoey, line 553: Maybe we could even simplify this to

```suggestion
    return ["POST", "PUT", "DELETE"].includes(method.toUpperCase());
```

kiliman:
I was curious it this would break if a client sent lowercase methods, but apparently `fetch` and `node-fetch` both covert method to uppercase before generating the request. So by the time it reaches Remix, the method should be normalized to uppercase.

packages/remix-server-runtime/server.ts
donavon: see new commit. no need to create/destroy the array each time

donavon:
> the method should be normalized to uppercase.

per the Fetch spec, yes

donavon:
Now using `Set`, which is perfect. In database terminology, execution is "indexed" with `Set` vs a "full table scan" with `includes`

MichaelDeBoey (approved):

sergiodxa (approved):

machour:
Should fix https://github.com/remix-run/remix/issues/145

machour (approved):

PR #2456, opened 2022-03-22
fix(remix-react): make `LiveReload` work for ports other than 8002

mcansh:
Prior to this change, if `config.devServerPort` was changed either manually or automatically (something already running on 8002), the port inlined to LiveReload was 8002 anyways. I went through some testing locally in `node_modules` and this was the solution. with our current testing infrastructure, I wasn't sure where to write a test for this, so if anyone has an idea, please let me know :)

Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1380,7 +1380,7 @@ export const LiveReload =
       }: {
         port?: number;
       }) {
-        let setupLiveReload = ((port: number) => {
+        let setupLiveReload = (() => {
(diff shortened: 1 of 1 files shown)


jacob-ebey (approved):

kentcdodds:
This caused an issue where `port` is `undefined` in the inline script. I noticed and made a PR to fix it, but then realized that this had been removed. Could you look into this further? https://github.com/remix-run/remix/pull/2662

I'm guessing my PR isn't enough, but this PR breaks livereload 😬

PR #2466, opened 2022-03-23
fix(transition): Retain index route match in layout form submission revalidation

brophdawg11:
When submitting an action from a layout route, we were losing the nested index route during revalidation.

Reproducible in https://github.com/jacob-ebey/bad-matches-bug

- [x] Docs - n/a
- [x] Tests

### DIFF
--- a/packages/remix-react/transition.ts
+++ b/packages/remix-react/transition.ts
@@ -629,7 +629,7 @@ export function createTransitionManager(init: TransitionManagerInit) {
   ) {
     let match = matches.slice(-1)[0];
 
-    if (!isIndexRequestUrl(url) && match.route.id.endsWith("/index")) {
+    if (!isIndexRequestUrl(url) && match.route.index) {
       return matches.slice(-2)[0];
     }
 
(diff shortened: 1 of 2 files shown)


packages/remix-react/transition.ts
```diff
@@ -1053,14 +1053,18 @@ export function createTransitionManager(init: TransitionManagerInit) {
...
+    // to run on layout/index routes.  We do not want to mutate the eventual
+    // matches used for revalidation
+    let actionMatches = matches;
     if (
-      !isIndexRequestAction(submission.action) &&
-      matches[matches.length - 1].route.id.endsWith("/index")
+      !isIndexRequestUrl(createUrl(submission.action)) &&
+      actionMatches[matches.length - 1].route.index
     ) {
-      matches = matches.slice(0, -1);
```
brophdawg11, line 1060: This as the bug - by mutating in place prior to getting the `leafMatch` we were causing the trimmed `matches` to persist through revalidation

brophdawg11: I also updated this to use the new `isIndexRequestUrl` which was doing the same thing as `isIndexRequestAction`

PR #2471, opened 2022-03-23
feat(codemod): replace-remix-imports

pcattori:
Throw an error if Remix is a:
- Side-effect import: `import 'remix'` (TODO)
- Default import: `import Remix from 'remix'`
- Namespace import: `import * as Remix from 'remix'`
The error should direct users to replace these^ imports with named imports.

Example input:

```tsx
import { isCookie, createCookie, RemixServer, Meta } from 'remix'
import { ScrollRestoration, createCookie as blah, yo, Link } from 'remix'
import type { ActionFunction } from 'remix'
import type { LoaderFunction as LF, FormProps as FP  } from 'remix'
import React, * as blah2 from 'react'

import { hook } from './lib'

const a = 1
console.log('blah')
```

- [ ] Docs
- [ ] Tests

### DIFF
--- a/examples/basic/app/entry.client.tsx
+++ b/examples/basic/app/entry.client.tsx
@@ -1,4 +1,4 @@
 import { hydrate } from "react-dom";
-import { RemixBrowser } from "remix";
+import { RemixBrowser } from "@remix-run/react";
 
 hydrate(<RemixBrowser />, document);
--- a/examples/basic/app/entry.server.tsx
+++ b/examples/basic/app/entry.server.tsx
@@ -1,6 +1,6 @@
 import { renderToString } from "react-dom/server";
-import { RemixServer } from "remix";
-import type { EntryContext } from "remix";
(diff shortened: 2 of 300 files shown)


pcattori:
Added a commit applying the codemod to `examples/`

mjackson:
Can we have the codemod also remove the `remix` entry in package.json? /cc @MichaelDeBoey

pcattori:
> Can we have the codemod also remove the `remix` entry in package.json? /cc @MichaelDeBoey

The transform itself is for modifying javascript packages, but the `codemod` command that calls the transform can certainly do that

MichaelDeBoey:
> the `codemod` command that calls the transform can certainly do that.

That's indeed what I was thinking too

MichaelDeBoey:
@pcattori Since #2430 is merged & we have #2575 & #2585 run with latest version of the codemod, I think we can close this one?

PR #2472, opened 2022-03-23
feat(remix-dev): support nested pathless layout routes

sergiodxa:
Because pathless layouts has a uniqueRouteId being `""` this adds the whole routeId, this way nested layouts will be use the actual routeId.

Closes: #

- [ ] Docs
- [x] Tests (pending)

### DIFF
--- a/integration/pathless-routes-test.ts
+++ b/integration/pathless-routes-test.ts
@@ -12,17 +12,22 @@ describe("rendering", () => {
     fixture = await createFixture({
       files: {
         "app/root.jsx": js`
-          import { Outlet, Scripts } from "remix";
+          import { Links, Meta, Outlet, Scripts } from "remix";
           export default function Root() {
             return (
-              <html>
-                <head />
+              <html lang="en">
+                <head>
+                  <Meta />
+                  <Links />
+                </head>
                 <body>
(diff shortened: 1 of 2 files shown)


sergiodxa:
Test seems to pass, but I'm not really sure if the test is correct, because this only happen on dev. At least it ensure the build for prod will not fail.

How should I write a test for the dev CLI?

integration/pathless-routes-test.ts
```diff
@@ -0,0 +1,66 @@
...
+            return (
+              <html>
+                <head />
+                <body>
+                  <Outlet />
+                  <Scripts />
+                </body>
+              </html>
+            )
+          }
```
MichaelDeBoey, line 26: See #2454

```suggestion
          import { Links, Meta, Outlet, Scripts } from "remix";
          export default function Root() {
            return (
              <html lang="en">
                <head>
                  <Meta />
                  <Links />
                </head>
                <body>
                  <main>
                    <Outlet />
                  </main>
                  <Scripts />
                </body>
              </html>
            );
          }
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

PR #2473, opened 2022-03-24
docs(examples/jokes): add validation to prevent open redirect on login

LewisArdern:
Hi Remix,

I was going through the [Remix Tutorial](https://remix.run/docs/en/v1/tutorials/jokes#optimistic-ui) and noticed that the login.tsx contains an open redirect issue. I realize this is a tutorial, but since this is setting developers up for the first time in the remix environment, its probably important to ensure that they do not introduce this type of issue into their apps.

This change validates the URL based on /jokes and / (since those are the two routes primarily used) in the tutorial.

If you want me to update the docs too, happy to do that -- just point me in the right direction.

Love Remix so far! Thank you for creating some awesome software. 

Lewis


MichaelDeBoey:
@LewisArdern Changes to `examples` need to go against `main` branch.
Can you please rebase your branch onto `main` & change base branch of this PR to `main` too?

kentcdodds:
Thanks for the contribution! There are likely a lot of little improvements we can make here and there to these tutorials, but at some point you have to draw a line at how far you go and we want to focus on learning Remix and making the app work, so I don't think this needs to be added to the example or tutorial.

Thanks anyway!

ryanflorence:
we should definitely validate the redirect location so we aren’t teaching people to expose their apps to “open redirect” vulnerabilities.

Could be as simple as using the pathname (no host) instead of the full URL. Would that feel less cumbersome @kentcdodds?

https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html

kentcdodds:
🤦‍♂️ I didn't read carefully enough and didn't notice the mention of a vulnerability. Yes, let's get this in 👍

kentcdodds:
We'll want to update the tutorial as well. Would you be interested in doing that @LewisArdern?

https://github.com/remix-run/remix/edit/main/docs/tutorials/jokes.md

Looks like there are three instances of this in the tutorial.

LewisArdern:
Yes, I'll make some changes today and update the docs and change to main like @[MichaelDeBoey](https://github.com/MichaelDeBoey) suggested

LewisArdern:
I think I messed up the pull request, just gonna submit a new pull request against master.

PR #2482, opened 2022-03-24
docs(examples): bring `google-analytics` example in line with templates

MichaelDeBoey:


kentcdodds (approved):
Thanks for doing this work! Especially for the ones that were started before we had the template in a good spot.

PR #2486, opened 2022-03-24
docs(examples): bring `infinite-scrolling` example in line with templates

MichaelDeBoey:
Follow-up of #2107


kentcdodds (approved):
Thanks!

PR #2489, opened 2022-03-25
feat: CSS modules (early experimental support)

chaance:
Hi


package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "clean": "git clean -fdX .",
     "build": "rollup -c && tsc -b",
+    "predev": "npm install --prefix ./fixtures/gists-app/",
     "dev": "yarn build && yarn --cwd ./fixtures/gists-app/ dev",
```
pcattori, line 9: @chaance `predev` and `dev` scripts still reference the `gists` app. I think we should just remove these two scripts.

packages/remix-dev/compiler/plugins/cssModulesPlugin.ts
```diff
@@ -0,0 +1,387 @@
...
+  name: string,
+  composes: CssModuleReference[]
+) {
+  return composes.reduce((className, composed) => {
+    // skip dependencies for now
+    if (composed.type === "dependency") {
+      console.log({ composed });
+      return className;
+    }
+    return className + " " + getPrefixedClassName(prefix, composed.name);
```
devongovett, line 268: FYI, you will need to do this recursively since locals can compose other locals. For example:

```css
.foo {
  color: red;
}

.bar {
  composes: foo;
}

.baz {
  composes: bar;
}
```

When using `baz`, the class names for `bar` and `foo` should also be applied.

It may be helpful to look at the implementation in Parcel here: https://github.com/parcel-bundler/parcel/blob/50893a7ec2b1490e6d0a2f5e1f9ab6a8fc57a113/packages/transformers/css/src/CSSTransformer.js#L153-L187

packages/remix-dev/compiler/plugins/cssModulesPlugin.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  // sort() to keep order consistent in different builds
+  for (let originClass of Object.keys(parcelExports).sort()) {
+    let props = parcelExports[originClass];
+    let patchedClass = props.name;
+    let prefixedClassName = getPrefixedClassName(classPrefix, patchedClass);
+    json[originClass] = props.composes.length
+      ? getComposedClassNames(classPrefix, prefixedClassName, props.composes)
+      : prefixedClassName;
+    cssModulesContent = cssModulesContent.replace(
```
devongovett, line 197: I guess if we added support for something like postcss-modules `generateScopedName` you could avoid this? https://github.com/parcel-bundler/parcel-css/issues/156

packages/remix-dev/compiler/plugins/cssModulesPlugin.ts
```diff
@@ -0,0 +1,387 @@
...
+
+  let cssWithSourceMap = cssModulesContent;
+  if (res.map) {
+    // TODO: Sourcemaps aren't working as expected because we are inlining the
+    // map at the end of each module. We need to merge them into a single
+    // inline sourcemap at the end of the build. We can probably use something
+    // like https://www.npmjs.com/package/merge-source-maps
+    // cssWithSourceMap += `\n/*# sourceMappingURL=data:application/json;base64,${res.map.toString(
+    //   "base64"
+    // )} */`;
```
devongovett, line 211: You could try out the Parcel source map library for this potentially: https://github.com/parcel-bundler/source-map

packages/remix-css-modules/README.md
```diff
@@ -0,0 +1,13 @@
+# Welcome to Remix!
```
wingleung, line 1: prolly one of the last tasks on the list but some documentation on how to use it might already be helpful for people like me wanting to test it out, this is what I figured out was the way to use it

```javascript
// in root.tsx
import CSSModules from "@remix-run/css-modules"

export const links: LinksFunction = () => {
  return [
    { rel: "stylesheet", href: CSSModules }
  ];
};
```

But I didn't manage to get the watch mode working, after a change in a css module I got a css-module-faker error. prolly WIP? 

big thanks for the work on this! we're very interested in using css modules in our remix app as well so if you need some help, let me know

packages/remix-dev/compiler/plugins/cssModulesPlugin.ts
devongovett: FYI working on that over here if you have any feedback: https://github.com/parcel-bundler/parcel-css/pull/180

wingleung:
@chaance just letting you know, I've been testing this in a PoC we're running and found out the `:global` and `:local` doesn't seem to resolve correctly.

for example, if the source css has the following code
```
:global .someGlobalClass :local .aLocalClass {
  background: red;
}
```

usually with parcel, which I think remix uses for css modules, it gets compiled to something like this
```
.someGlobalClass ._src_styles_module__aLocalClass {
  background: red;
}
```

however in this remix implementation I currently run into something like this, where `.header-module` is just the place where I use the `:global`/`:local` scope
```
:global .header-module__wSWGuW_someGlobalClass :local .header-module__wSWGuW_aLocalClass {
  background: red;
}
```

devongovett:
The `:global` and `:local` pseudo classes are not supported by Parcel CSS without the function syntax. Eg `:global(.someGlobalClass)`

wingleung:
thanks @devongovett! that seems to work idd.

It's odd because I tested it with Parcel 2 and it did seem to compile correctly with just `:global` and `:local`
👉 https://codesandbox.io/s/parcel-2-css-modules-example-forked-tcejyi?file=/src/styles.module.css
parcel-2-css-module

devongovett:
Yes, Parcel 2 falls back to postcss-modules when it detects legacy syntax for backward compatibility.

packages/remix-dev/compiler/plugins/cssModulesPlugin.ts
```diff
@@ -0,0 +1,362 @@
...
+  let classPrefix =
+    path.basename(filePath, path.extname(filePath)).replace(/\./g, "-") + "__";
+  let source = await fse.readFile(filePath);
+
+  let res = parcelTransform({
+    filename: path.relative(config.appDirectory, filePath),
+    code: source,
+    cssModules: true,
+    minify: process.env.NODE_ENV === "production",
+    analyzeDependencies: true,
```
wingleung, line 146: @chaance @devongovett when I use the `url()` function, the value in the `url()` gets hashed but never replaced again with the final result.

**input**
```
.container {
  background-image: url(https://images.unsplash.com/photo-1593288942460-e321b92a6cde)
}
```

**output**
```
.layout-module___9D6bia_container {
    background-image: url(mHxWpW);
}
```

This is because `analyzeDependencies` is set to `true` here but there is no process to replace those hashes with the actual results. [more info](https://github.com/parcel-bundler/parcel-css/blob/master/node/index.d.ts#L20-L26)

Does it need to be enabled? if so, we need to add a process to replace the hashes with the final urls again.

wingleung:
any updates on this? 🙂

chaance:
@wingleung This is on hold as we're working on some bigger compiler updates that may impact our implementation quite a bit. I'm working on a public roadmap for this and other work to provide some transparency here.

mikestopcontinues:
Been a while since the last update. Any news? 👀

chaance:
Closing in favor of https://github.com/remix-run/remix/pull/4852

PR #2493, opened 2022-03-25
fix(remix-dev): add missing deps that got removed during a rebase

mcansh:
lost them in a rebase #2317

Closes: #

- [ ] Docs
- [ ] Tests


packages/remix-eslint-config/internal.js
```diff
@@ -82,5 +82,13 @@ module.exports = {
         "prefer-let/prefer-let": OFF,
       },
     },
+    {
+      files: ["packages/**/*.*"],
+      excludedFiles: "**/__tests__/**/*.*",
+      rules: {
+        // Validate dependencies are listed in workspace package.json files
+        "import/no-extraneous-dependencies": "error",
```
brophdawg11, line 90: This will fail if a workspace file imports something that is not declared in the _workspace_ `package.json`.  This can go unnoticed currently if a workspace imports something that is a dependency of _another_ workspace and thus it exists in the root `node_modules`

brophdawg11: However - worth noting it will not fail in the PR status checks at the moment - we only run `lint --fix` on `main`/`dev` builds so it opens the potential to merge a PR that will break a dev build which is no bueno.  Should we run a subset (or all) lint rules on PRs?

.github/workflows/test.yml
```diff
@@ -41,6 +41,9 @@ jobs:
       - name: Install dependencies
         run: yarn --frozen-lockfile --network-timeout 1000000
 
+      - name: Lint
+        run: yarn lint
```
brophdawg11, line 45: Run `yarn lint` on PRs now - will only fail if ERROR rules fail, and we should only be using ERROR for functional eslint rules.  Stylistic eslint rules should be WARN and will get auto-fixed after merges

packages/remix-cloudflare-pages/package.json
```diff
@@ -14,7 +14,8 @@
     "url": "https://github.com/remix-run/remix/issues"
   },
   "dependencies": {
-    "@remix-run/cloudflare": "1.3.3"
+    "@remix-run/cloudflare": "1.3.3",
+    "@remix-run/server-runtime": "1.3.3"
```
brophdawg11, line 18: Added missing package

packages/remix-dev/package.json
```diff
@@ -16,6 +16,9 @@
     "remix": "cli.js"
   },
   "dependencies": {
+    "@babel/core": "7.17.8",
+    "@babel/plugin-syntax-jsx": "7.16.7",
+    "@babel/preset-typescript": "7.16.7",
```
brophdawg11, line 21: added missing packages

brophdawg11 (approved):

PR #2494, opened 2022-03-25
Bug: Routing failure with accents / non-ascii characters

machour:
Routes with accents or arabic characters seems to throw a 404.

If you encode the accent in the filename (ie, `routée.tsx` => `rout%C3%A9e.tsx`) the page loads, but the client-side is broken.

Bug reported on [discord](https://discord.com/channels/770287896669978684/770287896669978687/956848295127425095)


merelinguist:
I think this is a problem with react router too https://codesandbox.io/s/react-router-forked-pvdsdf?file=/index.js

jacob-ebey:
Has this been filed over in React Router land? This should be fixed over there.

herecydev:
Okay issue is literally this line: https://github.com/pillarjs/send/blob/master/index.js#L514

Dependency chain Remix => express => serve-static => send

Some explanation here and workarounds: https://github.com/pillarjs/send/issues/30#issuecomment-51772908

Does this seem sensible? Happy to rework some of this to mitigate this issue

herecydev:
@jacob-ebey any thoughts on how you want to proceed with this? The issue is two fold I think, we can solve it by enforcing that routes are encoded and the express server would need to be fixed. If accepting decoded routes is desirable then react router needs to be fixed (and possibly other areas).

herecydev:
I think the correct fix definitely lies in fixing the remix-serve package. It didn't occur to me that cloudflare (for example) uses a slightly different setup with assets served by [wrangler](https://github.com/cloudflare/wrangler2). This does work when the routes are encoded.

I think I'm going to put a PR together that replaces the usage of express.static as I've outlined comments that indiciate it's not usable for filesystems where encoding is used.

The `wrangler pages` does not provide any cache headers (which is negligable when running locally), but is this still needed for the `remix-serve` package? Does that actually get used "in production"?

machour:
Fixed

PR #2495, opened 2022-03-25
fix(remix-architect,remix-node): don't parse multipart `formData` to string

### LINKED ISSUE #2476: Architect multipart forms with file uploads result in corrupted files

1.3.3

### Steps to Reproduce

 - Create a new remix project with `npx create-remix`
 - Select "Just the basics"
 - Select "Architect (AWS Lambda)"
 - Select "Typescript"
 - Finish creating the project
 - Replace the contents of `app/routes/index.tsx` with the following code:

```typescript
import {
  ActionFunction,
  Form,
  unstable_createFileUploadHandler,
  unstable_parseMultipartFormData,
} from "remix";

  const uploadHandler = unstable_createFileUploadHandler({
    maxFileSize: 10_000_000,
    directory: "public/uploads",
    file: ({ filename }) => filename,
  });

  const formData = await unstable_parseMultipartFormData(
    request,
    uploadHandler
  );

  const file = formData.get("file-upload") as File;

  if (file) {
    console.log(`File uploaded to server/public/uploads/${file.name}`);
  } else {
    console.log("No file uploaded");
  }

  return {};
};

  return (
    <Form reloadDocument method="post" encType="multipart/form-data">
      <div>
        <label htmlFor="file-upload">
          <span>Upload a file</span>
          <input
            id="file-upload"
            name="file-upload"
            type="file"
          />
        </label>

        <button type="submit">Upload image</button>
      </div>
    </Form>
  );
}
```

- Run the server locally with `npm run dev`
- Open `http://localhost:3333` in your web browser and choose an image file to upload
- Submit the form
- View the file in `<root>/server/public/uploads`

### Expected Behavior

The file should be uploaded uncorrupted

### Actual Behavior

The file that is uploaded isn't the same file as the one selected, has additional data that is "corrupting" the file

hicksy:
A fix to resolve #2476

File uploads using Architect were producing corrupted files. I noted that the byte size was larger than expected on the corrupted file, and confirmed the same was true when using `unstable_createMemoryUploadHandler`. 

If i manually parsed the event body within a modified `createRemixRequest` in /server/node_modules/@remix-run/architect/server.js` I was able to obtain the correct byte length. 

(Documented my initial tests in issue[ #2476](https://github.com/remix-run/remix/issues/2476))

Further investigation, I've found that using `toString()` on the Buffer within `createRemixRequest` and `toString()` on the body when creating the Readable within `parseMultipartFormData` was modifying the buffer (presumably something to do with converting it to utf-8?)

I've modified `createRemixRequest` to do the following:

1. Determine if the event is a formData request (checking for content-type multipart/form-data)
2. if not, handle the base64 event in the current way
3. if it is a formData request, pass the Buffer as event.body rather than converting it to string. 

I've modified `parseMultipartFormData` by removing `toString` from body when creating the Readable. It'll eiether already be a string, or a Buffer

Closes: #2476

- [*] Tests


packages/remix-architect/server.ts
```diff
@@ -69,13 +69,21 @@ export function createRemixRequest(
...
   let search = event.rawQueryString.length ? `?${event.rawQueryString}` : "";
   let url = new URL(event.rawPath + search, `https://${host}`);
-
+  let isFormData = false;
+  if (
+    event.headers["content-type"] &&
+    event.headers["content-type"].includes("multipart/form-data")
+  ) {
+    isFormData = true;
+  }
```
MichaelDeBoey, line 78: This can be simplified to

```suggestion
  let isFormData = event.headers["content-type"]?.includes("multipart/form-data");
```

packages/remix-architect/server.ts
```diff
@@ -69,13 +69,21 @@ export function createRemixRequest(
...
   return new NodeRequest(url.href, {
     method: event.requestContext.http.method,
     headers: createRemixHeaders(event.headers, event.cookies),
     body:
       event.body && event.isBase64Encoded
-        ? Buffer.from(event.body, "base64").toString()
+        ? !isFormData
+          ? Buffer.from(event.body, "base64").toString()
+          : Buffer.from(event.body, "base64")
```
MichaelDeBoey, line 86: Making the condition positive, will create a lower cognitive load.

```suggestion
        ? isFormData
          ? Buffer.from(event.body, "base64")
          : Buffer.from(event.body, "base64").toString()
```

MichaelDeBoey (changes requested):

packages/remix-architect/server.ts
hicksy: Ah yes, of course. I had it that way originally and then switched it out to the longer form to produce an easier to follow example .js file for the issue. 

Much better like this, thanks

packages/remix-architect/server.ts
hicksy: Yes, much easier to read, thanks

hicksy (commented):
Thanks for the feedback. All good

hicksy:
@MichaelDeBoey do I have any more actions to take on this?

MichaelDeBoey (approved):

PR #2500, opened 2022-03-25
fix(remix-dev/cli): Remove excess fetching and improve error handling

chaance:
Instead of fetching early to detect template types in the CLI, we can simplify the code and speed up the setup process with a few heuristics:

1. Check if the user passed a local file. If they hand us an explicit file URL, we'll validate it first. Otherwise we just ping the filesystem to see if the string references a filepath and, if not, move on.
2. If the provided template contains no slashes, spaces, or special chars, we assume it is one of our templates and attempt to fetch it.
3. If the template is a GitHub URL or shorthand :org/:repo, we fetch the repo.
4. Any other URLs are assumed to be direct tarball URLs, so we try to fetch them.

Worth noting that there are probably more opportunities to improve error handling. We'll want to iterate here.

- [ ] Docs
- [x] Tests

### DIFF
--- a/packages/remix-dev/cli/create.ts
+++ b/packages/remix-dev/cli/create.ts
@@ -233,16 +233,55 @@ async function downloadAndExtractTemplateOrExample(
   if (type === "example") {
     name = name.split("/")[1];
   }
+  let typeDir = type + "s";
+  let templateUrl = `https://github.com/remix-run/remix/tree/main/${typeDir}/${name}`;
 
   let response: null | Awaited<ReturnType<typeof fetch>> = null;
+
+  // We need to ping the URL directly to make sure it exists as we fetch from
+  // codeload. attempting to extract the tarball from `main`. We can make these
+  // requests in parallel and reject if HEAD fetch returns a 404
   try {
-    response = await fetch(
-      "https://codeload.github.com/remix-run/remix/tar.gz/main",
-      options.token
-        ? { headers: { Authorization: `token ${options.token}` } }
-        : {}
-    );
-  } catch (_) {}
+    [, response] = await Promise.all([
+      fetch(templateUrl, { method: "HEAD" }).then((response) => {
+        if (response?.status !== 200) {
+          let message: string;
+          switch (response.status) {
+            case 408:
+              message = `🚨 There was a problem fetching the template from GitHub, and the request timed out. Please ensure you are connected to the internet and try again later.`;
+              break;
+            case 404:
+              message = `🚨 The requested template was not found. Valid templates must match the name of a directory inside of the \`templates\` directory in the \`remix-run/remix\` repo on the \`main\` branch.
+
+See https://github.com/remix-run/remix/tree/main/templates for available templates.`;
+              break;
+            default:
+              message = `🚨 Something went wrong when attempting to download the template. The request responded with a ${response.status} status.
+
+${HELP_TEXT}`;
+              break;
+          }
+
+          // throw the message to reject and log the message in `catch`
+          throw message;
+        }
+        return response;
+      }),
+      fetch(
+        "https://codeload.github.com/remix-run/remix/tar.gz/main",
+        options.token
+          ? { headers: { Authorization: `token ${options.token}` } }
+          : {}
+      ),
+    ]);
+  } catch (err) {
+    if (typeof err === "string") {
+      console.error(err);
+      process.exit(1);
+    }
+    // unknown error
+    throw err;
+  }
 
   if (!response || response.status !== 200) {
     console.error(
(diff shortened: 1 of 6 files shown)


packages/remix-dev/cli/create.ts
```diff
@@ -51,9 +64,10 @@ export async function createApp({
   let projectDirIsCurrentDir = relativeProjectDir === "";
   if (!projectDirIsCurrentDir) {
     if (fse.existsSync(projectDir)) {
-      throw new Error(
+      console.error(
```
chaance, line 67: Throwing known errors barfs out a call stack that isn't really useful to users. We know the problem, so let's just tell them and exit. They will open an issue if there's a problem and we can deal with the call stack. May want to pass the `--debug` flag here if they really want it.

packages/remix-dev/cli/create.ts
```diff
@@ -433,24 +544,66 @@ type TemplateType =
...
-  template: string,
-  useTypeScript: boolean,
-  token?: string
-): Promise<TemplateType> {
-  if (template.startsWith("file://") || fse.existsSync(template)) {
-    return "local";
+function detectTemplateType(template: string): TemplateType {
+  // 1. Check if the user passed a local file. If they hand us an explicit file
+  //    URL, we'll validate it first. Otherwise we just ping the filesystem to
+  //    see if the string references a filepath and, if not, move on.
```
brophdawg11, line 550: This ordered flow makes sense to me - nice work 👍

packages/remix-dev/cli/create.ts
mcansh: this technically doesn't have to be used only by the CLI which was the reason for throwing and having the CLI catch them and convert it to a log and process.exit

chaance: Hmm, do we document this somewhere? I don't know that an export should constitute public API but we can chat about that. Happy to change it if we're catching them elsewhere, but currently if a user hits any error they get all the logs.

packages/remix-dev/cli/create.ts
kentcdodds: Agreed 💯

mcansh (approved):

packages/remix-dev/cli/run.ts
```diff
@@ -278,6 +296,8 @@ export async function run(argv: string[] = process.argv.slice(2)) {
 
       await commands.create({
         appTemplate: flags.template ?? answers.appTemplate,
+        templateType:
+          templateType! || (answers.appType === "stack" ? "repo" : "template"),
         projectDir,
         remixVersion: flags.remixVersion,
         installDeps: flags.install ?? answers.install,
```
mcansh, line 303: can you change these to logical ORs and I'll close #2440?
using nullish coalescing prevents our CLI from showing the minimum node version error
```suggestion
        installDeps: flags.install || answers.install,
        useTypeScript: flags.typescript || answers.useTypeScript,
```

packages/remix-dev/cli/run.ts
```diff
@@ -305,7 +305,8 @@ export async function run(argv: string[] = process.argv.slice(2)) {
           templateType! || (answers.appType === "stack" ? "repo" : "template"),
         projectDir,
         remixVersion: flags.remixVersion,
-        installDeps: flags.install ?? answers.install,
+        installDeps: flags.install || answers.install,
+        useTypeScript: flags.typescript || answers.useTypeScript,
         useTypeScript: flags.typescript ?? answers.useTypeScript,
```
mcansh, line 310: heh oops, the commit suggestion didn't let me remove the previous `useTypeScript` 🤦

packages/remix-dev/cli/run.ts
```diff
@@ -1,11 +1,12 @@
 import * as path from "path";
-import meow from "meow";
 import inspector from "inspector";
+import meow from "meow";
 import inquirer from "inquirer";
-// import chalkAnimation from "chalk-animation";
```
kentcdodds, line 5: Could you remove this dep from the lockfile and package.json so we can close #2547?

kiliman:
Hmm.. just noticed that `create-remix@latest` doesn't even announce that you're creating a Remix project. Also, a previous version stopped displaying the version as well.

Any chance we can get those back in the CLI? It's nice knowing what version I'm going to be installing.

kentcdodds:
Once we get this merged, I'm going to be doing a bit more work in the CLI too. I may bring back the animated thing as well.

mcansh:
> Once we get this merged, I'm going to be doing a bit more work in the CLI too. I may bring back the animated thing as well.

sweet! i loved that little touch

brophdawg11:
> I may bring back the animated thing as well.

yes please!! 🙌

chaance:
Note about the animation: I like it as well but it creates some noise for screen-reader users. Want to make sure we ensure that the --no-color flag is respected there.

PR #2508, opened 2022-03-26
fix(docs/tutorials): fix Post links on Admin page in blog tutorial

kaspermroz:
Super-tiny change to the docs - I noticed that in the blog tutorial on Admin page there is an issue with URLs to the blog posts, resulting in a 404 error after clicking any of them (also newly created, which can create confusion and double check if creating posts actually works :) )
Screenshot 2022-03-26 at 19 43 46

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -146,6 +146,7 @@
 - kalch
 - kanermichael
 - karimsan
+- kaspermroz
 - KenanYusuf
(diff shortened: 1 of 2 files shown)


kaspermroz:
Closed it because later on there is actually a homework for fixing it yourself, so I don't want to spoil the fun, sorry! 😅

PR #2513, opened 2022-03-26
feat(remix-cloudflare-pages): add `GetLoadContextFunction` & `RequestHandler` types + remove `@remix-run/server-runtime` dependency

MichaelDeBoey:
Seems like this was forgotten in #2359, so wanted to make `remix-cloudflare-pages` consistent with all other adapter packages.

### DIFF
--- a/packages/remix-cloudflare-pages/worker.ts
+++ b/packages/remix-cloudflare-pages/worker.ts
@@ -1,9 +1,22 @@
-import type { ServerBuild, AppLoadContext } from "@remix-run/server-runtime";
-import { createRequestHandler as createRemixRequestHandler } from "@remix-run/server-runtime";
+import type { AppLoadContext, ServerBuild } from "@remix-run/cloudflare";
+import { createRequestHandler as createRemixRequestHandler } from "@remix-run/cloudflare";
+
+/**
+ * A function that returns the value to use as `context` in route `loader` and
+ * `action` functions.
+ *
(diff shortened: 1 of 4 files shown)


packages/remix-cloudflare-pages/worker.ts
```diff
@@ -1,17 +1,30 @@
...
+
+/**
+ * A function that returns the value to use as `context` in route `loader` and
+ * `action` functions.
+ *
+ * You can think of this as an escape hatch that allows you to pass
+ * environment/platform-specific values through to your loader/action.
+ */
+export type GetLoadContextFunction<Env = any> = (
+  context: EventContext<Env, any, any>
```
pcattori, line 12: Why does `EventContext` need to be parameterized for CF pages but [not for CF workers](https://github.com/remix-run/remix/blob/main/packages/remix-cloudflare-workers/worker.ts#L18)?

pcattori (approved):
LGTM

Just had a small question about `EventContext` type.

packages/remix-cloudflare-pages/worker.ts
MichaelDeBoey: It was already parameterized before, so I just kept it like this.
```ts
getLoadContext?: (context: EventContext<Env, any, any>) => AppLoadContext;
```

Happy to create another PR to do it for `remix-cloudflare-workers` as well if you want me to.

pcattori (approved):

PR #2527, opened 2022-03-28
feat: nightly releases

mcansh:

Closes: #

- [ ] Docs
- [ ] Tests


.github/workflows/release.yml
```diff
@@ -3,10 +3,82 @@ name: release
 on:
   release:
     types: [published]
+  schedule:
+    - cron: "0 21 * * *" # every day at 7PM EST
```
kentcdodds, line 7: Personally I'd put this at 12AM PST

kentcdodds (approved):
Imma fan 👍

.github/workflows/release.yml
mcansh: was just playing with times 😉

kentcdodds:
Do we know why the tests are failing here?

mcansh:
> Do we know why the tests are failing here?

looks unrelated to this PR, but I'll take a peek shortly

kentcdodds (approved):
This is good to go 👍

mcansh:
now the wait until 3 AM to make sure this works just as effortlessly as it did on my test repo 😅

mcansh:
actually, I don't think this will even run until we get it on main

MichaelDeBoey:
@mcansh It appears they _could_ run on other branches as well.
But I _think_ they'll only run on `main` as the `published` event is not a branch-specific event 🤔

> The following steps occur to trigger a workflow run:
> 
> 1. An event occurs on your repository. The event has an associated commit SHA and Git ref.
> 2. GitHub searches the `.github/workflows` directory in your repository for workflow files that are present in the associated commit SHA or Git ref of the event.
> 3. A workflow run is triggered for any workflows that have `on:` values that match the triggering event. Some events also require the workflow file to be present on the default branch of the repository in order to run.
> Each workflow run will use the version of the workflow that is present in the associated commit SHA or Git ref of the event. When a workflow runs, GitHub sets the `GITHUB_SHA` (commit SHA) and `GITHUB_REF` (Git ref) environment variables in the runner environment. For more information, see "[Using environment variables](https://docs.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables)."

https://docs.github.com/en/actions/using-workflows/triggering-a-workflow

kentcdodds:
Ah, good point. We probably should've merged it into `main`. It's no big deal. We can wait until next week.

If it works anyway as @MichaelDeBoey says, then that's a plus. We'll see in a few hours 😅

mcansh:
> Ah, good point. We probably should've merged it into `main`. It's no big deal. We can wait until next week.
> 
> 
> 
> If it works anyway as @MichaelDeBoey says, then that's a plus. We'll see in a few hours 😅

we could always cherry-pick the commit over to main too, shouldn't be too bad on conflicts if there are any.

kentcdodds:
I'm just worried about creating more work for @chaance on releases. It's already a big pain as it is.

PR #2528, opened 2022-03-28
fix(remix-serve): handle os signals by closing dev server

andrewbrey:
Add simple signal handlers to the `remix-serve` cli that will close down the dev server if it receives either a `TERM` or `INT` operating system signal. In the case of a failed shutdown of the dev server, log to stderr.


packages/remix-serve/cli.ts
```diff
@@ -36,9 +36,14 @@ let onListen = () => {
...
 let app = createApp(buildPath);
+let server: ReturnType<typeof app.listen>;
 
 if (process.env.HOST) {
-  app.listen(port, process.env.HOST, onListen);
+  server = app.listen(port, process.env.HOST, onListen);
 } else {
-  app.listen(port, onListen);
+  server = app.listen(port, onListen);
+}
```
MichaelDeBoey, line 45: This could be simplified to

```suggestion
let server = process.env.HOST
  ? app.listen(port, process.env.HOST, onListen)
  : app.listen(port, onListen);
```

packages/remix-serve/cli.ts
```diff
@@ -36,9 +36,14 @@ let onListen = () => {
...
+
+for (let signal of ["SIGTERM", "SIGINT"]) {
+  process.once(signal, () => server?.close(console.error));
 }
```
MichaelDeBoey, line 49:
```suggestion
["SIGTERM", "SIGINT"].forEach((signal) => {
  process.once(signal, () => server?.close(console.error));
});
```

MichaelDeBoey (changes requested):

packages/remix-serve/cli.ts
andrewbrey: Definitely! I was trying to make the fewest changes possible, and this style difference was previously an option too, so I figured it was not using a ternary on purpose - I'll go ahead and push a change for this :+1:

mcansh (approved):
LGTM :chefkiss:

andrewbrey:
@MichaelDeBoey I've pushed up changes for each of your two feedback items - let me know if you have anything else on this one!

MichaelDeBoey (approved):

machour (approved):

jacob-ebey:
Thank you for the contribution :)

PR #2529, opened 2022-03-28
fix(examples): use Node instead of npm for CMD process in `Dockerfile` of `jokes` example

andrewbrey:
In Docker, the `CMD` process is responsible for handling OS signals (including passing them on to child processes) and unfortunately `npm` does not correctly forward OS signals to scripts. This means that when you use an `npm` script as the `CMD` for a docker container, it will not respect `TERM` and `INT` signals and when the container needs to be shutdown, the docker daemon ends up having to force kill it after a timeout (more discussion here for example https://stackoverflow.com/questions/34441975/sigterm-does-not-reach-node-script-when-docker-runs-it-with-bin-sh-c)

The solution is fortunately pretty simple - just don't call an `npm` script in your `CMD`, and instead, use `node` (or something else like `tini` if you _really want to keep using npm scripts_). This PR updates the `CMD` in the `Jokes Tutorial` to use `node` instead of `npm`.

https://user-images.githubusercontent.com/34140052/160462620-53f257cf-62bc-4dc7-a8b5-fc62514b046c.mp4


kentcdodds (commented):
@mrkurt do you have thoughts on this? I think it's probably ok, but just wanted to check.

andrewbrey:
@kentcdodds for additional context, here is a repository from Bret Fisher, one of the Docker Captains, explaining why this is a good "best practice" for node docker containers: https://github.com/BretFisher/node-docker-good-defaults#production-minded-features

andrewbrey:
@kentcdodds one other item of note - I've created a "sibling" PR against `dev` to actually handle these signals in the `remix-serve` cli: https://github.com/remix-run/remix/pull/2528

kentcdodds (approved):
Thanks for putting this together 👍

Issue #1054, opened 2021-12-15
[Bug]: MetaFunction data differs on server vs client


Ehesp:
### What version of Remix are you using?

1.0.6

### What version of Node are you using? Minimum supported version is 14.

16.1.0

### Steps to Reproduce

Throw a 404 within a loader and log out the MetaFunction arguments:

```ts
// routes/foo.tsx
export const loader: LoaderFunction = async () => {
  throw json({ foo: 'bar' }, 404);
};

export const meta: MetaFunction = (args) => {
  console.log(args);
  return {
    title: 'Remix Starter',
    description: 'Welcome to remix!',
  };
};
```

### Expected Behavior

I expected the `data` in the `MetaFunction` to always be `{ foo: 'bar' }` on both client and server. 

### Actual Behavior

When loading the `/foo` page via a server request, the `data` argument of `MetaFunction` is `undefined`:

```js
{
  data: undefined,
  parentsData: { root: null },
  params: {},
  location: {
    pathname: '/foo',
    search: '',
    hash: '',
    state: null,
    key: 'default'
  }
}
```

However, if you client navigate to the `/foo` page, the data comes through as expected:

![image](https://user-images.githubusercontent.com/842078/146197801-5b608a02-cc70-4545-b17a-ecddd795943b.png)

Since I'm providing the data myself, I was experiencing an error accessing my object on the server (cannot read property of undefined), however on client navigation the page loaded as expected with the correct meta information.

I created a [reproduction](https://stackblitz.com/edit/node-xxkqxr?file=app/root.tsx) of this happening.

silvenon:
Also, semi-related, but why does the `meta` function run in the client at all? Am I missing some obvious use case where this would be useful?

giubatt:
I just ran into this bug while doing the Jokes App tutorial

After enabling javascript I tried throwing some errors on loaders to test the error boundaries and noticed this bug. When the route is rendered on the server and an error is thrown on the loader, the meta receives `data` as undefined, but after client-side navigation gets a thrown error on the loader, the meta function receives the thrown error in the `data`.

Is this expected behavior?

#2536 Seems to be on the right track to solve this

dmarkow:
[...]

PR #2536, opened 2022-03-29
fix(remix-react): prevent `CatchValue` & `Error` results to be passed into `MetaFunction`

swwind:
Prevented CatchValue and Error to be put into routeData, which maybe passed into data field in MetaFunction in clients when navigating.

It seems to work on my machine very well.

Closes: #1054

- [ ] Docs
- [ ] Tests

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -273,6 +273,7 @@
 - stephanerangaya
 - SufianBabri
 - supachaidev
+- swwind
 - tascord
 - TheRealAstoo
 - therealflyingcoder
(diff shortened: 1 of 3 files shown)


swwind:
For errors and catches, server uses `undefined` but client uses `CatchValue | Error` in MetaFunction data field, which is pretty weird.

This PR seems to break the old tests, but I think we can provide `CatchValue` and `Error` in another way to avoid writing

```ts
export const meta: MetaFunction = ({ data }: { data?: LoaderData | CatchValue | Error }) => {}
```

in every page. I think `{ data?: LoaderData }` is enough for generating all the useful meta data for the certain page, and we can handle error meta or catch meta in other places (or just ignore them).

chaance:
We've since removed the Remix transition manager as we migrated to React Router v6.4. If this is still an issue, please open a new issue and/or PR in the React Router repo.

Issue #2503, opened 2022-03-25
createCookieSessionStorage is no longer exported server-runtime in v1.3.3


garth:
### What version of Remix are you using?

v1.3.3

### Steps to Reproduce

Upgrade from v1.3.2 where server-runtime exported createCookieSessionStorage and now it is no longer there.

### Expected Behavior

It should still be exported.

### Actual Behavior

x

bndkt:
Same for createCloudflareKVSessionStorage, since upgrading from 1.3.2 to 1.3.3 it throws with:

`No matching export in "node_modules/@remix-run/cloudflare-pages/esm/index.js" for import "createCloudflareKVSessionStorage"`

JulianBuse:
I have this issue as well

pcattori:
We expect users to import from one of our concrete server runtime packages (i.e. `@remix-run/node` or `@remix-run/cloudflare`) rather than import directly from `@remix-run/server-runtime`.

We don't currently consider `@remix-run/server-runtime` to be an intended public API for users, so we refactored it. Specifically we replaced the functions that depend on internal crypto `sign`/`unsign` globals (like `createCookieSessionStorage` and `createCloudflareKVSessionStorage`) with factory functions that explicitly take crypto implementations and give you back the function (for example, [see this code](https://github.com/remix-run/remix/blob/dev/packages/remix-node/implementations.ts) in `@remix-run/node`).

---

If you have a specific use-case that somehow requires importing directly from `@remix-run/server-runtime` rather than one of our concrete server runtime packages, feel free to reopen.

garth:
@pcattori for everyone who cannot use the `remix` package directly (due to it modifying node_modules), the advice we got was to import from the packages behind that have the actually implementation. In this case it was `@remix-run/server-runtime`.

if I update the import from `@remix-run/server-runtime` to `@remix-run/node` I get the following error:

```
export const sessionStorage = createCookieSessionStorage({
                               ^
TypeError: (0 , import_server_runtime.createCookieSessionStorage) is not a function
```

pcattori:
@garth Previously we recommended importing from `remix` via "magic" imports that relied on modifying `node_modules/`, but we're moving away from that in favor of importing directly from server runtime packages (see #2359). We're actively working on updating our docs, examples, etc... to use this new recommended approach.

Could you provide more info about the error you're seeing e.g. what your usage looks like and what file the error points to?

pcattori:
[...]

PR #2537, opened 2022-03-29
docs: import from @remix-run/node instead of remix

machour:
Got a `createCookie() is not a function` today after upgrading my app to 1.3.3

According to https://github.com/remix-run/remix/issues/2503#issuecomment-1079963155, we should not import from `remix` or `@remix-run/server-runtime` anymore.

### DIFF
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -1726,7 +1726,7 @@ Let's say you have a banner on your e-commerce site that prompts users to check 
 First, create a cookie:
 
 ```js filename=app/cookies.js
-import { createCookie } from "remix";
+import { createCookie } from "@remix-run/node";
 
 export const userPrefs = createCookie("user-prefs", {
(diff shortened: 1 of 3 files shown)


MichaelDeBoey:
This should still be possible if you're not using PNPM/Yarn P'n'P, since the only recommended import has always been from the `remix` package with the `magicExports`.
For people using PNPM/Yarn P'n'P, it was indeed said to use `@remix-run/server-runtime` as an escape hatch.

However, that package (along with all other packages) was never meant to be a public (non-breaking) API.
This has changed since #2359 was released, but  the `remix` package is still backwards compatible.

Once @pcattori & I are finished with #2430, we'll also run the codemod on our `examples` & have a script ready to run the codemod on our docs too.

I think it's best to back off this PR for now, so we keep our docs consistent.

machour:
Ok.....

PR #2542, opened 2022-03-29
feat(remix-cloudflare): rename `createCloudflareKVSessionStorage` to `createWorkersKVSessionStorage`

mjackson:
Rename to `createWorkersKVSessionStorage` for consistency with the "Workers KV" product name.

### DIFF
--- a/packages/remix-cloudflare/index.ts
+++ b/packages/remix-cloudflare/index.ts
@@ -1,10 +1,22 @@
 import "./globals";
 
-export {
+import {createWorkersKVSessionStorage} from './sessions/workersKVStorage';
+
+const warn = <T extends Function>(fn: T, message: string): T =>
+  ((...args: unknown[]) => {
+    console.warn(message);
+
+    return fn(...args);
+  }) as unknown as T;
+
+
+/** @deprecated Use `createWorkersKVSessionStorage` instead. */
+export const createCloudflareKVSessionStorage = warn(
   createWorkersKVSessionStorage,
-  // TODO: Deprecate createCloudflareKVSessionStorage
-  createWorkersKVSessionStorage as createCloudflareKVSessionStorage,
-} from "./sessions/workersKVStorage";
+  "`createCloudflareKVSessionStorage` is deprecated. Please use `createWorkersKVSessionStorage` instead.",
+);
+
+export { createWorkersKVSessionStorage } from "./sessions/workersKVStorage";
 
 export {
   createCookie,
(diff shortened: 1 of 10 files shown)


packages/remix-cloudflare/index.ts
```diff
@@ -1,6 +1,10 @@
 import "./globals";
 
-export { createCloudflareKVSessionStorage } from "./sessions/cloudflareKVSessionStorage";
+export {
+  createWorkersKVSessionStorage,
+  // TODO: Deprecate createCloudflareKVSessionStorage
+  createWorkersKVSessionStorage as createCloudflareKVSessionStorage,
```
pcattori, line 6: I think it makes sense to have `createCloudflareKVSessionStorage` be a wrapper function for `createWorkersKVSessionStorage` that just `console.warn`s before delegating:

```ts
function deprecate<T extends Function>(fn: T, message: string): T {
    return <any>function(...args) {
        console.warn(message)
        return fn(...args)
    };
}

const createCloudflareKVSessionStorage = deprecate(createWorkersKVSessionStorage, "`createCloudflareKVSessionStorage` is deprecated. Use `createWorkersKVSessionStorage` instead.")
```

pcattori (changes requested):
Can we deprecate `createCloudflareKVSessionStorage` as part of this PR?

packages/remix-cloudflare/index.ts
MichaelDeBoey: Adding a `@deprecated` annotation to it would be great as well I think

```suggestion
function warn<T extends Function>(fn: T, message: string): T {
    return <any>function(...args) {
        console.warn(message)
        return fn(...args)
    };
}

/**
 * @deprecated Use `createWorkersKVSessionStorage` instead.
 */
const createCloudflareKVSessionStorage = warn(
  createWorkersKVSessionStorage,
  "`createCloudflareKVSessionStorage` is deprecated. Use `createWorkersKVSessionStorage` instead.",
)
```

MichaelDeBoey:
@mjackson It seems like we have some conflicts to resolve here

MichaelDeBoey:
@mjackson I took the liberty of resolving all conversations here, so we can get this one merged

mcansh (approved):

MichaelDeBoey:
@mcansh @pcattori Anything that's preventing us from merging this one?

MichaelDeBoey:
@mcansh @pcattori Anything that's preventing us from merging this one?

pcattori (approved):

PR #2548, opened 2022-03-29
docs(examples): add Docker example

BasharOsman:
Create simple example to run remix with docker-compse


examples/docker/README.md
```diff
@@ -0,0 +1,29 @@
+# Run Remix With docker-compose
+
+very simple example to run remix app with docker and docker-compose
+
+## Preview
```
kentcdodds, line 5: Probably can't do a preview of this one so we can remove this section.

examples/docker/README.md
```diff
@@ -0,0 +1,29 @@
+# Run Remix With docker-compose
+
+very simple example to run remix app with docker and docker-compose
```
kentcdodds, line 3: Could you explain here why someone would want to do this?

kentcdodds (changes requested):
Nice! Thanks for this example.

Just a few bits of feedback

examples/docker/README.md
MichaelDeBoey: If so, the `sandbox.config.json` should probably be removed too.

examples/docker/README.md
BasharOsman: ok i will add:

**With this example Remix can be run in any server have docker.**

examples/docker/README.md
BasharOsman: i will remove `sandbox.config.json` and will add:

This example don't have Preview

to Preview section

examples/docker/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Preview
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+[![Open in CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/docker)
+
+## How to use
+
+all what you need to do is to run the following command:
```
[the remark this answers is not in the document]
BasharOsman: i will change this too

examples/docker/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:16-alpine
```
c43721, line 1: If this is meant to be built for production, there's some suggestions here:

1) You should be specific on the version of node, such as `node:16.14.2-alpine`. 
This prevents possible regressions (if any occurred). There's some other arguments to be made here, so take this as you wish.

2) You should keep out dev-dependencies
Docker builds should be as slim as possible, and having all your files (including your .git, source and dist files) is not a good practice. 

3) You should switch to a non-root user, in this case node provides a `node` user.
Sandboxing!

I suggest this be switched to a more complex, but "production-hardened" dockerfile with another docker/compose file with development in mind. That way, you can have a secure but useful dockerfile for your production, but then have something a bit relaxed for development.

Maybe there's an argument to have a basic setup like this, but I think providing more than this will be super beneficial to everyone.

examples/docker/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+
+services:
+  remix:
+    #command: sh -c "npm rebuild esbuild" if ther any problem with esbuild try this
+    restart: unless-stopped
+    environment:
+      NODE_ENV: production #production or development
+    ports:
+      - 3000:3000
+    build:
```
c43721, line 11: This could be simplified to just `build: .` but may be needed if we separate the dev versus production Dockerfiles

examples/docker/.dockerignore
```diff
@@ -0,0 +1,4 @@
+**/build
+**/.cache
+**/node_modules
+**/npm-debug.log
```
c43721, line 4: You'll likely want to remove `.git` folder as well

c43721 (changes requested):
I'll also add some comments based on my docker experience. Most of these are nits and or strong suggestions, but I think this Docker example could be updated to accomondate more realistic Dockerfiles.

examples/docker/README.md
```diff
@@ -0,0 +1,29 @@
+# Run Remix With docker-compose
+
+Very simple example to run remix app with docker and docker-compose. 
+
+**With this example Remix can be run in any server have docker.**
```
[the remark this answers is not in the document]
BasharOsman: i don't know how to explain it but i will explain my use case:
if you have full control on server and want to run node js application you need to install nodejs and then you will need to test the the application in the new environment (like if you work on windows and the server is ubuntu or mac) so you need to spend time in this.
and after this you need to run application using PROCESS MANAGER like pm2,
all this can be solved by using docker. or this what i use docker for it

examples/docker/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Preview
+
+This example don't have Preview
+
```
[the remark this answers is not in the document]
BasharOsman: ok i will delete it

examples/docker/Dockerfile
BasharOsman: 1. I didn't think this would be a problem but it is better to specify the version. 
2. i think yarn install --production will solve this, 

> Yarn will not install any package listed in devDependencies if the NODE_ENV environment variable is set to production

4. for this:
```
RUN addgroup -g 1001 -S nodejs
RUN adduser -S remix -u 1001
USER remix 
```
add this code to dockerfile will be good ? 

i am not docker expert but i like use it and this my first time to contributing in project. so when i saw the examples and there no docker example i think why i don't add it. i want this example to be very simple.
i tryed to find how to write Docker for dev and Live Reloading the code but i could not find solution. all what i found is to use Volumes and did not work with me :(

examples/docker/.dockerignore
BasharOsman: i will add `.git` too

BasharOsman:
> 

oh I have not seen this examples, but i think there should be more simple example or example for any project not specified one. so the remix can be run in any environment : ).
i want it to be for prodction and development example. 
but didn't find a good example for development :(.

i would like to find official remix image on docker hub

chaance:

PR #2550, opened 2022-03-30
bug: <Form> submit changes url immediately upon submission

smerchek:
Submitting a normal `<form>` does not have this behavior. I'd expect the url to be unchanged until the action responds with a redirect or other response.

- [ ] Docs
- [X] Tests


zipme:
Any updates? Using `useFetch` doesn't have this problem though.

machour:
@mcansh will this be fixed by the change in 6.4 described by Ryan here? https://github.com/remix-run/remix/pull/2110#issuecomment-1180806310 🤔

kiliman:
@zipme Unlike forms, fetchers are **not** navigations, so they will never change the URL.

chaance:
If this is still an issue it should be fixed and tested in React Router. Mind checking that out and opening a new issue there if needed?

PR #2556, opened 2022-03-30
chore(remix-dev): always ask if they want ts or js during create

mcansh:
This PR re-adds the question during `create-remix` if the user wants to convert their project from TypeScript to JavaScript, it will first create the project like before, and then if a tsconfig.json file exists, it will prompt if you want to convert it to JavaScript.

I've also reshuffled around our node check and swapped `fs/promises` for `fs-extra` so that the node check will actually work

x-ref: #2317 

Closes: REM-927

- [ ] Docs
- [ ] Tests


MichaelDeBoey (approved):

kentcdodds (commented):
Could we pass this choice to the remix init script so templates can do their own updates as well (for example, we have GitHub actions that do type checking that should be removed in this case).

mcansh:
> Could we pass this choice to the remix init script so templates can do their own updates as well (for example, we have GitHub actions that do type checking that should be removed in this case).

that gets tricky due to createApp not requiring the to be ran by the CLI, but our tests no longer use createApp directly and we don't mention anywhere that you can use it, so im not personally opposed to bringing inquirer over there... but something to think about

mcansh:
> Could we pass this choice to the remix init script so templates can do their own updates as well (for example, we have GitHub actions that do type checking that should be removed in this case).

yeah i think we can move a little around to make this work, almost there, just ran into an issue with our new tsconfig path loader

mcansh:
> Could we pass this choice to the remix init script so templates can do their own updates as well (for example, we have GitHub actions that do type checking that should be removed in this case).

done!

mcansh:
I'll cut an experimental in a few minutes

mcansh:
https://user-images.githubusercontent.com/11698668/162055747-ddefd50d-eba0-4ef3-9d03-653c04115fd9.mp4

kentcdodds (approved):
Super 👍

kentcdodds (changes requested):
Could we get a test of this functionality please?

mcansh:
> Could we get a test of this functionality please?

now that we can test the interactive bits, of course you can :)

mcansh:
> Could we get a test of this functionality please?

do we want a test for remix.init with `isTypeScript` and `isDeno`?

MichaelDeBoey (approved):

MichaelDeBoey:
> do we want a test for remix.init with `isTypeScript` and `isDeno`?

You could maybe add a test to check if these values are passed correctly?
We can't really test the functions itself I think 🤔

pcattori:
To be clear, I'm not asking that we purposefully ignore Deno. Rather, that detecting both that (1) an app is a Deno app AND (2) that its using Typescript is tricky. Additionally, seems like Deno community leans more towards TS anyway. So the suggestion is to not bother handling Deno specially.

packages/remix-dev/cli/run.ts
```diff
@@ -315,15 +313,79 @@ export async function run(argv: string[] = process.argv.slice(2)) {
...
-        installDeps: flags.install !== false && answers.install !== false,
-        useTypeScript:
-          flags.typescript !== false && answers.useTypeScript !== false,
+        installDeps,
+        useTypeScript: flags.typescript !== false,
         githubToken: process.env.GITHUB_TOKEN,
       });
+
+      let isTypeScript = fse.existsSync(path.join(projectDir, "tsconfig.json"));
+      let isDeno = fse.existsSync(path.join(projectDir, "deno.json"));
```
pcattori, line 328: `isDeno` won't reliably detect Deno since [Deno's configuration file can be named anything](https://deno.land/manual/getting_started/configuration_file).

`isTypescript` will be `false` for Deno projects, but I think that's ok for now. We just don't support TS->JS conversion for Deno.

mcansh: ok cool, I think I misunderstood the original message from Pedro, sorry! aside, they can still convert if they _really_ want to by using `--no-typescript`

packages/remix-dev/cli/convert-to-javascript.ts
```diff
@@ -0,0 +1,91 @@
...
+      })
+    ),
+  };
+  // 4. Remove typecheck npm script from package.json
+  if (pkg.scripts && pkg.scripts.typecheck) {
+    delete pkg.scripts.typecheck;
+  }
+  fse.writeJSONSync(path.join(projectDir, "package.json"), newPackageJson, {
+    spaces: 2,
+  });
```
pcattori, line 90: Does this reorder `package.json`? E.g. if I had stuff after `devDependencies`, will `...pkg, devDependencies: ...` cause `devDependencies` to show up last? That could create a lot of `git` noise. Maybe mutating `pkg` (as much as I normally prefer immutable stuff) would be better if it preserves key ordering in `package.json`.

packages/remix-dev/cli/convert-to-javascript.ts
```diff
@@ -0,0 +1,91 @@
...
+    devDependencies: Object.fromEntries(
+      Object.entries(devDeps).filter(([name]) => {
+        return !name.startsWith("@types/") && name !== "typescript";
+      })
+    ),
+  };
+  // 4. Remove typecheck npm script from package.json
+  if (pkg.scripts && pkg.scripts.typecheck) {
+    delete pkg.scripts.typecheck;
+  }
```
pcattori, line 87: This doesn't remove the `typecheck` script from `newPackageJson`, so does the removal of the `typecheck` script ever get written to `package.json`?

packages/remix-dev/cli/convert-to-javascript.ts
mcansh: yeah we probably want to mutate or re-sort it

kentcdodds: I think we use https://www.npmjs.com/package/sort-package-json in sooner other places

packages/remix-dev/cli/convert-to-javascript.ts
mcansh: it does now after [dde9799](https://github.com/remix-run/remix/pull/2556/commits/dde9799390d844a3488a683b5cb50e92a785645e), but it seems to have still removed it using the 1.3.4 on the blues stack with `--no-typescript` - but looks like we should maybe also check each script to see if one of them also calls "typecheck" like the blues stack does `"validate": "run-p \"test -- --run\" lint typecheck test:e2e:run"`

kentcdodds: I think we can let the remix.init script handle that. It's impossible to handle every case for stacks authors, so we just do our best here and then stacks authors can do the rest

packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -179,8 +179,33 @@ describe("remix CLI", () => {
...
+        ],
+        { stdio: [null, null, null] }
+      );
+
+      await interactWithShell(proc, [
+        { question: /Where.*create.*app/i, type: [projectDir, ENTER] },
+        { question: /What type of app/i, answer: /basics/i },
+        { question: /Where.*deploy/i, answer: /express/i },
+        { question: /install/i, type: ["n", ENTER] },
+        { question: /typescript or javascript/i, answer: /javascript/i },
```
kentcdodds, line 208: Let's add some assertions to this one so we can make sure the conversion works

PR #2530, opened 2022-03-28
feat(cloudflare+cloudflare-pages+cloudflare-workers+templates): add support for `@cloudflare/workers-types` v3

edmundhung:
Follow up for #645 

I made a mistake and deleted the fork repository and now I cannot push any commit to the original PR. So I open a new one here:

> Cloudflare released an [update](https://github.com/cloudflare/workers-types/releases/tag/v3.0.0) for `@cloudflare/workers-types` last month. It is now automatically generated from the source as announced [here](https://blog.cloudflare.com/automatically-generated-types/) and this ensures typing to be always up-to-dated with the latest runtime.
>
> However, the current CF adapter has a peerDependency of v2 only. It would be nice if we could add support of v3 as well.
This will also makes experimenting with new ESM worker easier (as all new types are only defined on v3) before official support is ready. 
>
> Thanks for all the great work!

- This PR does not touch the `cloudflare-pages` adapter as it is already using `@cloudflare/workers-types` [v3](https://github.com/remix-run/remix/blob/dev/packages/remix-cloudflare-pages/package.json#L21).
- I have updated the `devDependencies` to v3 now so the tsconfig has to be updated as well as stated on the [release note](https://github.com/cloudflare/workers-types/releases/tag/v3.0.0) 

Closes #2504

- [ ] Docs
- [ ] Tests

PR #2565, opened 2022-03-31
fix: add `peerDependencies` to packages

illright:
Related to https://github.com/remix-run/blues-stack/issues/37

- [ ] Docs
- [ ] Tests

The repository packages had a few peer dependency issues which would **spam red** whenever any of these packages were installed through a package manager that cares about peer dependencies (like any package manager should).


packages/remix-cloudflare-pages/package.json
```diff
@@ -15,7 +15,9 @@
   },
   "dependencies": {
     "@remix-run/cloudflare": "1.3.3",
-    "@remix-run/server-runtime": "1.3.3"
+    "@remix-run/server-runtime": "1.3.3",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0"
```
MichaelDeBoey, line 20: Let's move these to `devDependencies` & add them to `peerDependencies` as well.

packages/remix-cloudflare/package.json
```diff
@@ -13,12 +13,14 @@
   },
   "dependencies": {
     "@cloudflare/kv-asset-handler": "^0.1.3",
-    "@remix-run/server-runtime": "1.3.3"
+    "@remix-run/server-runtime": "1.3.3",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0"
```
MichaelDeBoey, line 18: Let's move these to `devDependencies` & add them to `peerDependencies` as well.

packages/remix-node/package.json
```diff
@@ -22,6 +22,8 @@
     "cookie-signature": "^1.1.0",
     "form-data": "^4.0.0",
     "node-fetch": "^2.6.1",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
```
MichaelDeBoey, line 26: Let's move these to `devDependencies` & add them to `peerDependencies` as well.

packages/remix-dev/package.json
```diff
@@ -42,6 +42,8 @@
     "node-fetch": "^2.6.7",
     "prettier": "2.6.1",
     "pretty-ms": "^7.0.1",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
```
MichaelDeBoey, line 46: Let's move these to `devDependencies` & add them to `peerDependencies` as well.

packages/remix-cloudflare-workers/package.json
```diff
@@ -18,9 +18,9 @@
     "@remix-run/cloudflare": "1.3.3"
   },
   "peerDependencies": {
-    "@cloudflare/workers-types": "^2.2.2"
+    "@cloudflare/workers-types": "^3.2.0"
```
MichaelDeBoey, line 21: This would be a breaking change & is already handled by #2530

```suggestion
    "@cloudflare/workers-types": "^2.2.2"
```

packages/remix-cloudflare-workers/package.json
```diff
@@ -18,9 +18,9 @@
...
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^2.2.2"
+    "@cloudflare/workers-types": "^3.2.0"
```
MichaelDeBoey, line 24:
```suggestion
    "@cloudflare/workers-types": "^2.2.2"
```

packages/remix-cloudflare/package.json
```diff
@@ -13,12 +13,14 @@
...
   },
   "peerDependencies": {
-    "@cloudflare/workers-types": "^2.2.2"
+    "@cloudflare/workers-types": "^3.2.0"
```
MichaelDeBoey, line 21:
```suggestion
    "@cloudflare/workers-types": "^2.2.2"
```

packages/remix-cloudflare/package.json
```diff
@@ -13,12 +13,14 @@
...
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^2.2.2"
+    "@cloudflare/workers-types": "^3.2.0"
```
MichaelDeBoey, line 24:
```suggestion
    "@cloudflare/workers-types": "^2.2.2"
```

MichaelDeBoey (changes requested):

MichaelDeBoey:
@illright Could you please rebase your branch onto latest `dev` & resolve conflicts?

MichaelDeBoey:
Just thought about this one a bit more and the main problem is actually that `@remix-run/server-runtime` has `react` & `react-dom` in `peerDependencies`.

Since it feels like the server-runtime should be (frontend)framework agnostic, I think we should be move these packages to `devDependencies`, but that requires us to remove the usage of `react-router-dom`'s `matchRoutes` function and I'm not sure if that's preferred tbh. 🤔
https://github.com/remix-run/remix/blob/29af991b939d1a9efd654028179694dfd5e7a842/packages/remix-server-runtime/routeMatching.ts#L16

CC/ @ryanflorence @mjackson

MichaelDeBoey:
Discussed this with @ryanflorence and it seems the team will work towards a `@remix-run/router` package, which would include `matchRoutes` and won't include any DOM lib dependencies.

Once that package is available, we can see if we still need to do some things for your favorite package manager.

PR #2568, opened 2022-03-31
docs(loader): remove section about returning plain objects

machour:
Updated the documentation to remove bare objects returning from the loader.


kentcdodds (approved):
Stellar, thanks!

PR #2569, opened 2022-03-31
chore: use external server for express template

jacob-ebey:
If Remix bundles your server entry point:
- you get to use server.ts
- you get to share bundled deps between your server and app folder code
- have to use virtual module id to get the server build
- have to use a process manager, like nodemon or pm2 in dev
- have to reconnect to your database each time you change a file in dev

If your server.js is separate:
- you get to avoid side effects because server.js only runs once (e.g. pass db connections through using getLoadContext)
- have to require('./build') to get the server build
- changes surface to the browser quicker

Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/templates/express/server.js
+++ b/templates/express/server.js
@@ -28,15 +28,15 @@ app.all(
   "*",
   process.env.NODE_ENV === "development"
     ? (req, res, next) => {
-        purgeRequireCache(buildPath);
+        purgeRequireCache();
 
         return createRequestHandler({
-          build: require(buildPath),
+          build: require(BUILD_DIR),
           mode: process.env.NODE_ENV,
         })(req, res, next);
       }
     : createRequestHandler({
-        build: require(buildPath),
+        build: require(BUILD_DIR),
         mode: process.env.NODE_ENV,
       })
 );
--- a/templates/express/server.js
+++ b/templates/express/server.js
@@ -46,6 +46,15 @@ app.listen(port, () => {
   console.log(`Express server listening on port ${port}`);
 });
 
-function purgeRequireCache(path) {
-  delete require.cache[require.resolve(path)];
+function purgeRequireCache() {
+  // purge require cache on requests for "server side HMR" this won't let
+  // you have in-memory objects between requests in development,
+  // alternatively you can set up nodemon/pm2-dev to restart the server on
+  // file changes, but then you'll have to reconnect to databases/etc on each
+  // change. We prefer the DX of this, so we've included it for you by default
+  for (const key in require.cache) {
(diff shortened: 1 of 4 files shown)


templates/express/server.js
```diff
@@ -25,14 +26,26 @@ app.use(morgan("tiny"));
...
-
 const port = process.env.PORT || 3000;
 
 app.listen(port, () => {
   console.log(`Express server listening on port ${port}`);
 });
+
+function purgeRequireCache(path) {
+  delete require.cache[require.resolve(path)];
+}
```
kentcdodds, line 51:
```suggestion
function purgeRequireCache() {
  // purge require cache on requests for "server side HMR" this won't let
  // you have in-memory objects between requests in development,
  // alternatively you can set up nodemon/pm2-dev to restart the server on
  // file changes, but then you'll have to reconnect to databases/etc on each
  // change. We prefer the DX of this, so we've included it for you by default
  for (const key in require.cache) {
    if (key.startsWith(BUILD_DIR)) {
      delete require.cache[key];
    }
  }
}
```

templates/express/server.js
```diff
@@ -1,8 +1,9 @@
...
-import compression from "compression";
-import morgan from "morgan";
-import { createRequestHandler } from "@remix-run/express";
-import * as serverBuild from "@remix-run/dev/server-build";
+const express = require("express");
+const compression = require("compression");
+const morgan = require("morgan");
+const { createRequestHandler } = require("@remix-run/express");
+
+const buildPath = "./build";
```
kentcdodds, line 6:
```suggestion
const BUILD_DIR = path.join(process.cwd(), "build");
```

templates/express/server.js
```diff
@@ -25,14 +26,26 @@ app.use(morgan("tiny"));
...
+    ? (req, res, next) => {
+        purgeRequireCache(buildPath);
+
+        return createRequestHandler({
+          build: require(buildPath),
+          mode: process.env.NODE_ENV,
+        })(req, res, next);
+      }
+    : createRequestHandler({
+        build: require(buildPath),
```
kentcdodds, line 39:
```suggestion
        purgeRequireCache();

        return createRequestHandler({
          build: require(BUILD_DIR),
          mode: process.env.NODE_ENV,
        })(req, res, next);
      }
    : createRequestHandler({
        build: require(BUILD_DIR),
```

kentcdodds (commented):
Just a few small changes to make it more like the blues stack 👍

kentcdodds (approved):
Looks good 👍

PR #2571, opened 2022-03-31
fix: `HtmlMetaDescriptor` types

chaance:
The `HtmlMetaDescriptor` is typed incorrectly since we added support for object values. 

Given the existing type:

```ts
export interface HtmlMetaDescriptor {
  charset?: "utf-8";
  charSet?: "utf-8";
  title?: string;
  [name: string]:
    | null
    | string
    | string[]
    | undefined
    | Record<string, string>
    | Record<string, string>[];
}
```

A meta object can either have an array of strings or an array of objects, but not an array of both. However our implementation allows for that:

```ts
export function meta() {
  return {
    "og:type": [
      // this is totally fine
      "og:locale:alternate",
      {
        "property": "og:locale:alternate",
        "content": "es_ES",
      },
    ];
  };
}
```

Most users won't do this, of course, but I noticed that it breaks my `remix-seo` package, as I'm trying to push a value into an array that can either be an object or a string, but the value is always a string and `Array.prototype.push` expects an intersection of the possible types:

screenshot of the TS error: Argument of type 'string' is not assignable to parameter of type 'string & Record<string, string>'

### DIFF
--- a/packages/remix-react/routeModules.ts
+++ b/packages/remix-react/routeModules.ts
@@ -77,10 +77,9 @@ export interface HtmlMetaDescriptor {
   [name: string]:
     | null
     | string
-    | string[]
     | undefined
     | Record<string, string>
-    | Record<string, string>[];
+    | Array<Record<string, string> | string>;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


chaance:
Tests are failing on dev at the moment which is why they are failing here.

kentcdodds (approved):
Thanks!

PR #2581, opened 2022-04-01
fix(remix-react): add missing hash when redirecting client side

### LINKED ISSUE #1641: [Bug]: Cannot add hash to redirect url

- `remix` (Remix core)

1.1.3

16.13.0

### Steps to Reproduce

Add a hash to your loader function or action function redirect. It never reaches the url bar.

```typescript
export const loader = () => {
   return redirect('/#hash');
}
```

### Expected Behavior

The hash that is added to the redirect url should take effect in the browser and be accessible via the `useLocation` hook.

### Actual Behavior

The hash never makes it to the browser and cannot be accessed via the `useLocation` hook.

### Note

SearchParams work just fine. It's just the hash that is not working.

machour:
Closes #1641

### DIFF
--- a/integration/transition-test.ts
+++ b/integration/transition-test.ts
@@ -205,7 +205,6 @@ describe("rendering", () => {
   test("loader redirect with hash", async () => {
     await app.goto("/");
 
-    let responses = app.collectDataResponses();
     await app.clickLink(`/${REDIRECT_HASH}`);
 
     let url = new URL(app.page.url());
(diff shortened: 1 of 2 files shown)


integration/transition-test.ts
```diff
@@ -190,6 +202,17 @@ describe("rendering", () => {
     expect(html).toMatch(PAGE_INDEX_TEXT);
   });
 
+  test("loader redirect with hash", async () => {
+    await app.goto("/");
+
+    let responses = app.collectDataResponses();
```
MichaelDeBoey, line 208: It seems like the response isn't used, so we could just do the call without assigning it I guess

```suggestion
    app.collectDataResponses();
```

MichaelDeBoey (changes requested):
One tiny remark, but awesome first code PR overall! 👊

integration/transition-test.ts
machour: That was a copy/paste leftover, thanks for noticing !

PR #2587, opened 2022-04-01
refactor(remix-dev): `migrate` command instead of `codemod`

pcattori:
"codemod" and "transform" both correspond to changing source code (i.e. `.js`,`.jsx`,`.ts`,`.tsx`) whereas we want to also change app dependencies, config, etc... more apt terms are "migrate" and "migration".

Also, `replace-remix-imports` is a more apt name than `update-remix-imports`.
Also, using a flag (`-m`/`--migration`) borrows from the `jscodeshift` conventions and gives the CLI the ability to accept a project directory path without a migration, which will then prompt the user to interactively pick a migration.

Before:
```sh
npx @remix-run/dev codemod [TRANSFORM] [path/to/project]
```

After
```sh
npx @remix-run/dev migrate [-m/--migration MIGRATION] [path/to/project]
```

Closes: #

- [ ] Docs
- [ ] Tests


packages/remix-dev/cli/migrate/questions.ts
```diff
@@ -1,29 +1,29 @@
...
 
 export type Answers = {
   projectDir: string;
-  transform: string;
+  migration: string;
 };
 
 type QuestionsArgs = {
-  input: { projectDir: string; transform: string };
+  input: { projectDir: string; migration?: string };
```
MichaelDeBoey, line 11: Should have done this in my PR, but `projectDir` is optional too.

```suggestion
  input: { projectDir?: string; migration?: string };
```

packages/remix-dev/cli/migrate/run/index.ts
```diff
@@ -1,24 +1,24 @@
...
 import { updateRemixImports } from "./transforms";
 import { validateAnswers } from "./validate-answers";
 
 const transformFunctionByName: Record<
-  typeof transformOptions[number]["value"],
+  typeof migrationOptions[number]["value"],
   Transform
 > = {
-  "update-remix-imports": updateRemixImports,
+  "replace-remix-imports": updateRemixImports,
```
MichaelDeBoey, line 11: Probably a good idea to rename `updateRemixImports` to `replaceRemixImports` as well?

rollup.config.js
```diff
@@ -219,10 +219,10 @@ function remixDev() {
     },
     {
       external: (id) => isBareModuleId(id),
-      input: [`${sourceDir}/cli/codemod/run/transforms/index.ts`],
+      input: [`${sourceDir}/cli/migrate/run/transforms/index.ts`],
       output: {
-        banner: createBanner("@remix-run/codemod", version),
-        dir: `${outputDir}/cli/codemod/run/transforms`,
+        banner: createBanner("@remix-run/migrate", version),
```
MichaelDeBoey, line 224: This should have been done in my PR as well.
Probably an oversight when copy/pasting the config I had for the separate package 🙈

```suggestion
        banner: createBanner("@remix-run/dev", version),
```

MichaelDeBoey (changes requested):

packages/remix-dev/cli/migrate/run/index.ts
pcattori: Right now I just wanted to update the user-facing stuff. Will put up another PR later with some more refactoring to internals including this sort of internal renaming^.

packages/remix-dev/cli/migrate/questions.ts
pcattori: Actually, `meow` will not give `undefined` for `projectDir` when it is omitted, but rather `""`. Or at least that's what the types in my editor indicate.

rollup.config.js
MichaelDeBoey: Unfortunately they both are necessary, as transforms aren't used directly (they're used by their filePath).

This extra Rollup entry is getting the transforms index, which is exporting all transforms (since they're used in the `transformFunctionByName` object in `migrate/run/index.ts`), so is also exporting the escape-hatch, which is why the transform code is included in the build.

Was looking into doing it with a glob (something like `${sourceDir}/cli/migrate/run/transforms/**/jscode-transform/**` as `input`) so we don't need the escape-hatch in the code, but couldn't find how to do so.

MichaelDeBoey (approved):

PR #2598, opened 2022-04-01
feat(remix-react): Add array-syntax for `meta` export

kiliman:
This PR adds array-syntax to the `meta` export.

```ts
export const meta: MetaFunction = ({data}) => [
  { name: "description", content: data.description },
  { property: "og:image", content: "https://remix.run/logo.png" },
  { property: "og:type", content: "image/png" },
  { key: "http-equiv:refresh", httpEquiv: "refresh", content: "5;url=https://google.com" },
  { title: data.title },
  { name: "viewport", content: "width=device-width, initial-scale=1" },
];
```

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -56,7 +56,6 @@ import type {
   Fetcher,
   Submission,
 } from "./transition";
-import { keyBy } from "lodash";
 
 ////////////////////////////////////////////////////////////////////////////////
 // RemixEntry
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -688,6 +688,8 @@ function PrefetchPageLinksImpl({
   );
 }
 
+type MetaMap = Map<string, HtmlMetaDescriptor>;
+
 /**
  * Renders the `<title>` and `<meta>` tags for the current routes.
  *
@@ -697,7 +699,7 @@ export function Meta() {
   let { matches, routeData, routeModules } = useRemixEntryContext();
   let location = useLocation();
 
-  let meta: HtmlMetaDescriptor[] = [];
+  let meta: MetaMap = new Map();
   let parentsData: { [routeId: string]: AppData } = {};
 
   for (let match of matches) {
@@ -713,7 +715,7 @@ export function Meta() {
           ? routeModule.meta({ data, parentsData, params, location })
           : routeModule.meta;
       if (routeMeta) {
-        meta = processMeta(meta, routeMeta);
+        processMeta(meta, routeMeta);
       }
     }
 
@@ -722,13 +724,12 @@ export function Meta() {
 
   return (
     <>
-      {meta.map((metaDescriptor) => {
-        if (!metaDescriptor) {
-          return null;
+      {[...meta.entries()].map(([key, value]) => {
+        if (key === "title" && typeof value.content === "string") {
+          return <title key={key}>{value.content}</title>;
         }
-
-        if (metaDescriptor.hasOwnProperty("charset")) {
-          return <meta key="charSet" charSet={metaDescriptor.charset} />;
+        if (key === "charset" && typeof value.content === "string") {
+          return <meta key={key} charSet={value.content} />;
         }
 
         if (name === "title") {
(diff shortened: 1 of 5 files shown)


packages/remix-react/components.tsx
```diff
@@ -651,7 +651,7 @@ export function Meta() {
   let { matches, routeData, routeModules } = useRemixEntryContext();
   let location = useLocation();
 
-  let meta: HtmlMetaDescriptor = {};
+  let meta: HtmlMetaDescriptor[] = [];
```
donavon, line 654: let's change `meta` to a `Map`

```ts
type MetaMap = Map<string, HtmlMetaDescriptor>;
```

```ts
let meta: MetaMap = new Map();
```

packages/remix-react/components.tsx
```diff
@@ -666,52 +666,108 @@ export function Meta() {
...
       })}
     </>
   );
 }
 
+/*
+ * This function processes a meta descriptor and adds it to the meta array.
+ * This converts the original object syntax to new array syntax.
+ */
+export function processMeta(
```
donavon, line 705: change `HtmlMetaDescriptor` to this:

```ts
export type HtmlMetaDescriptor = {
  key?: string;
  name?: string;
  property?: string;
  title?: string;
  content?: string;
} & Record<string, string | string[] | null | undefined>;
```

and try this for `processMeta`. it sets `key` here

```tsx
export function processMeta(
  meta: MetaMap,
  routeMeta: HtmlMetaDescriptor | HtmlMetaDescriptor[]
) {
  let items: HtmlMetaDescriptor[] = Array.isArray(routeMeta)
    ? routeMeta
    : Object.entries(routeMeta)
        .map(([key, value]) => {
          if (!value) return [];

          let propertyName = key.startsWith("og:") ? "property" : "name";
          return key === "title"
            ? { key: "title", content: assertString(value) }
            : ["charset", "charSet"].includes(key)
            ? { key: "charset", content: assertString(value) }
            : Array.isArray(value)
            ? value.map((content) => ({
                key: `${key}.${content}`,
                [propertyName]: key,
                content,
              }))
            : { key, [propertyName]: key, content: value };
        })
        .flat();

  items.forEach((item) => {
    let [key, value] = computeKey(item);
    // only set if key doesn't exist (i.e. let the furthest route win)
    if (!meta.has(key)) {
      meta.set(key, value);
    }
  });
}
```

other needed helpers:

```tsx
function assertString(value: string | string[]): string {
  if (typeof value !== "string") {
    throw new Error("Expected string, got an array of strings");
  }
  return value;
}

function generateKey(item: HtmlMetaDescriptor) {
  console.warn("No key provided to meta", item);
  return JSON.stringify(item); // generate a reasonable key
}

function computeKey(item: HtmlMetaDescriptor): [string, HtmlMetaDescriptor] {
  let {
    name,
    property,
    title,
    key = name ?? property ?? title ?? generateKey(item),
    ...rest
  } = item;

  return [key, { name, property, title, ...rest }];
}
```

packages/remix-react/components.tsx
```diff
@@ -666,52 +666,108 @@ export function Meta() {
...
-      Object.assign(meta, routeMeta);
+      if (routeMeta) {
+        meta = processMeta(meta, routeMeta);
+      }
     }
 
     parentsData[routeId] = data;
   }
 
   return (
```
donavon, line 677: And here's the rendering part. The keys have been pre-computed in `processMeta`

```ts
  return (
    <>
      {[...meta.entries()].map(([key, value]) => {
        if (key === "title" && typeof value.content === "string") {
          return <title key={key}>{value.content}</title>;
        }
        if (key === "charset" && typeof value.content === "string") {
          return <meta key={key} charSet={value.content} />;
        }
        return <meta key={key} {...value} />;
      })}
    </>
  );
```

donavon (commented):
What do you think of my suggestions?

packages/remix-react/components.tsx
kiliman: I've made these changes, but a couple of comments:

1) I was trying not to change `HtmlMetaDescriptor` since it's a public API. Your version no longer supports the extended object syntax (ie., httpEquiv). My understanding is that's going to be reverted anyway.

2) The array syntax no longer supports the shorthand for `{ title: "my title"}` instead requiring `{ key: "title", content"my title" }`. Same with charset. Probably need Ryan to weigh-in on this change.

3) When you're merging meta map, you said only set if key doesn't exist. However, `processMap` is called from the root to the child route, so without subsequent sets, you can't override. I just removed the check.

With these changes, my local tests work. Still having issues with the integration test.

machour:
@kiliman Could you fix conflicts here please?

felquis:
Hello team, I want to mention a scenario I found, where open graph og:image accept multiple metatags with the same name, as described at: https://ogp.me/#array


> If a tag can have multiple values, just put multiple versions of the same <meta> tag on your page. The first tag (from top > to bottom) is given preference during conflicts.
> 
> ```html
> <meta property="og:image" content="https://example.com/rock.jpg" />
> <meta property="og:image" content="https://example.com/rock2.jpg" />
> ```
> Put structured properties after you declare their root tag. Whenever another root element is parsed, that structured property is considered to be done and another one is started.
> 
> For example:
> ```html
> <meta property="og:image" content="https://example.com/rock.jpg" />
> <meta property="og:image:width" content="300" />
> <meta property="og:image:height" content="300" />
> <meta property="og:image" content="https://example.com/rock2.jpg" />
> <meta property="og:image" content="https://example.com/rock3.jpg" />
> <meta property="og:image:height" content="1000" />
> ```
> means there are 3 images on this page, the first image is 300x300, the middle one has unspecified dimensions, and the last one is 1000px tall.

Just want to mention that this PR might allow this open graph usage. Thank you.

kiliman:
@machour I've rebased onto latest dev

@felquis I added support for duplicate property names. It dedupes by property+content. If you need to override dupes, you can provide a unique key for each.

https://github.com/remix-run/remix/blob/87594af3f1f6f0ef06c1a712b67f5ef1657b5137/integration/meta-test.ts#L146-L159

https://github.com/remix-run/remix/blob/87594af3f1f6f0ef06c1a712b67f5ef1657b5137/integration/meta-test.ts#L253-L260

chaance:
>  It dedupes by property+content. If you need to override dupes, you can provide a unique key for each.

@kiliman This seems like it could be error-prone across nested route boundaries. In the vast majority of cases you want your child route's meta to override the parent's. For this to work would users need to provide keys for every tag, or just tags that support duplicates for array values?

packages/remix-react/components.tsx
```diff
@@ -712,52 +714,103 @@ export function Meta() {
...
+    ? routeMeta.map((item) =>
+        item.title
+          ? { key: "title", content: item.title }
+          : { key: generateKey(item), ...item }
+      )
+    : Object.entries(routeMeta)
+        .map(([key, value]) => {
+          if (!value) return [];
+
+          let propertyName = key.startsWith("og:") ? "property" : "name";
```
chaance, line 766: This will need to be updated. See https://github.com/remix-run/remix/pull/4445

kiliman: Ok, I'll review this. Basically the key should be either `name` or `property` value. The only time you need an explicit `key` property is if you have duplicates, or there is no name/property like `<meta http-equiv="refresh" />`

chaance:
> Just want to mention that this PR might allow this open graph usage. Thank you.

@felquis Just so you know, you can already do this today with the existing API. You just need to pass an array to the tag's content.

```ts
export function meta() {
  return {
    "og:image": [
      "https://picsum.photos/300/300",
      "https://picsum.photos/200/200",
    ],
  };
}
```

felquis:
@chaance thank you for mentioning me.

Can you show me how the usage you showed me can be
applied to reproduce the following usage suggested by Open Graph?
```html
<meta property="og:image" content="https://example.com/rock.jpg" />
<meta property="og:image:width" content="300" />
<meta property="og:image:height" content="300" />
<meta property="og:image" content="https://example.com/rock2.jpg" />
<meta property="og:image" content="https://example.com/rock3.jpg" />
<meta property="og:image:height" content="1000" />
```

I don't want to make this discussion longer though, you can create another discussion/issue if you think it may be relevant to link in the future

Thank you so much for your help! ❤️

chaance:
For all interested, we have an RFC that will address a lot of the issues mentioned in this thread. https://github.com/remix-run/remix/discussions/4462

PR #2600, opened 2022-04-01
chore(test): Move `action` tests to `integration`

chaance:


kentcdodds (approved):
Super 👍 This is definitely an improvement. Thanks!

PR #2596, opened 2022-04-01
Docs: replace `remix` imports

pcattori:
Mostly syntactic/mechanical changes taking into account changes to the highlighted lines in codeblocks.

Semantic/prose changes within:
- `other-api/dev.md`
- `api/conventions.md`
- `api/remix.md` (TODO)

PR #2615, opened 2022-04-03
docs(jokes): fix LiveReload import

kleinfreund:
Fixes the LiveReload import using the package `remix` instead of `@remix-run/react`. The former isn’t installed in the recommended `create-remix@latest` project and causes the subsequent `npm run dev` step to fail.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -154,6 +154,7 @@
 - kgregory
 - kimdontdoit
 - klauspaiva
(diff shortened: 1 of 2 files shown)


kleinfreund:
Actually, this is completely superseded by https://github.com/remix-run/remix/pull/2596 which GitHub’s issue search did not yield on several searches.

PR #2629, opened 2022-04-04
feat(remix-dev/compiler): support native `.node` files

nicksrandall:
This PR enables support for users to include "native" (`.node`) modules into their server bundles (when using NodeJS). 

Libraries like `sharp` (awesome for for image processing) are written in a lower-level language than NodeJS and then are compiled as a "native" module to be used in a Node environment.

Discussion: #2560

- [ ] Docs
- [ ] Tests

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -359,6 +359,7 @@
 - nexxeln
 - ni554n
 - nicholaschiang
+- nicksrandall
 - nickytonline
 - niconiahi
 - nielsdb97
--- a/packages/remix-dev/compiler/loaders.ts
+++ b/packages/remix-dev/compiler/loaders.ts
@@ -28,6 +28,7 @@ export const loaders: { [ext: string]: esbuild.Loader } = {
   ".mov": "file",
   ".mp3": "file",
   ".mp4": "file",
+  ".node": "copy",
   ".ogg": "file",
   ".otf": "file",
   ".png": "file",


machour:
Sorry @nicksrandall, I did a whopsy while trying to change the base branch (code changes need to target `dev`).
I'll get this fixed unless you do.

packages/remix-dev/compiler/plugins/serverNativeNodeModules.ts
```diff
@@ -0,0 +1,32 @@
+import type { Plugin } from "esbuild";
+
+// See https://github.com/evanw/esbuild/issues/1051#issuecomment-806325487
+export const serverNativeNodeModulesPlugin: Plugin = {
```
MichaelDeBoey, line 4: Let's make this a function, so it's consistent with our other plugins

```suggestion
export const serverNativeNodeModulesPlugin = (): Plugin => ({
```

packages/remix-dev/compiler/plugins/serverNativeNodeModules.ts
```diff
@@ -0,0 +1,32 @@
...
+  name: 'native-node-modules',
+  setup(build) {
```
MichaelDeBoey, line 6:
```suggestion
  setup: (build) => {
```

MichaelDeBoey (changes requested):

packages/remix-dev/compiler/plugins/serverNativeNodeModules.ts
MichaelDeBoey, line 4: Let's make this a function, so it's consistent with our other plugins

```suggestion
export const serverNativeNodeModulesPlugin = (): Plugin => ({
```

packages/remix-dev/compiler/plugins/serverNativeNodeModules.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    // If a ".node" file is imported within a module in the "node-file" namespace, put
+    // it in the "file" namespace where esbuild's default loading behavior will handle
+    // it. It is already an absolute path since we resolved it to one above.
+    build.onResolve({ filter: /\.node$/, namespace: 'node-file' }, args => ({
+      path: args.path,
+      namespace: 'file',
+    }))
+  },
+}
```
MichaelDeBoey, line 32:
```suggestion
});
```

packages/remix-dev/compiler/plugins/serverNativeNodeModules.ts
MichaelDeBoey, line 6:
```suggestion
  setup: (build) => {
```

packages/remix-dev/compiler.ts
```diff
@@ -423,6 +424,7 @@ function createServerBuild(
     serverAssetsManifestPlugin(assetsManifestPromiseRef),
     serverBareModulesPlugin(config, options.onWarning),
     yarnPnpPlugin(),
+    serverNativeNodeModulesPlugin,
```
MichaelDeBoey, line 427:
```suggestion
    serverNativeNodeModulesPlugin(),
```

packages/remix-dev/compiler/plugins/serverNativeNodeModules.ts
```diff
@@ -0,0 +1,32 @@
+import type { Plugin } from "esbuild";
+
+// See https://github.com/evanw/esbuild/issues/1051#issuecomment-806325487
+export const serverNativeNodeModulesPlugin: Plugin = {
+  name: 'native-node-modules',
+  setup(build) {
+    // If a ".node" file is imported within a module in the "file" namespace, resolve
+    // it to an absolute path and put it into the "node-file" virtual namespace.
+    build.onResolve({ filter: /\.node$/, namespace: 'file' }, args => ({
+      path: require.resolve(args.path, { paths: [args.resolveDir] }),
```
MichaelDeBoey, line 10: @jacob-ebey's comment (https://github.com/remix-run/remix/pull/3263#discussion_r910178465)
> Add `sideEffects: false` here to make sure this can be optimized out of client bundles.

packages/remix-dev/compiler/plugins/serverNativeNodeModules.ts
```diff
@@ -0,0 +1,32 @@
...
+      namespace: 'node-file',
+    }))
+
+    // Files in the "node-file" virtual namespace call "require()" on the
+    // path from esbuild of the ".node" file in the output directory.
+    build.onLoad({ filter: /.*/, namespace: 'node-file' }, args => ({
+      contents: `
+        import path from ${JSON.stringify(args.path)}
+        try { module.exports = require(path) }
```
MichaelDeBoey, line 19: @jacob-ebey's comment (https://github.com/remix-run/remix/pull/3263#discussion_r910177928)
> This will break node ESM. Is it possible to load via `import` statements?

@benwis' reaction (https://github.com/remix-run/remix/pull/3263#discussion_r910203192)
> The trouble is getting the absolute path and then loading it. I can't use await import() because I can't have top level await. Unless there's a way to get the absolute path and the compiled path without importing it, I don't see how I resolve this

MichaelDeBoey (changes requested):
@nicksrandall I closed #3263 by @benwis in favor of this one, since you started first on this idea.

Could you please look into that PR & see if we also want to only enable this plugin on node based platforms?
Can you also check if we need [the extra compiler changes](https://github.com/remix-run/remix/pull/3263/files#diff-ef8b0a35f03b499672160742c2d6b9688717e8d6b923dde0e10d2631ab0c09d6R515) that were implemented.
Copying over the docs would be a good thing as well I think (please [add @benwis as a co-author](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) & in the `contributors` list)

benwis:
@MichaelDeBoey I don't particularly care which PR gets merged, but I have no ability to edit this here. I think you need the node system check, the compiler changes, the path detection code, and everything else I had to do in my PR. We're running this change in Prod right now to serve .node files, and it seems to work well.

nicksrandall:
@benwis I'll give you access to my branch. I also don't care which PR gets merged so if it's easier to close then and re-open yours, we can do that.

benwis:
@nicksrandall Thanks for inviting me! I'm happy to do it here, I'll dig in and address the comments today. I'd love to see this and the WASM changes merged!

MichaelDeBoey:
Let's keep this branch open then 👍

benwis:
@MichaelDeBoey @nicksrandall  I have updated this repo for the changes in Remix 1.7.1 and reapplied the changes made, plus some of mine. I'm not sure how to get this pull request to update based on the status of the dev branch of the parent repo though. It seems to be stuck.

nicksrandall:
@MichaelDeBoey any updates on this. This is a blocker for me.

chaance:
@nicksrandall I'd like to get this merged but there are comments left by @jacob-ebey that still haven't been addressed, and the server compiler code was moved into `compilerServer.ts` so some of the code needs to be moved over. If you can get this cleaned up and address outstanding issues, I'll merge.

nicksrandall:
@chaance The solutions here has gotten more simple since esbuild added support for the `copy` loader. 

See: https://github.com/evanw/esbuild/issues/1051#issuecomment-1156783762

This should just work when app imports a `.node` file directly.

One potential gotcha that I feel like I should point out is that packages that ship `.node` files have to compile a node file per OS/Architecture (like `linux/amd64`). Sometimes, the library author will then use dynamic require statements to require the appropriate node file. Esbuild can not know the path of a dynamic require statement so this will not work in those cases. 

In such a case, the user will just need to mark the package as "external" (or in Remix case, exclude it from being bundled) and then it should just work.

nicksrandall:
@benwis I know you also worked on this... You comfortable just using the `copy` loader for `.node` files?

chaance:
> In such a case, the user will just need to mark the package as "external" (or in Remix case, exclude it from being bundled) and then it should just work.

I that's probably fine for now. The same is true for any other dynamic import.

benwis:
I think that's probably fine as well, although I haven't found time to test it :)

MichaelDeBoey (approved):

chaance (approved):

PR #2630, opened 2022-04-04
Revert "docs: note about remix setup"

MichaelDeBoey:
Docs were updated in #2627, so this shouldn't be a problem anymore.

This reverts commit 25049aeb29fa9f577065e47ac2caefbb4b427af7.


kentcdodds (approved):
Thanks!

PR #2631, opened 2022-04-04
fix(templates/vercel): fix dependencies

MichaelDeBoey:
`@vercel/node` is in `@remix-run/vercel`'s `peerDependencies`

https://github.com/remix-run/remix/blob/22a618f93556d7cf57f3630273d8ae6c689656fc/packages/remix-vercel/package.json#L17-L19


kentcdodds (approved):
👍 Thanks!

PR #2636, opened 2022-04-04
feat(remix-dev): allow `data` to be returned from the MDX loader

donavon:
Some remark plugins (such as [remark-reading-time](https://github.com/mattjennings/remark-reading-time) and [@stefanprobst/remark-extract-toc](https://github.com/stefanprobst/remark-extract-toc)) mutate `data`. The problem is that Remix doesn't expose `data`. The PR will solve this.

When I have my `remix.config.js` file setup like this:

```js
mdx: async function () {
  const remarkPlugins = await Promise.all([
    import('remark-reading-time').then((mod) => mod.default),
  ]);

  return {
    remarkPlugins,
  };
},
```

The inside of `mdx.ts` we get the reading time (or anything else)  from `compiled.data` and expose it as an export.

Here is an example. See the new `data` export in the last line.

```ts
export const filename = "ink.mdx";
export const headers = typeof attributes !== "undefined" && attributes.headers;
export const meta = typeof attributes !== "undefined" && attributes.meta;
export const links = undefined;
export const data = {"readingTime":{"text":"1 min read","minutes":0.02,"time":1200,"words":4}};
```

This allows me to `import` the MDX file and do what I please with it.

```ts
import { data } from './ink.mdx';

console.log(ink.data.readingTime.text);
// outputs: 1 min read
```

On my blog, I use it like this:

image

Note: as this is an API change, I'm tagging @ryanflorence and @mjackson

### DIFF
--- a/packages/remix-dev/compiler/plugins/mdx.ts
+++ b/packages/remix-dev/compiler/plugins/mdx.ts
@@ -82,7 +82,7 @@ export const filename = ${JSON.stringify(path.basename(args.path))};
 export const headers = typeof attributes !== "undefined" && attributes.headers;
 export const meta = typeof attributes !== "undefined" && attributes.meta;
 export const links = undefined;
-export const data = ${JSON.stringify(compiled.data)})};
+export const data = ${JSON.stringify(compiled.data)};
 `;
 
           let errors: esbuild.PartialMessage[] = [];
--- a/packages/remix-dev/compiler/plugins/mdx.ts
+++ b/packages/remix-dev/compiler/plugins/mdx.ts
@@ -67,13 +67,6 @@ export function mdxPlugin(config: RemixConfig): esbuild.Plugin {
               break;
           }
 
-          let remixExports = `
-export const filename = ${JSON.stringify(path.basename(args.path))};
-export const headers = typeof attributes !== "undefined" && attributes.headers;
-export const meta = typeof attributes !== "undefined" && attributes.meta;
-export const links = undefined;
-          `;
-
           let compiled = await xdm.compile(fileContents, {
             jsx: true,
             jsxRuntime: "classic",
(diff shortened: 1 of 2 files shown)


donavon:
I was thinking about exposing this as a `loader` function that returns the same data. That way any components inside of the MDX file could get the data with `useLoaderData`. Thoughts on this alternate approach?

donavon:
I'm also excited for this as it would allow me to expose "out of band" Table of Contents in a hamburger menu.

Here is `data.toc` when using [@stefanprobst/remark-extract-toc](https://github.com/stefanprobst/remark-extract-toc)

```
[
  { depth: 2, value: 'Background' },
  { depth: 2, value: 'The problem' },
  { depth: 2, value: 'The solution' },
  { depth: 2, value: 'Specifying your preferred locales' },
  { depth: 2, value: 'Locale Provider' },
  { depth: 2, value: 'Wrapping the App on the server' },
  { depth: 2, value: 'Rendering dates in React' },
  { depth: 2, value: 'Client side rendering' },
  { depth: 2, value: "That's it!" }
]
```

donavon:
👊 bump

donavon:
If anyone stumbles upon this looking for a solution and this PR has been rejected or still remains open, check out the `mdx-data` patch that I'm using in production today.

https://github.com/donavon/remix-patches

kiliman:
@kentcdodds has a new project to eventually replace the existing MDX handler. I've got a PR that add support for loader/meta in MDX (data comes from frontmatter)

https://github.com/kentcdodds/remix-mdx/pull/2

kentcdodds:
We've decided to go a different direction with MDX. We're probably not going to support this. We'll do something else in the future that I think people will prefer.

kentcdodds:
Thanks for taking the time though @donavon!

coxley:
Is this going to be revisited soonish?

It doesn't seem to be possible according to the current docs and it's been awhile. I'm in a position where I've built things around the remix MDX integration, but ripping it out just for TOC extraction feels heavy handed. 😅

PR #2643, opened 2022-04-04
docs(api/remix): replace `remix` imports

pcattori:
Also, switch from `@remix-run/node` to `@remix-run/{runtime}` in API, Guides  and general pages to be server-runtime-agnostic.

### DIFF
--- a/docs/api/conventions.md
+++ b/docs/api/conventions.md
@@ -263,7 +263,7 @@ import { useParams } from "@remix-run/react";
 import type {
   LoaderFunction,
   ActionFunction,
-} from "@remix-run/node";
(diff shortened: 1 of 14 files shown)


MichaelDeBoey (approved):

kentcdodds (approved):
I wonder if there's anything we can do in the compiler to catch people who copy/paste without fixing the `{runtime}` 🤔 Like if the compiler sees an import for `@remix-run/{runtime}` we can handle that and tell them they need to update it to `@remix-run/node` or whatever runtime they're using.

But that would need to happen in a separate PR.

Issue #2620, opened 2022-04-03
Changing devServerPort in remix-config.js has no effect on dev server websocket port


eldarshamukhamedov:
### What version of Remix are you using?

1.3.4

### Steps to Reproduce

1. Create fresh Remix app with the `basic` template: `npx create-remix --template basic`
2. Run `npm install`
3. Add `devServerPort: 10890` to the `remix.config.js` file
4. Run Remix dev server with `npm run dev`

### Expected Behavior

No console errors:
![image](https://user-images.githubusercontent.com/1694252/161453872-6aed52da-6903-40a4-8053-0d8c38fbba07.png)

Modifying the "You!" string in the footer to "You!!!" live reloads the component.

### Actual Behavior

Console error:
![image](https://user-images.githubusercontent.com/1694252/161453926-5ed966ca-f868-4962-bdcc-af9ecab62c16.png)

FF attempts to establish socket connection to `8002`, the default port. `devServerPort: 10890` value from `remix.config.js` is ignored.

Modifying the "You!" string in the footer to "You!!!" does not live reload the component. Component is only updated with a page refresh (Ctrl + R).

alishi973:
For short-term solution, you can pass explicit ``port`` as props to ``<LiveReload />`` in ``/app/root.tsx``.
It seems ``devServerPort`` from ``remix.config.js`` will not inject into the browser bundle.

doppelmutzi:
In contrast to the documentation, the dev server port is not defaulting to 3000 with my indie stack project. The app server port in different every time I run `npm run dev`. 

I have the same problem with 1.3.4, that I couldn't manage to set the dev server port in the `remix.config.js`. What works for me as a workaround is to set the port explicitly in the root `package.json` like this

```json
// package.json
{
  // ...
  "scripts": {
    "dev": "PORT=3001 run-p dev:*"
    // ...
  }
}
```

alishi973:
@doppelmutzi, do not confuse the ``express`` static file serve ``port`` and ``LiveReload`` ``port``, this issue is related to the web-socket port on the browser bundle files, and your problem is related to the express port.

About your problem:
https://github.com/remix-run/remix/blob/981bb7621f7936b43612335c6c8945c634105468/packages/remix-dev/cli/commands.ts#L249-L251
As you can see, ``getPort`` will take ``process.env.PORT`` as the port number and if the specified port is not occupied by another process, return this number, or if the ``PORT`` variable is unset, it will be set to ``3000`` by default.

[...]

PR #2645, opened 2022-04-04
fix(remix-dev): pass `devServerPort` to React

alishi973:
Closes: #2620 

This issue demonstrates that ``devServerPort`` on ``remix.config.js`` will not inject into react-dev-server.

Then I reviewed #1352 by @kentcdodds and found something:

In this PR, we tried to pass remix config with ``esbuild`` module to react-dev-server:

https://github.com/remix-run/remix/blob/c3c3627374ef8b3e7b4eb7f30d614eeae48b37ef/packages/remix-dev/compiler.ts#L365-L370

According to ``esbuild`` test files, it's correct, and everything should be working fine and passes the parameter.
here is ``esbuild`` test files for the ``define`` property:
https://github.com/evanw/esbuild/blob/c02c99fc74373aed6d58e7a349b463cd26e1d86f/scripts/js-api-tests.js#L3847-L3861

Also, check this test:
https://github.com/remix-run/remix/blob/c3c3627374ef8b3e7b4eb7f30d614eeae48b37ef/packages/remix-react/__tests__/components-test.tsx#L48-L54
 
On line 49 we assign ``devServerPort`` directly to ``process.env``, so react-dev-server will pick this variable, but in a normal situation, we did not use ``process.env`` for passing variables and used ``define`` property on ``esbuild`` .

and finally, any change on this PR or suggestion about passing data from remix config to react dev server is appreciated✌️.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -11,6 +11,7 @@
 - Alarid
 - alex-ketch
 - alexuxui
+- alishi973
 - andrelandgraf
 - AndrewIngram
 - anishpras
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -455,10 +455,7 @@ function createServerBuild(
       assetNames: "_assets/[name]-[hash]",
       publicPath: config.publicPath,
       define: {
-        "process.env.NODE_ENV": JSON.stringify(options.mode),
-        "process.env.REMIX_DEV_SERVER_WS_PORT": JSON.stringify(
-          config.devServerPort
-        ),
+        "process.env.NODE_ENV": JSON.stringify(options.mode)
       },
       plugins,
     })
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -352,6 +352,8 @@ export async function readConfig(
   );
 
   let devServerPort = await getPort({ port: appConfig.devServerPort || 8002 });
+  // TODO: find better way for passing remix config to react, temporary assign to process directly.
+  process.env.REMIX_DEV_SERVER_WS_PORT = JSON.stringify(devServerPort);
   let devServerBroadcastDelay = appConfig.devServerBroadcastDelay || 0;
 
   let defaultPublicPath = "/build/";


kentcdodds (commented):
Thanks for this. How does the port value get to the client-side for `<LiveReload />` to work?

alishi973:
``<LiveReload />`` component gets ``port`` as props and uses it,
But in case of ``port`` has not been declared, it will use the default parameter:
https://github.com/remix-run/remix/blob/598a244e35c0cccaf1d5d5a858dde3551a5dd71d/packages/remix-react/components.tsx#L1388-L1392
As I mentioned, we expect ``process.env.REMIX_DEV_SERVER_WS_PORT`` to have the correct value but it has an ``undefined`` value.
I tried to set this variable in the ``readConfig`` function, then we can use this variable in the ``remix-react`` without the need to pass as a parameter on ``esbuild``.

mcansh:
my only concern is using this with Cloudflare and Deno where they don't have `process.env`, the esbuild define is _supposed_ to replace that with `config.devServerPort`, but it doesnt seem that it is happening

alishi973:
Thanks @mcansh for your tip, I tracked everything and found why ``esbuild`` won't pass the env parameter:

We have 2 types of build:
1. Server build
2. Browser Build

For browser build everything works fine, even you can see the ``port`` that comes from the ``esbuild`` ``define`` property, for proof, you can change the ``outdir`` in ``remix-dev/compiler.ts`` on the ``createBrowserBuild`` function and see the bundle files that have correct ``port`` value.

But for Server Build, we have only one file and that bundled file has not any ``<LiveReload />`` component, this file only has ``require("@remix-run/react")``.
As far as I know, ``esbuild`` only builds (or transforms) files and can't run the file itself so for using the ``define`` property, it should exist in the source file.
But in the server builds file's content we have not used that variable, so it will be ignored.
(``esbuild`` only replace ``process.env.XXX`` with ``XXX`` value on ``define`` property and will not execute files with that parameter).

So I think we have 2 options:
1. Use global variables like ``process.env`` (and make it compatible with deno)
2. Compile ``root.tsx`` (because it uses ``<LiveReload />``) with ``esbuild`` and put it on the server build file.

MichaelDeBoey:
Hi @alishi973!

If so, please rebase onto latest `dev` & resolve conflicts/remarks

alishi973:
Is it possible to reopen this? I’ve just been busy in recent days.
But yeah I’m interested and will resolve the issues ASAP.

brophdawg11:
@pcattori Is this still relevant in the new v2_dev server?

pcattori:
Obsoleted by new dev flow. With `v2_dev`, `remix dev --port` correctly sets the websocket port

PR #2650, opened 2022-04-05
lint: let to const where needed

load1n9:
Closes: #

- [ ] Docs
- [ ] Tests


MichaelDeBoey:
Going to close this one as ESLint is configured to use `let` & `const` where the teams want to use it & automatically fixed too.

Also: code-changes need to happen to `dev` instead of `main`

PR #2651, opened 2022-04-05
chore(test): Migrate `link` tests to integration

chaance:
Plz ignore the goofy commits and squash/merge 🙈


integration/helpers/create-fixture.tsx
```diff
@@ -166,10 +167,23 @@ export async function createAppFixture(fixture: Fixture) {
        * @param waitForHydration Will wait for the network to be idle, so
        * everything should be loaded and ready to go
        */
-      goto: async (href: string, waitForHydration?: true) => {
-        return page.goto(`${serverUrl}${href}`, {
-          waitUntil: waitForHydration ? "networkidle0" : undefined,
-        });
+      goto: (async (href: string, opts?: any) => {
```
chaance, line 170: I didn't actually change any of this. Did I miss something in a merge somewhere?

integration/helpers/create-fixture.tsx
```diff
@@ -23,6 +23,7 @@ export type Fixture = Awaited<ReturnType<typeof createFixture>>;
 export type AppFixture = Awaited<ReturnType<typeof createAppFixture>>;
 
 export const js = String.raw;
+export const css = String.raw;
```
chaance, line 26: Technically not needed but it helps w/ syntax highlighting plugins that support CSS-in-JS

kentcdodds:
We'll figure out the failing integration test issue once this is in my branch.

Thanks for doing all this!

PR #2652, opened 2022-04-05
fix(remix-dev): resource route module making it to the browser

jacob-ebey:
Closes: #

- [ ] Docs
- [x] Tests

### DIFF
--- a/integration/fetcher-test.ts
+++ b/integration/fetcher-test.ts
@@ -11,12 +11,41 @@ describe("useFetcher", () => {
   beforeAll(async () => {
     fixture = await createFixture({
       files: {
+        "app/routes/resource-route-action-only.ts": js`
+          import { json } from "@remix-run/node";
+          export function action() {
+            return json("${CHEESESTEAK}");
+          }
+        `,
+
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
This looks good to me 👍

brophdawg11 (approved):
👍 Nice work - this fixes my test case

ryanflorence:
Correct me if I'm wrong, but this looks like it solves the problem by adding a client side bundle for a resource route.

The root cause of this bug was that resource routes don't produce client bundles, but `fetcher.load` triggers code inside of Remix to load the bundle, it's not found, and we do a `window.reload()` becuase this is usually because the user was caught between deploys and the asset got removed.

Making an empty client bundle for the sake of being able to load nothing is the wrong approach. We should just not try to load bundles for resource routes.

We should be able to identify if it's coming from a `fetcher.load` or `fetcher.submit` here and just not try to load the bundle:

https://github.com/remix-run/remix/blob/46b96f53e8477dd9404d55e75b320775c3a8a7a4/packages/remix-react/routes.tsx#L153

PR #2653, opened 2022-04-05
fix(remix-dev/cli/migrate): bug fixes & improved DX

pcattori:
Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/remix-dev/cli/migrate/jscodeshift.ts
+++ b/packages/remix-dev/cli/migrate/jscodeshift.ts
@@ -11,13 +11,13 @@ const toFlags = (options: Options = {}) =>
         `--${key}${typeof value === "boolean" ? "" : `=${value}`}`
     );
 
-type Args<TransformOptions extends Options = {}> = {
+type Args<TransformOptions> = {
   transformPath: string;
   files: string[];
   flags: { dry?: boolean; print?: boolean; runInBand?: boolean };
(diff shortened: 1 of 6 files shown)


packages/remix-dev/cli/migrate/jscodeshift.ts
```diff
@@ -11,13 +11,13 @@ const toFlags = (options: Options = {}) =>
...
 
-type Args<TransformOptions extends Options = {}> = {
+type Args<TransformOptions> = {
   transformPath: string;
   files: string[];
   flags: { dry?: boolean; print?: boolean; runInBand?: boolean };
   transformOptions?: TransformOptions;
 };
-export const run = <TransformOptions extends Options = {}>({
+export const run = <TransformOptions>({
```
MichaelDeBoey, line 20: @pcattori This means we'll always need to provide the `TransformOptions` type.

Since the value can be optional, I think it would be best to keep the type optional too tbh.
What do you think?

kentcdodds (approved):
LGTM 👍

packages/remix-dev/cli/migrate/jscodeshift.ts
pcattori: We can always make it optional once we add a transform that doesn't need options. For now, I prefer to keep it like this to keep things simple.

PR #2659, opened 2022-04-05
chore: add @typescript-eslint/consistent-type-imports rule

brophdawg11:
Adds the typescript [consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md) rule to the eslint config to enforce consistent separation of `import type` and `import`.

- [ ] Docs
- [ ] Tests


kentcdodds (approved):
Yes!

PR #2661, opened 2022-04-05
docs: replace `[NBSP]` with normal space

MichaelDeBoey:
Schermafbeelding 2022-04-06 om 01 21 17


kentcdodds (approved):
Thanks! Could this explain the formatting issues we had on the other PR I just merged about using ASCII?

MichaelDeBoey:
@kentcdodds I'm not sure about that tbh, I just saw these in the files and thought it would be better to replace them with normal spaces to be consistent.

PR #2662, opened 2022-04-06
fix(remix-react): pass `port` to inline script of `LiveReload`

kentcdodds:
Without this you get an error in the console indicating `port` is `undefined`.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1404,7 +1404,7 @@ export const LiveReload =
         port?: number;
         nonce?: string;
       }) {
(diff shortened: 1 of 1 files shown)


kentcdodds:
This may be wrong...

MichaelDeBoey (changes requested):

mcansh:
yea this will just make it work again on 8002, but is something we'd want in todays' release..

mcansh:
investigating a few things, seems `process.env.REMIX_DEV_SERVER_WS_PORT` isn't being replaced in the server build

chaance (approved):

PR #2461, opened 2022-03-23
ci: normalize Node.js version across actions

ruisaraiva19:
This normalizes the usage of `.nvmrc` file to setup the correct Node.js version on GitHub actions.

PR #2667, opened 2022-04-06
chore: improve workflows a bit

kentcdodds:
Saw an interesting idea and wanted to see whether it could improve the performance of our workflows any.

https://linear.app/remix-run/issue/REM-962/improve-workflows

Primary improvements:
- Cancel workflow steps when a new workflow is kicked off - This should reduce the amount of CI usage we have
- Leverage caching of npm/yarn installs - Should speed up install steps
- Give workflows and steps emoji - Makes it easier to identify the workflow/step you're looking for at a glance
- Speed up `yarn test:integration` by only running the rollup build (instead of waiting a couple seconds for TypeScript)


MichaelDeBoey:
According to what @ruisaraiva19 did in #2461 & [`actions/setup-node` docs itself](https://github.com/actions/setup-node#caching-packages-dependencies), we don't need extra caching steps anymore when using the `cache` arg 🤔

kentcdodds:
Yup, that's what I thought as well. Just wanted to verify my thinking in a test :)

kentcdodds:
It probably *does* help whenever you *do* need to install new dependencies though. But that happens rarely enough to not need the extra stuff.

mcansh (approved):

PR #2670, opened 2022-04-06
feat(remix-dev/cli): add DX improvements for `migrate` command & `replace-remix-imports` migration

pcattori:
Migration should guide the user through 3 steps:

1. Replacing the `remix` package with `@remix-run/*` packages.
2. Removing `remix setup` from the `postinstall` script
3. Replacing `import {...} from "remix"` with `import {...} from "@remix-run/{some package}"`

## Try it!

```sh
git fetch origin # or whatever you call remix-run/remix
git checkout pedro/fix-migrate
yarn build
node ./build/node_modules/@remix-run/dev/cli.js migrate
```

## Example
Screen Shot 2022-04-10 at 10 09 38 AM


packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
```diff
@@ -1,30 +1,74 @@
...
+  // match `remix setup <runtime>` in `postinstall` script
+  let remixSetupMatch = scripts?.postinstall?.match(/remix setup (\w+)/);
+  if (remixSetupMatch && remixSetupMatch.length >= 2) {
+    let postinstallRuntime = remixSetupMatch[1];
+    if (isRuntime(postinstallRuntime)) {
+      return postinstallRuntime;
+    }
+  }
+
+  // otherwise, ask user for runtime
```
pcattori, line 22: One _tiny_ annoyance is that if users remove `remix setup` from the `postinstall` script, we can no longer auto-detect the runtime reliably, as the runtime is often not a direct dependency of the app.

So for most users, as they re-run `migrate`, we will keep auto-detecting their runtime, until they do what they are supposed to and remove the `remix setup` from `postinstall` script. At that point, the user will have to specify their runtime.

If anyone has a nice solution for this, I'm all ears. For now, I think this is fine, but could _maybe_ be better.

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
```diff
@@ -1,30 +1,74 @@
...
+  let { runtime } = await inquirer.prompt<{ runtime?: Runtime }>([
+    {
+      name: "runtime",
+      message: "Which server runtime is this project using?",
+      type: "list",
+      pageSize: runtimes.length + 1,
+      choices: [...runtimes, { name: "Nevermind...", value: undefined }],
+    },
+  ]);
+  if (runtime === undefined) process.exit(0);
```
MichaelDeBoey, line 32: This will make the code less error-prone when needing to change it

```suggestion
  if (runtime === undefined) {
    process.exit(0);
  }
```

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
```diff
@@ -1,30 +1,74 @@
...
+    );
+    console.log(
+      hint(
+        "You should only need one Remix server adapter. Uninstall unused server adapter packages and try again."
+      )
+    );
+    process.exit(1);
+  }
+
+  if (matched.length === 1) return matched[0];
```
MichaelDeBoey, line 64:
```suggestion
  if (matched.length === 1) {
    return matched[0];
  }
```

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -4,41 +4,116 @@ import { join } from "path";
...
 import * as jscodeshift from "../../jscodeshift";
-import { cleanupPackageJson } from "./cleanupPackageJson";
-import { getTransformOptions } from "./getTransformOptions";
+import {
+  findRemixDependencies,
+  getTransformOptions,
+} from "./getTransformOptions";
 import type { Options } from "./transform";
+import { isAdapter } from "./transform";
+import { isRuntime } from "./transform";
```
MichaelDeBoey, line 13:
```suggestion
import { isAdapter, isRuntime } from "./transform";
```

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -4,41 +4,116 @@ import { join } from "path";
...
+}: {
+  packageJson: PackageJson;
+  runtime: string;
+  adapter?: string;
+}): Generator<string> {
+  let remixDeps = findRemixDependencies(packageJson.dependencies);
+
+  // runtime not in deps
+  if (!remixDeps.includes(runtime)) {
+    yield `Add \`@remix-run/${runtime}\` to your dependencies`;
```
pcattori, line 33: I think I'm leaning more towards "Install"/"Uninstall" (instead of "Add"/"Remove") to emphasize that we don't want to just delete a string from package.json, but we want to user to use their package manager to manipulate the dependencies across package.json, lock files, etc...

So I'm pretty sure I want to change this to "Install"/"Uninstall"

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
MichaelDeBoey: I think it's a tradeoff we can't really have otherwise, so don't mind it will be semi-"broken" when people re-run the migration.

Migrations aren't really meant to be run multiple times anyways 🤷‍♂️

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
MichaelDeBoey: Using install/uninstall instead of add/remove sounds indeed like a good idea to me 👍

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -4,41 +4,116 @@ import { join } from "path";
...
+    .filter((dep) => dep !== adapter);
+  if (otherAdapters.length > 0) {
+    yield `Remove all unused adapters from your dependencies: (${otherRuntimes
+      .map((r) => "@remix-run/" + r)
+      .join(",")})`;
+  }
+
+  // remix in deps
+  if (Object.keys(packageJson.dependencies || {}).includes("remix")) {
+    yield "Remove `remix` from your dependencies";
```
MichaelDeBoey, line 63: I liked that fact that it was done automatically tbh.

Since we're not doing that anymore and are actually only changing code now, do we still want to keep the `migrate` name? 🤔
The main reason to rename `codemod` to `migration` was that we were doing more than only changing actual code.

MichaelDeBoey (changes requested):

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
MichaelDeBoey: I think it's a tradeoff we can't really have otherwise, so don't mind it will be semi-"broken" when people re-run the migration.

Migrations aren't really meant to be run multiple times anyways 🤷‍♂️

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
MichaelDeBoey: Using install/uninstall instead of add/remove sounds indeed like a good idea to me 👍

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
pcattori: The issue was that removing `remix` from `package.json` did not remove it from the user's lockfiles. And we can't know what package manager each user is using, so its better to ask them to do it.

As for the naming: I still think `migrate`/`migration` is good as its like a wizard/assistant guiding users through a full migration.

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -4,41 +4,120 @@ import { join } from "path";
...
+    yield `Install \`@remix-run/${adapter}\` as a dependency`;
+  }
+
+  // other adapters in deps
+  let otherAdapters = remixDeps
+    .filter(isAdapter)
+    .filter((dep) => dep !== adapter);
+  if (otherAdapters.length > 0) {
+    yield `Uninstall all unused adapters: (${otherRuntimes
+      .map((r) => "@remix-run/" + r)
```
MichaelDeBoey, line 57:
```suggestion
      .map((otherRuntime) => `@remix-run/${otherRuntime}`)
```

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -4,41 +4,120 @@ import { join } from "path";
...
+
+  console.log("\n💿 Checking if manual migration steps are necessary...");
+  // ask the user to do some post-migration tasks
+  let tasks = [...getTasks({ packageJson, runtime, adapter })];
+  if (tasks.length > 0) {
+    console.warn("⚠️  Manual migration steps are necessary");
+    console.log(
+      "\nYou're almost there! To finish the migration, please perform the following steps:"
+    );
+    tasks.forEach((task) => console.log("👉 " + task));
```
MichaelDeBoey, line 118:
```suggestion
    tasks.forEach((task) => console.log(`👉 ${task}`));
```

MichaelDeBoey (changes requested):

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
```diff
@@ -1,30 +1,74 @@
...
-const getRuntime = ({ scripts }: PackageJson): Runtime =>
-  (scripts?.postinstall?.replace("remix setup ", "") as Runtime) || "node";
-
-export const getTransformOptions = (packageJson: PackageJson): Options => ({
-  adapter: getAdapter(packageJson),
-  runtime: getRuntime(packageJson),
+const resolveRuntime = async ({ scripts }: PackageJson): Promise<Runtime> => {
+  // match `remix setup <runtime>` in `postinstall` script
+  let remixSetupMatch = scripts?.postinstall?.match(/remix setup (\w+)/);
+  if (remixSetupMatch && remixSetupMatch.length >= 2) {
```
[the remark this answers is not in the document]
pcattori: Can't use `?.` operator as that results in trying to compare `null` or `undefined` to `2` via `>=` operator, and Typescript doesn't like that.

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
pcattori: I do like the workflow of running the migration, doing some of what it says, re-running, etc... until everything is green. So in this workflow, `migrate` would ask for the runtime at least once.

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
MichaelDeBoey: We do have some places in our CLI that automatically is installing some things though, which is using `npm`.

#2562 is trying to auto-detect it and run the corresponding command accordingly.
We could maybe use that in here as well in a later phase.

pcattori: The `create` doesn't assume `npm`, but rather asks the user if `npm install` should be run. The user can say "No" and then use whatever package manager they want.

But yes, if we have a stable, reliable way to detect package manager in the future, we can revisit this.

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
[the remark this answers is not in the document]
MichaelDeBoey: TypeScript is really weird sometimes 😅

packages/remix-dev/cli/run.ts
```diff
@@ -344,10 +346,13 @@ export async function run(argv: string[] = process.argv.slice(2)) {
...
-        migrationId: flags.migration,
-        projectDir: input[1],
-      });
+      let projectDir = commands.migrate.resolveProjectDir(input[1]);
+      if (!flags.dry) {
+        checkGitStatus(projectDir, { force: flags.force });
+      }
+      let migrationId = await commands.migrate.resolveMigrationId(
+        flags.migration
+      );
```
MichaelDeBoey, line 355: Maybe move this back into a `resolveInput` function and do the check right there?

That way the CLI code stays clean:
1. resolve the input
2. run the command

pcattori: We want to show the git status check as soon as possible (i.e. before potentially asking the user for the migration they want to apply). That optimizes for the UX of the CLI.

MichaelDeBoey: If we have

```ts
export const resolveInput = () => {
  let projectDir = resolveProjectDir(input[1]);

  if (!flags.dry) {
    checkGitStatus(projectDir, { force: flags.force });
  }

  let migrationId = await resolveMigrationId(flags.migration);

  return { migrationId, projectId };
}
```

we would still have it as early as we have now, but the CLI file is cleaner and everything stays in the `migrate` folder

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
```diff
@@ -1,30 +1,85 @@
...
-  runtime: getRuntime(packageJson),
+const resolveRuntime = async ({
+  dependencies,
+  scripts,
+}: PackageJson): Promise<Runtime> => {
+  // match `remix setup <runtime>` in `postinstall` script
+  let remixSetupMatch = scripts?.postinstall?.match(/remix setup(\s+\w+)/);
+  if (remixSetupMatch && remixSetupMatch.length >= 2) {
+    // `remix setup` defaults to `node
+    if (remixSetupMatch[1] === undefined) return "node";
```
MichaelDeBoey, line 20:
```suggestion
    if (remixSetupMatch[1] === undefined) {
      return "node";
    }
```

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -4,41 +4,135 @@ import { join } from "path";
...
+  runtime,
+  adapter,
+}: {
+  packageJson: PackageJson;
+  runtime: string;
+  adapter?: string;
+}): Generator<string> {
+  let commandTask = (message: string, command: string) =>
+    `${message}:\n\t${colors.hint(command)}`;
+  let remixRunPrefix = (dep: string) => "@remix-run/" + dep;
```
MichaelDeBoey, line 31: These functions could be declared outside of `getTasks`

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -4,41 +4,135 @@ import { join } from "path";
...
+  // other adapters in deps
+  let otherAdapters = remixDeps
+    .filter(isAdapter)
+    .filter((dep) => dep !== adapter);
+  if (otherAdapters.length > 0) {
+    yield commandTask(
+      "Uninstall all unused adapters",
+      `npm uninstall ${otherAdapters.map(remixRunPrefix).join(" ")}`
+    );
+  }
```
MichaelDeBoey, line 70: This will never be shown, as `resolveAdapter` will fail when multiple adapters are present

MichaelDeBoey (changes requested):

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
```diff
@@ -1,30 +1,99 @@
...
+      return postinstallRuntime;
+    }
+  }
+
+  // @remix-run/serve uses node
+  if (findRemixDependencies(dependencies).includes("serve")) {
+    return "node";
+  }
+  // infer runtime from adapter
+  if (adapter) return adapter2runtime[adapter];
```
MichaelDeBoey, line 42:
```suggestion
  if (adapter) {
    return adapter2runtime[adapter];
  }
```

MichaelDeBoey (changes requested):

ryanflorence:
I ran it on a couple apps, works great!

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
pcattori: We have formatting/code style enforced by prettier/eslint, and we don't currently take a stance against this style. We have many other occurrences of this in the codebase (`grep -nri ") return" packages`).

If we decide we don't want this style at some point in the future, we will proceed by adding the corresponding formatting/linting rules in eslint and autofixing all occurrences in the codebase.

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
pcattori: See https://github.com/remix-run/remix/pull/2670#discussion_r845135466

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
pcattori: See https://github.com/remix-run/remix/pull/2670#discussion_r845135466

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/getTransformOptions.ts
pcattori: See https://github.com/remix-run/remix/pull/2670#discussion_r845135466

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
pcattori: We found [@npmcli/package-json](https://www.npmjs.com/package/@npmcli/package-json) as a reliable way to automatically modify the user's `package.json` files and also determined that all major package managers will update their lockfiles from changes in `package.json` when reinstalling.

So now we are back to automatically changing Remix dependencies and removing `remix setup` from `postinstall!

pcattori:
I added types for `@npmcli/package-json` in `packages/remiv-dev/types/`. VS Code picked those up just fine, but our build step calls `tsc -b` and that command cannot find our custom types.

Blocked on this^

MichaelDeBoey:
> I added types for `@npmcli/package-json` in `packages/remiv-dev/types/` and `tsconfig.json` is configured to include `**/*` from `remix-dev/`. VS Code finds these custom types just fine, but our build step calls `tsc -b` and that command cannot find our custom types.

@pcattori I had the same problem when I was first using `is-git-clean` & added types for it.
Creating a PR in https://github.com/DefinitelyTyped/DefinitelyTyped & using `@types/is-git-clean` fixed it.

I created https://github.com/DefinitelyTyped/DefinitelyTyped/pull/59806, so this PR isn't blocked anymore and we can investigate the root cause later.

MichaelDeBoey:
@pcattori I created #2718 as that was easier to do instead of typing it all out as a review here 🙈

pcattori:
> @pcattori I created #2718 (this PR is the base branch) as that was easier to do instead of typing it all out as a review here 🙈

@MichaelDeBoey I've incorporated the feedback from your PR into this PR.

MichaelDeBoey (approved):

packages/remix-dev/__tests__/fixtures/replace-remix-imports/.github/workflows/deploy.yml
```diff
@@ -0,0 +1,205 @@
+name: 🚀 Deploy
```
kentcdodds, line 1: Could we remove unnecessary files from the fixtures? I worry that including them will cause confusion and issues in the future when people are doing a big find/replace sort of thing in the codebase and they come across files like this and spend extra time to make sure they make proper changes when it actually doesn't even matter.

So let's include only the files that can justify their existence in the fixture (which is to say that most of the files here should probably be deleted or, at the very least, drastically simplified).

packages/remix-dev/__tests__/replace-remix-imports-test.ts
```diff
@@ -0,0 +1,101 @@
...
+    "migrate",
+    "--migration",
+    "replace-remix-imports",
+    projectDir,
+    "--force",
+  ]);
+};
+
+describe("`replace-remix-imports` migration", () => {
+  it("runs successfully", async () => {
```
kentcdodds, line 72: Let's add a couple assertions in here that look for the text `from "remix"` and ensures it no longer exists in the codebase.

kentcdodds: And perhaps we could run the `build` script in the fixture after the migration to make sure imports can be resolved properly.

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -1,44 +1,174 @@
...
+    if (!flags.debug) {
+      console.log("👉 Try again with the `--debug` flag to see what failed.");
+    }
+    process.exit(1);
+  }
+  console.log("✅ Your Remix imports look good!");
+
+  console.log("\n🚚 I've successfully migrated your project! 🎉");
+  console.log(
+    "\n👉 Reinstall from your new `package.json` to update your lockfile"
```
kentcdodds, line 171:
```suggestion
    "\n👉 Reinstall from your updated `package.json` to update your lockfile"
```

kentcdodds (changes requested):
Looking great! Just a few small things.

packages/remix-dev/__tests__/replace-remix-imports-test.ts
pcattori: I added checks that `from "remix"` is purged from the app post-migration.

As for running `build`, adding `npm install` and `npm run build` to the test makes the test go from sub-second to more than 1 minute on my machine (M1 Max) and becomes network dependent. I think I prefer leaving the `build` check out and keeping this test lean. What do you think @kentcdodds ?

MichaelDeBoey (approved):
I'm just wondering why we need baa2183 & 6e1bc21.

Since we're building the whole `@remix-run/dev` package, it shouldn't matter if `jscodeshift` supports it out of the box or not I think? 🤔

packages/remix-dev/__tests__/replace-remix-imports-test.ts
kentcdodds: Yeah, definitely don't want it to be network dependent 👍 We could probably simplify the app further (remove all the deps and stuff and make remix reference the local remix), but that's more work than it's worth. This is enough 👍

kentcdodds (approved):
This is great! Lots here that should hopefully help in future code transformations we will inevitably need to make 👍 Thanks!

PR #2671, opened 2022-04-06
ci: only comment on issues and PRs when `packages` change

mcansh:
This PR reduces which PRs (and in turn which issues) will receive comments when we cut a release, in this PR they get limited to `/packages/`, but I added a nice way to choose what makes it through. It also prevents "experimental" releases from being included in comments.

It also fixes a few bugs with nightly releases as well.

And finally it prevents duplicate comments on issues when more than one PR is linked to it.

playground: https://github.com/mcansh/nightly-release-test

BONUS: it also is possible to re-use the deployment workflow using curl and and a `repository_dispatch` trigger, SO GOOD

Closes: #

- [ ] Docs
- [ ] Tests


MichaelDeBoey (approved):

chaance (approved):

.github/workflows/release.yml
```diff
@@ -86,10 +71,14 @@ jobs:
           # need this token in order to have it trigger the comment workflow
           GITHUB_TOKEN: ${{ secrets.NIGHTLY_PAT }}
 
-  stable:
-    if: github.repository == 'remix-run/remix' && github.event.action == 'published' && github.event_name != 'schedule'
+  manual:
```
kentcdodds, line 74: From what it looks like, I think we could take the nightly release job and put it into its own workflow file right?

kentcdodds (changes requested):
Just one thing, otherwise I think we're set.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,299 @@
...
+        uses: actions/create-release@v1
+        with:
+          draft: false
+          prerelease: true
+          release_name: v${{ steps.version.outputs.NEXT_VERSION }}
+          tag_name: v${{ steps.version.outputs.NEXT_VERSION }}
+        env:
+          # need this token in order to have it trigger the comment workflow
+          GITHUB_TOKEN: ${{ secrets.NIGHTLY_PAT }}
+
```
kentcdodds, line 68: Would love to figure out how to avoid the duplication between our release scripts 😬

Maybe something will turn up for this: https://twitter.com/kentcdodds/status/1513601107875885059

kentcdodds (commented):
I didn't realize we were going to run through the pre-release deploy checks on nightlies, but I think that makes sense. Hopefully we can figure out how to remove this duplication. Otherwise we should probably have a giant comment letting people know they need to change it in two places.

.github/workflows/nightly.yml
kentcdodds: It did!

Wanna look into putting all this post-release test stuff in its own workflow and use `npm hook` to kick it off?

https://docs.npmjs.com/cli/v8/commands/npm-hook

mcansh: yes! definitely - the main reason for the deployments in the first place was for pre-releases/experimental/nightlies 😉

mcansh: npm's webhook unfortantely wont work as they don't align with the body that github wants for their `repository_dispatch` events, but(!!) we can use a `needs: []` and manually curl the event to github and have that kick off the commenting 🎉  [https://github.com/mcansh/nightly-release-test/blob/main/.github/workflows/nightly.yml#L76-L98](https://github.com/mcansh/nightly-release-test/blob/main/.github/workflows/nightly.yml#L76-L98)

aside, but splitting the action into a nightly and a manual one, we need to check if the new `github.ref` in the manual action contains `nightly` and abort if so

kentcdodds: Oh, that's a great idea!

.github/workflows/nightly.yml
```diff
@@ -9,6 +9,8 @@ jobs:
     name: 🌒 Nightly Release
     if: github.repository == 'remix-run/remix'
     runs-on: ubuntu-latest
+    outputs:
```
mcansh, line 12: this allows us to use this in the "comment" job below

.github/workflows/nightly.yml
```diff
@@ -9,6 +9,8 @@ jobs:
...
+      NEXT_VERSION: ${{ steps.version.outputs.NEXT_VERSION }}
```
mcansh, line 13: this allows us to use this in the "comment" job below

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: 🏗 Build
+        run: yarn build
+
+      - name: ⤴️ Update Version
+        id: version
+        run: |
+          SHA=$(git rev-parse HEAD)
+          SHORT_SHA=${SHA::7}
+          NEXT_VERSION=0.0.0-nightly-${SHORT_SHA}
```
mcansh, line 44: i'm in between using the date instead of the commit sha, i think some folks use both?

.github/workflows/release.yml
```diff
@@ -3,93 +3,14 @@ name: 🕊 Release
...
 
 jobs:
-  nightly:
-    name: 🌒 Nightly Release
-    if: github.repository == 'remix-run/remix' && github.event_name == 'schedule'
+  manual:
+    name: 📝 Manual Release
+    if: |
+      github.repository == 'remix-run/remix' &&
+      !contains(github.ref, 'nightly')
```
mcansh, line 12: actually! being that we use a dispatch event now i think we can use the standard GitHub action token when creating a release; please hold!

mcansh: nope, this workaround is needed to have the others run 😕

.github/workflows/nightly.yml
kentcdodds: Could you add a comment?

```suggestion
    outputs:
      # used in the comment job bellow
      NEXT_VERSION: ${{ steps.version.outputs.NEXT_VERSION }}
```

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,88 @@
...
+          # need this token in order to have it trigger the comment workflow
+          GITHUB_TOKEN: ${{ secrets.FORMAT_PAT }}
+
+  comment:
+    needs: [nightly]
+    name: Comment on Issues and PRs
+    if: github.repository == 'remix-run/remix'
+    runs-on: ubuntu-latest
+    steps:
+      - name: 🛴 Kick of comment and deployment test workflows
```
kentcdodds, line 77:
```suggestion
      - name: 🛴 Kick off comment and deployment test workflows
```

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: 🐱 Create GitHub release
+        uses: actions/create-release@v1
+        with:
+          draft: false
+          prerelease: true
+          release_name: v${{ steps.version.outputs.NEXT_VERSION }}
+          tag_name: v${{ steps.version.outputs.NEXT_VERSION }}
+        env:
+          # need this token in order to have it trigger the comment workflow
+          GITHUB_TOKEN: ${{ secrets.FORMAT_PAT }}
```
kentcdodds, line 69: What is `FORMAT_PAT`? Should this be `secrets.GITHUB_TOKEN`? And is that comment right? Should this be:

```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/release.yml
kentcdodds: Could you add a comment explaining that?

scripts/release/constants.mjs
```diff
@@ -15,3 +15,9 @@ export const [OWNER, REPO] = process.env.GITHUB_REPOSITORY.split("/");
 export const LATEST_RELEASE = process.env.GITHUB_REF.replace("refs/tags/", "");
 export const GITHUB_TOKEN = process.env.GITHUB_TOKEN;
 export const GITHUB_REPOSITORY = process.env.GITHUB_REPOSITORY;
+
+/**
+  if you're changing the paths to be counted,
+  make sure you also update /.github/workflows/release.yml
+ */
```
kentcdodds, line 22: Is this still correct?

kentcdodds (changes requested):
This is coming together nicely! Just a few things.

.github/workflows/nightly.yml
mcansh: it needs to be _any_ github PAT other than the default GITHUB_TOKEN, otherwise the deployment tests and comment workflows wont run

scripts/release/constants.mjs
mcansh: nope, maybe one day when i figure out how to get a proper commit diff as `$(git rev-list $GITHUB_SHA..HEAD --count)` results in 0 as $GITHUB_SHA and HEAD are the same commit, I have ideas, but for now this will work, we just may have a few days were unneeded nightlies are released

.github/workflows/nightly.yml
mcansh: updated to use `NIGHTLY_PAT` which i created from the remix-run-bot account :)

kentcdodds (approved):
Looks solid to me 👍

If you're done, feel free to merge.

kentcdodds (approved):
Super!

PR #2655, opened 2022-04-05
fix(cli): improve the CLI

kentcdodds:
Let's get these test improvements and CLI improvements.

PR #2676, opened 2022-04-07
chore(remix-dev/tests): run `replace-remix-imports` migration over stack fixture

MichaelDeBoey:
Seems like this one was missed in #2655


kentcdodds (approved):
Thanks!

Issue #1077, opened 2021-12-16
[Docs]: Did not expect server HTML to contain a <script> in <html>


chanan:
### What version of Remix are you using?

1.0.6

### What version of Node are you using? Minimum supported version is 14.

16

### Steps to Reproduce

1. Run `npx create-remix@latest`
2. Run `yarn` to install packages
3. run `yarn dev` to start the server
4. Observe the browser dev tools console

### Expected Behavior

No errors should be shown in the console

### Actual Behavior

The following error is shown in the browser dev console:

```
react-dom.development.js:67 
        
       Warning: Did not expect server HTML to contain a <script> in <html>.
    at head
    at html
    at Document (http://localhost:3000/build/root-KCVDQOQR.js:65:21)
    at App
    at RemixCatchBoundary (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:1518:10)
    at RemixErrorBoundary (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:1457:5)
    at RemixRoute (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:2947:3)
    at Routes (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:2930:7)
    at Router2 (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:457:21)
    at RemixCatchBoundary (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:1518:10)
    at RemixErrorBoundary (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:1457:5)
    at RemixEntry (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:2827:12)
    at RemixBrowser (http://localhost:3000/build/_shared/chunk-RDYF4RHT.js:3509:42)
```

girishk21:
hey @chanan, this is a [`hydration`](https://reactjs.org/docs/react-dom.html#hydrate) warning from React. This is most probably because of your browser extensions injecting scripts into the HTML Document before React could hydrate the Server rendered document. Check out in incognito, the warning should disappear

iwfan:
Hi @chanan, You can also add `suppressHydrationWarning` in the root.tsx to suppress this warning.

```diff
export default function App() {
  return (
-    <html lang='en'>
+    <html lang='en' suppressHydrationWarning={true}>
      <head>
  ...
```

kentcdodds:
[...]

PR #2682, opened 2022-04-07
docs(gotcha): Add an entry about hydration warnings

machour:
Closes: #1077

- [x] Docs
- [ ] Tests

### DIFF
--- a/docs/pages/gotchas.md
+++ b/docs/pages/gotchas.md
@@ -120,3 +120,15 @@ if (typeof document === "undefined") {
 This will work for all JS environments (Node.js, Deno, Workers, etc.).
 
 [esbuild]: https://esbuild.github.io/
+
+## Browser extensions injecting code
+
+You may run into this warning in the browser:
+
+```
+Warning: Did not expect server HTML to contain a <script> in <html>.
+```
+
+This is a hydration warning from React, and is most likely due to one of your browser extensions injecting scripts into the server-rendered HTML, creating a difference with the resulting HTML.
+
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
Thanks!

msakrejda:
What is this documentation actually telling us? To always browse in incognito mode (and have our users do so)? That's obviously not a solution, so I don't understand what this is suggesting. Never mind that extensions can be enabled in incognito mode. To not uninstall extensions that cause this problem? That's impractical for developers and impossible to require of end-users.

I'd like to suggest better guidance here (I ran across this myself), but I don't feel like I understand the intent behind this note.

machour:
@msakrejda You are making a totally fair point here. I'll look into expanding the documentation here to provide a better explanation and guidance 👍🏼

sergiodxa-silverback:
@machour it would be better to explain what React does when that happens, because I think most people think React will fail completely, when the reality is that React will do a new render as if you didn't SSR-ed your app, instead of just attaching event listener like the normal hydration process.

sergiodxa:
@machour it would be better to explain what React does when that happens, because I think most people think React will fail completely, when the reality is that React will do a new render as if you didn’t SSR-ed your app, instead of just attaching event listener like the normal hydration process.

machour:
@sergiodxa a couple questions here 🙏🏼 

- is what you're saying true for both React 17 et React 18 ?
- is the warning logged in production, or is it only for dev mode ?
- when react fails at hydrating, and does the new render, will Remix call loaders / actions again, or does it already have the data in the SSR response ?

Issue #1355, opened 2022-01-05
[Bug]: Top-level await not supported by remix/compiler 


vnikolov88:
### What version of Remix are you using?

v1.1.1

### What version of Node are you using? Minimum supported version is 14.

v16.13.1

### Steps to Reproduce

Changed the tsconfig.json to allow top-level await

`
"module": "es2022",
"lib": ["DOM", "DOM.Iterable", "ES2019"],
"target": "es2021",
`

and changed `serverModuleFormat: "esm"` in remix.config.js

the whole point is to have a top-level await in the db.server.ts file so when mongodb wants to connect it can do in a synchronous manner. also, it is interesting it is giving warnings for dotenv and MongoDB missing from the package.json as they are present and working properly although without the top-level await.

### Expected Behavior

To work :D

### Actual Behavior

[1] The path "dotenv" is imported in app\db.server.ts but dotenv is not listed in your package.json dependencies. Did you forget to install it?
[1] The path "mongodb" is imported in app\db.server.ts but mongodb is not listed in your package.json dependencies. Did you forget to install it?
[1] 
[1]  > app/db.server.ts:24:6: error: Top-level await is not available in the configured target environment ("es2021")
[1]     24 │       await global._mongoClient.connect();
[1]        ╵       ~~~~~
[1]    tsconfig.json:6:14: note: The target environment was set to "es2021" here
[1]      6 │     "target": "es2021",

Paddy-Hamilton:
I was getting this error when running through the blog tutorial [blog tutorial](https://remix.run/docs/en/v1/tutorials/blog), seeding the db .  
```
prisma/seed.ts:93:3 - error TS1378: Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', or 'nodenext', and the 'target' option is set to 'es2017' or higher.

```

I added your changes to the `tsconfig.json` and it's worked for me. Thanks. 

Would be good to get this updated.

machour:
Closed in favor of #4348

PR #2687, opened 2022-04-07
[Docs] Blog Tutorial Workaround for "Top-level 'await' expressions" error

davidkrisch:
I was working through the excellent Blog Tutorial and got this error 

```
TSError: ⨯ Unable to compile TypeScript:
prisma/seed.ts:93:3 - error TS1378: Top-level 'await' expressions are only allowed when the 'module' option is set to 'es2022', 'esnext', 'system', or 'nodenext', and the 'target' option is set to 'es2017' or higher. 
```
when I ran `npx prisma db seed`.  This PR works around the error by wrapping the top-level `await` in a function and then running it on the next line.

This is on node v14.18.

ref: #1355

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -58,6 +58,7 @@
 - danielweinmann
 - davecalnan
 - DavidHollins6
+- davidkrisch
 - davongit
 - denissb
 - derekr
--- a/docs/tutorials/blog.md
+++ b/docs/tutorials/blog.md
@@ -366,13 +366,16 @@ Isn't it great?
   },
 ];
 
-for (const post of posts) {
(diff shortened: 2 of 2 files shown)


kentcdodds:
Thanks for the PR. Unfortunately it looks like the wording made you misunderstand where that code snippet is supposed to go. It's supposed to be placed within the async `seed` function (as it says, right before the console.log): https://github.com/remix-run/remix/blob/7d799e2b6d71f5d7ff572addca623545f975af96/examples/blog-tutorial/prisma/seed.ts#L43

If you'd like to make another PR to improve the wording that would be welcome. Thanks!

PR #2695, opened 2022-04-08
chore(remix-dev/cli): simplify `validateNewProjectPath` in `create` command

XiNiHa:
Removes a dead branch found inside the CLI code for validating project directories.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -313,6 +313,7 @@
 - visormatt
 - weavdale
 - wladiston
+- XiNiHa
 - xstevenyung
 - yauri-io
 - yesmeck
--- a/packages/remix-dev/cli/create.ts
+++ b/packages/remix-dev/cli/create.ts
@@ -368,25 +368,12 @@ export async function validateNewProjectPath(input: string): Promise<void> {
   let projectDir = path.resolve(cwd, input);
   if (
     (await fse.pathExists(projectDir)) &&
-    (await fse.stat(projectDir)).isDirectory()
-  ) {
-    let contents = await fse.readdir(projectDir);
-    if (contents.length > 0) {
-      throw Error(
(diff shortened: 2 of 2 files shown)


MichaelDeBoey:
@XiNiHa Code changes go against `dev` branch. Please rebase your branch against latest `dev` & make sure to fix all tests.

XiNiHa:
Not sure if I did it right or not, but tried to do what you've requested

XiNiHa:
@MichaelDeBoey The diff is a bit confusing, but the code does the same thing (I just moved the `contents` part into the if conditions)

MichaelDeBoey:
@XiNiHa You also removed one of the possible errors.
I think you should revert that.

XiNiHa:
@MichaelDeBoey It was impossible to hit that error. The condition (`(await fse.pathExists(projectDir)) && (await fse.stat(projectDir)).isDirectory()`) was same for two `if`s, and the first `if` throws when `contents.length > 0`, and returned otherwise. Therefore it had no chance to hit the latter error, since the second `if` will never be visited. I believe that was intended to return on `!(contents.length > 0)` since it's safe to override the contents inside the directory in that case. I'm very sure that the code is functioning exactly same.

MichaelDeBoey:
@XiNiHa The second one was indeed never hit, but I'm quite certain that was never @kentcdodds' intention in #2655.

Since that wasn't his intention, the `return` should move 1 line higher to get the intended code.

XiNiHa:
@MichaelDeBoey I guess this is what you want, but I'm still not sure if it's a good idea to error when users try to create a project on an empty directory.

MichaelDeBoey (approved):

kentcdodds (approved):
Thank you!

PR #2703, opened 2022-04-08
docs(tutorials/jokes): filter RSS jokes by user

Noriller:
As I was doing the tutorial I saw that in the deployed version, the [RSS feed](https://remix-jokes.lol/jokes.rss) shows all jokes made by anyone.
The idea of some of the features in the example was to not have to have moderation, but the open RSS feed means that people could put things in the name of the joke or even in the user that would need moderation.

That said, to keep the example with minimal modifications, I've added a filter by username and I'm hoping that `kody` has some jokes in the deployed version.

By default, it will show only jokes by `kody`, but when logged, it adds a search param `jokester` with the username, so that you could potentially share only the jokes of your username in the feed.

Finally, I've updated the tutorial page with the last version of the `jokes[.]rss.tsx` file (that had an earlier version of the code).

### DIFF
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -5553,6 +5553,19 @@ import type { LoaderFunction } from "@remix-run/node";
 
 import { db } from "~/utils/db.server";
 
+function escapeCdata(s: string) {
+  return s.replace(/\]\]>/g, "]]]]><![CDATA[>");
+}
+
+function escapeHtml(s: string) {
+  return s
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&#039;");
+}
+
 export const loader: LoaderFunction = async ({ request }) => {
   const url = new URL(request.url);
   const username = url.searchParams.get("jokester") || "kody";
@@ -5560,7 +5573,6 @@ export const loader: LoaderFunction = async ({ request }) => {
   const jokes = await db.joke.findMany({
     take: 100,
     orderBy: { createdAt: "desc" },
-    include: { jokester: { select: { username: true } } },
     where: {
       jokester: {
         username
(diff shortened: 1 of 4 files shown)


examples/jokes/app/routes/jokes[.]rss.tsx
```diff
@@ -3,10 +3,22 @@ import type { LoaderFunction } from "@remix-run/node";
...
+  // in the official deployed version of the app, we don't want to deploy
+  // a site with unmoderated content, so we only show users their own jokes
+  // or jokes from kody, which are safe to show to everyone
+  // by using search params, you can still share and view jokes from other users
+  const username = url.searchParams.get("jokester") || "kody";
+
   const jokes = await db.joke.findMany({
     take: 100,
     orderBy: { createdAt: "desc" },
     include: { jokester: { select: { username: true } } },
```
Comment (machour) [inline examples/jokes/app/routes/jokes[.]rss.tsx:16]: We might as well get rid of this `include` to avoid a table JOIN since we already know the username, and update the RSS to use that variable.

docs/tutorials/jokes.md
```diff
@@ -5553,37 +5553,27 @@ import type { LoaderFunction } from "@remix-run/node";
...
+  const username = url.searchParams.get("jokester") || "kody";
 
-export const loader: LoaderFunction = async ({
-  request,
-}) => {
   const jokes = await db.joke.findMany({
     take: 100,
     orderBy: { createdAt: "desc" },
     include: { jokester: { select: { username: true } } },
+    where: {
```
machour, line 5564: See remark below about the `include` block

docs/tutorials/jokes.md
```diff
@@ -5553,37 +5553,27 @@ import type { LoaderFunction } from "@remix-run/node";
 
 import { db } from "~/utils/db.server";
 
-function escapeCdata(s: string) {
```
machour, line 5556: Why did those got removed? We definitely want to sanitize our RSS output.

machour (changes requested):
I remember noticing this, but totally forgot to get around it, thank you for taking care 🙌

docs/tutorials/jokes.md
Noriller: the file `jokes[.]rss.tsx` didn't have that.
But I'll add it in the example and back in this file.

kentcdodds:
Thanks for this! The jokes app that's deployed to production is actually here: https://github.com/kentcdodds/jokes-app

(I wanted to keep it separate from the main repo since there were changes I needed to make for the deployed version due to moderation issues). If you'd like to make a PR there be my guest :) Thanks!

kentcdodds:
Jk, that's not the right repo... Let me find it 😅

kentcdodds:
Here it is: https://github.com/remix-run/remix-jokes

I archived it since we can't really maintain it. I'll go ahead and make this change myself. Thanks again!

kentcdodds:
jk again, looks like I did already handle that. The person requesting the RSS feed will only see their own jokes coming back on it (so you couldn't really add it to an RSS feed reader, but that's not really the point).

Thanks again!

Issue #2248, opened 2022-03-07
Uncaught ReferenceError: Buffer is not defined


abdessamadelhamdany:
### What version of Remix are you using?

v1.2.3

### Steps to Reproduce

The examples in the docs https://remix.run/docs/en/v1/api/remix#uploadhandler

Just after adding this code inside a route

```javascript
let uploadHandler = unstable_createFileUploadHandler({
  maxFileSize: 5_000_000,
  file: ({ filename }) => filename,
})
```

### Expected Behavior

To work, as expected from the docs

### Actual Behavior

Log the error to the the browser console on page load

<img width="612" alt="Screen Shot 2022-03-07 at 23 30 32" src="https://user-images.githubusercontent.com/20196583/157131749-404bb25c-13a9-4a0b-b79b-ef208fff7571.png">

mzaatar:
I have had the same problem for a few days now. following...

abdessamadelhamdany:
I think some `node` code end-up in the client side, as Buffer available for node but not for the browser
But i can't find where's it's coming from, currently looking for a way to handle uploads with external express server

Jaikant:
I am facing a similar error, while trying to include [graphql-ws](https://github.com/enisdenjo/graphql-ws) in my app.
```
constants.js:5 Uncaught ReferenceError: Buffer is not defined
    at node_modules/ws/lib/constants.js (constants.js:5:17)
    at __require2 (chunk-FDP6WGZF.js:19:44)
    at node_modules/ws/lib/buffer-util.js (buffer-util.js:3:26)
    at __require2 (chunk-FDP6WGZF.js:19:44)
    at node_modules/ws/lib/permessage-deflate.js (permessage-deflate.js:5:20)
    at __require2 (chunk-FDP6WGZF.js:19:44)
    at node_modules/ws/lib/receiver.js (receiver.js:5:27)
    at __require2 (chunk-FDP6WGZF.js:19:44)
    at wrapper.mjs:2:22
```

Jaikant:
I think my issue is documented [here](https://remix.run/docs/en/v1/pages/gotchas) (I need to try it though)

kiliman:
I show a way to import the `buffer` polyfill package. NOTE: it does require a patch of Remix compiler since it complains about importing package of node built-ins.

https://github.com/kiliman/remix-walletconnect

fergusmeiklejohn:
yeah I have the same problem :-( 
[...]

PR #2704, opened 2022-04-08
docs: Fix upload handler examples + add context

chaance:
A few of our examples incorrectly called server-only functions at the top-level scope of a route module which will result in an error when they use Node globals. This PR fixes those examples and adds clarity where needed on the "gotchas" page.

This addresses the issue raised in #2248.


machour (approved):

jacob-ebey (approved):

docs/pages/gotchas.md
```diff
@@ -55,6 +55,31 @@ export default function SomeRoute() {
 
 Even better, send a PR to the project to add `"sideEffects": false` to their package.json so that bundlers that tree shake know they can safely remove the code from browser bundles.
 
+Similarly, you may run into a the same error if you call a function at the top-level scope of your route module that depends on server-only code.
```
derenkeskin, line 58:
```suggestion
Similarly, you may run into the same error if you call a function at the top-level scope of your route module that depends on server-only code.
```

PR #2725, opened 2022-04-11
bug: useActionData only returns data for associated component

rossipedia:
From @ryanflorence's [comment](https://discord.com/channels/770287896669978684/771068344320786452/962930725542305822) in Discord:

> **ryanflorence — Today at 9:59 PM**
> nah, useActionData() is a singleton, it's available in any route

useActionData should return the data that was POSTed, regardless of where in the render tree it's being called.

However, it seems to be associated only with the component that is defined in the same file as the `action()` handler itself:

- `Form` defined in `app/routes/foo/index.tsx`, with action of `/foo`
- `action()` defined in `app/routes/foo.tsx`
- `useActionData()` is invoked in both files.
- the call in `app/routes/foo.tsx` returns any data. the one in `app/routes/foo/index.tsx` does not.

see `integration/bug-report-test.ts` in this PR for details. 

Failing test error output:

image

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -1,4 +1,9 @@
-import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
+import {
+  createAppFixture,
+  createFixture,
+  js,
+  getElement,
+} from "./helpers/create-fixture";
 import type { Fixture, AppFixture } from "./helpers/create-fixture";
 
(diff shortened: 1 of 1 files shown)


chaance:
@rossipedia Can you confirm if this issue persists in 1.10+? I'm wondering if we may have fixed it in React Router after the 6.4 updates.

PR #2731, opened 2022-04-11
feat(remix-eslint-config): add ESLint warning when using deprecated `remix` imports

MichaelDeBoey:
This way people will know that all `remix` exports are considered deprecated & that they can use the `replace-remix-imports` migration to fix it all.

I needed to copy over all values from `packageExports` of `replace-remix-imports`' migration, since those values are only available in `@remix-run/dev`.

https://github.com/remix-run/remix/blob/main/packages/remix-dev/cli/migrate/migrations/replace-remix-imports/transform/mapNormalizedImports/packageExports.ts

### DIFF
--- a/packages/remix-eslint-config/rules/core.js
+++ b/packages/remix-eslint-config/rules/core.js
@@ -1,6 +1,44 @@
+const {
+  defaultAdapterExports,
+  defaultRuntimeExports,
+  architectSpecificExports,
+  cloudflareSpecificExports,
+  cloudflarePagesSpecificExports,
+  cloudflareWorkersSpecificExports,
+  nodeSpecificExports,
+  reactSpecificExports,
+} = require("./packageExports");
+
 // const OFF = 0;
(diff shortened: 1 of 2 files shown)


kentcdodds (approved):
This is excellent 👍 👏 Thanks!

chaance:
The goal here is great, but I think I'd like to build on this by creating a separate eslint plugin with our own rules, since there are many other Remix-specific warnings that would be helpful. Instead of lumping this in with `no-restricted-imports` I'd personally like to see a separate rule for working with deprecated imports that would allow the user to auto-fix since we already have a code-mod for that. 

I know it's already merged, but this would make a great follow-up IMO.

kentcdodds:
Oh for sure it'd be great to make our own rules. If someone wants to put time into a custom rule that's great, but I think this accomplishes our purposes without requiring a bunch of extra work 🤷‍♂️

Issue #2491, opened 2022-03-25
create-remix templates don't work if a local folder of the same name as the template exists


IanWitham:
### What version of Remix are you using?

1.3.3

### Steps to Reproduce

> npx create-remix@latest
> Just the basics
> Remix Application Server
> TS or JS
> Run npm install or don't

### Expected Behavior

A new Remix app is created.

### Actual Behavior

An empty package.json and package-lock.json are produced along with a "remix" folder containing the entirety of the remix-run/remix repo.

brophdawg11:
👋 What platform did you see this on?  I'm unable to reproduce on OSX locally:

<img width="858" alt="Screen Shot 2022-03-25 at 9 25 28 AM" src="https://user-images.githubusercontent.com/1609022/160130412-4fe23529-af78-411d-a0e5-d0df2dce6cb2.png">

cloudy9101:
Got the same issue.
`npx create-remix@latest` will download the mono repo remix-run/remix to the newly created project dir.

I tried to install create-remix globally and run it manually, that's works for me. It will **only download the remix-ts template** to the newly created project dir.
```
npm i -g create-remix
create-remix
```

cloudy9101:
Got the same issue.
This will download the mono repo remix-run/remix to remix-jokes
```
npx create-remix@latest
? Where would you like to create your app? ./remix-jokes
```

But if the path has `../` in the beginning, it will **only download the template** to remix-jokes ( of course it's located in the parent dir )
```
npx create-remix@latest
? Where would you like to create your app? ../remix-jokes
```

DavOnGit:
Got the same issue.
I'm on windows 10.

styxlab:
Workaround: `px create-remix@latest  --template remix-ts`

IanWitham:
I'm seeing the bug on macOS Catalina 10.15.7. the `../my-remix-app` workaround works.

I tried running `npm uninstall -g create-remix && npm cache clean --force` first and the problem persisted.

On Windows 10 I _don't_ see the bug.

cloudy9101:
I think the root cause it's that **there's a folder called `remix`** when we run create-remix@latest.
The template name of `Just the basics` is `remix`, the detectTemplateType function will return `local` since `fse.existsSync(template)` will return true if there's a folder called `remix`.

![image](https://user-images.githubusercontent.com/5436425/160223158-8112561f-ed3c-4650-8e67-4ac716bfe260.png)

[...]

PR #2733, opened 2022-04-11
fix(remix-dev): Prioritize remix templates in CLI create

chaance:
We've received a few issues where folks would step through the interactive CLI, select one of our templates and end up with an unexpected result, largely because we always prioritize local files when detecting the input type.

Since local files are likely going to be used less frequently than our built-in stacks and templates, this PR checks for those options first and ensures that those inputs are not matched to local files/directories to avoid conflicts. If a local file or forked repo is desired, the user can pass the full path/URL accordingly.

Closes: #2491

- [ ] Docs
- [x] Tests


packages/remix-dev/cli/create.ts
```diff
@@ -577,28 +592,17 @@ export function detectTemplateType(template: string): TemplateType | null {
...
-  }
-
   // 4. examples/<template> will use an example folder in the Remix repo
   if (/^examples?\/[\w-]+$/.test(template)) {
     return "example";
   }
 
-  // 5. If the string contains no slashes, spaces, or special chars, we assume
-  //    it is one of our remix-run/remix/templates.
-  if (/^[\w-]+$/.test(template)) {
```
chaance, line 592: We don't need to do this check since we are verifying our templates statically. I think this is better anyway since we don't have too many and they're easy to verify without fetching.

chaance:
Unsure what's going on with the tests but they are hanging indefinitely at the moment. Circling back in a bit when I can get tests to run on my machine.

mcansh (approved):

PR #2735, opened 2022-04-11
feat: deprecate all functions & types exported from `magicExports`

MichaelDeBoey:
As a follow-up of #2731, this will show deprecation warnings in the IDE + in console whenever people aren't using the ESLint config.

Inspired by @pcattori's remark (https://github.com/remix-run/remix/pull/2542#discussion_r838804227) on @mjackson's PR to rename `createCloudflareKVSessionStorage` to `createWorkersKVSessionStorage` (#2542). 

Since each package is on it's own, I had to copy/paste the `warn` & `getDeprecatedMessage` functions over to each package.

If this somehow could be simplified (as this is doing the same thing over & over again), I'd be happy to do so.

### DIFF
--- a/packages/remix-architect/magicExports/remix.ts
+++ b/packages/remix-architect/magicExports/remix.ts
@@ -1,5 +1,21 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
+import * as architect from "@remix-run/architect";
+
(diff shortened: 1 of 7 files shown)


chaance:
@MichaelDeBoey Heads up that some on the team are out next week, but I suspect we'll come back to this when they're back. We may not get the deprecation notices out until then.

PR #2736, opened 2022-04-11
feat: replace node-fetch with @remix-run/web-fetch

jacob-ebey:
This PR replaces the internal usage of `node-fetch` with `@remix-run/web-fetch` to be more spec compliant and allow for ReadableStream's to be used as a body in node.

This work is critical to enable a runtime agnostic way to implement streaming in the server-runtime package and provide React 18 support to everyone.

I have also re-worked the file upload handler API to operate on an AsyncIterable<Uint8Array> for the body content. This allows for better error propagation as well as bringing `unstable_parseMultipartFormData` to all supported platforms, not just Node.

- [x] Docs
- [x] Tests


packages/remix-architect/server.ts
```diff
@@ -53,10 +49,10 @@ export function createRequestHandler({
     let loadContext =
       typeof getLoadContext === "function" ? getLoadContext(event) : undefined;
 
-    let response = (await handleRequest(
+    let response = await handleRequest(
       request as unknown as Request,
       loadContext
-    )) as unknown as NodeResponse;
```
mjackson, line 59: Let's keep using node types in this file (and in all node adapters) if we can.

packages/remix-express/package.json
```diff
@@ -15,6 +15,7 @@
     "@remix-run/node": "1.4.2"
   },
   "peerDependencies": {
+    "@remix-run/web-stream": "^1.0.2",
```
mjackson, line 18: This should probably be a regular dep.

packages/remix-express/server.ts
```diff
@@ -133,11 +145,27 @@ export function sendRemixResponse(
     res.set("Connection", "close");
   }
 
-  if (Buffer.isBuffer(nodeResponse.body)) {
-    res.end(nodeResponse.body);
-  } else if (nodeResponse.body?.pipe) {
-    nodeResponse.body.pipe(res);
+  if (nodeResponse.body) {
```
mjackson, line 148: Add a `pipeReadableStreamToNodeReadable(readableStream, nodeReadable)` util (?)

packages/remix-express/server.ts
```diff
@@ -109,21 +110,32 @@ export function createRemixRequest(
   };
 
   if (req.method !== "GET" && req.method !== "HEAD") {
-    init.body = req.pipe(new PassThrough({ highWaterMark: 16384 }));
+    init.body = new ReadableStream({
```
mjackson, line 113: Add `createReadableStreamFromNodeReadable(nodeReadable)` util ?

packages/remix-node/__tests__/parseMultipartFormData-test.ts
```diff
@@ -1,7 +1,8 @@
-import { Blob, File } from "@web-std/file";
+import { FormData as NodeFormData } from "@remix-run/web-fetch";
+import { Blob, File } from "@remix-run/web-file";
 
 import { Request as NodeRequest } from "../fetch";
-import { FormData as NodeFormData } from "../formData";
+// import { FormData as NodeFormData } from "../formData";
```
mjackson, line 5: Remove

packages/remix-node/__tests__/parseMultipartFormData-test.ts
```diff
@@ -19,8 +20,8 @@ describe("internalParseFormData", () => {
 
     let uploadHandler = createMemoryUploadHandler({});
     let parsedFormData = await internalParseFormData(
-      req.headers.get("Content-Type"),
-      req.body as any,
+      req,
+      req.formData.bind(req),
```
mjackson, line 24: Remove this param and use req.formData() method.

packages/remix-node/fetch.ts
```diff
@@ -95,8 +31,8 @@ class NodeRequest extends BaseNodeRequest {
         /multipart\/form-data/.test(contentType))
     ) {
       return await internalParseFormData(
-        contentType,
-        this.body as Readable,
+        this,
+        super.formData.bind(this),
```
mjackson, line 35: Can simplify here as well.

packages/remix-node/fetch.ts
```diff
@@ -111,29 +47,3 @@ class NodeRequest extends BaseNodeRequest {
...
-  init = { compress: false, ...init };
-
-  if (init?.body instanceof NodeFormData) {
-    init = {
-      ...init,
-      body: formDataToStream(init.body),
-    };
-  }
-
-  // Default to { compress: false } so responses can be proxied through more
```
mjackson, line 135: TODO: Double check that web-std/fetch doesn't decompress the bytes in a response so you can return a fetch() for an encoded resource directly from your loader().

packages/remix-node/index.ts
```diff
@@ -5,10 +5,10 @@ sourceMapSupport.install();
 export { AbortController } from "abort-controller";
 
 export type {
-  HeadersInit,
-  RequestInfo,
+  // HeadersInit,
+  // RequestInfo,
   RequestInit,
-  ResponseInit,
+  // ResponseInit,
```
mjackson, line 11: Let's try and keep these exports if we can, even though they are really just re-exports of global DOM types, so we don't have a breaking change here.

mjackson (changes requested):
Great start. Keep going! Let's also look into getting rid of busboy and using our own multipart parser in web-std/form-data.

mjackson (changes requested):

packages/remix-node/upload/fileUploadHandler.ts
```diff
@@ -159,14 +158,43 @@ export class NodeOnDiskFile implements File {
...
+    }
+
+    let stats = statSync(this.filepath);
+    return stats.size;
+  }
+
+  slice(start?: number, end?: number, type?: string): Blob {
+    let startOffset = this.slicer?.start || 0;
+
+    start = startOffset + (start || 0);
```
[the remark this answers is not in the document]
jacob-ebey: The `stream-slice` library takes care of this for us. I've added some tests though.

packages/remix-node/upload/fileUploadHandler.ts
```diff
@@ -159,14 +158,43 @@ export class NodeOnDiskFile implements File {
...
+    end = startOffset + (end || this.size);
```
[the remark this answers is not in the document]
jacob-ebey: Didn't need Math.min but I did update the size getter to properly reflect the slice.

examples/file-and-cloudinary-upload/app/utils/utils.server.ts
```diff
@@ -1,27 +1,44 @@
...
       (error, result) => {
         if (error) {
           reject(error);
+          return;
         }
         resolve(result);
       }
     );
-    fileStream.pipe(uploadStream);
+    dataStream.pipe(uploadStream);
```
mjackson, line 25: Update to use `writeAsyncIterableToWritable(part.data, uploadStream)`.

packages/remix-node/stream.ts
```diff
@@ -0,0 +1,196 @@
+import type { Writable } from "stream";
+import { Stream } from "stream";
+
+const { readableHighWaterMark } = new Stream.Readable();
+
+export async function pipeReadableStreamToWritable(
```
mjackson, line 6: Let's implement these APIs in here.

```
  await writeReadableStreamToWritable(request.body, nodeWritable);
  await writeAsyncIterableToWritable(part.data, nodeWritable);
  
  const webReadableStream = createReadableStreamFromReadable(nodeReadable);
```

packages/remix-node/stream.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    await read();
+  }
+
+  await read();
+
+  return Buffer.concat(chunks).toString("base64");
+}
+
+export async function readableStreamToString(stream: ReadableStream<any>) {
```
mjackson, line 49: Add optional `encoding` arg for the `buffer.toString(encoding)` and just have one function instead of 2 (get rid of `readableStreamToBase64String`)

packages/remix-node/stream.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    await read();
+  }
+
+  await read();
+
+  return Buffer.concat(chunks).toString();
+}
+
+export const readableStreamFromStream = (
```
mjackson, line 70: Rename to `createReadableStreamFromReadable`.

packages/remix-node/stream.ts
```diff
@@ -0,0 +1,196 @@
...
+  source: Stream & { readableHighWaterMark?: number }
```
mjackson, line 71: This type should be `Readable`

packages/remix-node/stream.ts
```diff
@@ -0,0 +1,196 @@
...
+      writable.end();
+      return;
+    }
+
+    writable.write(value);
+
+    await read();
+  }
+
+  await read();
```
mjackson, line 25: try/catch this and propagate the error through to the writable?

packages/remix-node/stream.ts
mjackson: maybe this belongs in server-runtime? but it can't use Buffer if it is.

mjackson (changes requested):

packages/remix-node/magicExports/remix.ts
```diff
@@ -13,4 +13,4 @@ export {
   unstable_parseMultipartFormData,
 } from "@remix-run/node";
 
-export type { UploadHandler, UploadHandlerArgs } from "@remix-run/node";
+export type { UploadHandler, UploadHandlerPart } from "@remix-run/node";
```
MichaelDeBoey, line 16: We should also keep `UploadHandlerArgs` here, as it would otherwise be a breaking change

```suggestion
export type { UploadHandler, UploadHandlerArgs, UploadHandlerPart } from "@remix-run/node";
```

packages/remix-node/index.ts
```diff
@@ -76,4 +79,8 @@ export type {
   SessionData,
   SessionIdStorageStrategy,
   SessionStorage,
+  UploadHandlerPart,
+  UploadHandler,
```
MichaelDeBoey, line 83: `@remix-run/server-runtime` should also export it

```suggestion
  UploadHandler,
  UploadHandlerArgs,
```

MichaelDeBoey (changes requested):

packages/remix-node/magicExports/remix.ts
jacob-ebey: It's a breaking change anyways as the props on the type have changed. Upload handlers are unstable so this should be fine.

packages/remix-server-runtime/data.ts
```diff
@@ -33,7 +33,7 @@ export async function callRouteAction({
   let result;
   try {
     result = await action({
-      request: stripDataParam(stripIndexParam(request)),
+      request: stripDataParam(stripIndexParam(request.clone())),
```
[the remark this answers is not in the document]
jacob-ebey: After talking with @mcansh it seems we want to keep this. It was only removed due to an issue in the old fetch polyfill.

packages/remix-node/stream.ts
jacob-ebey: I'm going to say it should stay. Using the native Buffer is going to be significantly faster than doing this ourselves in userland.

packages/remix-server-runtime/__tests__/formData-test.ts
```diff
@@ -0,0 +1,175 @@
...
+      let formData = new NodeFormData();
+      formData.set("a", "value");
+      formData.set("blob", new Blob(["blob".repeat(1000)]), "blob.txt");
+      formData.set("file", new File(["file".repeat(1000)], "file.txt"));
+
+      let underlyingRequest = new NodeRequest("https://test.com", {
+        method: "post",
+        body: formData,
+      });
+      let underlyingBody = await underlyingRequest.text();
```
mjackson, line 104: Are you running this through a `new Request` just so you can get a URL-encoded string out? Because if you are you can just use `new URLSearchParams(formData).toString()` to get the same thing.

packages/remix-server-runtime/__tests__/formData-test.ts
```diff
@@ -0,0 +1,175 @@
...
```
mjackson, line 144: Same question as above. Could possibly simplify this code using `URLSearchParams`.

mjackson (changes requested):

packages/remix-server-runtime/__tests__/formData-test.ts
jacob-ebey: It's actually a multipart formdata body. It's to test what happens when things are interrupted.

packages/remix-node/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "exclude": ["__tests__"],
   "compilerOptions": {
-    "lib": ["ES2019", "DOM.Iterable"],
+    "lib": ["ES2019", "DOM", "DOM.Iterable"],
```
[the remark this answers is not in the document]
jacob-ebey: Aren't needed. Leftover from me playing with stuff.

mjackson (approved):
LGTM!

dahlbyk:
Any obvious reason I'd be getting this after upgrading from 1.4? If not, I'll try to submit a repro or fix.

```text
TypeError: this._readableStream.getReader is not a function
    at ReadableStreamSearch.[Symbol.asyncIterator] ({redacted}/node_modules/@web3-storage/multipart-parser/cjs/src/search.js:149:41)
    at AsyncGenerator.next (<anonymous>)
    at Object.streamMultipart ({redacted}/node_modules/@web3-storage/multipart-parser/cjs/src/index.js:119:29)
    at streamMultipart.next (<anonymous>)
    at parseMultipartFormData ({redacted}/node_modules/@remix-run/server-runtime/formData.js:48:18)
```

PR #2634, opened 2022-04-04
test: use playwright for integration tests

mcansh:
the main differences with playwright compared to jest + puppeteer are
- the lack of a "watch" mode
- no `toMatchInlineSnapshot`
- describe, beforeTest, beforeEach, etc need to be `test.` prefixed
- all tests need to use `test` and not `it`

Signed-off-by: Logan McAnsh <logan@mcan.sh>

Closes: #

- [ ] Docs
- [ ] Tests

PR #2738, opened 2022-04-11
chore: remove `pptr-testing-library` dependency

MichaelDeBoey:
Seems like this one was missed in #2634


mcansh (approved):
must've missed it during a merge, thanks!

kentcdodds:
I looked for it but forgot it's not in the `@testing-library/` org. Thanks!

Issue #2488, opened 2022-03-24
[BUG]: .DS_Store files in app/routes dir breaks app


eccentric-j:
### What version of Remix are you using?

1.3.3

### Steps to Reproduce

1. Create a new remix repo
2. Create a new folder in routes like `routes/about`
3. Copy a .DS_Store file (on Mac OS X) or any non-standard encoded file into `routes/about`
4. Run `npx remix routes --json`

### Expected Behavior

Remix routes command should ignore the .DS_Store file

### Actual Behavior

Remix throws an error:

```
Error: Invalid route module file: /home/runner/remix-ds-store-breaks-routes-cmd/app/routes/about/.DS_Store
    at /home/runner/remix-ds-store-breaks-routes-cmd/node_modules/@remix-run/dev/config/routesConvention.js:74:11
    at visitFiles (/home/runner/remix-ds-store-breaks-routes-cmd/node_modules/@remix-run/dev/config/routesConvention.js:214:7)
    at visitFiles (/home/runner/remix-ds-store-breaks-routes-cmd/node_modules/@remix-run/dev/config/routesConvention.js:212:7)
    at Object.defineConventionalRoutes (/home/runner/remix-ds-store-breaks-routes-cmd/node_modules/@remix-run/dev/config/routesConvention.js:63:3)
    at Object.readConfig (/home/runner/remix-ds-store-breaks-routes-cmd/node_modules/@remix-run/dev/config.js:158:47)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.routes (/home/runner/remix-ds-store-breaks-routes-cmd/node_modules/@remix-run/dev/cli/commands.js:136:18)
    at async Object.run (/home/runner/remix-ds-store-breaks-routes-cmd/node_modules/@remix-run/dev/cli/run.js:346:7)
```

### Demo

Created a reproduction demo at https://replit.com/@eccentric-j/remix-ds-store-breaks-routes-cmd#package.json. 

### Solution

Was able to solve this by updating the ignoreRoutes in remix.config.js:

```
	ignoredRouteFiles: ['**/.*'],
```

Would you be open to a PR to make that default in https://github.com/remix-run/remix/blob/d9ae8e6478bcee326895b8e61a28a3e8d7f19fe7/templates/remix-ts/remix.config.js#L5?

machour:
I don't think the suggested fix is valid, a route can begin with a `.`, like `.well-known`.

eccentric-j:
[...]

PR #2739, opened 2022-04-11
chore: Ignore dotfiles in route subdirs too

eccentric-j:
Updates the default ignoredRoutes pattern to include dot files in route subdirs too.

Currently the default is `".*"` which ignores them in the root directory, but not in any subdirectories.

The project I'm using remix on has team members that are not developers. One of which was using Finder on OS X to browse the routes dir and it generated .DS_Store files. Other users in the ticket referenced below mention programs like Emacs and I believe vanilla vim also generate dot files to store backups so it seems like it would be helpful to fix this.

Without that change, Remix breaks with a somewhat cryptic error if the generated dotfile is not correctly formatted as a route file.

Please let me know if there's any context I'm missing.

Closes: #2488

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -75,6 +75,7 @@
 - dwt47
 - dylanplayer
 - eastlondoner
+- eccentric-j
 - edgesoft
 - edmundhung
 - efkann
--- a/templates/arc/remix.config.js
+++ b/templates/arc/remix.config.js
@@ -4,7 +4,7 @@
 module.exports = {
   serverBuildTarget: "arc",
   server: "./server.js",
-  ignoredRouteFiles: [".*"],
+  ignoredRouteFiles: ["**/.*"],
   // appDirectory: "app",
   // assetsBuildDirectory: "public/build",
   // serverBuildPath: "server/index.js",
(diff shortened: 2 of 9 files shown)


machour:
Hi @eccentric-j and thank you for your contribution. 
Can you rebase your PR to target the `dev` branch instead of `main` ?

eccentric-j:
@MichaelDeBoey That's not quite correct though. The current default ONLY ignores dot files in the ROOT app/routes directory then breaks if any dotfiles get generated in `app/routes/<subdir>`

The current setting is worse than no default because it results in inconsistent, and unexpected behavior. It should either ignore all dotfiles regardless of subdirs or nothing and leave it to users to configure.

What's the value of only ignoring dotfiles in the root routes dir but not subdirs?

eccentric-j:
This PR only changes `".*"` to `"**/.*"`. It does not include any specific files by name as I agree maintaining a list of files would be a wrong headed direction. 

@MichaelDeBoey please reconsider

chaance:
Yeah this is fine, my main concern in #2488 was adding a bunch of individual one-off exceptions to the list.

chaance (approved):

MichaelDeBoey:
Since this is a change to the templates, this should have been a PR towards `main` though 🤔

eccentric-j:
Thanks for revisiting! The original issue derailed quite a bit which was why I went for the PR.

PR #2740, opened 2022-04-11
chore(remix-eslint-config): simplify `internal` config

MichaelDeBoey:
No need to copy over `node` config environment/rules


chaance:
This was another intentional choice, as not all of our projects are going to use Node so we don't want whatever is the Node config to get copied over. At the moment of course there are no Node rules, but the plan is to add our own specific to Remix to help ID when server code will end up in the browser bundle.

PR #357, opened 2021-11-01
feat: Introduce `@remix-run/eslint-config` package

chaance:
This PR creates a new package for providing simple ESLint support for new Remix projects. The goal is to be as unopinionated as possible while enabling rules that either deal with bugs or optimization. Like the rules for `create-react-app`, I also included an optional config for projects that use Jest.

In general, these rules are even less opinionated than the `react-app` config; we enable a subset of the rules they use and leave the rest up to users. Eventually I'd also like to create our own ESLint plugin that catches issues specific to Remix.

I have not updated `create-remix` to include this package/config by default, but would like to do that in a follow-up PR once this is merged.

PR #2741, opened 2022-04-11
chore: remove `remix-eslint-config` dependencies from root `package.json`

MichaelDeBoey:
I think this was forgotten in #357


chaance:
We aren't using the packaged version directly in our config (instead referencing it from the packages folder) and when I initially created it the dependencies weren't hoisted to the root. Looks like linting worked in CI so I think we'll be good, thanks!

PR #2742, opened 2022-04-11
chore(remix-eslint-config): cleanup `peerDependencies`

MichaelDeBoey:
I also added `engines` for convenience

### DIFF
--- a/packages/remix-eslint-config/package.json
+++ b/packages/remix-eslint-config/package.json
@@ -44,13 +44,17 @@
     "@types/eslint": "^8.4.1"
   },
   "peerDependencies": {
-    "eslint": ">=8.1.0",
-    "react": ">=17.0.0",
-    "react-dom": ">=17.0.0"
+    "eslint": "^8.0.0",
+    "react": "^17.0.0 || ^18.0.0",
+    "react-dom": "^17.0.0 || ^18.0.0",
+    "typescript": "^4.0.0"
   },
   "peerDependenciesMeta": {
     "typescript": {
       "optional": true
     }
+  },
+  "engines": {
+    "node": "^12.22.0 || ^14.17.0 || >=16.0.0"
   }
 }


packages/remix-eslint-config/package.json
```diff
@@ -44,13 +44,17 @@
     "@types/eslint": "^8.4.1"
   },
   "peerDependencies": {
-    "eslint": ">=8.1.0",
-    "react": ">=17.0.0",
-    "react-dom": ">=17.0.0"
+    "eslint": "^8.0.0",
```
MichaelDeBoey, line 47: Nothing is keeping us from using `v8.0.0` instead of `v8.1.0`

packages/remix-eslint-config/package.json
```diff
@@ -44,13 +44,17 @@
...
+    "react": "^17.0.0 || ^18.0.0",
+    "react-dom": "^17.0.0 || ^18.0.0",
+    "typescript": "^4.0.0"
```
MichaelDeBoey, line 50: It was in `peerDependenciesMeta`, but not in `peerDependencies`

chaance: @MichaelDeBoey This is intentional. Users who don't use TypeScript but want to use our rules shouldn't get bugged about missing peer deps that aren't applicable to them.

chaance (changes requested):
If we could address https://github.com/remix-run/remix/pull/2742/files#r851759590 I'll merge. Thanks!

packages/remix-eslint-config/package.json
MichaelDeBoey: > Users who don't use TypeScript but want to use our rules shouldn't get bugged about missing peer deps that aren't applicable to them.

That's why it's marked as `optional` in `peerDependenciesMeta`, no? 🤔
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

chaance: @MichaelDeBoey Gah, sorry about that. Misread the diff, thanks for clarifying 👍

chaance (approved):

PR #2759, opened 2022-04-12
chore(remix-dev): choose a better port if 3000 isn't available

mcansh:
getPort gets a crazy secondary port if the first choice is taken by default, this also bumps getPort to v6 and uses dynamic imports

Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/integration/helpers/create-fixture.ts
+++ b/integration/helpers/create-fixture.ts
@@ -2,7 +2,6 @@ import path from "path";
 import fse from "fs-extra";
 import type { Writable } from "stream";
 import express from "express";
-import getPort from "get-port";
 import stripIndent from "strip-indent";
 import chalk from "chalk";
 import { sync as spawnSync } from "cross-spawn";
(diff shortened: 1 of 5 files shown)


kentcdodds (changes requested):
This looks sweet! I think it's causing issues with the integration tests. Getting a `require()` of ES Module. I'm guessing because that dynamic import is being converted into a `require()`. I'm pretty sure we can use this in our build: https://stackoverflow.com/a/63563486/971592

That way babel will leave the dynamic imports as-is so at runtime they'll still be dynamic imports and will therefore work :)

I'm pretty certain Node 12 supports dynamic imports, but I'm not 100% positive. Could you double-check?

mcansh:
> I'm pretty certain Node 12 supports dynamic imports, but I'm not 100% positive. Could you double-check?

node 13.2 according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#browser_compatibility) 😢

kentcdodds:
Yeah, either that or we'll have to bundle it somehow 🤔

kentcdodds (approved):
Stellar 👍

PR #2768, opened 2022-04-12
fix(remix-server-runtime): keep memory storage data across cache clears

### LINKED ISSUE #1832: [Bug]: createMemorySessionStorage doesn't actually work in development

sheepsteak:
Fixes #1832 

This is a repost of #1878 as I'd deleted my fork since I'd done this PR and wasn't able to update the PR anymore.

Using an object stored on global the memory session storage will now keep data when the require cache is purged.

~~I started adding tests for this but I wasn't sure if they were useful. I can add them if needed.~~ @ryanflorence said this wasn't necessary in original PR.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -278,6 +278,7 @@
 - sean-roberts
 - selfish
 - sergiodxa
+- sheepsteak
 - shumuu
 - sidkh
 - sidv1905
--- a/packages/remix-server-runtime/sessions/memoryStorage.ts
+++ b/packages/remix-server-runtime/sessions/memoryStorage.ts
@@ -5,6 +5,21 @@ import type {
   CreateSessionStorageFunction,
 } from "../sessions";
 
+let memoryStorageData: {
+  map: Map<string, { data: SessionData; expires?: Date }>;
+  uniqueId: number;
+};
+
+declare global {
+  // eslint-disable-next-line prefer-let/prefer-let
+  var __memoryStorageData:
+    | {
+        map: Map<string, { data: SessionData; expires?: Date }>;
+        uniqueId: number;
+      }
+    | undefined;
+}
+
 interface MemorySessionStorageOptions {
   /**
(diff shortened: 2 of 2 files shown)


cliffordfajardo:
I think @pcattori 's recent dev server upgrades (https://github.com/remix-run/remix/pull/5133)  might solve or alleviate this?

pcattori:
> I think @pcattori 's recent dev server upgrades (#5133) might solve or alleviate this?

The new dev server doesn't do any require cache purging, so in-memory stuff sticks around as long as your app server does. That said, for dev mode, some/most app servers will need a way to pick up changes after rebuilds and won't have any fancy server-side HMR to do so. So they'll have to use something like `nodemon`, which will restart the app server and that _will_ cause a loss of in-memory data.

Definitely an improvement since now this will only happen for rebuilds, not for navigations in the browser between rebuilds. But if you want things to persist _across_ rebuilds, then your app server needs to be managed by some watcher process. You could roll your own with `chokidar` and then do the require cache purging yourself, which would let you do the `global` trick to preserve things.

Our current vision is to invest more into the new dev server as we explore things like HMR and JIT compilation by route. Will have a better idea of what "server-side HMR" will mean for us then, which will inform what "keep in-memory stuff around in the server across rebuilds" means or how best to do it.

PR #2771, opened 2022-04-13
ci(test): cache playwright binaries

mcansh:

Closes: #

- [ ] Docs
- [ ] Tests


kentcdodds (changes requested):
This is awesome!

Should we make this to `main` as well?

mcansh:
> This is awesome!
> 
> Should we make this to `main` as well?

for sure!

PR #2772, opened 2022-04-13
add fullyParallel setting to playwright

cliffordfajardo:
## Summary of Change
- add `fullyParallel` playwright setting; this is a relatively new setting announced < 2 weeks ago

## Motivation For Change
As the codebase's test suite grows running them parallel can help reduce overall time of test suite

NOTE: I tested on a 16in intel macbook; had tons of apps and tabs open; would be curious to see what other folks get for the time since I had lots of work stuff open; the test suite was reduce by a few seconds (better than nothing 😎 ), but I think it may be worth having it enabled for a bit since it doesn't seem to harm

image

## References
- Playwright Parallelization docs: https://playwright.dev/docs/test-parallel
- Playwright `fullyParallel` docs: https://playwright.dev/docs/api/class-testproject#test-project-fully-parallel

CC @mcansh 

Closes: (N/A)
- [ ] Docs
- [X] Tests

### DIFF
--- a/integration/playwright.config.ts
+++ b/integration/playwright.config.ts
@@ -12,10 +12,10 @@ const config: PlaywrightTestConfig = {
   },
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 2 : undefined,
   reporter: process.env.CI ? "github" : [["html", { open: "never" }]],
   use: { actionTimeout: 0 },
-  fullyParallel: false,
+  fullyParallel: true,
 
   projects: [
     {
--- a/integration/playwright.config.ts
+++ b/integration/playwright.config.ts
@@ -15,7 +15,7 @@ const config: PlaywrightTestConfig = {
   workers: process.env.CI ? 1 : undefined,
   reporter: process.env.CI ? "github" : [["html", { open: "never" }]],
   use: { actionTimeout: 0 },
-  fullyParallel: true,
+  fullyParallel: false,
 
   projects: [
     {
--- a/integration/playwright.config.ts
+++ b/integration/playwright.config.ts
@@ -15,7 +15,7 @@ const config: PlaywrightTestConfig = {
   workers: process.env.CI ? 1 : undefined,
   reporter: process.env.CI ? "github" : [["html", { open: "never" }]],
   use: { actionTimeout: 0 },
-  fullyParallel: false,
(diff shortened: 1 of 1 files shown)


cliffordfajardo:
Was running some tests trying `fullyParallel: false` vs `fullyParallel:true` by running some commits;
Forgot I need to tweat it for `CI` environment 🤦 ; lol it was working with multiple workers of course since its not in CI 😆 

Was going to update the workers key inside of `playwright.config.ts` to allow more workers, after realizing this ^^

@mcansh - do you have context for why the worker count is set to 1 worker for the CI environment?

```
# playwright.config.ts
workers: process.env.CI ? 1 : undefined,
```

If we're limiting the CI machine to 1 core, this won't work since fullyParallel automagically spawns N number of workers based for you with this setting


CleanShot 2022-04-12 at 20 42 12

mcansh:
> do you have context for why the worker count is set to 1 worker for the CI environment?

that's what given to me when i ran `npm init playwright` 🤷

cliffordfajardo:
> > do you have context for why the worker count is set to 1 worker for the CI environment?
> 
> that's what given to me when i ran `npm init playwright` 🤷


Can we increase the worker count number defined in the `playwright.config.ts` file?
Currently its 1, but was wondering if there was a good reason that value was set to 1, [here on line 15](https://github.com/cliffordfajardo/remix/blob/cfajardo/add-playwright-parralellization-setting/integration/playwright.config.ts#L15)


It seems like the bare minimum cores on github runners is 2 cores, per [this documentation link here](https://github.com/github/roadmap/issues/161)

↓↓ Do you know who on the Remix team set up the github action's plan, they might be able to see via the Github UI how many cores the machine; this would help ensure the integration tests are taking maximum advantage of the cores available to save time

image

cliffordfajardo:
**Test 1  `fullyParallel: true, workers:2`:**
Result: 72s for integration test

**Test 2  `fullyParallel: false, workers:1`:**

CleanShot 2022-04-13 at 06 42 34@2x

CleanShot 2022-04-13 at 06 48 03@2x

cliffordfajardo:
Update from Playwright team regarding `fullyParallel` performance tips
- https://discord.com/channels/770287896669978684/961293179208990800/965624167418441770

I don't think its urgent going `fullyParallel` mode at this time, so I'm going to close the issue for now

TLDR: 
- spawning multiple threads/process & context switching is expensive, 
- Maybe in the future when there are more tests it will be good to revist

PR #2547, opened 2022-03-29
chore(remix-dev): remove unused `chalk-animation` dependency

MichaelDeBoey:
@chaance already commented it out in 952b0ce3, but `chalk-animation` was still in `dependencies` list.

`chalk-animation` is also causing high severity vulnerabilities warnings, so let's just remove the dependency for now.
- https://github.com/remix-run/remix/discussions/2505
- https://github.com/remix-run/blues-stack/issues/24

Since we're using git, we can always get this back if we ever want to.

PR #2778, opened 2022-04-13
chore(remix-dev): remove `chalk-animation` dependency

alejandroperea:
`chalk-animation` package is using `meow` that it is using `trim-newlines` package, version 2.0.0 that has a security issue as you can see here:

Screenshot 2022-04-13 at 17 10 07

I was checking if this package can be updated, but looks like is unmaintained. After having a look into the code, looks like this package it is not used anymore. I think this is the last usage: febf2bf391b8cd15dfaa06c83fbd33cf0e814fdf

Correct me if I'm wrong or I'm missing something (first contribution)

### DIFF
--- a/packages/remix-dev/package.json
+++ b/packages/remix-dev/package.json
@@ -24,7 +24,6 @@
     "@npmcli/package-json": "^2.0.0",
     "cacache": "^15.0.5",
     "chalk": "^4.1.2",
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
Duplicate of #2547

alejandroperea:
Sorry @MichaelDeBoey I haven't seen that one.

Thanks!

PR #2780, opened 2022-04-13
ci: more nightly updates

mcansh:
update nightly releases to use version.js script

I'm not sure why we're getting a "dirty" working directory though.. perhaps we just `git commit -m "idk man"`

Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -48,7 +48,8 @@ jobs:
           echo ::set-output name=NEXT_VERSION::${NEXT_VERSION}
 
           git checkout -b nightly/${NEXT_VERSION}
-          npm version ${NEXT_VERSION}
+
+          yarn run version ${NEXT_VERSION} --skip-prompt
 
           git push origin --tags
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -48,7 +48,16 @@ jobs:
...
 
           git checkout -b nightly/${NEXT_VERSION}
-          npm version ${NEXT_VERSION}
+          
+          if [ -z "$(git status --porcelain)" ]; then
+            echo "✨"
+          else
+            echo "dirty working directory..."
+            git add .
+            git commit -m "dirty working directory..."
```
kentcdodds, line 57: Why is there a dirty working directory? What steps trigger file changes?

kentcdodds (changes requested):
Coolio, just one question

.github/workflows/nightly.yml
mcansh: great question, that i'd love to have an answer for!

in my test repo i was getting a changed package-lock.json, but not sure what the dirtiness is here.

might be fine to leave the commit for a day and see if we can remove it tomorrow after we know what changed?

kentcdodds: Sounds good 👍

PR #2783, opened 2022-04-13
fix: LiveReload only executes on client

jacob-ebey:
no more stringifying a func and dangerously setting innerHTML

Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1392,6 +1392,7 @@ export function useFetchers(): Fetcher[] {
   return [...fetchers.values()];
 }
 
+let liveReloadMounted = false;
 // Dead Code Elimination magic for production builds.
 // This way devs don't have to worry about doing the NODE_ENV check themselves.
 export const LiveReload =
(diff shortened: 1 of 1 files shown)


mcansh (approved):
🥵

kentcdodds (approved):
That makes a ton of sense. Thanks 👍

packages/remix-react/components.tsx
```diff
@@ -1424,17 +1428,16 @@ export const LiveReload =
...
-            }}
-          />
-        );
+        };
+
+        React.useEffect(() => {
+          if (!liveReloadMounted) {
+            setupLiveReload(port);
+            liveReloadMounted = true;
+          }
```
kentcdodds, line 1437: Instead of this, why don't we put the body of `setupLiveReload` in this effect and do proper cleanup? That way we don't need to have the mutable variable.

kentcdodds (changes requested):
One idea

packages/remix-react/components.tsx
jacob-ebey: I'd rather have the mutable global here than re-connect to a port if a boundary is hit personally. This should only ever happen once and stay connected for the duration of the session.

kentcdodds: Yeah, good point 👍

kentcdodds (approved):
LGTM 👍 Thank you!

axel-xitaso:
So now live reloading no longer works on pages that require no JavaScript, and thus do not get hydrated by React? That's seems unfortunate.

kentcdodds:
That's a great point I hadn't considered. We need to go back to our previous method I think.

PR #2786, opened 2022-04-13
fix(remix-dev): apply suggested and mandatory tsconfig when values are missing

mcansh:
Add some tsconfig (and jsconfig) compilerOptions, some of these are required by esbuild, others are our options from our existing templates. Being that our tsconfig resolver walks the fs, i have the baseUrl resolving from that

image


MichaelDeBoey (approved):

jacob-ebey:
Could we potentially set it to their app dir if it's not set? Or is it better to just alert about the misconfiguration?

mcansh:
> Could we potentially set it to their app dir if it's not set? Or is it better to just alert about the misconfiguration?

for sure, I was going back on forth on it

mcansh:
gonna implement this tomorrow, do we want to suggest/enforce any other compilerOptions? maybe the config from our templates?

mcansh:
main goal for the PR is done, but going to also check for the required tsconfig settings for esbuild

mcansh:
👌 
image

mcansh:
only downside, is the build is now "slow" as it has check for these settings, I'll look at just doing the check "remix dev"

mcansh:
actually, idk why that shows it being so slow, i'm seeing it a consistent ~520ms

kentcdodds:
What's the watch rebuild time like? We're not rechecking every build are we? 500ms for initial boot is ok, but I wouldn't want it to be that slow for iterative builds during dev.

mcansh:
> What's the watch rebuild time like? We're not rechecking every build are we? 500ms for initial boot is ok, but I wouldn't want it to be that slow for iterative builds during dev.

in the playground the rebuild time is almost identical, but we should be able to do a one time check

mcansh:
added some logging to make sure things were getting recalled, and looks like the loadSync is only happening once for the initial `dev` build

kentcdodds:
How does this impact the monorepo story?

mcansh:
> How does this impact the monorepo story?

not a ton if any, but the original reason for the PR was that tsconfig-paths which we use for path aliases doesn't work (and doesn't tell you why) if there isn't a baseUrl set

mcansh:
> How does this impact the monorepo story?

like any monorepo? or ours? the original reason for the PR was that tsconfig-paths which we use for path aliases doesn't work (and doesn't tell you why) if there isn't a baseUrl set

kentcdodds:
I'm just really hesitant to auto-update someone's tsconfig for them. react-scripts does this and it's super annoying if you have a good reason for making the config the way it is. I probably don't understand the problem this is solving well enough to comfortably review/merge.

mcansh:
totally fair, this PR started as just surfacing the missing baseUrl

kentcdodds:
Alright. Let's go for it.

evan-liu:
> How does this impact the monorepo story?

Hi @kentcdodds and @mcansh, this change does introduce a problem for nx monorepo where `paths` should on be extended from the top-level `tsconfig.base.json`. The change to force add `paths` to app `tsconfig` file breaks remix apps in an `nx` monorepo.

mcansh:
@evan-liu there's a nightly linked in https://github.com/remix-run/remix/pull/2961 you can try which properly uses the extended config

evan-liu:
I will try thanks @mcansh

Issue #2801, opened 2022-04-14
Github-Token missing from template validation call


FrecksterGIT:
### What version of Remix are you using?

1.3.5

### Steps to Reproduce

When trying to use a stack from a private GitHub repository the template validation step fails.

The authentication header is missing:
https://github.com/remix-run/remix/blob/7492b9beabb75f2f3af108942253e0a28b57c1ef/packages/remix-dev/cli/create.ts#L446

### Expected Behavior

Private Stacks should work out of the box.

### Actual Behavior

Error: 🚨 There was a problem verifying the template file. Please ensure you are connected to the internet and try again later.

machour:
Fixed by @bgschiller ❤️

PR #2804, opened 2022-04-14
fix(remix-dev): fix private stack validation

bgschiller:
Closes: #2801

- [ ] Docs
- [x] Tests

I'm not feeling like this needs new documentation, since it corrects something that's already described in the docs.


mcansh (approved):

mcansh:
LGTM, im debating passing the token token via an argument instead of using the process for when inevitably change it to a CLI arg

zkdiff:
Is there any further work to be done prior to releasing this fix? I'm running into this issue and would be happy to assist if something more is needed.

machour:
I had this problem today and came up with the same patch as this PR, but it still doesn't work for me.

I'm in the last case, with a private repo provided as `owner/name` on the command line.

I'm testing using 
```shell
yarn build # after applying the patch
GITHUB_TOKEN=ghp_MY_PERSONAL_ACCESS_TOKEN node build/node_modules/create-remix/cli.js --template machour/remix-starter --yes
```

If I comment out the template validation step, it works as expected and fetches my private repository, so the token is okay..
Any clue what might be going on @bgschiller? Is this patch working for you as of now?

bgschiller:
@machour good call. It looks like GITHUB_TOKEN is ignored for web requests. I just added a commit to hit api.github.com instead. It's working for me locally with that change.

A few snapshot tests changed though. I'm not sure how much y'all care about the specific message.

packages/remix-dev/cli/create.ts
```diff
@@ -490,7 +491,11 @@ export async function validateTemplate(input: string) {
       let spinner = ora("Validating the template file…").start();
       let response;
       try {
-        response = await fetch(input, { method: "HEAD" });
+        let headers: Record<string, string> = {};
+        if (input.startsWith("https://github.com/")) {
+          headers.Authorization = `token ${process.env.GITHUB_TOKEN}`;
+        }
+        response = await fetch(input, { method: "head", headers });
```
machour, line 498: Should be cased `HEAD` (remix convention)

packages/remix-dev/cli/create.ts
```diff
@@ -518,10 +523,27 @@ export async function validateTemplate(input: string) {
...
+      let method = "HEAD";
+      if (branch) {
+        apiUrl += `/git/trees/${branch}?recursive=1`;
+      }
+      if (filePath) {
+        // When filePath is present, we need to examine the response json to see
+        // if that path exists in the repo.
+        invariant(
+          branch,
+          "expected branch to be present, since the github URL structure requires it (and we need the recursive tree of files to validate)"
```
machour, line 538: I think we should rephrase this as a less-internal explanation for the end user:
"Expecting branch to be present when specifying a path."

packages/remix-dev/cli/create.ts
```diff
@@ -533,6 +555,24 @@ export async function validateTemplate(input: string) {
 
       switch (response.status) {
         case 200:
+          if (filePath && filePath !== "/") {
+            // if a filePath is included there must also be a branch, because of how github structures
+            // their URLs. That means the api results list all files. Not directories though, so we'll
+            // try to find some file that starts with the path we're after
+            let filesWithinRepo = await response.json();
+            if (
+              !filesWithinRepo.tree.some((file: any) =>
```
machour, line 564: filesWithinRepo.tree may be undefined (see test failure mentioning `some`) and thus should be tested before access

machour (changes requested):
Almost looking good, but we definitely need to do something about the error messages. 

At the present state, I will receive a message about me not being connected to internet when I really am, so the error won't be helpful at all, and will bring us a ton of new GitHub issues 😅

bgschiller:
This is working (and tests passing) for repo URLs, but not yet for remoteTarballs that are github releases. I'm trying to tackle that now, but happy to move it into another PR if you prefer.

machour:
Thank you for walking those extra miles 🙌
If it's remote tarball from private repositories, let's finish the work in this PR. If not, let's move to another one as soon as this one is merged.

bgschiller:
:+1: will do. It is indeed remote tarball for private repos, so I'll keep truckin' on it

packages/remix-dev/cli/run.ts
```diff
@@ -186,7 +186,9 @@ export async function run(argv: string[] = process.argv.slice(2)) {
       // as possible, but inquirer will allow users to retry input rather than
       // stop the process.
       if (flags.template) {
-        await validateTemplate(flags.template);
+        await validateTemplate(flags.template, {
+          githubToken: process.env.GITHUB_TOKEN,
+        });
```
bgschiller, line 191: I'm happy to revert this and read the process value from inside `validateTemplate`, but I thought I'd be consistent with `createApp` if possible.

packages/remix-dev/cli/create.ts
```diff
@@ -380,6 +411,40 @@ function getGithubUrl(info: Omit<RepoInfo, "url">) {
...
+}
+interface ReleaseAssetInfo {
+  browserUrl: string;
+  owner: string;
+  name: string;
+  asset: string;
+  tag: string;
+}
+function getGithubReleaseAssetInfo(browserUrl: string): ReleaseAssetInfo {
+  // for example, https://github.com/deptagency/dash/releases/download/v0.0.1/stack.tar.gz
```
machour, line 427: do you mind using an anonymous example here? (this could be perceived as promoting stuff)

```suggestion
  // for example, https://github.com/owner/repository/releases/download/v0.0.1/stack.tar.gz
```

packages/remix-dev/cli/create.ts
```diff
@@ -284,10 +285,40 @@ async function downloadAndExtractTarball(
...
+    if (!assetId) {
+      throw Error(
+        "🚨 There was a problem fetching the file from GitHub. No asset was " +
+          "found at that url. Please try again later."
+      );
+    }
+    resourceUrl = `https://api.github.com/repos/${info.owner}/${info.name}/releases/assets/${assetId}`;
+    headers.Accept = "application/octet-stream";
+  }
+  let response = await fetch(resourceUrl, { headers });
```
machour, line 321: This still doesn't work for me in the private repo case (I'm just pointing to `machour/remix-starter` which doesn't have releases, so it isn't affected by the previous if branch).

In my case, headers are still an empty object. It works when I add this before this fetch:

```js
 headers = {
      Authorization: `token ${token}`,
      Accept: "application/vnd.github.v3+json",
 };
```

Could you provision headers before the `isGithubReleaseAssetUrl()` test?

machour (changes requested):
We're really close to a working solution, there's still a case to cover.

packages/remix-dev/cli/create.ts
```diff
@@ -533,6 +641,23 @@ export async function validateTemplate(input: string) {
 
       switch (response.status) {
```
machour, line 642: Also a potential DX improvement, if I run my command without providing the GITHUB_TOKEN env variable, I get this error:

```🚨 The template could not be verified. The server returned a response with a 401 status. Please double check that the template is a valid GitHub repository and try again.```

We could add a `401` case here explaining that there's an Authorization problem, and a valid GitHub token with access to this repo must be provided.

File: packages/remix-dev/cli/create.ts (hunk 5 above)
Comment (bgschiller) [reply packages/remix-dev/cli/create.ts:427]: no problem, that's just the url I was testing with.

File: packages/remix-dev/cli/create.ts (hunk 6 above)
Comment (bgschiller) [reply packages/remix-dev/cli/create.ts:320]: Fixed. Looks like there was a bug in my github-mocks that was hiding this issue. Thanks for pointing it out :)

File: packages/remix-dev/cli/create.ts (hunk 7 above)
Comment (bgschiller) [reply packages/remix-dev/cli/create.ts:641]: for sure, done.

**Review by machour — APPROVED** [review]
Awesome work @bgschiller, thank you so much!

File: packages/remix-dev/cli/run.ts (hunk 4 above)
Comment (machour) [reply packages/remix-dev/cli/run.ts:191]: @mcansh will give you feedback on this item

**Comment by machour** [comment]
@mcansh this is quiet a blocker for enterprise adoption, anything else that needs to be addressed before the merge?

**Comment by mcansh** [comment]
> @mcansh this is quite a blocker for enterprise adoption, anything else that needs to be addressed before the merge?

gonna get this one over the finish line now, thanks for the patience 🙏

**Review by mcansh — APPROVED** [review]

File: packages/remix-dev/cli/create.ts [hunk 8]
```diff
@@ -518,10 +616,27 @@ export async function validateTemplate(input: string) {
...
       let response;
+      let apiUrl = `https://api.github.com/repos/${owner}/${name}`;
+      let method = "HEAD";
+      if (branch) {
+        apiUrl += `/git/trees/${branch}?recursive=1`;
+      }
+      if (filePath) {
+        // When filePath is present, we need to examine the response json to see
+        // if that path exists in the repo.
+        invariant(
```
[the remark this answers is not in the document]
mcansh: on second thought in this context, it's next to impossible to have a file path, but not a branch, so the invariant is fine

mcansh:
> @mcansh this is quite a blocker for enterprise adoption, anything else that needs to be addressed before the merge?

I've requested @chaance to give this a look as well, as he's been in here quite a bit recently as well 😃

Issue #2402, opened 2022-03-19
[Deno] [ERROR]: Property 'reason' does not exist on type 'AbortSignal'


CanRau:
### What version of Remix are you using?

1.3.2

### Steps to Reproduce

1. npx create-remix@latest --template deno-ts
2. npm i (if not run during first step)
3. npm run dev

```bash
$ deno --version
deno 1.20.1 (release, x86_64-apple-darwin)
v8 10.0.139.6
typescript 4.6.2
```

### Expected Behavior

Running dev server

### Actual Behavior

```bash
error: TS2339 [ERROR]: Property 'reason' does not exist on type 'AbortSignal'.
    return Promise.reject(createAbortError(signal.reason));
                                                  ~~~~~~
    at https://deno.land/std@0.130.0/async/abortable.ts:25:51

TS2339 [ERROR]: Property 'reason' does not exist on type 'AbortSignal'.
  const abort = () => waiter.reject(createAbortError(signal.reason));
                                                            ~~~~~~
    at https://deno.land/std@0.130.0/async/abortable.ts:28:61

TS2339 [ERROR]: Property 'reason' does not exist on type 'AbortSignal'.
    throw createAbortError(signal.reason);
                                  ~~~~~~
    at https://deno.land/std@0.130.0/async/abortable.ts:43:35

TS2339 [ERROR]: Property 'reason' does not exist on type 'AbortSignal'.
  const abort = () => waiter.reject(createAbortError(signal.reason));
                                                            ~~~~~~
    at https://deno.land/std@0.130.0/async/abortable.ts:46:61

Found 4 errors.
```

machour:
Seeing the same thing

```
deno 1.19.3 (release, aarch64-apple-darwin)
v8 9.9.115.8
typescript 4.5.2
```

F3n67u:
I think it's caused by `compilerOptions.lib`  option in `deno.json` which is come from `deno-ts` template, If I remove the `compilerOption.lib` in `deno.json`, then the TS ERROR is gone.

```json
{
  "compilerOptions": {
    "lib": ["dom", "dom.iterable", "dom.asynciterable", "deno.ns"]
  }
}
```

=>

```json
{}
```

pcattori:
I'm unable to reproduce with:

```
deno 1.19.1 (release, aarch64-apple-darwin)
v8 9.9.115.7
typescript 4.5.2
```

[...]

PR #2805, opened 2022-04-14
fix(templates/deno): `build` and `start` run without errors

pcattori:
Closes: #2402

### DIFF
--- a/templates/deno/app/deps/@remix-run/react.ts
+++ b/templates/deno/app/deps/@remix-run/react.ts
@@ -7,4 +7,4 @@ export {
   RemixServer,
   Scripts,
   ScrollRestoration,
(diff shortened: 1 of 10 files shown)


templates/deno/app/deps/@remix-run/react.ts
```diff
@@ -1,10 +1,10 @@
...
\ No newline at end of file
+  Links,
+  LiveReload,
+  Meta,
+  Outlet,
+  RemixBrowser,
+  RemixServer,
+  Scripts,
+  ScrollRestoration,
+} from "https://esm.sh/@remix-run/react@latest?pin=v77&dev";
```
jacob-ebey, line 10: Can we use a version number instead of latest?

jacob-ebey (approved):

PR #2808, opened 2022-04-14
docs: add docs on ?index query param

brophdawg11:
Adds a section to the routing docs on the `?index` parameter used on Index Routes

- [x] Docs

### DIFF
--- a/docs/guides/routing.md
+++ b/docs/guides/routing.md
@@ -171,7 +171,7 @@ Index routes are "leaf routes". They're the end of the line. If you think you ne
 
 This usually comes up when folks are just getting started with Remix and put their global nav in `app/routes/index.jsx`. Move that global nav up into `app/root.jsx`. Everything inside of `app/routes/*` is already a child of `root.tsx`.
 
-**What is the `?index` query param?**
+### What is the `?index` query param?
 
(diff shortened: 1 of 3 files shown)


docs/guides/routing.md
```diff
@@ -171,6 +171,22 @@ Index routes are "leaf routes". They're the end of the line. If you think you ne
 
 This usually comes up when folks are just getting started with Remix and put their global nav in `app/routes/index.jsx`. Move that global nav up into `app/root.jsx`. Everything inside of `app/routes/*` is already a child of `root.tsx`.
 
+**What is the `?index` query param?**
+
+You may notice an `?index` query parameter showing up on your URLs from time to time, particularly when you are submitting a `<Form>` from an index route. This is required to differentiate index routes from their parent layout routes. Consider the following structure, where a URL such as `/sales/invoices/` would be ambiguous. Is that referring to the `routes/sales/invoices.jsx` file? Or is it referring to the `routes/sales/invoices/index.jsx` file? In order to avoid this ambiguity, Remix uses the `?index` parameter to indicate that a UL refers to the index route for a given URL.
```
machour, line 176:
```suggestion
You may notice an `?index` query parameter showing up on your URLs from time to time, particularly when you are submitting a `<Form>` from an index route. This is required to differentiate index routes from their parent layout routes. Consider the following structure, where a URL such as `/sales/invoices/` would be ambiguous. Is that referring to the `routes/sales/invoices.jsx` file? Or is it referring to the `routes/sales/invoices/index.jsx` file? In order to avoid this ambiguity, Remix uses the `?index` parameter to indicate that a URL refers to the index route for a given URL.
```

machour (commented):
Nice addition. What do you think about making its title a markdown heading so that we can link to it in an easy way?

I'd also think we could add a link to this new section from both places currently mentioning `?index` + useFetcher()

machour (approved):

kentcdodds (approved):
Stellar 👍

PR #2822, opened 2022-04-15
chore: replace meow with arg

jacob-ebey:
This removes 4 high severity vulnerabilities introduced by using such an old version. The reason for not just updating is that they only ship ESM now 😤

Closes: #

- [ ] Docs
- [x] Tests


package.json
```diff
@@ -73,6 +73,7 @@
     "@typescript-eslint/parser": "^5.12.1",
     "abort-controller": "^3.0.0",
     "abortcontroller-polyfill": "^1.7.3",
+    "arg": "^5.0.1",
```
mjackson, line 76: Why do we need this dependency in the root? I think we can remove it? It looks like meow was only ever being used in remix-dev.

mjackson (approved):

MichaelDeBoey:
I have **0** severity vulnerabilities when installing a fresh Remix project.

Also: the change to ESM was only as of `meow` v10, so we should be able to update to v9 without any problems.
https://github.com/sindresorhus/meow/releases/tag/v10.0.0

Since all the reasons mentioned in the OP seem to be obsolete, I don't know if we still want to change dependencies & do more things ourselves?

jacob-ebey:
> Since all the reasons mentioned in the OP seem to be obsolete, I don't know if we still want to change dependencies & do more things ourselves?

I still think it's worth doing just to remove some dependencies. `meow` has 60 deps and `arg` is self contained.

MichaelDeBoey (approved):

machour (approved):

MichaelDeBoey:
@jacob-ebey It seems like we have some conflicts to resolve here

Issue #2859, opened 2022-04-19
LiveReload doesn't work when TLS is enabled


isaacs:
### What version of Remix are you using?

1.3.5

### Steps to Reproduce

1. Create a server with TLS keys
2. Attach remix route handler to it
3. Load site
4. Error appears in console:

    ```
    WebSocket connection to 'wss://moxy.cow-augmented.ts.net:8002/socket' failed: An SSL error has occurred and a secure connection to the server cannot be made.
    ```

5. Change files, observe that they are not reloaded live.  (Server rebuilds, but the ws push doesn't work.)

### Expected Behavior

Expect that WebSocket.Server would be able to listen over https somehow.

### Actual Behavior

> WebSocket connection to 'wss://moxy.cow-augmented.ts.net:8002/socket' failed: An SSL error has occurred and a secure connection to the server cannot be made.

Tracked it down to this bit in `./node_modules/@remix-run/dev/cli/commands.js`

```js
  let wss = new WebSocket__default["default"].Server({
    port: config$1.devServerPort
  });
```

It appears that the only way for a ws server to speak tls is to attach to an existing server.

Something like this:

```js
  const server = config.devServerCert && config.devServerKey ? https.createServer({
    cert: config.devServerCert,
    key: config.devServerKey,
  }) : http.createServer()
  server.listen(config.devServerPort)
  let wss = new WebSocket.Server({ server });
```

mstaicu:
@isaacs [this](https://github.com/traefik/traefik/issues/7111#issuecomment-1094679207) _might_ help. I'm running a different setup than yours, but I still serve the client via HTTPS and the live reload server has to connect via `wss` due to `https` being used as the protocol

tadeaspetak:
Being able to run a `dev` server on an `https` domain (`localhost`, `lvh.me`) would be wonderful 👌

EBruchet:
Is there any chance of the above commit being accepted in the near future? Using `https` configuration locally with an Express server setup is blocking access to the `wss://` for hot reloading at the moment.

Not sure if there's another known workaround.

kiliman:
Have you tried simply including the `<LiveReload/>` component in your project and removing the protocol check (i.e., always using `ws:`).

https://github.com/remix-run/remix/blob/40a7a390063836607c76aad9754b7881f8c82303/packages/remix-react/components.tsx#L1498-L1561

EBruchet:
[...]

PR #2861, opened 2022-04-19
feat(remix-dev): add `devServerCert` & `devServerKey` config option

isaacs:
Fix: https://github.com/remix-run/remix/issues/2859

Closes: #2859

- [ ] Docs
- [ ] Tests

Lacking docs and tests, provided more as a clearer sort of bug report for #2859.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -142,6 +142,7 @@
 - illright
 - imzshh
 - isaacrmoreno
+- isaacs
 - ishan-me
 - IshanKBG
 - jacob-ebey
--- a/packages/remix-dev/__tests__/readConfig-test.ts
+++ b/packages/remix-dev/__tests__/readConfig-test.ts
@@ -27,6 +27,8 @@ describe("readConfig", () => {
         "assetsBuildDirectory": Any<String>,
         "cacheDirectory": Any<String>,
         "devServerBroadcastDelay": 0,
+        "devServerCert": null,
+        "devServerKey": null,
         "devServerPort": Any<Number>,
         "entryClientFile": "entry.client.tsx",
         "entryServerFile": "entry.server.tsx",
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -7,6 +7,9 @@ import ora from "ora";
 import prettyMs from "pretty-ms";
 import WebSocket from "ws";
 import type { Server } from "http";
+import type { Server as HTTPSServer } from "https";
+import http from "http";
+import https from "https";
 import type * as Express from "express";
 import type { createApp as createAppType } from "@remix-run/serve";
 import getPort, { makeRange } from "get-port";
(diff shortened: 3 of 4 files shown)


pascal-codetaal:
@isaacs  nice work, we can this pr :-)

roblom:
@pascal-codetaal Merging this would be **much** appreciated! Currently hacking the command.js inside node_modules does work, but well, yeah, please? :-)

kiliman:
@roblom Instead of hacking Remix, you can simply use the Express adapter and configure your server any way you want. I wrote a script to eject out of Remix App Server to Express.

https://github.com/kiliman/rmx-cli#eject-ras

chaance:
I don't think this is the solution we're looking for here. We can explore better support for custom domains, but you should be able to achieve this with a custom server and a forked `<LiveReload>` component in the mean time. Looks like @kiliman has an example: https://github.com/remix-run/remix/issues/2859#issuecomment-1220057500

In the future, we'd prefer new APIs/features to be requested as a Discussion before introducing code changes in a PR. This helps us prioritize and get things on the roadmap, which makes them much more likely to land!

isaacs:
@chaance Totally understand.

> In the future, we'd prefer new APIs/features to be requested as a Discussion before introducing code changes in a PR. This helps us prioritize and get things on the roadmap, which makes them much more likely to land!

I'll keep that in mind.  For what it's worth, in this case, I wrote a PR rather than requesting via Discussion because it was easier to express in JavaScript than in prose.  No resentment or unpleasant minutes spent on this, and my apologies if it added any friction on your end.

kyleharper:
@chaance with your feedback here would you prefer me close my related PR too?

https://github.com/remix-run/remix/pull/4107

We can manage the patch manually for now until the dev server gets some attention.

kyleharper:
> I don't think this is the solution we're looking for here. We can explore better support for custom domains, but you should be able to achieve this with a custom server and a forked `<LiveReload>` component in the mean time. Looks like @kiliman has an example: [#2859 (comment)](https://github.com/remix-run/remix/issues/2859#issuecomment-1220057500)
> 
> In the future, we'd prefer new APIs/features to be requested as a Discussion before introducing code changes in a PR. This helps us prioritize and get things on the roadmap, which makes them much more likely to land!

@chaance with your feedback here would you prefer me close my related [PR](https://github.com/remix-run/remix/pull/4107) too? (it attempts to solve the same issue)

My team can manage the patch manually for now until the dev server gets a new release (which I believe is in the works).

machour:
@kyleharper I think closing the PR and going through the proposal process will be the better course of action for now 👍🏼

chaance:
> I'll keep that in mind. For what it's worth, in this case, I wrote a PR rather than requesting via Discussion because it was easier to express in JavaScript than in prose. No resentment or unpleasant minutes spent on this, and my apologies for any added friction on your end.

No apology needed. Thanks for understanding and contributing nonetheless!

It would be great to find a way to make this work in the Remix dev server. Even nicer if we could do so without adding new config options. Environment variables could be another option to look at.

PR #2864, opened 2022-04-19
feat(remix-dev/cli): let `replace-remix-imports` migration re-install to update lockfile

MichaelDeBoey:
Now that we have `packageManager` available since #2562, we can also use it in the migrations

### DIFF
--- a/packages/remix-dev/__tests__/replace-remix-imports-test.ts
+++ b/packages/remix-dev/__tests__/replace-remix-imports-test.ts
@@ -1,10 +1,10 @@
+import os from "os";
 import path from "path";
 import fse from "fs-extra";
-import os from "os";
+import glob from "fast-glob";
+import shell from "shelljs";
 import stripAnsi from "strip-ansi";
 import type { PackageJson } from "type-fest";
-import shell from "shelljs";
-import glob from "fast-glob";
 
 import { run } from "../cli/run";
 import { readConfig } from "../config";
@@ -110,6 +110,5 @@ describe("`replace-remix-imports` migration", () => {
     expect(result.code).toBe(0);
 
     expect(output).toContain("successfully migrated");
-    expect(output).toContain("npm install");
-  }, 25_000);
+  }, 200_000);
 });
--- a/packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
+++ b/packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
@@ -1,3 +1,4 @@
+import { execSync } from "child_process";
 import { join } from "path";
 import NpmCliPackageJson from "@npmcli/package-json";
 import glob from "fast-glob";
@@ -6,6 +7,7 @@ import semver from "semver";
 
 import * as colors from "../../../../colors";
 import { readConfig } from "../../../../config";
+import { getPreferredPackageManager } from "../../../getPreferredPackageManager";
 import * as jscodeshift from "../../jscodeshift";
 import type { MigrationFunction } from "../../types";
 import type { Dependency } from "./dependency";
@@ -74,13 +76,13 @@ const shouldKeepPostinstall = (original?: string): boolean => {
 };
 
 export const replaceRemixImports: MigrationFunction = async ({
-  projectDir,
   flags,
+  projectDir,
 }) => {
   let pkg = await NpmCliPackageJson.load(projectDir);
 
   // 0. resolve runtime and adapter
-  let { runtime, adapter } = await resolveTransformOptions(pkg.content);
+  let { adapter, runtime } = await resolveTransformOptions(pkg.content);
 
   let deps = depsToEntries(pkg.content.dependencies);
   let remixDeps = deps.filter(({ name }) => isRemixPackage(name));
@@ -142,13 +144,20 @@ export const replaceRemixImports: MigrationFunction = async ({
   // write updates to package.json
   await pkg.save();
 
-  // 3. Run codemod
+  // 3. Update lockfile for new dependencies by reinstalling
+  console.log("\n💿 I'm updating your lockfile");
+  console.log(because("your dependencies changed."));
+  let packageManager = getPreferredPackageManager();
+  execSync(`${packageManager} install`, { cwd: projectDir, stdio: "inherit" });
+  console.log("✅ Your lockfile looks good!");
+
+  // 4. Run codemod
   console.log("\n💿 I'm replacing any `remix` imports");
   console.log(because("importing from `remix` is deprecated."));
   let config = await readConfig(projectDir);
   let files = glob.sync("**/*.+(js|jsx|ts|tsx)", {
-    cwd: config.appDirectory,
     absolute: true,
+    cwd: config.appDirectory,
   });
   let codemodOk = await jscodeshift.run<Options>({
     files,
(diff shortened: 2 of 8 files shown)


packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -142,13 +144,23 @@ export const replaceRemixImports: MigrationFunction = async ({
...
 
-  // 3. Run codemod
+  // 3. Reinstall from your updated `package.json` to update your lockfile
+  console.log("\n💿 I'm reinstalling your dependencies");
+  console.log(
+    because(
+      "your lockfile needs to be updated as well now that your `package.json` is updated."
+    )
+  );
+  execSync(`${packageManager} install`, { cwd: projectDir, stdio: "inherit" });
```
pcattori, line 154: Are we guaranteed that all package managers have an `install` command and that it updates the corresponding lockfile? I know its true for `npm` and `yarn`, but this puts us on the hook for getting it right for _all_ package managers.

What happens if someone is using a bespoke package manager (i.e. not `npm`, `yarn` or `pnpm`)?

MichaelDeBoey: We do the exact same call in a couple of other places.

https://github.com/remix-run/remix/blob/404f3c3f8f666a5a02fa4c48028df7bbdd8823fa/packages/remix-dev/cli/create.ts#L214-L217
https://github.com/remix-run/remix/blob/404f3c3f8f666a5a02fa4c48028df7bbdd8823fa/packages/remix-dev/cli/commands.ts#L73-L76

We currently only account for `npm`, `pnpm` or `yarn`, which all have an `install` command that also updates their corresponding lockfile.
https://github.com/remix-run/remix/blob/404f3c3f8f666a5a02fa4c48028df7bbdd8823fa/packages/remix-dev/cli/run.ts#L22-L26

I can't think of any other package manager tbh, but we can always update these install commands if people are at some point complaining about their package manager not working as expected.

pcattori: What about `lerna`?

MichaelDeBoey: `lerna` is not a package manager, but a package for managing monorepos.

`lerna` is also officially deprecated
https://github.com/lerna/lerna/issues/3121

pcattori: Ok that makes sense, though my other comment about returning `undefined` when package manager detection fails still stands.

MichaelDeBoey: Returning `undefined` when we can't figure it out ourselves has the following implications:

- `message` of `install` option is referring to it
  Should we remove the question for this option?
  Should we log something to the console?
  https://github.com/remix-run/remix/blob/ba6de76510464a7ce3b5745ab55377ef3ca2c72e/packages/remix-dev/cli/run.ts#L307-L315

- The warning that's shown when terminal isn't interactive is referring to it
  https://github.com/remix-run/remix/blob/ba6de76510464a7ce3b5745ab55377ef3ca2c72e/packages/remix-dev/cli/run.ts#L318-L327

- The `create` command is using & referring to it
  https://github.com/remix-run/remix/blob/ba6de76510464a7ce3b5745ab55377ef3ca2c72e/packages/remix-dev/cli/create.ts#L200-L205
  https://github.com/remix-run/remix/blob/ba6de76510464a7ce3b5745ab55377ef3ca2c72e/packages/remix-dev/cli/create.ts#L206-L212
  https://github.com/remix-run/remix/blob/ba6de76510464a7ce3b5745ab55377ef3ca2c72e/packages/remix-dev/cli/create.ts#L214-L217

- The `init` command is using it & passing it to `remix.init`
  https://github.com/remix-run/remix/blob/ba6de76510464a7ce3b5745ab55377ef3ca2c72e/packages/remix-dev/cli/commands.ts#L72-L77

So returning `undefined` is simpler said than done.

This feels to me like a more structural problem than the one I'm trying to solve here tbh.
So I'm suggesting to merge this PR as is and we can discuss this further on Discord with the rest of the team?

pcattori: Yea given that there aren't other major package managers I'm ok with this as-is.

I do think that returning `undefined` (or similar sentinel value) is the right thing to do, but you're right that it could be considered out-of-scope for this PR.

packages/remix-dev/__tests__/replace-remix-imports-test.ts
```diff
@@ -1,10 +1,10 @@
+import os from "os";
 import path from "path";
 import fse from "fs-extra";
-import os from "os";
+import glob from "fast-glob";
+import shell from "shelljs";
 import stripAnsi from "strip-ansi";
 import type { PackageJson } from "type-fest";
-import shell from "shelljs";
-import glob from "fast-glob";
```
pcattori, line 7: nit: we don't have linting/formatting for import ordering intentionally and generally want to avoid `git` noise from rearranging imports, just to make the diff simpler and reviewing easier. No big deal and we can merge this fine, but just mentioning it for future changes.

packages/remix-dev/cli/migrate/types.ts
```diff
@@ -1,8 +1,9 @@
 import type { Flags } from "./flags";
 
 export type MigrationFunction = (args: {
-  projectDir: string;
   flags: Flags;
+  packageManager: "npm" | "yarn" | "pnpm";
```
pcattori, line 5: Can you make a `type PackageManager = "npm" | "yarn" | "pnpm"` in the same file as `getPreferredPackageManager` instead of duplicating this type?

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
MichaelDeBoey: I agree that it would be "more correct" to have `getPreferredPackageManager` return `undefined`.

Just don't want to make this PR too big if it's unnecessary & make the implications clear to everyone.

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts
```diff
@@ -142,13 +144,23 @@ export const replaceRemixImports: MigrationFunction = async ({
...
+  console.log("✅ Your dependencies are reinstalled!");
```
pcattori, line 155: Can you change the wording to refer to the user-facing task of "updating the lockfile" rather than the implementation detail of "by reinstalling dependencies"?

For example:

```suggestion
  // 3. Update lockfile for new dependencies by reinstalling
  console.log("\n💿 I'm updating your lockfile");
  console.log(
    because(
      "your dependencies changed."
    )
  );
  execSync(`${packageManager} install`, { cwd: projectDir, stdio: "inherit" });
  console.log("✅ Your lockfile looks good!");
```

pcattori (changes requested):

packages/remix-dev/cli/migrate/types.ts
MichaelDeBoey: Was thinking about this too, but this would create a circular dependency if we start importing it everywhere.

Maybe we should extract the function (together with the type) into a separate file instead?

MichaelDeBoey: If we do se, we could even stop passing it and just call `getPreferredPackageManager` whenever we need it.

pcattori: Yea I like that!

MichaelDeBoey: Wasn't necessary to have a separate type anymore, since they all use the function now.

Did it anyway, so it's a bit cleaner.

pcattori (approved):

PR #2865, opened 2022-04-19
docs(pages/FAQ): ad another nested object solution for form, using Ramda

alireza-bonab:
Closes: Here is a solution to have nested structured object out of HTML form using [Ramda js](https://ramdajs.com/)

- [x] Docs
- [ ] Tests


chaance (changes requested):
I think we'd prefer to keep detailed implementations with third-party libs to a minimum. I don't mind linking out to Ramda as a tool that can be used, but its implementation could change at any time and this adds to the maintenance overhead of our own docs. 

If you could update this PR to remove everything except for the link to the package, I'll be happy to merge.

alireza-bonab:
Since there is already a solution on using the `query-string` package, I thought it would be helpful to add another solution using `ramda` this time.

docs/pages/faq.md
```diff
@@ -204,6 +205,69 @@ export async function action({ request }) {
...
+  const formData = await request.formData();
+  const pairs = Array.from(formData);
+
+  // create nested structure object
+  const data = reduce(reduceFn, {}, pairs);
+
+  return json(data);
+};
+```
+### Other solutions
```
chaance, line 270:
```suggestion
```

Remove lib-specific docs

chaance (changes requested):
Requested changes. Thanks again!

docs/pages/faq.md
```diff
@@ -177,6 +177,7 @@ export async function action({ request }) {
 
 Using the same input name and `formData.getAll()` covers most cases for wanting to submit structured data in your forms.
 
+### Structured data using [`query-string`][query-string]
```
chaance, line 180:
```suggestion
```

chaance (changes requested):

chaance (approved):

PR #2866, opened 2022-04-19
feat: add support for Netlify Edge Functions

ascorbic:
This PR adds a Remix adapter for the beta release of Netlify Edge Functions. It is implemented as a package `@remix-run/netlify-edge` and uses the Deno runtime with some small but important changes. Most significantly, this adapter bundles all of the generated server code (except the adapter itself), and by default expects that user code will use regular node-style imports rather than URL imports as expected by Deno. The generated server code is Deno-compatible, and can be used directly as a Netlify Edge Function. It is emitted into `.netlify/edge-functions/server.js`. It also generates a manifest in `.netlify/edge-functions/manifest.json`. The Deno runtime code is currently inlined inside the package, because it isn't published anywhere at present.

This PR also adds some tooling to make life easier when working on Deno and Node code in the same repo. This includes defining `deno.enablePaths` in the VS Code settings, so that it only runs on Deno code, and adding an import map so that Deno code that uses bare imports (such as the imports of `build`) works as expected.

It extends the default Netlify template, adding a `remix.init` script that asks if the user want to use Edge Functions, and setting the project up accordingly.

Closes: #

- [ ] Docs
- [ ] Tests


packages/remix-netlify-edge/mod.ts
```diff
@@ -0,0 +1,48 @@
...
+  SessionIdStorageStrategy,
+  SessionStorage,
+} from "https://esm.sh/@remix-run/server-runtime@1.4.1";
+export {
+  createSession,
+  isCookie,
+  isSession,
+  json,
+  redirect,
+} from "https://esm.sh/@remix-run/server-runtime@1.4.1";
```
MichaelDeBoey, line 42:
```suggestion
} from "https://esm.sh/@remix-run/server-runtime?pin=v77";
```

packages/remix-netlify-edge/server.ts
```diff
@@ -0,0 +1,58 @@
+import type { ServerBuild } from "https://esm.sh/@remix-run/server-runtime@1.3.5?pin=v77";
+import { createRequestHandler as createRemixRequestHandler } from "https://esm.sh/@remix-run/server-runtime@1.3.5?pin=v77";
```
MichaelDeBoey, line 2: These are referencing older versions.

We should at least update to latest version

```suggestion
import type { ServerBuild } from "https://esm.sh/@remix-run/server-runtime@1.4.1?pin=v77";
import { createRequestHandler as createRemixRequestHandler } from "https://esm.sh/@remix-run/server-runtime@1.4.1?pin=v77";
```

Or remove version number all together

```suggestion
import type { ServerBuild } from "https://esm.sh/@remix-run/server-runtime?pin=v77";
import { createRequestHandler as createRemixRequestHandler } from "https://esm.sh/@remix-run/server-runtime?pin=v77";
```

templates/netlify/remix.init/edge-server.js
```diff
@@ -0,0 +1,8 @@
+// Import path interpreted by the Remix compiler
+import * as build from "@remix-run/dev/server-build";
+import { createRequestHandler } from "@remix-run/netlify-edge";
```
MichaelDeBoey, line 3: Shouldn't we have both `@remix-run/dev` & `@remix-run/netlify-edge` in our `dependencies` when choosing for `"Netlify Edge Functions"`? 🤔

packages/remix-netlify-edge/package.json
```diff
@@ -0,0 +1,29 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/remix-run/remix/issues"
+  },
+  "files": [
+    "**/*.ts",
+    "index.js"
+  ],
+  "peerDependencies": {
+    "@remix-run/server-runtime": "*"
```
MichaelDeBoey, line 27: This should probably go into `dependencies` I think? 🤔

templates/netlify/package.json
```diff
@@ -10,7 +10,7 @@
     "start": "cross-env NODE_ENV=production netlify dev"
   },
   "dependencies": {
-    "@netlify/functions": "^0.10.0",
+    "@netlify/functions": "latest",
     "@remix-run/netlify": "*",
     "@remix-run/node": "*",
```
MichaelDeBoey, line 15: We should probably replace this with `@remix-run/server-runtime` when choosing for `"Netlify Edge Functions"` in `remix.init`

Issue #2714, opened 2022-04-10
Create-Remix not working


ZetraDev:
### What version of Remix are you using?

/

### Steps to Reproduce

yarn create remix > YOUR_PATH > Just the basic > Cloudflare Pages > TypeScript > No
using npx still cause error
using other deploy target also cause error

### Expected Behavior

Work

### Actual Behavior

It returns an error:
```🚨 The provided template must be a Remix project with a `package.json` file, but that file does not exist in D:\Project\test-remix\package.json```

**Comment by anantoghosh** [comment]
https://github.com/remix-run/remix/discussions/2709#discussioncomment-2539300

Seems like it only affects windows. @kentcdodds.

**Comment by machour** [comment]
Tracked in #2691 
For now you can use `npx create-remix@1.3.4`

**Comment by anantoghosh** [comment]
@machour that looks like an unrelated issue. This issue is related to npx create-remix tar extraction skipping all files.

quote from the discussion
https://github.com/remix-run/remix/blob/7d799e2b6d71f5d7ff572addca623545f975af96/packages/remix-dev/cli/create.ts#L255-L271
At least on windows, header.name.startsWith(filePath) is always false causing all files to get ignored.

**Comment by machour** [comment]
ReOpening so that @mcansh can take a look

**Comment by SangeetAgarwal** [comment]
I'm seeing the same issue on my windows machine, strangely enough, the `pre-configured stack, ready for production` works just fine

**Comment by codthing** [comment]
> Tracked in #2691 For now you can use `npx create-remix@1.3.4`

I don't understand why it's not friendly to Window users, it seems that only Apple is you guys choice.

**Comment by kentcdodds** [comment]
> I don't understand why it's not friendly to Window users, it seems that only Apple is you guys choice.

This is a bug. It's not intentional. It will be fixed.

**Comment by mmattbtw** [comment]
> I don't understand why it's not friendly to Window users, it seems that only Apple is you guys choice.

I'd recommend just using WSL for the meantime.

**Comment by chaance** [comment]
Anyone on Windows able to give the latest pre-release a run and see if it fixes your issues?

```sh
npx create-remix@1.4.1-pre.1
```

washedupdeveloper:
> Anyone on Windows able to give the latest pre-release a run and see if it fixes your issues?
> 
> ```shell
> npx create-remix@1.4.1-pre.1
> ```

I am able to get it working just fine, having created the remix app server deployment option.
[...]

PR #2867, opened 2022-04-19
fix(remix-dev/create): use path.split and path.join to always make path posix style

mcansh:
the CLI needs to work on Node 12, but replaceAll wasn't added until [Node 15](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility)

Closes: #

- [ ] Docs
- [ ] Tests


chaance:
This should address https://github.com/remix-run/remix/issues/2714#issuecomment-1101835952

chaance (approved):

PR #2874, opened 2022-04-19
fix: re-inline live reload script

jacob-ebey:
fix: manually set REMIX_DEV_SERVER_WS_PORT in config for users running `remix dev`
fix: read devServerPort from ENV at build time

Closes: # https://github.com/remix-run/remix/pull/2783#issuecomment-1102169177

- [ ] Docs
- [x] Tests

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1409,15 +1409,18 @@ export const LiveReload =
          */
         nonce?: string;
       }) {
+        let js = String.raw;
         return (
           <script
             nonce={nonce}
             suppressHydrationWarning
             dangerouslySetInnerHTML={{
-              __html: `
+              __html: js`
                 let protocol = location.protocol === "https:" ? "wss:" : "ws:";
                 let host = location.hostname;
-                let socketPath = \`\${protocol}//\${host}:${port}/socket\`;
+                let socketPath = protocol + "//" + host + ":" + ${String(
+                  port
+                )} + "/socket";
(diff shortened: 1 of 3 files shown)


packages/remix-react/components.tsx
```diff
@@ -1408,36 +1409,35 @@ export const LiveReload =
...
-        }, []);
-
-        return null;
+        return (
+          <script
+            nonce={nonce}
+            suppressHydrationWarning
+            dangerouslySetInnerHTML={{
+              __html: `
+                let protocol = location.protocol === "https:" ? "wss:" : "ws:";
```
kentcdodds, line 1418: Are we concerned about using `let` instead of `var` here? It won't be compiled, but maybe we don't care?

Also, I can't remember whether `let` variables will be added to the global scope. I think they won't, but we'll want an iife otherwise.

kentcdodds (commented):
Just some quick questions. Can you explain what was broken about our previous implementation of the inline script and how this is different?

packages/remix-react/components.tsx
jacob-ebey: They do cross script boundaries and end up on the global scope, good catch. Just iife'd it.

kentcdodds (commented):
Looks good to me. Did you try it out in a playground? If so, go ahead and merge it. I hope this is the last time we have to fix live reload 😅

kentcdodds (approved):

PR #2875, opened 2022-04-19
feat: clean directories before initial builds

jacob-ebey:
Closes: #

- [ ] Docs
- [ ] Tests

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -317,7 +317,6 @@ function cleanBuildDirectories(config: RemixConfig) {
     fse.emptyDirSync(config.assetsBuildDirectory);
     fse.emptyDirSync(path.dirname(config.serverBuildPath));
   } catch (err) {
-    console.error(err);
     console.error("Could not clean build directories");
   }
 }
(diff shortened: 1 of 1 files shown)


jacob-ebey:
> Could we put this in `buildEverything` in `compiler.ts`? That way it can be in one place and it'll get cleared between builds in watch mode as well.

`buildEverything` is actually called for re-builds as well in dev mode. We cleared out inbetween builds in the past and it caused some issues with missing assets, etc. I'd rather keep it explicit in the CLI commands layer, not compiler layer.

jacob-ebey:
> Could we put this in `buildEverything` in `compiler.ts`? That way it can be in one place and it'll get cleared between builds in watch mode as well.

`buildEverything` is actually called for re-builds as well in dev mode. We cleared out inbetween builds in the past and it caused some issues with missing assets, etc. I'd rather keep it explicit in the CLI commands layer, not compiler layer.

kentcdodds (approved):
Sounds fine to me 👍 Thanks for explaining that to me.

nicksrandall:
@kentcdodds @jacob-ebey this breaks most of the "stacks" as the "server.js" file is cleared before the server can start. IMHO, I think this change needs to be reverted until we figure out how to handle these cases.

kentcdodds:
Thanks @nicksrandall.

Let's get this reverted.

PR #2876, opened 2022-04-19
enhancement: Export additional types from `@remix-run/react`

chaance:
This PR exposes additional types that are helpful when passing values via context or props.

Complex return types from functions should be exposed directly IMO. Usage is a little simpler than relying on `ReturnType<typeof someFunc>`, and in the case of generics like `Fetcher` and `FetcherWithComponents` it is necessary.

```ts
// instead of this...
type Ctx = {
  fetcher: ReturnType<typeof useFetcher>;
  location: ReturnType<typeof useLocation>;
};

// ...I can do this
type Ctx = {
  fetcher: FetcherWithComponents<SomeData>;
  location: Location;
};
```

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -8,7 +8,7 @@ import type {
   TouchEventHandler,
 } from "react";
 import * as React from "react";
-import type { Navigator } from "react-router";
+import type { Navigator, Params } from "react-router";
 import {
   Router,
   Link as RouterLink,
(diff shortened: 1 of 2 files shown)


mikeybinnswebdesign:
Another good type to export would be `Transition`, instead of `ReturnType<typeof useTransition>`.
I used to be able to import it from `import type { Transition } from "@remix-run/react/transition";` but this is no longer possible, and there's no export for `@remix-run/react`.

I'm not sure if you can update this PR with this very similar request or if I should make a new one, let me know 😄

Issue #2896, opened 2022-04-19
Back navigation from ErrorBoundary does not work properly if ErrorBoundary does not include <Scripts />


fnimick:
### What version of Remix are you using?

1.3.5

### Steps to Reproduce

- Clone and run https://github.com/machour/francis-issue (thanks to @machour for helping to investigate this!)
- Navigate from `/` to `/boom` via the link, confirm that the error boundary renders
- Navigate back via browser controls, confirm that the root page correctly renders.
- Navigate to `/boom` via the link, *reload the browser tab*, then navigate back. The error boundary will remain mounted and the app is unusable.

This issue has been confirmed intermittently on Chrome, and repeatably on Safari and Firefox on OSX 10.12.

Of note: if `<Scripts />` is added to the `ErrorBoundary`, back button behavior after a refresh works as expected. This only occurs when the top-level error boundary does not load javascript.

### Expected Behavior

Navigating back after reloading should return to the preceding page.

### Actual Behavior

Navigating back after reloading, to any number of parent pages, leaves the error boundary mounted. A full reload is required to interact with the app.

machour:
Interesting fact: the issue only happens if there's no `<Scripts/>` in the ErrorBoundary 👀

machour:
Opened a bug-report PR at #2919.
Oddly enough, the bug fails with or without `<Scripts />` in the boundary

machour:
Closing this in favor of #2919

brophdawg11:
Cross-referencing https://github.com/remix-run/remix/pull/2919#issuecomment-1629665909 from here - this was determined to be expected behavior

PR #2919, opened 2022-04-19
Bug: Back navigation from ErrorBoundary does not work properly #2896

machour:
Reproduce case for #2896


brophdawg11:
I don't think this is a bug, if you reload after a `pushState` the browser still has that history stack, so when you click the back button you get a `popstate` event instead of a document reload.  If you don't include `<Scripts/>` in the `ErrorBoundary` then there's no client-side JS to detect the `popstate` and update accordingly.

I moved this test into the existing error boundary tests to ensure it's covered when `<Scripts />` is present.

PR #2931, opened 2022-04-20
fix: bundle netlify-edge

ascorbic:
Using URL imports for server runtime seems to be causing problems with conflicting versions of React. The best workaround looks like it is to bundle everything at build time, rather than trying to use Deno URL imports. This PR changes the runtime from being distributes as TypeScript source for Deno, and instead bundles it like other runtimes. The generated server functions are still Deno-compatible, but it no longer imports anything from esm.sh etc.

Closes: #

- [ ] Docs
- [ ] Tests


.vscode/settings.json
```diff
@@ -2,10 +2,6 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "deno.enablePaths": [
     "./templates/deno/",
-    "./templates/netlify/remix.init/edge-server.js",
-    "./packages/remix-netlify-edge/mod.ts",
-    "./packages/remix-netlify-edge/server.ts",
-    "./packages/remix-netlify-edge/remix-deno/"
```
ascorbic, line 8: These are now Node-compatible files, rather than Deno

packages/remix-dev/compiler.ts
```diff
@@ -444,46 +444,6 @@ function createServerBuild(
...
-          "@remix-run/netlify-edge": runtimePath,
-          "@remix-run/dev/server-build": `file://${buildPath}`,
-          "@remix-run/server-runtime":
-            "https://esm.sh/@remix-run/server-runtime@1.4.1?pin=v77",
-        },
-      };
-
-      fse.ensureDirSync(path.dirname(importMapPath));
-      fse.writeJSONSync(importMapPath, importMap);
-    }
```
ascorbic, line 486: Don't need import maps, because it's all bundled

chaance (approved):

PR #2938, opened 2022-04-20
fix(remix-react): avoid duplicate `loader` calls when using `prefetch-intent` 

### LINKED ISSUE #1249: [Bug]: redundant loader invocation when using prefetch="intent"

1.1.1

>14

### Steps to Reproduce

1. Visit this repro -> https://stackblitz.com/edit/node-huqahk 
2. go to "users" page (you'll see a list of "users" displayed) (url: `/users`)
3. click on a a name to see some user details (url: `/users/$userId`)
    - check out the console, you'll see that a http call has been made to fetch the user details
4. now, click on an another user name
    - check out the console, you'll see that a http call has been made to fetch the new user details
    - **but another call has been made, with the previous id**

### Expected Behavior

When switching to a different user (url: `/users/$userId`), I expect the loader (`app/routes/users/$userId.tsx`) to be called only once, not twice

### Actual Behavior

When switching to a different user (url: `/users/$userId`), the loader (`app/routes/users/$userId.tsx`) is called twice:
- one time to fetch the current user details
- and a second time with the previous user id
- as mentioned in the title, this phenomenon occurs only when the link has its **prefetch** attribute set to "intent" (or "render" for that matter)

### 📽 video
Kapture 2021-12-27 at 18 20 40

brophdawg11:
This fixes a nuanced issue when using `<Link prefetch="intent" />` due to prefetch cache behavior.  Subsequent `<link rel="prefetch">` elements to the same locations as previous prefetch elements will not leverage prior prefetch cache entries unless the endpoint returned a `cache-control` header.  Couldn't find much official docs on this, but https://github.com/GoogleChromeLabs/quicklink/issues/21#issuecomment-446710070 was the most relevant bit I found.

So, in the scenario described by #1249 when a loader does not return `cache-control`, the behavior ended up being:

1. Hover on `/users/1`
   * Adds prefetch links to call the loader for `/users/1` which has no `cache-control` header
2. Navigate to `/users/1`
   * Standard loader calls use the prefetch cache entry for `/users/1`
   * Prefetch link elements are removed since it's now an active route
3. Navigate to `/users/2` 
   * Prefetch links were being added back for `/users/1` since it was still in a `shouldPrefetch` state from it's prior hover and is no longer active
   * These subsequent prefetch links would not use the pre-existing prefetch cache and without cache-control would cause another hit to the server for the now inactive route.

This PR updates it such that `shouldPrefetch` is unset upon navigating to the link in question, thus requiring another hover/focus intent to re-add `<link rel="prefetch">` elements.  If developers wish to avoid subsequent hits to their loaders on _those_ calls they should leverage a `cache-control` header  which will be respected by the prefetch links.

Closes #1249

- [x] Tests


packages/remix-react/components.tsx
```diff
@@ -423,6 +425,17 @@ function usePrefetchBehavior(
...
 
+  React.useEffect(() => {
+    // If we navigate to a route that we potentially prefetched for, unset
+    // shouldPrefetch so that when we navigate away from this route the links
+    // don't get re-inserted and cause a dup call to the loader.  We should
+    // require another intent to re-insert
+    if (prefetch === "intent" && linkHref === currentHref) {
+      setMaybePrefetch(false);
+      setShouldPrefetch(false);
+    }
```
brophdawg11, line 436: note: look into moving this to `cancelIntent` instead

integration/prefetch-test.ts
```diff
@@ -185,7 +185,29 @@ test.describe("prefetch=intent (hover)", () => {
       "#nav link[rel='modulepreload'][href^='/build/routes/without-loader-']",
       { state: "attached" }
     );
-    expect(await page.locator("#nav link").count()).toBe(3);
+    expect(await page.locator("#nav link").count()).toBe(1);
```
brophdawg11, line 188: @ryanflorence Now that we `setShouldPrefetch(false)` in `cancelIntent` these tests are invalid since the link tags remove - just confirming that's expected 👍

MichaelDeBoey (approved):

Issue #184, opened 2021-05-24
type `LinkDescriptor` shouldn't require href for preloading responsive images


Graham42:
For a [responsive image](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) with a `srcset` and `sizes`, sites should be able to preload as "image" with `imagesizes` and `imagesrcset`.  For this use case it's common to omit the `href` on the link element so that unsupporting browsers won't request an unnecessary image.

Currently the `LinkDescriptor` type requires the `href` attribute, but I think it should be optional when and `imagesrcset` is provided.

Remix version: 0.17.2

Graham42:
This looks like it has been fixed in [1.6.0](https://github.com/remix-run/remix/releases/tag/v1.6.0) by https://github.com/remix-run/remix/pull/2944

Thank you @chaance!

PR #2944, opened 2022-04-20
fix: Better support for `imageSrcSet` + `imageSizes` in route `links` (#184)

chaance:
This PR improves support for returned links that include an `imagesrcset` attribute to preload images when matched against a responsive set of URLs. This only works when `rel === 'preload'` and `as === 'image'`.  https://html.spec.whatwg.org/commit-snapshots/cb4f5ff75de5f4cbd7013c4abad02f21c77d4d1c/#attr-link-imagesrcset

Prior to this PR, a few problems with our implementation:

1. Our types for `HtmlLinkDescriptor` make `href` required. This is true *unless* a link includes `imagesrcset` and `imagesizes`, though `imagesrcset` and `href` will also work with or without `imagesizes` according to the spec. This PR uses an intersection type to account for these cases.
2. React 17 was released prior to these attributes being accepted into the spec and implemented by browsers, so passing the props as camel-cased will throw a warning about unsupported DOM props. Our types recognized this and supported the all-lowercase version to skip the warning and forward the attribute to the node, but React 18 now recognizes the camel-cased version. This PR updates our types to use the camel-cased prop names, but it will support either at runtime and normalize the object based on the detected version of React.
3. Because we use `href` + `rel` as a key, links without an href will have duplicate keys. So `imageSrcSet` is added as a fallback to prevent that.

---

Closes: #184

- [ ] Docs
- [x] Tests

### DIFF
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -16,16 +16,11 @@ type LiteralUnion<LiteralType, BaseType extends Primitive> =
   | LiteralType
   | (BaseType & Record<never, never>);
 
-/**
- * Represents a `<link>` element.
- *
- * WHATWG Specification: https://html.spec.whatwg.org/multipage/semantics.html#the-link-element
- */
-export interface HtmlLinkDescriptor {
+interface HtmlLinkProps {
   /**
    * Address of the hyperlink
    */
-  href: string;
+  href?: string;
 
   /**
    * How the element handles crossorigin requests
@@ -92,16 +87,6 @@ export interface HtmlLinkDescriptor {
   sizes?: string;
 
   /**
-   * Images to use in different situations, e.g., high-resolution displays, small monitors, etc. (for rel="preload")
-   */
-  imagesrcset?: string;
-
-  /**
-   * Image sizes for different page layouts (for rel="preload")
-   */
-  imagesizes?: string;
-
-  /**
    * Potential destination for a preload request (for rel="preload" and rel="modulepreload")
    */
   as?: LiteralUnion<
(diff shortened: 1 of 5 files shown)


ryanflorence:
I’ve found these kinds of tests to be flakey because the browser can decide not to prefetch given different memory/cpu/who-knows-what.  Maybe we can assert on the rendered link tags instead of requests?

ryanflorence:
Also, maybe we can just use the DOM types here?

chaance:
> Also, maybe we can just use the DOM types here?

Unfortunately the DOM types also require `href` so this wouldn't really fix things for our end users, and React's `LinkHTMLAttributes` type is different for 17 vs. 18. I think our custom types are a big DX improvement for both, personally.

packages/remix-react/links.ts
```diff
@@ -91,16 +86,6 @@ export interface HtmlLinkDescriptor {
...
-  /**
-   * Images to use in different situations, e.g., high-resolution displays, small monitors, etc. (for rel="preload")
-   */
-  imagesrcset?: string;
-
-  /**
-   * Image sizes for different page layouts (for rel="preload")
-   */
-  imagesizes?: string;
-
```
MichaelDeBoey, line 103: Removing this will technically be a breaking change I think 🤔😕

chaance: Yeah, I can add these back and just mark them as `@deprecated`

MichaelDeBoey: We can remove them in Remix v2 👍

chaance: Appended here so they are easy to remove later https://github.com/remix-run/remix/blob/6e5a89c97a5a67bcd135cce88c9d0787ec7b3e08/packages/remix-react/links.ts#L186

packages/remix-server-runtime/links.ts
```diff
@@ -1,13 +1,14 @@
...
- * Represents a `<link>` element.
- *
- * WHATWG Specification: https://html.spec.whatwg.org/multipage/semantics.html#the-link-element
- */
-export interface HtmlLinkDescriptor {
+type Primitive = null | undefined | string | number | boolean | symbol | bigint;
+
+type LiteralUnion<LiteralType, BaseType extends Primitive> =
+  | LiteralType
+  | (BaseType & Record<never, never>);
```
MichaelDeBoey, line 5: Is there a reason why we're not using `type-fest` here? 🤔

mcansh: at the time the only type we used from type-fest was this, but i think we actually have the dep now anyways

https://github.com/remix-run/remix/pull/1123#issuecomment-1035921934

MichaelDeBoey: Just looked into this & we have `type-fest` in `@remix-run/dev`, but it is in `devDependencies` though.

For `@remix-run/server-runtime` it should be in `dependencies` as we're exporting the types.

chaance: May as well just inline it. No reason to worry about carrying over the dependency for this IMO.

MichaelDeBoey (approved):

PR #2945, opened 2022-04-20
ci: add content-type to curl, rename `GITHUB_REF` to `VERSION`

mcansh:
I'll need to cherry-pick this over to `dev` as well due to changing `scripts/release`


WaleedxxxxjJz (commented):
Good

PR #2951, opened 2022-04-21
Update node version required to generate project

michaelkoelewijn91:
The docs mention the project can be generated using node v14.
Running ```npx create-remix@latest``` using v14 resolves in a ```.replaceAll function is not defined```.
This function has been introduced as of node v15  (see https://nodejs.org/en/blog/release/v15.0.0/#v8-8-6-35415).
I have updated the docs in this PR to include the correct minimum version.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -334,3 +334,4 @@
 - zachdtaylor
 - zainfathoni
 - zhe
+- michaelkoelewijn91
--- a/docs/tutorials/blog.md
+++ b/docs/tutorials/blog.md
@@ -19,7 +19,7 @@ Click this button to create a [Gitpod](https://gitpod.io) workspace with the pro
 
 If you want to follow this tutorial locally on your own computer, it is important for you to have these things installed:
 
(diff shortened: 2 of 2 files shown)


kiliman:
It would probably be better to simply replace the `.replaceAll` call with `.replace(/pattern/g, replacement)` than to force everyone to update their node version.

sergiodxa:
> It would probably be better to simply replace the `.replaceAll` call with `.replace(/pattern/g, replacement)` than to force everyone to update their node version.

Why I agree, specially because it's not used that much, I think Remix should think how they plan to support Node versions.

Is Remix supporting the current LTS or greater?
Is Remix supporting all currently maintained version of Node? In this case it should support v14 until next year, and v12 should be supported until the end of the current month.
Is Remix supporting a version based on some vendor? Like maybe some service only uses v14 so it can't upgrade to greater versions.
If a version of Node is not LTS and a new current was released, should Remix still support it?

MichaelDeBoey:
Closing this as #2867 has been merged and we're Node v14 compatible again

PR #2961, opened 2022-04-22
chore(remix-dev): resolve `extends` in tsconfig when applying defaults

mcansh:
follow up to #2786

the merging worked prior to 02b60f5 in the PR due to parsing the config before writing the defaults, but now it happens after that due to the tsconfig loader being there temporarily. the writing of _just_ the missing options to the resolved tsconfig is new behavior

### DIFF
--- a/integration/tsconfig-test.ts
+++ b/integration/tsconfig-test.ts
@@ -1,0 +1,94 @@
+import { test, expect } from "@playwright/test";
+import fse from "fs-extra";
+import path from "path";
+import JSON5 from "json5";
+
+import { createFixture, json } from "./helpers/create-fixture";
+
+const DEFAULT_CONFIG = {
+  include: ["remix.env.d.ts", "**/*.ts", "**/*.tsx"],
+  compilerOptions: {
+    allowJs: true,
+    esModuleInterop: true,
+    forceConsistentCasingInFileNames: true,
+    isolatedModules: true,
+    jsx: "react-jsx",
+    lib: ["DOM", "DOM.Iterable", "ES2019"],
+    moduleResolution: "node",
+    noEmit: true,
+    resolveJsonModule: true,
+    strict: true,
+    target: "es2019",
+    baseUrl: ".",
+    paths: {
+      "~/*": ["./app/*"],
+    },
+  },
+};
+
+// besides baseUrl, due to an upstream issue in tsconfig-paths
+// https://github.com/dividab/tsconfig-paths/pull/208
+test("should output default tsconfig if file is empty", async () => {
+  let fixture = await createFixture({
+    files: {
+      "tsconfig.json": json({ compilerOptions: { baseUrl: "." } }),
+    },
+  });
+
+  let content = JSON5.parse(
+    await fse.readFile(path.join(fixture.projectDir, "tsconfig.json"), "utf8")
+  );
+  expect(content).toEqual(DEFAULT_CONFIG);
+});
+
+test("should add/update mandatory config", async () => {
+  let fixture = await createFixture({
+    files: {
+      "tsconfig.json": json({
+        include: ["remix.env.d.ts", "**/*.ts", "**/*.tsx"],
+        compilerOptions: {
+          allowJs: true,
+          baseUrl: ".",
+          esModuleInterop: true,
+          forceConsistentCasingInFileNames: true,
+          isolatedModules: false, // true is required by esbuild
+          jsx: "react-jsx",
+          lib: ["DOM", "DOM.Iterable", "ES2019"],
+          // moduleResolution: "node", // this is required by esbuild
+          noEmit: true,
+          resolveJsonModule: true,
+          strict: true,
+          target: "es2019",
+        },
+      }),
+    },
+  });
+
+  let content = JSON5.parse(
+    await fse.readFile(path.join(fixture.projectDir, "tsconfig.json"), "utf8")
+  );
+  expect(content).toEqual(DEFAULT_CONFIG);
+});
+
+test("shouldn't change suggested config if set", async () => {
+  let config = {
+    ...DEFAULT_CONFIG,
+    compilerOptions: {
+      ...DEFAULT_CONFIG.compilerOptions,
+      strict: false,
+    },
+  };
+
+  let fixture = await createFixture({
+    files: {
+      "tsconfig.json": json(config),
+    },
+  });
+
+  let content = JSON5.parse(
+    await fse.readFile(path.join(fixture.projectDir, "tsconfig.json"), "utf8")
+  );
+  expect(content).toEqual(config);
+});
+
+// TODO: add test for nested tsconfig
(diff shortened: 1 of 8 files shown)


integration/tsconfig-test.ts
```diff
@@ -0,0 +1,136 @@
+import { test, expect } from "@playwright/test";
```
kentcdodds, line 1: Love this test. Do you think you could help me understand which bits and pieces are important in the assertions via something like this: https://kentcdodds.com/blog/aha-testing

kentcdodds (changes requested):
I think this looks good to me. I just have one ask. Thanks!

packages/remix-dev/compiler/utils/tsconfig/write-config-defaults.ts
```diff
@@ -37,15 +38,33 @@ function objectKeys<Type extends object>(value: Type): Array<ObjectKeys<Type>> {
...
+  // this will be the user's actual tsconfig file
+  let configContents = fse.readFileSync(configPath, "utf8");
+
+  let config: TsConfigJson | undefined;
+  try {
+    config = JSON5.parse(configContents);
+  } catch (error: unknown) {}
+
+  if (!fullConfig || !config) {
+    // how did we get here?
```
kentcdodds, line 55: Maybe add a `console.warn` saying: "This should never happen, please open an issue with a reproduction." or something.

integration/tsconfig-test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+test("should add/update mandatory config", async () => {
+  let fixture = await createFixture({
+    files: {
+      "tsconfig.json": json({
+        ...DEFAULT_CONFIG,
+        compilerOptions: {
+          ...DEFAULT_CONFIG.compilerOptions,
+          isolatedModules: false, // true is required by esbuild
+          // moduleResolution: "node", // this is required by esbuild
```
kentcdodds, line 56: Should this commented-out code be here?

integration/tsconfig-test.ts
```diff
@@ -0,0 +1,125 @@
...
+      lib: ["DOM", "DOM.Iterable", "ES2019"],
+      moduleResolution: "node",
+      noEmit: true,
+      resolveJsonModule: true,
+      strict: true,
+      target: "es2019",
+      paths: {
+        "~/*": ["./app/*"],
+      },
+    },
```
kentcdodds, line 123: Where do all these values come from? Is there a way to indicate where they come from by the test code itself?

kentcdodds (changes requested):
Thanks! More thoughts.

integration/tsconfig-test.ts
mcansh: i actually wanted to remove the compilerOption alltogether, but that doesn't happen due to the spread, will fix

integration/tsconfig-test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+import { createFixture, json } from "./helpers/create-fixture";
+
+async function getTsConfig(projectDir: string) {
+  let tsconfigPath = path.join(projectDir, "tsconfig.json");
+  let config = await fse.readFile(tsconfigPath, "utf8");
+  return JSON5.parse(config);
+}
+
+// this is the default tsconfig.json that is shipped with `create-remix` templates
```
mcansh, line 14: after https://github.com/remix-run/remix/pull/3013 is merged we can do this instead 


```ts
const DEFAULT_CONFIG_CONTENT = fse.readFileSync(
  path.join(process.cwd(), "templates/remix/tsconfig.json"),
  "utf8"
);

const DEFAULT_CONFIG = JSON5.parse(DEFAULT_CONFIG_CONTENT);
```

integration/tsconfig-test.ts
```diff
@@ -0,0 +1,125 @@
...
+    noEmit: true,
+    resolveJsonModule: true,
+    strict: true,
+    target: "ES2019",
+    baseUrl: ".",
+    paths: {
+      "~/*": ["./app/*"],
+    },
+  },
+};
```
mcansh, line 34: after https://github.com/remix-run/remix/pull/3013 is merged we can do this instead 


```ts
const DEFAULT_CONFIG_CONTENT = fse.readFileSync(
  path.join(process.cwd(), "templates/remix/tsconfig.json"),
  "utf8"
);

const DEFAULT_CONFIG = JSON5.parse(DEFAULT_CONFIG_CONTENT);
```

kentcdodds: I'd rather not use stuff in `templates`. It's unclear to future maintainers that those things are used in tests and false breakages could lead to confusion. I'm fine with the object here.

integration/tsconfig-test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  let fixture = await createFixture({
+    files: {
+      "tsconfig.json": json(config),
+      "tsconfig.base.json": json(baseConfig),
+    },
+  });
+
+  let tsconfig = await getTsConfig(fixture.projectDir);
+  // our base config only sets a few options, so our local config should fill in the missing ones
```
kentcdodds, line 105: It should be obvious from the code which were "the missing ones" from the way this code is written. Is this just basically "the config the project had before plus some filled in properties?" If that's the case, then I should be able to tell just by reading the test code.

mcansh: in this case its the `DEFAULT_CONFIG` minus https://github.com/remix-run/remix/pull/2961/files/22e522bf19dfe9bfc60e9a365fa9223491c302de#diff-1d51fc796c50fdfde6dfe89b8fb099b6c8b25edba2b4e7da0e861a8ea8e35e1eR91-R94 as that was defined in a different tsconfig which was extended – so we could spread the default config in and delete those 2 options by hand instead of a big object

kentcdodds: I see. Thanks for explaining that and updating the code. So if people have `allowJs` and `baseUrl` set, then we remove those for them? When do we make this change again? I'm still unsure why we want to do that.

mcansh: so in this test, we're verifying that their tsconfig and any they're extending are merged properly, we're not removing any properties. 

this will run during the initial build of remix dev/watch and remix build

kentcdodds: Oh I see. So this test is just saying: "If their main config doesn't have these properties, we also check any extends to make sure those properties are in there so we don't have to add those properties manually"?

mcansh: yup! if the extend has it, we can rely on that

kentcdodds: > this will run during the initial build of remix dev/watch and remix build

Ok, where are we running the build in this test to make sure the code under test is actually being run?

Also, I guess I'm a bit confused still... So first we create a `tsconfig.json` with the following contents:

```json
{
  extends: "./tsconfig.base.json",
};
```

Then we create the fixture.

Then we get the file and make sure it's contents are equal to the default config. But they shouldn't be right? Because we didn't modify them.

kentcdodds: We talked offline and I get now. This is good 👍

kentcdodds (approved):
Super! Thanks 👍

Issue #2836, opened 2022-04-17
remix 1.4.0 breaks integration with turborepo & workspaces libraries


nareshbhatia:
### What version of Remix are you using?

1.4.0

### Steps to Reproduce

Here's my sample repo which is working with remix 1.3.5: https://github.com/nareshbhatia/custom-react-stack.

Build & run to make sure everything is working ok:

```bash
npm install
npm run dev
```

Point your browser to the remix app at http://localhost:3002/.

Now change remix dependencies to 1.4.0 in `apps/movie-magic-remix/package.json`. Do a clean build and run.

```bash
npm run clean
rm package-lock.json
npm install
npm run dev
```

### Expected Behavior

Build does not produce any errors and application runs.

### Actual Behavior

Build produces the following error and application does not run. Note that `packages/ui-lib` is a library package in the monorepo. The remix server is not able to resolve its path.

```
movie-magic-remix:dev: ✘ [ERROR] [plugin server-bare-modules] Couldn't resolve the base path of "ui-lib". Searched inside the resolved main file path "/Users/naresh/projects/custom-react-stack/packages/ui-lib/dist/index.js" using "node_modules/ui-lib"
movie-magic-remix:dev:
movie-magic-remix:dev:     ../../node_modules/@remix-run/dev/compiler/plugins/serverBareModulesPlugin.js:181:10:
movie-magic-remix:dev:       181 │     throw new Error(`Couldn't resolve the base path of "${package...
movie-magic-remix:dev:           ╵           ^
movie-magic-remix:dev:
movie-magic-remix:dev:     at resolveModuleBasePath (/Users/naresh/projects/custom-react-stack/node_modules/@remix-run/dev/compiler/plugins/serverBareModulesPlugin.js:181:11)
movie-magic-remix:dev:     at warnOnceIfEsmOnlyPackage (/Users/naresh/projects/custom-react-stack/node_modules/@remix-run/dev/compiler/plugins/serverBareModulesPlugin.js:126:20)
movie-magic-remix:dev:     at /Users/naresh/projects/custom-react-stack/node_modules/@remix-run/dev/compiler/plugins/serverBareModulesPlugin.js:96:11
movie-magic-remix:dev:     at callback (/Users/naresh/projects/custom-react-stack/node_modules/esbuild/lib/main.js:920:34)
movie-magic-remix:dev:     at processTicksAndRejections (node:internal/process/task_queues:96:5)
[...]

PR #2965, opened 2022-04-22
chore(remix-dev): change `resolveModuleBasePath` to use a lookup for a package.json

mcansh:


MichaelDeBoey (approved):

mcansh:
sounds like packages are resolving fine in monorepos now using [1.4.3](https://github.com/remix-run/remix/issues/2836#issuecomment-1111726881), leaving open for now

mcansh:
> @mcansh Since the problem seems to be fixed, do we still want to implement this?

i think we're good on skipping this :)

PR #2974, opened 2022-04-25
feat: enabled node-esm target

jacob-ebey:
chore: added express-esm template

Closes: #

- [ ] Docs
- [ ] Tests


MichaelDeBoey:
@jacob-ebey It seems like we have some conflicts to resolve here

Issue #2983, opened 2022-04-26
Dev mode is not re building / reloading in monorepo context (nx)


npirotte:
### What version of Remix are you using?

1.4.1

### Steps to Reproduce

* Create a monorepo with NX (I assume it's the same with turbo repo)
* Create a remix application
* Create a library
* Import something from the library into the application
* Start the application in dev mode
* Save some change in the library file.

Alternative that might go quicker:
* Fork https://github.com/npirotte/hello-visx
* run `npm i`
* run `nx dev demo`
* do some changes in `packages/bar-chart/src/lib/bar-chart.tsx`

### Expected Behavior

App rebuild and reload.

### Actual Behavior

Nothing happens, I have to do some change in the app itself to have it reloading.

npirotte:
After more testing, I understand that remix listen to events on the `appDirectory`, while my expectation are that it listen to `appDirectory` + dependencies changes (in monorepo context, most of the files will be out of appDirectory).

npirotte:
As an easy and simple solution, I can propose to add an optional configuration in the remix config:

```ts
export interface AppConfig {
  /**
  * List of extra paths or globs to watch changes on in addition to the main app directory.
  * relative to `remix.config.js`. Defaults to `[]`.
  */
  watch?: string | string[]
}
```

This would the be merge to the current list of watched directories:

```
// compiler.ts
let extraWatch = Array.isArray(config.watch) ? config.watch : [config.watch];
let toWatch = [config.appDirectory, ...extraWatch];
if (config.serverEntryPoint) {
  toWatch.push(config.serverEntryPoint);
}
```

npirotte:
Already created an example implementation here: https://github.com/remix-run/remix/pull/2984

fknop:
Having the same issue with NX. Is there an eventual workaround until the fix lands?

npirotte:
Didn't found anything yet, currently I simply keep the root.tsx file opened, and I hit save on that one when I want to rebuild the app.

jdnichollsc:
@fknop maybe using [patch-package](https://www.npmjs.com/package/patch-package) with the changes created by @npirotte in that PR 🤔

MichaelDeBoey:
Closed by #3188

PR #2985, opened 2022-04-26
feat(remix-dev): add `extraPathsToWatch` config option

npirotte:
This adds the possibility to watch extra directories/files/globs in dev mode.

This should make integration with mono repos lix NX easier.

Will spend more time documenting and testing if necessary and if feature is approved.

Closes: #2983

- [x] Docs
- [ ] Tests

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -331,3 +331,4 @@
 - zachdtaylor
 - zainfathoni
 - zhe
+- npirotte
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -215,7 +215,7 @@ export async function watch(
     if (onRebuildFinish) onRebuildFinish();
   }, 100);
 
-  let toWatch = [config.appDirectory];
+  let toWatch = [config.appDirectory, ...config.extraPathsToWatch];
   if (config.serverEntryPoint) {
     toWatch.push(config.serverEntryPoint);
   }
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -40,6 +40,12 @@ export interface AppConfig {
   appDirectory?: string;
 
   /**
+  * List of extra paths or globs to watch changes on in addition to the main app directory.
+  * relative to `remix.config.js`. Defaults to `[]`.
+  */
+  extraPathsToWatch?: string | string[];
+
+  /**
    * The path to a directory Remix can use for caching things in development,
    * relative to `remix.config.js`. Defaults to `".cache"`.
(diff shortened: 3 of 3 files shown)


npirotte:
Yes, I don't mind both option will fulfil my needs :)

philipjfulcher:
Chiming in from the Nx side, I think the function might work better for us as well. That way, we could use the project graph to determine what other directories need to be watched instead of having a static list of directories we'd need to update.

I'm curious how often this config is re-evaluated. If a dev imports from a new lib that's not yet added to the watch, are we able to supply new directories to watch over time? Or do you need to re-start the `dev` script entirely to get the updated config?

philipjfulcher:
Thinking through this a little more, I have a specific, rather than a hypothetical example of where the function would have advantages over the array.

We can look at how `@nrwl/react-native` does something similar working with `metro-resolver` in [with-nx-metro.ts](https://github.com/nrwl/nx/blob/master/packages/react-native/plugins/with-nx-metro.ts)


```typescript
let watchFolders = config.watchFolders || [];
  watchFolders = watchFolders.concat([
    join(workspaceRoot, 'node_modules'),
    join(workspaceRoot, workspaceLayout().libsDir),
    join(workspaceRoot, '.storybook'),
  ]);
  if (opts.watchFolders?.length) {
    watchFolders = watchFolders.concat(opts.watchFolders);
  }

  watchFolders = watchFolders.filter((folder) => existsSync(folder));
```

To provide all of the directories to watch, we use some utility functions from Nx to determine where the libs folder is actually located. We do this same sort of thing with a static array, it's just that we could introduce issues if the libs location ever changes or sin't set correctly to begin with. We also include changes in `node_modules` and `.storybook` directories to rebuild on those changes.

npirotte:
Make sense, to me, maybe @christophertrudel can you create a PR with your commit? Then I can close this one. Just make sure to link https://github.com/remix-run/remix/issues/2983 to it :)

Would be nice to have some remix developer feedback on that too :)

guillaumebailleul:
This solution works for my project. Hope this one or this https://github.com/christophertrudel/remix/commit/763b6acdabda6a855c3776df698159cba95f68d4 would be merged soon.

christophertrudel:
> Make sense, to me, maybe @christophertrudel can you create a PR with your commit? Then I can close this one. Just make sure to link #2983 to it :)
> 
> Would be nice to have some remix developer feedback on that too :)

https://github.com/remix-run/remix/pull/3188

MichaelDeBoey:
Closed by #3188

PR #2994, opened 2022-04-27
Bug: getDependenciesToBundle fails on pkg missing main

Bushuo:
- [ ] Docs
- [x] Tests

Testing Strategy:
I tested this by altering the bug report test to include a package, as it is done in the compiler integration tests.

if the `package.json` is altered to include a `"main": "submodule.js"` instead of the `"exports": ...` it works.


Bushuo:
Hi Remix Team,
first of all, I really enjoy using remix as a developer and would be glad to provide a contribution to such an awesome product :rainbow:

I would also like to point to this discussion, where I made a comment about other issues regarding this function.
#2594

PR #3001, opened 2022-04-27
test: increase timeout of test for windows

kentcdodds:
- [x] Docs
- [x] Tests

Testing Strategy: I ran the tests locally and they run fast on my machine. I'm guessing the windows powershell thing is really slow or something, so my testing strategy is CI here to see whether the test times out in CI on windows.


MichaelDeBoey (approved):

kentcdodds:
Looks like it only took 5 seconds on CI 🤷‍♂️ Sometimes it times out after 10 seconds so I think we'll leave it where it's at. This isn't a very important thing to be very fast anyway.

PR #3018, opened 2022-04-29
docs: React Router migration docs

chaance:
This PR introduces a new comprehensive docs page for migrating a CRA app to Remix.


docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+export function Logo() {
+  return <MyLogo />;
+}
+```
+
+If you want to use SVG files as React components, you'll need to first create the components and import them directly. [React SVGR](https://react-svgr.com/) is a great toolset that can help you generate these components from the [command line](https://react-svgr.com/docs/cli/) or in an [online playground](https://react-svgr.com/playground/) if you prefer to copy and paste.
+
+### Style imports
+
+CRA supports importing styles in your components in many ways. While this is common practice in the React ecosystem, it's not supported the same way in Remix for a few different reasons.
```
kentcdodds, line 57: I think this section could be expanded to list a few of the common issues people will come up against due to this and what they should do about it.

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+First, you'll need a few of our packages to build on Remix. Follow the instructions below, running all commands from the root of your CRA project.
+
+```shell
+npm install @remix-run/react @remix-run/node
+```
+
+Because CRA apps are client-rendered by default, you may not have a server yet. All Remix apps run on a server, so we'll install a few packages to create one quickly.
+
+```shell
+npm install @remix-run/express express@4 compression@1 morgan@1
```
kentcdodds, line 70: Is there a reason they can't just use remix-serve? That would simplify this guide considerably I think. If there's a change remix-serve needs to support CRA apps then maybe we should make that change.

kentcdodds (commented):
Only got to the `Setting up the server` part, but here's my feedback so far.

This will be enormously helpful for people!

docs/guides/migrating-cra.md
chaance: This guide was written to basically emulate our Express template, so I copied all of the scripts and dependencies directly from it. We could probably simplify it but I'd like to be as close to the template as possible, as we'll be linking out to it.

docs/guides/migrating-cra.md
chaance: Yeah I meant to include a link to the section below where we talk about styles in relation to `links` but I just missed it 😅

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+│   └── expenses.jsx
+│   └── index.jsx
+│   └── invoices.jsx
+│   └── $userId.jsx
+└── root.jsx
+```
+
+### The `root` route
+
+The root route (or the "root root" if you're Wes Bos) is responsible for providing the structure of the application. Its default export is a component that renders the full HTML tree.
```
kentcdodds, line 503: @wesbos, just want to double-check with you that you're ok with this joke 😅

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+        <div id="root"></div>
+      </body>
+    </html>
+  );
+}
+```
+
+Notice that we don't need to use `%PUBLIC_URL%` as our server serves relative paths from the `public` directory.
+
+We can also get rid of the `noscript` tag because we're server rendering now, which means users who disable JavaScript will still be able to see our app (even if they can't interact with it in exactly the same way).
```
kentcdodds, line 579:
```suggestion
We can also get rid of the `noscript` tag because we're server rendering now, which means users who disable JavaScript will still be able to see our app and thanks to [progressive enahancement](../pages/philosophy#progressive-enhancement) most of the app will still work anyway.
```

I'm not sure if that URL is correct. I *think* it is.

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+      <footer>&copy; Remix Software</footer>
+    </div>
+  );
+}
+```
+
+While this is a convenient API, consider a few questions:
+
+- How do the styles actually end up on the page? Do you get a `<link />` or an inline `<style />` in the `<head>`?
+- If other components also import CSS, where do they end up in relation other component styles?
```
kentcdodds, line 650:
```suggestion
- If other components also import CSS, where do they end up in relation other component styles? This has important implications on how the styles are applied due to the cascading nature of CSS.
```

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+      </head>
+      <body>
+        <App />
+      </body>
+    </html>
+  );
+}
+```
+
+You'll notice on line 32 that we've rendered a `<Links />` component that replaced all of our individual `<link />` components. This is inconsequential if we only ever use links in the root route, but it's all child routes may export their own links that will also be rendered here. The `links` function can also return a [`PageLinkDescriptor` object](../api/conventions#pagelinkdescriptor) that allows you to prefetch the resources for a page the user is likely to navigate to.
```
kentcdodds, line 712:
```suggestion
You'll notice on line 32 that we've rendered a `<Links />` component that replaced all of our individual `<link />` components. This is inconsequential if we only ever use links in the root route, but all child routes may export their own links that will also be rendered here. The `links` function can also return a [`PageLinkDescriptor` object](../api/conventions#pagelinkdescriptor) that allows you to prefetch the resources for a page the user is likely to navigate to.
```

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+
+If you inject `<link />` tags into your page client-side in your existing route components, either directly or via an abstraction like [`react-helmet`](https://www.npmjs.com/package/react-helmet), you'll get to delete a lot of code and possibly a dependency or two!
```
kentcdodds, line 714:
```suggestion
If you currently inject `<link />` tags into your page client-side in your existing route components, either directly or via an abstraction like [`react-helmet`](https://www.npmjs.com/package/react-helmet), you can stop doing that and instead use the `links` export. You get to delete a lot of code and possibly a dependency or two!
```

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+
+### Route `<meta />` tags
+
+Similar to `links`, each route can also export a `meta` function that—you guessed it—returns a value responsible for rendering `<meta />` tags for that route. This is useful because each route often has its own
+
+The API is slightly different for `meta`. Instead of an array, it returns an object where the keys represent the meta `name` attribute (or `property` in the case of OpenGraph tags) and the value is the `content` attribute. The object can also accept a `title` property that renders a `<title />` component specifically for that route.
+
+Again—no more weird dances to get meta into your routes from deep in the component tree. Export them at the route level and let the server handle it. ✨
+
```
kentcdodds, line 723: Can we add a simple example here or a link to the docs for the `meta` component? Would probably be good to mention that meta accepts loader data so your meta tags can be data-driven. Would probably be good to move the `<title>` in the root.tsx to `meta` and explain that's a special case.

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+## Using the `.jsx` extension for React component files
+
+In CRA you can use JSX in either `.jsx` or `.js` files. You'll need to rename any components with JSX to `.jsx`.
```
kentcdodds, line 726:
```suggestion
In CRA you can use JSX in either `.jsx` or `.js` files. In Remix, you *must* use `.jsx` for all files that use JSX.
```

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,774 @@
...
+  "dev:node": "cross-env NODE_ENV=development nodemon ./server.js --watch ./server.js",
+  "dev:remix": "remix watch"
+}
+```
+
+These scripts are a bit more verbose, but that's because we don't abstract the processes. We give that power back to you. After all, you know how to run your app better than anyone on our team!
+
+## Final Thoughts
+
+While we've done our best to provide a comprehensive migration guide, it's important to note that we built Remix from the ground up with a few key principles that differ significantly from how many React apps are currently built. While your app will likely run at this point, as you dig through our docs and explore our APIs, we think you'll be able to drastically reduce the complexity of your code and improve the end user experience of your app. It might take a bit of time to get there, but you can eat that elephant one bite at a time. 🐘
```
kentcdodds, line 772:
```suggestion
While we've done our best to provide a comprehensive migration guide, it's important to note that we built Remix from the ground up with a few key principles that differ significantly from how many React apps are currently built. While your app will likely run at this point, as you dig through our docs and explore our APIs, we think you'll be able to drastically reduce the complexity of your code and improve the end user experience of your app. It might take a bit of time to get there, but you can eat that elephant one bite at a time. 🐘 In particular, after you've shipped this first upgrade to your app, you'll want to start moving your data fetching and mutations to [`loaders`](../api/conventions#loader) and [`actions`](../api/conventions#action). Get ready to delete a lot of code!
```

kentcdodds (commented):
Super 👍 Just a few other comments.

docs/guides/migrating-cra.md
chaance: @kentcdodds Discussed w/ @mjackson before I wrote this and he agreed the Express template would make for a better teaching opportunity. We could discuss again, just want to make sure we're all on the same page.

docs/guides/migrating-cra.md
chaance: Going to hold on this. @mjackson suggested that we should probably support JSX in `.js` files since esbuilt supports that directly now. May try to land that PR first!

chaance: Related PR is up: https://github.com/remix-run/remix/pull/3059

docs/guides/migrating-cra.md
wesbos: haha yep, I didn't realize I was known for the way I say Route :)

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,790 @@
...
+<docs-warning><strong>Important:</strong> be sure to delete the `index.html` from your `public` directory after you've created your root route. Keeping the file around will cause your server to send that HTML instead of your Remix app when accessing the `/` route.</docs-warning>
+
+### Route stylesheets
+
+At this point, you _might_ be able to run your app with no changes, but you probably import a stylesheet somewhere in your `App` component (and probably from others, I'm sure). Remember though—Remix does not handle CSS imports the same way CRA does, and we think for good reason.
+
+Assume you have a plain CSS import in your `App` component:
+
+```jsx filename=app.jsx lines=[6]
+import * as React from "react";
```
MichaelDeBoey, line 628:
```suggestion
```jsx filename=app.jsx lines=[5]
```

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,790 @@
...
+</docs-info>
+
+### Route `links` exports
+
+In Remix, stylesheets can only be loaded from route component files. Importing them does not do anything magical with your styles, rather it returns a URL that can be used to load the stylesheet as you see fit. You can render the stylesheet directly in your component or use our `links` export.
+
+Let's move our app's stylesheet and a few other assets to the `links` function in our root route:
+
+```jsx filename=root.jsx lines=[2,5,7-16,32]
+import * as React from "react";
```
MichaelDeBoey, line 676:
```suggestion
```jsx filename=root.jsx lines=[1,4,6-15,31]
```

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,790 @@
...
+      </head>
+      <body>
+        <App />
+      </body>
+    </html>
+  );
+}
+```
+
+You'll notice on line 32 that we've rendered a `<Links />` component that replaced all of our individual `<link />` components. This is inconsequential if we only ever use links in the root route, but all child routes may export their own links that will also be rendered here. The `links` function can also return a [`PageLinkDescriptor` object](../api/conventions#pagelinkdescriptor) that allows you to prefetch the resources for a page the user is likely to navigate to.
```
MichaelDeBoey, line 718:
```suggestion
You'll notice on line 31 that we've rendered a `<Links />` component that replaced all of our individual `<link />` components. This is inconsequential if we only ever use links in the root route, but all child routes may export their own links that will also be rendered here. The `links` function can also return a [`PageLinkDescriptor` object](../api/conventions#pagelinkdescriptor) that allows you to prefetch the resources for a page the user is likely to navigate to.
```

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,790 @@
...
+
+If you currently inject `<link />` tags into your page client-side in your existing route components, either directly or via an abstraction like [`react-helmet`](https://www.npmjs.com/package/react-helmet), you can stop doing that and instead use the `links` export. You get to delete a lot of code and possibly a dependency or two!
+
+### Route `meta` exports
+
+Similar to `links`, each route can also export a `meta` function that—you guessed it—returns a value responsible for rendering `<meta />` tags for that route. This is useful because each route often has its own
```
MichaelDeBoey, line 724: > This is useful because each route often has its own

Seems like this sentence is missing some words? 🤔

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,790 @@
...
+
+**Before:**
+
+```jsx bad
+import { Link, Outlet } from "react-router-dom";
+```
+
+**After:**
+
+```jsx
```
MichaelDeBoey, line 746:
```suggestion
```jsx good
```
This way it will show a green border & is consistent with the bad example (which shows with a red border)

MichaelDeBoey (changes requested):
Really comprehensive migration guide! 👀

Added some small fixes here and there.

kentcdodds:
I talked with @mjackson and he and I came up with a migration strategy we're really happy with that doesn't involve making your own express server. I need to test a few things out then I'll make a loom to show how it works.

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,790 @@
+---
+title: Migrating from Create React App
+description: Remix makes integrating MDX into your project a breeze with built in routes and "import" support.
+---
+
+# Migrating from Create React App with React Router
+
+Most apps bootstrapped with [Create React App](https://create-react-app.dev/) (CRA) already use [React Router](https://reactrouter.com/) for client-side routing. Because Remix is built on top of React Router, migration should be a relatively easy process.
+
+If you aren't already using React Router, we think there are several compelling reasons to reconsider! History management, dynamic path matching, nested routing, and much more. Take a look at the [React Router docs](https://reactrouter.com/docs/en/v6/getting-started/concepts) and see what all we have to offer.
```
machour, line 10:
```suggestion
If you aren't already using React Router, we think there are several compelling reasons to reconsider! History management, dynamic path matching, nested routing, and much more. Take a look at the [React Router docs](https://reactrouter.com/docs/en/v6/getting-started/concepts) and see all what we have to offer.
```

docs/guides/migrating-cra.md
```diff
@@ -0,0 +1,790 @@
...
+  createRequestHandler({
+    build: require(BUILD_DIR),
+    mode: process.env.NODE_ENV,
+  })
+);
+```
+
+We can improve the experience during development quite a bit here by purging the [require cache](https://nodejs.org/api/modules.html#caching) before each request. This is a good idea to avoid memory leaks and other issues that can arise from caching, especially when using live reload.
+
+We'll write a function called `purgeRequireCache` that will delete the cache for anything that requires our build output, and we'll call it cnoditionally when `NODE_ENV` is set to `development`.
```
machour, line 269:
```suggestion
We'll write a function called `purgeRequireCache` that will delete the cache for anything that requires our build output, and we'll call it conditionally when `NODE_ENV` is set to `development`.
```

machour (commented):
Awesome work @chaance !

chaance:
@kentcdodds I changed the title of the page to reflect that these instructions now cover any React Router app with a Webpack bundler instead of just CRA. Removed all of the custom server stuff and the tutorial now mirrors the migration repo you created.

docs/guides/migrating-react-router-app.md
```diff
@@ -0,0 +1,518 @@
...
+
+We also suggest renaming this directory to make it clear that this is your old code so that, eventually, you can delete it after migrating all of its contents. The beauty of this approach is that you don't have to do it all at once for your app to run as usual. In our demo project we name this directory `old-app`.
+
+Lastly, in your root `App` component (the one that would have been mounted to the `root` element), remove the `<BrowserRouter>` from React Router. Remix takes care of this for you without needing to render the provider directly.
+
+## Creating a catch-all route
+
+Remix needs routes beyond the root route to know what to render in `<Outlet />`. Fortunately you already render `<Route>` components in your app, and Remix can use those as you migrate to use our [routing conventions](./routing.md).
+
+To start, create a new directory in `app` called `routes`. In that directory, create a file called `$.jsx`. This is called a **catch-all route** and it will be useful to let your old app handle routes that you haven't moved into the `routes` directory yet.
```
kentcdodds, line 180:
```suggestion
To start, create a new directory in `app` called `routes`. In that directory, create a file called `$.jsx`. This is called [a **catch-all route**](https://remix.run/docs/en/v1/guides/routing#splats) and it will be useful to let your old app handle routes that you haven't moved into the `routes` directory yet.
```

docs/guides/migrating-react-router-app.md
```diff
@@ -0,0 +1,518 @@
...
+
+### A note about non-standard imports
+
+At this point, you _might_ be able to run your app with no changes. If you are using Create React App or a highly-configured Webpack app, you likely use `import` to include non-JavaScript modules like stylesheets and images.
+
+Remix does not support most non-standard imports, and we think for good reason. Below is a non-exhaustive list of some of the differences you'll encounter in Remix, and how to refactor as you migrate.
+
+#### Asset imports
+
+Many bundlers use plugins to allow importing various assets like images and fonts. These typically come into your component as string representing the filepath of the asset. This works the same way in Remix.
```
kentcdodds, line 317:
```suggestion
Many bundlers use plugins to allow importing various assets like images and fonts. These typically come into your component as string representing the filepath of the asset.
```

We say that after the code block.

kentcdodds (changes requested):
This is looking great! Just a few thoughts. Thanks!

docs/guides/migrating-react-router-app.md
chaance: This needs to be a relative path. Will update in a separate commit 👍

docs/guides/migrating-react-router-app.md
```diff
@@ -0,0 +1,518 @@
...
+
+**Note:** Remix does not currently support CSS processing directly. If you use preprocessors like Sass, Less, or PostCSS, you can run those as a separate process in development.
+
+We also do not yet support CSS Modules, as that requires compiler integration and current approaches are not aligned with our design philosophy. We are actively working on a solution and plan to have an API for CSS Modules very soon.
+
+</docs-info>
+
+### Route `links` exports
+
+In Remix, stylesheets can only be loaded from route component files. Importing them does not do anything magical with your styles, rather it returns a URL that can be used to load the stylesheet as you see fit. You can render the stylesheet directly in your component or use our `links` export.
```
chaance, line 427:
```suggestion
In Remix, stylesheets can only be loaded from route component files. Importing them does not do anything magical with your styles, rather it returns a URL that can be used to load the stylesheet as you see fit. You can render the stylesheet directly in your component or use our [`links` export](../api/conventions#links).
```

docs/guides/migrating-react-router-app.md
```diff
@@ -0,0 +1,518 @@
...
+
+<docs-info>
+
+**Note:** Remix does not currently support CSS processing directly. If you use preprocessors like Sass, Less, or PostCSS, you can run those as a separate process in development.
+
+We also do not yet support CSS Modules, as that requires compiler integration and current approaches are not aligned with our design philosophy. We are actively working on a solution and plan to have an API for CSS Modules very soon.
+
+</docs-info>
+
+### Route `links` exports
```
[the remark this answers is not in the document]
chaance: I'm going to add this link to the first paragraph in the body just for consistency's sake.

docs/guides/migrating-react-router-app.md
```diff
@@ -0,0 +1,518 @@
...
+  );
+}
+```
+
+Once you create this file, you can delete the `<Route>` component from your `App`. After all of your routes have been migrated you can delete `<Routes>` and ultimately all of the code in `old-app`.
+
+## Gotchas and next steps
+
+At this point you might be able to say you are done with the initial migration. Congrats! However Remix does things a bit differently than your typical React app. If it didn't, why would we have bothered building it in the first place? 😅
+
```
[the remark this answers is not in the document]
chaance: Great suggestion, added a few sections here 👍

docs/guides/migrating-react-router-app.md
```diff
@@ -0,0 +1,621 @@
...
+If you are lazy-loading components with [`React.lazy`](https://reactjs.org/docs/code-splitting.html#reactlazy) and [`React.Suspense`](https://reactjs.org/docs/react-api.html#reactsuspense), you may run into issues depending on the version of React you are using. Until React 18, this would not work on the server as `React.Suspense` was originally implemented as a browser-only feature.
+
+If you are using React 17, you have a few options:
+
+- Upgrade to React 18
+- Use the [client-only approach](#client-only-components) outlined above
+- Use an alternative lazy-loading solution such as [Loadable Components](https://loadable-components.com/docs/loadable-vs-react-lazy/)
+- Remove `React.lazy` and `React.Suspense` altogether
+
```
kentcdodds, line 328:
```suggestion

Keep in mind that Remix automatically handles code-splitting for all your routes that it manages, so as you move things into the `routes` directory you should rarely if ever need to use `React.lazy` manually.
```

PR #3052, opened 2022-05-01
Reproduce a bug when an index route is next to a pathless directory

silvenon:
In the following route structure:

```
app/routes
└── nested
    ├── __pathless
    │   ├── bar.jsx
    │   └── foo.jsx
    ├── __pathless.jsx
    └── index.jsx
```

Remix incorrectly figures out what to show in the `/nested` route. Instead of showing `nested/index.jsx`, it shows an empty `nested/__pathless` route, i.e. the contents of the `nested/__pathless.jsx` layout with nothing in the `<Outlet />`. The misinterpretation is broken down in https://github.com/remix-run/remix/discussions/3014#discussioncomment-2666749.

This only happens for routes nested at least one level deep, that's why the `nested` directory is here.

Reproduces: https://github.com/remix-run/remix/discussions/3014, but it's just a discussion, should I create a bug report as well?

- [ ] Docs
- [x] Tests

Testing Strategy: `npm run bug-report-test`

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -47,28 +47,30 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/index.jsx": js`
-        import { json } from "@remix-run/node";
-        import { useLoaderData, Link } from "@remix-run/react";
-
-        export function loader() {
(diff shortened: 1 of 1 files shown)


brophdawg11:
Thanks for the bug report @silvenon!  I'm going to close this out as I've copied the bug over into https://github.com/remix-run/remix/pull/3986

PR #3059, opened 2022-05-02
fix(remix-dev): support JSX in `.js` route files

chaance:
We currently support JSX syntax in `.js` files *except* for route modules. This PR should fix that. 

- [x] Docs
- [x] Tests

### DIFF
--- a/integration/js-routes-test.ts
+++ b/integration/js-routes-test.ts
@@ -1,0 +1,42 @@
+import { test } from "@playwright/test";
+
+import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
+import type { AppFixture } from "./helpers/create-fixture";
+import { PlaywrightFixture } from "./helpers/playwright-fixture";
+
+test.describe(".js route files", () => {
+  let appFixture: AppFixture;
+
+  test.beforeAll(async () => {
+    appFixture = await createAppFixture(
+      await createFixture({
+        files: {
+          "app/routes/js.js": js`
+            export default () => <div data-testid="route-js">Rendered with .js ext</div>;
+          `,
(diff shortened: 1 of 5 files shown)


rollup.config.js
```diff
@@ -887,9 +887,12 @@ function copyToPlaygrounds() {
...
         // tickle live reload by touching the server entry
-        let serverEntry = ["entry.server.tsx", "entry.server.jsx"].find(
-          (entryPath) =>
-            fse.existsSync(path.join(playgroundDir, "app", entryPath))
+        let serverEntry = [
+          "entry.server.tsx",
+          "entry.server.js",
+          "entry.server.jsx",
+        ].find((entryPath) =>
+          fse.existsSync(path.join(playgroundDir, "app", entryPath))
```
MichaelDeBoey, line 895: You could simplify this to

```suggestion
        let serverEntry = ["tsx", "js", "jsx"].find(
          (entryPathExtension) =>
            fse.existsSync(path.join(playgroundDir, "app", `entry.server.${entryPathExtension}`))
```

MichaelDeBoey (dismissed):

integration/js-routes-test.ts
```diff
@@ -0,0 +1,42 @@
+import { test } from "@playwright/test";
```
chaance, line 1: I went to add a unit test to our build in `remix-dev` but it looks like that test is completely skipped and relies on the old Gists app that no longer exists. I tried to adapt it with a new fixture but couldn't get it working, so I went with this for now. We should probably fix that test file though.

MichaelDeBoey (approved):

mcansh (approved):

Issue #83, opened 2021-01-19
Clear app-build when remix build is run


NejcZdovc:
When I rename one route old route was still in `app-build`. Because of that when I did `vercel` old route was pushed to the server and vercel returned error that that old route was not in manifest. Probably what we need is that we clear that folder on every build so that old files are removed.

Aprillion:
Workaround I use in my project (Cloudflare Workers) until this is fixed: https://github.com/Aprillion/stampy-ui/commit/ca95346d65090b4558587ba955b54c8ce602963a


```diff
package.json
-"build": "cross-env NODE_ENV=production remix build",
+"build": "rimraf public/build && cross-env NODE_ENV=production remix build",
```

andrew-secret:
@Aprillion  Maybe you can help me. I'm trying to set up a remix project with Cloudflare workers. Each time I spin up the dev server with Miniflare I always see that my build directory is full of hashed files which are generated on the fly on each file change. Is this related?

brophdawg11:
This should be fixed with #3061, which will be in next week's release 👍

PR #3061, opened 2022-05-02
fix: clear build directory before new builds

brophdawg11:
This clears the `public/build` directory before the build when running `remix build`, to avoid stale artifacts from persisting on subsequent builds/deploys

Closes: #83

Testing Strategy: Not a great way to write a unit test for this, but it can be tested locally via:

* `npm run build`
* Inspect `public/build/routes/` for a given route JS file
* Rename the route TS/JS file
* `npm run build`
* Inspect `public/build/routes/` and confirm the old route file is no longer present and the new route exists

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -144,6 +144,7 @@ export async function build(
 
   let start = Date.now();
   let config = await readConfig(remixRoot);
+  fse.emptyDirSync(config.assetsBuildDirectory);
   await compiler.build(config, {
     mode: mode,
     sourcemap,


packages/remix-dev/cli/commands.ts
```diff
@@ -144,6 +144,7 @@ export async function build(
 
   let start = Date.now();
   let config = await readConfig(remixRoot);
+  fse.emptyDirSync(config.assetsBuildDirectory);
```
pcattori, line 147: Do we need to do the same thing for `watch` command too?

brophdawg11: I left out `watch` since the issue at hand here is related for not deploying stale assets on new builds.  I'm unsure if folks run into the same issue when running watch.  Assuming the watcher only does incremental re-builds, I'm not sure we could safely delete everything?

kentcdodds (approved):
Looks great to me 👍

PR #3064, opened 2022-05-02
fix: provide clear error messages for create-remix errors

brophdawg11:
This PR clears up some error messaging when we cannot validate `--template` options.  We had some nested `try`/`catch`'s where specific errors were getting swallowed by an outer `catch` with a more generic error.

Before - with a non-existent github template that would return a 404:
```sh
~> npx create-remix@latest --template brophdawg11/junk
🚨 There was a problem verifying the template. Please ensure you are connected to the internet and try again later.
```

After (stack trace is only because of the direct `node` execution):
```sh
~> node ../remix/build/node_modules/@remix-run/dev/cli create --template brophdawg11/junk
Error: 🚨 The template could not be verified. Please double check that the template is a valid GitHub repository and try again.
    at Object.validateTemplate (/Users/matt/me/remix/repos/remix/build/node_modules/@remix-run/dev/cli/create.js:492:19)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.run (/Users/matt/me/remix/repos/remix/build/node_modules/@remix-run/dev/cli/run.js:274:11)
```

- [x] Tests

Testing Strategy:
Added unit tests against specific `msw` mocks to force various HTTP errors (400, 403, 404)

### DIFF
--- a/packages/remix-dev/__tests__/create-test.ts
+++ b/packages/remix-dev/__tests__/create-test.ts
@@ -542,6 +542,88 @@ describe("the create command", () => {
     expect(output).toContain(getOptOutOfInstallMessage("pnpm exec remix init"));
     process.env.npm_user_agent = originalUserAgent;
   });
+
+  describe("errors", () => {
+    it("identifies when a github repo is not accessible (403)", async () => {
+      let projectDir = await getProjectDir("repo");
+      await expect(() =>
+        run([
+          "create",
+          projectDir,
+          "--template",
+          "error-username/403",
+          "--no-install",
+          "--typescript",
+        ])
+      ).rejects.toMatchInlineSnapshot(
+        `[Error: 🚨 The template could not be verified because you do not have access to the repository. Please double check the access rights of this repo and try again.]`
+      );
+    });
+
+    it("identifies when a github repo does not exist (404)", async () => {
+      let projectDir = await getProjectDir("repo");
+      await expect(() =>
+        run([
+          "create",
+          projectDir,
+          "--template",
+          "error-username/404",
+          "--no-install",
+          "--typescript",
+        ])
+      ).rejects.toMatchInlineSnapshot(
+        `[Error: 🚨 The template could not be verified. Please double check that the template is a valid GitHub repository and try again.]`
+      );
+    });
+
+    it("identifies when something unknown goes wrong with the repo request (4xx)", async () => {
+      let projectDir = await getProjectDir("repo");
+      await expect(() =>
+        run([
+          "create",
+          projectDir,
+          "--template",
+          "error-username/400",
+          "--no-install",
+          "--typescript",
+        ])
+      ).rejects.toMatchInlineSnapshot(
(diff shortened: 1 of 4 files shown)


packages/remix-dev/__tests__/create-test.ts
```diff
@@ -542,6 +542,88 @@ describe("the create command", () => {
...
+
+  describe("errors", () => {
+    it("identifies when a github repo is not accessible (403)", async () => {
+      let projectDir = await getProjectDir("repo");
+      await expect(() =>
+        run([
+          "create",
+          projectDir,
+          "--template",
+          "error-username/403",
```
brophdawg11, line 554: Added an `msw` mock to pick up on this pattern and error with the indicated status code

packages/remix-dev/__tests__/create-test.ts
```diff
@@ -542,6 +542,88 @@ describe("the create command", () => {
...
+    });
+
+    it("identifies when a remote tarball does not exist (404)", async () => {
+      let projectDir = await getProjectDir("remote-tarball");
+      await expect(() =>
+        run([
+          "create",
+          projectDir,
+          "--template",
+          "https://example.com/error/404/remix-stack.tar.gz",
```
brophdawg11, line 602: Same here for remote tarballs

packages/remix-dev/cli/create.ts
```diff
@@ -502,69 +502,73 @@ export async function validateTemplate(input: string) {
...
-            throw Error(
-              "🚨 The template file could not be verified. The server returned " +
-                `a response with a ${response.status} status. Please double ` +
-                "check the URL and try again."
-            );
-        }
-      } catch (err) {
-        spinner.stop();
+        response = await fetch(input, { method: "HEAD" });
+      } catch (_) {
```
brophdawg11, line 508: This is best viewed with whitespace hidden, but it just lifts the outer generic `catch` up to wrap only the `fetch` call.  This allows the thrown status-specific errors to propagate outwards

packages/remix-dev/cli/create.ts
```diff
@@ -576,34 +580,36 @@ export async function validateTemplate(input: string) {
       }
       let typeDir = templateType + "s";
       let templateUrl = `https://github.com/remix-run/remix/tree/main/${typeDir}/${name}`;
+      let response;
```
brophdawg11, line 583: I had some trouble coming up with a way to test the failure scenario for in-repo example/template since they're already validated against known-good values.  Any ideas?  Or OK with the current tests?

packages/remix-dev/__tests__/create-test.ts
kentcdodds: Brilliant

packages/remix-dev/cli/create.ts
kentcdodds: I think we're ok with current tests 👍

kentcdodds (approved):
Love what I see here 👍 Thanks!

PR #3071, opened 2022-05-03
chore: bring `playground` in line with `blues-stack`

MichaelDeBoey:


kentcdodds (approved):
Good call 👍 thanks!

kentcdodds (commented):
Looks like a linting issue. I hate that rule

PR #3073, opened 2022-05-03
test(replace-remix-imports): use cross-platform `grep` to avoid conditional `expect`

pcattori:
fixes lint warnings about conditional `expect`

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,6 @@
     "jsonfile": "^6.0.1",
     "meow": "^7.1.1",
     "npm-run-all": "^4.1.5",
-    "node-powershell": "^5.0.1",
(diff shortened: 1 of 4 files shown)


MichaelDeBoey (changes requested):
This means we can remove the `node-powershell` dependency as well, as it's only used in this test

pcattori:
> This means we can remove the `node-powershell` dependency as well, as it's only used in this test

`node-powershell` isn't listed as a dependency in `remix-dev/package.json`.

MichaelDeBoey:
@pcattori It's listed in the root `package.json`

MichaelDeBoey (approved):

PR #2412, opened 2022-03-20
fix: can't use tsconfig 'paths' aliases other than `~`

F3n67u:
Closes: https://github.com/remix-run/remix/issues/2298

- [ ] Docs
- [ ] Tests

PR #3074, opened 2022-05-03
refactor(remix-dev): clean up custom `tsconfig` loader

F3n67u:
Follow-up of #2412


MichaelDeBoey:
@F3n67u Please look into why the integration test are failing.

F3n67u:
> @F3n67u Please look into why the integration test are failing.

ok.

F3n67u:
After I bump `@remix-run/dev`'s `tsconfig-paths` version to `^4.0.0`, there will exist two different versions on `node_modules`.
1. `<workspace-root>/packages/remix-dev/node_modules/tsconfig-paths`: v4.0.0
2. `<workspace-root>/node_modules/tsconfig-paths`: v3.14.1, introduced by `eslint-plugin-import` and `eslint-import-resolver-typescript`

When running `yarn build`, `<workspace-root>/packages/remix-dev/node_modules` is not copied to `build/node_modules/@remix-run/dev`. so when running test the `tsconfig-paths` deps is resolved to `<workspace-root>/node_modules/tsconfig-paths(v3.14.1)` instead of `<workspace-root>/packages/remix-dev/node_modules/tsconfig-paths(v4.0.0)`, this is why the test is failed.

I don't know how to solve this problem, any thought? @MichaelDeBoey

F3n67u:
I make pr to bump tsconfig-paths version to v4 on https://github.com/import-js/eslint-plugin-import/pull/2447 and https://github.com/alexgorbatchev/eslint-import-resolver-typescript/pull/104. If our upstream deps all bump their `tsconfig-paths` to v4, then the problem will be solved either.

MichaelDeBoey:
@F3n67u Having `eslint-import-resolver-typescript` & `eslint-plugin-import` update their dependency on `tsconfig-paths` would be the easiest solution.

If they don't want to (because of loss of Node v4 compatibility) or it takes too long, maybe @kentcdodds can help out with figuring out the best approach here.

kentcdodds:
Well, if the latest version of tsconfig-paths requires Node v14 then we can't use it either because we still support Node v12 I believe 😬

F3n67u:
> After I bump `@remix-run/dev`'s `tsconfig-paths` version to `^4.0.0`, there will exist two different versions on `node_modules`.
> 
> 1. `<workspace-root>/packages/remix-dev/node_modules/tsconfig-paths`: v4.0.0
> 2. `<workspace-root>/node_modules/tsconfig-paths`: v3.14.1, introduced by `eslint-plugin-import` and `eslint-import-resolver-typescript`
> 
> When running `yarn build`, `<workspace-root>/packages/remix-dev/node_modules` is not copied to `build/node_modules/@remix-run/dev`. so when running test the `tsconfig-paths` deps is resolved to `<workspace-root>/node_modules/tsconfig-paths(v3.14.1)` instead of `<workspace-root>/packages/remix-dev/node_modules/tsconfig-paths(v4.0.0)`, this is why the test is failed.
> 
> I don't know how to solve this problem, any thought? @MichaelDeBoey
> 
> ### here is how to reproduce this resolve problem.
> ```shell
> # git checkout <this-pr>
> $ yarn
> $ yarn build
> $ touch build/node_modules/@remix-run/dev/test.js
> $ echo "console.log(require.resolve('tsconfig-paths'))" >> build/node_modules/@remix-run/dev/test.js
> $ node build/node_modules/@remix-run/dev/test.js                                                    
> /Users/feng/dev/remix/node_modules/tsconfig-paths/lib/index.js
> ```
> 
>  ```
> $ npm list tsconfig-paths                       
> remix-monorepo@ /Users/feng/dev/remix
> ├─┬ @remix-run/dev@1.4.3 -> ./packages/remix-dev
> │ └── tsconfig-paths@4.0.0
> └─┬ @remix-run/eslint-config@1.4.3 -> ./packages/remix-eslint-config
>   ├─┬ eslint-import-resolver-typescript@2.5.0
>   │ └── tsconfig-paths@3.14.1
>   └─┬ eslint-plugin-import@2.25.4
>     └── tsconfig-paths@3.14.1 deduped
> ```

@kentcdodds the real problem is current build dir cannot resolve a npm packages which is not hoisted to top node_modules dir.

MichaelDeBoey:
> Well, if the latest version of tsconfig-paths requires Node v14 then we can't use it either because we still support Node v12 I believe 😬

The minimum required Node version is v14 for Remix

But we're talking about these packages dropping v4 (yes 10! version below) here 🙈

kentcdodds:
Oh wow

MichaelDeBoey:
It doesn't seem like https://github.com/import-js/eslint-plugin-import/pull/2447 is going to be merged (very soon).

@kentcdodds This however shouldn't matter though.
Code/tests shouldn't fail because 2 dependencies have a dependency on another version of the same package though 🤷‍♂️
So it might be worth it to look at fixing that instead of waiting for `eslint-plugin-import` to be updated.

MichaelDeBoey:
@F3n67u @kentcdodds This problem seems to be fixed already without us knowing. 🤔

So I think we can just review the code & merge it if all is good.

MichaelDeBoey (approved):

F3n67u:
> @F3n67u @kentcdodds This problem seems to be fixed already without us knowing. 🤔
> 
> So I think we can just review the code & merge it if all is good.

wow. that's great.

kentcdodds (approved):
I love deleting code. Do we have tests for this? How was this change verified to have worked?

MichaelDeBoey:
@kentcdodds Tests were introduced in #2412 and they're still passing 🥳🎉

kentcdodds:
Thanks!

MichaelDeBoey:
@F3n67u I just see that your tests created in #2412 doesn't include a `jsconfig.json` file.
Are you willing to include such a test please?
If not, I'm happy to do it myself, but since it's all your code/tests & it's probably an oversight, I wanted to ask you first.

F3n67u:
> @F3n67u I just see that your tests created in #2412 doesn't include a `jsconfig.json` file.
> Are you willing to include such a test please?
> If not, I'm happy to do it myself, but since it's all your code/tests & it's probably an oversight, I wanted to ask you first.

@MichaelDeBoey good point. I will add a test for jsconfig.json later.

F3n67u:
> @F3n67u @kentcdodds This problem seems to be fixed already without us knowing. 🤔
> 
> So I think we can just review the code & merge it if all is good.

I think it's because we have not covered `jsconfig.json` scenario anymore, so the test passed even though the `tsconfig-paths` resolved to the v3 version which does not have `jsconfig.json` support  when testing .

kentcdodds:
Sorry, I haven't taken the time to fully understand the constraints and issues here. Can you help me understand you suggest we do?

kentcdodds:
I see. And why are we using v3 of that package in the tests?

MichaelDeBoey:
That's the actual question 🙈
Imo it should use v4, as `@remix-run/dev` is using that version of the package

According to @F3n67u's explanation, we're not copying over `node_modules` when running the build, which is a normal/expected behavior imo.
Our (integration) tests however use the local build files.
This means that `npm` will search for the nearest version of a certain package when needing to import it.

Since v4 is only used in `@remix-run/dev` and v3 is used by dependencies of other packages, v4 will be in `packages/remix-dev/node_modules` & **not** in the root's `node_modules`.
This has the implication that `npm` will find v3, so when we're using v4 specific API's/functionality, it will fail to do so.

kentcdodds:
Oh, I see. Yeah the way we're managing `node_modules` for our integration tests and things is ... not the best. We're hoping to eventually move to Nx which would help with this I think. But until then... I guess we'll just not be able to have tests for this without a lot of work. I think it's probably not worth the effort. If we tested it manually and it worked, let's just be satisfied for that for now. Maybe we could write the tests and add a `.skip` on them with a `// TODO: ` comment explaining why they're skipped for now.

01taylop:
Hello, I stumbled here from the `eslint-plugin-import` issue which I also want resolved. Unfortunately I don't think this will happen any time soon but I had a similar hoisting issue with a Babel dependency not so long ago. I think you can get around the issue by switching back to `npm` or by upgrading to Yarn 2/3 (whichever it's called). Just wanted to share that in case it is a path you want to explore.

PR #3075, opened 2022-05-03
fix: fix rollup playground livereload

brophdawg11:
We made a small change in #3059 that broke the playground live reload trigger from the rollup build.  This fixes that and also adds the same live reload for usages of `REMIX_LOCAL_DEV_OUTPUT_DIRECTORY`

Testing Strategy:
* Test out live-reload of playground apps
* Or try using `REMIX_LOCAL_DEV_OUTPUT_DIRECTORY=../path/to/remix/app yarn watch` and confirm live reload works there as well

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -869,8 +869,9 @@ export default function rollup(options) {
 }
 
 async function triggerLiveReload(appDir) {
-  // Tickle live reload by touching the server entry.  Include all extensions
-  // since they may or may not use TS and/or JSX
+  // Tickle live reload by touching the server entry
+  // Consider all of entry.server.{tsx,ts,jsx,js} since React may be used 
+  // via `React.createElement` without the need for JSX.
   let serverEntryPaths = [
     "entry.server.ts",
     "entry.server.tsx",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -868,37 +868,44 @@ export default function rollup(options) {
   return builds;
 }
 
+async function triggerLiveReload(appDir) {
+  // tickle live reload by touching the server entry
+  let serverEntryPaths = [
+    "entry.server.tsx",
+    "entry.server.js",
+    "entry.server.jsx",
+  ];
+  let serverEntryPath = serverEntryPaths
+    .map((entryFile) => path.join(appDir, "app", entryFile))
+    .find((entryPath) => fse.existsSync(entryPath));
+  let date = new Date();
+  await fs.promises.utimes(serverEntryPath, date, date);
+}
+
 function copyToPlaygrounds() {
   return {
     name: "copy-to-remix-playground",
     async writeBundle(options, bundle) {
-      let playgroundsDir = path.join(__dirname, "playground");
-      let playgrounds = await fs.promises.readdir(playgroundsDir);
-      let writtenDir = path.join(__dirname, options.dir);
-      for (let playground of playgrounds) {
-        let playgroundDir = path.join(playgroundsDir, playground);
-        if (!fse.statSync(playgroundDir).isDirectory()) {
-          continue;
+      if (activeOutputDir === "build") {
+        let playgroundsDir = path.join(__dirname, "playground");
+        let playgrounds = await fs.promises.readdir(playgroundsDir);
+        let writtenDir = path.join(__dirname, options.dir);
+        for (let playground of playgrounds) {
+          let playgroundDir = path.join(playgroundsDir, playground);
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -868,37 +868,44 @@ export default function rollup(options) {
...
     name: "copy-to-remix-playground",
     async writeBundle(options, bundle) {
-      let playgroundsDir = path.join(__dirname, "playground");
-      let playgrounds = await fs.promises.readdir(playgroundsDir);
-      let writtenDir = path.join(__dirname, options.dir);
-      for (let playground of playgrounds) {
-        let playgroundDir = path.join(playgroundsDir, playground);
-        if (!fse.statSync(playgroundDir).isDirectory()) {
-          continue;
+      if (activeOutputDir === "build") {
```
brophdawg11, line 889: Only update playgrounds when not using `REMIX_LOCAL_DEV_OUTPUT_DIRECTORY`

rollup.config.js
```diff
@@ -868,37 +868,44 @@ export default function rollup(options) {
...
+  let serverEntryPaths = [
+    "entry.server.tsx",
+    "entry.server.js",
+    "entry.server.jsx",
+  ];
+  let serverEntryPath = serverEntryPaths
+    .map((entryFile) => path.join(appDir, "app", entryFile))
+    .find((entryPath) => fse.existsSync(entryPath));
+  let date = new Date();
+  await fs.promises.utimes(serverEntryPath, date, date);
```
brophdawg11, line 882: Updated to just touch the file instead of having to read and re-write the file

rollup.config.js
```diff
@@ -868,37 +868,44 @@ export default function rollup(options) {
   return builds;
 }
 
+async function triggerLiveReload(appDir) {
+  // tickle live reload by touching the server entry
+  let serverEntryPaths = [
+    "entry.server.tsx",
```
pcattori, line 874: Is `entry.server.ts` intentionally omitted from this list?

brophdawg11: Hm, unsure.  This matches current behavior after https://github.com/remix-run/remix/pull/3059.  Looks like we used to only check `jsx`/`tsx`but esbuild allows `JSX` inside `.js` files so we added JSX.  I'll check if `.ts` supports JSX syntax and add if so

pcattori: In either case, I think it'd be good to add a comment explaining that esbuild allows JSX in `.js` and that's why `.js` is included here.

brophdawg11: Agreed - added one for clarity

kentcdodds: People can technically write `React.createElement` manually to avoid JSX. I don't see a reason to not just list all possible options.

rollup.config.js
```diff
@@ -871,10 +871,10 @@ export default function rollup(options) {
 async function triggerLiveReload(appDir) {
   // tickle live reload by touching the server entry
   let serverEntryPaths = [
+    "entry.server.ts",
```
pcattori, line 874: Can we add a similar comment to this line explaining that `React.createElement` can be used instead of JSX?

jacob-ebey (approved):

rollup.config.js
```diff
@@ -869,7 +869,8 @@ export default function rollup(options) {
 }
 
 async function triggerLiveReload(appDir) {
-  // tickle live reload by touching the server entry
+  // Tickle live reload by touching the server entry.  Include all extensions
+  // since they may or may not use TS and/or JSX
```
pcattori, line 873: nit: I think explicitly calling out that all these extensions are valid is good, but explaining _why_ they are valid is the tricky part.

Maybe something like:

```js
// Tickle live reload by touching the server entry.
// Consider all of `entry.server.{tsx,ts,jsx,js}` since React may be used via `React.createElement` without the need for JSX.
```

pcattori (changes requested):

rollup.config.js
```diff
@@ -868,37 +868,46 @@ export default function rollup(options) {
   return builds;
 }
 
+async function triggerLiveReload(appDir) {
+  // Tickle live reload by touching the server entry.  Include all extensions
+  // since they may or may not use TS and/or JSX
```
brophdawg11, line 873:
```suggestion
  // Tickle live reload by touching the server entry
  // Consider all of entry.server.{tsx,ts,jsx,js} since React may be used 
  // via `React.createElement` without the need for JSX.
```

pcattori (approved):

MichaelDeBoey (approved):

PR #3076, opened 2022-05-03
feat(remix-dev): pass `packageManager` to `remix.init` script

MichaelDeBoey:
This way, the `remix.init` script can call `npm/yarn/pnpm install` if needed.

The use-case I'm thinking about is having different dependencies when using JavaScript or TypeScript.


mcansh (approved):

kentcdodds (approved):
Perfect 👍

PR #3078, opened 2022-05-03
chore(remix-dev): default to "just the basics"

### LINKED ISSUE #3000: Suggestion: Default installation should point to "Just the Basics" by default

1.4.1

### Steps to Reproduce

Run npx create-remix@latest

### Expected Behavior

You are asked, "What type of app do you want to create?"
The default one that is selected is the more advanced option.

It might be better for beginners to remix if the default option was "Just the Basics"

### Actual Behavior

My suggestion is for the question "What type of app do you want to create?" make the default option "Just the Basics"

mcansh:
resolves #3000

Testing Strategy:

I ran the build and ran `node ./build/node_modules/create-remix/cli.js`, then i ran `yarn test` to make sure they all still passed

### DIFF
--- a/packages/remix-dev/cli/run.ts
+++ b/packages/remix-dev/cli/run.ts
@@ -259,12 +259,12 @@ export async function run(argv: string[] = process.argv.slice(2)) {
             },
             choices: [
               {
-                name: "A pre-configured stack ready for production",
-                value: "stack",
-              },
-              {
                 name: "Just the basics",
                 value: "template",
+              },
+              {
+                name: "A pre-configured stack ready for production",
+                value: "stack",
               },
             ],
           },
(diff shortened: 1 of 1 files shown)


MichaelDeBoey:
Maybe we could also add the following (like we do with `install`? 
```js
default: 'template',
```

mcansh:
> Maybe we could also add the following (like we do with `install`?
> 
> ```js
> default: 'template',
> ```

that's how i originally had it, but i think it's better to have the default at the top

MichaelDeBoey:
@mcansh My suggestion was to have both, not one or the other.

MichaelDeBoey (approved):

machour (approved):

kentcdodds (approved):
Sounds good 👍

PR #3079, opened 2022-05-03
ci: bump cli test process timeout

brophdawg11:
We had a 6 second timeout in `interactWithShell` that seems to be firing on windows test fairly often making them a bit flakey.  It also seems that when the timeout fires, we didn't kill the process which left it potentially running in the background and continuing to execute, holding up the jest process from completing.  This PR bumps the timeout from 6->10 seconds and calls `proc.kill()` when the timeout elapses

### DIFF
--- a/packages/remix-dev/__tests__/cli-test.ts
+++ b/packages/remix-dev/__tests__/cli-test.ts
@@ -345,9 +345,10 @@ async function interactWithShell(
   // output as possible.
   let timeout = setTimeout(() => {
     if (deferred.state.current === "pending") {
+      proc.kill();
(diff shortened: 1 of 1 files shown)


MichaelDeBoey (approved):

MichaelDeBoey:
@brophdawg11 I think it would be good to also cherry-pick this onto `dev`

pcattori (approved):

kentcdodds (approved):
Good call 👍

brophdawg11:
@kentcdodds @MichaelDeBoey I cherry-picked this into `dev` here to help with in-progress PRs: https://github.com/remix-run/remix/pull/3115

PR #3084, opened 2022-05-03
chore(remix-eslint-config): comment out replaceRemixImportsOptions as it's not currently used

mcansh:


mjackson (changes requested):
Looks like you also need to comment out everything you `require` from `packageExports` too, since it's also unused after this change.

mjackson (approved):

MichaelDeBoey (approved):

Issue #2677, opened 2022-04-07
Monorepo and dependency installation check


npirotte:
### What version of Remix are you using?

remix@1.3.5-pre.1

### Steps to Reproduce

* Setup a mono repo (for example with nx) with a remix application, having it's own package.json, and at least another library that have some third party dependency (simply install and import lodash, that will be enough).
* From remix application, import the component from the other library (so the third party is included into the build)
* Build or dev the remix application

### Expected Behavior

App is built or dev mode is started, no warning.

### Actual Behavior

App is build or dev mode is started, application is running fine, but a warning is printed in the console for each dependency that are imported by some of the monorepo component but not by remix application itself.

example:
```
> npx remix build

Building Remix app in production mode...
The path "react-dom/server" is imported in app/entry.server.tsx but react-dom is not listed in your package.json dependencies. Did you forget to install it?
The path "react" is imported in app/routes/demos/actions.tsx but react is not listed in your package.json dependencies. Did you forget to install it?
The path "@react-hook/size" is imported in ../bar-chart/src/lib/bar-chart.tsx but @react-hook/size is not listed in your package.json dependencies. Did you forget to install it?
The path "@visx/scale" is imported in ../bar-chart/src/lib/bar-chart.tsx but @visx/scale is not listed in your package.json dependencies. Did you forget to install it?
The path "@ngneat/falso" is imported in ../bar-chart/src/lib/bar-chart.tsx but @ngneat/falso is not listed in your package.json dependencies. Did you forget to install it?
The path "@visx/group" is imported in ../bar-chart/src/lib/bar-chart.tsx but @visx/group is not listed in your package.json dependencies. Did you forget to install it?
The path "@visx/shape" is imported in ../bar-chart/src/lib/bar-chart.tsx but @visx/shape is not listed in your package.json dependencies. Did you forget to install it?
The path "lodash/fp" is imported in ../bar-chart/src/lib/bar-chart.tsx but lodash is not listed in your package.json dependencies. Did you forget to install it?
The path "@visx/axis" is imported in ../bar-chart/src/lib/bar-chart.tsx but @visx/axis is not listed in your package.json dependencies. Did you forget to install it?
Built in 357m
```

Here is a link to the repo that generate those logs (running nx build demo)

F3n67u:
@npirotte could you create a reproduction repo for this issue?

npirotte:
ha, I forgot to paste the link: https://github.com/npirotte/hello-visx

Just run `nx build demo`

[...]

PR #3085, opened 2022-05-03
fix(remix-dev): update `serverBareModulesPlugin` to use `require.resolve` instead of using `package.json` to determine if deps are installed

mcansh:
Closes: #2677 

- [ ] Docs
- [ ] Tests

Testing Strategy:

trying to think of a way to test this as the build will fail due to the dep not being installed, so we can't use an `integration` test currently


lensbart:
This PR introduces spurious warnings when running `remix build` using [Yarn PnP](https://yarnpkg.com/features/pnp):

```
The path "@remix-run/deno" is imported in server.ts but @remix-run/deno is not listed in your package.json dependencies. Did you forget to install it?
The path "@remix-run/react" is imported in src/entry.server.tsx but @remix-run/react is not listed in your package.json dependencies. Did you forget to install it?
```

in my case, both dependencies are present in `package.json`. I think `require.resolve` is supposed to work with Yarn PnP, so not sure what’s wrong here.

Would it be sufficient to add `process.versions.pnp == null` check so that we don’t check if dependencies are installed when using Yarn PnP?

I also think the warning message itself may have to be updated, because it doesn’t check the presence of dependencies using `package.json` anymore, but rather whether the dependency is installed.

lensbart:
Warnings muted in https://github.com/remix-run/remix/pull/3599

Issue #2848, opened 2022-04-18
Some docs & examples use top-level `Date.now()` for cookie expiration


guillaumervls:
### What version of Remix are you using?

1.4.0

### Steps to Reproduce

(Problem is the docs, see below)

### Expected Behavior

Hi and thanks for the great work on Remix !

In the docs https://remix.run/docs/en/v1/api/remix#using-sessions and in some examples https://github.com/remix-run/remix/tree/main/examples/redis-upstash-session the proposed session storage code uses calls to top-level `Date.now()` for cookie expiration date.

While it may seem to work in development because it looks like the modules are re-imported for every request, it does not in production. `Date.now()` then does not have much sense, since it is only the timestamp when the server was started, not the timestamp of the request. 

Also, I found it confusing to create a cookie _outside_ the request/response lifecycle. As I understand, it's not _actually_ a cookie, but _cookie options_ instead. But it's weird, because the `expires` option doesn't make sense. (Unless we recreate the session storage for every request. And the docs & examples don't seem to go that way).

Hope I'm not missing something...

Cheers !

### Actual Behavior

(Problem is the docs, see above)

machour:
Fixed by #3091

PR #3091, opened 2022-05-04
chore(remix-server-runtime): warn against using `expires` in `createCookieSessionStorage` and `createSessionStorage`

mcansh:
specifying an 'expires' will cause all uses of `commitSession` to use that date.

~~this only warns for `createCookieSessionStorageFactory` and `createSessionStorageFactory` (which is used by `createCookieSessionStorage` and `createSessionStorage`) as our generic `createCookie` util doesn't warn about signing cookies, so i left it out here, but easy enough to add it for that too~~

Closes: #2848

- [ ] Docs
- [x] Tests

Testing Strategy:

This test covers this code: https://github.com/remix-run/remix/blob/29cf8ff5288d97405e09812dfc8f30efa10bb1d2/packages/remix-server-runtime/__tests__/sessions-test.ts#L142

### DIFF
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -251,9 +251,11 @@ function myUnescape(value: string): string {
 
 function warnOnceAboutExpiresCookie(name: string, expires?: Date) {
   warnOnce(
-    typeof expires === "undefined",
+    !(expires instanceof Date),
     `The "${name}" cookie has an "expires" property set. ` +
       `This will cause the expires value to not be updated when the session is committed. ` +
-      `Instead, use \`serialize("value", { expires })\` to set the expires value.`
+      `Instead, you should set the expires value when serializing the cookie. ` +
+      `You can use \`commitSession(session, { expires })\` if using a session abstraction, ` +
+      `or \`serialize("value", { expires })\` if you're creating a cookie yourself.`
   );
 }
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -255,7 +255,7 @@ function warnOnceAboutExpiresCookie(name: string, expires?: Date) {
     `The "${name}" cookie has an "expires" property set. ` +
       `This will cause the expires value to not be updated when the session is committed. ` +
       `Instead, you should set the expires value when serializing the cookie. ` +
-      `You can use \`commitSession(session, { expires })\` if using a session abstraction, ` +
-      `or \`serialize("value", { expires })\` if you're creating a cookie yourself.`
+      `You can use \`commitSession(session, { expires })\` if using a session storage object, ` +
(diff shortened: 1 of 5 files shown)


MichaelDeBoey (approved):

packages/remix-server-runtime/sessions.ts
```diff
@@ -260,6 +261,15 @@ export const createSessionStorageFactory =
     };
   };
 
+export function warnOnceAboutExpiresSessionCookie(cookie: Cookie) {
+  warnOnce(
+    typeof cookie.expires === "undefined",
+    `The "${cookie.name}" cookie has an "expires" property set. ` +
+      `This will cause the expires value to not be updated when the session is committed. ` +
+      `Instead use the second argument to commitSession \`commitSession(session, { expires })\` to set the expires value.`
```
mjackson, line 269:
```suggestion
      `Instead, use \`commitSession(session, { expires })\` to set the expires value.`
```

mjackson (changes requested):
Let's also warn when using `createCookie`. I can't think of any reason you'd want to create a cookie right off the bat with an `expires` value.

mcansh:
adding a test for the warning for "missing secrets" while im in here, and looks like the sessionStorages call out to createCookie so we end up with 2 warnings, moving the linear issue back to `in development` and reassigning it while i look into that 😬

packages/remix-server-runtime/cookies.ts
```diff
@@ -245,3 +248,14 @@ function myUnescape(value: string): string {
   }
   return result;
 }
+
+function warnOnceAboutExpiresCookie(name: string, expires?: Date) {
+  warnOnce(
+    !(expires instanceof Date),
```
MichaelDeBoey, line 254: It would lower the cognitive load if we just check if `expires` is set

```suggestion
    !expires,
```

MichaelDeBoey (changes requested):

packages/remix-server-runtime/cookies.ts
```diff
@@ -245,3 +248,14 @@ function myUnescape(value: string): string {
...
 }
+
+function warnOnceAboutExpiresCookie(name: string, expires?: Date) {
+  warnOnce(
+    !expires,
+    `The "${name}" cookie has an "expires" property set. ` +
+      `This will cause the expires value to not be updated when the session is committed. ` +
+      `Instead, you should set the expires value when serializing the cookie. ` +
+      `You can use \`commitSession(session, { expires })\` if using a session abstraction, ` +
+      `or \`serialize("value", { expires })\` if you're creating a cookie yourself.`
```
mjackson, line 259:
```suggestion
      `You can use \`commitSession(session, { expires })\` if using a session storage object, ` +
      `or \`cookie.serialize("value", { expires })\` if you're using the cookie directly.`
```

mjackson (approved):

PR #3093, opened 2022-05-04
feat: create new `config` package

mcansh:

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


packages/remix-dev/compiler/plugins/serverAssetsManifestPlugin.ts
```diff
@@ -1,8 +1,8 @@
 import type { Plugin } from "esbuild";
 import jsesc from "jsesc";
+import { assetsManifestVirtualModule } from "@remix-run/config";
```
mjackson, line 3: Why did this need to move to remix-config?

mjackson (changes requested):

packages/remix-serve/index.ts
```diff
@@ -1,30 +1,55 @@
...
+    // TODO: Deprecate using a string as 1st argument. Keep this behavior for
+    // now to avoid breaking changes.
+    serverBuildPath = serverBuildPathOrConfig;
+    assetsBuildDirectory = "public/build";
+    publicPath = "/build/";
+  } else {
+    serverBuildPath = serverBuildPathOrConfig.serverBuildPath;
+    assetsBuildDirectory = serverBuildPathOrConfig.assetsBuildDirectory;
+    publicPath = serverBuildPathOrConfig.publicPath;
+  }
```
MichaelDeBoey, line 29: This could be simplified to

```suggestion
  let { assetsBuildDirectory, publicPath, serverBuildPath } = typeof serverBuildPathOrConfig === "string"
    // TODO: Deprecate using a string as 1st argument. Keep this behavior for
    // now to avoid breaking changes.
    ? { 
        assetsBuildDirectory: "public/build",
        publicPath: "/build/",
        serverBuildPath: serverBuildPathOrConfig,
      }
    : serverBuildPathOrConfig
```

MichaelDeBoey (changes requested):

packages/remix-dev/compiler/plugins/serverAssetsManifestPlugin.ts
mcansh: we don't, it was due to just moving the `config` dir from dev over wholesale 🙈

mcansh: wait, we need `serverBuildVirtualModule` from it - we could just move that over though https://github.com/remix-run/remix/blob/d8180da3ddac5789fd3cd4feb820b28612140bf7/packages/remix-config/config.ts#L401

MichaelDeBoey:
@mcansh It seems like we have some conflicts to resolve here

PR #3096, opened 2022-05-04
docs(tutorials/blog): add warning about date serialization in blog tutorial

MMDF:
Add a small info box on the docs page about Date serializations

Perhaps there should be other additions to the docs page, but I wanted to start with the blog tutorial. This is a point that got me for a second and I think first-comers would benefit from knowing such information.

I don't think Remix should implement a feature to specially serialize and deserialize dates to overcome this issue, instead a simple reminder would go a long way.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -240,6 +240,7 @@
 - mirzafaizan
 - mjackson
 - mkrtchian
+- MMDF
 - mochi-sann
 - mohammadhosseinbagheri
 - monitaure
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
Not sure if we _really_ want this in our docs (especially the link to a non-Remix gist) tbh as this is a more generic "problem" that's not only about `Date`s.

kentcdodds:
I agree. It may be good to talk about this, but this isn't the place to do so and we don't want to recommend people do something that's in a gist. If we really want to support what's there then it should be built-in instead. Feel free to open a GitHub discussion to talk about that.

PR #3106, opened 2022-05-05
[bug] `getDependenciesToBundle` fails on `package.json` with no `main` field present

Bushuo:
Hi Remix Team,
thanks for this awesome library, I really enjoy using it in my side project.
I already opened this pull request a week ago and made a mistake in updating the branch.
So here it is again in its original form.

Please also look at #2594
I will write another pull request with the changes, I made locally as requested in the discussion.

- [ ] Docs
- [x] Tests

Testing Strategy:
This test shows, that getDependenciesToBundle fails if the dependency has no `main` field in the `package.json`

Other bugs I would like to mention and provide tests and fixes for:
* getDependenciesToBundle does not recurse and only outputs the input dependency
* getDependenciesToBundle can not resolve `@types/*` packages

What would be the best way to provide tests for this?
Another pull request or separate test files inside this one?


ZipBrandon:
@Bushuo  I want to anecdotally add that this PR has resolved the issues that I was experiencing with packages that I was dependent upon which did not have the `main` field in their `package.json`.  Thank you for your diligent work.

markdalgleish:
This particular scenario has been fixed in https://github.com/remix-run/remix/pull/7272. Note that the package must provide a `"./package.json"` export in order to resolve the path.

PR #3104, opened 2022-05-05
feat: add `@remix-run/netlify-edge` package + update Netlify template

nickytonline:
## Description

Adds Edge support for Netlify.

This brings in the work that currently resides in the [`experimental-netlify-edge`](https://github.com/remix-run/remix/tree/experimental-netlify-edge) branch and updates the Netlify template for Remix with the more up to date Edge configuration/files found in https://github.com/netlify/remix-edge-template.

It also introduces the `@remix-run/netlify-edge` package to Remix.

Relates to:
- #3092
- https://github.com/netlify/remix-edge-template/issues/71

- [ ] Docs
- [x] Tests (added deployment test)

## Getting Started

1. Checkout this branch, e.g. via the GitHub CLI `gh co 3104`
2. Ensure that the [Netlify CLI is installed or up to date](https://docs.netlify.com/cli/get-started/).
3. Run `yarn` to install all the dependencies of the project
4. Copy all the files in `templates/netlify` to a new folder you need to create called`scripts/playground/template.local`. This is required so that we can run the Netlify template in the Remix testing playground. (see https://github.com/remix-run/remix/blob/main/docs/pages/contributing.md#playground)
5. Run `yarn watch`. You'll see some output like this:

```sh
rollup v2.75.7
bundles packages/remix-netlify-edge/index.ts → build/node_modules/@remix-run/netlify-edge/dist...
created build/node_modules/@remix-run/netlify-edge/dist in 24ms
watch.onEnd $ node scripts/copy-build-to-dist.mjs
  🛠  Copying build/node_modules/@remix-run/architect/dist to packages/remix-architect/dist
  🛠  Copying build/node_modules/@remix-run/cloudflare/dist to packages/remix-cloudflare/dist
  🛠  Copying build/node_modules/@remix-run/cloudflare-pages/dist to packages/remix-cloudflare-pages/dist
  🛠  Copying build/node_modules/@remix-run/cloudflare-workers/dist to packages/remix-cloudflare-workers/dist
  🛠  Copying build/node_modules/@remix-run/dev/dist to packages/remix-dev/dist
  🛠  Copying build/node_modules/@remix-run/express/dist to packages/remix-express/dist
  🛠  Copying build/node_modules/@remix-run/netlify/dist to packages/remix-netlify/dist
  🛠  Copying build/node_modules/@remix-run/netlify-edge/dist to packages/remix-netlify-edge/dist
  🛠  Copying build/node_modules/@remix-run/node/dist to packages/remix-node/dist
[...]

PR #3107, opened 2022-05-06
feat(remix-netlify): broaden `peerDependencies`

MichaelDeBoey:
As suggested in https://github.com/remix-run/remix/pull/3104#discussion_r866382710, extracted this to not pollute #3104


kentcdodds (approved):
Definitely 👍

PR #3111, opened 2022-05-06
chore(examples): use helper functions as much as possible in `jokes` example

youbicode:
Hello,

A tiny change to use the already defined function `getUser` to get the `userId`

### DIFF
--- a/examples/jokes/app/utils/session.server.ts
+++ b/examples/jokes/app/utils/session.server.ts
@@ -60,7 +60,7 @@ export async function requireUserId(
   request: Request,
   redirectTo: string = new URL(request.url).pathname
 ) {
-  const userId = await getUser(request);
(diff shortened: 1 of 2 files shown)


examples/jokes/app/utils/session.server.ts
```diff
@@ -60,8 +60,7 @@ export async function requireUserId(
   request: Request,
   redirectTo: string = new URL(request.url).pathname
 ) {
-  const session = await getUserSession(request);
-  const userId = session.get("userId");
+  const userId = await getUser(request);
```
MichaelDeBoey, line 63: I think you meant to use `getUserId`?

```suggestion
  const userId = await getUserId(request);
```

MichaelDeBoey (changes requested):

Issue #934, opened 2021-12-07
[Bug]: Destroying an empty file session tries to delete the entire session directory


clarkdave:
### What version of Remix are you using?

1.0.5

### What version of Node are you using? Minimum supported version is 14.

17.0.1

### Steps to Reproduce

```ts
const sessionStorage = createFileSessionStorage({
  dir: path.join(process.cwd(), '.cache', 'sessions'),
  cookie: {
    name: '_session',
    sameSite: 'lax',
    path: '/',
    httpOnly: true,
  }
})

// create a new session (or forget to pass in a `request`...)
const session = await sessionStorage.getSession()

// delete the session
sessionStorage.destroySession(session)
```

### Expected Behavior

I'd expect either:

1. Nothing happens, as the session is empty
2. An error is thrown stating that an empty session cannot be destroyed

### Actual Behavior

Node will throw an error:

```
Error: EPERM: operation not permitted, unlink './.cache/sessions'
```

This is because the file storage adapter's `deleteData` method does not check if `id` is missing before trying to unlink, so the line `unlink(getFile(dir, id))` ends up being `unlink(dir)`. That results in the `EPERM` error from trying to unlink a directory instead of a file.

Although it of course doesn't make sense to delete an empty session, the typings do allow this to happen and the resulting error is confusing. I discovered this by accident when I was trying to delete a _non-empty_ session, but forgot to pass `request` into the `getSession()` function.

machour:
Still happening on latest release

machour:
Fixed in 1.7.0

PR #3113, opened 2022-05-06
fix(remix-node): don't delete entire session directory when destroying empty session file

abereghici:
The file storage adapter's deleteData method does not check if id is missing before trying to unlink, so the line unlink(getFile(dir, id)) ends up being unlink(dir).

That results in the EPERM error from trying to unlink a directory instead of a file.

This PR fixes [#934](https://github.com/remix-run/remix/issues/934) and covers it with a test.


MichaelDeBoey (approved):

machour (approved):

mcansh (approved):

kentcdodds:
Haha, I keep forgetting to add changeset files to these 🤦‍♂️  I'll get it eventually. Here's this one's: https://github.com/remix-run/remix/commit/27a7aa869163e1f1667f852412ff358811b4328e

PR #3114, opened 2022-05-06
fix(remix-dev/cli/migrate): directly use `jscodeshift` instead of via CLI

MichaelDeBoey:
Fixed the problem @lensbart mentioned in https://github.com/remix-run/remix/issues/683#issuecomment-1103507342

### DIFF
--- a/packages/remix-dev/cli/migrate/jscodeshift.ts
+++ b/packages/remix-dev/cli/migrate/jscodeshift.ts
@@ -1,11 +1,10 @@
-import { sync as execaSync } from "execa";
+// @ts-expect-error `@types/jscodeshift` doesn't have types for this
+import { run as jscodeshift } from "jscodeshift/src/Runner";
 
 import type { Flags } from "./flags";
 
-const jscodeshiftExecutable = require.resolve(".bin/jscodeshift");
-
 type Options = Record<string, unknown>;
-const toFlags = (options: Options = {}) =>
+const toCLIOptions = (options: Options = {}) =>
   Object.entries(options)
     .filter(([_key, value]) => Boolean(value))
     .map(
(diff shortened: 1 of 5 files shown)


packages/remix-dev/cli/migrate/jscodeshift.ts
```diff
@@ -1,60 +1,51 @@
-import { sync as execaSync } from "execa";
+// @ts-ignore `@types/jscodeshift` doesn't have types for this
```
pcattori, line 1: Can we use `@ts-expect-error` here instead so we are alerted if/when `@types/jscodeshift` does include types for the runner?

```suggestion
// @ts-expect-error `@types/jscodeshift` doesn't have types for this
```

pcattori: Also I figured out how to add types locally for a 3rd party package, so we could do so for this:


```ts
// types/jscodeshift.d.ts
declare module "jscodeshift/src/Runner" { // I think this will work even though there are `/` in the module specifier.
  interface Options {...} // TODO
  export const run: (transformPath: string, files: string[], options: Options) => Promise<{ error: number }>;
}
```

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/transform/options.ts
```diff
@@ -7,7 +7,7 @@
 import type { Adapter } from "./adapter";
 import type { Runtime } from "./runtime";
 
-export interface Options {
-  runtime: Runtime;
+export type Options = {
   adapter?: Adapter;
-}
+  runtime: Runtime;
+};
```
pcattori, line 13: I looks like this change doesn't semantically change anything. Can you keep this code as-is?

I personally like having `runtime` and then `adapter` since the `runtime` is required.

pcattori (changes requested):

packages/remix-dev/cli/migrate/migrations/replace-remix-imports/transform/options.ts
MichaelDeBoey: I needed to make it a `type` instead of `interface`, as `Option` should now extend `Record<string, unknown>`.

If I'm not making it a `type`, I get `TypeError`s

packages/remix-dev/cli/migrate/jscodeshift.ts
MichaelDeBoey: I'll create a PR to DT to fix these types.
Once it's merged, I'll update and remove the `@ts-expect-error`

packages/remix-dev/cli/migrate/jscodeshift.ts
```diff
@@ -1,60 +1,51 @@
...
   }
 
   try {
-    let result = execaSync(jscodeshiftExecutable, args, {
-      stdio: debug ? "inherit" : "ignore",
-      stripFinalNewline: false,
-    });
-    return result.exitCode === 0;
+    let { error } = await jscodeshift(transformPath, files, options);
+    return error === 0;
```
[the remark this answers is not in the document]
MichaelDeBoey: No it's not, it's the amount of errors
The result object is looking like

```js
{
  stats: {},
  timeElapsed: '0.001',
  error: 0,
  ok: 0,
  nochange: 0,
  skip: 0
}
```

See https://github.com/facebook/jscodeshift#usage-js

pcattori (approved):

PR #3117, opened 2022-05-06
feat(compiler): officially support Deno

pcattori:
## Docs

I will update `docs/` with mentions to `@remix-run/deno` as a follow-up PR targeting the `main` branch since there are changes there that we want to add to.

## Testing Strategy

Integration test: `deno-compiler-test.ts`

## Manually testing

Have to jump through a few hoops to get the locally built packages all plumbed through...

1. Build `remix-run/remix`:

```sh
yarn build
```

2. Package `@remix-run/deno` locally:

```sh
cd build/node_modules/@remix-run/deno
npm pack
```

3. Create a Remix app with the Deno template, without running `npm install`:

```sh
npx create-remix@latest --template <path/to/templates/deno>
# answer "n" (no) when prompted if `npm install` should be run
```

4. Install the locally packaged `@remix-run/deno`:

```sh
npm i <path/to/remix/build/node_modules/@remix-run/deno/package.tar.gz>
```

5. Use the local compiler to build, run, and dev

```sh
node <path/to/remix/build/node_modules/@remix-run/dev/cli.js> build
node <path/to/remix/build/node_modules/@remix-run/dev/cli.js> start
```

```sh
node <path/to/remix/build/node_modules/@remix-run/dev/cli.js> watch
# in another terminal window...
npm run dev:deno
```


packages/remix-dev/compiler/plugins/urlImportsPlugin.ts
```diff
@@ -0,0 +1,15 @@
+import type { Plugin } from "esbuild";
+
+/**
+ * Mark all URL imports as external so that the URL import is preserved in the build output.
+ */
+export const urlImportsPlugin = (): Plugin => {
+  return {
+    name: "url-imports",
+    setup(build) {
+      build.onResolve({ filter: /^https?:\/\// }, ({ path }) => {
```
MichaelDeBoey, line 10: Shouldn't we make this more specific, so we _really_ only support https://deno.land & https://esm.sh ? 🤔

templates/deno/package.json
```diff
@@ -16,5 +16,11 @@
   },
   "engines": {
     "node": ">=14"
+  },
+  "dependencies": {
+    "@remix-run/deno": "*",
+    "@remix-run/react": "*",
+    "react": "^18.1.0",
+    "react-dom": "^18.1.0"
```
MichaelDeBoey, line 24: I think we should keep using v17 for now, so we're in line with all other templates.

Shouldn't we add `@types/react`, `@types/react-dom` & `typescript` in `devDependencies` too?

```suggestion
    "react": "^17.0.2",
    "react-dom": "^17.0.2"
```

packages/remix-deno/globals.ts
```diff
@@ -0,0 +1,12 @@
+/*
+Remix provides `process.env.NODE_ENV` at compile time.
+Declare types for `process` here so that they are available in Deno.
+*/
+
+interface ProcessEnv {
```
MichaelDeBoey, line 6: I think these should all be inside a `declare global`, right?

MichaelDeBoey (changes requested):
We'll probably also need `magicExports`, like we have for `cloudflare` & `node` runtimes?

packages/remix-deno/globals.ts
pcattori: No, `declare global` is only valid syntax in a typescript module (a TS file with `import` / `export` in it).
TS files that are not module (like this one) are automatically in the global scope.

packages/remix-dev/compiler/plugins/urlImportsPlugin.ts
pcattori: No, Deno could import from arbitrary URLs, not just those ones.

MichaelDeBoey: Your commit description made it sound like we only wanted to support these official ones, that's why I was a bit confused I guess.

MichaelDeBoey:
> We don't need nor want magic exports for Deno since we don't have to support existing Deno projects that `import {...} from "remix"`.

This would break our "just copy over your `app` directory to a new project with different adapter/runtime" promise though.

pcattori:
> > We don't need nor want magic exports for Deno since we don't have to support existing Deno projects that `import {...} from "remix"`.
> 
> This would break our "just copy over your `app` directory to a new project with different adapter/runtime" promise though.

We can't promise that. E.g. if someone writes `Deno.readFile` in their `loader`, that won't work on Node. A completely portable `app/` folder would be ideal, but its not something we currently do.

templates/deno/package.json
pcattori: We don't need `@types/*` nor `typescript` in `devDependencies` because users will be relying on Deno tools (like the VSCode Deno extension) to provide types. These tools will use import maps to find the appropriate types (e.g. `templates/deno/.vscode/resolve_npm_imports.json`).

MichaelDeBoey:
> We can't promise that. E.g. if someone writes `Deno.readFile` in their `loader`, that won't work on Node. A completely portable `app/` folder would be ideal, but its not something we currently do.

Good point on that one.

If I recall correctly, thie "advice" @mjackson used to give for changing adapter/runtime has always been "create a new project with the required adapter/runtime & copy over your `app` folder" (I even thought it was somewhere in the docs, but can't find it), that's why I was thinking the `magicExports` would be needed

docs/decisions/0001-use-npm-to-manage-npm-dependencies-for-deno-projects.md
```diff
@@ -0,0 +1,97 @@
...
+
+Use `npm` and `node_modules/` to manage NPM dependencies like `react` for Remix projects, even when using Deno with Remix.
+
+Deno module dependencies (e.g. from `https://deno.land`) can still be managed via URL imports.
+
+### Allow URL imports
+
+Remix will preserve any URL imports in the built bundles as external dependencies,
+letting your browser runtime and server runtime handle them accordingly.
+That means thay you may:
```
MichaelDeBoey, line 56:
```suggestion
That means that you may:
```

MichaelDeBoey (changes requested):

packages/remix-deno/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "type": "git",
+    "url": "https://github.com/remix-run/remix",
+    "directory": "packages/remix-deno"
+  },
+  "bugs": {
+    "url": "https://github.com/remix-run/remix/issues"
+  },
+  "sideEffects": false,
+  "dependencies": {
+    "@remix-run/server-runtime": "*",
```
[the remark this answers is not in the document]
pcattori: The version script will automatically overwrite this to the correct version for us before the first release with this feature is published.

https://github.com/remix-run/remix/blob/4675aefede5b91f02187b7ef564d1cf5effbb800/scripts/utils.js#L187

CanRau:
😍 Very excited about this, can't wait to give it a try 🥳

integration/helpers/deno-template/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "remix-template-deno",
+  "private": true,
+  "sideEffects": false,
+  "dependencies": {
+    "@remix-run/deno": "*",
+    "@remix-run/react": "*",
+    "react": "^18.1.0",
+    "react-dom": "^18.1.0"
```
MichaelDeBoey, line 9: Just like in the `remix-deno` template, I think this should still be v17 to be consistent with all other templates we have.

```suggestion
    "react": "^17.0.2",
    "react-dom": "^17.0.2"
```

MichaelDeBoey (changes requested):

integration/deno-compiler-test.ts
```diff
@@ -0,0 +1,231 @@
...
+      "node_modules/npm-component/package.json": json({
+        name: "component",
+        version: "1.0.0",
+        sideEffects: false,
+      }),
+      "node_modules/npm-component/index.js": js`
+        module.exports = { npmComponent: () => "NPM_COMPONENT" };
+      `,
+      "node_modules/npm-util/package.json": json({
+        name: "util",
```
MichaelDeBoey, line 103: Shouldn't this be `npm-util` to be correct? 🤔

```suggestion
        name: "npm-util",
```

integration/deno-compiler-test.ts
```diff
@@ -0,0 +1,231 @@
...
+        version: "1.0.0",
+        sideEffects: false,
+      }),
+      "node_modules/npm-util/index.js": js`
+        module.exports = { npmUtil: () => "NPM_UTIL" };
+      `,
+      "node_modules/npm-server-only/package.json": json({
+        name: "server-only",
```
MichaelDeBoey, line 111: Shouldn't this be `npm-server-only` to be correct? 🤔

```suggestion
        name: "npm-server-only",
```

integration/deno-compiler-test.ts
```diff
@@ -0,0 +1,231 @@
...
+        version: "1.0.0",
+        sideEffects: false,
```
MichaelDeBoey, line 113: Since it's a server-only package, this should be set to `true` to make it throw whenever it's used on the client I think?

```suggestion
        sideEffects: true,
```

MichaelDeBoey (changes requested):

integration/deno-compiler-test.ts
pcattori: No, this is supposed to model a library with no side-effects so that we can test treeshaking, so [`sideEffects` needs to be `false`](https://esbuild.github.io/api/#conditionally-injecting-a-file).

MichaelDeBoey (approved):

integration/deno-compiler-test.ts
```diff
@@ -0,0 +1,231 @@
...
+  expect(utilFiles.length).toBeGreaterThanOrEqual(1);
+
+  let componentFiles = await searchFiles(
+    importPattern("https://deno.land/x/component.ts"),
+    browserCodeFiles
+  );
+  expect(componentFiles.length).toBeGreaterThanOrEqual(1);
+
+  /*
+  url imports _could_ have side effects,
```
mjackson, line 173: Let's add `sideEffects: false` in our plugin that resolves HTTP imports since it is the most common thing for http imports to not have side effects. https://esbuild.github.io/plugins/#on-resolve-results

integration/deno-compiler-test.ts
```diff
@@ -0,0 +1,231 @@
...
+  let serverOnlyImports = await searchFiles(
+    importPattern("npm-server-only"),
+    browserCodeFiles
+  );
+  expect(serverOnlyImports.length).toBe(0);
+  let serverOnlyFiles = await searchFiles("NPM_SERVER_ONLY", browserCodeFiles);
+  expect(serverOnlyFiles.length).toBe(0);
+});
+
+test("compiles bundles deno export of 3rd party package", async () => {
```
pcattori, line 226:
```suggestion
test("compiler bundles deno export of 3rd party package", async () => {
```

packages/remix-deno/server.ts
```diff
@@ -47,7 +46,7 @@ export async function serveStaticFiles(
     cacheControl?: string | ((url: URL) => string);
     publicDir?: string;
     assetsPublicPath?: string;
-  }
+  },
 ) {
   const url = new URL(request.url);
```
mjackson, line 52: Let's throw a `FileNotFoundError` instead of whatever `Deno.readFile` throws when it can't find something.

templates/deno/README.md
```diff
@@ -1,69 +1,62 @@
...
+- ✅ You may use inlined URL imports or [deps.ts](https://deno.land/manual/examples/manage_dependencies#managing-dependencies) for Deno modules.
+  ```ts
+  import { copy } from "https://deno.land/std@0.138.0/streams/conversion.ts";
+  ```
+- ❌ Do not use [import maps](https://deno.land/manual/linking_to_external_code/import_maps).
 
-```sh
-npm run start
-```
+Read about [why we use `npm` to manage dependencies for Deno projects](https://github.com/remix-run/remix/blob/main/docs/decisions/0001-use-npm-to-manage-npm-dependencies-for-deno-projects.md
```
mjackson, line 28: Let's move this up so people see it before they start installing stuff.

templates/deno/package.json
```diff
@@ -7,7 +7,9 @@
     "dev": "remix build && run-p dev:*",
     "dev:remix": "remix watch",
     "dev:deno": "cross-env NODE_ENV=development deno run --unstable --watch --allow-net --allow-read --allow-env ./build/index.js",
-    "start": "cross-env NODE_ENV=production deno run --unstable --allow-net --allow-read --allow-env ./build/index.js"
+    "start": "cross-env NODE_ENV=production deno run --unstable --allow-net --allow-read --allow-env ./build/index.js",
+    "lint": "deno lint --ignore=node_modules",
+    "fmt": "deno fmt --ignore=node_modules"
```
mjackson, line 12: Let's call this `format` instead of `fmt`

mjackson (changes requested):

packages/remix-deno/server.ts
```diff
@@ -37,6 +36,12 @@ export function createRequestHandler<Context = unknown>({
   };
 }
 
+class StaticFileNotFoundError extends Error {
+  constructor(message: string) {
+    super(message);
+  }
```
mjackson, line 42:
```suggestion
  constructor(file: string) {
    super(`No such file or directory: ${file}`);
  }
```

packages/remix-deno/server.ts
```diff
@@ -37,6 +36,12 @@ export function createRequestHandler<Context = unknown>({
   };
 }
 
+class StaticFileNotFoundError extends Error {
```
mjackson, line 39:
```suggestion
class FileNotFoundError extends Error {
```

packages/remix-deno/server.ts
```diff
@@ -65,9 +70,18 @@ export async function serveStaticFiles(
...
-  return new Response(file, { headers });
+  const filePath = path.join(publicDir, url.pathname);
+  try {
+    const file = await Deno.readFile(filePath);
+    return new Response(file, { headers });
+  } catch (error) {
+    if (error.code === "EISDIR" || error.code === "ENOENT") {
+      throw new StaticFileNotFoundError(
+        `No such file or directory: ${filePath}`,
+      );
```
mjackson, line 81:
```suggestion
      throw new FileNotFoundError(filePath);
```

packages/remix-deno/server.ts
```diff
@@ -94,7 +108,7 @@ export function createRequestHandlerWithStaticFiles<Context = unknown>({
     try {
       return await serveStaticFiles(request, staticFiles);
     } catch (error) {
-      if (error.code !== "EISDIR" && error.code !== "ENOENT") {
+      if (!(error instanceof StaticFileNotFoundError)) {
```
mjackson, line 111:
```suggestion
      if (!(error instanceof FileNotFoundError)) {
```

mjackson (changes requested):

mjackson (approved):

MichaelDeBoey:
@pcattori The Cloudflare & Node runtimes both also export the following functions:

- `MaxPartSizeExceededError`
- `unstable_composeUploadHandlers`
- `unstable_createMemoryUploadHandler`
- `unstable_parseMultipartFormData`

and the following types:

- `MemoryUploadHandlerFilterArgs`
- `MemoryUploadHandlerOptions`
- `UploadHandler`
- `UploadHandlerPart`

Do we want to implement them for Deno as well, so we have feature parity?

PR #3119, opened 2022-05-07
docs(example): add twind example

nexxeln:
This example shows how to use [Twind](https://twind.dev) with Remix. 
Thank you. Remix rocks

demo gif


kentcdodds (changes requested):
Sweet! Thanks! Does this server render? It doesn't look like it does. Is it possible to make it so we don't get a flash of unstyled content before the JS loads?

nexxeln:
Twind does support server side rendering: 
- https://twind.dev/usage-guides/ssr.html

I'll implement this in a bit.

nexxeln:
Turns out, for SSR in Remix they have a package. I'll use that and commit the changes.

examples/twind/app/root.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <html lang="en" className="bg-[#121212] text-slate-100">
+      <head>
+        <Meta />
+        <Links />
+      </head>
+      <body>
+        <Outlet />
+        <ScrollRestoration />
+        <Scripts />
+        {process.env.NODE_ENV === 'development' && <LiveReload />}
```
MichaelDeBoey, line 33:
```suggestion
        <LiveReload />
```

MichaelDeBoey (changes requested):

kentcdodds (approved):
Sweet! Thanks!

PR #3127, opened 2022-05-07
fix(remix-react): detect index route from manifest in `useFormAction`

penspinner:
The `useFormAction` hook detects the file ID ending in "/index" to determine whether or not to return a form action with the  `?index` search param. This holds true in Remix's default route convention. However, in the case where the consumer defines custom routes, this may not hold true anymore.

---

Testing Strategy:

I'm using the [Remix Flat Routes](https://github.com/kiliman/remix-flat-routes/) flat-folders convention in my project.

Let's say I have a route `_auth.dashboard.index/_routes.tsx`. That's considered to be an index route in the flat-folders convention. However, rendering a `<Form>` in that route results in a form with the action `"/dashboard"` while the action should be `"/dashboard?index"`.

### DIFF
--- a/integration/form-test.ts
+++ b/integration/form-test.ts
@@ -1064,3 +1064,61 @@ test.describe("Forms", () => {
     });
   }
 });
+
+test.describe("Forms in an app with custom routes", () => {
+  let fixture: Fixture;
+  let appFixture: AppFixture;
+
+  test.beforeAll(async () => {
+    fixture = await createFixture({
+      files: {
+        "remix.config.js": js`
+          module.exports = {
+            routes(defineRoutes) {
+              return defineRoutes((defineRoute) => {
+                defineRoute("form-custom-route", "routes/form-custom-route/_route.jsx", () => {
+                  defineRoute("", "routes/form-custom-route.index/_route.jsx", { index: true });
+                });
+              });
+            },
+          };
+        `,
+        "app/routes/form-custom-route/_route.jsx": js`
+          import { Form, Outlet } from "@remix-run/react"
+          export default function FormCustomRoute() {
+            return (
+              <>
+                <Form id="form" />
+                <Outlet />
+              </>
+            )
+          }
+        `,
+        "app/routes/form-custom-route.index/_route.jsx": js`
+          import { Form } from "@remix-run/react"
+          export default function FormCustomRouteIndex() {
+            return <Form id="form-index" />
+          }
+        `,
+      },
+    });
+    appFixture = await createAppFixture(fixture);
+  });
+
+  test.afterAll(async () => {
+    await appFixture.close();
+  });
+
+  test("Form has correct action in a layout route and an index route", async ({
+    page,
+  }) => {
+    let app = new PlaywrightFixture(appFixture, page);
+    await app.goto("/form-custom-route");
+    expect((await app.getElement("#form")).attr("action")).toBe(
+      "/form-custom-route"
+    );
+    expect((await app.getElement("#form-index")).attr("action")).toBe(
+      "/form-custom-route?index"
+    );
+  });
+});
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1150,6 +1150,7 @@ export function useFormAction(
   method: FormMethod = "get"
 ): string {
   let { id } = useRemixRouteContext();
+  let { manifest } = useRemixEntryContext();
   let resolvedPath = useResolvedPath(action ? action : ".");
 
   // Previously we set the default action to ".". The problem with this is that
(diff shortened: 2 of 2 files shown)


packages/remix-react/components.tsx
```diff
@@ -959,9 +959,10 @@ export function useFormAction(
   method: FormMethod = "get"
 ): string {
   let { id } = useRemixRouteContext();
+  let { manifest } = useRemixEntryContext();
   let path = useResolvedPath(action);
   let search = path.search;
-  let isIndexRoute = id.endsWith("/index");
+  let isIndexRoute = manifest.routes[id].index;
```
MichaelDeBoey, line 965: Since `isIndexRoute` suggests a boolean value, this should be converted to a boolean value.

```suggestion
  let isIndexRoute = Boolean(manifest.routes[id].index);
```

MichaelDeBoey (changes requested):

MichaelDeBoey:
@penspinner Could you please also add a test for this change?

integration/form-test.ts
```diff
@@ -576,3 +576,53 @@ test.describe("Forms", () => {
...
   });
 });
+
+test.describe("Forms in an app with custom defineRoutes", () => {
+  let fixture: Fixture;
+  let appFixture: AppFixture;
+
+  test.beforeAll(async () => {
+    fixture = await createFixture({
+      template: "node-custom-routes-template",
```
MichaelDeBoey, line 586: @penspinner Is there a specific reason you're not just using `node_template` (the default) here?

You could also use [Pathless Layout Routes](https://remix.run/docs/en/v1/guides/routing#pathless-layout-routes) if you want to remove certain sections from the URL.

penspinner: I have to test for the case where the route does not end in `"/index"` but `useFormAction` still adds the `"?index"` to the form action. Unless I'm mistaken, I think this only happens when the user is defining custom routes. There doesn't seem to be a way to pass a `defineRoutes` to `createFixture`, so I had to create a new template to test this.

MichaelDeBoey: You can always create a `remix.config.js` file entry in your test if you want change the behaviour of it.

penspinner: Oh, I see what you mean now. Didn't know we could define the Remix config in the `createFixture`.

integration/form-test.ts
```diff
@@ -576,3 +576,64 @@ test.describe("Forms", () => {
...
+test.describe("Forms in an app with custom routes", () => {
+  let fixture: Fixture;
+  let appFixture: AppFixture;
+
+  test.beforeAll(async () => {
+    fixture = await createFixture({
+      files: {
+        "remix.config.js": `
+          module.exports = {
+            ignoredRouteFiles: ["**/.*"],
```
MichaelDeBoey, line 589: We're not defining any dotfiles, so we can remove this line and make the test as clean as possible

```suggestion
```

integration/form-test.ts
```diff
@@ -576,3 +576,64 @@ test.describe("Forms", () => {
...
 });
+
+test.describe("Forms in an app with custom routes", () => {
+  let fixture: Fixture;
+  let appFixture: AppFixture;
+
+  test.beforeAll(async () => {
+    fixture = await createFixture({
+      files: {
+        "remix.config.js": `
```
MichaelDeBoey, line 587: Make sure to use our helper function, for better IDE code highlighting

```suggestion
        "remix.config.js": js`
```

integration/form-test.ts
```diff
@@ -576,3 +576,64 @@ test.describe("Forms", () => {
...
+            routes(defineRoutes) {
+              return defineRoutes((defineRoute) => {
+                defineRoute("form-custom-route", "routes/form-custom-route/_route.jsx", () => {
+                  defineRoute("", "routes/form-custom-route.index/_route.jsx", { index: true });
+                });
+              });
+            },
+          };
+        `,
+        "app/routes/form-custom-route/_route.jsx": `
```
MichaelDeBoey, line 599:
```suggestion
        "app/routes/form-custom-route/_route.jsx": js`
```

integration/form-test.ts
```diff
@@ -576,3 +576,64 @@ test.describe("Forms", () => {
...
+          export default function FormCustomRoute() {
+            return (
+              <>
+                <Form id="form" />
+                <Outlet />
+              </>
+            )
+          }
+        `,
+        "app/routes/form-custom-route.index/_route.jsx": `
```
MichaelDeBoey, line 610:
```suggestion
        "app/routes/form-custom-route.index/_route.jsx": js`
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

kiliman:
Awesome work @penspinner. BTW: I did fix remix flat routes to work around this issue. I append `/index` to all the index route ids so Remix works properly. Once this PR lands, I'll be able to remove the hack. Thanks!

MichaelDeBoey:
Hi @penspinner!

penspinner:
@MichaelDeBoey This seems to have been fixed in [React Router](https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L1046-L1050). Gonna close this.

PR #3132, opened 2022-05-08
fix(remix-dev): dont use nullish coalescing in cli to support node 12

mcansh:

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/cli/run.ts
+++ b/packages/remix-dev/cli/run.ts
@@ -19,11 +19,11 @@ import { validateNewProjectPath, validateTemplate } from "./create";
  * that can be used to determine which package manager ran
  * the command.
  */
-function getPreferredPackageManager() {
-  return ((process.env.npm_user_agent ?? "").split("/")[0] || "npm") as
-    | "npm"
-    | "yarn"
-    | "pnpm";
+function getPreferredPackageManager(): "npm" | "yarn" | "pnpm" {
+  let npmUserAgentString = process.env.npm_user_agent || "";
+  let [npmUserAgent] = npmUserAgentString.split("/");
+  if (!npmUserAgent) return "npm";
+  return npmUserAgent as "npm" | "yarn" | "pnpm";
 }
 
 const helpText = `


MichaelDeBoey:
I don't think this is necessary, as the very first thing the CLI is checking is if the user is using Node v14 or higher 🤔

https://github.com/remix-run/remix/blob/eb4e899b1129fd96eba744c2342dbe61669227d8/packages/remix-dev/cli/run.ts#L145-L152

MichaelDeBoey (commented):
I don't think this is necessary, as the very first thing the CLI is checking is if the user is using Node v14 or higher 🤔

https://github.com/remix-run/remix/blob/eb4e899b1129fd96eba744c2342dbe61669227d8/packages/remix-dev/cli/run.ts#L145-L152

mcansh:
it is indeed needed image

MichaelDeBoey:
Hmmm that's weird, it should already stop at the thrown error, no?
The `getPreferredPackageManager` function isn't called yet there 🤔

mcansh:
> The `getPreferredPackageManager` function isn't called yet there 🤔

i don't think it matters if or when it's called, but when it's parsed it errors out

MichaelDeBoey:
> but when it's parsed it errors out

Does that mean we can _never_ have v14 syntax?
Or is that only in the `packages/remix-dev/cli/run.ts` file?

sergiodxa:
Node v12 LTS maintenance life ended last month, I don’t think Remix should support it, even if it fails without showing an specific error

mcansh:
> Node v12 LTS maintenance life ended last month, I don’t think Remix should support it, even if it fails without showing an specific error

we only support it enough to say we don't support it 😅

MichaelDeBoey:
Questions remains: what about earlier versions?

MichaelDeBoey:
@mcansh This function has been extracted into a separate file in #2864, so I think this isn't necessary anymore? 🤔

mcansh:
yup looks good now :) image

PR #3134, opened 2022-05-09
tests: don't call `remix init` when creating fixtures

MichaelDeBoey:
All tests are using the new imports, so there's no need to run `remix init` in the tests when creating fixtures


MichaelDeBoey:
We still have tests to check that `magicExports` are working

PR #3141, opened 2022-05-09
docs(guides/styling): clarify Tailwind example

janhoogeveen:
Small change, but this makes the documentation on Tailwind easier to grasp for newcomers.


docs/guides/styling.md
```diff
@@ -448,6 +448,7 @@ Finally, import the generated CSS file into your app:
 
 ```tsx filename=app/root.tsx
 // ...
+import type { LinksFunction } from "@remix-run/node";
```
MichaelDeBoey, line 451: I think it would be better to do it like this then

```suggestion
import type { LinksFunction } from "@remix-run/node"; // or "@remix-run/cloudflare"

// ...

```

MichaelDeBoey (changes requested):

machour (approved):

MichaelDeBoey (approved):

PR #3142, opened 2022-05-09
ci: use tags instead of releases

mcansh:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey:
@mcansh Seems like some conflicts occurred.

MichaelDeBoey:
@mcansh It seems like we have some conflicts to resolve here

mcansh:
i think this is just about ready to go, i hit the api limit in testing that we compare the right previous version https://github.com/mcansh/nightly-release-test/actions/runs/2511849483, so i'll verify again tomorrow

scripts/release/github.ts
```diff
@@ -0,0 +1,337 @@
...
+  );
+
+  let prs = await getMergedPRsBetweenTags(
+    owner,
+    repo,
+    previousTag,
+    currentTag,
+    currentTag.isPrerelease && previousTag.isPrerelease
+      ? NIGHTLY_BRANCH
+      : DEFAULT_BRANCH
```
mcansh, line 48: i think this is my main area of question, and i think i need an additional check for when going from a stable to nightly

nightly > nightly -> 'dev'
nightly > stable -> 'main'
stable -> nightly -> 'main' // but should be dev

mcansh: however this run makes me think otherwise 🙈 https://github.com/mcansh/nightly-release-test/runs/6937790319?check_suite_focus=true

kentcdodds (approved):
Looks good 👍 Thanks @mcansh!

PR #3012, opened 2022-04-28
chore(remix-dev): when using jsconfig, include js in jsconfig files instead of ts

mcansh:
Depends on #2961

PR #3150, opened 2022-05-10
feat(remix-dev/cli/migrate): add `convert-to-javascript` migration

MichaelDeBoey:
When working on supporting the `isTypeScript` variable in `remix.init` in our default stacks (see https://github.com/remix-run/indie-stack/pull/63, https://github.com/remix-run/blues-stack/pull/57 & https://github.com/remix-run/grunge-stack/pull/50), I needed a way to transform all import statements in files outside the `app` folder from

```js
import Prisma from 'prisma';
import { foo } from 'bar';
```

to

```js
const Prisma = require('prisma');
const { foo } = require('bar');
```

As discussed with @kentcdodds & @mcansh, it would be a good idea to do this transform for _all_ stacks, so we came to the conclusion that we should have something inside the `@remix-run/dev` package for that.

We're currently already have a [`convertTemplateToJavaScript` function](https://github.com/remix-run/remix/blob/4675aefede5b91f02187b7ef564d1cf5effbb800/packages/remix-dev/cli/convert-to-javascript.ts), that's used in our `create` command
https://github.com/remix-run/remix/blob/4675aefede5b91f02187b7ef564d1cf5effbb800/packages/remix-dev/cli/create.ts#L206-L208
but as it's name says, that's only converting all files from TS to JS (which in practice is just removing all type-related things and leave the rest of the code as-is).
https://github.com/remix-run/remix/blob/4675aefede5b91f02187b7ef564d1cf5effbb800/packages/remix-dev/cli/convert-to-javascript.ts#L16-L23

This PR will add the necessary changes to the import statements in all files **outside** the `app` folder.

I took the liberty to also include @mcansh's changes to `convertTemplateToJavaScript` from #3012, as these are also some changes that all stacks could benefit from.

To see how this transform works, you can watch it on [AST explorer](https://astexplorer.net/#/gist/ee39a24bb9986c1a6aad14d1dbf9f83e/e4ec9da2ca5a1db8a5c483e32f6238b7ae1ab8a8).

---

I'll create a follow-up PR to remove the `convertTemplateToJavaScript` function from the `create` command & replace it by this new migration.


packages/remix-dev/cli/migrate/migrations/convert-to-javascript/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // 2. Remove @types/* & TypeScript dependencies + typecheck script from package.json
+  await cleanupPackageJson(projectDir);
+
+  // 3. Run codemod
+  let config = await readConfig(projectDir);
+  let files = glob.sync("**/*.+(ts|tsx)", {
+    absolute: true,
+    cwd: config.appDirectory,
+    ignore: ["./app/**/*"],
```
mcansh, line 28: might want to also use the appDirectory here
```suggestion
    ignore: [`./${path.dirname(config.appDirectory)}/**/*`],
```

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    cwd: config.appDirectory,
+    ignore: [`./${dirname(config.appDirectory)}/**/*`],
+  });
+  let codemodOk = await jscodeshift.run({
+    files,
+    flags,
+    transformPath: TRANSFORM_PATH,
+  });
+  if (!codemodOk) {
+    console.error("❌ I couldn't replace all of your `remix` imports.");
```
pcattori, line 36: Probably need to change this error message to match the context of this migration.

```suggestion
    console.error("❌ I couldn't convert your files to Javascript.");
```
...or something like that.

pcattori (changes requested):

packages/remix-dev/__tests__/migrations/fixtures/indie-stack/.gitpod.yml
```diff
@@ -0,0 +1,48 @@
+# https://www.gitpod.io/docs/config-gitpod-file
```
kentcdodds, line 1: In an effort to reduce noise in the repo, let's delete as much as we can from the test fixture. That will help maintainers in the future know what's important to the test and what is not.

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  convertTSConfigs(config.rootDirectory);
+
+  // 2. Remove @types/* & TypeScript dependencies + typecheck script from package.json
+  await cleanupPackageJson(config.rootDirectory);
+
+  // 3. Run codemod
+  let files = glob.sync("**/*.+(ts|tsx)", {
+    absolute: true,
+    cwd: config.rootDirectory,
+    ignore: [`./${config.appDirectory}/**/*`],
```
kentcdodds, line 29: Should we also ignore node_modules?

kentcdodds (changes requested):
This is looking good!

packages/remix-dev/__tests__/migrations/fixtures/indie-stack/.gitpod.yml
MichaelDeBoey: Removed a bunch of things.

Let me know if you want other things to be removed as well.

kentcdodds:
I think we want to remove *as much as possible* without reducing the level of confidence that the feature works. I should write a blog post about this, but it's related to what I talk about here: https://kcd.im/aha-testing. Most of those routes don't need to be there for example. Let's reduce it down to as little as possible please. Thanks!

MichaelDeBoey:
Kept the index route from the templates in `app/routes` + removed all non-JS/TS-files outside `app`

pcattori:
@MichaelDeBoey : Organizationally, this doesn't fit into `migrations` since migrations are for moving users towards compatibility with newer Remix versions and conceptually are moving users towards what we think is a better way.

But `convert-to-javascript` is more of a stand-alone utility for users who prefer JS. Maybe this is just a naming thing, but I'd like to explore a different namespace for this other than `migrations`. We can still reuse all the migration utils for it. Maybe we don't find something else better than `migrations`, but I think its worth considering.

MichaelDeBoey:
@pcattori I can't think of another name tbh.
If you don't want it to be public, we can always remove it from the list?

I wouldn't say migrations are only for updating `remix` versions though, as we can do all sort of things with it I think.

The idea came from a codemod kind of view and because we were doing more than only changing code, it made sense to rename it to `migrations`, but I don't think we should limit it to only update `remix` versions.
We could even create a migration to automate a transition from other frameworks (like CRA of Next) to Remix at one point if we ever want to have that.

kentcdodds:
I've asked @chaance to work on moving stuff out of the CLI this week (including the migrations stuff). We'll still keep the option, but we're going to move the code into a separate package (which the CLI would basically run `npx` for). The goal is to slim down the CLI package and remove everything that's not frequently used.

I think this could probably go into that package when he's gotten that set up. I'll ping Chance about this so he's aware that the package he's working on may be for more than just "migrations." It's more for "one-time-scripts" generally. Maybe it could be called `@remix-run/scripts`.

MichaelDeBoey:
Once the refactor is finished, I'm happy to rebase my PR & move files accordingly.

On the naming part: `@remix-run/scripts` sounds more like an internal package to me tbh.
Can't think of a better name than `migrations` or `scripts` though 🤔😕

mcansh (approved):
I'm not super familiar (or at all 🙈 actually) with jscodeshift, but this LGTM

kentcdodds:
I'm sorry I haven't had a chance to review this PR yet. We're going to need to hold off for a little while longer while we decide some stuff on the scripts package.

kentcdodds:
We decided to wait on the scripts package. Reviewing now.

kentcdodds (approved):
This looks good to me 👍 Thank you for working on this!

MichaelDeBoey:
@kentcdodds This migration will be internally used once #3529 is merged

Issue #1038, opened 2021-12-14
[Feature]: Rename @remix-run/architect adapter to @remix-run/aws-lambda


shamsup:
### What is the new or updated feature that you are suggesting?

1. I think it would be beneficial in the long run to rename this adapter to `@remix-run/aws-lambda` to avoid confusion. 
2. Remove the peer dependency on `@architect/architect` since it is not used

### Why should this feature be included?

1. The `@remix-run/architect` adapter is actually a generic adapter for AWS Lambda behind API Gateway. This package can be used as the adapter for projects using AWS CDK, AWS SAM, and serverless framework in addition to architect (and lambda without a framework).
2. Currently, the architect adapter has no need for a peer dependency of architect because it is not used within the package.

I think this will avoid confusion moving forward for people who want to run Remix within AWS Lambda but are using another framework to work with the service.

PR #3173, opened 2022-05-12
feat(remix-architect): support API gateway v1

wingleung:
Closes: #2696

- [ ] Docs
- [x] Tests

Testing Strategy:

added jest tests for...
- the new parts (v1-test.ts and server-test.ts)
- moved existing tests to logical location (v2-test.ts and server-test.ts)

Just a proposal to get this started. I need API Gateway V1 integration because it supports features we need that are not available in v2


mcansh:
sweet! from first glance i love it - I'll take a more in depth look in a few minutes :)

mcansh:
do you have an example repo of this? I'm not having much luck using arc.codes with it, but i could be doing something wrong.

wingleung:
> do you have an example repo of this? I'm not having much luck using arc.codes with it, but i could be doing something wrong.

Thanks for reviewing!

I was a bit behind on dev so this PR was indeed broken, my bad. I updated this PR again, I don't use architect but I've tested it with the grunge stack which uses remix-architect and I have it running here in a separate temp branch 👉  wingleung/remix/pull/1 could you try again? 

sidenote, we use sam cli as local dev tool and a custom pipeline on jenkins to deploy our stack.

packages/remix-architect/server.ts
```diff
@@ -39,99 +48,26 @@ export function createRequestHandler({
   build,
   getLoadContext,
   mode = process.env.NODE_ENV,
+  apiGatewayVersion = APIGatewayVersion.v2
 }: {
   build: ServerBuild;
   getLoadContext?: GetLoadContextFunction;
   mode?: string;
+  apiGatewayVersion?: APIGatewayVersion;
```
wingleung, line 56: this would mean we use the api like this

```javascript
import * as build from '@remix-run/dev/server-build'
import { createRequestHandler } from '@sporza/architect'

export const handler = createRequestHandler({
  build,
  mode: process.env.NODE_ENV,
  apiGatewayVersion: 'v1' // or even APIGatewayVersion.v1 if you import the enum
})
```

wingleung: @mcansh wdyt, string `v1`/`v2` in favour of numbers `1`/`2` would be in line with how AWS documents the [new API Gateway](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/AWS_ApiGatewayV2.html)

mcansh: yup i agree

MichaelDeBoey (changes requested):

packages/remix-architect/server.ts
```diff
@@ -26,10 +30,15 @@ import { isBinaryType } from "./binaryTypes";
...
  */
 export type GetLoadContextFunction = (
-  event: APIGatewayProxyEventV2
+  event: APIGatewayProxyEventV2 | APIGatewayProxyEvent
 ) => AppLoadContext;
 
-export type RequestHandler = APIGatewayProxyHandlerV2;
+export type RequestHandler = APIGatewayProxyHandlerV2 | APIGatewayProxyHandler;
+
+export enum APIGatewayVersion {
```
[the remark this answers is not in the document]
mcansh: definitely in index, not sure it's needed as a magic export though

packages/remix-architect/index.ts
```diff
@@ -4,3 +4,8 @@ export { createArcTableSessionStorage } from "./sessions/arcTableSessionStorage"
 
 export type { GetLoadContextFunction, RequestHandler } from "./server";
 export { createRequestHandler } from "./server";
+
+export enum APIGatewayVersion {
+  v1 = "v1",
+  v2 = "v2",
+}
```
MichaelDeBoey, line 11: This should be in `server` & exported in this file again

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

wingleung:
@mcansh any idea when this one can be merge? I'm using a custom build now but it's getting a little harder to keep everything up to date 😅 let me know if you want more info

mcansh:
sorry @wingleung! this keeps getting lost in my sea of tabs - gonna give it another look over and re-verify it's something we want first party support for.

I've also haven't been able to run an example thus far, but I'll try again today

ryanflorence:
Appreciate the work here. It's difficult for us to know all of the different deployment targets well and don't want to be in people's way like this. I personally barely knew enough to make this adapter in the first place and to now support two versions of API Gateway is stretching us even farther.

We had to ship with a few "first party" adapters ourselves (because who cares about a new javascript framework?) but the design of Remix is that you can make your own adapter. Long term the adapters should be maintained by the community and the hosts.

What if you made your own API Gateway adapter instead? Then more people can deploy to it w/o needing Architect?

wingleung:
@ryanflorence thanks for the insight on the long term strategy here, I definitely understand you want to focus on the core of Remix. And I have no problem splitting this into a separate community project in a leaner form without architect or move it into an `/examples` folder with other example integrations like you guys do in Remix, I'll get it up and running in the next few days and update this PR and https://github.com/remix-run/remix/discussions/2696

keep you posted!

RichiCoder1:
> @ryanflorence thanks for the insight on the long term strategy here, I definitely understand you want to focus on the core of Remix. And I have no problem splitting this into a separate community project in a leaner form without architect or move it into an `/examples` folder with other example integrations like you guys do in Remix, I'll get it up and running in the next few days and update this PR and #2696
> 
> keep you posted!

Let me know how you get along with this! I'd be happy to chip in. Def an interesting problem, especially since there's arguably _three_ relevant payload and response formats: API Gateway v1 & v2, as well as ALB. I've heard rumors that there might be a new streaming response format too, which would also justify a seperate implementation. (Plus I find it personally confusing the package is named _architect_ when it's technically compatible with Serverless, SST, or any other AWS Lambda framework.)

MichaelDeBoey:
> Plus I find it personally confusing the package is named _architect_ when it's technically compatible with Serverless, SST, or any other AWS Lambda framework.

@RichiCoder1 We have [an open issue to rename `@remix-run/architect` to `@remix-run/aws-lambda`](https://github.com/remix-run/remix/issues/1038) by @shamsup for this.

wingleung:
> Let me know how you get along with this! I'd be happy to chip in. Def an interesting problem, especially since there's arguably _three_ relevant payload and response formats: API Gateway v1 & v2, as well as ALB. I've heard rumors that there might be a new streaming response format too, which would also justify a seperate implementation. (Plus I find it personally confusing the package is named _architect_ when it's technically compatible with Serverless, SST, or any other AWS Lambda framework.)

@RichiCoder1 Haven't got around to separate it yet but idd the architect would be removed to keep the adapter lean. I think having a separate adapter per aws service would also keep things lean so in the ALB vs API Gateway scenario we would have

- `remix-aws-api-gateway`
- `remix-aws-application-load-balancer` or `remix-aws-alb``

if we would go with `remix-lambda` then I guess we would add support for all AWS proxies which might be too big of a scope.

The remix guys also have split their server adapters according to third party service for example [the cloudflare adapters](https://remix.run/docs/en/v1/other-api/adapter)

- `@remix-run/cloudflare-pages`
- `@remix-run/cloudflare-workers`

Having said that, I wonder if there is any overlap in the request en response structure that an ALB and API GW requires and if AWS will keep this overlap going further 🤔 
Looking at the [ALB docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#receive-event-from-load-balancer) there is some similarities with the [API GWv1 structure](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)

So there is an advantage if all the integration code is together in a `remix-aws-lambda` package 😄  WDYT?

RichiCoder1:
> So there is an advantage when all the integration code is together in a `remix-aws-lambda` package 😄 WDYT?

@wingleung I don't have data to back this up, but I think the events share enough commonality and don't differ wildly enough to separate in their own packages. A converged package with event detection should be more than good enough!

wingleung:
@RichiCoder1 I kickstarted [Remix AWS](https://github.com/wingleung/remix-aws) where I removed the Architect part and focused on adding support for the application load balancer.

Almost every property the ALB needs is already being used in the API Gateway v1. I've added some more info about this in the readme.

Event detection itself could be an iteration on this idd.

wingleung:
@kamtugeza raises a good point in https://github.com/wingleung/remix-aws/pull/2. the only thing we would want to move to another package would be the adapters. a `createRequestHandler` could potentially stay in `remix-core`, it might keep the integrations aligned in `remix-core`

chaance:
Housekeeping! I'm going to close this for now, as it seems like the consensus is that this should be maintained as a community package outside of Remix. Feel free to ping if you feel I missed any context and we need to reconsider.

PR #3181, opened 2022-05-13
chore(remix-dev): remove suggested `paths` from default tsconfig

mcansh:
some folks dont like path aliases and "forcing" them to have it screws up auto imports. however i do think we should keep it in the templates by default. 

an alt approach folks can do for now is set `"paths": {}` in their tsconfig

- [x] Tests


MichaelDeBoey (approved):

kentcdodds (approved):
👍

Issue #1186, opened 2021-12-22
[Bug]:  Error "process is not defined" when accessing environment variables


ronb1982:
### What version of Remix are you using?

1,1,1

### What version of Node are you using? Minimum supported version is 14.

14.16.1

### Steps to Reproduce

1. Add environment variables to .bash_profile file: (i.e. `export BASE_URL="https://test.com"
2. Open project and create a utility / helper file, and add a method like this:

In helpers.js:

```
// Note that this method is not a function component, but rather, a simple file containing helper functions
export const getBaseUrl = () => {
    return process.env.BASE_URL
}
```

RESULT:
I get the following error: "process is not defined". However, `process` is defined on the client-side.

NOTE: This same code works in both React and Next.js, but for some reason, I'm getting this error only in Remix Run.

### Expected Behavior

Env var works in helper file and no errors are thrown.

### Actual Behavior

Env var cannot be accessed - specifically, `process` is not defined.

How can I configure the Remix Run project to successfully pull environment variables that are stored on my MacOS's  .bash_profile file into my project.

As mentioned, the same code works in React and Next.js.

sergiodxa:
You can only access `process` inside the server-side code.

In Next, they are inlining them for the browser, and that's why it works.
In React, I suppose you mean CRA because React does nothing with env variables and Remix uses React, they do the same, any env is inlined in the JS code.

So both, Next and React are actually doing build variables instead of env variables since that doesn't exists in browser, open the devtools and write `process` and you will get an error.

Check this for more information on how to work with env variables in Remix: https://remix.run/docs/en/v1/guides/envvars

hrgui:
Do environment variables work for `remix watch`?

I tried following https://remix.run/docs/en/v1/guides/envvars, except for remix watch from the Cloudflare Pages starter

```
"dev:remix": "node -r dotenv/config node_modules/.bin/remix watch",
```

and I get this error:

```
Watching Remix app in development mode...
💿 Built in 166ms
Compiling worker to "/var/folders/pd/ltcn90yj4v96117p2thv__fc0000gn/T/functionsWorker.js"...
Compiled Worker successfully.
[pages:err] ReferenceError: process is not defined
    at build/index.js (/Users/hrgui/remix-cf-pages-poc/app/services/auth.ts:12:25)
Could not start Miniflare.
ERROR: "dev:wrangler" exited with 1.
```

[...]

PR #3186, opened 2022-05-13
feat(templates/cloudflare-pages): bind local `.env` variables in development

theeomm:
- Passed `--binding $(cat .env)` argument to the `dev:wrangler` script in the cloudflare-pages template as suggested [here](https://github.com/remix-run/remix/issues/1186#issuecomment-1019492493)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -330,6 +330,7 @@
 - SufianBabri
 - supachaidev
 - tascord
+- theeomm
 - TheRealAstoo
 - therealflyingcoder
 - thomasheyenbrock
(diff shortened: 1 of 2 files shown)


theeomm:
Unfortunately passing `$(cat .env)` does not work on windows. Probably this should be documented.

theeomm:
I've been running remix in WSL for all my projects previously. I only bumped into this issue a few hours ago. Once I have it figured out, I will push an update.

MichaelDeBoey:
Going to close this one we now point people to use the `.dev.vars` file in the docs (see #3804)

PR #3190, opened 2022-05-14
feat(remix-dev): import `.sql` files as text

jacob-ebey:
Adds the ability to import `.sql` files as text. This is useful when writing raw SQL instead of using an ORM.


sergiodxa (approved):

MichaelDeBoey (approved):

packages/remix-dev/compiler/loaders.ts
```diff
@@ -22,6 +22,7 @@ export const loaders: { [ext: string]: esbuild.Loader } = {
   ".ogg": "file",
   ".otf": "file",
   ".png": "file",
+  ".sql": "text",
```
sergiodxa, line 25:
```suggestion
  ".gql": "text",
```
What do you think about also adding `.gql` as text? That will let people import a `.gql` file with a GraphQL query/mutation/subscription and then use it as a variable directly in their code

jacob-ebey: Let's do a separate PR for that.

PR #3202, opened 2022-05-16
fix(remix-dev): ensure devServerPort config value is the same across reloads

tessellator:
Closes: #2958

- [x] Tests

If the devServerPort changes values across config reloads then LiveReload will attempt to connect to an incorrect port. Files being created, deleted, or renamed under the app folder trigger config reloads.

Testing Strategy:

A test is included in the commit that ensures that the same devServerPort is always returned.

Additionally, I tested a new project with this code using the following steps:

1. Create a new project using create-remix@latest and replace @remix-run/dev with a build including the code in the attached commit.
2. Run `npm run dev`
3. Browse to localhost:3000. Verify LiveReload is working by editing `app/routes/index.tsx` and observing the change.
4. Create a new file `app/components/SomeComponent.tsx` and observe page refresh. Validate no websocket errors in browser console.
5. Verify LiveReload is still working by editing `app/routes/index.tsx` and observing the change.

Issue #296, opened 2021-10-04
Cannot run `remix run` on multiple apps at once


chaance:
Because the port used by the WebSocket server is hard-coded, there's currently no way to concurrently run dev for two Remix apps at once.

https://github.com/remix-run/remix/blob/c1821e4f2c7e844f3b7462bbd6ae7711479c1c7d/packages/remix-dev/cli/commands.ts#L52

The `run` and `watch` commands would need to be updated to accept an alternative argument to set the WebSocket's server port. It would be nice if we could also pass these as command line arguments as needed.

```bash
# easily override either the port passed to @remix-run/serve or the watch command
# bikeshed the APIs, obviously!
$ remix run --port 3030 --socket-port 3002
```

Thoughts?

jamesmall:
Pilfered through the source code and found that you can set it via command line. If you update the devServerPort in the remix.config.js file in conjunction with setting REMIX_DEV_SERVER_WS_PORT via command line you can run multiple apps at once. You would also have to set the PORT to be different than the other app, of course.


```
// remix.config.js
...
module.exports = {
  appDirectory: 'app',
  browserBuildDirectory: 'public/build',
  publicPath: '/build/',
  serverBuildDirectory: 'build',
  devServerPort: [NEW WS PORT],
};


// command line
PORT=3001 REMIX_DEV_SERVER_WS_PORT=[NEW WS PORT] remix dev
```

The issue can be found in @remix-run/dev/config.js file. The config is being read but for whatever reason, the devServerPort isn't being set on appConfig causing the devServerPort to be the new port when compiled but the ws port remains 8002.

mcansh:
> Pilfered through the source code and found that you can set it via command line. If you update the devServerPort in the remix.config.js file in conjunction with setting REMIX_DEV_SERVER_WS_PORT via command line you can run multiple apps at once with live reload. You would also have to set the PORT to be different than the other app, of course.
> 
> ```
> // remix.config.js
> ...
> module.exports = {
>   appDirectory: 'app',
>   browserBuildDirectory: 'public/build',
>   publicPath: '/build/',
>   serverBuildDirectory: 'build',
>   devServerPort: [NEW WS PORT],
> };
> 
> 
> // command line
> PORT=3001 REMIX_DEV_SERVER_WS_PORT=[NEW WS PORT] remix dev
> ```
> 
[...]

PR #3206, opened 2022-05-16
fix(remix-dev): re-use `REMIX_DEV_SERVER_WS_PORT` when reloading config

### LINKED ISSUE #2958: Remix dev asset server web socket error

1.4.1

### Steps to Reproduce

1.npm i create-remix mydemo
2.remix dev
3.Modify or add files in routes
4.chrome error(Remix dev asset server web socket error:)

image

### Expected Behavior

Automatically refresh the page to match the new route.
To fix this I had to reboot.

### Actual Behavior

image

mcansh:
closes REM-1091

closes #2958

- [ ] Docs
- [x] Tests

Testing Strategy:

I opened up a new playground (`npm run playground:new`) with a basic node server running on port 8002 and then created a few new files to trigger the config reload while logging the current `REMIX_DEV_SERVER_WS_PORT`.
image

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -353,7 +353,7 @@ export async function readConfig(
       path.join("public", "build")
   );
 
-  let devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT) || await getPort({ port: Number(appConfig.devServerPort) || null });
+  let devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT) || await getPort({ port: Number(appConfig.devServerPort) || undefined });
   // set env variable so un-bundled servers can use it
   process.env.REMIX_DEV_SERVER_WS_PORT = `${devServerPort}`;
   let devServerBroadcastDelay = appConfig.devServerBroadcastDelay || 0;


tessellator:
I think this can still get you in trouble wrt #296. If you set `REMIX_DEV_SERVER_WS_PORT` to the same value in two different projects, you would lose the advantage of using `getPort`. I will test to verify. I provided a similar solution in #3202 that will work in two projects simultaneously.

tessellator:
Confirmed, but on second thought this behavior might be desirable. This PR makes `REMIX_DEV_SERVER_WS_PORT` explicit and will fail fast if the port is already taken, whereas the other PR is a "best effort" and will rewrite `REMIX_DEV_SERVER_WS_PORT` rewritten if the port is already taken.

jacob-ebey (approved):

packages/remix-dev/config.ts
```diff
@@ -359,6 +359,10 @@ export async function readConfig(
     );
   }
   let devServerPort = await getPort({ port: appConfig.devServerPort });
+  // check if the port is already in use (e.g this file is reloaded after a dep install or config change)
+  if (process.env.REMIX_DEV_SERVER_WS_PORT) {
+    devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT || "8002");
+  }
```
kentcdodds, line 365: Doing things this way avoids the unnecessary `getPort` call:

```suggestion
  let devServerPort;
  // check if the port is already in use (e.g this file is reloaded after a dep install or config change)
  if (process.env.REMIX_DEV_SERVER_WS_PORT) {
    devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT || "8002");
  } else {
    devServerPort = await getPort({ port: appConfig.devServerPort })
  }
```

kentcdodds (changes requested):
Thanks! What do you think about this?

packages/remix-dev/config.ts
tessellator: Is the `|| "8002"` necessary? `REMIX_DEV_SERVER_WS_PORT` has already been determined to be truthy.

packages/remix-dev/config.ts
```diff
@@ -353,12 +353,19 @@ export async function readConfig(
...
-    appConfig.devServerPort = Number(
-      process.env.REMIX_DEV_SERVER_WS_PORT || 8002
-    );
+  let devServerPort: number;
+
+  // check if the port is already in use as this file is reloaded after a dep install or config change
+  if (process.env.REMIX_DEV_SERVER_WS_PORT) {
+    try {
+      devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT);
+    } catch (error: unknown) {
```
tessellator, line 362: I'm unsure if `Number` ever throws an error. In most cases it will return `NaN` instead. 

In this part of the code, it would mean that someone explicitly set the env var to an invalid value. If it wasn't set explicitly, then it would either 1) fall through to `getPort` on the initial load, or 2) have been set to the output of `getPort` on a previous load. 

Is `getPort` as default the right thing to do if someone explicitly sets the env var to an invalid value? (Genuinely curious as to the philosophy of the Remix team.)

mcansh: yup, you're right on the NaN, Mondays amirite 😅. i'll wait for @kentcdodds' thoughts on the default behavior, i could see throwing an error as the desired behavior

tessellator: Awesome. :) Thanks so much for working on this. I'm excited to see LiveReload working again!

packages/remix-dev/config.ts
```diff
@@ -353,12 +353,16 @@ export async function readConfig(
...
-    );
+  let devServerPort: number;
+
+  // check if the port is already in use as this file is reloaded after a dep install or config change
+  if (process.env.REMIX_DEV_SERVER_WS_PORT) {
+    devServerPort =
+      Number(process.env.REMIX_DEV_SERVER_WS_PORT) ||
+      (await getPort({ port: appConfig.devServerPort }));
+  } else {
+    devServerPort = await getPort({ port: appConfig.devServerPort });
```
kentcdodds, line 364: How about this:

```suggestion
  let devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT) || await getPort({ port: Number(appConfig.devServerPort) || null });
```

Only thing is it wouldn't work with a port of `0`, but I don't think that's a concern... We can't use `??` because `NaN`.

This also handles the case of `devServerPort` not being configured properly.

kentcdodds (changes requested):
Another thought

kentcdodds (approved):
Sweet! Thanks 👏👏

packages/remix-dev/config.ts
```diff
@@ -353,12 +353,8 @@ export async function readConfig(
       path.join("public", "build")
   );
 
-  if (typeof appConfig.devServerPort !== "number") {
-    appConfig.devServerPort = Number(
-      process.env.REMIX_DEV_SERVER_WS_PORT || 8002
-    );
+  let devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT) || await getPort({ port: Number(appConfig.devServerPort) || null });
   }
```
kentcdodds, line 357:
```suggestion
```

kentcdodds (approved):
Whoops 😅 Ok, should be good now!

packages/remix-dev/config.ts
```diff
@@ -353,12 +353,7 @@ export async function readConfig(
       path.join("public", "build")
   );
 
-  if (typeof appConfig.devServerPort !== "number") {
-    appConfig.devServerPort = Number(
-      process.env.REMIX_DEV_SERVER_WS_PORT || 8002
-    );
-  }
-  let devServerPort = await getPort({ port: appConfig.devServerPort });
+  let devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT) || await getPort({ port: Number(appConfig.devServerPort) || null });
```
kentcdodds, line 356:
```suggestion
  let devServerPort = Number(process.env.REMIX_DEV_SERVER_WS_PORT) || await getPort({ port: Number(appConfig.devServerPort) || undefined });
```

kentcdodds (approved):
I really shouldn't use our CI as my development environment 😅

EkaanshArora:
Hi team,

This breaks out of the box hot-reload with some of the templates (eg: [just the basics +  express](https://github.com/remix-run/remix/issues/3237)) as the used port is different from the expected `8002` default port for these configurations. (#3237)

This is because the new code gets a random available port if the environment variable isn't set - it's undefined for the templates as `remix.config.js` does not include a `devServerPort`.

Should the templates be updated to have `"devServerPort": 8002` in the `remix.config.js`?

kentcdodds:
@EkaanshArora, thanks. @mcansh is going to look into this.

danditomaso:
Any updates on this issue?

Issue #3003, opened 2022-04-27
TypeError: Body is unusable when using Remix(experimental-netlify-edge) Actions


nickytonline:
### What version of Remix are you using?

0.0.0-experimental-fd9fa7f4

### Steps to Reproduce

1. Use Node 16.x or 18.x, the result is the same.
2. Ensure that you have the [Netlify CLI installed](https://docs.netlify.com/cli/get-started/).
3. Run `npx create-remix --template https://github.com/netlify/remix-edge-template`. Use the default answers when the Remix CLI asks you questions. Note: You can create a TypeScript project (default) or a JavaScript project. It doesn't matter, the result is the same in regards to the error.
4. Modify the file `/app/routes/index.js` with the following code:

```jsx
import type { LoaderFunction, ActionFunction } from "@remix-run/server-runtime"
import { useLoaderData } from "@remix-run/react";

let projectName: string | null = null;

export const loader: LoaderFunction = async () => {
  return { name: projectName ?? 'Create a Project Name' };
};

export const action: ActionFunction = async({ request }) => {
  const body = await request.formData();

  console.log('body',body);
  projectName = body.get("projectName") as string;
  console.log('projectName pulled from form data', projectName)
  console.log(`request body is used ${request.bodyUsed}`)

  console.log('action ran')

  return null
}

export default function Index() {
  let project = useLoaderData();

  return (
    <div style={{ fontFamily: "system-ui, sans-serif", lineHeight: "1.4" }}>
      <h1>Welcome to Remix {project.name}</h1>
      <ul>
        <li>
          <a
            target="_blank"
            href="https://remix.run/tutorials/blog"
            rel="noreferrer"
          >
            15m Quickstart Blog Tutorial
          </a>
        </li>
        <li>
          <a
            target="_blank"
            href="https://remix.run/tutorials/jokes"
            rel="noreferrer"
          >
            Deep Dive Jokes App Tutorial
          </a>
        </li>
        <li>
          <a target="_blank" href="https://remix.run/docs" rel="noreferrer">
            Remix Docs
          </a>
        </li>
      </ul>
    </div>
  );
}
```

6. run `ntl dev` to start the application in development mode.
7. Things build without any failures.
8. Navigate to http://localhost:3000 and the Welcome to Remix page loads.

![CleanShot 2022-04-27 at 14 15 40](https://user-images.githubusercontent.com/833231/165592731-c5916704-bd55-4594-9b82-fbcccd743fda.png)

[...]

PR #3207, opened 2022-05-16
feat(remix-server-runtime): strip `body` from `request` before calling `loader`s

jacob-ebey:
chore: remove duplicate method definition

Since loaders map to "GET" and "HEAD" requests, those by definition can't have a body and therefore there is no reason to pass it in to use-code. This also avoids exclusive locks on the body when multiple loaders are executed in parallel.

Changes:
- Action recieves OG request and can choose to read the body.
- handleDocumentRequest recieves OG request, if the action read the body it will be locked and `bodyUsed` will be true. Atempting to read the body again will throw
- Loaders recieve a copy of the OG request without the body.

Closes: https://github.com/remix-run/remix/issues/3003

- [ ] Docs
- [ ] Tests

### DIFF
--- a/docs/decisions/0002-do-not-clone-request.md
+++ b/docs/decisions/0002-do-not-clone-request.md
@@ -6,14 +6,14 @@ Status: accepted
 
 ## Context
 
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (approved):

chaance:
> I would love to but none of this is broken on node

👍 

> we don't have a way to easily run integration tests against other runtimes

😭

mjackson (approved):

PR #3208, opened 2022-05-16
chore: remove unused dependencies from root `package.json`

MichaelDeBoey:
They're either not used or already in the package where it's used


kentcdodds (approved):
Thanks!

Issue #3133, opened 2022-05-08
`Form` post submission does not retain search parameters


TheRealFlyingCoder:
### What version of Remix are you using?

1.4.3

### Steps to Reproduce

Impement a route that includes search params in the url.

`/my-route?param=1234`

On the page include a post form

```
<Form method="post">
        <button type="submit>Submit</button>
</Form>
```

Native browser behaviour can be seen here thanks to @kiliman

https://codesandbox.io/s/form-action-params-8r86qw?file=/app/routes/test.tsx:732-987

### Expected Behavior

Following the native browser behaviour, this should post back to `/my-route?param=1234`

Note that a `method="get"` post would natively replace the search params by the nature of the submission method

### Actual Behavior

The form will post to the base path `/my-route` and trim off the search params

kiliman:
@manV Yes, fixing this issue will solve your scenario.

You can "fix" it now, by manually specifying your action.
```ts
const location = useLocation()
return <Form action={`${location.pathname}${location.search}`}/>
```
The actual fix is to essentially do this automatically if action is not specified, as that is standard browser behavior.

nivekithan:
I think problem lies in `useFormAction`, 

https://github.com/remix-run/react-router/blob/931cf23a54373485607b3727dbd5f9575fe58ebe/packages/react-router/lib/hooks.tsx#L244-L256

As of now code looks like this 
```typescript
export function useFormAction(
  action = ".",
  // TODO: Remove method param in v2 as it's no longer needed and is a breaking change
  method: FormMethod = "get"
): string {
  let { id } = useRemixRouteContext();
  let path = useResolvedPath(action);
  let search = path.search;
  let isIndexRoute = id.endsWith("/index");

  if (action === "." && isIndexRoute) {
    search = search ? search.replace(/^\?/, "?index&") : "?index";
  }

  return path.pathname + search;
}
```

According to docs for `useResolvedPath` from  https://reactrouter.com/docs/en/v6/hooks/use-resolved-path

```
This hook resolves the pathname of the location in the given to value against the pathname of the current location.
```

I think what that means is that only resolves only `pathName` and  ignores the `search` key. Therefore returned value from `useResolvedPath` looks like this

```ts
{hash: ""
pathname: "/login"
search: ""}
```

even when the `url` is `/login?redirectTo=%2Fsettings`. I am not sure whether it is intended behaviour or not.

At commit  a3ae536f25bbb1d0cf93d06082b04741c8edae34 

[...]

PR #3211, opened 2022-05-17
tests(remix-react): add failing test for default `Form` action

ionut-botizan:
Issue: #3133

The default `<Form />` action should match the current location. Currently, only the `path` is preserved, but the `search` parameters are lost.

E.g. For the `/login?redirectTo=/profile` location and a `<Form method="post" />` element, the generated form node looks like
`<form method="post" action="/login">`.

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -1,6 +1,6 @@
 import { test, expect } from "@playwright/test";
 
-import { getElement, PlaywrightFixture } from "./helpers/playwright-fixture";
+import { PlaywrightFixture } from "./helpers/playwright-fixture";
 import type { Fixture, AppFixture } from "./helpers/create-fixture";
 import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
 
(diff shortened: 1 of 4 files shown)


integration/bug-report-test.ts
```diff
@@ -47,30 +47,19 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/index.jsx": js`
```
chaance, line 50: This test file is used as a template. Please move this test to `integration/form-test.ts`

packages/remix-react/components.tsx
```diff
@@ -961,9 +961,14 @@ export function useFormAction(
 ): string {
   let { id } = useRemixRouteContext();
   let path = useResolvedPath(action);
+  let location = useLocation();
   let search = path.search;
   let isIndexRoute = id.endsWith("/index");
 
+  if (action === ".") {
```
chaance, line 968: This should check if `action === undefined` instead. If the action is explicitly set to `.` it will be resolved to strip the search, which is the expected behavior.

We only want to use the location's search if the action is left undefined in the Form. This means we need to remove the default value of the action in `FormImpl` as well.

See https://github.com/remix-run/remix/issues/3133#issuecomment-1128776268

chaance (changes requested):
A few changes needed before we can merge. If an action value of `.` is provided, it is correct to strip the search params when resolving the route path. We only need to preserve the location's search if no action is provided.

integration/bug-report-test.ts
ionut-botizan: Oopsie! 😊

packages/remix-react/components.tsx
ionut-botizan: Got it!

maranomynet:
Please someone resolve and merge this PR.

Always having to do this feels so daft.

```ts
const { pathname, search } = useLocation();

return <Form action={pathname+search}
```

ionut-botizan:
Hey guys! Are there any other changes or additions that you would like me to provide? I think I did everything that was asked in the initial review...
I don't mean to bother you. I realize there are more important issues out there I was just thinking that maybe you missed the fact that I responded to the review. 🙂

chaance:
Hi @ionut-botizan, sorry for the radio silence on this one. We actually have duplicated efforts a bit on this bug. We are building Remix's form components and hooks into React Router, so @brophdawg11 and I have been working in tandem on that separately. I think I must have forgotten about your work and started my own over in https://github.com/remix-run/remix/pull/3697. Very sorry about that. 

To clean things up, I'm going to close this one. But I added you as a co-committer for [`783c435` (#3697)](https://github.com/remix-run/remix/pull/3697/commits/783c43573f37bb89e1cf25be5ec87b72aee5186f) and made sure your handle is included in the contributor list. I definitely want to make sure you're credited with identifying the problem and proposing the initial fix!

PR #3215, opened 2022-05-17
feat(remix-deno): export new `@remix-run/server-runtime` upload functions & types

MichaelDeBoey:
This PR brings feature parity with the Cloudflare & Node runtimes

### DIFF
--- a/packages/remix-deno/index.ts
+++ b/packages/remix-deno/index.ts
@@ -18,7 +18,11 @@ export {
   isCookie,
   isSession,
   json,
+  MaxPartSizeExceededError,
   redirect,
+  unstable_composeUploadHandlers,
+  unstable_createMemoryUploadHandler,
+  unstable_parseMultipartFormData,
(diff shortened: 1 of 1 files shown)


pcattori:
This raises two thoughts:
1. If we need to make sure we export certain things, we should have a test for that.
2. I personally think we should move towards splitting `server-runtime` pkg into two packages: one that provides utils for making server runtime packages (e.g. `createCookieFactory` or other factory functions) and another one that provides shared, runtime-agnostic implementations (e.g. `json` or the implementations added to the exports in this PR).

If we go with (2), then (1) is obviated, but that requires some design thinking of the implications. Let me chat with the team and get back to you on this.

pcattori (approved):
I'm going to merge this as-is and will track writing tests for this as a separate task. In the meantime, I'll start the conversation around splitting up `server-runtime` into two packages, but don't want to wait for that decision to get this improvement in.

PR #3216, opened 2022-05-17
fix(templates/vercel): add `index.js.map` to `.gitignore`

jiahao-c:
In the Vercel template,  `/api/index.js.map` should also be ignored along with `/api/index.js`. Otherwise, that mapping file would be committed into version control, which is not desired.

This PR only touches `/templates`, therefore requesting merge at `main` instead of `dev` branch.

Testing Strategy: `git commit`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -155,6 +155,7 @@
 - jesse-deboer
 - jesseflorig
 - jgarrow
+- jiahao-c
 - jkup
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (approved):

kentcdodds (approved):
Thanks!

PR #3221, opened 2022-05-18
chore(templates): bring `deno` template in line with other templates

MichaelDeBoey:


templates/deno/.vscode/resolve_npm_imports.json
```diff
@@ -5,11 +5,11 @@
...
-    "react": "https://esm.sh/react@18.0.0",
-    "react-dom": "https://esm.sh/react-dom@18.0.0",
-    "react-dom/server": "https://esm.sh/react-dom@18.0.0/server",
-    "@remix-run/dev/server-build": "https://esm.sh/@remix-run/dev@1.4.3/server-build",
+    "react": "https://esm.sh/react@17.0.2",
+    "react-dom": "https://esm.sh/react-dom@17.0.2",
+    "react-dom/server": "https://esm.sh/react-dom@17.0.2/server",
     "@remix-run/deno": "https://esm.sh/@remix-run/deno@1.4.3",
+    "@remix-run/dev/server-build": "https://esm.sh/@remix-run/dev@1.4.3/server-build",
     "@remix-run/react": "https://esm.sh/@remix-run/react@1.4.3"
```
pcattori, line 13: We should update the `@remix-run/*` versions to `1.5.0` so that it references a version where `@remix-run/deno` will be available

templates/deno/app/entry.server.tsx
```diff
@@ -1,16 +1,16 @@
...
 export default function handleRequest(
   request: Request,
   responseStatusCode: number,
   responseHeaders: Headers,
-  remixContext: EntryContext,
+  remixContext: EntryContext
 ) {
   const markup = renderToString(
-    <RemixServer context={remixContext} url={request.url} />,
+    <RemixServer context={remixContext} url={request.url} />
```
pcattori, line 13: We use `deno lint` and `deno fmt` for Deno code, so we should not remove the trailing `,`

templates/deno/package.json
```diff
@@ -1,15 +1,17 @@
 {
   "name": "remix-template-deno",
   "private": true,
+  "description": "",
+  "license": "",
```
pcattori, line 5: Why are `description` and `license` added here?

templates/deno/remix.config.js
```diff
@@ -1,12 +1,13 @@
...
-
-  /*
-  If live reload causes page to re-render without changes (live reload is too fast),
-  increase the dev server broadcast delay.
-
-  If live reload seems slow, try to decrease the dev server broadcast delay.
-  */
+  // If live reload causes page to re-render without changes (live reload is too fast),
+  // increase the dev server broadcast delay.
+  // If live reload seems slow, try to decrease the dev server broadcast delay.
```
pcattori, line 6: I prefer a multi-line comment here. Could we keep as-is?

templates/deno/remix.config.js
```diff
@@ -1,12 +1,13 @@
...
   devServerBroadcastDelay: 300,
+  ignoredRouteFiles: ["**/.*"],
+  // appDirectory: "app",
+  // assetsBuildDirectory: "public/build",
+  // serverBuildPath: "build/index.js",
+  // publicPath: "/build/",
```
pcattori, line 12: Can you include a link to the Remix config docs instead of commenting out defaults?

templates/deno/remix.config.js
```diff
@@ -1,12 +1,13 @@
...
-  If live reload causes page to re-render without changes (live reload is too fast),
-  increase the dev server broadcast delay.
-
-  If live reload seems slow, try to decrease the dev server broadcast delay.
-  */
+  // If live reload causes page to re-render without changes (live reload is too fast),
+  // increase the dev server broadcast delay.
+  // If live reload seems slow, try to decrease the dev server broadcast delay.
   devServerBroadcastDelay: 300,
+  ignoredRouteFiles: ["**/.*"],
```
pcattori, line 8: Why is `ignoredRouteFiles` added?

templates/deno/app/entry.client.tsx
```diff
@@ -1,5 +1,5 @@
-import React from "react";
-import ReactDOM from "react-dom";
 import { RemixBrowser } from "@remix-run/react";
+import * as React from "react";
```
pcattori, line 2: Why is `* as React` used here instead of just `React`?

pcattori (changes requested):
Most of these changes are preferences (like import ordering and such) that don't affect the correctness, readability, or maintainability of the user's code.

I'd prefer if this PR just included:
- updates to the version specs in the import map
- link to the Remix config docs in `remix.config.js`
- importing `hydrate` as a named export in `entry.client.tsx`

templates/deno/.vscode/resolve_npm_imports.json
MichaelDeBoey: Probably a good idea to update this each time we release

templates/deno/remix.config.js
MichaelDeBoey: We have these in all other templates as well

templates/deno/app/entry.client.tsx
MichaelDeBoey: This is the recommended way of importing `react`

templates/deno/remix.config.js
MichaelDeBoey: We have these in all other templates as well

templates/deno/.vscode/resolve_npm_imports.json
MichaelDeBoey: This is implemented in #3225

templates/deno/package.json
pcattori: They aren't used at all, so they should be omitted.

templates/deno/remix.config.js
pcattori: I'd prefer if we linked to the Remix config docs across all of our templates. I don't think there's value in having consistency in commented out defaults across templates as long as the information is accessible. Plus the docs would explain why the defaults are what they are, and would also show all the other possible options.

Could you replace this with a link to the Remix config docs?

templates/deno/remix.config.js
pcattori: I don't see the value in making comment style consistent. If I had to pick one, I would argue that multi-line comments are better for comments spanning multiple lines (e.g. they can be edited without needing to add or remove intermediate `//`).

Can you keep this as-is?

templates/deno/remix.config.js
pcattori: I don't think we should aim to make this consistent across templates. It's not needed by this template. If it's common enough that we'd want it in every template, we should just make it the default value instead of including it here.

templates/deno/app/entry.client.tsx
pcattori: Is that recommendation from the React docs? Or from us? or elsewhere?

templates/deno/app/entry.server.tsx
pcattori: Good to know! I agree that it'd be nice to ignore `prettier` here. In any case, these changes shouldn't be included in the PR.

templates/deno/package.json
MichaelDeBoey: That should be done in a separate PR imo.

Let's merge these changes so all templates are consistent & we can remove it again in another PR if we want to.

templates/deno/remix.config.js
MichaelDeBoey: Don't have a strong opinion, just wanted to make it consistent.

templates/deno/app/entry.server.tsx
MichaelDeBoey: Seems like it's not ignored by default because we use the root `.eslintignore` file as ignore file in our `format` script.

templates/deno/remix.config.js
MichaelDeBoey: That should be done in a separate PR imo.

Let's merge these changes so all templates are consistent & we can remove it again in another PR if we want to.

templates/deno/remix.config.js
MichaelDeBoey: All templates have set `ignoredRouteFiles` like this, so they all have a sensible default value.
People could still choose to remove it if they want to.

Same reason those other commented values are there.

If we want to remove it, I think it should be done for all templates consistently.

So let's merge these changes so all templates are consistent & we can remove it again in another PR if we want to.

pcattori: I have similar thoughts about this. If we can directly enunciate the purpose of setting `ignoredRouteFiles`, I'm ok to keep it. But I don't want to include it just because other templates do.

Let's instead keep as-is and make another PR to fix in other templates.

templates/deno/app/entry.client.tsx
MichaelDeBoey: The React team is recommending it since v17 I think.
They even have a codemod for it if I'm not mistaken + there's an issue about deprecating default export in their repo.

Don't have the time to search for them currently, but I know I @kentcdodds has updated these imports in his courses in the past.

pcattori: Ok yes I recall that using `* as React` means that `tsconfig.json` doesn't need to be configured with `allowSyntheticDefaultImports`. Agreed, these should be `* as React` 👍

templates/deno/package.json
MichaelDeBoey: They're not used by us, but people will use them in their apps usually.

templates/deno/remix.config.js
MichaelDeBoey: The value we set in `ignoredRouteFiles` by default is to comply to the case where `.DS_Store` was causing errors.
That's why we set all dot files to be ignored as the sensible default for an app, as most people will want that.

templates/deno/package.json
pcattori: `description`, `license` and similar fields in `package.json` are designed for providing package metadata when that package is installed as a dependency. They are used at all for apps.

MichaelDeBoey: You could say the same about `private` & `sideEffects` I think?
And these are also in all `package.json`.

I'll revert these & we can discuss further in another PR to remove them in all other templates.

pcattori: `private` is used by `npm`, `yarn` and other tools so that they know this package isn't a library that will be published to NPM, so it is appropriate for apps to set `private: true`.

I think you're right about `sideEffects`. I believe that's not used if the package is actually an app.

templates/deno/remix.config.js
pcattori: Ok I buy that. Let's keep your `ignoredRouteFiles` changes in the PR.

MichaelDeBoey: Than I think we can merge this PR.

templates/deno/remix.config.js
```diff
@@ -1,12 +1,16 @@
...
-  increase the dev server broadcast delay.
-
-  If live reload seems slow, try to decrease the dev server broadcast delay.
-  */
+  /**
+   * If live reload causes page to re-render without changes (live reload is too fast),
+   * increase the dev server broadcast delay.
+   *
+   * If live reload seems slow, try to decrease the dev server broadcast delay.
+   */
```
pcattori, line 9: nit: using jsdoc-style comment is overkill and runs into the same problem as multiple `//` in that editing the comment requires messing with comment indicators on each line. Multi-line comment should be fine (`/*` and `*/`)

pcattori (approved):

pcattori:
@MichaelDeBoey we probably want to target `main` now since 1.5.0 got released and this is a template change

pcattori (approved):

PR #3224, opened 2022-05-18
feat(remix-server-runtime): make `Cookie` & `CreateCookieFunction` more type-safe

mikeybinns:
- [ ] Docs
- [ ] Tests

Testing Strategy:

I'm not really sure about how to automate testing this because it only adds an optional type. `npm run lint` and `yarn test` run with no issues with my code. I don't see a way to run a specific type check.

I can use Typescript Intellisense in VS Code to see that the type is correctly applied when it is specified using the generic parameter for createCookie and the type is set to any when it is not (current behaviour).

e.g.
```ts
let cookie = createCookie(); 
let variable = cookie.parse(); // variable type = any. Current behaviour. 
```
```ts
let cookie = createCookie<string>(); 
let variable = cookie.parse(); // variable type = string
```
```ts
let cookie = createCookie<{some_value: string}>(); 
let variable = cookie.parse(); // variable type = {some_value: string}
```

fix note:
Adding type safety to the serialise function produced an error in the function. 
`This condition will always return 'false' since the types 'Type' and 'string' have no overlap.`
I fixed it by adding a type assertion to the value to make it read as "any". I don't know why the error occurred, to me it looks like a shortcoming with Typescript rather than an actual bug, but I think this should be double checked.

### DIFF
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -32,7 +32,7 @@ export type CookieOptions = CookieParseOptions &
  *
  * @see https://remix.run/api/remix#cookie-api
  */
-export interface Cookie<Value extends any = any> {
+export interface Cookie<Value extends unknown = unknown> {
   /**
    * The name of the cookie, used in the `Cookie` and `Set-Cookie` headers.
    */
@@ -64,10 +64,13 @@ export interface Cookie<Value extends any = any> {
    * Serializes the given value to a string and returns the `Set-Cookie`
    * header.
    */
-  serialize(value: Value, options?: CookieSerializeOptions): Promise<string>;
+  serialize(
+    value: Value | null,
+    options?: CookieSerializeOptions
+  ): Promise<string>;
 }
 
-export type CreateCookieFunction = <Value extends any = any>(
+export type CreateCookieFunction = <Value extends unknown = unknown>(
   name: string,
   cookieOptions?: CookieOptions
 ) => Cookie<Value>;
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -32,7 +32,7 @@ export type CookieOptions = CookieParseOptions &
  *
  * @see https://remix.run/api/remix#cookie-api
  */
-export interface Cookie<Type = any> {
+export interface Cookie<Value = any> {
   /**
    * The name of the cookie, used in the `Cookie` and `Set-Cookie` headers.
    */
@@ -58,19 +58,19 @@ export interface Cookie<Type = any> {
   parse(
     cookieHeader: string | null,
     options?: CookieParseOptions
-  ): Promise<Type | null>;
+  ): Promise<Value | null>;
 
   /**
    * Serializes the given value to a string and returns the `Set-Cookie`
    * header.
    */
-  serialize(value: Type, options?: CookieSerializeOptions): Promise<string>;
+  serialize(value: Value, options?: CookieSerializeOptions): Promise<string>;
 }
 
-export type CreateCookieFunction = <Type = any>(
+export type CreateCookieFunction = <Value = any>(
   name: string,
   cookieOptions?: CookieOptions
-) => Cookie<Type>;
+) => Cookie<Value>;
 
 /**
  * Creates a logical container for managing a browser cookie from the server.
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -120,7 +120,9 @@ export const createCookieFactory =
       async serialize(value, serializeOptions) {
         return serialize(
           name,
-          value === "" ? "" : await encodeCookieValue(sign, value, secrets),
+          (value as any) === ""
+            ? ""
+            : await encodeCookieValue(sign, value, secrets),
           {
             ...options,
             ...serializeOptions,
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -32,7 +32,7 @@ export type CookieOptions = CookieParseOptions &
  *
  * @see https://remix.run/api/remix#cookie-api
  */
-export interface Cookie<Value extends unknown = unknown> {
+export interface Cookie<Value extends any = any> {
   /**
    * The name of the cookie, used in the `Cookie` and `Set-Cookie` headers.
    */
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -32,7 +32,7 @@ export type CookieOptions = CookieParseOptions &
  *
  * @see https://remix.run/api/remix#cookie-api
  */
-export interface Cookie<Value extends unknown = unknown> {
+export interface Cookie<Value extends any = any> {
   /**
    * The name of the cookie, used in the `Cookie` and `Set-Cookie` headers.
    */
@@ -70,7 +70,7 @@ export interface Cookie<Value extends unknown = unknown> {
   ): Promise<string>;
 }
 
-export type CreateCookieFunction = <Value extends unknown = unknown>(
+export type CreateCookieFunction = <Value extends any = any>(
   name: string,
   cookieOptions?: CookieOptions
 ) => Cookie<Value>;
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -67,7 +67,7 @@ export interface Cookie<Value extends any = any> {
   serialize(value: Value, options?: CookieSerializeOptions): Promise<string>;
 }
 
-export type CreateCookieFunction = <Value extends unknown = unknown>(
+export type CreateCookieFunction = <Value extends any = any>(
   name: string,
   cookieOptions?: CookieOptions
 ) => Cookie<Value>;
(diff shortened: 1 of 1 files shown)


packages/remix-server-runtime/cookies.ts
```diff
@@ -57,19 +57,19 @@ export interface Cookie {
...
     options?: CookieParseOptions
-  ): Promise<any>;
+  ): Promise<Type | null>;
 
   /**
    * Serializes the given value to a string and returns the `Set-Cookie`
    * header.
    */
-  serialize(value: any, options?: CookieSerializeOptions): Promise<string>;
+  serialize(value: Type, options?: CookieSerializeOptions): Promise<string>;
```
MichaelDeBoey, line 66: Probably a good idea to rename `Type` to `Value`

```suggestion
  serialize(value: Value, options?: CookieSerializeOptions): Promise<string>;
```

packages/remix-server-runtime/cookies.ts
```diff
@@ -57,19 +57,19 @@ export interface Cookie {
   parse(
     cookieHeader: string | null,
     options?: CookieParseOptions
-  ): Promise<any>;
+  ): Promise<Type | null>;
```
MichaelDeBoey, line 60:
```suggestion
  ): Promise<Value | null>;
```

packages/remix-server-runtime/cookies.ts
```diff
@@ -31,7 +31,7 @@ export type CookieOptions = CookieParseOptions &
  *
  * @see https://remix.run/api/remix#cookie-api
  */
-export interface Cookie {
+export interface Cookie<Type = any> {
```
MichaelDeBoey, line 34:
```suggestion
export interface Cookie<Value = any> {
```

packages/remix-server-runtime/cookies.ts
```diff
@@ -57,19 +57,19 @@ export interface Cookie {
...
   /**
    * Serializes the given value to a string and returns the `Set-Cookie`
    * header.
    */
-  serialize(value: any, options?: CookieSerializeOptions): Promise<string>;
+  serialize(value: Type, options?: CookieSerializeOptions): Promise<string>;
 }
 
-export type CreateCookieFunction = (
+export type CreateCookieFunction = <Type = any>(
```
MichaelDeBoey, line 69:
```suggestion
export type CreateCookieFunction = <Value = any>(
```

packages/remix-server-runtime/cookies.ts
```diff
@@ -57,19 +57,19 @@ export interface Cookie {
...
   name: string,
   cookieOptions?: CookieOptions
-) => Cookie;
+) => Cookie<Type>;
```
MichaelDeBoey, line 72:
```suggestion
) => Cookie<Value>;
```

MichaelDeBoey (changes requested):

packages/remix-server-runtime/cookies.ts
```diff
@@ -116,7 +116,9 @@ export const createCookieFactory =
       async serialize(value, serializeOptions) {
         return serialize(
           name,
-          value === "" ? "" : await encodeCookieValue(sign, value, secrets),
+          (value as any) === ""
+            ? ""
+            : await encodeCookieValue(sign, value, secrets),
```
MichaelDeBoey, line 121: This can be reverted

```suggestion
          value === "" ? "" : await encodeCookieValue(sign, value, secrets),
```

packages/remix-server-runtime/cookies.ts
```diff
@@ -31,7 +31,7 @@ export type CookieOptions = CookieParseOptions &
  *
  * @see https://remix.run/api/remix#cookie-api
  */
-export interface Cookie {
+export interface Cookie<Value = any> {
```
MichaelDeBoey, line 34: In this case it doesn't really matter, but it's a good practice to say what type it should extend.

```suggestion
export interface Cookie<Value extends any = any> {
```

MichaelDeBoey (changes requested):

mikeybinnswebdesign:
@MichaelDeBoey I don't know if I need to request a review after I resolve these changes or if you get notified anyway, let me know if it's spamming you and I'll stop 👍

packages/remix-server-runtime/cookies.ts
```diff
@@ -57,19 +57,19 @@ export interface Cookie {
...
   /**
    * Serializes the given value to a string and returns the `Set-Cookie`
    * header.
    */
-  serialize(value: any, options?: CookieSerializeOptions): Promise<string>;
+  serialize(value: Value, options?: CookieSerializeOptions): Promise<string>;
 }
 
-export type CreateCookieFunction = (
+export type CreateCookieFunction = <Value extends any = any>(
```
MichaelDeBoey, line 69:
```suggestion
export type CreateCookieFunction = <Value extends unknown = unknown>(
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

packages/remix-server-runtime/cookies.ts
MichaelDeBoey: This is not the problem now

MichaelDeBoey: The problem is within `sessions.ts` now
https://github.com/remix-run/remix/runs/6507483095?check_suite_focus=true#step:6:125

packages/remix-server-runtime/cookies.ts
```diff
@@ -32,7 +32,7 @@ export type CookieOptions = CookieParseOptions &
  *
  * @see https://remix.run/api/remix#cookie-api
  */
-export interface Cookie {
+export interface Cookie<Value extends unknown = unknown> {
```
MichaelDeBoey, line 35: Let's keep `any` (for now), as it's causing some other type errors, which would bring us really far on this.
https://github.com/remix-run/remix/runs/6507483095?check_suite_focus=true#step:6:125

```suggestion
export interface Cookie<Value extends any = any> {
```

MichaelDeBoey (approved):

mikeybinnswebdesign:
@MichaelDeBoey Should the serialise function also allow null? This way, you can serialise null as a form of deleting the cookie, because parsing the cookie would return null, same as if it didn't exist.
I don't know if there's a better way to handle clearing / deleting a specific cookie. there's no Delete-Cookie header to do so. I can only think of setting an expires that's in the past, but that's not exactly clean.

packages/remix-server-runtime/cookies.ts
```diff
@@ -58,19 +58,22 @@ export interface Cookie {
...
 
   /**
    * Serializes the given value to a string and returns the `Set-Cookie`
    * header.
    */
-  serialize(value: any, options?: CookieSerializeOptions): Promise<string>;
+  serialize(
+    value: Value | null,
+    options?: CookieSerializeOptions
+  ): Promise<string>;
```
MichaelDeBoey, line 70: @mikeybinnswebdesign As explained in https://stackoverflow.com/a/53573622/4001231, setting the cookie with an expired date in the past is the way [the spec](https://tools.ietf.org/search/rfc6265) defines how you should delete a cookie.

> Finally, to remove a cookie, the server returns a Set-Cookie header with an expiration date in the past.  The server will be successful in removing the cookie only if the Path and the Domain attribute in the Set-Cookie header match the values used when the cookie was created.

```suggestion
  serialize(value: Value, options?: CookieSerializeOptions): Promise<string>;
```

MichaelDeBoey (changes requested):

mikeybinnswebdesign:
Sorry @MichaelDeBoey I missed your response to this, all makes sense 😄

mikeybinnswebdesign:
@MichaelDeBoey I have done the work to make sessions work with a default value of unknown. Am I okay to rebase the patch-5 branch onto the dev branch and then commit those changes?

mikeybinns:
@sergiodxa has raised a good point in [this discussion](https://github.com/remix-run/remix/discussions/5100#discussioncomment-4688549) that cookies are user input and therefore could be different from what we expect. This shouldn't be a problem with encrypted cookies, so I'm going to update this PR to only type the cookie if a secret value is set.

mikeybinns:
I've tried to do this over the weekend but I'm not able to work out how to do it because the function is returned by the factory function, rather than being its own named function. Perhaps someone with more Typescript knowledge may have better luck?

chaance:
`createCookie<string>()` isn't really type-safe, it's basically just a fake promise in this case. Even with encrypted cookies, you still want to validate the input somewhere on your server.

My suggestion is to create a wrapper for `createCookie` that handles runtime validation for you. And if you *really* want to skip runtime validation you're free to do so, I just don't think we want to encourage that behavior!

```ts
import {
  createCookie as remixCreateCookie,
  type CookieOptions,
  type CookieParseOptions,
} from "@remix-run/node";

interface ValidatedCookieValue {
  whatever: string;
}

export function createCookie(
  name: string,
  cookieOptions?: CookieOptions
) {
  let cookie = remixCreateCookie(name, cookieOptions);
  return {
    ...cookie,
    async parse(
      cookieHeader: string | null,
      parseOptions?: CookieParseOptions
    ): Promise<ValidatedCookieValue> {
      let parsed = await cookie.parse(cookieHeader, parseOptions);
      assertValidCookieValue(parsed);
      return parsed;
    },
  };
}
```

This could also be done in a more flexible way, say if `parse` was generic and accepted its own validation function through which you could infer the return type. But you should be able to achieve what you're after as long as you're validating input.

mikeybinns:
Okay, thanks @chaance, @sergiodxa echoed the same concerns and you're both right, we should be validating this cookie and not relying on false types 👍

ansavchenco:
Maybe i'm missing something but can the result of `cookie.parse` be anything other than `string` (or `null`/`undefined`) if cookie is not there? Why generic or `any`?

PR #3225, opened 2022-05-18
chore: update Deno import maps on release

MichaelDeBoey:


scripts/utils.js
```diff
@@ -176,6 +176,49 @@ async function updateDeploymentScriptVersion(nextVersion) {
...
+
+  return importName.startsWith("@")
+    ? [`${part1}/${part2}`, rest.join("/")]
+    : [part1, [part2, ...rest].join("/")];
+};
+/**
+ * @param {string} importMapPath
+ * @param {string} nextVersion
+ */
+const updateDenoImportMap = async (importMapPath, nextVersion) => {
```
pcattori, line 194: Is this guaranteed to preserve key ordering, both for the comments and for the imports?

scripts/utils.js
```diff
@@ -176,6 +176,49 @@ async function updateDeploymentScriptVersion(nextVersion) {
...
+
+  let newImports = Object.fromEntries(
+    Object.entries(imports).map(([importName, path]) => {
+      let [packageName, importPath] = getPackageNameFromImportName(importName);
+
+      return remixPackagesFull.includes(packageName)
+        ? [
+            importName,
+            `https://esm.sh/${packageName}@${nextVersion}${
+              importPath ? `/${importPath}` : ""
```
pcattori, line 208: We don't have any imports using paths beyond the package name, and those wouldn't be part of our stable API anyway.

Could you simplify this part so that we don't account for an import path after the package name?

pcattori: Oh good point, I forgot about virtual modules like this one.

scripts/utils.js
```diff
@@ -176,6 +176,49 @@ async function updateDeploymentScriptVersion(nextVersion) {
...
+ * @param {string} importName
+ * @returns {[string, string]} [packageName, importPath]
+ */
+const getPackageNameFromImportName = (importName) => {
+  let [part1, part2, ...rest] = importName.split("/");
+
+  return importName.startsWith("@")
+    ? [`${part1}/${part2}`, rest.join("/")]
+    : [part1, [part2, ...rest].join("/")];
+};
```
pcattori, line 189: You coul use the `limit` parameter of `String.split` to make the logic easier to follow:

```suggestion
const getPackageNameFromImportName = (importName) => {
  if (importName.startsWith("@")) {
    let [ scope, pkg, path] = importName.split("/", 2)
    return { package: `${scope}/{pkg}`, path }
  }
  let [pkg, path] = importName.split("/", 1)
  return { package: pkg, path }
};
```

pcattori: @MichaelDeBoey ^ not sure if you saw my second comment. I'm pretty sure `split` with a limit doesn't return things after the limit unfortunately.

MichaelDeBoey: Oops, I missed that comment 🙈

It doesn't seem to include the rest indeed
> [...] Any leftover text is not included in the array at all.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#limit

Will revert back to what I had.

scripts/utils.js
```diff
@@ -176,6 +176,48 @@ async function updateDeploymentScriptVersion(nextVersion) {
   );
 }
 
+/**
+ * @param {string} importName
+ * @returns {[string, string]} [packageName, importPath]
+ */
+const getPackageNameFromImportName = (importName) => {
```
pcattori, line 183: nit: ASTExplorer calls the package name in `"` the "import specifier". I think that is more apt here than `importName`:

```suggestion
const getPackageNameFromImportSpecifier = (importSpecifier) => {
```

scripts/utils.js
```diff
@@ -176,6 +176,48 @@ async function updateDeploymentScriptVersion(nextVersion) {
...
+  let [part1, part2, ...rest] = importName.split("/");
+  return importName.startsWith("@")
+      ? [`${part1}/${part2}`, rest.join("/")]
+      : [part1, [part2, ...rest].join("/")];
```
pcattori, line 187: I think even though we can't use the `limit` parameter here, the code still benefits from readability by hoisting the `startsWith("@")` logic:

```suggestion
  if (importSpecifier.startsWith("@")) {
    let [scope, pkg, ...path] = importSpecifier.split("/");
    return [`${scope}/${pkg}`, path.join("/")];
  }
  let [pkg, ...path] = importSpecifier.split("/");
  return [pkg, path.join("/")];
```

pcattori (approved):

pcattori:
@MichaelDeBoey this should target `dev` instead of `main` to that this code gets run for the next release.

PR #3229, opened 2022-05-18
tests: bring `deno` fixture template in line with actual template

MichaelDeBoey:


integration/helpers/deno-template/.eslintignore
```diff
@@ -0,0 +1 @@
+*
```
kentcdodds, line 1: Could we add a comment explaining why this file is here?

kentcdodds (changes requested):
Looks good. Just one thing.

integration/helpers/deno-template/.eslintignore
pcattori: Yea I think we can just remove this file and be ok.

integration/helpers/deno-template/package.json
```diff
@@ -2,14 +2,15 @@
   "name": "remix-template-deno",
   "private": true,
   "sideEffects": false,
+  "scripts": {},
```
pcattori, line 5: Why is `scripts` needed?

MichaelDeBoey: Leftover from cleanup, will remove

kentcdodds (approved):
👍

PR #3234, opened 2022-05-18
feat(remix-express): remove dependency on `express`

nicksrandall:
This PR removes express as a dependency from the "express" handler (using only vanilla `http` module). This allows the handler to be used by any "express like" router (like polka) or using `http.createServer` directly! This should improve remix's compatibility with many node based http server frameworks. 

## Example (using http.createServer & sirv)

```ts
import * as http from 'http';
import * as path from 'path';
import sirv from 'sirv';
import { createRequestHandler } from '@remix-run/express';

const remixHandler = createRequestHandler({ build: require(BUILD_DIR) });

const server = http.createServer(sirv("public", {
	  maxAge: 3600,
	  setHeaders(res, pathname) {
		    if (pathname.startsWith("/build/")) {
		      res.setHeader("Cache-Control", "public, max-age=31536000, immutable");
		    }
	  },
	  onNoMatch: remixHandler
}));

server.listen(process.env.PORT || 3000);
```

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

All tests pass.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -341,6 +341,7 @@
 - ni554n
 - nicholaschiang
 - nickytonline
+- nicksrandall
 - niconiahi
 - nielsdb97
 - ninjaPixel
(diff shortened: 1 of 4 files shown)


chaance:
Looks like this PR got borked a bit with some unrelated changes related to `.node` files in https://github.com/remix-run/remix/pull/3234/commits/f724f0aaeaff9bbd30c3d62360409217deb3e20a. Mind fixing those up so we can take a look?

nicksrandall:
This is the strangest thing but maybe I mixed up my PRs? This is code for a completely unrelated PR. Let me fix this.

chaance:
@nicksrandall no worries, I've done the same 😅  Usually happens because you try to fix two issues on the same branch and accidentally end up force-pushing to the wrong one.

nicksrandall:
Closing in favor of #5237

PR #3235, opened 2022-05-18
chore(vscode): add recommended extensions

pcattori:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


mcansh:
perhaps we recommend eslint too?

.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+  "recommendations": [
+    "esbenp.prettier-vscode",
+    "denoland.vscode-deno"
```
mcansh, line 4:
```suggestion
    "esbenp.prettier-vscode",
    "denoland.vscode-deno",
    "dbaeumer.vscode-eslint"
```

pcattori: Hmm, for some reason I remember that "format on save" for eslint extension caused me some headaches before (made some unnecessary git noise), but we run eslint in CI anyway, so I think it should be fine.

pcattori: Added eslint extension to the list.

mcansh: i *think* that's due to both trying to format on save?

i think this is what solved it for me, i have a few eslint related things in my settings

```json
{
	"editor.defaultFormatter": "esbenp.prettier-vscode",
	"editor.codeActionsOnSave": { "source.fixAll.eslint": true }
}

PR #3239, opened 2022-05-19
fix(templates/deno): use `Deno.env` instead of `process.env`

CanRau:
`process.env.NODE_ENV` works, but VSCode unnecessarily complains and it also doesn't feel Deno-y enough so I would propose changing it to the appropriate `Deno.env.get("NODE_ENV")`.

Please correct me if I'm mistaken 🙏

Also so happy to finally get Deno Remix just working, with the latest Deno and all 🥳 Thanks so much @pcattori  👏

### DIFF
--- a/templates/deno/server.ts
+++ b/templates/deno/server.ts
@@ -5,7 +5,7 @@ import * as build from "@remix-run/dev/server-build";
 
 const remixHandler = createRequestHandlerWithStaticFiles({
   build,
-  mode: process.env.NODE_ENV,
+  mode: Deno.env.get("NODE_ENV"),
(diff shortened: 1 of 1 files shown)


CanRau:
I'm sorry I have no idea what this is, I just edited one file in the GitHub UI and switched to the dev branch when comparing before creating the PR 😳

MichaelDeBoey:
Since this is a change to the templates, this can go to `main`

MichaelDeBoey (approved):

CanRau:
Thought about it but want sure 😇
Thank you 🙏

PR #3243, opened 2022-05-19
fix(templates/cloudflare-pages): fix runtime globals reference

Dueen:
The globals file got removed in a previous commit and the d.ts has not been updated to reflect that change.


templates/cloudflare-pages/remix.env.d.ts
```diff
@@ -1,3 +1,3 @@
 /// <reference types="@remix-run/dev" />
-/// <reference types="@remix-run/cloudflare-pages/globals" />
+/// <reference types="@remix-run/cloudflare-pages" />
```
MichaelDeBoey, line 2: This should point to the runtime globals instead

```suggestion
/// <reference types="@remix-run/cloudflare/globals" />
```

MichaelDeBoey (changes requested):

PR #3249, opened 2022-05-19
fix(remix-dev): improve `appTemplate` message

nvh95:
When running `npx create-remix@latest`, The CLI asks me 

"Where do you want to deploy? Choose **Remix** if you're unsure; it's easy to  change deployment targets."

But there is no option such as **Remix**, but **Remix App Server**. This PR is to make users not confused when first using `create-remix`

using create-remix to bootstrap a remix project

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -266,6 +266,7 @@
 - nobeeakon
 - nordiauwu
 - nurul3101
+- nvh95
 - nwalters512
 - octokatherine
 - omamazainab
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -103,8 +103,8 @@ R E M I X
 
 ? Where would you like to create your app? remix-jokes
 ? What type of app do you want to create? Just the basics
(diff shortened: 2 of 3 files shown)


machour (approved):

MichaelDeBoey (approved):

nvh95:
> @nvh95 Code changes need to go against `dev`.

@MichaelDeBoey I think if any code changes need to go against `dev`, then `remix` repository should set the default branch to `dev`, instead of `main`, it will be easier to contributors to create PRs.
cc: @kentcdodds

kentcdodds:
I'm pretty sure the majority of PRs are for docs/examples/templates which need to go into `main`. Additionally, folks contributing code normally take more care about something like that than someone contributing to docs. I think we've got the right default.

nvh95:
@kentcdodds I see, it's clear that I haven't checked out [the CONTRIBUTING guide](https://github.com/remix-run/remix/blob/main/docs/pages/contributing.md). Thanks.
contributing guide for remix

In short,
- docs/ example/ templates PRs => PR to `main`
- Touches code => PR to `dev`

PR #3250, opened 2022-05-19
fix(examples): fix `README` of `stitches` example

himorishige:
We have removed the reference to `server.context.tsx` from the documentation, as it appears that `server.context.tsx` is not used in the sample source files.


machour (approved):

machour:
Thank you @himorishige 🙌🏼

PR #3252, opened 2022-05-19
tests: add jsconfig path mapping integration test

F3n67u:
Follow up of https://github.com/remix-run/remix/pull/3074#issuecomment-1129081462

- [ ] Docs
- [x] Tests

Testing Strategy:


MichaelDeBoey:
@F3n67u As suggested by @kentcdodds in https://github.com/remix-run/remix/pull/3074#issuecomment-1132127958:

> [...] write the tests and add a `.skip` on them with a `// TODO: ` comment explaining why they're skipped for now.

integration/helpers/create-fixture.ts
```diff
@@ -18,7 +18,7 @@ interface FixtureInit {
   buildStdio?: Writable;
   sourcemap?: boolean;
   files: { [filename: string]: string };
-  template?: "cf-template" | "deno-template" | "node-template";
+  template?: "cf-template" | "deno-template" | "node-template" | "node-javascript-template";
```
MichaelDeBoey, line 21: Not sure if we want an extra template for this though. 🤔

Maybe we should wait for the `convert-to-javascript` migration (see #3150) & call the migration before we do anything else in our tests?

kentcdodds: Hmmm... I think I'd prefer to duplicate than make a dependency here.

MichaelDeBoey: This means we'll potentially have them double.

What about adding an extra (optional) key to `FixtureInit` that says if we want to use TypeScript or not.
If not, the `createFixture` function could run the migration automatically?
That way we don't need double templates + we can choose to use JS in our tests if we want to.

kentcdodds: I'm ok with that 👍

chaance:
I noticed this is still a draft PR, so I'm closing for now while we clean up some old issues/PRs. Feel free to ping me if and when you're ready to see this through!

MichaelDeBoey:
I think these tests are still valuable to have

@F3n67u Could you please open a new PR with these tests?

PR #3256, opened 2022-05-19
chore(templates/deno): add missing typings to `devDependencies`

machour:


MichaelDeBoey:
According to @pcattori, this isn't necessary (see https://github.com/remix-run/remix/pull/3117#discussion_r868543212)

> We don't need `@types/*` nor `typescript` in `devDependencies` because users will be relying on Deno tools (like the VSCode Deno extension) to provide types. These tools will use import maps to find the appropriate types (e.g. `templates/deno/.vscode/resolve_npm_imports.json`).
> 
> In short, Deno does typechecking via a custom TS compiler implementation (not via `tsc`). That implementation doesn't grab types from `node_modules/` at all.

machour:
I am using WebStorm instead of VS Code, and it was complaining to me about this.
Turns out I needed to install the Deno plugin, sorry for the noise!

MichaelDeBoey:
@machour Had the same problem before indeed

PR #703, opened 2021-11-27
chore: add support for native .node modules to remix-dev (making tailwind integration pain free)

sync:
Was checking out remix-tailwind and it had issues on OSX due to fsevents use of .node modules in OSX. I think remix-tailwind is a super nice way to do tailwind in remix, pain free and it would be nice if our OSX users could also use it.

Here you can find more details about the issue:
https://github.com/itsMapleLeaf/remix-tailwind/issues/1

This code was originated from here and written by Evan:
https://github.com/evanw/esbuild/issues/1051

For anyone interested trying this fix, feel free to:
- install [patch-package](https://github.com/ds300/patch-package#readme)
- create a patches folder in the root
- unzip and drop the patched files into the patches folder

[@remix-run+dev+1.0.6.patch.zip](https://github.com/remix-run/remix/files/7612991/%40remix-run%2Bdev%2B1.0.6.patch.zip)

PR #3261, opened 2022-05-19
feat(remix-dev): add `.node` file support for Node based platforms

benwis:
## Changes Made
This pull requests adds an esbuild plugin and some tweaks to the Remix compiler to allow the import of `.node` files in Remix's loaders and actions. While I was motivated by the desire to write Rust helper functions for server side modules, this was also mentioned in this [other pull request](https://github.com/remix-run/remix/pull/703) for .node imports for remix-tailwind, and in [a discussion](https://github.com/remix-run/remix/discussions/2560) by @nicksrandall for sharp support.

I have started [a discussion](https://github.com/remix-run/remix/discussions/3201) here, but as of now hasn't generated much traffic.

## Benefits
The primary benefit I see is the ability to run other language's code on your server easily in Remix. This could be done in Rust using napi-rs, in Python, or in C. That offers substantial benefits in performance, opens possibilities for easy concurrency, and other possibilities.

## Limitations
One of the biggest limitations is that it can only run on Node based platforms. Napi-rs is currently working on Deno support, so we may see this come there as well. 

## Testing
I have modified the Rust example in the main repo with to generate a sum() function with [napi-rs](https://napi.rs/) and use that to perform the addition that example does. The repo for this is [here](https://github.com/benwis/rust-remix). There are a few build steps as described in that README that must be performed, depending on the platform you're running Remix on.

- [x] Docs

### DIFF
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -89,6 +89,7 @@ export const createCookieFactory =
     let { secrets, ...options } = {
       secrets: [],
       path: "/",
+      sameSite: "lax" as const,
       ...cookieOptions,
     };
 


MichaelDeBoey:
@benwis It seems like you have unrelated commits in this PR 
Could you please remove them?

benwis:
@MichaelDeBoey Yeah I noticed that, no idea how that happened. Let me delete those

MichaelDeBoey:
New version: #3263

PR #3262, opened 2022-05-19
docs(decisions/0003): package for shared server features

pcattori:
and let each adapter export adapter-specific features, obsoleting the
need for direct imports from server runtime packages.

### DIFF
--- a/decisions/0003-package-for-shared-server-features.md
+++ b/decisions/0003-package-for-shared-server-features.md
@@ -1,0 +1,113 @@
+# Package for shared server features
+
+Date: 2022-06-01
+
+Status: proposed
+
+## Context
+
+Currently, users need to install 4 Remix dependencies:
+
+- Dev (`@remix-run/dev`)
+- Server runtime (`@remix-run/{cloudflare,deno,node}`)
+- Adapter (`@remix-run/{arc,cloudflare-pages,cloudflare-workers,express,fly,netlify,vercel}`)
+- Renderer (`@remix-run/react`)
+
+User code currently looks like this:
+```tsx
+// server runtime
+import { json, createFileSessionStorage } from "@remix-run/node";
+// adapter
+import { createRequestHandler } from "@remix-run/express"
+// renderer
+import { useLoaderData } from "@remix-run/react";
+```
+
+### Goals
+
+- make DX and docs simpler: more intuitive, less magic, eliminate some user errors
+- reduce bugs from `remix` internal complexity
+
+### Non-goals
+
+- maintainability of `remix` repo
+- more portable `app/` directory
+- reduce or increase number of packages
+
+Some of these are nice-to-haves, but not the primary reasons for this proposal.
+
+### Issues
+
+1. Users can import features from their runtime that aren't supported by their adapter:
+
+```tsx
+import { createFileSessionStorage } from "@remix-run/node";
+// Implementation only depends on server runtime, but _availability_ depends on the adapter!
+// ❌ This will crash in netlify nor in vercel!
+```
+
+2. Docs, examples, and videos refer to `@remix-run/node` by default for shared functionality like `json`:
+
+```tsx
+import { json } from "@remix-run/node";
+// Need docs to tell users that they should import from their corresponding runtime package, 
+// and not from `@remix-run/server-runtime`.
+```
+
+3. Keeping shared exports consistent across runtime packages is [error-prone](https://github.com/remix-run/remix/pull/3215) and tedious.
+
+## Decision
+
+Users should import features from _adapter packages_, not from runtime packages. Each adapter can export only the features it supports. Shared features will be imported from a new package, `@remix-run/server`.
+
+This means there won't be any imports needed from the runtime packages (`@remix-run/{cloudflare,deno,node`}) and these will become an implementation detail for most users. Only developers that need to develop adapters will import from the runtime packages directly.
+
+### User-facing packages
+
+Users still need to import from 4 Remix packages:
+
+- Dev `@remix-run/dev` (unchanged)
+- Shared server features `@remix-run/server`
+- Adapter-specific server features (`@remix-run/{arc,cloudflare-pages,cloudflare-workers,express,fly,netlify,vercel}`)
+- Renderer `@remix-run/react` (unchanged)
+
+User code would look like this:
+```tsx
+// server
+import { json } from "@remix-run/server";
+// adapter
+import { createRequestHandler, createFileSessionStorage } from "@remix-run/fly"
+// renderer
+import { useLoaderData } from "@remix-run/react";
+```
+
+### Internal packages
+
+- Runtime packages for authoring adapters (`@remix-run/{cloudflare,deno,node}`)
+- Package for authoring new runtimes (`@remix-run/server-runtime`), only including factory functions like:
+  - `createCookieFactory`
+  - `createSessionStorageFactory`
+  - `createCookieSessionStorageFactory`
+  - `createMemorySessionStorageFactory`
+
+### Migration
+
+Include a new migration for updating imports.
+
+Can reuse and refactor code from the existing [replace-remix-imports](https://github.com/remix-run/remix/blob/main/packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts) migration with [small tweaks to the import maps](https://github.com/remix-run/remix/blob/main/packages/remix-dev/cli/migrate/migrations/replace-remix-imports/transform/mapNormalizedImports/packageExports.ts) and affected imports (`@remix-run/*` instead of `remix`).
+
+## Consequences
+
+DX will be simplified as it will be more obvious to users what package to import from. Less magic, complexity, abstraction, etc...
+
+1. Users can only import features supported by their adapter, **eliminating a category of user errors**.
+2. Docs, examples, and videos will unambiguously use `@remix-run/server` for shared features and particular adapter features from adapter packages when necessary.
+3. All re-exports from runtime packages are obsoleted, **reducing bugs and missing features from Remix packages**.
+
+Additionally,
+- migrating imports will be automated via `npx @remix-run/dev migrate`.
+- VS Code will recommend a _single_ package when importing values like `json`.
+- Code in `app/` will be more portable and non-portable, adapter-specific features will be more obvious.
+
+Note: While having portable `app/` code might be useful for users switching their runtime/adapter, the much more common upside is that it will be much easier for users, especially beginners, to copy code from docs, examples, and videos.
+


decisions/0003-create-a-server-runtime-agnostic-package.md
```diff
@@ -0,0 +1,71 @@
...
+
+Specific package names TBD (`@remix-run/server` might be confusing with `@remix-run/serve` already existing and doing something categorically different).
+
+`@remix-run/server-runtime-toolkit` should not be considered public API, while `@remix-run/server` should be.
+
+## Consequences
+
+### How to import from Remix server packages
+
+Users would import runtime-agnostic values and functions from `@remix-run/server`:
```
MichaelDeBoey, line 44: I don't know if this is desired usage tbh.

Benefits I see:
- We're more transparant about everything
- We need no tests to see that all functions/types/values are exported by all 3 runtimes.

However, the disadvantages are:
- People need to care about more than just the runtime they're using
- It will become harder for us to have runtime specific implementations of certain things
  We currently have runtime specific implementations for `createCookie`, `createCookieSessionStorage`, `createMemorySessionStorage`, ` createSessionStorage` & `createRequestHandler` (only specific in `deno` runtime).
  We would either loose the disability to have this because we're putting all common functions/types/values in 1 package
  _OR_ we would cause confusion because people won't know what would be in the generic package & what would be in the runtime package.

Especially the last disadvantage is a huge one imo.

pcattori: To me it sounds like your points of "People need to care..." and "we would cause confusion because..." are the same thing, but correct me if I'm wrong.

I don't think it would be confusing nor limiting. We can simply follow the rule "If any function has at least 1 runtime-specific implementation, its part of the server runtime interface". For example, `createRequestHandler`.

But the _all_ types from `server-runtime` and the vast majority of values/functions are _never_ runtime-specific.

For users, that means if you are trying to use:
- `createCookie`
- `createCookieSessionStorage`
- `createMemorySessionStorage`
- `createSessionStorage`
- `createRequestHandler`
then grab those from your server runtime package.

For everything else, grab it from `@remix-run/server`.

MichaelDeBoey: What about the fact that we start as an agnostic package (which was the case with `createRequestHandler` before we had support for `deno`) and later on decide it needs a platform specific implementation?

kiliman: In hindsight, I think `createRequestHandler` is something that should have always been considered platform specific. With the experience we have over the past several months, I think it would be possible to determine if a new API is a common utility function, or something that could be platform specific.

I just don't think it would be a good idea to ditch this proposal because of past "mistakes".

pcattori: > What about the fact that we start as an agnostic package (which was the case with `createRequestHandler` before we had support for `deno`) and later on decide it needs a platform specific implementation?

I think those cases will be rare considering there aren't more than a handful of JS runtimes, but if we find out something needs to be runtime-specific, we can put it into each runtime and provide a codemod for it. Should be easy now that we have existing codemods for changing where things get imported from 😄 But yea, I expect that to be quite rare if it happens at all.

decisions/0003-blah.md
```diff
@@ -0,0 +1,57 @@
...
+
+Goals:
+- make DX and docs simpler: more intuitive, less magic
+- reduce bugs from user error
+- reduce bugs from `remix` internal complexity
+
+Non-goals:
+- maintainability of `remix` repo
+- reduce or increase number of packages
+- more portable `app/` directory
```
MichaelDeBoey, line 55: It is currently still the recommended way of changing adapters to just create a new project and copy over the `app` directory though.

https://twitter.com/mjackson/status/1532361967133831174

decisions/0003-package-for-shared-server-features.md
```diff
@@ -0,0 +1,108 @@
...
+### Goals
+
+- make DX and docs simpler: more intuitive, less magic, eliminate some user errors
+- reduce bugs from `remix` internal complexity
+
+### Non-goals
+
+- maintainability of `remix` repo
+- reduce or increase number of packages
+- more portable `app/` directory
```
MichaelDeBoey, line 35: It is currently still the recommended way of changing adapters to just create a new project and copy over the `app` directory though.

https://twitter.com/mjackson/status/1532361967133831174

Not saying we should stick to it, but people will still want to have an easy way to change adapters, so I think we should at least take this into account.

MichaelDeBoey (commented):
I like the general idea of this RFC! 👍

The simplicity it creates for users + the fact that users won't be able to use unsupported stuff will prevent a lot of bugs I think.

One small things I think we should keep in our minds though.

decisions/0003-package-for-shared-server-features.md
pcattori: Added some prose explaining this^ in the proposal.

MichaelDeBoey (approved):

decisions/0003-package-for-shared-server-features.md
```diff
@@ -0,0 +1,111 @@
...
+- Runtime packages for authoring adapters (`@remix-run/{cloudflare,deno,node}`)
+- Package for authoring new runtimes (`@remix-run/server-runtime`), only including factory functions like:
+  - `createCookieFactory`
+  - `createSessionStorageFactory`
+  - `createCookieSessionStorageFactory`
+  - `createMemorySessionStorageFactory`
+
+### Migration
+
+Include a new migration for updating imports. Should be a straightforward copy of the existing [replace-remix-imports](https://github.com/remix-run/remix/blob/main/packages/remix-dev/cli/migrate/migrations/replace-remix-imports/index.ts) migration with [small tweaks to the import maps](https://github.com/remix-run/remix/blob/main/packages/remix-dev/cli/migrate/migrations/replace-remix-imports/transform/mapNormalizedImports/packageExports.ts).
```
MichaelDeBoey, line 95: We could use the `replace-remix-imports` migration for this instead of copy/pasting the whole thing over.

Only things needed to change are:
- The imports map (as already mentioned)
- The imports to change
  Currently we're only looking at `remix` imports. This should be changed to also include `@remix-run/*` packages.

decisions/0003-package-for-shared-server-features.md
```diff
@@ -0,0 +1,113 @@
...
+
+This means there won't be any imports needed from the runtime packages (`@remix-run/{cloudflare,deno,node`}) and these will become an implementation detail for most users. Only developers that need to develop adapters will import from the runtime packages directly.
+
+### User-facing packages
+
+Users still need to import from 4 Remix packages:
+
+- Dev `@remix-run/dev` (unchanged)
+- Shared server features `@remix-run/server`
+- Adapter-specific server features (`@remix-run/{arc,cloudflare-pages,cloudflare-workers,express,fly,netlify,vercel}`)
```
MichaelDeBoey, line 71: What would be the recommended adapter when just figuring things out locally?

Right now if you're using the `fly` or `remix` template, you don't have an adapter installed and are just importing functions from `@remix-run/node`.
Since we want people to import as much as possible from an adapter, should we create a `remix` adapter?
Or do we want to recommend another adapter (like `@remix-run/fly`, which currently also doesn't exist) instead?

The `deno` template has the same problem. We however have another concern here as `@remix-run/deno` is the runtime, so should we create a `@remix-run/deno-deploy` adapter for this?

pcattori: The `remix` template uses `@remix-run/serve`, and consequently, `@remix-run/express`. So it should either (a) re-export adapter-specific features from `@remix-run/express` or (b) declare `@remix-run/express` as a peer dependency and users can import directly from `@remix-run/express` for adapter-specific features. 

---

I think it does make sense to have a `@remix-run/fly` adapter that only exports the features available on Fly.

---

For Deno: yes, I think we should create a `@remix-run/deno-deploy` adapter.

MichaelDeBoey: `@remix-run/serve` isn't really an adapter, but a way to serve the code, right?

It's used in both `fly` & `remix` templates to do so.

When using `@remix-run/netlify` adapter, we also need an extra package (`netlify-cli` globally installed) to serve the code, so it's also not coming from the adapter directly.
Same when using `@remix-run/cloudflare-pages`, we use `wrangler` to serve the code.

So I think we still need an adapter package for the `remix` template (next to the new `fly` adapter) 🤔

decisions/0003-package-for-shared-server-features.md
```diff
@@ -0,0 +1,113 @@
...
+Users should import features from _adapter packages_, not from runtime packages. Each adapter can export only the features it supports. Shared features will be imported from a new package, `@remix-run/server`.
+
+This means there won't be any imports needed from the runtime packages (`@remix-run/{cloudflare,deno,node`}) and these will become an implementation detail for most users. Only developers that need to develop adapters will import from the runtime packages directly.
+
+### User-facing packages
+
+Users still need to import from 4 Remix packages:
+
+- Dev `@remix-run/dev` (unchanged)
+- Shared server features `@remix-run/server`
```
MichaelDeBoey, line 70: We currently already have a draft version of [`@remix-run/router`](https://github.com/remix-run/react-router/tree/19d5fa6ab66300ef529dcf39f8089f3869d6e843/packages/router) (which will be official once `react-router` v6.4 is released), which exports a lot of `@remix-run/server-runtime` (including functions like `json` & `redirect`).

Do we want to keep the functionality in `@remix-run/router` or do we still want to extract it into `@remix-run/server`?

sergiodxa: I thing those functions only make sense on the server when using Remix, since you will only use them there. So it will make sense to export them from `@remix-run/server`

MichaelDeBoey: @sergiodxa `loaders` could return a `Response`, so _at least_ the `json` helper could be interesting to use on the frontend when using `react-router` as well.

CC/ @brophdawg11 @ryanflorence

sergiodxa: When using React Router you can use `json` and `redirect` no the client, but when using Remix there's no reason, your loader and action functions only run server-side.

Why would you want to create a Response using those helpers client-side in Remix?

MichaelDeBoey: @sergiodxa It seems a bit weird to me as well, but it's explained in the docs
https://beta.reactrouter.com/en/dev/route/loader#returning-responses

sergiodxa: Yeah I understand that, my point is that makes sense in React Router, but in Remix there’s no need for using those response helpers client side, only server side, so they should be exported from the proposed @remix-run/server package only, not client side

brophdawg11: I'm not too familiar with the server adapter landscape - but `json`/`redirect` will indeed be exported from `@remix-run/router`.  I'm unsure if we'll actually leverage that package to do SSR data fetching when we bring remix on top of RR 6.4 but at the very least we can re-export them through the remix server packages so we're not duping them

pcattori: Yea I think `json` and similar shared server features from the proposed `@remix-run/server` will need some SSR-specific implementation on the Remix side. So then having RR export a client-side/SPA version of `json` (and friends) is fine alongside Remix's SSR-specific implementations.

If we find that we can generalize the implementation so that it doesn't depend on whether that feature is used client- or server-side, then we can implement where ever  we think is best and re-export it as needed.

sergiodxa: But there's no server-side or client-side versions of `json` and `redirect`, they just create a `Response`, that could work server-side and client-side, and the `Response` is global on both the server (in Remix) and the browser

pcattori: Yea I'm just referring to `json` and `redirect` as examples of shared server features/exports. Those ones aren't server-side nor client-side specific, but I'm just saying that even if there _were_ features that needed to be client/server-side specific, we'd still have a way to deal with that too.

So just saying that we aren't painting ourselves into a corner.

decisions/0003-package-for-shared-server-features.md
pcattori: > @remix-run/serve isn't really an adapter, but a way to serve the code, right?

That's right. The revelant adapter for `@remix-run/serve` and the `remix` template is `@remix-run/express`. That's why I'm leaning towards having `@remix-run/serve` specify `@remix-run/express` as a peer dependency, so that users can still import adapter-specific features from `@remix-run/express`.

MichaelDeBoey: `@remix-run/serve` is also used in the `fly` & `vercel` template, so especially for the `vercel` template we can't use `@remix-run/express`.

pcattori: Yea there's really a conflation between two things:
1) The request/response interface, which is specific to e.g. express
2) The server features available for a host, which is specific to e.g fly or vercel or etc...

I'll think a bit about what we can do to capture this distinction more clearly and what the API for that would be, i.e. where to import what.

decisions/0003-package-for-shared-server-features.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+### Internal packages
+
+- Runtime packages for authoring adapters (`@remix-run/{cloudflare,deno,node}`)
+- Package for authoring new runtimes (`@remix-run/server-runtime`), only including factory functions like:
+  - `createCookieFactory`
+  - `createSessionStorageFactory`
+  - `createCookieSessionStorageFactory`
+  - `createMemorySessionStorageFactory`
```
MichaelDeBoey, line 91: I think `createRequestHandler` should still be here as well.

I agree with what @kiliman said in https://github.com/remix-run/remix/pull/3262#discussion_r877519747, that each adapter should have it's own `createRequestHandler`, but they all use the generic `createRequestHandler` (renamed to `createRemixRequestHandler` internally) under the hood, so we should keep that in `@remix-run/server-runtime` I think.

My reasoning in https://github.com/remix-run/remix/pull/3262#discussion_r877459283 about `createRequestHandler` only being specific in `@remix-run/deno` is actually wrong, as we combine both a runtime + adapter in 1 package.
If we would have created a separate `@remix-run/deno-deploy` adapter from the start, `@remix-run/deno` would have just re-exported `createRequestHandler` from `@remix-run/server-runtime`.

pcattori:
This proposal is really about two orthogonal things:
1. Having a shared package where users would import runtime-agnostic things like `json`, `redirect`, `ActionFunction` etc...
2. Making _adapters_ export server features as a guard rail for users e.g. so that they don't try to use the filesystem on hosts that don't provide that.

After talking with the team, we need to think more about how to guide users towards existing server features without relying on the adapter export since hosts may be able to run more than one JS runtime so we'd need a way to export the correct server feature implementation (e.g. `createFileSessionStorage`) based on the runtime _and_ the host's available features.

Additionally, we've moved the shared "server" features into React Router. In the future, we'll have that implementation available as `@remix-run/router` (or something similar) and _that_ will be the shared, runtime-agnostic package. So we don't need to create a new one (`@remix-run/server`), we just need to adopt the existing one (React Router -> `@remix-run/router`). I also put "server" in quotes above because really `json`, `redirect`, `LoaderFunction`, `ActionFunction` _can_ be run by the server, but can also be emulated by the client, so these are more "router" features than "server" features.

brophdawg11:
@pcattori Would it be worth merging the ADR in a "rejected" status with your above explanation?  Would likely be a bit easier to locate there in the future than digging back through the PRs?

pcattori:
@brophdawg11 maybe? I'm leaning towards not merging it since:
1. it never corresponded to a prototype or implementation or any sort of work
2. it would be more of a distraction than a help for anyone looking through the ADRs once we have the Remix Router work in.

MichaelDeBoey:
> we'd need a way to export the correct server feature implementation (e.g. `createFileSessionStorage`) based on the runtime _and_ the host's available features.

@pcattori The current approach for this is to have a separate adapter package (see t[he `@remix-run/netlify-edge` PR](https://github.com/remix-run/remix/pull/3104)).

pcattori:
> @pcattori The current approach for this is to have a separate adapter package (see [the `@remix-run/netlify-edge` PR](https://github.com/remix-run/remix/pull/3104)).

AFAIK Netlify and Netlify Edge are two different product offerings, not one hosting option that can run either Node or Deno. My point is more for cases where a single product offering/hosting option could run multiple JS runtimes.

MichaelDeBoey:
> My point is more for cases where a single product offering/hosting option could run multiple JS runtimes.

Do you have an example of such a product offering/hosting option then?
As far as I know, all of them just support 1 runtime then if you see Netlify & Netlify Edge as 2 different product offerings/hosting options.

pcattori:
To be clear, I think Netlify sees standard Netlify and Netlify Edge as separate products with different feature sets, so I don't view that as just my perspective.

As an example, Fly is a host that runs both Node and Deno. Plus we don't know if tomorrow Netlify will announce that standard Netlify now runs Node or Deno. There's nothing inherent about a set of features provided by a host that prevents them from implementing those features in multiple JS runtimes, so I don't think its an apt assumption to make that one host corresponds to one JS runtime.

pcattori:
Maybe, but maybe not. We'd have host-specific features that are runtime-agnostic, and we'd have runtime-specific features that are host-agnostic, and we'd also potentially have host-specific features that are runtime-specific. This matrix of different combinations ends up being quite large, so we don't want to commit to maintaining a package for each of these combinations for each host and runtime combo, at least not yet.

PR #3268, opened 2022-05-20
docs(examples): add link to `EdgeDB` variant of `jokes` example

colinhacks:
Adding a variant of the `jokes` example that uses [EdgeDB](https://www.edgedb.com) instead of SQLite/Prisma.


machour:
Thank you for pulling this example together @colinhacks !

My first sentiment is that the example should focus solely on the EdgeDb example (and drop the metas, etc) but since this is an adaptation of the jokes tutorials, not sure how we would like to handle this. 

What would @kentcdodds do?

MichaelDeBoey:
I would personally make all the examples as small as possible & only focus on what technology your trying to show.

`blog` & `jokes` examples are an exception imo, as they both have a full blown tutorial written for them as well.

kentcdodds:
Agreed!

I do feel like there's value here though. Here's what I suggest @colinhacks: put this example in your personal repo and update the README of the jokes example to point to yours as a comparison.

May be interesting to start your git repo with the original jokes app and make a branch with that, then update to EdgeDB and link people to a diff.

examples/jokes/README.md
```diff
@@ -22,10 +22,12 @@ This example demonstrates some of the basic features of Remix, including:
 - Resource Routes
 - Deployment
 
-This is the finished version of [the tutorial](https://remix.run/tutorials/jokes)
+This is the finished version of [the tutorial](https://remix.run/tutorials/jokes).
 
 - [Remix Docs](https://remix.run/docs)
 
+> For comparison, a variant of this sample app using EdgeDB instead of SQLite/Prisma is [available here](https://github.com/edgedb/edgedb-examples/tree/main/remix) ([see the Git diff](https://github.com/edgedb/edgedb-examples/compare/d654bcc1...ad86e66b)).
```
kentcdodds, line 29: Would you rather make this compare link compare branches instead? That way if you realize there's an easier/better way to do it you can update the branch without having to update this link.

If not that's fine too. Happy to merge as-is!

kentcdodds (changes requested):
Thanks! Just one thing.

examples/jokes/README.md
colinhacks: I wasn't sure how to do that exactly, because the latest version of the `remix` app will always be in the `main` branch. But if I do a comparison from a `jokes-initial...main`, there will be a bunch of changes mixed in relating to other non-Remix examples (because our `edgedb/edgedb-examples` repo holds sample projects for a variety of frameworks. Am I missing something obvious here?

kentcdodds: Ah, I didn't look carefully enough and didn't notice the fact that this is in your examples repo. I thought you were going to make a completely separate repo.

I think you could accomplish the same with tags. But again, it's not a big deal. Let me know if you'd rather I just merge it as-is :)

colinhacks: Ok I decided we should get rid of the link entirely for the time being 😅 

- It's probably a bit too much to include in a random aside in this README
- I'm worried about that diff going stale as Remix and EdgeDB change their APIs/workflows
- I don't want to bother you guys with PRs down the road just to update this darn link lol

I'll include the link on our end instead so people who want to dig deeper can explore it. Anyway - ready to merge. Thanks Kent :)

MichaelDeBoey (approved):

kentcdodds (approved):
Sounds good! Thanks!

PR #3269, opened 2022-05-20
docs(jokes): Clarify restart of dev server when resetting prisma db

sndrem:
Adding an additional line about remembering to restart your dev server when resetting the local prisma database.


kentcdodds (approved):
Thanks!

PR #3273, opened 2022-05-20
fix(remix-server-runtime): fix `JsonFunction` type

8balloon:
## problem
`@remix-run/node`'s `json()` function accepts non-json types, which commonly leads to mistypings.

## example of the problem
```tsx
import { useLoaderData } from "@remix-run/react";

function getLoaderData() {
  return {
    date: new Date(), // this could be a `createdAt` field, or something similar
  };
}

type LoaderData = ReturnType<typeof getLoaderData>;

  const data = getLoaderData();
  return json(data);
};

export default function DateDisplay() {

  // This will cause a runtime error, because data.date is a string, although it's typed as a function.
  // The reason is because `json()` uses `JSON.stringify()`, which automatically converts dates to ISO strings
  return <div>Day from date is {data.date.getDay()}</div>;
}

```
## solution
`json(data)` would result in a type error in the above example, which would prevent a runtime error due to mistyping.

Also remove unused `Data` type parameterization (could easily revert this if we wish)

### DIFF
--- a/packages/remix-server-runtime/responses.ts
+++ b/packages/remix-server-runtime/responses.ts
@@ -6,18 +6,16 @@ type Json =
   | Array<Json>
   | { [key: string]: Json };
 
-export type JsonFunction = (
-  data: Json,
-  init?: number | ResponseInit
-) => Response;
-
 /**
  * This is a shortcut for creating `application/json` responses. Converts `data`
  * to JSON and sets the `Content-Type` header.
  *
  * @see https://remix.run/api/remix#json
  */
-export const json: JsonFunction = (data, init = {}) => {
+export const json = <Data extends Json>(
+  data: Data,
+  init: number | ResponseInit = {}
+) => {
   let responseInit = typeof init === "number" ? { status: init } : init;
 
   let headers = new Headers(responseInit.headers);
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
@8balloon Code changes go against `dev`.
Can you please rebase your branch onto latest `dev` & set base branch to `dev` as well?

8balloon:
@MichaelDeBoey it looks like this solution may not work, as JSON-compatible interfaces are incompatible with the `Json` type I've defined. Open to ideas.

If this were my own codebase, I'd replace `interface` uses with `type` uses through the codebase, as those play nicely with the `Json` type I've defined.

packages/remix-server-runtime/responses.ts
```diff
@@ -1,5 +1,13 @@
...
+type Json =
+  | null
+  | boolean
+  | number
+  | string
+  | Array<Json>
+  | { [key: string]: Json };
+
+export type JsonFunction = (
+  data: Json,
```
sergiodxa, line 10:
```suggestion
export type JsonFunction = <Data extends Json>(
  data: Data,
```
This way, `json` will support a generic, but the generic needs to match the Json type. Without this, there's no way to pass a generic like right now and this doing that change will cause a breaking change.

Technically, doing this will also be a breaking change but since it's more similar to the correct behavior you could consider the previous one allowing things like Date a bug.

8balloon: Good callout 👍

PR #3276, opened 2022-05-21
Loader inference

colinhacks:
Discussed in  https://github.com/remix-run/remix/pull/1254

- [X] Tests

### Testing Strategy:

- Test `remix-server-runtime > json > infers input type` ensures that the return type of `json` is a generic `Response`
- Test `remix-react > hook-types-test` checks the type logic for `useLoaderData`

### Guidance needed:

**Documentation**

In my opinion, all TypeScript examples should be changed to use `LoaderArgs`, which allows TypeScript to properly infer the return type:
```ts
// OLD
const loader: LoadFunction = (args)=>{ ... }
type LoaderData = ...

function App(){
}


// NEW
const loader = (args: LoaderArgs)=>{ ... }

function App(){
  const data  = useLoaderData<typeof loader>();
}
```

There are lot of instances of the old approach in the docs. I'm happy to change them but wanted to get confirmation first.

**`makeLoader`**

Relatedly, my proposal in #1254 also included the idea of a `makeLoader` HOC for writing loaders. It would take some tweaking to the compiler to guarantee this was always eliminated during tree shaking, e.g. I wasn't sure if the team was okay with this so I held off for now.

```ts
let plugins = [
  // ...
  emptyModulesPlugin(config, /^@remix-run/(node|cloudflare)$/)
];
```


packages/remix-react/components.tsx
```diff
@@ -20,6 +20,7 @@ import {
   useResolvedPath,
 } from "react-router-dom";
 import type { LinkProps, NavLinkProps } from "react-router-dom";
+import type { LoaderFunction } from "@remix-run/server-runtime";
```
MichaelDeBoey, line 23: This makes `@remix-run/server-runtime` a dependency of `@remix-run/react`, which is currently not the case.

colinhacks: It's not a runtime dependency since `import type` gets stripped out entirely at build time. Put differently, this works regardless of whether `@remix-run/server-runtime` is in the `package.json` of `remix-react`. That said, I only did this avoid avoid a redundant definition of `LoaderFunction` but it's easy enough to duplicate it - is that preferred? I wasn't sure what the preferred pattern was for sharing types across package boundaries.

kiliman:
How about restricting `json<Data>()` type to only extend from `JSONValue`. This will ensure the developer won't send non-JSON values like `Date` or `BigInt`
```ts
type JSONValue =
    | string
    | number
    | boolean
    | { [x: string]: JSONValue }
    | Array<JSONValue>;
 ```

File: packages/remix-react/components.tsx (hunk 1 above)
Comment (MichaelDeBoey) [reply packages/remix-react/components.tsx]: `LoaderFunction` is used in a type that's exported (`DataOrLoader`, which is used in exported `UseLoaderData` type), so that makes `@remix-run/server-runtime` a dependency.

Comment (colinhacks) [reply packages/remix-react/components.tsx]: I've updated the PR to remove this import.

**Comment by colinhacks** [comment]
> How about restricting `json<Data>()` type to only extend from `JSONValue`. This will ensure the developer won't send non-JSON values like `Date` or `BigInt`
> 
> ```ts
> type JSONValue =
>     | string
>     | number
>     | boolean
>     | { [x: string]: JSONValue }
>     | Array<JSONValue>;
> ```

I suggested this in #1254 but I'd want confirmation from the team before I implement it. It's possible it'll break some codebases if the return type of loader is `unknown` or contains `unknown`:

```ts
type JSON =
  | string
  | number
  | boolean
  | {[x: string]: JSON}
  | Array<JSON>;

const arg: JSON = {asdf: 'asdf' as unknown};
//    ^ not assignable to type JSON
```

Not sure how common this is but I don't think it's a big problem. Worth the improved type safety IMO.

itsMapleLeaf:
Can we extend this to work with `useActionData` as well?

I'd suggest `useFetcher`, but that seems a bit more error-prone; it's a lot easier to mismatch the fetcher `action` with the `loader` function you'd have to pass in, but I'd be open to it if others feel different

packages/remix-react/components.tsx
```diff
@@ -1296,7 +1296,45 @@ export function useMatches() {
...
+    }
+  : T extends (infer U)[]
+  ? SerializeType<U>[]
+  : {
+      [k in keyof T as T[k] extends NonJsonPrimitives
+        ? never
+        : k]: SerializeType<T[k]>;
+    };
+
+export type UseDataFunction<T extends DataOrFunction> = T extends (
```
MichaelDeBoey, line 1330: Shouldn't this be called `UseDataFunctionReturn`, since it's the return type of `useActionData` & `useLoaderData`?

MichaelDeBoey (changes requested):

kentcdodds (commented):
Thank you so much for taking the time to implement this @colinhacks. This looks really good to me. I appreciate the added tests as well.

Could you help me understand why we need a `TypedResponse` type? As far as the user (developer) is concerned, they don't ever call `response.json()` themselves, so the fact that it returns a promise of the correct type shouldn't make a difference right?

EDIT: Actually, I think I understand why it's necessary. It's for the `useLoaderData` generic to be able to get the type out of `typeof loader`.

Next question, what happens when the loader returns a `redirect` or a `new Response()`?

packages/remix-react/components.tsx
colinhacks: Changed to your recommendation

colinhacks:
Oops got distracted with travel and weddings! Back to answer 

> Next question, what happens when the loader returns a `redirect` or a `new Response()`?

Good question. If you don't use `return json(...)`, inference isn't possible and the return type of `useLoader<typeof loader>()` will be `any`. This is true for both `return redirect(...)` and `return new Response({ ... })`.

One interesting case: if the user `throws` the redirect (which I believe is recommended?), that code path isn't factored into the loader's inferred return type. This is how TypeScript works, not something I implemented. This is desirable, since you can `throw` redirects without messing up loader type inference.

```ts
export const loader = async (args: LoaderArgs) => {
  const sessionCookie = args.request.headers.get("cookie");

  if (!sessionCookie) {
    throw redirect("/login");
  }

  return json({ ok: true });
};

export default function Home(){
  const data = useLoaderData<typeof loader>();  // { ok: boolean }
}
```

Unfortunately conditionally *returning* a redirect also breaks type inference:

```ts
export const loader = (args: LoaderArgs) => {
  if (Math.random()) return redirect("/");
  return json({ ok: true });
}

export default function Home(){
  const data = useLoaderData<typeof loader>();    // any
}
```

You may have assumed that the return type of this loader is `Response | TypedResponse<{ok: boolean}>` but because `TypedResponse<T>` is a subtype of `Response`, TypeScript collapses this union type down to simply `Response`. And when the loader returns a plain untyped `Response`, `useLoaderData<typeof loader>()` will return `any`. 

For this reason, the docs should always recommend `throw`ing redirects, IMO>

itsMapleLeaf:
Regarding `redirect`, would it work here to return `TypedResponse<never>`? In a union `TypedResponse<never | SomeType>`, the `never` goes away, and you get `TypedResponse<SomeType>`. [I rely on this in my own version of typed helpers.](https://gist.github.com/itsMapleLeaf/0521a21e5e24419ce90d447e7ceede70) I also think it's intuitive; if you navigate away from the page, that data will _never_ actually arrive to the page.

kentcdodds:
So, to be clear, we'd be fine if we do:

```ts
export const loader = (args: LoaderArgs) => {
  if (Math.random()) throw redirect("/");
  return json({ ok: true });
}
```

I think I can live with that. I'm on a family trip right now (just hitting GitHub on some down-time), but when I get the chance I'll talk with Ryan and Michael about this. Thanks for your patience!

colinhacks:
Ooh good idea @itsMapleLeaf! I updated the PR to use that solution. This way, users don't need to remember to `throw` redirects, and type inference will work no matter what. Seems unambiguously better IMO. 

> when I get the chance I'll talk with Ryan and Michael about this. Thanks for your patience

🤙 no rush

joshmedeski:
Thanks, @colinhacks! I took a [rudimentary attempt](https://github.com/remix-run/remix/discussions/2937#discussioncomment-2602237) at this earlier this year and will be glad to see this get merged. 

Let me know if you need any help with testing edge cases.

colinhacks:
> Thanks, @colinhacks! I took a [rudimentary attempt](https://github.com/remix-run/remix/discussions/2937#discussioncomment-2602237) at this earlier this year and will be glad to see this get merged.
> 
> Let me know if you need any help with testing edge cases.

Oh wow, hadn't seen that discussion! I see you also chose a simplified `typeof loader` approach (without `Awaited<ReturnType<...>>`) and renamed `DataFunctionArgs` to `LoaderArgs`. We're definitely on the same wavelength 😛

packages/remix-react/components.tsx
```diff
@@ -1296,7 +1296,45 @@ export function useMatches() {
...
+type NonJsonPrimitives = undefined | Function | symbol;
+type SerializeType<T> = T extends JsonPrimitives
+  ? T
+  : T extends undefined
+  ? undefined
+  : T extends { toJSON(): infer U }
+  ? U
+  : T extends []
+  ? []
+  : T extends [any, ...any[]]
```
MichaelDeBoey, line 1316: Could we use `unknown` instead?

```suggestion
  : T extends [unknown, ...unknown[]]
```

packages/remix-react/components.tsx
```diff
@@ -1296,7 +1296,45 @@ export function useMatches() {
...
+  : T extends (infer U)[]
+  ? SerializeType<U>[]
+  : {
+      [k in keyof T as T[k] extends NonJsonPrimitives
+        ? never
+        : k]: SerializeType<T[k]>;
+    };
+
+export type UseDataFunctionReturn<T extends DataOrFunction> = T extends (
+  ...args: any[]
```
MichaelDeBoey, line 1331:
```suggestion
  ...args: unknown[]
```

MichaelDeBoey (changes requested):

packages/remix-react/components.tsx
colinhacks: I've changed all the other `any`s  to `unknown` except this one. 

This specific change would break the code because TypeScript correctly treats function types are [contravariant](https://www.stephanboyer.com/post/132/what-are-covariance-and-contravariance) in their argument type. In other words, no function type is considered a subtype of `(...args: unknown[])=>any` because `unknown` is not considered a subtype of anything but `any` and `unknown`.

```ts
type myFunc = (arg: string) => boolean;
type test1 = myFunc extends (...args: any[]) => any ? true : false;
// true
type test2 = myFunc extends (...args: any[]) => unknown ? true : false;
// true
type test3 = myFunc extends (...args: unknown[]) => any ? true : false;
// false
```

kentcdodds:
Having talked about this with Pedro, I'm convinced option 1 is the option I "hate the least" so I vote that one.

pcattori:
Summarizing a conversation I had with @ryanflorence about this:

The root cause here is that `loader` is an implicit input for `useLoaderData`. So modeling that would be `useLoaderData(loader)`. But we want to call the loader outside of rendering (that's like the whole point of Remix!), so its called before `useLoaderData`. That means we can't directly infer types from `loader` through `useLoaderData` without a type cast or using `any` somewhere in the data flow from `loader` to `useLoaderData`.

One thought is that for option (3), type purists could still omit the generic to get `unknown` back from `useLoaderData`:
```ts
let data = useLoaderData() // unknown
let product = Product.parse(data) // use `zod` or `invariant` or another runtime checking library to get type safety
```

Unfortunately if we use (2) or (3), for purists that are just following docs/examples that means TS will never warn them that data from `useLoaderData` was never validated/parsed and that they should do so themselves.

```ts
let product = useLoaderData<typeof loader>() // TS won't warn purists that we're never validating this data to make sure it aligns with the return type of `loader`
```

But for semi-purists or users who don't care as much about strict, 100% type safety, it'd be nice if they weren't required to import `InferLoaderData` and do the incantation of `useLoaderData() as InferLoaderData<typeof loader>`.

Plus `loader` is defined _right there_ next to the usage of `useLoaderData` AND usage of the generic seems very error-averse: just do `useLoaderData<typeof loader>`.

---

This all makes me lean toward (3) as long as:
- we allow omission of the generic causing `useLoaderData` to return `unknown`
- we clearly document that purists can omit the generic and use a runtime validation/parsing library to get strict type safety

I _think_ the best place for documenting that the generic can be omitted is in the jsdoc for `useLoaderData` since purists will be prone to look at the definition for `useLoaderData`.

colinhacks:
> we allow omission of the generic causing useLoaderData to return unknown

I believe it's `any` currently, but `unknown` would be a better choice in that `loaderData` would requires explicit type casting before it can really use used for anything. This would break a lot of code though. You can access properties on an `any` value `("asdf" as any).whatever` but you can't with `unknown`.

> we clearly document that purists can omit the generic and use a runtime validation/parsing library to get strict type safety

A true type purist would be using a typesafe ORM in their loader so it would be redundant to revalidate `loaderData` on the client 🙃 But this is definitely a documentable pattern. (Aside: You wouldn't want to validate inline since validation can be CPU-intensive; you'd use `useMemo` here.)

pcattori:
> This would break a lot of code though.

We have some internal project tracking for breaking changes for the eventual Remix v2, so we could keep as `any` and mark that as an action item in the v2 project. That said, its a very minor breaking change since users can just cast with `as any` if we switch to return `unknown` so I'm tempted to make that breaking change, but I'll think about this some more and discuss with the team.

> A true type purist would be using a typesafe ORM in their loader so it would be redundant to revalidate `loaderData` on the client 🙃

Good point! I started thinking of use-cases where I'd still want to revalidate on the client, but couldn't think of a non-contrived one. I'm convinced.

MichaelDeBoey:
> But we want to call the loader outside of rendering (that's like the whole point of Remix!), so its called before `useLoaderData`.

Isn't the "the hook can also return `null`"-problem (so not always the return type from `loader`, because we're calling it async) fixable by making the return type of `useLoaderData` be `null | InferLoaderData<typeof loader>`? 🤔

MichaelDeBoey:
Now that we have `unknown`, we should use it as much as possible instead of `any`.
It's also recommended: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any

If we do that in v2 of before is up to the team, but I personally would start doing that for input types already in v1.
The team can decide if we want to do that for return types as well or if we wait for v2 to do that.

kentcdodds:
After reading the arguments (thanks a ton @colinhacks and @MichaelDeBoey!), here's what I say:

1. Let's go with `useLoaderData<typeof loader>();`
2. Let's stick with `any` by default and switch to `unknown` in v2
3. Let's create a codemod to migrate `useLoaderData();` to `useLoaderData() as any;`

Another thing we should consider is the impact of @jacob-ebey's work on the new `deferred` API on these types (for example: https://github.com/jacob-ebey/stupid-simple-remix-deferred-demo/blob/8dbd5726db7ff8236d48dbd29fed8cbe9d83f803/app/routes/__blog.tsx#L17)

Aside from that, where do we stand in your opinion @colinhacks. Is this PR ready to review/merge as far as you're concerned?

colinhacks:
> Aside from that, where do we stand in your opinion @colinhacks. Is this PR ready to review/merge as far as you're concerned?

Woo! Aside from the lack of a proper solar flare mitigation strategy, I believe this is ready to merge as is :)

-------

Looking at the `deferred` proof-of-concept, I don't think it presents any issues here. It's not hard to detect that the returned value is a "deferred response" and tweak the `useLoaderData` type logic however we need (e.g. append `| null` or whatnot).

kentcdodds:
That's great to hear @colinhacks. Can't tell you how much we appreciate the amount of time you've spent working on this. If you would be willing to work on the documentation for this that would be stellar. @MichaelDeBoey may also be able to help with updating examples and things both code examples in the docs and the examples in the repo (he's very good at that ❤️) so I don't think you need to worry about that in this PR.

If you could update the core documentation we have about TypeScript on our TypeScript page (https://github.com/remix-run/remix/blob/main/docs/guides/typescript.md, https://remix.run/docs/en/v1/guides/typescript) that would be very helpful. Thank you!

MichaelDeBoey:
Since this is a non-breaking change, I think we don't need to update docs/examples in this PR.
It will only make it bloated imo.

I can definitely create a follow-up PR with updated docs/examples if that's wanted, but don't have a problem with @colinhacks doing it as well 🤷‍♂️
Most important is that it's done at some point after this PR is merged.

packages/remix-server-runtime/errors.ts
```diff
@@ -57,14 +57,16 @@ export interface ThrownResponse<T = any> {
...
-}
+};
 
-export async function serializeError(error: Error): Promise<SerializedError> {
+export async function serializeError(error: Error) {
   return {
     message: error.message,
     stack: error.stack,
-  };
+  } as SerializedError;
```
MichaelDeBoey, line 71: Is this type cast necessary?
I think we could keep the `: Promise<SerializedError>` return value, no?

packages/remix-server-runtime/magicExports/remix.ts
```diff
@@ -19,9 +19,11 @@ export type {
   HtmlLinkDescriptor,
   PageLinkDescriptor,
   ErrorBoundaryComponent,
+  ActionArgs,
```
MichaelDeBoey, line 22: We don't have these files anymore on latest `dev`, it's now all configured in root's `rollup.config.js`

packages/remix-server-runtime/responses.ts
```diff
@@ -1,7 +1,26 @@
-export type JsonFunction = <Data>(
+type SerializablePrimitives =
+  | string
+  | number
+  | boolean
+  | null
+  | { toJSON(): any }
```
MichaelDeBoey, line 6: This should be `unknown` I think?

```suggestion
  | { toJSON(): unknown }
```

packages/remix-server-runtime/responses.ts
```diff
@@ -1,7 +1,26 @@
...
+  | Serializable[];
+export type JsonFunction = <Data extends Serializable>(
   data: Data,
   init?: number | ResponseInit
-) => Response;
+) => TypedResponse<Data>;
+
+// must be a type since this is a subtype of response
+// interfaces must conform to the types they extend
+export type TypedResponse<T = any> = Response & {
```
MichaelDeBoey, line 21: Doesn't really matter in this case, but I think it's a good practice to say what type it needs to extend.

Also: let's make it `unknown` as well for the same reason

```suggestion
export type TypedResponse<T extends unknown = unknown> = Response & {
```

MichaelDeBoey (changes requested):

frontsideair:
Awesome job, I can't wait for this to land! 🤩 

One thing I want to suggest that was already mentioned in the thread; a JSON encode/decode will cause some type unsafety. The suggestion was to restrict the return type to be a JSONValue, which would cause backwards incompatibility and cause a burden to the developers.

My suggestion is to use a helper type to convert or omit stuff that wouldn't survive an encode/decode. I recently made [such a helper type myself](https://gist.github.com/frontsideair/dfd696bc927514ba4a78e1c5ed4acd1c) for my own consumption and would be happy to see it get merged. I currently use it like this:

```tsx
type LoaderData = {
  // stuff
};

export const loader: LoaderFunction = async ({ request, params }) => {
  return json<LoaderData>({/* stuff */});
};

  const data = useLoaderData<Jsonify<LoaderData>>();
  // stuff
}
```

I would be happy if I didn't have to remember to wrap `LoaderData` with `Jsonify`. I also [opened an issue](https://github.com/microsoft/TypeScript/issues/48697) to the TypeScript repo since this helper is not very robust right now and it's not really possible in userland.

I'm really excited about this and would love to help if there's anything I can help with.

packages/remix-react/components.tsx
```diff
@@ -1296,7 +1296,45 @@ export function useMatches() {
...
+        ? null
+        : SerializeType<T[k]>;
+    }
+  : T extends (infer U)[]
+  ? SerializeType<U>[]
+  : {
+      [k in keyof T as T[k] extends NonJsonPrimitives
+        ? never
+        : k]: SerializeType<T[k]>;
+    };
```
MichaelDeBoey, line 1328: After reading @frontsideair's comment about his `Jsonify` helper function, I thought about [the one from `type-fest`](https://github.com/sindresorhus/type-fest#json).
Couldn't we use that one instead here? 🤔

frontsideair: I would advise strongly against using `Jsonify` from `type-fest` for now. I tried it and in edge cases it's less robust than the one in my gist. I opened [an issue ](https://github.com/sindresorhus/type-fest/issues/407) with my findings and will contribute my improvements if they decide to go forward with it.

I haven't tried the one in this PR yet, so I have no comments about it.

sindresorhus: https://github.com/sindresorhus/type-fest/releases/tag/v2.15.0

MichaelDeBoey: @colinhacks `type-fest` now has an updated `Jsonify` generic.
https://github.com/sindresorhus/type-fest/releases/tag/v2.15.0

Let's use that one instead of creating our own variant.

kentcdodds: Agreed. That's sufficiently complex that I'd rather not maintain it 😅

Thanks @sindresorhus!

kentcdodds:
Agreed @MichaelDeBoey. We can add docs after merging this.

@colinhacks, let us know how we can help you get this across the finish line. Really eager to get this! `useLoaderData<typeof loader>();` is going to be amazing.

Anyone have an appetite for working on a codemod to take `useLoaderData();` to `useLoaderData() as any;`? Start here: https://github.com/remix-run/remix/tree/main/packages/remix-dev/cli/migrate/migrations

pcattori:
> Anyone have an appetite for working on a codemod to take `useLoaderData();` to `useLoaderData() as any;`? Start here: https://github.com/remix-run/remix/tree/main/packages/remix-dev/cli/migrate/migrations

Since `useLoaderData() as any` will be redundant (same as `useLoaderData()`) until we switch to `unknown` in v2, I think we probably want to hold off on _releasing_ the migration until the switch to `unknown` is in. That said, we can _write_ the migration now and just wait to release it. What do you think @kentcdodds ?

kentcdodds:
Agreed @pcattori 👍

packages/remix-react/components.tsx
colinhacks: There are still some issues with `Jsonify` that cause failures in the type tests I added to `hook-types-types.tsx`. I've fixed the issues with this PR's implementation discovered by @frontsideair and think we should merge as-is. In parallel I'll create another PR to type-fest fixing some of it's issues; once that gets merged I'll submit a final PR to Remix switching over to typefest's `Jsonify`.

This issues with `Jsonify`:

```ts
Jsonify<unknown>;   
// never
// should be unknown (IMO)

Jsonify<symbol>;   
// never
// should be undefined

Jsonify<Function>;   
// never
// should be undefined

Jsonify<[string, number, Date]>;   
// (string | number)[]
// tuple type not preserved

Jsonify<{ arg1: Function}> 
// {arg: null}, should be {arg: never}
// non-serializable values are omitted from objects
```

The `unknown` problem is a major issue and preserving tuple types is a nice-to-have, everything else is pretty minor. 


As for the issues @frontsideair discovered with `SerializeType`:

> Conversion to primitives for Number, String, Boolean

Fixed

> Turning NonJsonPrimitives to null when in array

Fixed

> Rejecting bigint

Fixed, assuming you by "rejecting" you mean assigning to `never`. In some sense in doesn't matter what the static type is in these scenarios because this will throw a runtime error.

> Omitting entries with symbol keys

Didn't bother with this one yet. I'll fix this in my PR to typefest this is quite minor.
> Rejecting plain NonJsonPrimitives

These should actually be `undefined`. The current version implements this correctly.

MichaelDeBoey (changes requested):

packages/remix-react/components.tsx
```diff
@@ -1320,7 +1320,54 @@ export function useMatches(): RouteMatch[] {
...
+    }
+  : T extends (infer U)[]
+  ? (U extends NonJsonPrimitives ? null : SerializeType<U>)[]
+  : T extends object
+  ? {
+      [k in keyof T as T[k] extends NonJsonPrimitives
+        ? never
+        : k]: SerializeType<T[k]>;
+    }
+  : never;
```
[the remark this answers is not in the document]
colinhacks: Discussed here: https://github.com/remix-run/remix/pull/3276#discussion_r914391105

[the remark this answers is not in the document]
colinhacks: I'm fine if you'd rather wait until we can land those fixes in type-fest though I don't have a ton of bandwidth unfortunately. Perhaps @frontsideair would want to take point on that? Seems PRs to typefest get merged in a timely manner.

packages/remix-server-runtime/magicExports/remix.ts
pcattori: @colinhacks : this file should be excluded from this PR as we no longer use `magicExports/remix.ts` files.

pcattori (changes requested):

pcattori (approved):

alavkx:
oh no...now im going to have to re-evaluate my decision to use next + trpc for my upcoming greenfield project. woe is me
great work on this yall. happy to see e2e type safety catching on in the ecosystem. 
image

pcattori:
Ok, this feature is now merged into `dev` along with a [fix for what would've been breaking change](https://github.com/remix-run/remix/pull/3677).

nickmccurdy:
Has that patch been published yet?

kentcdodds:
It's probably in `v0.0.0-nightly-c3ec279-20220708`

lemol:
I think I have found a bug 🐛 

Using the Final API provided in the descriptions of this PR, just include the `return { count: 1 };` in the `loader` function. The `loader` function return type is `Promise<TypedResponse<{ hello: string }> | { count: number }>`. The `data` returned by `useLoaderData<typeof loader>` is expected to be inferred as `{ hello: string } | { count: number }` but it is incorrectly being included with `Response` fields.

```tsx
import {LoaderArgs} from "@remix-run/node";

const loader = async (args: LoaderArgs)=>{
  return { count: 1 };
  return json({ hello: "data" }); // TypedResponse<{hello: string}>
}

export default App(){
  // ⚠️ data is not inferred correctly here
  const data = useLoaderData<typeof loader>()
  return <div>{data}</div>
}
```

kentcdodds:
Looks like this is breaking [`useLoaderData() as LoaderData`](https://github.com/remix-run/remix/runs/7277569011?check_suite_focus=true#step:7:10) which we need to support at least until the next breaking change:

```
Error: app/routes/notes/$noteId.tsx(36,16): error TS2352: Conversion of type '{ note: { id: string; title: string; body: string; createdAt: string; updatedAt: string; userId: string; }; }' to type 'LoaderData' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

How can we make sure we don't break that?

lemol:
> @lemol you mean you have two return statements in your loader? Why would you do that?

That was just to illustrate when there is a mix of returned types. Actual useful example may be returning conditionally:

```ts
const loader = async (args: LoaderArgs)=>{
  if (some condition) {
    return { count: 1 };
  }

  return json({ hello: "data" }); // TypedResponse<{hello: string}>
}
```

joshmedeski:
@lemol I would consider the bug in this example to be not returning the `count` object in the `json` object. It properly types the function if you fix it.

```diff
const loader = async (args: LoaderArgs)=>{
  if (some condition) {
-    return { count: 1 };
+    return json({ count: 1 });
  }

  return json({ hello: "data" }); // TypedResponse<{count: number} | {hello: string}>
}
```

joshmedeski:
P.S. @colinhacks this works really well for me on the dev branch, can't wait to see this released! 🥳

frontsideair:
Good catch @joshmedeski, just came to say the exact thing!

@kentcdodds This is interesting. The loader data that the hook returns wouldn't have a `Date` type, so I'm not sure if it's a breaking change or a fix. The error is this:

```
The types of 'note.createdAt' are incompatible between these types.
    Type 'string' is not comparable to type 'Date'.
```

which is correct, depending on `note.createdAt` to be a `Date` would actually be an error. If there are people who depend on `useLoaderData() as LoaderData`, this is a time bomb. The correct usage would be `useLoaderData() as SerializeType<LoaderData>`.

But real life is messy and it's not great for people to update Remix and have type errors (even for wrong things). I guess it would be a breaking change that could wrap `LoaderData` with `SerializeType` as a codemod. 🤷

kentcdodds:
@joshmedeski is correct. This is a new feature which you opt-into so it's non-breaking. We can say: "if you want to use this feature (`typeof loader`) then you must use `json`" which we want to encourage anyway and in v2 we'll probably remove the ability to return a plain object.

kentcdodds:
@frontsideair

> real life is messy

Correct. It's embarrassing that I didn't notice that in the actual error and also that we've gone so long having that time bomb in our starters.

I think what we should do in this situation is push forward and we'll call out this in the notes. In a world where bug fixes can break people's CI, every new version could be considered a major version bump. I don't want to live in a world like that, especially when it's just CI and not production code since it's just TypeScript compilation that will be affected.

kentcdodds:
Ah, I didn't realize that we won't be using `LoaderFunction` anymore. At first I wasn't excited about this. I'm still not. But I'm willing to go with it 👍

Thanks 🎉

joshmedeski:
@kentcdodds I had to do the same thing in my original solution. I don't see much benefit from `LoaderFunction` other than typing the arguments (since the return type is just `any` or `Promise<any>`) but `LoaderArgs` does the same thing.🤷

theMosaad:
Should this approach provide an easy type for `data` that's [passed to the Meta function](https://remix.run/docs/en/v1/api/conventions#page-context-in-meta-function)?

Having a hard time narrowing this type

```ts
type LoaderData = Awaited<ReturnType<typeof loader>> // Response & { json(): Promise<{ name: string; }>; }
```

kentcdodds:
@theMosaad I think you're looking for this:

```
type LoaderData = UseDataFunctionReturn<typeof loader>
```

:)

But yeah, I think we could probably do something to help with this in the `MetaFunction` type 🤔

kubahasek:
Hi! Is there any official documentation for this or only in this PR? Thanks!

machour:
@kubahasek not yet, but it's being worked on. All the docs/examples/stacks should get an update soon.

tshddx:
> @theMosaad I think you're looking for this:
> 
> ```ts
> type LoaderData = UseDataFunctionReturn<typeof loader>
> ```
> 
> :)
> 
> But yeah, I think we could probably do something to help with this in the `MetaFunction` type 🤔

@kentcdodds This is working great for me! Is it normal/expected that I should `import { UseDataFunctionReturn } from '@remix-run/react/dist/components'`? It seems like the type is only available via `dist` because [@remix-run/react/index.tsx](https://github.com/remix-run/remix/blob/c5d6cd276b7446747339abdc0f66360617086fea/packages/remix-react/index.tsx#L22-L30) doesn't re-export it from the `components` module.

kentcdodds:
Oh, we should definitely export that. PR welcome! 😅

MichaelDeBoey:
@kentcdodds PR to export `UseDataFunctionReturn` type: #3764

packages/remix-react/components.tsx
tshddx: > > > Rejecting plain NonJsonPrimitives
> 
> > These should actually be undefined. The current version implements this correctly.
> 
> I'm not sure about this. Stringifying any of these returns `undefined` and `JSON.parse` throws, similar to the `bigint` case, so my argument applies here as well. But I'm open to discussion, I'd like to understand why would you like these to be `undefined`?

@frontsideair It's true that an actual `undefined` literal, e.g. inside an array like `[1, 2, undefined]` won't make it back from JSON.stringify and JSON.parse (in that example you'd get `[1, 2, null]`, not a thrown error). But shouldn't simple optional fields in an object be supported? It appears that `undefined` is even being stripped from union types, so an optional field returned from `loader` will come back from `useLoaderData` as if it's always defined, which will very easily lead to a runtime error. For example:

```ts
import { json } from '@remix-run/node';
import { useLoaderData } from '@remix-run/react';

export const loader = () => {
  const user = Math.random() < 0.5 ? { username: 'jo', id: 123 } : undefined;
  return json({ user });
};

  // TypeScript thinks `data.user` is always defined,
  // thus this is a runtime error!
  return <h1>Hello {data.user.username}</h1>;
}

```

This seems like a fairly critical bug in 1.6.5 that can very easily lead to runtime errors.

PR #3281, opened 2022-05-21
build(templates/netlify): remove unused `@netlify/functions` dependency

pcattori:
template code does not use it and it is not a peer dependency. `@remix-run/netlify` is the only
thing that depends on `@netlify/functions`, but it already declares it as a dependency.

### DIFF
--- a/templates/netlify/package.json
+++ b/templates/netlify/package.json
@@ -10,7 +10,6 @@
     "start": "cross-env NODE_ENV=production netlify dev"
   },
   "dependencies": {
-    "@netlify/functions": "^1.0.0",
     "@remix-run/netlify": "*",
     "@remix-run/node": "*",
     "@remix-run/react": "*",


MichaelDeBoey:
It's in `@remix-run/netlify`'s `peerDependencies`

https://github.com/remix-run/remix/blob/d19d888a6dd191a84d1defb1003d1004bc9c101f/packages/remix-netlify/package.json#L17-L19

pcattori:
> It's in `@remix-run/netlify`'s `peerDependencies`
> 
> https://github.com/remix-run/remix/blob/d19d888a6dd191a84d1defb1003d1004bc9c101f/packages/remix-netlify/package.json#L17-L19

Yea you're right, strange that `npm install` doesn't show a warning for missing peer dependencies when I remove it from `package.json`, delete `node_modules/` and `package-lock.json` and reinstall. `npm ls` also shows it coming from `@remix-run/netlify` in this case:

```sh
❯ npm ls @netlify/functions
project@ /path/to/project
└─┬ @remix-run/netlify@1.5.1
  └── @netlify/functions@1.0.0
```

In any case, it _is_ a peer dependency, so I'm closing this.

PR #3284, opened 2022-05-22
feat: deprecate all functions & types exported from `magicExports`

MichaelDeBoey:
I wanted to re-submit #2735 again, as:

- The first release supporting separate packages was `1.3.3`, which was released on 23 Mar 2022
  https://github.com/remix-run/remix/releases/tag/v1.3.3
- The `replace-remix-imports` migration was added in `1.4.0`, which was released on 14 Apr 2022
  https://github.com/remix-run/remix/releases/tag/v1.4.0
- We now export functions/types that aren't available anymore through the `remix` package, as they're not included in the `magicExports` anymore (see both `@remix-run/cloudflare` & `@remix-run/node`) since `1.5.0`
- We now have a runtime package (`@remix-run/deno`) that doesn't have *any* magic exports *at all* since `1.5.0`
https://github.com/remix-run/remix/releases/tag/v1.5.0
- Our ESLint config is warning with a deprecated message when importing from the `remix` package since `1.6.0`
https://github.com/remix-run/remix/releases/tag/v1.6.0

---

Original PR description:

> As a follow-up of #2731, this will show deprecation warnings in the IDE + in console whenever people aren't using the ESLint config.
> 
> Inspired by @pcattori's remark (https://github.com/remix-run/remix/pull/2542#discussion_r838804227) on @mjackson's PR to rename `createCloudflareKVSessionStorage` to `createWorkersKVSessionStorage` (#2542). 
> 
> Since each package is on it's own, I had to copy/paste the `warn` & `getDeprecatedMessage` functions over to each package.
> 
> If this somehow could be simplified (as this is doing the same thing over & over again), I'd be happy to do so.

PR #3285, opened 2022-05-22
feat(remix-eslint-config): enable deprecated imports warnings

MichaelDeBoey:
Companion PR of #3284

I think we can enable these again, as:

> - The first release supporting separate packages was `1.3.3`, which was released on 23 Mar 2022
>   https://github.com/remix-run/remix/releases/tag/v1.3.3
> - The `replace-remix-imports` migration was added in `1.4.0`, which was released on 14 Apr 2022
>   https://github.com/remix-run/remix/releases/tag/v1.4.0
> - We now export functions/types that aren't available anymore through the `remix` package, as they're not included in the `magicExports` anymore (see both `@remix-run/cloudflare` & `@remix-run/node`) since `1.5.0`
> - We now have a runtime package (`@remix-run/deno`) that doesn't have *any* magic exports *at all* since `1.5.0`
> https://github.com/remix-run/remix/releases/tag/v1.5.0

### DIFF
--- a/packages/remix-eslint-config/rules/core.js
+++ b/packages/remix-eslint-config/rules/core.js
@@ -1,45 +1,45 @@
-// const {
-//   defaultAdapterExports,
-//   defaultRuntimeExports,
-//   architectSpecificExports,
-//   cloudflareSpecificExports,
-//   cloudflarePagesSpecificExports,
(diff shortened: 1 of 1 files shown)


kentcdodds:
Thanks!

PR #3287, opened 2022-05-22
chore(templates): remove `description`, `license` & `name` keys from `package.json`

MichaelDeBoey:
As mentioned by @pcattori in https://github.com/remix-run/remix/pull/3221#discussion_r876255891, these fields aren't necessary for apps.

### DIFF
--- a/templates/arc/package.json
+++ b/templates/arc/package.json
@@ -1,8 +1,5 @@
 {
-  "name": "remix-template-arc",
   "private": true,
-  "description": "",
-  "license": "",
   "sideEffects": false,
   "scripts": {
     "build": "remix build",
--- a/templates/cloudflare-pages/package.json
+++ b/templates/cloudflare-pages/package.json
@@ -1,8 +1,5 @@
 {
-  "name": "remix-template-cloudflare-pages",
   "private": true,
-  "description": "",
-  "license": "",
   "sideEffects": false,
   "scripts": {
     "build": "remix build",
--- a/templates/cloudflare-workers/package.json
+++ b/templates/cloudflare-workers/package.json
@@ -1,8 +1,5 @@
 {
-  "name": "remix-template-cloudflare-workers",
   "private": true,
-  "description": "",
-  "license": "",
   "sideEffects": false,
   "main": "build/index.js",
   "scripts": {
--- a/templates/deno/package.json
+++ b/templates/deno/package.json
@@ -1,5 +1,4 @@
 {
-  "name": "remix-template-deno",
   "private": true,
   "sideEffects": false,
   "scripts": {
--- a/templates/express/package.json
+++ b/templates/express/package.json
@@ -1,8 +1,5 @@
 {
-  "name": "remix-template-express",
   "private": true,
-  "description": "",
-  "license": "",
   "sideEffects": false,
   "scripts": {
     "build": "remix build",
--- a/templates/fly/package.json
+++ b/templates/fly/package.json
@@ -1,8 +1,5 @@
 {
-  "name": "remix-template-fly",
   "private": true,
-  "description": "",
-  "license": "",
   "sideEffects": false,
   "scripts": {
     "build": "remix build",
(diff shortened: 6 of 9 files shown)


pcattori:
`sideEffects` is actually required for bundling to work as expected. Setting `sideEffects: false` in an app's `package.json` marks all imports from that app (e.g. relative imports) as side-effect free.

MichaelDeBoey:
@pcattori I've added `sideEffects` back to `package.json` 👍

pcattori:
@MichaelDeBoey `name` should also be removed.

MichaelDeBoey:
@pcattori Since each app has a certain name, I think we should keep `name` for convenience.

pcattori:
> @pcattori Since each app has a certain name, I think we should keep `name` for convenience.

I'm not sure I follow. What convenience does the `name` provide?

MichaelDeBoey:
If I'm not mistaken (@mcansh can probably provide more info about this), tests will also fail if examples/templates have the same/no `name` 🤔

pcattori:
> patchup-example-names

If we already have a script for handling `name` collisions, that's a smell that `name` isn't the right thing for the job in the first place. We should instead just use the filename in the `patchup-example-names` script to populate the `name` field dynamically for the examples/tests. That would guarantee uniqueness (due to filesystem), whereas the `name` field does not prevent duplicate names.

MichaelDeBoey:
Thinking about this a bit more and maybe we should just keep these fields, as `npm` will warn the following otherwise 🤔

```sh
npm WARN example-app No description
npm WARN example-app No repository field.
npm WARN example-app No license field.
```

pcattori:
> Thinking about this a bit more and maybe we should just keep these fields, as `npm` will warn the following otherwise 🤔
> 
> ```shell
> npm WARN example-app No description
> npm WARN example-app No repository field.
> npm WARN example-app No license field.
> ```

Does that app have `private: true`?

MichaelDeBoey:
> Does that app have `private: true`?

It was indeed missing that one, my bad 😅🙈

MichaelDeBoey:
> I'm not sure I follow. What convenience does the `name` field provide?

@pcattori People are used to have a `name` in all their `package.json`s, so I think we'll get PRs to add it (again) if we decide to remove it.

`create-react-app` & `create-next-app` also add `name` to `package.json` btw.
https://github.com/facebook/create-react-app/blob/c20ccecfa1cf130a47d37908dc54959c618ce8ea/packages/create-react-app/createReactApp.js#L266-L270
https://github.com/vercel/next.js/blob/863db9b0e2a7b038348202c040f90f050fd12855/packages/create-next-app/create-app.ts#L202-L212

mcansh:
tests won't fail without the name, we'll just get bombarded with jest haste warnings

pcattori:
> > I'm not sure I follow. What convenience does the `name` field provide?
> 
> @pcattori People are used to have a `name` in all their `package.json`s, so I think we'll get PRs to add it (again) if we decide to remove it.
> 
> `create-react-app` & `create-next-app` also add `name` to `package.json` btw. https://github.com/facebook/create-react-app/blob/c20ccecfa1cf130a47d37908dc54959c618ce8ea/packages/create-react-app/createReactApp.js#L266-L270 https://github.com/vercel/next.js/blob/863db9b0e2a7b038348202c040f90f050fd12855/packages/create-next-app/create-app.ts#L202-L212

I don't think we'll get PRs to add it back in. People can just add their app name if they want to their app's `package.json`. We don't have an issue with that. But the template itself shouldn't include a name.

`create-react-app` and `create-next-app` also set `name` to be the _app_ name, not the _template_ name.

MichaelDeBoey:
> They would have needed to edit it anyway if we had the template name set as the `name`, so omitting the name not any more onerous than what we currently have (the template name).

Good point!
Would it be a good idea to automatically add the app name as `name` instead?

pcattori:
> > They would have needed to edit it anyway if we had the template name set as the `name`, so omitting the name not any more onerous than what we currently have (the template name).
> 
> Good point! Would it be a good idea to automatically add the app name as `name` instead?

I think its better to leave it out. Keeps `create-remix` simpler and makes it clear that its not required.

pcattori (approved):

PR #3289, opened 2022-05-22
tests: remove `description` & `license` keys from `package.json`

MichaelDeBoey:
Follow-up of #3287


kentcdodds (approved):
Does this need to be a draft? It's good to go IMO 👍

kentcdodds (approved):
Thanks!

PR #3296, opened 2022-05-24
feat(remix-serve): allow passing `getLoadContext` to `createApp`

TillaTheHun0:
This PR allows passing `getLoadContext` into `@remix-run/express` from `@remix-run/serve`. It doesn't change current functionality, just exposes `getLoadContext` on `@remix-run/serve`

The use case is that as a developer, I want to programmatically invoke `createApp`, to get the built `express` app, and then manually start the server, or even modify the app further ie. more routes, middleware, etc. Allowing to pass `getLoadContext` seemed kosher to me.

This of course would not be available via the `remix serve` CLI.

An alternative approach would be for a developer to reimplement what `@remix-run/serve` is doing, if they want to provide a `getLoadContext`. Though I would rebut that since `@remix-run/serve` interfaces with `@remix-run/dev` and wires all of the Remix build outputs into express for the developer, I think exposing here would help avoid potential "issue bloat" from folks implementing a million different ways. They could just use `@mirex-run/serve`.

- [ ] Docs
- [ ] Tests

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -332,6 +332,7 @@
 - therealflyingcoder
 - thomasheyenbrock
 - thomasrettig
+- TillaTheHun0
 - tjefferson08
 - tombyrer
 - toyozaki
--- a/packages/remix-serve/index.ts
+++ b/packages/remix-serve/index.ts
@@ -2,8 +2,13 @@ import express from "express";
 import compression from "compression";
 import morgan from "morgan";
 import { createRequestHandler } from "@remix-run/express";
+import type { GetLoadContextFunction } from "@remix-run/express";
 
-export function createApp(buildPath: string, mode = "production") {
+export function createApp(
+  buildPath: string,
+  mode = "production",
+  getLoadContext?: GetLoadContextFunction
+) {
   let app = express();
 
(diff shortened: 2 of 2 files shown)


kiliman:
Hmm.. it seems like if you need to use context, it would be simpler to just use the Express adapter. Remix App Server is meant to be a batteries-include no-config server.

You can eject from RAS to Express using this script: https://github.com/kiliman/rmx-cli#eject-ras

ZipBrandon:
Is this accomplishing the same objective as https://github.com/remix-run/remix/pull/2045 and https://github.com/remix-run/remix/pull/1606 ?  If so, then #2045 is the most conditioned PR to merge.  Hopefully that will get merged shortly now that Remix Conf is in the rearview.

chaance:
Echoing @kiliman's sentiments here, based on your use-case this is exactly why we have our lower-level Express adapter. `remix-serve` is designed to get you up-and-running, but it isn't meant to satisfy the need for load context and high levels of customization. I generally expect that most apps will ultimately end up with the need to call `createRequestHandler` in their own server, and it sounds like you're already there 🙂

PR #3299, opened 2022-05-24
feat(remix-dev): add `.wasm` as a supported file type

benwis:
This is a truly massive change, that tells esbuild to bundle .wasm files with the file loader type in both the browser and server bundles.  Big thanks to @jacob-ebey for helping me debug this at the Remix Conf bar!
## Discussion
There was a brief discussion [here](https://github.com/remix-run/remix/discussions/2752), started by @seanaye.
## Testing
If you're using Rust and wasm-bindgen, you'll want to bundle your Rust code with `wasm-pack --target web`. Then once you've done that, you'll want to import both the wasm file itself, the auto generated init() function, and the name of any functions you've made that you'd like to use within the route or component.
```ts
import init, {greet} from "../../browser-rust-functions/pkg/browser-rust-functions"
```
Once you have it loaded, you need to instantiate your WASM, which loads it into the browser's memory. You can either do it in your entry.client.tsx like this
```ts
import wasm from "../../browser-rust-functions/pkg/browser-rust-functions_bg.wasm"
import init from "../../browser-rust-functions/pkg/browser-rust-functions"
init(wasm).then(() => {
    hydrate(<RemixBrowser />, document);
})
```
You may want to make wasm loading conditional to pages that contain it, as the above method does it on every page;
Then you can add a useEffect to your route or component to call the funtions you've created.
```ts
import {greet} from "../../browser-rust-functions/pkg/browser-rust-functions"
useEffect(() => {
      greet();
  }, [])
```
Alternatively you can do it all in one step inside your route
```ts
import wasm from "../../browser-rust-functions/pkg/browser-rust-functions_bg.wasm"
import init, {greet} from "../../browser-rust-functions/pkg/browser-rust-functions"
useEffect(() => {
init(wasm).then(() => {
      greet();
})
  }, [])
```
Feel free to check out a working example of this in my demo repo [here.](https://github.com/benwis/rust-remix)


MichaelDeBoey (approved):

packages/remix-dev/modules.ts
```diff
@@ -68,6 +68,10 @@ declare module "*.ttf" {
   let asset: string;
   export default asset;
 }
+declare module "*.wasm" {
```
jacob-ebey, line 71: Will these need to be copied to the server output directory for node or will they be loaded from the node_modules folder?

shripadk:
Can this pull request please be merged? I am stuck because an authorization library I use imports wasm module which is pretty much blocking using Remix at this point.

JSanchezIO:
Is work still being done on this? I'd like to leverage what this aims to solve. Not trying to rush this, mostly just curious about any blockers or if we should be seeing if anyone else has spare time to assist.

benwis:
As far as I know, there's nothing that needs to be done here, other than merging the change.
WASM support pretty much just needs .wasm added to the compiler and defined as a module.

We're using this patched into Remix now, but would love to have it in the next release.
------- Original Message -------

JSanchezIO:
Should we @ anyone in particular to get this merged?

JSanchezIO:
@MichaelDeBoey are you able to guide us on getting this merged?

chaance:
@benwis I think you inadvertently merged the `main` branch into here which added a bunch of unrelated commits. Any chance you could back out of that so I can get this merged?

benwis:
Yeah, I should be able to. Give me a bit and I'll see what I can do

benwis:
@chaance I have removed that merge and things should be good to go.

chaance (approved):

benwis:
YAY! Thanks @chaance!

machour:
Published in 1.8.0

swamidass:
What about server side WASM? Does that work? 

E.g...

https://github.com/vercel/og-image

PR #3315, opened 2022-05-27
chore: run `sort-package-json` on packages

MichaelDeBoey:
Making it consistent on all packages & in line with how it's done for templates & examples.


kentcdodds (approved):
Thanks 👍

PR #3326, opened 2022-05-29
docs(tutorial/blog): Add guidance in case of GitHub auth error

HenokT:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -363,3 +363,4 @@
 - zachdtaylor
 - zainfathoni
 - zhe
+- henokt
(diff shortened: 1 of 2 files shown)


kentcdodds (changes requested):
This should not happen. Can you provide reproduction steps?

HenokT:
> This should not happen. Can you provide reproduction steps?

@kentcdodds I it happened to me because I had an expired  `GITHUB_TOKEN` environment variable in my _~/.bash_profile_ and had to override it with an inline valid token. If I first unset the expired token then the `create-remix` command works as is:

```sh
unset GITHUB_TOKEK 
npx create-remix --template remix-run/indie-stack blog-tutorial
```

kentcdodds:
Glad you figured out the issue.

PR #3329, opened 2022-05-29
docs(examples): update `.env.example` of `firebase-auth-firestore` example

johnpolacek:
This is an update to `.env.example` for the Firebase Auth Example that makes it more clear what goes where and how to format.

- [ ] Docs
- [ ] Tests
- [x] Examples


kentcdodds (approved):
Yes, I definitely appreciate this 👍 Thank you!

Issue #3120, opened 2022-05-07
React Markdown is Broken On Remix


mehulmpt:
### What version of Remix are you using?

1.4.3

### Steps to Reproduce

- ReactMarkdown component errors out with the following error when deploying to Cloudflare Pages + Remix

![Screenshot 2022-05-07 at 3 27 10 AM](https://user-images.githubusercontent.com/10388889/167221485-8cfcdceb-dcc0-4568-8ae8-0b2afefddc07.png)

Steps:
1. Clone this repo: https://github.com/mehulmpt/react-markdown-cloudflare-remix (this repo is Remix barebones TS + Cloudflare Pages starter template)
2. `npm install` and `npm run dev`
3. You'll see this error:

<img width="749" alt="Screenshot 2022-05-07 at 3 34 24 AM" src="https://user-images.githubusercontent.com/10388889/167222075-4db5a824-0856-4376-aa31-e241916d97d1.png">

It seems like Remix ships wrong node polyfills: https://github.com/remarkjs/react-markdown/issues/688#issuecomment-1120053163

### Expected Behavior

ReactMarkdown should work fine

### Actual Behavior

It crashes with an error

MichaelDeBoey:
This seems to be a problem with [`@esbuild-plugins/node-modules-polyfill`](https://github.com/remorses/esbuild-plugins), which is actually relying on [`rollup-plugin-node-polyfills`](https://github.com/ionic-team/rollup-plugin-node-polyfills).

- https://github.com/remorses/esbuild-plugins/blob/53b216168bb480db8ccd7887506763924175c670/node-modules-polyfill/src/polyfills.ts#L45-L48
- https://github.com/ionic-team/rollup-plugin-node-polyfills/blob/9b5fe1a9cafffd4871e6d65613ed224f807ea251/polyfills/url.js

rgdelato:
I'm using a different library, but I'm seeing effectively this same issue when using Remix with Cloudflare Pages.

It looks like `rollup-plugin-node-polyfills` is no longer maintained, though this fork seems to be carrying the torch: https://github.com/FredKSchott/rollup-plugin-polyfill-node

From what I can tell, there doesn't seem to be a great way to hook into Remix's build system to swap out this dependency on the user's side in the short term.

mwawrusch:
So on netlify I get this:

```
[...]

PR #3330, opened 2022-05-29
fix(remix-dev): set `esbuild`'s `serverPlatform` to `browser` for Cloudflare & Deno

rgdelato:
Setting esbuild [platform](https://esbuild.github.io/api/#platform) to "browser" for Cloudflare and Deno to fix errors with using various libraries on those platforms.

Closes: #3120, https://github.com/algolia/react-instantsearch/issues/3547

For additional context, this setting was originally set to `neutral` in #1504

- [ ] Tests

Testing Strategy:
I'm having some difficulty writing a test for this. I've been able to locally run two playground apps that are fixed by this change (one that imports `react-markdown` and one that imports `algoliasearch`), but I would appreciate some help with writing an integration test!


RyKilleen:
Looking forward to this landing! Just ran into it in the context of trying to use Sanity

schpet:
is it possible to use these changes before this is merged? i.e. by adding something to remix.config.js? or do i need to run a fork of remix until this is merged?

machour:
@schpet There is no configuration in `remix.config.js` that would use this PR, but you can use something like https://www.npmjs.com/package/patch-package to achieve what you want

schpet:
@machour thanks so much for this info!

with [this patch](https://gist.github.com/schpet/9b74730b54eb5151173ce4e9c759ed2a) i was able to successfully get react-markdown working on remix deployed to cloudflare pages (and locally via wrangler)

apalumbo:
@MichaelDeBoey sorry for tagging you directly, from the source code it seems that you are the last one that worked on the file, it would be great if this fix can be integrated . Thank you

acoreyj:
I'm honestly a bit concerned this has been an issue for so long, is there a better way to flag important community pull requests to the Remix team?

MichaelDeBoey (commented):
I'm honestly not sure if we should support `"browser"` as value to `serverPlatform` as [the ESbuild `platform` docs](https://esbuild.github.io/api/#platform) themselves are saying we should set it to `"node"` when we're generating code for the (Node) server.

> If your bundled code is intended to run in node instead, you should set the platform to `node`

@rgdelato @RyKilleen @schpet @apalumbo @acoreyj I think the solution for #3120, algolia/react-instantsearch#3547, nextauthjs/next-auth#6270, Sanity, ... would be to wrap your components in something like @sergiodxa's [`remix-utils`](https://github.com/sergiodxa/remix-utils)' [`ClientOnly` component](https://github.com/sergiodxa/remix-utils#clientonly).

acoreyj:
> set it to "node" when we're generating code for the (Node) server.

Hey @MichaelDeBoey for more details I think the issue stems from how [use V8 but not node](https://developers.cloudflare.com/workers/learning/how-workers-works/), they do implement a lot of web apis so it seems it's more similar to deno than node? I'm guessing that's why remix currently the esbuild to "neutral" instead of "node" for cloudflare (same as for deno).

The error that happens though is because the wrangler package for building to cloudflare uses a package "node-modules-polyfills" which when esbuild is in neutral can't seem to properly find the exports. So right now in my project if esbuild is "neutral" I get the error

```
✘ [ERROR] No matching export in "node-modules-polyfills:crypto" for import "createHmac"

    node_modules/@panva/hkdf/dist/node/esm/runtime/fallback.js:1:9:
      1 │ import { createHmac } from 'crypto';
        ╵          ~~~~~~~~~~
```

But if it is set to "browser" it works.

MichaelDeBoey:
Tagging @evanw (the creator of ESbuild) to see if ESbuild's `platform` should support `worker`/`deno` as a value or if it's intended to use `browser` in these cases. 🤔

rgdelato:
@MichaelDeBoey As far as I can tell, one of the main issues is that these packages specify various entry points for different platforms in their package.json ([here's an example](https://github.com/facebook/react/blob/main/packages/react-dom/package.json#L48-L85) from `react-dom`, which uses both the `exports` and `browser` versions of specifying their exports (also here's [an associated PR](https://github.com/facebook/react/pull/23252))).

My understanding is that Cloudflare wants to resolve "worker" or "browser" exports when they're specified. I think esbuild's [conditions](https://esbuild.github.io/api/#conditions) can cover the `exports` version and [main fields](https://esbuild.github.io/api/#main-fields) can resolve the top-level `browser` version.

Maybe esbuild should support `worker` as a platform value, not sure there, but in the meantime, if we don't want to set platform to "browser", would it make sense to instead have a config like `{ mainFields: ["browser", "module" , "main"], conditions: ["worker", "browser", "module"] }`?

silence48:
is this still an issue??

MichaelDeBoey:
Hi @rgdelato!

If so, please rebase onto latest `dev` & resolve conflicts.
I would also suggest to update the templates with the necessary changes as setting `serverBuildTarget` is now deprecated and will be removed in v2

rgdelato:
Apologies for forgetting about this, I was on an anniversary trip! But also, I wrote this PR like a year ago, and I would gladly defer to anyone else who has more familiarity with the codebase. (If not, I might try to take another look at this issue in a couple weeks.)

Issue #3327, opened 2022-05-29
[Bug]: Layout route with more than one pathless layout causes invalid conflict


pokedotdev:
### What version of Remix are you using?

1.5.1

### Steps to Reproduce

Create a route structure like so:
```
app/routes
├── slug
│  ├── __bar
│  │  └── bar.tsx
│  ├── __foo
│  │  ├── foo.tsx
│  │  └── index.tsx
│  ├── __bar.tsx
│  └── __foo.tsx
└── slug.tsx
```

### Expected Behavior

Layout routes can have more than one pathless layout route.

### Actual Behavior

The error is thrown when starting remix:
```
Error: Path "slug/" defined by route "routes/slug/__foo" conflicts with route "routes/slug/__bar"
```
[You can see reproduction here](https://stackblitz.com/edit/github-3s7qm8?file=app/routes/$slug.tsx)

goldo:
Hi, I can confirm this, I have the same problem. (I wouldn't flag it as "feat" tho, it looks like a bug to me)

binajmen:
I have the same problem. @machour, could you flag it as bug instead?

machour:
Will be fixed by #4421

brophdawg11:
#4421 is merged and should be available once `v1.17.0` is released

PR #3331, opened 2022-05-30
fix(remix-dev): support multiple nested pathless layout routes

pokedotdev:
Closes: #3327

- [ ] Docs
- [x] Tests


packages/remix-dev/config/routesConvention.ts
```diff
@@ -62,13 +62,16 @@ export function defineConventionalRoutes(
...
 
-      let isIndexRoute = routeId.endsWith("/index");
+      let isIndexRoute = lastSegment === "index";
+      let isLayoutRoute = lastSegment.startsWith("__");
       let fullPath = createRoutePath(routeId.slice("routes".length + 1));
-      let uniqueRouteId = (fullPath || "") + (isIndexRoute ? "?index" : "");
+      let uniqueRouteId = fullPath || "";
+
+      if (isIndexRoute) uniqueRouteId += "?index";
+      else if (isLayoutRoute) uniqueRouteId += lastSegment;
```
MichaelDeBoey, line 74: Let's make this less error-prone

```suggestion
      if (isIndexRoute) {
        uniqueRouteId += "?index";
      } else if (isLayoutRoute) {
        uniqueRouteId += lastSegment;
      }
```

MichaelDeBoey (changes requested):

apalumbo:
@MichaelDeBoey is this PR still to approve?

brophdawg11:
Oh interesting, I didn't know we had a separate PR for this. This is also fixed by #4421 which was initially addressing #3014 but uncovered this same issue during testing.  I think I lean towards the solution in #4421 since rather than mess with the `id` just to make it unique, we skip the check entirely for pathless layout routes which, by definition, create the potential for path collisions.  I'll check with the team as well and we'll get one of these merged to fix #3327.  

It may make sense to keep the integration tests from this PR even if we go with #4421 - I'll take a closer look and make sure we're properly covered.

integration/multiple-pathless-routes-test.ts
```diff
@@ -0,0 +1,91 @@
...
+});
+
+test.afterAll(async () => appFixture.close());
+
+test("slug index & foo layout", async () => {
+  let res = await fixture.requestDocument("/slug");
+  let text = await res.text();
+  expect(text).toMatch(TEXTS.SLUG_LAYOUT);
+  expect(text).toMatch(TEXTS.FOO_LAYOUT);
+  expect(text).toMatch(TEXTS.SLUG_LAYOUT);
```
Vivalldi, line 76:
```suggestion
  expect(text).toMatch(TEXTS.SLUG_CHILD);
```

Vivalldi (commented):
Noticed a small thing regarding the `slug index & foo layout` test

chaance:
This got a bit outdated -- very sorry it took us a bit to follow up. As @brophdawg11 mentioned this should have been taken care of already, and most of the logic has been moved to React Router. If there is still an issue I'd suggest opening one in that repo so we can give it a fresh look.

PR #3338, opened 2022-05-30
feat(remix-dev): load env vars sooner

herecydev:
Move loading of .env file to top of dev function, which allows access to `process.env` as early as possible. This more closely resembles conventions of environment variables by being available sooner and permits using the variables within the `remix.config.js` file.

A good use case might be that a developer needs environment variables to influence programmatic creation of the "routes" within `remix.config.js`

Testing Strategy:

Ran locally with `remix dev` to confirm correct behavior.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -126,6 +126,7 @@
 - hardingmatt
 - helderburato
 - HenryVogt
+- herecydev
 - hicksy
 - himorishige
 - hkan
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -249,10 +249,14 @@ export async function dev(remixRoot: string, modeArg?: string) {
     );
   }
 
+  if (!remixRoot) {
+    remixRoot = process.env.REMIX_ROOT || process.cwd();
+  }
+
+  await loadEnv(path.resolve(remixRoot));
+
   let config = await readConfig(remixRoot);
   let mode = isBuildMode(modeArg) ? modeArg : BuildMode.Development;
-
-  await loadEnv(config.rootDirectory);
 
   let port = await getPort({
     port: process.env.PORT ? Number(process.env.PORT) : makeRange(3000, 3100),


MichaelDeBoey (approved):

kiliman:
I think the reason is that Remix treats environment variables as *runtime* variables and not *buildtime* variables. The exception is `NODE_ENV` as so many tools depend on it. 

What kind of values would you depend on at build time?

herecydev:
I've outlined an example [here](https://github.com/remix-run/remix/discussions/3341)

sergiodxa:
I'm not sure but I think `remix dev` code will not run with `remix watch`, so this will mean remix.config has access to .env when using remix app server but not when using another server, which makes remix.config behavior not reliable.

I think you should require dotenv yourself in remix.config if you know you want access to env variables.

herecydev:
@sergiodxa how is one supposed to address that inconsistency anyway? Sometimes you have env variables (dev) and sometimes they're absent (watch). I would have assumed it was never intended to run `watch` with a "local environment" in mind.

If I were to agree with you then I think [this part](https://remix.run/docs/en/v1/guides/envvars#server-environment-variables) of the documentation could do with some updating:

> If your host doesn't have any conventions for environment variables during development, the remix dev server can help out as it provides built-in support for dotenv.

sergiodxa:
> remix dev server can help out as it provides built-in support for dotenv

It mentions Remix Dev Server, when you use `remix dev` you are using that.

But if you use Express for example, you don't use RDS anymore, you use `remix watch` to build the app on every change and `node server.js` to run your server. In that case, you need to setup `dotenv` in your Express server, and if this PR is merged and `remix dev` gives remix.config.js access to the `.env` variables, that will only happen on RDS and not Express or another adapter.

herecydev:
Right I follow that, but if I specifically highlight:

> If your host doesn't have any conventions for environment variables during development

What hosts do provide a convention for environment variables? I'm sorry if I'm misunderstanding this, but from my knowledge any host would provide env variables as soon as you run. For example if you run with netlify would that any environment variables be surfaced to remix?

sergiodxa:
> Right I follow that, but if I specifically highlight:
> 
> > If your host doesn't have any conventions for environment variables during development
> 
> What hosts do provide a convention for environment variables? I'm sorry if I'm misunderstanding this, but from my knowledge any host would provide env variables as soon as you run. For example if you run with netlify would that any environment variables be surfaced to remix?

Vercel for example, reads the `.env` file if you run the `vercel dev`, in that case your code doesn't need to require dotenv. In production, you can configure your env variables on a dashboard and make them available for your app.

Netlify also has support for env variables AFAIK. And Cloudflare makes env variables available through the `context.env` key on loaders and actions.

And technically, any hosting provider has support since env variables are a system thing, you don't need dotenv to use them, if you do `SOMETHING=value remix dev` your app will have access to `process.env.SOMETHING` in the server code, or if your add `export SOMETHING=value` in your `.zshrc` or `.bashrc` file, and re-open the terminal it will also be available.

So, what the docs mean is that some postings have some ways to populate env variables for your from a config or a file, you should use that, but with Remix Dev Server you can use `.env`, with Express you can setup it yourself because you own the code.

herecydev:
Thanks for the thorough breakdown. With those adapters like vercel and netlify that make those env variables available, they should be able to be accessed within the remix config file then? I think the point I was trying to get across in the pr any my comments is that the env variables should be loaded as early as possible. I would imagine all adapters would surface those variables immediately when running any remix code so I'm just shooting for a consistent baseline in when they're available.

sergiodxa:
> Thanks for the thorough breakdown. With those adapters like vercel and netlify that make those env variables available, they should be able to be accessed within the remix config file then? I think the point I was trying to get across in the pr any my comments is that the env variables should be loaded as early as possible. I would imagine all adapters would surface those variables immediately when running any remix code so I'm just shooting for a consistent baseline in when they're available.

AFAIK and understand, yes, you should be able to access them in remix.config.js, but you will have to try, specially in local, inside their servers it's most likely to work.

chaance:
Hi @herecydev, thanks for the suggestion and sorry it took us a while to come back to this one. I think I'm a bit hesitant here for many of the reasons @sergiodxa mentioned. It gets tricky to set the right expectations for production since the prod build will also rely on the Remix config, and it will get a bit too easy for a mismatch between the config and runtime values.

I'm going to close this for now and make a note to update our docs to be more clear on this. Thanks for bringing it up!

PR #3358, opened 2022-06-01
fix(examples): use `post` action in `firebase-auth-firestore` example

penx:
Fix the Firebase example when JavaScript is disabled.

`method="delete"` is not a valid form action.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method

While this works when JavaScript is enabled, it does not fall back gracefully.

Instead, use specify form action using name + value on buttons, as shown in FAQs:

https://remix.run/docs/en/v1/pages/faq#how-do-i-handle-multiple-forms-in-one-route

Closes: a bug I encountered while working on the Firebase example - no open issue.

- [ ] Docs N/A
- [ ] Tests

Testing Strategy:

- is it possible to write integration tests for the examples?

In order to test this manually, I ran the Firebase example with JavaScript enabled and ensured:

- when submitting new TODO: entries with a blank input, the validation error shows
- when deleting TODO: entries, the entries are removed

### DIFF
--- a/examples/firebase-auth-firestore/app/routes/index.tsx
+++ b/examples/firebase-auth-firestore/app/routes/index.tsx
@@ -34,18 +34,21 @@ export type ActionData = {
 export const action: ActionFunction = async ({ request }) => {
   const { uid } = await requireAuth(request);
   const form = await request.formData();
-  if (request.method === "POST") {
+  const action = form.get("action");
+  if (action === "create") {
     const title = form.get("title");
-    if (typeof title !== "string")
+    if (typeof title !== "string" || !title.length) {
       return json<ActionData>({ error: "title is required" }, { status: 400 });
+    }
 
     await addTodo(uid, title);
     return redirect("/");
   }
-  if (request.method === "DELETE") {
+  if (action === "delete") {
     const id = form.get("id");
-    if (typeof id !== "string")
+    if (typeof id !== "string") {
       return json<ActionData>({ error: "id is required" }, { status: 400 });
+    }
     await removeTodo(uid, id);
     return redirect("/");
   }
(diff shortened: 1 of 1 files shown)


examples/firebase-auth-firestore/app/routes/index.tsx
```diff
@@ -34,32 +34,44 @@ export type ActionData = {
...
+    case "delete": {
+      const id = form.get("id");
+      if (typeof id !== "string") {
+        return json<ActionData>({ error: "id is required" }, { status: 400 });
+      }
+      await removeTodo(uid, id);
+      return redirect("/");
+    }
+    default: {
+      return json<ActionData>({ error: "unknown method" }, { status: 400 });
```
MichaelDeBoey, line 60: Since we only have `create` or `delete`, we can keep the 2 `if`-statements like they were
Only need to change them to check for the `action`.

MichaelDeBoey (changes requested):

examples/firebase-auth-firestore/app/routes/index.tsx
```diff
@@ -34,18 +34,21 @@ export type ActionData = {
 export const action: ActionFunction = async ({ request }) => {
   const { uid } = await requireAuth(request);
   const form = await request.formData();
-  if (request.method === "POST") {
+  const action = form.get("action");
+  if (action === "create") {
     const title = form.get("title");
-    if (typeof title !== "string")
+    if (typeof title !== "string" || !title.length) {
```
MichaelDeBoey, line 40:
```suggestion
    if (typeof title !== "string" || title.length === 0) {
```

MichaelDeBoey (changes requested):

examples/firebase-auth-firestore/app/routes/index.tsx
penx: Thanks, is this for performance, personal preference or consistency with the rest of the code base?

MichaelDeBoey: It makes it more clear what you're actually checking.

MichaelDeBoey (approved):

kentcdodds (approved):
Thanks!

PR #3359, opened 2022-06-01
ci: update deployment testing infra

mcansh:
let's clean up after ourselves, some providers limit how many resources you can have,
AWS has a 100 S3 Bucket limit and CF Pages limits you to 100 pages projects – others may have limits as well, so lets just remove the app when we're done

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/scripts/deployment-test/_shared.mjs
+++ b/scripts/deployment-test/_shared.mjs
@@ -105,8 +105,3 @@ export async function validatePackageVersions(directory) {
     })
   );
 }
-
-// check if the file was called directly
-export function isMainModule(module) {
-  return module.url.endsWith(process.argv[1]);
-}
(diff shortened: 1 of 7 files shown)


MichaelDeBoey (approved):

kentcdodds:
Does the client need to be reused? Maybe we can break things out into other files if necessary?

mcansh:
> 

it was just a one time thing to clear out older deployments, we have about 6 days worth since i opened and ran the deletion script (that I've since removed 😅), but i can manually remove those

kentcdodds:
Got it 👍 Sounds good then. I think we can merge this. Thanks!

kentcdodds (approved):
Solid 👍 Thank you!

PR #3360, opened 2022-06-01
chore: add `engines` field to `package.json`

MichaelDeBoey:


kentcdodds (approved):
Thanks!

Dmitry-Sokolov:
@MichaelDeBoey Could you please help, why did you put 14 and not 12? Do you have some special requirements for the node js? I try to run on the 12 version and it works, so I wonder, why 14?

machour:
@Dmitry-Sokolov Officially, Remix supports Node version ^14.17.0, or >=16.0.0

Dmitry-Sokolov:
Thank you @machour 
Need to say, that 6.6.1 can work with the last 12 node js, just need to make the correct configuration for dev dependencies and some other small fixes, in any case, not all old projects can be super fast upgrade nodes :(

PR #3362, opened 2022-06-01
feat(examples/firebase-auth-firestore): use rest api for server auth

penx:
Use the [Firebase Auth REST API](https://firebase.google.com/docs/reference/rest/auth) for authenticating server side, instead of the [Firebase client SDK](https://github.com/firebase/firebase-js-sdk).

The [Firebase Client Auth SDK is stateful](https://firebase.google.com/docs/auth/web/auth-state-persistence). It seems to have been built for the purpose of using it within an application where only one user is logged in at a time (i.e. client applications). It _may_ not pose a security risk if it is _only_ used for authentication calls, but login state is persisted in the module and this could lead to security issues if other features such as `firebase/firestore` were used, as these would behave as the last user to log in.

The Firebase Auth REST API gives us the same functionality with no external dependencies and control over state.

This is the API that the client SDK is calling, so it is unnecessary to include the client sdk to call a single endpoint. 

As far as I'm aware, there are rate limiting issues with both approaches, which I plan to address in a follow up PR by calling the REST API client side and falling back to server side. 

This was previously discussed at https://github.com/remix-run/remix/pull/1811#discussion_r884011463

Closes: discussion at https://github.com/remix-run/remix/pull/1811#discussion_r884011463

- [x] Docs
- [ ] Tests

Testing Strategy:

- ensuring examples/firebase-auth-firestore works with and without JavaScript against emulator and production instances


kentcdodds:
> which I plan to address in a follow up PR by calling the REST API client side and falling back to server side.

*really* don't want to do that. Is it possible to avoid this at all? Sounds like the rate-limiting doesn't impact this PR anyway, so I can probably merge as-is. I'm just really disappointed that Firebase seems to require that we do auth on the client. Is there really no way to avoid that?

kentcdodds (approved):
Definitely an improvement over what we've got currently. Thank you!

penx:
> really don't want to do that. Is it possible to avoid this at all? Sounds like the rate-limiting doesn't impact this PR anyway, so I can probably merge as-is. I'm just really disappointed that Firebase seems to require that we do auth on the client. Is there really no way to avoid that?

@kentcdodds pretty sure I’d need someone from Firebase to give a definitive answer. I could try support. 

Would be good to know your objections as this would fall back to server side auth and would have a very small client side bundle impact. The only thing I can think of is we have to event.preventDefault() before the async function that does auth is called.

PR #3367, opened 2022-06-02
docs(api/remix): fix `NavLink` example

poteirard:
I think the types forces you to return an empty object instead of `undefined`

Cursor_and_secrets-vault_–_NavList_tsx

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -282,6 +282,7 @@
 - penx
 - phishy
 - plastic041
+- poteirard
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
I think this is a typing-issue in `react-router-dom` tbh.
`style` should accept returning `undefined` if it's a function, as it also accepts `undefined` as value.

poteirard:
Ok, close then Thank you

Issue #1285, opened 2021-12-30
[Bug]: `process` is not defined when build with cloudflare pages


willin:
### Which Remix packages are impacted?

- [ ] `remix` (Remix core)
- [ ] `create-remix`
- [ ] `@remix-run/architect`
- [X] `@remix-run/cloudflare-workers`
- [ ] `@remix-run/dev`
- [ ] `@remix-run/express`
- [ ] `@remix-run/netlify`
- [ ] `@remix-run/node`
- [ ] `@remix-run/react`
- [ ] `@remix-run/serve`
- [ ] `@remix-run/server-runtime`
- [ ] `@remix-run/vercel`

### What version of Remix are you using?

"@remix-run/cloudflare-pages": "^1.1.1"

### What version of Node are you using? Minimum supported version is 14.

17.2

### Steps to Reproduce

https://github.com/js-cool/remix-blog-demo/blob/main/app/utils/enviroment.ts#L17

```ts
/**
 * Get the value from an environment variable and throw a
 * MissingEnvironmentError exception if it is not found, optinally you can pass
 * a fallback value to avoid throwing if the environment variable is not defined
 * @param {string} env The environment variable name.
 * @param {string} [fallback] The fallback value in case it's not defined.
 * @throws {MissingEnvironmentError} If the environment variable is not found.
 * @returns {string} The value of the environment variable.
 */
export function env(name: string, fallback?: string): string {
  const value = process.env[name] ?? fallback;
  if (!value) throw new MissingEnvironmentError(name);
  return value;
}
```

### Expected Behavior

work properly

### Actual Behavior

cannot build

```
💿 Built in 167ms

warn - The `purge`/`content` options have changed in Tailwind CSS v3.0.
warn - Update your configuration file to eliminate this warning.
Compiling worker to "/var/folders/td/zr87rwn546n4q0y3zddn95zw0000gn/T/functionsWorker.js"
Finished in 55ms.
wrangler pages dev [directory] [-- command]

🧑‍💻 Develop your full-stack Pages application locally

Positionals:
  directory  The directory of static assets to serve  [string]
  command    The proxy command to run  [string]

Flags:
  -c, --config   Path to .toml configuration file  [string]
  -h, --help     Show help  [boolean]
  -v, --version  Show version number  [boolean]

Options:
  -l, --local        Run on my machine  [boolean] [default: true]
      --port         The port to listen on (serve from)  [number] [default: 8788]
[...]

PR #3368, opened 2022-06-02
docs(guides/envvars): clarify docs for local/production env variables

dgurns:
When I recently spun up a project using the `@remix-run/cloudflare-pages` adapter, it was not clear how to set up and use env variables, both locally and when deployed to Cloudflare Pages.

Thankfully @jemjam figured it out in [this issue comment](https://github.com/remix-run/remix/issues/1285#issuecomment-1124463689)

This updates the Environment Variables docs with that information. This should hopefully save people a lot of pain when initially setting up a new project with the Cloudflare Pages adapter.

### DIFF
--- a/docs/guides/envvars.md
+++ b/docs/guides/envvars.md
@@ -20,21 +20,17 @@ If your experience with web development is primarily with the JS frameworks in t
 
 ## Server Environment Variables
 
-Environment variables on your server will be handled by your host, for example:
+### Local Development
 
-- [Netlify](https://docs.netlify.com/configure-builds/environment-variables/)
-- [Fly.io](https://fly.io/docs/reference/secrets/)
-- [Cloudflare Workers](https://developers.cloudflare.com/workers/platform/environment-variables)
-- [Vercel](https://vercel.com/docs/environment-variables)
-- [Architect](https://arc.codes/docs/en/reference/cli/env)
+If you're using the `remix dev` server to run your project locally, it has built-in support for [dotenv](https://www.npmjs.com/package/dotenv).
 
-If your host doesn't have any conventions for environment variables during development, the `remix dev` server can help out as it provides built-in support for [dotenv](https://www.npmjs.com/package/dotenv).
-
-If you're using the `remix dev` server, you can do this very quickly:
+First, create an `.env` file in the root of your project:
 
 ```sh
 touch .env
 ```
+
+<docs-error>Do not commit your <code>.env</code> file to git, the point is that it contains secrets!</docs-error>
 
 Edit your `.env` file.
 
@@ -50,13 +46,9 @@ export async function loader() {
 }
 ```
 
-Note that `dotenv` is only for development, you should not use it in production, so Remix doesn't load these when running `remix serve`. You'll need to follow your host's guides on adding secrets to your production server.
+If you're using the `@remix-run/cloudflare-pages` adapter, env variables work a little differently. You'll need to manually pass your `.env` file to `wrangler` by doing this:
 
-<docs-error>Do not commit your <code>.env</code> file to git, the point is that it contains secrets!</docs-error>
-
-If you're using the `@remix-run/cloudflare-pages` adapter, env variables work a little differently. When running locally, you'll need to manually pass your `.env` file to `wrangler` by doing this:
-
-```sh
+```json
 // package.json
 
 {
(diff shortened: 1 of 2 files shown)


machour:
@dgurns Thank you for opening this PR !

This seems a little too specific to me, and perhaps should be added in a new, CF specific, section. @kentcdodds WDYT?

docs/guides/envvars.md
```diff
@@ -54,6 +54,27 @@ Note that `dotenv` is only for development, you should not use it in production,
...
+	}
+}
+```
+
+When deploying to Cloudflare, you can use the Cloudflare Pages Project Settings UI to set env variables for Preview and Production.
+
+Then, in your `loader` functions, both locally and when deployed on Cloudflare, you can access environment variables directly on `context`:
+```
+export const loader = async ({ context }) => {
+  console.log('My env:', context.ENV_VAR);
```
jemjam, line 74: Nit: Maybe it's just me, but I never like to leave a console log in the docs when the example could focus on more appropriate values. (I know my example in the thread included it, but docs examples tend to be taken more literally, folks will leave context logging in places it shouldn't be.) 

Maybe we can just focus on the env vars being passed verbatim by default, with an example like grabbing MY_SECRET_KEY out of `context.MY_SECRET_KEY`?

docs/guides/envvars.md
```diff
@@ -54,6 +54,27 @@ Note that `dotenv` is only for development, you should not use it in production,
 
 <docs-error>Do not commit your <code>.env</code> file to git, the point is that it contains secrets!</docs-error>
 
+If you're using the `@remix-run/cloudflare-pages` adapter, env variables work a little differently. When running locally, you'll need to manually pass your `.env` file to `wrangler` by doing this:
```
jemjam, line 57: I wonder if this would be better in the readme for `@remix-run/cloudflare-pages` rather than in the general docs for envvars?

docs/guides/envvars.md
```diff
@@ -54,6 +54,27 @@ Note that `dotenv` is only for development, you should not use it in production,
...
+
+```sh
+// package.json
+
+{
+	"scripts": {
+		"dev:wrangler": "cross-env NODE_ENV=development wrangler pages dev ./public --binding $(cat .env)"
```
jemjam, line 64: aha, I didn't even think to cat the env file, I've been binding these values individually (like a fool)

jemjam:
Just threw some minor thoughts up, take or leave my opinions if you choose to use any of this. ❤️

dgurns:
Thanks for the great feedback folks. I updated the doc to be better organized, dividing "Server Environment Variables" into "Local Development" and "Production" sections.

docs/guides/envvars.md
dgurns: A Readme per adapter does seem like a good idea but they just seem to be boilerplate at the moment, so I left them as is.

docs/guides/envvars.md
dgurns: Makes sense to me, however @ryanflorence's example above used the `console.log` so I figured I'd keep it consistent.

docs/guides/envvars.md
dgurns: I gotta give a h/t to @theeomm on that one! https://github.com/remix-run/remix/pull/3186#issue-1235573117

dgurns:
> @theeomm already updated the `cloudflare-pages` template in #3186, so I don't think these doc changes are necessary anymore once his PR is merged? 🤔

That PR looks great and will help, but I think this one is important as well. If you're a new user with a Cloudflare Pages project:
- It's not clear that you can use a `.env` file - usually Cloudflare projects use a `wrangler.toml` locally for env variables, if I understand correctly.
- Once you have the env variables loaded in, it's not clear how to access them. I initially reached for `process.env` by default, and then top-level global variables like Cloudflare Workers, before realizing by trial and error that they were on `context`.

theeomm:
> > @theeomm already updated the `cloudflare-pages` template in #3186, so I don't think these doc changes are necessary anymore once his PR is merged? 🤔
> 
> That PR looks great and will help, but I think this one is important as well. If you're a new user with a Cloudflare Pages project:
> 
> * It's not clear that you can use a `.env` file - usually Cloudflare projects use a `wrangler.toml` locally for env variables, if I understand correctly.
> * Once you have the env variables loaded in, it's not clear how to access them. I initially reached for `process.env` by default, and then top-level global variables like Cloudflare Workers, before realizing by trial and error that they were on `context`.

I think it is necessary for this PR to be merged too as it really documents how to handle environment variables with CF, which I should have probably done in my PR in the first place.

docs/guides/envvars.md
```diff
@@ -50,9 +46,37 @@ export async function loader() {
...
+```json
+// package.json
+
+{
+	"scripts": {
+		"dev:wrangler": "cross-env NODE_ENV=development wrangler pages dev ./public --binding $(cat .env)"
+	}
+}
+```
+
```
kentcdodds, line 60: We should probably not assume people will know about `cross-env`.

```suggestion

<docs-info>Learn more about <a href="https://npm.im/cross-env"><code>cross-env</code></a></docs-info>

```

kentcdodds (approved):

kentcdodds:
Thanks everyone!

dgurns:
Cool! Gotta say I am super impressed with the Remix community so far. Thank you all for jumping in with feedback so quickly on the PR, and seeing it merged in a few days. Amazing stuff! Excited to keep digging in to Remix.

Issue #3372, opened 2022-06-02
Error. Redirects to relative path in the actions are not working, the are made absolute


luispuig:
### What version of Remix are you using?

1.5.1

### Steps to Reproduce

* Create a nested route with a button that calls the action. For instance `/admin/login`
* In the action do a redirect with a relative path, for instance `export const action = async () => redirect("relative");`
* The redirection is made to "/relative" instead of to `/admin/relative` or `/admin/login/relative`

```
import type { ActionFunction } from "@remix-run/node";
import { redirect } from "@remix-run/node";
import { Form, Link } from "@remix-run/react";

export const action: ActionFunction = async ({ request }) => {
  return redirect("relative");
};

export default function ExamplePage() {
  return (
    <>
      <Link to="relative">Go to relative with Link</Link>
      <Form method="post">
        <button type="submit">Go to relative with Action</button>
      </Form>
    </>
  );
}
```

### Expected Behavior

The redirections in an action to a relative path are made relative and not absolute.

![Jun-03-2022 00-39-41](https://user-images.githubusercontent.com/24297787/171749864-5b7f9ef6-1aa6-47de-b765-8f41dd36189f.gif)

### Actual Behavior

The redirections in an action to a relative path are made absolute.

![Jun-03-2022 00-40-05](https://user-images.githubusercontent.com/24297787/171749899-f443f5de-1881-484b-86b7-f7ea593215e8.gif)

luispuig:
A typical use case is the `/notes/new` action, redirecting to the note page after creating the new one. `redirect(note.id)`. Using the relative way, the notes "module/folder" doesn't need to know its position in the whole tree.

luispuig:
I found this related issue: https://github.com/remix-run/remix/issues/91

machour:
Would you mind opening a PR with a failing test case? https://github.com/remix-run/remix/blob/dev/integration/bug-report-test.ts

luispuig:
Created

machour:
CI ran and it's failing, perfect 👌🏼 

Closing this issue in favor of the bug-PR

PR #3373, opened 2022-06-03
BUG: Relative redirect action test

luispuig:
ISSUE: https://github.com/remix-run/remix/issues/3372

1.5.1

### Steps to Reproduce

* Create a nested route with a button that calls the action. For instance `/admin/login`
* In the action do a redirect with a relative path, for instance `export const action = async () => redirect("relative");`
* The redirection is made to "/relative" instead of to `/admin/relative` or `/admin/login/relative`

```
import type { ActionFunction } from "@remix-run/node";
import { Form, Link } from "@remix-run/react";

  return redirect("relative");
};

export default function ExamplePage() {
  return (
    <>
      <Link to="relative">Go to relative with Link</Link>
      <Form method="post">
        <button type="submit">Go to relative with Action</button>
      </Form>
    </>
  );
}
```

### Expected Behavior

The redirections in an action to a relative path are made relative and not absolute.

Jun-03-2022 00-39-41

### Actual Behavior

The redirections in an action to a relative path are made absolute.

Jun-03-2022 00-40-05


mcansh:
under the hood this is a standard HTTP redirect using the [Location](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location) Header, but i do see the confusion / desire for the expected behavior - however, that would more than likely be a "breaking" change as we don't currently need the current url in our redirect helper function 


for now a fix would be the following (with some checks around whether or not your redirect is external)

```ts
import { resolvePath } from "react-router-dom";

export const loader: LoaderFunction = ({ request }) => {
  let resolved = resolvePath("relative", request.url);

  return redirect(resolved.pathname + resolved.search);
};
```

luispuig:
Thanks for the answer, that could work for my use-case.

Related to this, I opened this idea https://github.com/remix-run/remix/discussions/3722 to evaluate different ways to get the absolute path of a route.

jgconway:
Is there a more general issue here with relative redirects handled by the client-side router?

With the following setup:
- `/app/routes/parent/has-link` -> renders `<Link to="/parent/has-redirect">`
- `/app/routes/parent/has-redirect` -> loader returns `redirect("destination")`

Navigating to `/parent/has-link` and then clicking the link results in a redirect to `/destination`. However, opening the link in a new tab or navigating directly to `/parent/has-redirect` results in a redirect to `/parent/destination`.

The way Remix handles the redirect isn't consistent with the way the browser does.

I guess this is because the router resolves the redirect relative to `window.location.origin` rather than the full URL. https://github.com/remix-run/remix/blob/32337757eba981e5d9705e40ad084d9d5c2d2bf2/packages/remix-react/routes.tsx#L216-L219

I'm wondering what the reason is for this, or if it's actually a bug?

aaronadamsCA:
It looks like this was never merged/fixed, is that correct? I think I've just encountered it myself.

machour:
@aaronadamsCA This is a pull request reproducing the problem. It's still open, so the problem probably still exist.

Although I know @brophdawg11 have been fixing various redirect issues, so I'm pinging him for feedback 🙏🏼

brophdawg11:
This was fixed with the integration of React Router 6.4+ in 1.10.0.  The test here is a false negative because it's missing an `<Outlet/>`, and thus never rendering the expected child contents after the redirect.  Here's a working example of this test with the `Outlet`: https://stackblitz.com/edit/remix-run-remix-g9e4gs.

I'm going to close this out since we have plenty of tests around relative redirects over in React Router.  

@aaronadamsCA If you are having related issues would you mind opening a new issue with a reproduction from https://remix.new?

aaronadamsCA:
@brophdawg11 Thanks for the explanation. I'll see if I can find the time to file a reproduction.

In the meantime, for the benefit of anyone searching, I _think_ this is what I observed:

- Route `/route/` has `<form action="action">`
- Action `/route/action/` has `redirect("..")`
  - Expected: User is redirected to `/route/`
  - Actual: User is redirected to `/`

I think this happens because `redirect()` resolves relative to the caller instead of the action. I worked around it by switching to absolute paths in redirects.

Issue #3105, opened 2022-05-05
use of source-map-support causes the wrong line numbers to appear in @swc/jest tests


parisholley:
### What version of Remix are you using?

1.4.0

### Steps to Reproduce

https://github.com/kulshekhar/ts-jest/issues/2372

if for any reason `@remix-run/node` is pulled into a test, this issue will happen. should either ENV flag outside of test, or version flag to only include on older node versions?

### Expected Behavior

should show correct typescript line numbers

### Actual Behavior

line numbers get mangled

esamattis:
I'm seeing this with babel-jest when ever `@remix-run/node` gets imported in the test code. I guess both have some sort of source map support which causes conflicts.

esamattis:
I think I figured it out. The issue is here

https://github.com/remix-run/remix/blob/b0b544d5372abece57055a39f257ff79e987def9/packages/remix-node/index.ts#L3

Jest probably installs this as well and they seem to collide. The line numbers are correct if this line is removed. I created a repro here with babel-jest:

https://github.com/esamattis/remix-jest-line-numbers-bug

I  also created a [patch-package](https://www.npmjs.com/package/patch-package) patch [here](https://github.com/esamattis/remix-jest-line-numbers-bug/blob/master/patches/%40remix-run%2Bnode%2B1.5.1.patch) which can be used to fix this in projects.

But this is not probably not the best way to fix this since it means coding jest specific things into Remix. Maybe we can add a global flag to disable Remix source maps in the test frameworks 🤔

esamattis:
Just to note: this issue happens with ts-jest too and the same workaround works for it too.

yesmar12:
Since there has not been any progress I'm commenting to say that this worked for me with ts-jest as you say. Would love to see the PR above be put into the application.

gorbak25:
If anyone encounters this issue you may use `moduleNameMapper` in jest to ensure the `source-map-support` module is is replaced in tests by a harmless stub:
```js
// file: ./app/jest/source-map-support.stub.ts
/* eslint-disable no-console */
export const install = () => {
  let caller = "[UNKNOWN]";
  // Play some 4D chess
  const o = Error.prepareStackTrace;
  try {
    Error.prepareStackTrace = (_err, stackTraces) => {
      const c = stackTraces[1];
      if (c) caller = `${c.getFileName()}:${c.getLineNumber()}:${c.getColumnNumber()}`;
    };
    const a = new Error();
    void a.stack;
[...]

PR #3374, opened 2022-06-03
fix(remix-node): add `REMIX_DISABLE_SOURCE_MAPS` environment variable support

esamattis:
Closes: #3105

The `source-map-support` module can collide with other tools using it or tools like it causing line numbers to be off. One example of these tools is Jest. See the #3105 issue.

This PR allows users of such tools to disable source maps from Remix.

Here's a repro with Jest with patch demonstration:
https://github.com/esamattis/remix-jest-line-numbers-bug

Testing Strategy:

Manual. I have no idea how one could write a unit tests for this.

### DIFF
--- a/packages/remix-node/index.ts
+++ b/packages/remix-node/index.ts
@@ -1,6 +1,8 @@
 import sourceMapSupport from "source-map-support";
 
-sourceMapSupport.install();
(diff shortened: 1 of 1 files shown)


brophdawg11:
Closing per https://github.com/remix-run/remix/issues/3105#issuecomment-1662901267

PR #3375, opened 2022-06-03
ci: update needs-response workflow to use official stale action

mcansh:

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -35,4 +35,3 @@ jobs:
           days-before-stale: -1
           any-of-labels: needs-response
           labels-to-remove-when-unstale: needs-response
-          repo-token: ${{ github.token }}
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -18,7 +18,7 @@ jobs:
 
     steps:
       - name: 🥺 Handle Ghosting
-        uses: actions/stale@v5.0.0
+        uses: actions/stale@v5
         with:
           close-issue-message: >
             This issue has been automatically closed because we haven't received a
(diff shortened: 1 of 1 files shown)


.github/workflows/no-response.yml
```diff
@@ -7,20 +7,32 @@ on:
...
+
 jobs:
   noResponse:
     if: github.repository == 'remix-run/remix'
     runs-on: ubuntu-latest
 
     steps:
       - name: 🥺 Handle Ghosting
-        uses: lee-dohm/no-response@v0.5.0
+        uses: actions/stale@v5.0.0
```
MichaelDeBoey, line 21: You can just use latest version by doing

```suggestion
        uses: actions/stale@v5
```

.github/workflows/no-response.yml
```diff
@@ -7,20 +7,32 @@ on:
...
+            This PR has been automatically closed because we haven't received a
+            response from the original author 🙈. This automation helps keep the issue
+            tracker clean from PRs that are unactionable. Please reach out if you
+            have more information for us! 🙂
+          days-before-close: 10
+          # don't automatically mark issues/PRs as stale
+          days-before-stale: -1
+          any-of-labels: needs-response
+          labels-to-remove-when-unstale: needs-response
+          repo-token: ${{ github.token }}
```
MichaelDeBoey, line 38: This is the default value, so no need to pass it

```suggestion
```

MichaelDeBoey (changes requested):

.github/workflows/no-response.yml
mcansh: 👍 copy pasta 🤪

kentcdodds (approved):
Looks good to merge. Why is this still a draft?

mcansh:
> Looks good to merge. Why is this still a draft?

oops!

MichaelDeBoey:
@mcansh We probably want to have this in `main` as well?

kentcdodds:
I think we can wait on this one. No rush. Thanks for the reminder though @MichaelDeBoey!

PR #3376, opened 2022-06-03
Move rollup configs to individual packages

brophdawg11:
This PR splits the root `rollup.config.js` file out to individual `packages/{name}/rollup.config.js` files, and adjusts the root `rollup.config.js` file to read all `packages/{name}/rollup.config.js` files to auto-generat the root config.  This is similar in nature to the `jest` approach this repo is using.

This has two primary benefits:
* In preparation for the repo merge, I'll be making the same change in `react-router` (once we merge the `remixing` branch) and aligning `react-router` to use the same exact approach and `rollup.utils.js` file.  This means that we do not have to merge the root `rollup.config.js` file from `remix` into `react-router` at all.  When we migrate the `remix` `packages/*` folders, they will bring along the package-specific `rollup.config.js` files, and the `react-router` root `rollup.config.js` file will just automatically pick up on them 😄 
* We can now do individual package builds via `npx rollup -c` directly from within a `packages/{name}` directory if we don't need to build everything.  This also "just works" with things like `LOCAL_BUILD_DIRECTORY` and `rollup -w` - both from the root and from individual packages

To test, I ran fresh builds on both `dev` and this branch and ensured `diff -rq build-dev/ build-branch/` didn't report any differences.


package.json
```diff
@@ -91,8 +91,8 @@
     "prompt-confirm": "^2.0.4",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "rollup": "^2.36.1",
-    "rollup-plugin-copy": "^3.3.0",
+    "rollup": "^2.75.5",
+    "rollup-plugin-copy": "^3.4.0",
```
brophdawg11, line 95: Updated to the latest rollup dependencies while I was in here

packages/create-remix/rollup.config.js
```diff
@@ -0,0 +1,52 @@
...
+  buildDir,
+  copyToPlaygrounds,
+  createBanner,
+  executableBanner,
+  getVersion,
+} = require("../../rollup.utils");
+
+/** @returns {import("rollup").RollupOptions[]} */
+module.exports = function rollup() {
+  let sourceDir = path.relative(process.cwd(), __dirname) || ".";
```
brophdawg11, line 17: This changes from what used to be a static `packages/{name}`.  We now need `sourceDir` to be relative to the current working directory, so that configs work fine regardless of whether they're run from the root or the individual `packages/{name}` folder

packages/create-remix/rollup.config.js
```diff
@@ -0,0 +1,52 @@
...
+  let outputDir = path.join(buildDir, "node_modules/create-remix");
```
brophdawg11, line 18: Using `buildDir` from `rollup.utils` ensure's we respect `LOCAL_BUILD_DIRECTORY`

rollup.utils.js
```diff
@@ -0,0 +1,201 @@
+const fs = require("fs");
```
brophdawg11, line 1: This file holds utilities used by all package-specific `rollup.config.js` files, since we can't export them directly from the root `rollup.config.js` because `rollup` expects a single function export when using a CJS module :/.  We may be able to move to a `rollup.config.mjs` file and then move the utils into there but I figured keeping the surface area of this change smallest pre-merge makes sense, and we can look into the `.mjs` file after the merge if we want to get rid of the utils file.  Happy to look into it ahead of time if folks don't have concerns though.

mcansh (approved):
LGTM :) but might be worth having Michael give it a look

packages/remix-cloudflare-pages/rollup.config.js
```diff
@@ -0,0 +1,49 @@
...
+  isBareModuleId,
+} = require("../../rollup.utils");
+
+/** @returns {import("rollup").RollupOptions[]} */
+module.exports = function rollup() {
+  let sourceDir = path.relative(process.cwd(), __dirname) || ".";
+  let outputDir = path.join(
+    buildDir,
+    "node_modules/@remix-run/cloudflare-pages"
+  );
```
MichaelDeBoey, line 21: Shouldn't we make sure Windows won't fail here?

```suggestion
  let outputDir = path.join(
    buildDir,
    "node_modules",
    "@remix-run",
    "cloudflare-pages"
  );
```

packages/remix-cloudflare-workers/rollup.config.js
```diff
@@ -0,0 +1,49 @@
...
+  isBareModuleId,
+} = require("../../rollup.utils");
+
+/** @returns {import("rollup").RollupOptions[]} */
+module.exports = function rollup() {
+  let sourceDir = path.relative(process.cwd(), __dirname) || ".";
+  let outputDir = path.join(
+    buildDir,
+    "node_modules/@remix-run/cloudflare-workers"
+  );
```
MichaelDeBoey, line 21:
```suggestion
  let outputDir = path.join(
    buildDir,
    "node_modules",
    "@remix-run",
    "cloudflare-workers"
  );
```

packages/remix-cloudflare/rollup.config.js
```diff
@@ -0,0 +1,93 @@
...
+  copyToPlaygrounds,
+  createBanner,
+  getVersion,
+  isBareModuleId,
+} = require("../../rollup.utils");
+
+/** @returns {import("rollup").RollupOptions[]} */
+module.exports = function rollup() {
+  let sourceDir = path.relative(process.cwd(), __dirname) || ".";
+  let outputDir = path.join(buildDir, "node_modules/@remix-run/cloudflare");
```
MichaelDeBoey, line 18:
```suggestion
  let outputDir = path.join(buildDir, "node_modules", "@remix-run", "cloudflare");
```

packages/remix-deno/rollup.config.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const path = require("path");
+const copy = require("rollup-plugin-copy");
+
+const { buildDir, copyToPlaygrounds } = require("../../rollup.utils");
+
+/** @returns {import("rollup").RollupOptions[]} */
+module.exports = function rollup() {
+  let sourceDir = path.relative(process.cwd(), __dirname) || ".";
+  let outputDir = path.join(buildDir, "node_modules/@remix-run/deno");
```
MichaelDeBoey, line 10:
```suggestion
  let outputDir = path.join(buildDir, "node_modules", "@remix-run", "deno");
```

packages/remix-react/rollup.config.js
```diff
@@ -0,0 +1,130 @@
...
+  copyToPlaygrounds,
+  createBanner,
+  getVersion,
+  isBareModuleId,
+} = require("../../rollup.utils");
+
+/** @returns {import("rollup").RollupOptions[]} */
+module.exports = function rollup() {
+  let sourceDir = path.relative(process.cwd(), __dirname) || ".";
+  let outputDir = path.join(buildDir, "node_modules/@remix-run/react");
```
MichaelDeBoey, line 18:
```suggestion
  let outputDir = path.join(buildDir, "node_modules", "@remix-run", "react");
```

packages/remix-serve/rollup.config.js
```diff
@@ -0,0 +1,76 @@
...
+  createBanner,
+  executableBanner,
+  getVersion,
+  isBareModuleId,
+} = require("../../rollup.utils");
+
+/** @returns {import("rollup").RollupOptions[]} */
+module.exports = function rollup() {
+  let sourceDir = path.relative(process.cwd(), __dirname) || ".";
+  let outputDir = path.join(buildDir, "node_modules/@remix-run/serve");
```
MichaelDeBoey, line 19:
```suggestion
  let outputDir = path.join(buildDir, "node_modules", "@remix-run", "serve");
```

packages/remix/rollup.config.js
```diff
@@ -0,0 +1,74 @@
...
+  buildDir,
+  copyToPlaygrounds,
+  createBanner,
+  getVersion,
+} = require("../../rollup.utils");
+
+/** @returns {import("rollup").RollupOptions[]} */
+module.exports = function rollup() {
+  let sourceDir = path.relative(process.cwd(), __dirname) || ".";
+  let outputDir = path.join(buildDir, "node_modules/remix");
```
MichaelDeBoey, line 16:
```suggestion
  let outputDir = path.join(buildDir, "node_modules", "remix");
```

rollup.utils.js
```diff
@@ -0,0 +1,201 @@
...
+
+/** @returns {import("rollup").RollupOptions[]} */
+function getAdapterConfig(adapterName) {
+  let packageName = `@remix-run/${adapterName}`;
+  let sourceDir =
+    path.relative(
+      process.cwd(),
+      path.join(__dirname, "packages", `remix-${adapterName}`)
+    ) || ".";
+  let outputDir = path.join(buildDir, `node_modules/${packageName}`);
```
MichaelDeBoey, line 61:
```suggestion
  let outputDir = path.join(buildDir, "node_modules", packageName);
```

MichaelDeBoey (changes requested):

packages/remix-cloudflare-pages/rollup.config.js
brophdawg11: Yeah.  I noticed a handful of spots in the current config (i.e., https://github.com/remix-run/remix/blob/main/rollup.config.js#L159) that were hardcoding separators so I figured it might not be currently windows friendly and that could be a task in it's own right - but it does seem to be causing issues in CI so I can get those updated :thumbsup:

packages/create-remix/rollup.config.js
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
MichaelDeBoey, line 1: Maybe we should disable this globally in our root `.eslintrc` file for these files?

DEVELOPMENT.md
```diff
@@ -118,7 +118,7 @@ By default, the Remix `rollup` build will strip any `console.debug` calls to avo
 REMIX_DEBUG=true yarn watch
 ```
 
-**`REMIX_LOCAL_DEV_OUTPUT_DIRECTORY`**
+**`LOCAL_BUILD_DIRECTORY`**
```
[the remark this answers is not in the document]
pcattori: I do think its nice to have a framework-specific prefix for env vars to avoid conflicts with existing env vars. Since the company name is Remix, I think `REMIX_` is good. Maybe there's something better.

rollup.config.js
```diff
@@ -1,938 +1,18 @@
...
+      } catch (e) {
+        return null;
       }
-    },
-  };
-}
+      let packageBuild = require(`.${path.sep}${configPath}`);
+      return packageBuild(options);
+    })
+    .filter((p) => p);
```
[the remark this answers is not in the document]
brophdawg11: even better - we can just return `[]` since it's a `flatMap` and drop the `filter` entirely

packages/remix-dev/rollup.config.js
```diff
@@ -0,0 +1,129 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
MichaelDeBoey, line 1: Seems like we forgot one here 🙈

```suggestion
```

packages/remix/rollup.config.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
MichaelDeBoey, line 1:
```suggestion
```

MichaelDeBoey (approved):

rollup.utils.js
```diff
@@ -298,4 +350,9 @@ module.exports = {
  *   sourceDir: string;
  *   version: string;
  * }} BuildInfo
+ *
+ * @typedef {{
+ *   vars?: Record<string, string[]>;
+ *   types?: Record<string, string[]>;
+ * }} MagicExports
```
MichaelDeBoey, line 357: More in line with the actual `importKind` would be to use `value` & `type`, like we do in `packageExports` of the `replace-remix-imports` migration & of the ESLint warning

https://github.com/remix-run/remix/blob/ec49faa9d8155542e157dbac7977393003f0c7a7/packages/remix-dev/cli/migrate/migrations/replace-remix-imports/transform/mapNormalizedImports/packageExports.ts#L27-L30
https://github.com/remix-run/remix/blob/ec49faa9d8155542e157dbac7977393003f0c7a7/packages/remix-eslint-config/rules/packageExports.js#L1-L4

```suggestion
 * @typedef {{
 *   value?: Record<string, string[]>;
 *   type?: Record<string, string[]>;
 * }} MagicExports
```

MichaelDeBoey (changes requested):

packages/remix-dev/cli/setup.ts
```diff
@@ -74,9 +82,9 @@ export async function setupRemix(platform: SetupPlatform): Promise<void> {
     ".js"
   );
 
-  await fse.writeFile(path.join(remixPkgDir, "index.d.ts"), magicTypes);
-  await fse.writeFile(path.join(remixPkgDir, "index.js"), magicCJS);
-  await fse.writeFile(path.join(remixPkgDir, "esm/index.js"), magicESM);
+  await fse.writeFile(path.join(remixPkgDir, "dist/index.d.ts"), magicTypes);
+  await fse.writeFile(path.join(remixPkgDir, "dist/index.js"), magicCJS);
+  await fse.writeFile(path.join(remixPkgDir, "dist/esm/index.js"), magicESM);
```
MichaelDeBoey, line 87: This will probably make Windows happy again

```suggestion
  await fse.writeFile(path.join(remixPkgDir, "dist", "index.d.ts"), magicTypes);
  await fse.writeFile(path.join(remixPkgDir, "dist", "index.js"), magicCJS);
  await fse.writeFile(path.join(remixPkgDir, "dist", "esm", "index.js"), magicESM);
```

MichaelDeBoey (changes requested):

DEVELOPMENT.md
```diff
@@ -118,7 +118,7 @@ By default, the Remix `rollup` build will strip any `console.debug` calls to avo
 REMIX_DEBUG=true yarn watch
 ```
 
-**`REMIX_LOCAL_DEV_OUTPUT_DIRECTORY`**
+**`REMIX_LOCAL_BUILD_DIRECTORY`**
```
chaance, line 121: This will need to be tested still

brophdawg11: I gave it a quick run and it looks good to me - copying files to the right spots in my sibling app directory:

Screen Shot 2022-06-20 at 9 01 55 PM


Saw one rollup complaint (in `yarn watch` only) about the empty `deno` chunk but I see that on dev as well so I don't think it's related.  It also doesn't break the `yarn watch` command.

Screen Shot 2022-06-20 at 9 00 21 PM

integration/cf-compiler-test.ts
```diff
@@ -192,7 +192,7 @@ test.describe("cloudflare compiler", () => {
       "useTransition",
     ];
     let magicRemix = await fs.readFile(
-      path.resolve(projectDir, "node_modules/remix/index.js"),
+      path.resolve(projectDir, "node_modules/remix/dist/index.js"),
```
MichaelDeBoey, line 195:
```suggestion
      path.resolve(projectDir, "node_modules", "remix", "dist", "index.js"),
```

integration/compiler-test.ts
```diff
@@ -313,7 +313,7 @@ test.describe("compiler", () => {
       "useTransition",
     ];
     let magicRemix = await fs.readFile(
-      path.resolve(fixture.projectDir, "node_modules/remix/index.js"),
+      path.resolve(fixture.projectDir, "node_modules/remix/dist/index.js"),
```
MichaelDeBoey, line 316:
```suggestion
      path.resolve(fixture.projectDir, "node_modules", "remix", "dist", "index.js"),
```

MichaelDeBoey (changes requested):

DEVELOPMENT.md
pcattori: @brophdawg11 The "Generated an empty chunk" part is expected, but the "You must specify "output.file"..." part looks like it might be new since the Rollup upgrade. I think previously Rollup required an input, but not an output.


For Deno we just want to copy things with the `copy` plugin, so we can set `output.file` to some dummy location and that should work. A little hacky, but let's use use Rollup to build _all_ our packages instead of treating `remix-deno` specially.

brophdawg11: It's got something do to with `watch` mode - it happens both on `dev` and this branch but only when you run `yarn watch`, not `yarn build`.  Digging in a bit more now

integration/deno-compiler-test.ts
```diff
@@ -2,7 +2,7 @@ import { test, expect } from "@playwright/test";
 import * as fse from "fs-extra";
 import path from "path";
 import shell from "shelljs";
-import glob from "glob";
+import glob from "fast-glob";
```
chaance, line 5: Updated for consistency, but it should also make the test a tiiiiiiiiiny bit faster

MichaelDeBoey (changes requested):

chaance:
@pcattori I ended up backing out of the `index` abstraction you added. There were a few subtle differences between them that seemed made a difference in the output without passing several additional arguments, so I figured I'd keep the original in tact for now.

Let me know if you think I missed anything and we can re-visit!

packages/remix/rollup.config.js
```diff
@@ -0,0 +1,68 @@
...
+      ],
+    },
+    {
+      external() {
+        return true;
+      },
+      input: `${sourceDir}/index.ts`,
+      output: {
+        format: "esm",
+        dir: path.join(outputDist, "esm"),
```
MichaelDeBoey, line 54: This should be the unused `outputEsmDist` I guess?

```suggestion
        dir: outputEsmDist,
```

MichaelDeBoey (changes requested):

packages/remix/rollup.config.js
chaance: I actually meant to delete the variable, no need for it here IMO

packages/remix/rollup.config.js
```diff
@@ -0,0 +1,68 @@
...
+  // Don't blow away remix magic exports on local builds, since they've
+  // already been configured by postinstall
+  if (process.env.REMIX_LOCAL_BUILD_DIRECTORY) {
+    return [];
+  }
+
+  let sourceDir = "packages/remix";
+  let outputDir = getOutputDir(packageName);
+  let outputDist = path.join(outputDir, "dist");
+  let outputEsmDist = path.join(outputDist, "esm");
```
chaance, line 24:
```suggestion
```

scripts/copy-build-to-dist.mjs
```diff
@@ -7,16 +7,13 @@ const PACKAGES_PATH = path.join(ROOT_DIR, "packages");
 const DEFAULT_BUILD_PATH = path.join(ROOT_DIR, "build");
 
 let activeOutputDir = DEFAULT_BUILD_PATH;
-if (process.env.REMIX_LOCAL_DEV_OUTPUT_DIRECTORY) {
-  let appDir = path.join(
-    ROOT_DIR,
-    process.env.REMIX_LOCAL_DEV_OUTPUT_DIRECTORY
-  );
+if (process.env.REMIX_LOCAL_BUILD_DIRECTORY) {
+  let appDir = path.join(ROOT_DIR, process.env.REMIX_LOCAL_BUILD_DIRECTORY);
```
brophdawg11, line 11: I don't think `copyBuildToDist` plays quite right with this - it seems to copy any `node_modules/*/dist` directory into `packages/*/dist` - not just remix packages.  Might just need something like the following in `getPackageBuildPaths`:

```js
      if (path.basename(moduleDir) === "@remix-run") {
        packageBuilds.push(...(await getPackageBuildPaths(moduleDir)));
      } else if (
        /node_modules\/@remix-run\//.test(moduleDir) ||
        /node_modules\/create-remix/.test(moduleDir) ||
        /node_modules\/remix/.test(moduleDir)
      ) {
        packageBuilds.push(moduleDir);
      }
```


I also think I noticed that `node_modules/remix` wasn't getting created inside of `REMIX_LOCAL_BUILD_DIRECTORY`.  That might only be an issue if you're actively changing `magicExports` stuff though so I don't think that's a big issue if it's a pain to handle

pcattori:
> @pcattori I ended up backing out of the `index` abstraction you added. There were a few subtle differences between them that seemed made a difference in the output without passing several additional arguments, so I figured I'd keep the original in tact for now.
> 
> Let me know if you think I missed anything and we can re-visit!

The main value add were the `magicExport` helpers and those have been obsoleted by the work to move the magic exports out of their source packages. Was nice to see which things had magic exports in a declarative way. `index` and `cli` were just some low-hanging fruit, but not the main value for the abstractions IMO.

So yea, this makes sense!

brophdawg11:
Hey this is passing windows CI tests now?? 🥳   Any idea why?

chaance:
@brophdawg11 Should be good for another pass here.

> Hey this is passing windows CI tests now?? 🥳 Any idea why?

My rule of thumb is to never ask this question and just be thankful (meaning, I have no idea why 😅)

packages/remix-dev/rollup.config.js
```diff
@@ -96,11 +96,20 @@ module.exports = function rollup() {
...
+          banner: createBanner("@remix-run/dev", version, true),
+          dir: outputDir,
+          format: "cjs",
+        },
+        {
+          banner: createBanner("@remix-run/dev", version, true),
+          dir: outputDist,
+          format: "cjs",
+        },
+      ],
```
brophdawg11, line 112: Looks like this didn't get carried through from https://github.com/remix-run/remix/pull/3607 with the merges

brophdawg11:
Looking good!  I ran a fresh build on dev and this branch and found a few small magic exports deltas - but unsure if I know which is correct:

* This branch has `export type { UploadHandler, UploadHandlerPart } from '@remix-run/node'` in `@remix-run/architect/dist/magicExports/remix.d.ts` and `dev` does not 
* `dev` exports `ActionArgs` and `LoaderArgs` in `@remix-run/server-runtime/dist/magicExports/remix.d.ts` but this branch does not


Otherwise the builds looked identical minus two small `/* webpackIgnore: true */` comments on some dynamic imports

MichaelDeBoey:
> * This branch has `export type { UploadHandler, UploadHandlerPart } from '@remix-run/node'` in `@remix-run/architect/dist/magicExports/remix.d.ts` and `dev` does not
> * `dev` exports `ActionArgs` and `LoaderArgs` in `@remix-run/server-runtime/dist/magicExports/remix.d.ts` but this branch does not

This should be like we have on `dev`

chaance:
@brophdawg11 Addressed both of those issues. Confirmed that magic exports match what's on dev, and the webpack comment should be included here 👍

ronnylt:
> We can now do individual package builds via npx rollup -c directly from within a packages/{name} directory if we don't need to build everything. This also "just works" with things like LOCAL_BUILD_DIRECTORY and rollup -w - both from the root and from individual packages

I am trying to build individual packages but it's not working:
```

remix/packages/remix-vercel (dev ✔) ᐅ npx rollup -c

packages/remix-vercel/index.ts → build/node_modules/@remix-run/vercel/dist...
[!] Error: Could not resolve entry module (packages/remix-vercel/index.ts).
Error: Could not resolve entry module (packages/remix-vercel/index.ts).
    at error (/Users/ronnylt/Workspace/opensource/remix/node_modules/rollup/dist/shared/rollup.js:198:30)
    at ModuleLoader.loadEntryModule (/Users/ronnylt/Workspace/opensource/remix/node_modules/rollup/dist/shared/rollup.js:22527:20)
    at async Promise.all (index 0)
```

PR #3391, opened 2022-06-05
fix(remix-dev): Allow debugging for the remix shorthand

machour:
I simply merged the default case with the dev case.
This allow to use the `--debug` CLI flag when using the `remix dev` shorthand. 

Testing Strategy:

I've built and tested from one of my projects using `node ../remix/build/node_modules/\@remix-run/dev/cli.js --debug`

### DIFF
--- a/packages/remix-dev/cli/run.ts
+++ b/packages/remix-dev/cli/run.ts
@@ -126,6 +126,12 @@ const npxInterop = {
   pnpm: "pnpm exec",
 };
 
+async function dev(projectDir: string, flags: { debug: boolean|undefined }) {
+  if (!process.env.NODE_ENV) process.env.NODE_ENV = "development";
+  if (flags.debug) inspector.open();
+  await commands.dev(projectDir, process.env.NODE_ENV);
+}
+
 /**
  * Programmatic interface for running the Remix CLI with the given command line
  * arguments.
@@ -426,9 +432,10 @@ export async function run(argv: string[] = process.argv.slice(2)) {
       break;
     }
     case "dev":
-    default: // `remix ./my-project` is shorthand for `remix dev ./my-project`
-      if (!process.env.NODE_ENV) process.env.NODE_ENV = "development";
-      if (flags.debug) inspector.open();
-      await commands.dev(input[1], process.env.NODE_ENV);
+      await dev(input[1], flags);
+      break;
+    default:
+      // `remix ./my-project` is shorthand for `remix dev ./my-project`
+      await dev(input[0], flags);
   }
 }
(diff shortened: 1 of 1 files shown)


packages/remix-dev/cli/run.ts
```diff
@@ -426,13 +426,10 @@ export async function run(argv: string[] = process.argv.slice(2)) {
       break;
     }
     case "dev":
+    default: // `remix ./my-project` is shorthand for `remix dev ./my-project`
       if (!process.env.NODE_ENV) process.env.NODE_ENV = "development";
       if (flags.debug) inspector.open();
       await commands.dev(input[1], process.env.NODE_ENV);
       break;
```
MichaelDeBoey, line 433: You can delete this last `break` now, since it's not necessary in the `default` case.

```suggestion
```

MichaelDeBoey (changes requested):

packages/remix-dev/cli/run.ts
```diff
@@ -426,13 +426,9 @@ export async function run(argv: string[] = process.argv.slice(2)) {
...
     }
     case "dev":
+    default: // `remix ./my-project` is shorthand for `remix dev ./my-project`
       if (!process.env.NODE_ENV) process.env.NODE_ENV = "development";
       if (flags.debug) inspector.open();
       await commands.dev(input[1], process.env.NODE_ENV);
```
pcattori, line 432: `input[1]` will return `"dev"` for `remix dev ./my-project`, no? So we need to account for that here.

---

I think it might make more sense to keep the cases separate, but factor out the common logic:

```ts
async function dev(projectDir: string) {
  if (!process.env.NODE_ENV) process.env.NODE_ENV = "development";
  if (flags.debug) inspector.open();
  await commands.dev(projectDir, process.env.NODE_ENV);
}


// ...
case "dev": { await dev(input[1]); break; }
case default: { await dev(input[0]); break; }
```

pcattori (changes requested):

packages/remix-dev/cli/run.ts
MichaelDeBoey: Indeed important oversight! 👀

pcattori: Also, we should have a CLI test for the shorthand to make sure we catch this sort of thing in the future. @machour : could you add a test to that effect?

machour: glad you caught that @pcattori 😱 I'll make the requested change and test

machour: I definitely need some guidance for the test part, as the `dev` command itself is not currently tested, and I'm a bit clueless about how it should be done.

machour: I could test that both `remix non-existing-dir` and `remix dev non-existing-dir` print an error matching `Error loading Remix config in .*/non-existing-dir/remix.config.js` confirming that they both really do the same thing with args, but that seem a bit flaky to me 😅

packages/remix-dev/cli/run.ts
```diff
@@ -126,6 +126,12 @@ const npxInterop = {
   pnpm: "pnpm exec",
 };
 
+async function dev(projectDir: string, flags: { debug: boolean|undefined }) {
```
pcattori, line 129: nit: 
```suggestion
async function dev(projectDir: string, flags: { debug?: boolean }) {
```

Issue #211, opened 2021-06-28
Firebase functions & Express - Empty request body


TheRealFlyingCoder:
### Issue
Action Request body is empty after making the journey through Firebase `https.onRequest()` -> Express -> `createRequestHandler()`

### Notes
Express request object has the body before `createRequestHandler` as you can successfully pass it in the Context object

### Replication
I've created a little solution that replicates the problem as simply as possible.

The button posts a hidden field value to the action, if the value exists in the body it redirects to "hasToken" otherwise "noToken"

There are 3 tests which simply change the server configuration:

1. Remix run -> Should work successfully
2. Express -> Should work successfully
3. Firebase Emulator -> Should fail

It needs the Firebase Emulator suite to do the failing test unfortunately, but it's not too big a hassle to install.

[ExpressTest.zip](https://github.com/remix-run/remix/files/6722575/ExpressTest.zip)

ccssmnn:
I was running into this when I tried to prepare a Remix + Firebase example. In the Cloud Function [docs](https://firebase.google.com/docs/functions/http-events#read_values_from_the_request), they say, that request bodies are parsed before the request is processed by the Cloud Function handler. That's why it works with Remix run and standalone Express, but not in the emulator.

As answered [here](https://stackoverflow.com/questions/70992856/disable-undo-firebase-cloud-functions-default-request-parsing), there is no way to disable this behavior.

gustavopch:
For anyone having this issue, you can make it work with this patch:
```diff
diff --git a/node_modules/@remix-run/express/dist/server.js b/node_modules/@remix-run/express/dist/server.js
index 4ea186f..2c63e13 100644
--- a/node_modules/@remix-run/express/dist/server.js
+++ b/node_modules/@remix-run/express/dist/server.js
@@ -74,7 +74,7 @@ function createRemixRequest(req, res) {
     signal: controller.signal
   };
   if (req.method !== "GET" && req.method !== "HEAD") {
-    init.body = req;
+    init.body = req.rawBody ?? req;
   }
   return new node.Request(url.href, init);
 }
```

PR #3397, opened 2022-06-06
feat: add `@remix-run/google-cloud-functions` package

penx:
Required adapter in order to provide example of deploying to Firebase functions

Closes: #125
Closes: #211

- [ ] Docs
- [x] Tests

Testing Strategy:

1. Added unit tests
2. [Created a branch](https://github.com/penx/remix/tree/firebase-adapter-integration) that [uses this adapter for all integration tests](https://github.com/penx/remix/commit/a9d8fa694b913a76adeb45e500f69d3da7e852fe)


package.json
```diff
@@ -82,7 +83,7 @@
     "eslint-plugin-markdown": "^2.2.1",
     "eslint-plugin-prefer-let": "^3.0.1",
     "express": "^4.17.1",
-    "jest": "^27.5.1",
+    "jest": "^28.1.0",
```
penx, line 86: This is required in order to 

```ts
import { getTestServer } from "@google-cloud/functions-framework/testing";
```
https://github.com/penx/remix/blob/29be5a6c90e9d1a02db06d6e9f306947ef89d452/packages/remix-google-cloud-functions/__tests__/server-test.ts#L10-L11

MichaelDeBoey:
@penx This seems to be a duplicate of #2266?
Could you please elaborate on the difference?

package.json
penx:
```ts
import { getTestServer } from "@google-cloud/functions-framework/testing";
```

This import makes use of [package exports](https://nodejs.org/api/packages.html#exports):

https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/759aac18ad297363c4651efcf19ab2bf8fa9625a/package.json#L13

Full support for this [was added in Jest 28](https://jestjs.io/blog/2022/04/25/jest-28#packagejson-exports).

Without Jest 27,  server-test.ts will fail saying it cannot find `@google-cloud/functions-framework/testing`.

MichaelDeBoey: Let's do this in a separate PR instead, so we keep this PR focussed.

penx:
> @penx This seems to be a duplicate of #2266?
> Could you please elaborate on the difference?

Sure, I was working on this independently and was going to name it `@remix-run/firebase` as per https://github.com/remix-run/remix/issues/125 before realising a more generic cloud-functions was probably a better route given this is what Firebase functions are running on. As such I hadn't seen the above PR until earlier today.

It appears #2266 has the same purpose but with a slightly different approach.

I would suggest using my approach here for various reasons:


## Implementation

### rawBody

#2266 has the following code which I think is the main difference:


```ts
/**
  * Google Cloud Functions includes middleware that processes incoming requests based on their headers and sets the request body to
  * a javascript object. But Remix doesn't like that, remix has its own request processing logic, so we have to turn the body back into
  * something that Remix is expecting. In this case a Buffer with URLSearchParams encoded data works.
  *
  * @param req the request passed in from the Google Cloud Function
  */
 function createRemixBody(req: GcfRequest) {
   let s = new Readable()
   s.push((new URLSearchParams(req.body).toString()).toString());
   s.push(null);
   return s;
 }
```
https://github.com/remix-run/remix/pull/2266/files#diff-c7fe10f958a1c448a6f62cb813a2868c7b075f470304598ca79e26eb167f018cR141-R153

We don't need to create a new buffer on each request as `rawBody` is available as a property on `req` that provides this.

Here's the equivalent code in my version:

https://github.com/penx/remix/blob/29be5a6c90e9d1a02db06d6e9f306947ef89d452/packages/remix-google-cloud-functions/server.ts#L94-L96

This comes from @ryanflorence 's recommendation on Discord:

https://discord.com/channels/770287896669978684/771068344320786452/905168966706675782

https://gist.github.com/ryanflorence/8a12cef023765bd0520a04cf25d07cf7#file-remix-firebase-js-L49-L51

### Types

I've used the official types from `@google-cloud/functions-framework` and avoided referencing Express directly due to subtle differences such as body + rawBody. This should help identify issues if the official types move further away from Express types in the future.

## Testing

- I've added unit tests based on the Express adapter but using getTestServer from [GoogleCloudPlatform/functions-framework-nodejs](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/docs/testing-functions.md)
- In a [separate branch](https://github.com/penx/remix/tree/firebase-adapter-integration) I've [confirmed all integration tests pass using this adapter](https://github.com/penx/remix/commit/a9d8fa694b913a76adeb45e500f69d3da7e852fe), and would be happy to integrate these in to this PR if a maintainer wants to suggest a way to swap out and test with different adapters.

## Naming

- This adapter works with both google cloud functions v1 and v2, which I expect #2266 does too, but I think there's no need to include v2 in the adapter name in

package.json
MichaelDeBoey: Let's do this in a separate PR instead, so we keep this PR focussed.

I think the Jest 28 change is also causing some tests to fail, so we could solve them in isolation in the new PR.

MichaelDeBoey:
@penx I suggest you to join forces with @eastlondoner on #2266, since he started first on this idea.
You can add suggestions on his PR and/or create a PR against his branch (on his repo).

This way we can focus on keeping all related discussions on the same PR.

penx:
I've left feedback on #2266 and opened https://github.com/eastlondoner/remix/pull/1 with the unit tests

package.json
MichaelDeBoey: @penx Now that https://github.com/GoogleCloudPlatform/functions-framework-nodejs/pull/461 is merged, do we still need `jest` v28?

Issue #3314, opened 2022-05-27
Live reload doesn't work with a custom Express setup


illright:
### What version of Remix are you using?

1.5.1

### Steps to Reproduce

Create a new Remix app and configure a custom Express server as described in [this example](https://github.com/remix-run/remix/tree/main/examples/socket.io).

Run the development server and change some code. See the error in the browser console about not being able to connect to port 8002 and see no live reloading happen.

Terminate the development server and restart it, this time with an explicitly set environment variable, `REMIX_DEV_SERVER_WS_PORT=8002`. See no errors in the console and see live reloading happen upon code changes.

### Expected Behavior

Live reloading defaults to port 8002 and works without explicitly setting the port, even for custom Express-based server setups.

Alternatively, I'd like this behaviour to be mentioned in the README of the repos that have a custom Express server because I had to figure this out by reading the source code. 

### Actual Behavior

Live reloading requires extra steps on every run for custom Express-based server setups.

kiliman:
I did come up with a work-around if you need random ports with the Express adapter. The trick is to set the random port *before* calling `run-p dev:*`.

There's a CLI version of `get-port` which output the random port to *stdout*.
`npm i -D get-port-cli`

I wrote a script that sets the port, before calling remix/express (had to rename it to `start:dev`).

```json
  "scripts": {
    "build": "remix build",
    "dev": "cross-env-shell NODE_ENV=development ./start-remix-dev",
    "start:dev": "run-p dev:*",
    "dev:node": "nodemon ./server.js --watch ./server.js",
    "dev:remix": "remix watch",
    "start": "cross-env NODE_ENV=production node ./server.js"
  },
```
`start-remix-dev` script
```bash
#!/usr/bin/env bash
REMIX_DEV_SERVER_WS_PORT=$(npx get-port)
echo "REMIX_DEV_SERVER_WS_PORT: $REMIX_DEV_SERVER_WS_PORT"
export REMIX_DEV_SERVER_WS_PORT
npm run start:dev
```

illright:
Thanks for your input! I'd rather stick to a hardcoded port in `remix.config.js` for now to keep complexity out of my repo and keep it cross-platform.

```diff
 /**
  * @type {import('@remix-run/dev').AppConfig}
  */
 module.exports = {
   cacheDirectory: "./node_modules/.cache/remix",
+  devServerPort: 8002,
[...]

PR #3411, opened 2022-06-07
chore(remix-react): make LiveReload port required

mcansh:
without this change, unbundled servers wouldn't properly connect to the web socket connection

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/examples/basic/app/root.tsx
+++ b/examples/basic/app/root.tsx
@@ -75,7 +75,7 @@ function Document({
         <RouteChangeAnnouncement />
         <ScrollRestoration />
         <Scripts />
-        <LiveReload />
+        <LiveReload port={Number(process.env.REMIX_DEV_SERVER_WS_PORT)} />
       </body>
     </html>
   );
--- a/examples/blog-tutorial/app/root.tsx
+++ b/examples/blog-tutorial/app/root.tsx
@@ -47,7 +47,7 @@ export default function App() {
         <Outlet />
         <ScrollRestoration />
         <Scripts />
-        <LiveReload />
+        <LiveReload port={Number(process.env.REMIX_DEV_SERVER_WS_PORT)} />
       </body>
     </html>
   );
--- a/examples/bullmq-task-queue/app/root.tsx
+++ b/examples/bullmq-task-queue/app/root.tsx
@@ -25,7 +25,7 @@ export default function App() {
         <Outlet />
         <ScrollRestoration />
         <Scripts />
-        <LiveReload />
+        <LiveReload port={Number(process.env.REMIX_DEV_SERVER_WS_PORT)} />
       </body>
     </html>
   );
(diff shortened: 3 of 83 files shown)


kiliman:
Ugh. Not a fan. 😒

Also, this is a breaking change for *every* Remix app out there.

mcansh:
> Ugh. Not a fan. 😒
> 
> Also, this is a breaking change for _every_ Remix app out there.

hoping to find a better solution (which is why this is still a draft), however it will still try to use infer `process.env.REMIX_DEV_SERVER_WS_PORT` and only throw if it's still undefined

kiliman:
I think for the most part, *remix.config* `devServerPort` is an acceptable solution. Sure, it's nice to be able to have a random port, but not at the expense of breaking code.

I have posted a workaround for random ports with multiple processes, but it required a "parent" process to get the port and set the environment variable, so it could be inherited by child processes. It definitely complicated the startup.

mcansh:
> I think for the most part, _remix.config_ `devServerPort` is an acceptable solution. Sure, it's nice to be able to have a random port, but not at the expense of breaking code.
> 
> I have posted a workaround for random ports with multiple processes, but it required a "parent" process to get the port and set the environment variable, so it could be inherited by child processes. It definitely complicated the startup.

devServerPort sets the environment variable, which is where the issue comes in as it does set it, but the web socket is set up in the browser where the variable isn't available and then defaults to 8002 - the fact that it's random and not 8002 by default is a different bug

kiliman:
Yes, I understand that. I even posted about it in #3314. And even included a "solution" to support random ports with environment variables for dual-process setups like Express.

All I'm saying is that everything worked fine when we simply used `devServerPort`. Granted that makes it harder to have multiple apps running, so you need to constantly update the config.

I do like the random port, and I realize that to share the port with Remix *and* the web server, you need to use an environment variable. The issue is you can't share the environment variable across sibling processes.

Perhaps `remix dev` can have a parameter that allows you to specify the process to run for the "web" server, where it defaults to `remix-serve` RAS.

`remix dev --server "node ./server.js"`. 

That way it since it is the one launching the web process, it can share environment variables across the compiler and web and everything will work properly.

Anyway, just a thought.

kentcdodds:
I need help understanding why we have this problem in the first place.

This solution won't work. We'll need to come up with something else.

mcansh:
> I need help understanding why we have this problem in the first place.

The issue comes down to us not bundling the `@remix-run/react` package in node targets. This leaves the reading of the port a runtime thing requiring `process.env.REMIX_XXX_PORT` var to be set on the process running the server.

```ts
<LiveReload port={Number(process.env.REMIX_DEV_SERVER_WS_PORT)} />
``` is not visually pleasing, but it does work

more info: https://discord.com/channels/770287896669978684/983424417352777728/983831766508908544

PR #3417, opened 2022-06-08
feat(examples/firebase-auth-firestore): avoid auth rate limiting

penx:
Use the Firebase Auth REST API client side to avoid rate limiting.

This:
- does not require the Firebase SDK client side
- falls back to a standard form post and server side auth
- has a very small client side bundle impact.

Reservations have already been made at https://github.com/remix-run/remix/pull/3362#issuecomment-1149193412, and I have no objection to this being closed if it's not wanted, however if server side auth is rate limited to 150 requests/IP address/hour (which I understand to be the case based on reports [here](https://twitter.com/1fiws/status/1521426483037839362) and [here](https://discord.com/channels/770287896669978684/777863722592501780/963845752986161162) but have not tested) then the current example would not be suitable for many production applications. I am opening this as draft to discuss issues with and alternatives to this approach.

This is in draft:

- pending information from Firebase support on rate limiting from the SignInWithPassword Firebase REST endpoint (I have contacted Firebase support)
- if SignInWithPassword is rate limited, to see if Firebase support can suggest alternatives for server side auth

The best official source I can find is the following. Though specific rates for login per IP address are not mentioned, I think it may fall under "Verification requests":

- https://firebase.google.com/docs/auth/limits

Closes: Discussion at:

- https://twitter.com/1fiws/status/1521426483037839362
- https://discord.com/channels/770287896669978684/777863722592501780/963845752986161162

f

- [ ] Docs
- [ ] Tests


examples/firebase-auth-firestore/app/routes/login.tsx
```diff
@@ -44,12 +70,38 @@ export const action: ActionFunction = async ({ request }) => {
...
 
 export default function Login() {
+  const [clientAction, setClientAction] = useState<ActionData>();
   const action = useActionData<ActionData>();
+  const restConfig = useLoaderData<LoaderData>();
+  const submit = useSubmit();
+
+  const handleSubmit = useCallback(
+    async (event: React.FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
```
penx, line 80: this line is my main issue with this approach, as if something fails after this point it will not fall back to the form post + server side auth

kentcdodds:
I'm ready to merge this once you mark it as not a draft.

PR #3422, opened 2022-06-08
types(remix-server-runtime): make AppLoadContext an empty interface instead of any

marwan38:
👋 Hey, 

This PR builds on top of another PR that was created https://github.com/remix-run/remix/pull/1876.
I found that there were some issues with typing due to the original change of the `AppLoadContext` type to an interface.

The issue was this:
```typescript
- type AppLoadContext = any;
+ interface AppLoadContext {
+    [key: string]: unknown;
+ }

// Other functions that rely on the context
const someLoaderThing = (loadContext: unknown) => { // ... } // type error, type unknown conflicts with the new interface
```

Based on the fact that the documentation shows that the `getLoadContext` function returns an object https://remix.run/docs/en/v1/other-api/adapter#createrequesthandler, I thought that original change to `AppLoadContext` was suitable.

I cherry picked the original commits to ensure that the credits to the original author remain.

Closes: #1876

Testing Strategy:

I initiated a new playground and added the typescript override within the default `remix.env.d.ts`.

The override looks like this:
```typescript
/// <reference types="@remix-run/node/globals" />
import "@remix-run/server-runtime";

declare module "@remix-run/server-runtime" {
  interface AppLoadContext {
    foo: {
      bar: true;
    };
  }
}

```

Success! Context within action and loaders are now strongly typed.

### DIFF
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -6,7 +6,9 @@ import { json, isResponse, isRedirectResponse } from "./responses";
  * An object of arbitrary for route loaders and actions provided by the
  * server's `getLoadContext()` function.
  */
-export interface AppLoadContext {};
+export interface AppLoadContext {
+  [key: string]: unknown;
+};
 
 /**
  * Data for a route that was returned from a `loader()`.
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -7,7 +7,7 @@ import { json, isResponse, isRedirectResponse } from "./responses";
  * server's `getLoadContext()` function.
  */
 export interface AppLoadContext {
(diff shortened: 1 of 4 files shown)


packages/remix-server-runtime/data.ts
```diff
@@ -18,7 +20,7 @@ export async function callRouteAction({
   match,
   request,
 }: {
-  loadContext: unknown;
+  loadContext?: AppLoadContext;
```
marwan38, line 23: I did not take too much time in learning what the implementation details were of changing this (and the following) types like this. Relying on reviewers to correct me if I'm wrong here.

MichaelDeBoey:
Please don't just create a new PR with any changes, but make a suggestion on the original PR instead.

Going to close this one to keep all related discussions in the same PR.

PR #3433, opened 2022-06-09
docs(examples): add file upload to S3 example

GSt4r:
Since no example on the internet seemed to handle `AsyncIterable` correctly, I decided to create a simple example on how to upload a file to an S3 bucket. This example is based on the cloudinary one.


examples/file-and-s3-upload/app/routes/s3-upload.tsx
```diff
@@ -0,0 +1,59 @@
...
+    });
+  }
+  return json({
+    imgSrc,
+    imgDesc,
+  });
+};
+
+export default function Index() {
+  const data = useActionData<ActionData>();
```
kentcdodds, line 38: Because this is not a navigation we should use `useFetcher` instead. Could you update this example to use that instead?

kentcdodds (changes requested):
This looks awesome! Thank you for taking the time to put this together. Could you make one change?

examples/file-and-s3-upload/app/routes/s3-upload.tsx
GSt4r: Hey Kent, thanks for reviewing this! I've updated the example to use `useFetcher` instead

kentcdodds (approved):
Thank you!

PR #3434, opened 2022-06-10
feat: the real implementation of deferred on the old codebase that's no longer valid and superseded by yet another PR built on top of React Router where the real deferred implementation actually lives

jacob-ebey:
🚨 **IMPORTANT:** Revert any changes to `version` and `release` scripts before merging 🚨

---


packages/remix-serve/index.ts
```diff
@@ -8,7 +8,26 @@ export function createApp(buildPath: string, mode = "production") {
...
+          req.headers["accept"] &&
+          req.headers["accept"].indexOf("text/html") !== -1
+        ) {
+          return false;
+        }
+
+        // No compression for _data requests to allow for streaming
+        if (req.query._data) {
+          return false;
+        }
```
sergiodxa, line 25: What if a resource route needs streaming? Wouldn't it be better to apply the compression middle only on the static files?

packages/remix-server-runtime/__tests__/responses-test.ts
```diff
@@ -1,4 +1,230 @@
...
+    expect(initialData).toEqual({
+      foo: "remix",
+      bar: DEFERRED_PROMISE_PREFIX + "bar",
+      baz: DEFERRED_PROMISE_PREFIX + "baz",
+    });
+
+    let [secondChunkType, secondChunkData] = chunkTypeAndData(decodedChunks[1]);
+    let secondChunkDataKeys = Object.keys(secondChunkData);
+    expect(secondChunkDataKeys.length).toBe(1);
+    if (secondChunkType === "error") {
```
mjackson, line 192: Maybe store all chunks in an array, then go find which one is the error/data to avoid duplicating if/else if/else statements.

packages/remix-serve/index.ts
```diff
@@ -8,7 +8,26 @@ export function createApp(buildPath: string, mode = "production") {
 
   app.disable("x-powered-by");
 
-  app.use(compression());
+  app.use(
+    compression({
```
mjackson, line 12: Maybe utilitize this function (for other remix-express users that aren't using one of our templates) and possibly check res content-type to know what to compress instead of checking the req.

packages/remix-serve/index.ts
jacob-ebey: I'm going to explore switching on the response content-type instead.

packages/remix-react/components.tsx
```diff
@@ -371,6 +375,219 @@ export function RemixRoute({ id }: { id: string }) {
...
+          ctx.data = error;
+        }
+      }
+    );
+  }
+
+  if (data instanceof Error) {
+    if (
+      typeof errorBoundary === "undefined" &&
+      typeof document !== "undefined"
```
mjackson, line 530: Could we make this a little more obvious with a `isSsr` module var?

docs/guides/streaming.md
```diff
@@ -0,0 +1,349 @@
...
+      : await packageLocationPromise,
+  });
+};
+```
+
+That `shouldDeferPackageLocation` could be implemented to check the user making the request, whether the package location data is in a cache, the status of an A/B test, or whatever else you want. This is pretty sweet 🍭
+
+## When to not use streaming
+
+TODO: Talk about no streaming on action reloads.
```
kentcdodds, line 342: @jacob-ebey, can you help me come up with an example of skipping deferring on revalidations?

kentcdodds (commented):
Just pushed some docs changes. Feedback welcome.

docs/guides/streaming.md
brophdawg11: Jacob should confirm but I believe `deferred` should be automatically "disabled" on revalidations as we just await all the promises before resolving the `loader` if it's not an initial request

docs/guides/streaming.md
```diff
@@ -0,0 +1,349 @@
+---
+title: Streaming
+description: When, why, and how to stream with React 18 and Remix's deferred API.
+---
+
+# Streaming
+
+Remix supports the [web streaming API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) as a first-class citizen. Additionally, JavaScript server runtimes have support for streaming responses to the client.
+
+One problem streaming in React 18 solves is for lazy-loading chunks of your code you don't want in your server render (for example, if a component doesn't support server-rendering). Once you enable React 18 streaming, you can use [`React.lazy`][react-lazy] on the server and the fallback will be rendered on the server and the client will fetch the code and render the components.
```
jacob-ebey, line 10: @kentcdodds I don't believe this is 100% accurate. React.lazy *will* server render in React 18 when using one of the streaming renderer's instead of renderToString. We will be requiring one of those streaming renderers when using deferred, so I would re-word this or remove it.

kentcdodds: Wait. To be clear... If you're using `renderToString` (even with React 18) you can't use `React.lazy` right? That's what I'm trying to communicate here. That enabling streaming by using `renderToPipeableStream` is required if you want to use `React.lazy` on the server render. Am I mistaken?

jacob-ebey: It's correct that `renderToString` will bail out on the server due to usage of streaming features (suspense) without using a streaming compatible rendering method . That is a byproduct of attempting to use streaming features in an invalid context and isn't really a "streaming feature", more of a byproduct of improper use and isn't behavior I would encourage relying on in general as as soon as you do actually start streaming that use-case breaks.

packages/remix-server-runtime/__tests__/responses-test.ts
jacob-ebey: These test have been cleaned up and moved to the shared package.

docs/guides/typescript.md
```diff
@@ -76,4 +76,135 @@ Remix has TypeScript type definitions built-in as well. The starter templates cr
 
 <docs-info>Note that the types referenced in `remix.env.d.ts` will depend on which environment you're running your app in. For example, there are different globals available in Cloudflare</docs-info>
 
+## Type Utilities
+
+Remix's built-in inference between `loader` function + `useLoaderData` generic and the `action` function + `useActionData` generic is quite good and you shouldn't often have to use thse utilities. That said they can be quite handy in some cases which is why they are exported.
```
tatchi, line 81:
```suggestion
Remix's built-in inference between `loader` function + `useLoaderData` generic and the `action` function + `useActionData` generic is quite good and you shouldn't often have to use those utilities. That said they can be quite handy in some cases which is why they are exported.
```

kentcdodds: Thanks, but it's correct as-is.

tatchi: Hey, what does `thse` mean? Is that an abbreviation for `these` ?

kentcdodds: 🤦‍♂️ sorry. I meant to write `these` and when I reviewed your PR I failed to see that it was missing the `e` 😅

kentcdodds, line 81:
```suggestion
Remix's built-in inference between `loader` function + `useLoaderData` generic and the `action` function + `useActionData` generic is quite good and you shouldn't often have to use these utilities. That said they can be quite handy in some cases which is why they are exported.
```

kentcdodds: Fixed. Thank you very much!

docs/guides/streaming.md
```diff
@@ -0,0 +1,369 @@
...
+
+          responseHeaders.set("Content-Type", "text/html");
+
+          resolve(
+            new Response(body, {
+              status: didError ? 500 : responseStatusCode,
+              headers: responseHeaders,
+            })
+          );
+          pipe(body);
```
nicksrandall, line 179: It's a best practice to add the `<!DOCTYPE html>` tag to prevent browser from going into "quirks mode". 

```suggestion
          body.write("<!DOCTYPE html>");
          pipe(body);
```

docs/guides/streaming.md
```diff
@@ -0,0 +1,369 @@
...
+    const { pipe, abort } = renderToPipeableStream(
+      <RemixServer
+        context={remixContext}
+        url={request.url}
+      />,
+      {
+        [callbackName]() {
+          let body = new PassThrough();
+
+          responseHeaders.set("Content-Type", "text/html");
```
nicksrandall, line 171: Lighthouse audits say that you should add the `charset` in the first 1024 bytes streamed from the server. This can also be done via a meta tag but I feel like this is the easiest place to add it. 

```suggestion
          responseHeaders.set("Content-Type", "text/html; charset=UTF-8");
```

kentcdodds: Perhaps we should update all the example to do this 🤔

MichaelDeBoey: According to @mathiasbynens, it's a best practice to use the `<meta charset="utf-8">` (see https://github.com/GoogleChrome/lighthouse/issues/10023#issuecomment-558521860)

nicksrandall: Ok. I can see the argument to keep things as is. 

I using a CCS-in-JS library (I know it’s something the Remix community thinks should die) which automatically injects style tags into the top of the head in document. In my case, this bumped the meta tag to be too low in the document to work. Again, I understand that is an edge case so I will close this.

packages/remix-react/components.tsx
```diff
@@ -1335,8 +1601,38 @@ type JsonPrimitives =
   | Number
   | Boolean
   | null;
+
 type NonJsonPrimitives = undefined | Function | symbol;
-type SerializeType<T> = T extends JsonPrimitives
+
+type SerializeType<T, UseDeferred = false> = UseDeferred extends false
```
kentcdodds, line 1607: @colinhacks and @mattpocock would either of you be able to review some of this TypeScript magic? 😅

jacob-ebey: Same network serialization type guard but with promise support:
```ts
type Test = SerializeType<{ a: Promise<Date>, b: Date }, true>;
// Test should equal Test2
type Test2 = { a: Promise<string>, b: string };
```

jacob-ebey:
deferred-benefits

docs/guides/streaming.md
```diff
@@ -0,0 +1,372 @@
...
+
+But there's a more nuanced problem that streaming solves. Let's look into that:
+
+## The problem
+
+Imagine a scenario where one of your routes' loaders needs to retrieve some data that for one reason or another is quite slow. For example, let's say you're showing the user the location of a package that's being delivered to their home:
+
+```tsx
+import type { LoaderFunction } from "@remix-run/node";
+import { json } from "@remix-run/node";
```
MichaelDeBoey, line 20:
```suggestion
import type { LoaderFunction } from "@remix-run/node"; // or cloudflare/deno
import { json } from "@remix-run/node"; // or cloudflare/deno
```

docs/guides/streaming.md
```diff
@@ -0,0 +1,372 @@
...
+
+With just that in place, you're unlikely to see any significant performance improvement. But with that alone you can now use [`React.lazy`][react-lazy] to SSR components but delay hydration on the client. This can open up network bandwidth for more critical things like styles, images, and fonts leading to a better LCP and TTI.
+
+### Using `deferred`
+
+With React streaming setup, now you can start adding `Deferred` usage for your slow data requests where you'd rather render a fallback UI. Let's do that for our example above:
+
+```tsx lines=[3,5,6,11,18,23,33-46]
+import type { LoaderFunction } from "@remix-run/node";
+import { deferred } from "@remix-run/node";
```
MichaelDeBoey, line 219:
```suggestion
import type { LoaderFunction } from "@remix-run/node"; // or cloudflare/deno
import { deferred } from "@remix-run/node"; // or cloudflare/deno
```

jacob-ebey:
> Could you double-check the highlighted lines?

Just went through them and they look good to me.

packages/remix-react/data.ts
```diff
@@ -54,11 +57,41 @@ export async function fetchData(
...
+    ) {
+      return deferred.criticalData;
+    }
+
+    let isArray = Array.isArray(deferred.criticalData);
+    let data = deferred.criticalData as Record<string | number, unknown>;
+    for (let entry of Object.entries(deferred.deferredData)) {
+      let key = isArray ? Number(entry[0]) : entry[0];
+      data[key] = await entry[1];
+    }
```
mjackson, line 91: Let's extract this into its own utility method in the remix-deferred package.

packages/remix-react/transition.ts
```diff
@@ -771,6 +822,22 @@ export function createTransitionManager(init: TransitionManagerInit) {
...
+      pendingNavigationDeferredControllers
+    );
+    // let deferredLoaderData, monitorDeferred;
+    // [results, deferredLoaderData, monitorDeferred] =
+    //   await makeDeferredLoaderData(
+    //     getState,
+    //     update,
+    //     results,
+    //     pendingNavigationDeferredControllers
+    //   );
```
mjackson, line 839: TODO: remove

packages/remix-react/components.tsx
```diff
@@ -859,10 +1056,107 @@ window.__remixRouteModules = {${matches
...
+  let scripts = React.useMemo<null | React.ReactNode[]>(() => {
+    if (!deferredLoaderData) {
+      return null;
+    }
+
+    let scripts: React.ReactNode[] = [];
+    for (let [routeId, deferredKeys] of Object.entries(
+      deferredLoaderData || {}
+    )) {
+      if (!routeData[routeId] || typeof routeData[routeId] !== "object") {
```
mjackson, line 1076: Needs a promise check.

decisions/0004-deferred.md
```diff
@@ -0,0 +1,53 @@
...
+### Server API:
+
+Introduces a new response helper called `defer()` that's similar to the existing `json()` function. `defer()` allows you to return an object / array with Promises as entries. This will ultimately return a `Response` that sends the critical data (non-promise entries), then as the promises resolve, streams the results or errors.
+
+#### **defer(result, init)**
+
+Definition:
+
+```ts
+export type DeferredFunction = <Data extends unknown = unknown>(
```
kentcdodds, line 36: Should this be renamed to `DeferFunction` now since the function is now called `defer` instead of `deferred`?

docs/guides/typescript.md
```diff
@@ -76,4 +76,55 @@ Remix has TypeScript type definitions built-in as well. The starter templates cr
 
 <docs-info>Note that the types referenced in `remix.env.d.ts` will depend on which environment you're running your app in. For example, there are different globals available in Cloudflare</docs-info>
 
+## Type Utilities
+
+Remix's built-in inference between `loader` function + `useLoaderData` generic and the `action` function + `useActionData` generic is quite good and you shouldn't often have to use these utilities. That said they can be quite handy in some cases which is why they are exported.
+
+### `UseDataFunctionReturn`
```
pcattori, line 83: `UseDataFunctionReturn` was renamed to `SerializedFrom` and can currently be imported from any one of the server runtime pkgs.

pcattori (changes requested):

docs/guides/streaming.md
```diff
@@ -0,0 +1,377 @@
...
+
+const ABORT_DELAY = 5000;
+
+export default function handleRequest(
+  request: Request,
+  responseStatusCode: number,
+  responseHeaders: Headers,
+  remixContext: EntryContext
+) {
+  const callbackName = isbot(
```
mjackson, line 155: Please branch the code example below. Something like:

```ts
if (isbot(...)) {
  return serveTheBots(req);
} else {
  return serveBrowsers(req);
}
```

docs/guides/streaming.md
```diff
@@ -0,0 +1,377 @@
...
+        </Await>
+      </Suspense>
+    </main>
+  );
+}
+
+function PackageLocation() {
+  const packageLocation =
+    useAsyncValue<
+      UseDataFunctionReturn<
```
mjackson, line 294: We renamed this type.

docs/guides/streaming.md
```diff
@@ -0,0 +1,377 @@
...
+      : await packageLocationPromise,
+  });
+}
+```
+
+That `shouldDeferPackageLocation` could be implemented to check the user making the request, whether the package location data is in a cache, the status of an A/B test, or whatever else you want. This is pretty sweet 🍭
+
+Also, because this happens at request time (even on client transitions), makes use of the URL via nested routing (rather than requiring you to render before you know what data to fetch), and it's all just regular HTTP, we can prefetch and cache the response! Meaning client-side transitions can be _much_ faster (in fact, there are plenty of situations when the user may never be presented with the fallback at all).
+
+Another powerful thing that's not immediately recognizable is if your server can finish loading deferred data before the client can load the javascript and hydrate, the server will stream down the HTML and "pop" it into place before react is even on the page, increasing performance for those on slow networks.
```
mjackson, line 352:
```suggestion
Another thing that's not immediately recognizable is if your server can finish loading deferred data before the client can load the javascript and hydrate, the server will stream down the HTML and add it to the document *before React hydrates*, thereby increasing performance for those on slow networks. This works even if you never add `<Scripts />` to the page thanks to React 18's support for out-of-order streaming.
```

docs/guides/streaming.md
```diff
@@ -0,0 +1,377 @@
...
+}
+```
+
+</details>
+
+## Evaluating the solution
+
+So rather than waiting for the whole `document -> JavaScript -> hydrate -> request` cycle, with streaming we start the request for the slow data as soon as the document request comes in. This can significantly speed up the user experience.
+
+Additionally, the API that Remix exposes for this is extremely ergonomic. You can literally switch between whether something is going to be deferred or not based on whether you include the `await` keyword:
```
mjackson, line 314:
```suggestion
Remix treats any `Promise` values as deferred data. These will be streamed to the client as each `Promise` resolves. If you'd like to prevent streaming for critical data, just use `await` and it will be included in the initial presentation of the document to the user.
```

docs/guides/streaming.md
```diff
@@ -0,0 +1,377 @@
...
+
+```tsx
+return defer({
+  // not deferred:
+  packageLocation: await packageLocationPromise,
+  // deferred:
+  packageLocation: packageLocationPromise,
```
mjackson, line 321:
```suggestion
return defer({
  // critical data (not deferred):
  packageLocation: await packageLocationPromise,
  // non-critical data (deferred):
  packageLocation: packageLocationPromise,
```

packages/remix-express/compression.ts
```diff
@@ -0,0 +1,36 @@
+import type { CompressionFilter } from "compression";
```
mjackson, line 1: Let's get rid of this and tests.

packages/remix-express/index.ts
```diff
@@ -1,4 +1,7 @@
 import "./globals";
 
+export type { CreateCompressionFilterOptions } from "./compression";
+export { createCompressionFilter } from "./compression";
```
mjackson, line 4: Also remove these.

packages/remix-express/package.json
```diff
@@ -17,6 +17,7 @@
     "@remix-run/node": "1.6.8"
   },
   "devDependencies": {
+    "@types/compression": "^1.7.0",
```
mjackson, line 20: Also remove this.

packages/remix-react/data.ts
```diff
@@ -54,11 +57,32 @@ export async function fetchData(
...
+    response.body &&
+    contentType &&
+    /\btext\/remix-deferred\b/.test(contentType)
+  ) {
+    let deferred = await parseDeferredReadableStream(response.body);
+    if (!resolveDeferred) {
+      return deferred;
+    }
+
+    if (await deferred.resolveData(signal)) {
```
mjackson, line 79: Would you mind adding a comment here that explains what this is doing?

mjackson (changes requested):
Partial review. Please update so we can review again tomorrow!

packages/remix-react/components.tsx
jacob-ebey: Not actually needed after double checking.

decisions/0004-deferred.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Context
+
+With React 18 landing, we are looking to take full advantage of the out-of-order streaming renderers and selective hydration without introducing a huge new API surface or requiring fundamental architectural changes for existing apps to leverage the new functionality.
+
+## Decision
+
+### Behavior
+
+The `<Suspense />` fallback prop only renders on the initial render of the `<Await />` component. It will not render the fallback if props change. Effectively, this means that you will not get a fallback rendered when a user submits a form and loader data is revalidated and you will not get a fallback rendered when the user navigates to the same route with different params.
```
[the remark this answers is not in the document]
jacob-ebey: Matt got a doc on dev already so I've removed this.

packages/remix-react/data.ts
jacob-ebey: I want to review this behavior more. It's and edge case so we might just want this to throw but I'd have to look deeper.

packages/remix-deferred/index.ts
```diff
@@ -0,0 +1,449 @@
+export const CONTENT_TYPE = "text/remix-deferred";
```
tom-sherman, line 1: I would love if `DeferredData` could be implemented in a more "webby" way, instead of relying on a custom Content-Type. Also, I totally get why you'd restrict `deferred` to JSON responses only as a first pass as this solves most use cases but it would be great if you could defer other content types eg. blobs and forms.

I think both of these needs can be met by changing `DeferredData`'s internal representation to use `Content-Type: multipart/mixed`, or maybe it should be `multipart/parallel` as deferred parts indeed do not have any particular order - they're just streamed when ready.

Because each part of a multipart response has it's own headers it allows `DeferredData` to contain many different content types, including nested `DeferredData`s. I'm not sure how the latter could be represented in JS (probably not possible with promises) but it's fun to think about.

The other benefit here is that you're not then inventing your own internal content type and you can take advantage of prior art such as https://github.com/maraisr/meros and https://github.com/tom-sherman/response-multipart. This has a benefit to consumers in that devtools can display multipart responses in a nice way which can aid debugging.

Would love to hear thoughts!

tom-sherman: Using `multipart/mixed` actually makes the response body super readable in plaintext:

```
... other headers here ...
Content-type: multipart/mixed; boundary="remix boundary" 
--remix boundary
Content-Type: application/json
Remix-Event-Key: data
Remix-Settled-Key: someData

{"count": 1}
--remix boundary
Content-Type: application/octet-stream
Remix-Event-Key: data
Remix-Settled-Key: anImage

... imagine some image data here ...

--remix boundary-- 
```

Note here I've used the custom HTTP headers `Remix-Settled-Key` and `Remix-Event-Key`, these are just examples of how a multipart body could carry the key name.

tom-sherman: Finally, just as a note, part of this request is because I feel like there is a lot of overlap with what you're doing with DeferredData and what others are doing with multipart bodies. Your `readStreamSections` implementation for example could be lifted and shifted into [my multipart library](https://github.com/tom-sherman/response-multipart) to support full streaming of each parts.

At the end of the day this is why standards and Using The Platform™️ are so great!

docs/guides/typescript.md
```diff
@@ -76,4 +76,55 @@ Remix has TypeScript type definitions built-in as well. The starter templates cr
...
+export const loader: LoaderFunction = () => {
+  return json({ cats: await getCats() });
+};
+
+export const action: ActionFunction = () => {
+  return json({ error: "Not yet implemented" });
+};
+
+export default function CatsRoute() {
+  const data = useLoaderData<typeof loader>();
```
grabbou, line 99: I did some local testing and I think the type of `data` will be `any`. Changing to `ActionArgs` and `LoaderArgs` instead of setting type for the entire function does the trick.

```
export const action = async ({ request }: ActionArgs) => {
};
```

grabbou: Just got redirected to this PR by @kentcdodds to check some examples around typings and figured I'd give some feedback! If this is WIP, please consider my comment irrelevant 👋

lili21:
can anyone share the status of the deferred API? when will this PR be merged ?

kiliman:
It will be part of "Remixing React Router" where they port Remix on top of React Router 6.4 (which was just released). They are working on the port now. I'm not sure of the timeline. If you want to try out the features, you can install the remix packages using the _deferred_ version tag.
image

https://remix.run/blog/remixing-react-router

brophdawg11:
@lili21 Yep!  I would refer to the React Router docs for now as the eventual stable Remix API should match: https://reactrouter.com/en/main/guides/deferred and https://reactrouter.com/en/main/components/await

kentcdodds:
Heya 👋 

I found a TypeScript bug. In a situation where you want to A/B test deferring some data, the render prop doesn't seem to know what to do with the type. Here's my loader:

```tsx
export async function loader({ request, params }: LoaderArgs) {
  const user = await requireUser(request);
  const { customerId } = params;
  invariant(customerId, "customerId param is required");
  const customerInfo = await getCustomerInfo(customerId);
  if (!customerInfo) {
    throw new Response("not found", { status: 404 });
  }
  const invoiceDetailsPromise = getCustomerInvoiceDetails(customerId);
  const deferInvoices = await shouldDeferInvoices({
    customerId,
    userId: user.id,
  });
  return defer({
    customerInfo,
    invoiceDetails: deferInvoices
      ? invoiceDetailsPromise
      : await invoiceDetailsPromise,
  });
}
```

And here's the `Await` usage:

```tsx
<Suspense fallback={<InvoiceDetailsFallback />}>
  <Await
    resolve={data.invoiceDetails}
    errorElement={
      <div className="relative h-full">
        <ErrorFallback />
      </div>
    }
  >
    {(invoiceDetails) => (
      <table className="w-full">
        <tbody>
          {/* @ts-expect-error This needs to be fixed in Remix */}
          {invoiceDetails.map((details) => (
            <Etc />
          ))}
        </tbody>
      </table>
    )}
  </Await>
</Suspense>
```

The TypeScript error I get on `invoiceDetails` is:

```
Property 'map' does not exist on type 'SerializeObject<{ totalAmount: number; totalDeposits: number; daysToDueDate: number; dueStatus: DueStatus; dueStatusDisplay: string; id: string; number: number; }>[] | SerializeObject<...>'.
  Property 'map' does not exist on type 'SerializeObject<{ then: <TResult1 = { totalAmount: number; totalDeposits: number; daysToDueDate: number; dueStatus: DueStatus; dueStatusDisplay: string; id: string; number: number; }[], TResult2 = never>(onfulfilled?: ((value: { totalAmount: number; ... 5 more ...; number: number; }[]) => TResult1 | PromiseLike<...>...'.
```

Interestingly, if I change my loader for `invoiceDetails` to either *only* await or *only not* await, then it works fine. It's just not jazzed about using both.

Here's the spot in my repo: https://github.com/kentcdodds/stream-away-the-wait-talk/blob/2526cafd0a758c20c3ef70ba78b008bacabd19bc/apps/11-levers/app/routes/__app/sales/customers/%24customerId.tsx#L70

Feel free to reproduce via:

```
npx create-remix ./defer-types-issue-repro --template=https://github.com/kentcdodds/stream-away-the-wait-talk/tree/main/apps/11-levers --typescript --install
cd ./defer-types-issue-repro
open "app/routes/__app/sales/customers/\$customerId.tsx"
# delete the ts-expect-error on line 70
npm run typecheck
```

lili21:
does it work inside worker runtime, like cloudflare worker?

cliffordfajardo:
I'm using the same branch/version Kent is using for the `defer` branch  `0.0.0-experimental-9b7f37c9a`
When I run my page which is using `defer`
-  the first time I render the page during development it works development it works
-  but when I refresh the page, its breaks and I get a message like:

```ts
Error: The render was aborted by the server without a reason.
    at abortTask (/Users/cfajardo/Repos/remix-ssr-vs-ssr-streaming/node_modules/react-dom/cjs/react-dom-server.node.development.js:6436:43)
    at /Users/cfajardo/Repos/remix-ssr-vs-ssr-streaming/node_modules/react-dom/cjs/react-dom-server.node.development.js:7002:14
    at Set.forEach (<anonymous>)
    at abort (/Users/cfajardo/Repos/remix-ssr-vs-ssr-streaming/node_modules/react-dom/cjs/react-dom-server.node.development.js:7001:20)
    at Timeout.abort [as _onTimeout] (/Users/cfajardo/Repos/remix-ssr-vs-ssr-streaming/node_modules/react-dom/cjs/react-dom-server.node.development.js:7051:7)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
```


This is my **package.json** file - and here is a stackblitz reproducing the error: https://stackblitz.com/edit/node-whsnyv

Video reproducing the error in the slackblitz (same behavior happens on my Mac)
https://user-images.githubusercontent.com/6743796/200146863-71ac5799-eb70-44c2-8137-5ff6f4db2d77.mp4


### Steps to Reproduce
1. Load route which is using `defer`
2. Refresh the route which is using `defer` and you'll see the 2nd time it no longer works

lili21:
> does it work inside worker runtime, like cloudflare worker?

```bash
✘ [ERROR] Could not resolve "fs/promises"
```

when I'm trying to use `defer` in the cloudflare worker template.

lili21:
@cliffordfajardo try decrease the `getPeopleDelayTime` value. or increase the `ABORT_DELAY` value inside `entry.server.tsx`.

lili21:
if the deferred promise reject, it will throw the `unhandlePromiseRejection` error, which will break the application by defaut.

kentcdodds:
It was a good run 😅

For anyone looking for the new PR: https://github.com/remix-run/react-router/pull/9760

PR #3436, opened 2022-06-10
fix(remix-server-runtime): use correct error/catch boundary on SSR action errors

brophdawg11:
Closes: #599

- [x] Tests

Testing Strategy: I tested using the repo in #599 but the repro setup is basically:
* Parent route with loader data
* Child route that exports `ErrorBoundary` or `CatchBoundary`
* Child contains `form` or `<Form reloadDocument={true}>` submitting to child action that throws
  * This error only happens in server-side form submissions 
* Proper boundary should be used and the parent loaderData should not be wiped away on submission

### DIFF
--- a/packages/remix-server-runtime/__tests__/server-test.ts
+++ b/packages/remix-server-runtime/__tests__/server-test.ts
@@ -880,7 +880,8 @@ describe("shared server runtime", () => {
       let result = await handler(request);
       expect(result.status).toBe(400);
       expect(testAction.mock.calls.length).toBe(1);
-      expect(rootLoader.mock.calls.length).toBe(1);
+      // Should not call root loader since it is the boundary route
+      expect(rootLoader.mock.calls.length).toBe(0);
       expect(testLoader.mock.calls.length).toBe(0);
       expect(build.entry.module.default.mock.calls.length).toBe(1);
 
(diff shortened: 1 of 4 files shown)


packages/remix-server-runtime/server.ts
```diff
@@ -273,24 +273,21 @@ async function handleDocumentRequest({
...
     matchesToLoad = getMatchesUpToDeepestBoundary(
-      // get rid of the action, we don't want to call it's loader either
-      // because we'll be rendering the catch boundary, if you can get access
-      // to the loader data in the catch boundary then how the heck is it
-      // supposed to deal with thrown responses?
-      matchesToLoad.slice(0, -1),
+      matchesToLoad,
       "CatchBoundary"
-    );
+    ).slice(0, -1);
```
brophdawg11, line 285: We don't want to `slice` _and then_ look for the boundary, otherwise we miss leaf boundaries.  Instead, find the boundary then `slice` so we don't run the loader for the boundary route.

packages/remix-server-runtime/__tests__/server-test.ts
```diff
@@ -880,7 +880,8 @@ describe("shared server runtime", () => {
       let result = await handler(request);
       expect(result.status).toBe(400);
       expect(testAction.mock.calls.length).toBe(1);
-      expect(rootLoader.mock.calls.length).toBe(1);
+      // Should not call root loader since it is the boundary route
+      expect(rootLoader.mock.calls.length).toBe(0);
```
brophdawg11, line 884: These unit tests were previously invalid - if we throw from the child and it gets handled by the root - we do not want to re-run the root loader

MichaelDeBoey (approved):

PR #3438, opened 2022-06-10
feat(templates/deno): add triple-slash reference for client code

justjavac:
- `/// <reference no-default-lib="true" />` can prevent developers from writing Deno code on the client side.

- `<reference lib="dom" />` can fix using document in Deno code. Without this line of code, vscode produces a warning:

  > Cannot find name 'document'. Do you need to change your target library? Try changing the 'lib' compiler option to include 'dom'. **deno-ts(2584)**


templates/deno/app/entry.client.tsx
```diff
@@ -1,3 +1,6 @@
+/// <reference no-default-lib="true" />
+/// <reference lib="dom" />
```
MichaelDeBoey, line 2: I think this should be in a `tsconfig.json` instead? 🤔

CC/ @pcattori

MichaelDeBoey: Other templates have it in its `tsconfig.json`.

I know @pcattori removed the file on purpose, but I can't remember the reasoning behind it.

MichaelDeBoey:
@justjavac Template changes go against `main`.
Could you please rebase your branch onto latest `main` & set base branch to `main` as wel?

justjavac:
@MichaelDeBoey I opened a new pr #3449

PR #3443, opened 2022-06-10
ci: add script to convert links in docs to references

mcansh:

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


.github/workflows/format.yml
```diff
@@ -34,6 +34,9 @@ jobs:
       - name: 👔 Format
         run: npm run format --if-present
 
+      - name: 🔗 Convert Docs links to references
```
MichaelDeBoey, line 37: I think we should do this before the `format` step, as all JS/TS in our MD will be in the wrong format otherwise

MichaelDeBoey (changes requested):

.github/workflows/format.yml
MichaelDeBoey: That's correct, but it's using `markdown` for everything it seems.
Even if we have code examples in there.

Don't know how to fix it in another way though. 🤔

MichaelDeBoey:
@mcansh Probably a good idea to re-run `format` manually so we can see everything stays as it should be?

scripts/markdown-references.mjs
```diff
@@ -0,0 +1,29 @@
+import path from "node:path";
+import fsp from "node:fs/promises";
```
MichaelDeBoey, line 2: These imports could be simplified I guess?

```suggestion
import path from "path";
import fsp from "fs/promises";
```

MichaelDeBoey (changes requested):

docs/api/conventions.md
```diff
@@ -1307,27 +1252,23 @@ Another common case is when you've got nested routes and a child component has a
 
 Consider these routes:
 
-```
```
MichaelDeBoey, line 1310: @mcansh I think your plugin needs some work (probably a good idea to add tests for these), as it appears that's the one removing code block start/end (` ``` `).

MichaelDeBoey (changes requested):

docs/api/conventions.md
mcansh: looks like we lost the code fencing, but markdown still renders it the same

MichaelDeBoey: Fencing gets removed & 4 spaces are put into place, which should indeed give the same rendered result normally.

However, I think having the fences makes it more clear though.

MichaelDeBoey: Thinking about this, this could be a (semi) `prettier` problem as well, as code blocks aren't formatted consistently, but I don't know what the expected behavior is for them.

mcansh: looks like fences are an option of remark - adding them back

docs/api/conventions.md
```diff
@@ -1389,9 +1383,9 @@ You need to be very careful here, though. That project (or its nested relationsh
 
 Any files inside the `app` folder can be imported into your modules. Remix will:
 
-1. Copy the file to your browser build directory
-2. Fingerprint the file for long-term caching
-3. Return the public URL to your module to be used while rendering
+1.  Copy the file to your browser build directory
+2.  Fingerprint the file for long-term caching
+3.  Return the public URL to your module to be used while rendering
```
MichaelDeBoey, line 1394: This also seems something that shouldn't happen 🤔

```suggestion
1. Copy the file to your browser build directory
2. Fingerprint the file for long-term caching
3. Return the public URL to your module to be used while rendering
```

docs/guides/data-loading.md
```diff
@@ -357,15 +357,15 @@ export default function ProductFilters() {
 
 If the user only has one selected:
 
-- [x] Nike
-- [ ] Adidas
+- \[x] Nike
+- \[ ] Adidas
```
MichaelDeBoey, line 361: Same here 🤔

MichaelDeBoey (changes requested):

scripts/markdown-references.mjs
mcansh: personally, im fine with it, indicates that these are built ins

docs/guides/data-loading.md
MichaelDeBoey: It seems like this is a `remark` problem.

@wooorm Is this expected default behavior? If so, How could we make sure the backslash (`\`) is removed?

wooorm: Us remark-gfm to enable gfm in markdown

MichaelDeBoey (approved):

kentcdodds (approved):
Solid 👍 Thanks!

docs/api/conventions.md
```diff
@@ -191,9 +191,9 @@ app/
 
 The file in `app/root.tsx` is your root layout, or "root route" (very sorry for those of you who pronounce those words the same way!). It works just like all other routes:
 
-- You can export a [`loader`](#loader), [`action`](#action), [`meta`](#meta), [`headers`](#headers), or [`links`](#links) function
-- You can export an [`ErrorBoundary`](#errorboundary) or [`CatchBoundary`](#catchboundary)
-- Your default export is the layout component that renders the rest of your app in an [`<Outlet />`](https://reactrouter.com/docs/en/v6/api#outlet)
+- You can export a [`loader`][3], [`action`][4], [`meta`][5], [`headers`][6], or [`links`][7] function
```
kentcdodds, line 194: Is there any way your package could try to use the content of the thing that's being linked for the name rather than a number? So this would be:

```suggestion
- You can export a [`loader`][loader], [`action`][action], [`meta`][meta], [`headers`][headers], or [`links`][links] function
```

It would need to strip characters that don't work in a link (like back-ticks and `<`s I think) and probably truncate at some max-length. And you'd want to handle duplicates with post-fixed numbers.

docs/api/conventions.md
```diff
@@ -191,9 +191,9 @@ app/
...
+- You can export an [`ErrorBoundary`][8] or [`CatchBoundary`][9]
+- Your default export is the layout component that renders the rest of your app in an [`<Outlet />`][reactrouter-1]
```
kentcdodds, line 196: Maybe we could combine some of the logic of the anchor tags I describe in the previous comment with the domain logic of external links. So this could be:

```suggestion
- Your default export is the layout component that renders the rest of your app in an [`<Outlet />`][reactrouter-outlet]
```

Then maybe we could reduce the number of times we need a post-fix number.

Also, on that note, let's not add a post-fix for the first link. So if there's a duplicate we'd only get a `-2` but we wouldn't get a `-1`.

docs/api/conventions.md
```diff
@@ -285,7 +285,7 @@ export default function PostRoute() {
 
 Nested routes can also contain dynamic segments by using the `$` character in the parent's directory name. For example, `app/routes/blog/$postId/edit.tsx` might represent the editor page for blog entries.
 
-See the [routing guide](../guides/routing.md) for more information.
+See the [routing guide][11] for more information.
```
kentcdodds, line 288: For internal links I think we can do similar treatment that I described for the anchor links above. Use the content of the link. You may be able to just use something like https://www.npmjs.com/package/@sindresorhus/slugify for this.

kentcdodds (changes requested):
Thanks! The script itself looks great. I have some ideas for the package you developed for this. Let me know if you've got questions. Thanks!

mcansh:
> Thanks! The script itself looks great. I have some ideas for the package you developed for this. Let me know if you've got questions. Thanks!

great feedback, i was thinking about this a bit, but never looked into implementing it, i'll take a stab at it, would be slick for sure

package.json
```diff
@@ -91,15 +94,18 @@
...
     "rollup": "^2.36.1",
     "rollup-plugin-copy": "^3.3.0",
     "semver": "^7.3.4",
     "simple-git": "^3.2.4",
     "sort-package-json": "^1.54.0",
     "strip-indent": "^3.0.0",
+    "to-vfile": "7.2.3",
     "type-fest": "^2.11.1",
-    "typescript": "^4.5.5",
-    "unified": "^9.2.0"
```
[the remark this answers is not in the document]
mcansh: mistakenly removed tbh, i added it myself and then removed it when i switched 🤷 - i can add it back to be safe

mcansh:
sweet, i think i have it mostly there @kentcdodds, if you could check https://github.com/mcansh/remark-definition-links/blob/main/__tests__/fixtures/conventions-after.md?plain=1#L1368-L1412 that'd be stellar!

docs/guides/data-writes.md
```diff
@@ -516,9 +516,17 @@ From your components perspective, all that happened was the `useTransition` hook
...
+- [useTransition][use-transition]
+- [Actions][actions]
+- [Loaders][loaders]
+- [`useSubmit()`][use-submit]
+- [`useFetcher()`][use-fetcher]
+
+[form]: ../api/remix#form
+[use-submit]: ../api/remix#usesubmit
+[use-fetcher]: ../api/remix#usefetcher
+[form-2]: ../api/remix#form
```
MichaelDeBoey, line 529: This is the same as the first link.

Probably a bit hard, but could we maybe keep it at 1 reference?

MichaelDeBoey (approved):
Looks awesome! 🔥

package.json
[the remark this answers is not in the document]
MichaelDeBoey: Don't think it's necessary, as we're not using it directly + CI is green without it

docs/guides/data-writes.md
mcansh: definitely should be one, checking 🙂

docs/guides/migrating-react-router-app.md
```diff
@@ -174,9 +174,9 @@ Lastly, in your root `App` component (the one that would have been mounted to th
 
 ## Creating a catch-all route
 
-Remix needs routes beyond the root route to know what to render in `<Outlet />`. Fortunately you already render `<Route>` components in your app, and Remix can use those as you migrate to use our [routing conventions](./routing.md).
+Remix needs routes beyond the root route to know what to render in `<Outlet />`. Fortunately you already render `<Route>` components in your app, and Remix can use those as you migrate to use our [routing conventions][routing-conventions].
 
-To start, create a new directory in `app` called `routes`. In that directory, create a file called `$.jsx`. This is called [a **catch-all route**](./routing#splats) and it will be useful to let your old app handle routes that you haven't moved into the `routes` directory yet.
+To start, create a new directory in `app` called `routes`. In that directory, create a file called `$.jsx`. This is called [a **catch-all route**][a] and it will be useful to let your old app handle routes that you haven't moved into the `routes` directory yet.
```
kentcdodds, line 179: Probably should handle this better.

kentcdodds (commented):
Yeah, this is really close. Just one thing.

docs/guides/migrating-react-router-app.md
mcansh: ah yup, i can broaden my "aggregate" function to accept all sorts of things https://github.com/mcansh/remark-definition-links/blob/main/src/index.ts#L8

mcansh: ah yup, i can broaden my "aggregate" function to accept all sorts of things https://github.com/mcansh/remark-definition-links/blob/main/src/index.ts#L8

docs/tutorials/jokes.md
```diff
@@ -6351,10 +6351,118 @@ This will build the docker image and deploy it on Fly in the region you selected
...
+[a-greeting-from-the-index-route]: /jokes-tutorial/img/index-route-greeting.png
+[jokes]: http://localhost:3000/jokes
+[a-random-joke-on-the-jokes-page-i-was-wondering-why-the-frisbee-was-getting-bigger-then-it-hit-me]: /jokes-tutorial/img/random-joke.png
+[jokes-new]: http://localhost:3000/jokes/new
+[a-new-joke-form]: /jokes-tutorial/img/new-joke.png
+[jokes-anything-you-want]: http://localhost:3000/jokes/hippos
+[a-new-joke-form-2]: /jokes-tutorial/img/param-route.png
+[links]: ../api/conventions#links
+[links-2]: ../api/remix#link
+[http-localhost-3000-2]: http://localhost:3000/
```
MichaelDeBoey, line 6391: This is the same as `http-localhost-3000`, except it has a trailing slash

docs/tutorials/jokes.md
```diff
@@ -6351,10 +6351,118 @@ This will build the docker image and deploy it on Fly in the region you selected
...
+[supabase]: https://supabase.com/
+[airtable]: https://www.airtable.com/
+[hasura]: https://hasura.io/
+[google-spreadsheets]: https://www.google.com/sheets/about/
+[cloudflare-workers-kv]: https://www.cloudflare.com/products/workers-kv/
+[fauna]: https://fauna.com/features
+[postgre-sql]: https://www.postgresql.org/
+[a-vs-code-extension]: https://marketplace.visualstudio.com/items?itemName=Prisma.prisma
+[sq-lite]: https://sqlite.org/index.html
+[prisma-2]: https://www.prisma.io
```
MichaelDeBoey, line 6407: Same for `prisma`

docs/tutorials/jokes.md
mcansh: which makes it different 🥴

docs/tutorials/jokes.md
mcansh: we're now removing trailing slashes in the plugin 😄

kentcdodds:
Ok, we've decided we're not going to do the separate package thing for now. I'll give this a look today.

kentcdodds:
Whoops. Wrong PR 😅 I'll still get this merged soon.

kentcdodds (approved):
Looks good to me! Thanks!

kentcdodds:
I wonder if we should get this in `main` as well?

mcansh:
> I wonder if we should get this in `main` as well?

ah yup! i always know what base branch to pick, and always end up picking the wrong one 😑

mcansh:
back at my computer, here it is! https://github.com/remix-run/remix/pull/3525

PR #3444, opened 2022-06-10
chore(remix-dev): always minify

mcansh:
some folks are running into issues in dev that are solved by minification. 
also esbuild is faster when minifying

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -371,7 +371,7 @@ async function createBrowserBuild(
     incremental: options.incremental,
     mainFields: ["browser", "module", "main"],
     treeShaking: true,
-    minify: options.mode === BuildMode.Production,
+    minify: true,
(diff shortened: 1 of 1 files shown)


jacob-ebey (approved):

MichaelDeBoey:
Won't this make debugging harder in `dev` mode? 🤔

sergiodxa:
> Won't this make debugging harder in `dev` mode as we now don't have actual function names, ... anymore? 🤔

If it generates a source map it shouldn't be an issue

MichaelDeBoey:
> If it generates a source map it shouldn't be an issue

That's another option, but it isn't enabled by default if I'm correct.

MichaelDeBoey (approved):

mjackson:
Let's not do this. There should be no need to minify in development. If there are issues that are solved with minification, we should be able to address those separately.

PR #3445, opened 2022-06-10
enhancement(cli): Ask about TypeScript before `npm install`

chaance:
This PR improves the CLI by asking about using TypeScript or JavaScropt *before* running `npm install`. This is a better UX because:

- We ask all questions up front, so it just flows a bit better
- If we convert the template first, `npm install` won't include all of the TS-related modules previously in the `package.json`

---

- [ ] Docs
- [x] Tests


MichaelDeBoey (approved):
LGTM!

Be sure to remove unused imports from `packages/remix-dev/cli/run.ts` though 😉

MichaelDeBoey (approved):

mcansh (approved):

PR #3447, opened 2022-06-10
enhancement(dev/cli): Add `--port` flag to `remix dev`

chaance:
This PR adds a `--port` flag to `remix dev` that takes precedence over `process.env.PORT`. This makes it easier to choose a new port on the fly.

Note, I updated the snapshot but didn't create any new tests yet. We don't currently have a test suite for `remix dev` commands and I need to follow up internally before adding a new process-intensive test that could further slow things down in CI. I did test it manually and things work as expected.

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -234,7 +234,11 @@ export async function watch(
   });
 }
 
-export async function dev(remixRoot: string, modeArg?: string) {
+export async function dev(
+  remixRoot: string,
+  modeArg?: string,
+  portArg?: number
+) {
   let createApp: typeof createAppType;
   let express: typeof Express;
   try {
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (approved):

packages/remix-dev/cli/commands.ts
```diff
@@ -234,7 +234,11 @@ export async function watch(
   });
 }
 
-export async function dev(remixRoot: string, modeArg?: string) {
+export async function dev(
+  remixRoot: string,
+  modeArg?: string,
+  portArg?: number
+) {
```
kentcdodds, line 241: If we need to add one more arg to this in the future let's turn this into a single `options` argument.

kentcdodds (commented):
This is great.

PR #3448, opened 2022-06-11
refactor: Move migration script to `@remix-run/scripts`

chaance:
This PR creates a new package `@remix-run/scripts` for useful tools like the `migrate` script currently in `@remix-run/dev`. Motivation for this change documented in https://github.com/remix-run/remix/pull/3150#issuecomment-1148907520.


packages/remix-scripts/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "directory": "packages/remix-scripts"
+  },
+  "license": "MIT",
+  "bin": {
+    "remix": "cli.js"
+  },
+  "dependencies": {
+    "@esbuild-plugins/node-modules-polyfill": "^0.1.4",
+    "@npmcli/package-json": "^2.0.0",
+    "@remix-run/dev": "1.5.1",
```
MichaelDeBoey, line 21: How would you imagine the new `convert-to-javascript` migration to be run from within `@remix-run/dev` if people choose to do so with the CLI prompt?

This would cause a circular dependency if we keep it this way.

packages/remix-scripts/cli/run.ts
```diff
@@ -0,0 +1,85 @@
...
+  --force             Bypass Git safety checks and forcibly run migration
+  --migration, -m     Name of the migration to run
+
+${colors.heading("Values")}:
+  - ${colors.arg(
+    "migration"
+  )}         One of the choices from https://github.com/remix-run/remix/tree/dev/packages/remix-scripts/cli/migrate/migration-options
+`;
+
+const MINIUMUM_NODE_VERSION = 14;
```
MichaelDeBoey, line 33: Typo

```suggestion
const MINIMUM_NODE_VERSION = 14;
```

packages/remix-scripts/cli/run.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Programmatic interface for running the Remix CLI with the given command line
+ * arguments.
+ */
+export async function run(argv: string[] = process.argv.slice(2)) {
+  // Check the node version
+  let versions = process.versions;
+  if (
+    versions &&
+    versions.node &&
+    semver.major(versions.node) < MINIUMUM_NODE_VERSION
```
MichaelDeBoey, line 45:
```suggestion
    semver.major(versions.node) < MINIMUM_NODE_VERSION
```

packages/remix-scripts/cli/run.ts
```diff
@@ -0,0 +1,85 @@
...
+  ) {
+    throw new Error(
+      `️🚨 Oops, Node v${versions.node} detected. Remix requires a Node version greater than ${MINIUMUM_NODE_VERSION}.`
```
MichaelDeBoey, line 48:
```suggestion
      `️🚨 Oops, Node v${versions.node} detected. Remix requires a Node version greater than ${MINIMUM_NODE_VERSION}.`
```

MichaelDeBoey (changes requested):

packages/remix-scripts/package.json
chaance: It would no longer be available under `@remix-run/dev`. The idea is to get all of this out of that package, and the script would change as a result (`remix-scripts migrate`). The `convert-to-javascript` stuff is outside the scope of `remix migrate`. Right now the only command under `migrate` is `https://github.com/remix-run/remix/blob/main/packages/remix-dev/cli/migrate/migrations/index.ts`. https://github.com/remix-run/remix/blob/main/packages/remix-dev/cli/migrate/migrations/index.ts

MichaelDeBoey: At least for backwards compatibility, we should keep `remix migrate` imo.

This can then run `remix-scripts migrate` under the hood & even first show a warning message, but we can't just remove it I think.

MichaelDeBoey: The idea behing the new `convert-to-javascript` migration/script was also to move it to this new `@remix-run/scripts` package btw.

That's why @kentcdodds was suggesting `scripts` as package name, as it covers both `migrations` and other scripts.

Having it in `@remix-run/dev` would mean we would copy/paste a lot of the migration code we now have in this new package too.

chaance: Sorry, what I meant is that we aren't touching the existing JS conversion script that's already in the `dev` package *in this PR*. Step one, I believe, is just to move the `migrate` script out of the `dev` package. This lets us shed some of the weight of dependencies of a core module and keeping utility scripts that won't be used very often in a separate package.

I see the `convert-to-javascript` script you started as additive rather than replacing the current behavior in `dev`. It should go in this new package, but that doesn't mean we need to change the current behavior in `dev` at the moment.  Ultimately I think @kentcdodds will be the final word on this, but that's my take.

In the long run I think it might make sense to move other scripts to this package. It's a little tricky right now since the build/dev scripts are sort of intertwined with our compiler logic, so I think that'll need to be cleaned up first. But alternatively, those scripts are necessary for remix development, whereas other scripts are less used utilities. I don't think the scope for each package is clearly defined yet, hence why I'm just starting with the one script.

As for back-compat, I don't think that's a huge concern for the migrate script. Again, Kent will say for sure. He's out next week so we likely won't merge this until he's back anyway.

MichaelDeBoey: I'm fine with whatever is decided, just wanted to state my concerns/remarks & make sure we take everything into account 👍

MichaelDeBoey (approved):

PR #3449, opened 2022-06-11
chore(templates/deno): update types config

justjavac:
via https://github.com/remix-run/remix/pull/3438#issuecomment-1152518309

### DIFF
--- a/templates/deno/.vscode/resolve_npm_imports.json
+++ b/templates/deno/.vscode/resolve_npm_imports.json
@@ -8,8 +8,10 @@
     "@remix-run/deno": "../node_modules/@remix-run/deno/index.ts",
     "@remix-run/dev/server-build": "https://esm.sh/@remix-run/dev@1.6.0/server-build",
     "@remix-run/react": "https://esm.sh/@remix-run/react@1.6.0",
+    "@remix-run/server-runtime": "https://esm.sh/@remix-run/server-runtime@1.6.0",
     "react": "https://esm.sh/react@17.0.2",
     "react-dom": "https://esm.sh/react-dom@17.0.2",
-    "react-dom/server": "https://esm.sh/react-dom@17.0.2/server"
+    "react-dom/server": "https://esm.sh/react-dom@17.0.2/server",
+    "mime": "https://esm.sh/mime@3.0.0"
   }
 }
--- a/templates/deno/app/entry.client.tsx
+++ b/templates/deno/app/entry.client.tsx
@@ -1,3 +1,6 @@
+/// <reference no-default-lib="true" />
+/// <reference lib="dom" />
+
 import { RemixBrowser } from "@remix-run/react";
 import * as React from "react";
 import { hydrate } from "react-dom";
--- a/templates/deno/app/entry.client.tsx
+++ b/templates/deno/app/entry.client.tsx
@@ -1,6 +1,3 @@
-/// <reference no-default-lib="true" />
-/// <reference lib="dom" />
-
 import { RemixBrowser } from "@remix-run/react";
 import * as React from "react";
 import { hydrate } from "react-dom";
--- a/templates/deno/package.json
+++ b/templates/deno/package.json
@@ -5,11 +5,11 @@
     "build": "remix build",
     "deploy": "deployctl deploy --prod --include=build,public --project=<your deno deploy project> ./build/index.js",
     "dev": "remix build && run-p dev:*",
-    "dev:deno": "cross-env NODE_ENV=development deno run --unstable --watch --allow-net --allow-read --allow-env ./build/index.js",
+    "dev:deno": "cross-env NODE_ENV=development deno run --unstable --watch --allow-net --allow-read --allow-env --no-check=remote ./build/index.js",
     "dev:remix": "remix watch",
     "format": "deno fmt --ignore=node_modules,build,public/build",
     "lint": "deno lint --ignore=node_modules,build,public/build",
-    "start": "cross-env NODE_ENV=production deno run --unstable --allow-net --allow-read --allow-env ./build/index.js"
+    "start": "cross-env NODE_ENV=production deno run --unstable --allow-net --allow-read --allow-env --no-check=remote ./build/index.js"
   },
   "dependencies": {
     "@remix-run/deno": "*",
--- a/templates/deno/package.json
+++ b/templates/deno/package.json
@@ -10,7 +10,7 @@
     "format": "deno fmt --ignore=node_modules,build,public/build",
     "lint": "deno lint --ignore=node_modules,build,public/build",
     "start": "cross-env NODE_ENV=production deno run --unstable --allow-net --allow-read --allow-env --no-check ./build/index.js",
-    "typecheck": "deno check --import-map .vscode/resolve_npm_imports.json server.ts app/*.tsx app/*/*.tsx"
+    "typecheck": "deno check --import-map .vscode/resolve_npm_imports.json server.ts app/*.ts* app/*/*.ts*"
   },
   "dependencies": {
     "@remix-run/deno": "*",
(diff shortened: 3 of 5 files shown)


templates/deno/app/entry.client.tsx
```diff
@@ -1,3 +1,6 @@
+/// <reference no-default-lib="true" />
+/// <reference lib="dom" />
```
MichaelDeBoey, line 2: I think this should be in a `tsconfig.json` instead? 🤔

CC/ @pcattori

pcattori: Deno doesn't use `tsconfig.json`, but we could instead include this in a [`deno.json` config file](https://deno.land/manual/getting_started/configuration_file).

pcattori:
@justjavac : we were previously using the `lib` option with the Deno config file, but that was [causing some errors](https://github.com/remix-run/remix/issues/2402). Maybe it was related to the Deno version, but I'm not really sure. 

@CanRau : do you have any insights into why the errors from #2402 were happening in the first place and if re-adding these references will cause issues?

I personally have experience both cases locally: (1) getting `AbortSignal` errors when using references/`lib` option in Deno config file and (2) getting errors about `document` type when not using references/`lib` option.

justjavac:
@pcattori Deno implements `AbortSignal.reason` ahead of typescript. Deno defines `AbortSignal.reason` in its `.d.ts` file. When configuring `lib: ["dom", "deno.ns"]`, typescript's `.d.ts` will overwrite deno's.

In Deno v1.21.2, it adds a `lib.dom.extras.d.ts` file and modifies the `lib.dom.d.ts` file to include it. https://github.com/denoland/deno/pull/14430

Now we can use `deno.json`:

```json
{
  "compilerOptions": {
    "lib": ["dom", "dom.iterable", "dom.asynciterable", "deno.ns"]
  }
}
```

For previous versions, you can add `--no-check=remote` flag. (added in deno [v1.17.0](https://github.com/denoland/deno/releases/tag/v1.17.0))

justjavac:
I create a repo. testing in v1.17.0, 1.18.0, 1.19.0, 1.20.1, 1.21.0, 1.22.0, latest and canary https://github.com/justjavac/remix-deno-demo/actions/runs/2480815351

-------------

1. `npx create-remix@latest --template deno`

1. `npm i https://gitpkg.now.sh/justjavac/remix/packages/remix-deno?deno-std-143`: in this branch, I upgrade deno `std@0.143.0`

1. create a `deno.json` file

1. add `--no-check=remote` flag in package.json

MichaelDeBoey:
Maybe also a good idea to add minimum deno version to `engines` in `package.json`?

justjavac:
If we use the `--no-check=remote` flag, we can set the minimum version to `v1.17.0`. If we use the `--no-check` flag, we don't need to set the minimum version, which is what deployctl does. deployctl validates 3 versions in GitHub action: old(1.16.0), stable(v1.x), canary.

https://github.com/denoland/deployctl/blob/main/.github/workflows/ci.yml#L24-L28

Under the hood, `deno run` uses tsc, and `deno run --no-check` uses swc. In the future `--no-check` will become the default behavior of `deno run`, and provide a separate `deno check` command for type checking.

I think we can use `--no-check` to run deno code because vscode-deno extension already checks for type issues during development. Developers can also use `deno check`(added in v1.21.0) to manually check types, or use `deno cache`(ddded in v0.x) to check types, which is what deployctl does.

pcattori:
@justjavac I like the idea of using `--no-check` so that we have compatibility across Deno versions and rely on `deno check` instead. Ideally, I'd like to add a `typecheck` script to `package.json` that just calls `deno check` (with any relevant args/flags). My only concern is that users of Deno <1.21 won't have a programmatic way to typecheck. I think this a concern both for users who might not be using `vscde-deno` (or not using VS Code at all) OR for users who need a programmatic typecheck e.g. for CI.

That said, I'm still leaning towards using `--no-check` and an explicit `"typecheck": "deno check"` script. We just wouldn't support programmatic typechecking for users on Deno <1.21 🤷 if they want typechecking they'll just need to upgrade Deno.

pcattori:
@justjavac I like the idea of using `--no-check` so that we have compatibility across Deno versions and rely on `deno check` instead. Ideally, I'd like to add a `typecheck` script to `package.json` that just calls `deno check` (with any relevant args/flags). My only concern is that users of Deno <1.21 won't have a programmatic way to typecheck. I think this a concern both for users who might not be using `vscde-deno` (or not using VS Code at all) OR for users who need a programmatic typecheck e.g. for CI.

That said, I'm still leaning towards using `--no-check` and an explicit `"typecheck": "deno check"` script. We just wouldn't support programmatic typechecking for users on Deno <1.21 🤷 if they want typechecking they'll just need to upgrade Deno.

templates/deno/.vscode/resolve_npm_imports.json
```diff
@@ -5,11 +5,13 @@
...
-    "@remix-run/deno": "https://esm.sh/@remix-run/deno@1.6.0",
+    "@remix-run/deno": "https://esm.sh/@remix-run/deno@1.6.0/index.ts",
+    "@remix-run/server-runtime": "https://esm.sh/@remix-run/server-runtime@1.6.0",
     "@remix-run/dev/server-build": "https://esm.sh/@remix-run/dev@1.6.0/server-build",
     "@remix-run/react": "https://esm.sh/@remix-run/react@1.6.0",
     "react": "https://esm.sh/react@17.0.2",
     "react-dom": "https://esm.sh/react-dom@17.0.2",
-    "react-dom/server": "https://esm.sh/react-dom@17.0.2/server"
+    "react-dom/server": "https://esm.sh/react-dom@17.0.2/server",
+    "mime": "https://esm.sh/mime@3.0.0"
```
MichaelDeBoey, line 15: Let's keep these alphabetical, so it's easier to find the correct ones if we ever have a long list here

MichaelDeBoey (changes requested):

templates/deno/package.json
```diff
@@ -5,11 +5,12 @@
...
     "deploy": "deployctl deploy --prod --include=build,public --project=<your deno deploy project> ./build/index.js",
     "dev": "remix build && run-p dev:*",
-    "dev:deno": "cross-env NODE_ENV=development deno run --unstable --watch --allow-net --allow-read --allow-env ./build/index.js",
+    "dev:deno": "cross-env NODE_ENV=development deno run --unstable --watch --allow-net --allow-read --allow-env --no-check ./build/index.js",
     "dev:remix": "remix watch",
     "format": "deno fmt --ignore=node_modules,build,public/build",
     "lint": "deno lint --ignore=node_modules,build,public/build",
-    "start": "cross-env NODE_ENV=production deno run --unstable --allow-net --allow-read --allow-env ./build/index.js"
+    "start": "cross-env NODE_ENV=production deno run --unstable --allow-net --allow-read --allow-env --no-check ./build/index.js",
+    "typecheck": "deno check --import-map .vscode/resolve_npm_imports.json server.ts app/*.tsx app/*/*.tsx"
```
pcattori, line 13: Does `deno check` support globs like `**/*.tsx`?

Also, we need to check `.ts` files too, not just `.tsx` files, so maybe we'd need something like:

```suggestion
    "typecheck": "deno check --import-map .vscode/resolve_npm_imports.json server.ts app/**/*.ts app/**/*.tsx"
```

CanRau:
@pcattori sorry for late response also no idea 😬

pcattori:
I'm in favor of using a dedicated `typecheck` script and using `--no-check` for all other Deno commands, but right now Deno does not support globs for its commands.

Closing this PR as we don't want to stop typechecking deeply nested source files. If/when Deno supports globs, we can reopen.

Issue #3302, opened 2022-05-25
Redirects cause root route loader invocation but root route component does not receive loader data


jangerhofer:
### What version of Remix are you using?

1.5.1

### Steps to Reproduce

A repo with a minimal reproduction is available here: https://github.com/jangerhofer/remixRootRouteRedirectReproduction along with some directions.

### Expected Behavior

I expect that a server-side redirect (whether performance only once or multiple times) would cause a fresh root route `loader` function invocation and that the (serialized) value returned from the `loader` would be made available to the root route component via the `useLoaderData` hook.

### Actual Behavior

Instead, I see only the invocation the root route `loader`; the `useLoaderData` hook's value remains an empty object despite the `loader` returning data.

amanolar94:
@jangerhofer did you maybe have any luck with this? running onto the same problem as well...

jangerhofer:
@amanolar94: I did not, unfortunately.

Instead, I moved around my route structure to avoid having to perform two redirects. I used a [pathless layout route](https://remix.run/docs/en/v1/guides/routing#pathless-layout-routes) in the process.

amanolar94:
I see..In our case pathless layout route won't do since it is a [dynamic segment](https://remix.run/docs/en/v1/guides/routing#dynamic-segments).

jamesilva:
@jangerhofer I was having a similar bug, but in the opposite way, where a user that no longer was logged in was still showing up in the root navbar. Then I started digging and found this  post, and in your repo you say that it works if

> "the login form redirects to /, but the / index route's loader does not redirect but rather renders a component itself"

so I tried your code and only changed a few things in the index.tsx, leaving the routing to the component:
```
export const loader: LoaderFunction = async ({ request }) => {
  const cookieHeader = request.headers.get('Cookie');
  const { email } = (await session.parse(cookieHeader)) || {};

  return { email };
};

export default function Index() {
  let { email } = useLoaderData();
  let navigate = useNavigate();

  useEffect(() => {
    if (email) {
      navigate('/content');
    } else {
      navigate('login');
    }
  }, [email]);
  return <></>;
}
```
which worked as intended. 
So, after a bit more digging, I finally got to the remix page about resource routes [https://remix.run/docs/en/v1/guides/resource-routes](url), which says:

[...]

PR #3460, opened 2022-06-13
fix: Handle revalidations across multi-redirects

brophdawg11:
There is some stateful information driving whether or not we revalidate inside transition manager which was getting lost across a redirect boundaries in certain scenarios.  This PR fixes two potential flows:

1. action submission redirects, then loader redirects
   1. We go `actionRedirect -> normalRedirect` and the second loader execution skips reused parent routes
   2. This is fixed by keeping subsequent redirects marked as `actionRedirect`
2. loader redirects with a cookie
   1. We go normalRedirect -> normalRedirect but we lose the `location.state.setCookie` flag on the second redirect
   2. This is fixed by carrying forth the `setCookie` flag from the current navigation when kicking off a new `normalRedirect`

Closes: #3302 

- [x] Tests

Testing Strategy:
* Tested locally with https://github.com/jangerhofer/remixRootRouteRedirectReproduction
* Added 2 new integration tests

### DIFF
--- a/integration/redirects-test.ts
+++ b/integration/redirects-test.ts
@@ -11,15 +11,15 @@ test.describe("redirects", () => {
   test.beforeAll(async () => {
     fixture = await createFixture({
       files: {
-        "app/routes/action-submission.jsx": js`
+        "app/routes/action.jsx": js`
           import { Outlet, useLoaderData } from "@remix-run/react";
 
-          if (typeof global.count === "undefined") {
-            global.count = 0;
+          if (typeof global.actionCount === "undefined") {
+            global.actionCount = 0;
           }
 
           export async function loader({ request }) {
-            return { count: ++count };
+            return { count: ++global.actionCount };
           };
 
           export default function Parent() {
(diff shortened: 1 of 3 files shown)


integration/redirects-test.ts
```diff
@@ -0,0 +1,174 @@
...
+      files: {
+        "app/routes/action.jsx": js`
+          import { Outlet, useLoaderData } from "@remix-run/react";
+
+          if (typeof global.actionCount === "undefined") {
+            global.actionCount = 0;
+          }
+
+          export async function loader({ request }) {
+            return { count: ++global.actionCount };
```
brophdawg11, line 22: We use a count to confirm how many times this parent loader runs on the redirects

integration/redirects-test.ts
```diff
@@ -0,0 +1,174 @@
...
+
+          export async function action({ request }) {
+            return redirect("/action/1");
+          };
+
+          export default function Login() {
+            return (
+              <Form method="post">
+                <button type="submit">Submit</button>
+              </Form>
```
brophdawg11, line 48: Submit the form, which redirects to `/action/1` and then to `/action/2`

integration/redirects-test.ts
```diff
@@ -0,0 +1,174 @@
...
+                <Outlet/>
+              </div>
+            );
+          }
+        `,
+
+        "app/routes/loader/link.jsx": js`
+          import { Link } from "@remix-run/react";
+          export default function Parent() {
+            return <Link to="/loader/redirect">Redirect</Link>;
```
brophdawg11, line 101: Click this link, which redirects to `/loader/1` then to `/loader/2`

packages/remix-react/transition.ts
```diff
@@ -1284,7 +1286,11 @@ export function createTransitionManager(init: TransitionManagerInit) {
       // loader redirected during an action reload, treat it like an
       // actionRedirect instead so that all the loaders get called again and the
       // submission sticks around for optimistic/pending UI.
-      if (state.transition.type === "actionReload") {
+      if (
+        state.transition.type === "actionReload" ||
+        ((location.state as any)?.isRedirect &&
+          (location.state as any)?.type === "action")
+      ) {
```
brophdawg11, line 1293: Stay on the `actionRedirect` path if the current load is an action redirect

packages/remix-react/transition.ts
```diff
@@ -1217,7 +1217,9 @@ export function createTransitionManager(init: TransitionManagerInit) {
     invariant(
       state.transition.type === "actionSubmission" ||
         // loader redirected during action reload
-        state.transition.type === "actionReload",
+        state.transition.type === "actionReload" ||
+        // loader redirected during action redirect
+        state.transition.type === "actionRedirect",
```
brophdawg11, line 1222: Allow `actionRedirect` to kick off `actionRedirect` as well

PR #3463, opened 2022-06-13
chore(remix-dev): set default `devServerPort` to 8002

mcansh:
before this, we would set it to a random port always - setting it to 8002 by default allows LiveReload to work in most cases

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey (approved):

MichaelDeBoey:
@mcansh What does this mean for #3411?

mcansh:
> @mcansh What does this mean for #3411?

we should probably still do it, this PR should fix a lot of the problems though

ZackWard:
Thank you for this PR. This would have saved me some time this morning. 👍

PR #3467, opened 2022-06-14
docs(examples): add firestore rules to `firebase-auth-firestore` example

penx:
Add a firestore.rules file ([docs](https://firebase.google.com/docs/firestore/security/get-started)) that disables public access to all Firestore data.

All reads and writes are done via firebase-admin, which skips Firebase rules.

Also added a auth/config.json file to silence auth warning during emulator startup:

```sh
$ yarn emulators
...
⚠  auth: Skipped importing config because /remix/examples/firebase-auth-firestore/firebase-fixtures/auth/config.json does not exist.
```

### DIFF
--- a/examples/firebase-auth-firestore/firebase-fixtures/auth/config.json
+++ b/examples/firebase-auth-firestore/firebase-fixtures/auth/config.json
@@ -1,0 +1,6 @@
+{
+  "signIn": {
+    "allowDuplicateEmails": false
+  },
+  "usageMode": "DEFAULT"
+}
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (approved):

examples/firebase-auth-firestore/firebase-fixtures/auth/config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "signIn": {
+    "allowDuplicateEmails": false
+  },
+  "usageMode": "DEFAULT"
```
penx, line 5: this is generated by a Firebase emulators export, when looking for documentation on what usageMode I couldn't find any but found that references to it were removed 12 days ago

https://github.com/firebase/firebase-tools/pull/4607

...I'll remove this the emulators start fine without it

kentcdodds (approved):
Nice! Thanks!

PR #3478, opened 2022-06-15
chore: lint Deno files

MichaelDeBoey:


packages/remix-deno/globals.ts
```diff
@@ -9,4 +9,5 @@ interface ProcessEnv {
 interface Process {
   env: ProcessEnv;
 }
+// deno-lint-ignore no-unused-vars no-var
 var process: Process;
```
MichaelDeBoey, line 13: @pcattori Do we want to convert this to a `const` or is this the way it should be used & we want to keep the ignore?

pcattori: We're mirroring the types from `@types/node` here and [they use `var` for `process`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/process.d.ts#L5), so let's keep as-is.

MichaelDeBoey: Then I think this is good to be merged

pcattori (approved):

PR #3480, opened 2022-06-15
bug reproduction: returned variable from fetch() is not an instance of global `Response`

manV:
- [ ] Docs
- [x] Tests

Testing Strategy:

The returned variable from fetch should be an instance of global variable `Response` but it does not seem to be the case.

however it seems to be an instance of `Response` from `@remix-run/web-fetch`

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -288,6 +288,7 @@
 - manan30
 - manosim
 - mantey-github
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
This will be fixed by #4148

PR #3482, opened 2022-06-15
chore(docs,examples,templates): use short JSDOC type notation in configs

MichaelDeBoey:
We do this in all our configs, except examples & templates


machour (approved):

kentcdodds:
Thanks!

PR #3491, opened 2022-06-16
docs: update incorrect fetcher.type "load" to "normalLoad"

JeffBeltran:
updated the docs for `fetcher.type` from `load` to `normalLoad`


kentcdodds (approved):
Thanks!

PR #3500, opened 2022-06-16
docs(guides/not-found): add notice about catch propagation requiring a React component export

rowanc1:
Scenario:
* I had an index page that redirected to another page, and so I didn't have an exported react component `Outlet`
* This meant that the catch boundary didn't propagate to the root, and it was really hard to track down :(

I think it makes sense as a feature/design, but could use a note in the docs!

Not sure about the wording, but gave it a shot ... !

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -429,6 +429,7 @@
 - rossipedia
 - RossJHagan
 - RossMcMillan92
+- rowanc1
 - rowinbot
 - rphlmr
 - rtabulov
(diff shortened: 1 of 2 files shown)


brophdawg11:
I'm not sure I understand the scenario being described here? 

If I set this up as an `_index.tsx` file I properly end up in the root boundary showing a 404:

```jsx

export function loader() {
  return redirect("/junk");
}
```

PR #3501, opened 2022-06-17
ci(deployments): define required secrets

mcansh:
https://github.com/remix-run/remix/actions/runs/2513898715

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.github/workflows/deployments.yml
+++ b/.github/workflows/deployments.yml
@@ -2,6 +2,31 @@ name: 🚀 Deployment Tests
 
 on:
   workflow_call:
+    secrets:
+      TEST_AWS_ACCESS_KEY_ID:
+        required: true
+      TEST_AWS_SECRET_ACCESS_KEY:
+        required: true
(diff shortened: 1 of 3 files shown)


MichaelDeBoey:
@mcansh I think this PR causes an error on all forks that don't have these secrets set 🤔
https://github.com/MichaelDeBoey/remix/actions/runs/2523109212

mcansh:
> @mcansh I think this PR causes an error on all forks that don't have these secrets set 🤔
> 
> https://github.com/MichaelDeBoey/remix/actions/runs/2523109212

hm, i'd expect the "if" statement to make that not matter, we can change them to not required

MichaelDeBoey:
> i'd expect the "if" statement to make that not matter

Each step will probably be skipped, but the workflow itself will already error out because they're all required inputs

mcansh:
this is related to a misconfiguration im pretty sure - fixed here https://github.com/remix-run/remix/pull/3520

PR #3503, opened 2022-06-17
chore(docs): Build out CLI docs

chaance:
I was tasked with adding documentation for the new `--port` option for `remix dev`, but in the process realized the CLI docs needed a bit more in general.

This still isn't entirely comprehensive, but I think it's an improvement and at least more fully explains the commands currently included on the docs page.


MichaelDeBoey (approved):

docs/other-api/dev.md
```diff
@@ -5,32 +5,100 @@ order: 1
...
+
+Alternatively, a port can be assigned to the `PORT` environment variable.
+
+<aside aria-label="Running multiple instances of Remix">
+<docs-info>
+
+If you are changing the port to run multiple instances of Remix, you must also set unique [websocket ports](../api/conventions#devserverport) via `remix.config.js`, or by setting the `REMIX_DEV_SERVER_WS_PORT` environment variable.
+
+</docs-info>
+</aside>
```
kentcdodds, line 104: This should not be necessary anymore since we determine this dynamically and devs shouldn't have to worry about it anymore.

```suggestion

```

docs/other-api/dev.md
```diff
@@ -5,32 +5,100 @@ order: 1
...
+The same as `watch`, but also boots the [Remix App Server](serve.md) in development mode if it's installed.
 
 ```sh
 remix dev
 ```
 
-_Note: The default port is `3000`, but can be changed by setting the `PORT` environment variable. If you are changing this to run multiple instances of Remix, you must also set unique [Dev websocket ports](https://remix.run/docs/en/v1/api/conventions#devserverport) via `remix.config.js` or setting `REMIX_DEV_SERVER_WS_PORT`._
+### `remix dev --debug`
+
+Attaches a [Node inspector](https://nodejs.org/en/docs/guides/debugging-getting-started/) to develop your app in debug mode.
```
kentcdodds, line 84: Thank you! `debug` is actually a common query in our docs search, so it'll be good to get at least *something* in the output for that.

kentcdodds (approved):
This is great. Thanks! Just one thing that I'll handle.

kentcdodds (approved):
Thanks!

PR #3508, opened 2022-06-18
Example of using feathersjs, a real time framework, with remix

ryan-snyder:
An example of how to use feathersjs with remix.run. This example shows how you can use feathers to receive both socket.io events/emits and real-time events sent by feathers.  This was heavily adopted from the current existing socket.io example, but changing it to work with feathers and showing some feathersjs specific features. This does not show how to authenticate users with feathers in remix, but that could perhaps be added at a later date, or in a separate example. 

The main reason behind wanting to make this example, is that my current project that I'm trying to transition to remix, uses feathersjs for the backend, and it took a little bit of work trying to use it with remix, so hopefully this will be helpful to other people doing the same thing

I tried to make it as simple and small as possible, but the code for the feathersjs server is perhaps larger than it should be, but 99 percent of it was autogenerated by cli tools, the same cli tools that most people either experienced or getting started with feathersjs would use, but any suggestions are welcome, of course


examples/feathersjs/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Feathersjs is a framework for real-time node applications and this example will show you how to connect to the feathers backend from remix and use it to view events, borrowing heavily from the socket.io example and adapting it for feathersjs. 
+
+To use run `npm run watch` to build and watch for remix changes and `npm run start` to start our server
+
+## Preview
+
+Open this example on [CodeSandbox](https://codesandbox.com):
+
+<!-- TODO: update this link to the path for your example: -->
```
machour, line 11: The codesandbox link must be updated to match your example path

examples/feathersjs/README.md
```diff
@@ -0,0 +1,36 @@
...
+Now, the real cool part is if you open up two clients, messages will get synced between multiple clients without reloads. 
+For this to work, we need to tell feathers to publish an event anytime a message is created. This is done in server/channels.ts.
+As well, we use our client socket to listen for the message created event and call `fetcher.load('/?index')` function
+
+
+For number two, we simply do the same thing done in the socket.io example and send a ping event.
+
+## Related Links
+
+Link to documentation or other related examples.
```
machour, line 36: This needs to be updated as well

examples/feathersjs/README.md
ryan-snyder: done

examples/feathersjs/README.md
ryan-snyder: and done

chaance:

PR #3511, opened 2022-06-19
docs(guides/data-writes): fix `form`/`Form` explanation

vlindhol:
The issue here is that `<Form reloadDocument />` hasn't been mentioned yet, neither in text nor code. So the reader following along with the guide will have `<form />` implemented in their editor at this point.

Note that the next paragraph goes into why `Form` needs the `reloadDocument` prop to be equivalent to `form`, which makes sense, and then we can forget about `form` altogether.

### DIFF
--- a/docs/guides/data-writes.md
+++ b/docs/guides/data-writes.md
@@ -292,7 +292,7 @@ You can ship this code as-is. The browser will handle the pending UI and interru
 
 ### Graduate to `<Form>` and add pending UI
 
-Let's use progressive enhancement to make this UX a bit more fancy. By changing it from `<Form reloadDocument>` to `<Form>`, Remix will emulate the browser behavior with `fetch`. It will also give you access to the pending form data so you can build pending UI.
+Let's use progressive enhancement to make this UX a bit more fancy. By changing it from `<form>` to `<Form>`, Remix will emulate the browser behavior with `fetch`. It will also give you access to the pending form data so you can build pending UI.
 
 ```tsx [2, 11]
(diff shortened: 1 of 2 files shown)


docs/guides/data-writes.md
```diff
@@ -292,7 +292,7 @@ You can ship this code as-is. The browser will handle the pending UI and interru
 
 ### Graduate to `<Form>` and add pending UI
 
-Let's use progressive enhancement to make this UX a bit more fancy. By changing it from `<Form reloadDocument>` to `<Form>`, Remix will emulate the browser behavior with `fetch`. It will also give you access to the pending form data so you can build pending UI.
+Let's use progressive enhancement to make this UX a bit more fancy. By changing it from `<form>` to `<Form>`, Remix will emulate the browser behavior with `fetch`. It will also give you access to the pending form data so you can build pending UI.
```
machour, line 295: I think the idea was to imply that `<form/>` is equivalent to `<Form reload document />`. Might be worth adding this in a parenthesis (or.. ).

WDYT @kentcdodds ?

vlindhol: The issue here is that `<Form reloadDocument />` hasn't been mentioned even once, either in text nor code. So the reader following along with the guide will have `<form />` implemented in their editor at this point.

The paragraph below this one does mention the differences between `form` and `Form`, and the necessity of adding `reloadDocument` to get better UX. This mini-PR doesn't take a stance on that part :)

machour (approved):

kentcdodds (approved):
Thanks!

PR #3521, opened 2022-06-20
chore(integration): fix type errors and start enforcing type checks

jenseng:
Fix errors that currently show up in VSCode with typescript 4.5.5 (the current version Remix uses). Playwright only transpiles typescript, it doesn't do any type checking. In addition to fixing the errors, start type checking `integration` when we build.

### DIFF
--- a/integration/helpers/playwright-fixture.ts
+++ b/integration/helpers/playwright-fixture.ts
@@ -22,10 +22,15 @@ export class PlaywrightFixture {
    * @param waitForHydration Will wait for the network to be idle, so
    * everything should be loaded and ready to go
    */
-  async goto(href: string, waitForHydration?: true) {
-    return this.page.goto(this.app.serverUrl + href, {
+  async goto(href: string, waitForHydration?: true): Promise<Response> {
+    let response = await this.page.goto(this.app.serverUrl + href, {
       waitUntil: waitForHydration ? "networkidle" : undefined,
     });
+    if (response == null)
+      throw new Error(
+        "Unexpected null response, possible about:blank request or same-URL redirect"
+      );
+    return response;
   }
 
   /**
@@ -42,7 +47,7 @@ export class PlaywrightFixture {
       throw new Error(`Could not find link for ${selector}`);
     }
     if (options.wait) {
-      await doAndWait(this.page, () => el.click());
+      await doAndWait(this.page, () => el!.click());
     } else {
       await el.click();
     }
@@ -94,7 +99,7 @@ export class PlaywrightFixture {
       }
     }
     if (options.wait) {
-      await doAndWait(this.page, () => el.click());
+      await doAndWait(this.page, () => el!.click());
     } else {
       await el.click();
     }
@@ -108,7 +113,7 @@ export class PlaywrightFixture {
     if (!el) {
       throw new Error(`Can't find element for: ${selector}`);
     }
-    await doAndWait(this.page, () => el.click());
+    await doAndWait(this.page, () => el!.click());
   }
 
   /**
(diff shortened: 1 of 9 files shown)


MichaelDeBoey:
This seems to cause the build to fail, could you have a look at fixing that please?

jenseng:
✅  Fixed, there was a [new typecheck failure](https://github.com/remix-run/remix/pull/3521/files#diff-0b29475230aa2db12b48d45be9b162817bc2ff6e75909d869ed973e7d9d60f59) after I rebased due to things now being [built into `dist`](https://github.com/remix-run/remix/commit/0f8279d3fd2f14096a78b592b2790c70c935808a) ... I had re-tested locally, but hadn't cleaned out the old built files, so it still worked 🙃

integration/tsconfig-test.ts
```diff
@@ -27,7 +28,7 @@ const DEFAULT_CONFIG = {
     strict: true,
     target: "ES2019",
     baseUrl: ".",
-  },
+  } as Omit<TsConfigJson.CompilerOptions, "plugins">,
```
MichaelDeBoey, line 31: Wouldn't it be better to type `DEFAULT_CONFIG` instead?

jenseng: sure, I can do that, it might just be a smidge more verbose since I have to exclude the nested `plugins` which have `unknown` properties and make the whole thing not JSON.stringifiable

jenseng: Done

integration/compiler-test.ts
```diff
@@ -165,7 +165,7 @@ test.describe("compiler", () => {
   test("removes server code with `*.server` files", async ({ page }) => {
     let app = new PlaywrightFixture(appFixture, page);
     let res = await app.goto("/", true);
-    expect(res.status()).toBe(200); // server rendered fine
+    expect(res?.status()).toBe(200); // server rendered fine
```
kentcdodds, line 168: Thanks for this! Is there any way we can make it so `app.goto` always returns a `res`? I can't think of a situation where it'd be ok if it didn't 😬

integration/helpers/playwright-fixture.ts
```diff
@@ -42,7 +42,7 @@ export class PlaywrightFixture {
       throw new Error(`Could not find link for ${selector}`);
     }
     if (options.wait) {
-      await doAndWait(this.page, () => el.click());
+      await doAndWait(this.page, () => el!.click());
```
kentcdodds, line 45: Hey @ryanflorence, I'm pretty sure if we used `const el` above, we wouldn't have to do this :P

kentcdodds (commented):
This is very helpful @jenseng, thank you! Just one question.

integration/compiler-test.ts
MichaelDeBoey: We're just re-using `playwright`'s `app.goto` under the hood, which is returning `null | Response`.

```ts
  /**
   * Returns the main resource response. In case of multiple redirects, the navigation will resolve with the response of the
   * last redirect.
   *
   * [...]
   * 
   * > NOTE: The method either throws an error or returns a main resource response. The only exceptions are navigation to
   * `about:blank` or navigation to the same URL with a different hash, which would succeed and return `null`.
   *
   * [...]
   */
```

Since we're not going to use `about:blank`, I think we can make our helper function just always return `Response`

integration/helpers/playwright-fixture.ts
jenseng: Yup, using `const` fixes it (but then I'd have to use `eslint-disable-next-line`). Alternatively I could remove the `!` and add `@ts-expect-error` (until such time as [narrowing is improved](https://github.com/microsoft/TypeScript/issues/35124)). Happy to switch to one of these if we prefer them stylistically.

integration/compiler-test.ts
jenseng: Done

kentcdodds (approved):
This looks great. Thank you so much for working on this.

PR #3528, opened 2022-06-21
feat: extract `@remix-run/deno-deploy` out of `@remix-run/deno`

MichaelDeBoey:
As mentioned before in #3442 & #3262, we should have done this from the start instead of putting the adapter in the same package as the runtime.

To not break backwards compatibility, `@remix-run/deno` now depends on `@remix-run/deno-deploy`.
I can remove this and do the other way around (as it's supposed to be) if wanted.
That will break backwards compatibility though.


pcattori:
@MichaelDeBoey : Until #3262 (or similar) is merged _and accepted_ (right now its just proposed), we should hold off on reworking the package structure for adapters. I actually think #3262 is proposing two orthogonal things: (1) shared server package and (2) adapters export server features, and we might not accept either or might just accept (1) but not (2) etc...

pcattori:
> Deno one is the odd one and needs to be used differently than all other ones.

How so?

MichaelDeBoey:
`@remix-run/deno` is now **both** an adapter + a runtime 1 one package, whereas with all other adapter/runtimes they're split into 2 packages

MichaelDeBoey:
Gotcha, thanks for making me understand!

pcattori:
> Gotcha, thanks for making me understand!

For sure! You asked good questions 💯

PR #3529, opened 2022-06-21
refactor(remix-dev/cli): use `convert-to-javascript` migration internally

MichaelDeBoey:
Follow-up of #3150


packages/remix-dev/cli/migrate/migrations/convert-to-javascript/index.ts
```diff
@@ -34,16 +34,20 @@ export const convertToJavaScript: MigrationFunction = async ({
     transformPath: TRANSFORM_PATH,
   });
   if (!codemodOk) {
-    console.error("❌ I couldn't update your imports to JS.");
-    if (!flags.debug) {
-      console.log("👉 Try again with the `--debug` flag to see what failed.");
+    if (!flags.silent) {
```
kentcdodds, line 37: I think if there's an error we should have output even if silent is set to true right?

kentcdodds (changes requested):
Looks good. Just one question.

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/index.ts
MichaelDeBoey: The reason why I want to hide the `👉 Try again with the `--debug` flag to see what failed.` message, is that we're running the migration programmatically instead of via the `migrate` CLI

kentcdodds: I see what you mean. Maybe we should do some code duplication then. Seems like we have two different use cases for this code.

MichaelDeBoey: That would mean a lot of duplication (like almost the whole folder) I'm afraid 😕

pcattori: `--no-interactive` is a common flag for doing that sort of thing, so we could use that

pcattori: Or actually, even better, we can auto-detect if we're running via command line or as a function via `require.main === module`. see https://stackoverflow.com/a/6398335/1490091

kentcdodds: That sounds great @pcattori 👍

MichaelDeBoey: Just to be clear: I should put the `isInteractive` check _only_ around the `--debug` log, right?

kentcdodds: Yep

packages/remix-dev/cli/create.ts
```diff
@@ -195,8 +195,8 @@ export async function createApp({
     !useTypeScript &&
     fse.existsSync(path.join(projectDir, "tsconfig.json"))
   ) {
-    let spinner = ora("Converting template to JavaScript…").start();
-    await convertTemplateToJavaScript(projectDir);
+    let spinner = ora("Migrating template to JavaScript…").start();
+    await convertToJavaScript({ flags: {}, projectDir });
```
pcattori, line 199: Can you make the `flags` parameter optional? Also might be good to rename to `options` instead of `flags` since "flags" is more of a CLI term and `convertToJavascript` can be called programmatically like we do here.

Also, I think a nicer function signature for `convertToJavascript` might be:

```ts
function convertToJavascript(projectDir: string, options?: { /* relevant options corresponding to flags */} = {})
```

So that calling it here would be:

```suggestion
    await convertToJavaScript(projectDir);
```

pcattori (changes requested):

packages/remix-dev/cli/create.ts
MichaelDeBoey: This would mean we'll have to refactor the `MigrationFunction` signature as well.
Do we want this for all of the migrations then?
https://github.com/remix-run/remix/blob/f3691d51027b93caa3fd2cdfe146d7b62a6eb8f2/packages/remix-dev/cli/migrate/types.ts#L3-L6

Also: do you want the `Flags` type to be renamed to `Options` as well?
If so, we would need to rename all `flags` args to `options` as well, which isn't always suited imo.

In `packages/remix-dev/cli/migrate/jscodeshift.ts` for instance, it's actually representing JSCodeShift flags
https://github.com/remix-run/remix/blob/f3691d51027b93caa3fd2cdfe146d7b62a6eb8f2/packages/remix-dev/cli/migrate/jscodeshift.ts#L15-L20

In `packages/remix-dev/cli/migrate/resolveInput.ts` it's representing CLI flags
https://github.com/remix-run/remix/blob/f3691d51027b93caa3fd2cdfe146d7b62a6eb8f2/packages/remix-dev/cli/migrate/resolveInput.ts#L38-L54

kentcdodds: Yeah, I didn't have an issue with `flags` because that's the JS API for jscodeshift. I don't mind switching it to `options` though. Maybe any flags that go to `jscodeshift` would be on a sub-option called "flags" and all our own options would go as properties on the options object?

I do like the `projectDir` as the first argument though.

pcattori (approved):

packages/remix-dev/cli/migrate/types.ts
```diff
@@ -1,9 +1,11 @@
...
 
-export type MigrationFunction = (args: {
-  flags: Flags;
-  projectDir: string;
-}) => Promise<void>;
+export type MigrationFunction = (
+  projectDir: string,
+  options?: {
+    flags: Flags;
+  }
```
pcattori, line 7: nit: `flags` are already a map of options, so flattening this would simplify. If we need non-flag options later, then we can add the `options` wrapper, but right now we don't need it:

```suggestion
  flags?: Flags,
```

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/index.ts
```diff
@@ -10,16 +10,17 @@ import { convertTSFilesToJS } from "./convertTSFilesToJS";
...
 
-export const convertToJavaScript: MigrationFunction = async ({
-  flags,
+export const convertToJavaScript: MigrationFunction = async (
   projectDir,
-}) => {
+  flags = {}
+) => {
   let config = await readConfig(projectDir);
+  let isInteractive = require.main === module;
```
MichaelDeBoey, line 18: @pcattori This seems to break our test 🤔

https://github.com/remix-run/remix/runs/7009823608?check_suite_focus=true#step:6:163

packages/remix-dev/__tests__/create-test.ts
```diff
@@ -275,16 +275,10 @@ describe("the create command", () => {
...
-    expect(fse.existsSync(path.join(projectDir, "app/root.jsx"))).toBeTruthy();
     expect(fse.existsSync(path.join(projectDir, "app/root.tsx"))).toBeFalsy();
+    expect(fse.existsSync(path.join(projectDir, "app/root.jsx"))).toBeTruthy();
     expect(fse.existsSync(path.join(projectDir, "tsconfig.json"))).toBeFalsy();
     expect(fse.existsSync(path.join(projectDir, "jsconfig.json"))).toBeTruthy();
-    expect(fse.existsSync(path.join(projectDir, "app/utils.js"))).toBeTruthy();
-    let pkgJSON = JSON.parse(
-      fse.readFileSync(path.join(projectDir, "package.json"), "utf-8")
-    );
-    expect(Object.keys(pkgJSON.devDependencies)).not.toContain("typescript");
```
MichaelDeBoey, line 287: Made this test smaller, as we're already testing this in the migration test itself.

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/index.ts
MichaelDeBoey: Tests still seem to fail 😕

https://github.com/remix-run/remix/runs/7011654283?check_suite_focus=true#step:6:190

packages/remix-dev/__tests__/migrations/convert-to-javascript-test.ts
```diff
@@ -68,56 +103,53 @@ const makeApp = () => {
...
-    );
-    expect(packageJson.devDependencies).not.toContain("@types/react");
-    expect(packageJson.devDependencies).not.toContain("@types/react-dom");
-    expect(packageJson.devDependencies).not.toContain("typescript");
-    expect(packageJson.scripts).not.toContain("typecheck");
+    await checkMigrationRanSuccessfully(projectDir);
+
+    console.log("debug: ---START---");
+    console.log("debug: output", output);
+    console.log("debug: ---END---");
```
kentcdodds, line 139: I think we want to remove these

MichaelDeBoey: Indeed.
Just left them here for @pcattori to be able to have a look at what's going wrong with what he proposed in https://github.com/remix-run/remix/pull/3529#discussion_r902850714

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/index.ts
pcattori: Are the tests passing for you locally?

MichaelDeBoey: No they're failing locally as well with this 😕

pcattori: If I run `yarn jest __tests__/replace-remix-imports-test.ts` within `packages/remix-dev`, the test passes, but running `yarn jest` (either in `packages/remix-dev` or in repo root) causes the test to fail.

pcattori:
@MichaelDeBoey I think the best way to proceed is to introduce a `--interactive` flag that to the CLI that defaults to `require.main === module`, but can be set explicitly by users or by tests. Note that we [aren't catering to the test user](https://kentcdodds.com/blog/avoid-the-test-user), since _actual_ users may also want to set `--interactive=false` in their automation.

MichaelDeBoey:
@kentcdodds @pcattori `--interactive` & `--no-interactive` flags are implemented in the `convert-to-javascript` migration now 👍

packages/remix-dev/__tests__/migrations/convert-to-javascript-test.ts
pcattori: @MichaelDeBoey : we can remove these debugging statements now, right?

packages/remix-dev/__tests__/migrations/convert-to-javascript-test.ts
```diff
@@ -68,56 +103,54 @@ const makeApp = () => {
...
```
pcattori, line 140: Let's delete the debugging statements

```suggestion
```

pcattori (changes requested):

pcattori (approved):

kentcdodds (approved):
Super! Thank you so much for working on this @MichaelDeBoey and @pcattori!

PR #3532, opened 2022-06-21
feat: force bundle LiveReload to prevent process.env variables not being replaced

mcansh:
supersedes #3411 

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


packages/remix-react/components.tsx
```diff
@@ -1417,56 +1417,3 @@ export function useFetchers(): Fetcher[] {
   let { fetchers } = transitionManager.getState();
   return [...fetchers.values()];
 }
-
-// Dead Code Elimination magic for production builds.
-// This way devs don't have to worry about doing the NODE_ENV check themselves.
-// If running an un-bundled server outside of `remix dev` you will still need
-// to set the REMIX_DEV_SERVER_WS_PORT manually.
-export const LiveReload =
```
kentcdodds, line 1425: Can we add `export { LiveReload } from "./LiveReload";`? That way this isn't a breaking change and it should still work right?

mcansh: if you import from root it won't be bundled, looking into a "workaround" though

mcansh: alright, using [`db38a4b` (#3532)](https://github.com/remix-run/remix/pull/3532/commits/db38a4bd229e2988a9973ae3f2df22d9d35e9137) the root import is now changed to the deep import, i just need `serverBareModulesPlugin` to treat it as the deep import and not the root one 🙃

mcansh: onLoad i think runs after onResolve, not vice versa, so i got some more spelunking to do

PR #3540, opened 2022-06-22
chore: Add script for copying build output to package `dist` directories

chaance:
This will be useful for publishing via Changesets. We can remove this once we've reconfigured Rollup accordingly.


mcansh:
> @mcansh Any clue why we'd suddenly be running out of memory in CI? I don't feel like anything has substantially changed about the build other than including the Rollup TS plugin, and I have no such issues w/ Node's default memory limit running locally.

nothing's jumping out at me, but i can take a look

scripts/copy-build-to-dist.mjs
```diff
@@ -0,0 +1,88 @@
+import path from "path";
+import fse from "fs-extra";
+import chalk from "chalk";
+
+const ROOT_DIR = process.cwd();
+const BUILD_PATH = path.join(ROOT_DIR, "build", "node_modules");
```
brophdawg11, line 6: I don't think it's anything we. need to deal with, but this doesn't take into account `REMIX_LOCAL_DEV_OUTPUT_DIRECTORY` which makes for an interesting flow:

* Assume no existing `build/` directory
* `REMIX_LOCAL_DEV_OUTPUT_DIRECTORY=../some-app yarn build`
* `rollup` builds output to the specific app folder
* `tsc` doesn't have awareness of that env var (a known current issue), so it still builds types to `build/`
* This script then copies only the updates types to `dist/` leaving them potentially out of sync with the source code in `dist/`

If the rollup ts plugin wasn't so slow I think this would be a non-issue.   And I think it's also fine to ignore since it's just a local dev shortcut and we can just work around it.  Just wanted to call it out for visibility though

brophdawg11 (approved):
Builds looked ok to me locally.  Think it's worth publishing an experimental from here and actually npm installing a few things to make sure it all looks good?

scripts/copy-build-to-dist.mjs
chaance: I went ahead and updated this to take that env variable into account. Better to copy just the js files than just the types! Hopefully we can figure out a decent solution for types later.

chaance:
@brophdawg11 Successfully tested with `npx create-remix@0.0.0-experimental-c5bb5cb31`. If you want to spin it up you'll need to force install to deal w/ peer deps (need to fix that in the version script), but should be good to go.

PR #3542, opened 2022-06-22
ci: reduce node_versions we test against

mcansh:
on PRs/commits to main or dev, we'll test against the "latest" version of node, but for releases (nightlies, experimental, pre-releases, stable) we'll test against v14, v16, and v18

i've also moved the reusable workflows into a sub directory of the workflow directory to more easily tell that aren't run on their own.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


mcansh:
dang it https://github.com/remix-run/remix/actions/runs/2551054712

PR #3543, opened 2022-06-22
chore(refactor): Create `magicExports` in Rollup build

chaance:
This PR removes all `magicExports` modules from the source code and emits the files via Rollup instead. This eliminates any issues with TS or Node module resolution that will occur from copying the build output into a `dist` directory alongside the package source. 

This will unblock https://github.com/remix-run/remix/pull/3540 which in turn will unblock the work needed to implement Changesets.


package.json
```diff
@@ -91,6 +92,7 @@
     "jest-watch-select-projects": "^2.0.0",
     "jest-watch-typeahead": "^0.6.5",
     "jsonfile": "^6.0.1",
+    "lodash": "^4.17.21",
```
MichaelDeBoey, line 95: Could we use `lodash.camelcase` instead?

https://www.npmjs.com/package/lodash.camelcase

```suggestion
    "lodash.camelcase": "^4.3.0",
```

package.json
```diff
@@ -65,6 +65,7 @@
     "@types/glob": "7.2.0",
     "@types/jest": "^27.4.1",
     "@types/jsonfile": "^6.1.0",
+    "@types/lodash": "^4.14.182",
```
MichaelDeBoey, line 68:
```suggestion
    "@types/lodash.camelcase": "^4.3.7",
```

rollup.config.js
```diff
@@ -4,6 +4,7 @@ import nodeResolve from "@rollup/plugin-node-resolve";
 import copy from "rollup-plugin-copy";
 import fse from "fs-extra";
 import fs from "fs";
+import camelCase from "lodash/camelCase";
```
MichaelDeBoey, line 7:
```suggestion
import camelCase from "lodash.camelcase";
```

rollup.config.js
```diff
@@ -633,61 +540,271 @@ function getAdapterConfig(adapterName) {
...
+  let magicExports = null;
+  switch (packageName) {
+    case "@remix-run/architect":
+      magicExports = {
+        values: {
+          "@remix-run/architect": ["createArcTableSessionStorage"],
+        },
+        types: {
+          "@remix-run/node": ["UploadHandler", "UploadHandlerPart"],
+        },
```
MichaelDeBoey, line 573: These looks like a left-over from a copy/paste?

https://github.com/remix-run/remix/blob/e85757e84cf7e83cf754681917cf4bcce717e544/packages/remix-architect/magicExports/remix.ts#L1-L5

```suggestion
```

rollup.config.js
```diff
@@ -633,61 +540,271 @@ function getAdapterConfig(adapterName) {
...
+      };
+      break;
+    case "@remix-run/cloudflare-pages":
+      magicExports = {
+        values: {
+          "@remix-run/cloudflare": ["createCloudflareKVSessionStorage"],
+        },
+      };
+      break;
```
MichaelDeBoey: I don't think this one is necessary, because we already have this exported by `@remix-run/cloudflare`

```suggestion
```

rollup.config.js
```diff
@@ -633,61 +540,271 @@ function getAdapterConfig(adapterName) {
...
+    case "@remix-run/cloudflare-workers":
+      magicExports = {
+        values: {
+          "@remix-run/cloudflare": ["createCloudflareKVSessionStorage"],
+        },
+      };
+      break;
```
MichaelDeBoey: Same

```suggestion
```

MichaelDeBoey (changes requested):

package.json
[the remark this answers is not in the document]
MichaelDeBoey: Oops you were faster than I was reviewing 🙈

brophdawg11:
The only thing I notice on a quick diff of a dev `build/` versus here is these 2 lines are missing at the top of the generated files in this branch.  We don't need `use strict` - but unsure if the `__esModule` is relevant for anything in the bundle landscape or just rollup cruft though

```js
'use strict';

Object.defineProperty(exports, '__esModule', { value: true });
```

brophdawg11 (approved):

MichaelDeBoey (changes requested):

rollup.config.js
```diff
@@ -633,61 +540,266 @@ function getAdapterConfig(adapterName) {
...
+        values: {
+          "@remix-run/architect": ["createArcTableSessionStorage"],
+        },
+      };
+    case "@remix-run/cloudflare-pages":
+      return {
+        values: {
+          "@remix-run/cloudflare": ["createCloudflareKVSessionStorage"],
+        },
+      };
```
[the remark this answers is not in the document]
chaance: I'm going to leave these as-is, but we can revisit in a follow-up PR if needed. The goal here is to change the config but leave the published output alone.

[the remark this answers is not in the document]
MichaelDeBoey: I created https://github.com/remix-run/remix/pull/3546 as a follow-up

PR #3547, opened 2022-06-23
docs: jokes app tutorial in fetching jokes part

justinsalasdev:
at point where reader is only done with building types for `Joke`, the example shows to import `type User` from `@prisma/client` where it is not there yet.

the succeeding code continuation also didn't use the built type from `@prisma/client`

### DIFF
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -1620,20 +1620,23 @@ Remix and the `tsconfig.json` you get from the starter template are configured t
 
 <summary>app/routes/jokes.tsx</summary>
 
-```tsx filename=app/routes/jokes.tsx lines=[3,4,6,12,19-21,23-28,31,55-59]
+```tsx filename=app/routes/jokes.tsx lines=[3,6,8,16,23-25,27-32,35,59-63]
 import type {
   LinksFunction,
   LoaderFunction,
-  json,
 } from "@remix-run/node";
+
 import type { Joke } from "@prisma/client";
+
+import { json } from "@remix-run/node";
+
 import {
   Link,
   Outlet,
   useLoaderData,
 } from "@remix-run/react";
 
-import { db } from "~/utils/db.server";
+import { db } from "~/utils/db.server"; 
 import stylesUrl from "~/styles/jokes.css";
 
 export const links: LinksFunction = () => {
@@ -1652,7 +1655,7 @@ export const loader: LoaderFunction = async () => {
 };
 
 export default function JokesRoute() {
-  const data = useLoaderData<LoaderData>();
+  const data = useLoaderData<LoaderData>();35
 
   return (
     <div className="jokes-layout">
@@ -1708,7 +1711,7 @@ I want to call out something specific in my solution. Here's my loader:
 
 ```tsx lines=[2,8-10]
 type LoaderData = {
-  jokeListItems: Array<Pick<Joke,"id" | "name">>;
+  jokeListItems: Array<{ id: string; name: string }>;
 };
 
 export const loader: LoaderFunction = async () => {
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -1655,7 +1655,7 @@ export const loader: LoaderFunction = async () => {
 };
 
 export default function JokesRoute() {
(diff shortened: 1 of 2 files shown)


docs/tutorials/jokes.md
```diff
@@ -1620,12 +1620,13 @@ Remix and the `tsconfig.json` you get from the starter template are configured t
 
 <summary>app/routes/jokes.tsx</summary>
 
-```tsx filename=app/routes/jokes.tsx lines=[3,5,12,19-21,23-28,31,55-59]
+```tsx filename=app/routes/jokes.tsx lines=[3,4,6,12,19-21,23-28,31,55-59]
 import type {
   LinksFunction,
   LoaderFunction,
+  json,
```
machour, line 1627: This is a `type` import, so json shouldn't be imported from there

docs/tutorials/jokes.md
```diff
@@ -1705,9 +1706,9 @@ And here's what we have with that now:
 
 I want to call out something specific in my solution. Here's my loader:
 
-```tsx lines=[8-10]
+```tsx lines=[2,8-10]
 type LoaderData = {
-  jokeListItems: Array<{ id: string; name: string }>;
+  jokeListItems: Array<Pick<Joke,"id" | "name">>;
```
machour, line 1711: Although this is correct, it may be confusing for a new comer to typescript in this on boarding tutorial.

docs/tutorials/jokes.md
```diff
@@ -1620,12 +1620,13 @@ Remix and the `tsconfig.json` you get from the starter template are configured t
 
 <summary>app/routes/jokes.tsx</summary>
 
-```tsx filename=app/routes/jokes.tsx lines=[3,5,12,19-21,23-28,31,55-59]
+```tsx filename=app/routes/jokes.tsx lines=[3,4,6,12,19-21,23-28,31,55-59]
```
machour, line 1623: Needs to be reverted, see next comment

machour (changes requested):
Thank you for your contribution @justinsalasdev !

Left you a couple of remarks to address.

docs/tutorials/jokes.md
justinsalasdev: @machour, thanks for your feedbacks ser - had reflected them in recent commits
* separated `json` import from `types` import
* imported `Joke` type on `findMany` example - placed directly below `import type`'s for grouping (updated line highlight for new lines added)
* removed `Pick` from projected example and replaced with `{ id:string, name:string }` - add line highlight so reader is aware that original `Joke` type needs to be edited

kentcdodds (changes requested):
Hi @justinsalasdev. Thanks for the PR.

As noted in the text right before your changes:

> this is just an example. No need to copy/paste this 😄

It's only there to give you an idea of how things work as an example so you can finish the goal without looking at the finished version. We purposefully don't use `Joke` there. I think using Joke would be *more* confusing because it would lead people to believe that they need to make those changes to their code which they do not.

I'd be more in favor of changing the example completely to be something like `Sandwich` or something.

justinsalasdev:
> Hi @justinsalasdev. Thanks for the PR.
> 
> As noted in the text right before your changes:
> 
> > this is just an example. No need to copy/paste this 😄
> 
> It's only there to give you an idea of how things work as an example so you can finish the goal without looking at the finished version. We purposefully don't use `Joke` there. I think using Joke would be _more_ confusing because it would lead people to believe that they need to make those changes to their code which they do not.
> 
> I'd be more in favor of changing the example completely to be something like `Sandwich` or something.

@kentcdodds, still can't believe I got an actual pr review from you ^__^
I must admit I have overlooked that example caveat. I skimmed the tutorial and found the `User` model so when I decided to do the tutorial, I got confused when it's importing `User` even do I haven't built it. 

Overall, here are the changes I've done
* changed example type from `User` to `Sandwich`
* imported `Joke` type in answer portion and used it for `findMany`, also adjusted highlight numbers due to spaces added
* used `{ id:string, name:string }` type on fetch overload example, and highlight its change from `Joke`

kentcdodds (approved):
Stellar. Thank you so much @justinsalasdev 👏

PR #3553, opened 2022-06-23
ci: debug failing deployment tests

mcansh:
this pr addresses a few deployment issues in CI among a few other things

- replaces `jsonfile` with `@npmcli/package-json` for updating package json
- re-adds the commit sha to app names in the event it’s taken due to unforeseen reasons (sans deno as you can’t delete a project via the CLI)
- updates env handling to be explicit
- adds exponential waiting to fetchRetrying a url (deployment or npm package) (capped at 10s)
- adds a bit of logging to fetchRetry (see above)
- actually throws an error if the url isn’t reachable via fetchRetry
- updates how the script is exited so the cleanup function is ran
- moves various deployment cli’s to the deployment app itself instead of having them all live in the root deployment scripts directory


kentcdodds:
> adds exponential waiting to fetchRetrying a url (deployment or npm package)

How long before this times out?

chaance (approved):

mcansh:
> Should we also add a random couple characters at the end just in case we try to re-deploy the same commit?

just added

> How long before this times out?

uh.. before it was about 4.5 minutes if it made it to the tenth attempt, so a total of 8.5 minutes, I added a max wait of 10s, so it should be about 1 minute now, but perhaps thats a bit short (we shouldn't make it to any retries unless something happens to npm i imagine)

kentcdodds (approved):
Thanks.

PR #3566, opened 2022-06-24
fix(remix-dev): fix config call when using Yarn PnP

lensbart:
Related to #683 (already closed)

* Lists `react` and `react-dom` as peer dependencies for all packages that consume `@remix-run/server-runtime` and don’t already have `react`(`-dom`) as a dependency, because `@remix-run/server-runtime` itself [lists](https://github.com/remix-run/remix/blob/d397eaffbf4e8dc4288aed2e6421c7001690dc6a/packages/remix-server-runtime/package.json#L31-L34) `react`(`-dom`) as peer dependencies. This prevents the following warning when running `yarn install`:
   ```
   ➤ YN0002: │ @remix-run/deno doesn't provide react (p1dd1e), requested by @remix-run/server-runtime
   ➤ YN0002: │ @remix-run/deno doesn't provide react-dom (p56944), requested by @remix-run/server-runtime
   ➤ YN0002: │ @remix-run/dev [bc263] doesn't provide react (pbd2ca), requested by @remix-run/server-runtime
   ➤ YN0002: │ @remix-run/dev [bc263] doesn't provide react-dom (pb7dc6), requested by @remix-run/server-runtime
   ```
* Silences harmless error when `yarn config get @remix-run:registry` throws because the setting can’t be found

### DIFF
--- a/packages/remix-cloudflare/package.json
+++ b/packages/remix-cloudflare/package.json
@@ -21,9 +21,7 @@
     "@cloudflare/workers-types": "^3.4.0"
   },
   "peerDependencies": {
-    "@cloudflare/workers-types": "^2.0.0 || ^3.0.0",
-    "react": ">=16.8",
-    "react-dom": ">=16.8"
+    "@cloudflare/workers-types": "^2.0.0 || ^3.0.0"
   },
   "engines": {
     "node": ">=14"
--- a/packages/remix-cloudflare/package.json
+++ b/packages/remix-cloudflare/package.json
@@ -21,7 +21,9 @@
     "@cloudflare/workers-types": "^3.4.0"
   },
   "peerDependencies": {
-    "@cloudflare/workers-types": "^2.0.0 || ^3.0.0"
+    "@cloudflare/workers-types": "^2.0.0 || ^3.0.0",
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
   },
   "engines": {
     "node": ">=14"
--- a/packages/remix-deno/package.json
+++ b/packages/remix-deno/package.json
@@ -17,10 +17,6 @@
     "@remix-run/server-runtime": "1.6.1",
     "mime": "^3.0.0"
   },
-  "peerDependencies": {
-    "react": ">=16.8",
-    "react-dom": ">=16.8"
-  },
   "engines": {
     "node": ">=14"
   }
--- a/packages/remix-deno/package.json
+++ b/packages/remix-deno/package.json
@@ -17,6 +17,10 @@
     "@remix-run/server-runtime": "1.6.1",
     "mime": "^3.0.0"
   },
+  "peerDependencies": {
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
+  },
   "engines": {
     "node": ">=14"
   }
--- a/packages/remix-dev/cli/create.ts
+++ b/packages/remix-dev/cli/create.ts
@@ -205,12 +205,14 @@ export async function createApp({
   if (installDeps) {
     let packageManager = getPreferredPackageManager();
 
-    let npmConfig = execSync(
-      `${packageManager} config get @remix-run:registry`,
-      {
+    let npmConfig;
+    try {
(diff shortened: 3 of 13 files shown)


packages/remix-cloudflare/package.json
```diff
@@ -21,7 +21,9 @@
     "@cloudflare/workers-types": "^3.4.0"
   },
   "peerDependencies": {
-    "@cloudflare/workers-types": "^2.0.0 || ^3.0.0"
+    "@cloudflare/workers-types": "^2.0.0 || ^3.0.0",
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
```
MichaelDeBoey, line 26: We already had a couple of PRs doing this and we decided we don't want this as these aren't necessary in any way.
This is only there because we're relying on `react-router`, which has these in `peerDependencies` as well.

Once we're switch to `@remix-run/router` (once `react-router` v6.4 is released), these won't be necessary anymore in `peerDependencies`, so all problems will be gone as well.

```suggestion
    "@cloudflare/workers-types": "^2.0.0 || ^3.0.0"
```

packages/remix-deno/package.json
```diff
@@ -17,6 +17,10 @@
     "@remix-run/server-runtime": "1.6.1",
     "mime": "^3.0.0"
   },
+  "peerDependencies": {
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
+  },
```
MichaelDeBoey, line 23:
```suggestion
```

packages/remix-dev/package.json
```diff
@@ -76,7 +76,9 @@
     "type-fest": "^2.12.2"
   },
   "peerDependencies": {
-    "@remix-run/serve": "1.5.1"
+    "@remix-run/serve": "1.5.1",
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
```
MichaelDeBoey, line 81:
```suggestion
    "@remix-run/serve": "1.5.1"
```

packages/remix-node/package.json
```diff
@@ -29,6 +29,10 @@
     "@types/cookie-signature": "^1.0.3",
     "@types/source-map-support": "^0.5.4"
   },
+  "peerDependencies": {
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
+  },
```
MichaelDeBoey, line 35:
```suggestion
```

packages/remix-dev/cli/create.ts
```diff
@@ -205,12 +205,14 @@ export async function createApp({
...
-      `${packageManager} config get @remix-run:registry`,
-      {
+    let npmConfig;
+    try {
+      npmConfig = execSync(`${packageManager} config get @remix-run:registry`, {
         encoding: "utf8",
-      }
-    );
+      });
+    } catch (error: unknown) {
```
MichaelDeBoey, line 213: Since we're not using `error` at all, we can remove it

```suggestion
    } catch {
```

MichaelDeBoey (changes requested):

packages/remix-cloudflare/package.json
lensbart: Thank you for that context. The desired result (no warning when using Yarn) can also be achieved by setting 
```json
  "peerDependenciesMeta": {
    "react": {
      "optional": true
    },
    "react-dom": {
      "optional": true
    }
  }
```

in the `package.json` of `remix-server-runtime`. I’ve made this change now, let me know if you disagree

MichaelDeBoey: Since it's only a warning, I think we can just leave it as-is
The warning will disappear automatically once we made the switch to `@remix-run/router`.

lensbart: While this might be fixed in the future, I’d still err on the side of correctness in the meantime.

This is a warning when running `yarn install`, but an error (via `@yarnpkg/esbuild-plugin-pnp` when running `yarn build`):
Screenshot 2022-06-25 at 17 41 40

lensbart:
Closing in favour of https://github.com/remix-run/remix/pull/3594

PR #3569, opened 2022-06-24
chore: Introduce Changesets

chaance:


MichaelDeBoey (approved):

package.json
```diff
@@ -41,7 +41,11 @@
     "test:integration": "playwright test --config ./integration/playwright.config.ts",
     "posttest:integration": "rimraf ./.tmp/integration",
     "test:primary": "jest",
+    "changeset": "changeset",
```
chaance, line 44: Keeping the original scripts around as a fallback for now. I want to use the Changesets GitHub action to handle next week's release, but if anything goes awry we can still easily release manually.

.changeset/config.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@2.0.0/schema.json",
+  "changelog": ["@changesets/cli/changelog", { "repo": "remix-run/remix" }],
+  "commit": false,
+  "fixed": [
+    [
```
chaance, line 6: Every package will remain version-fixed for now *except* for `create-remix`, which is never installed as a dependency and should generally be releasable independently.

kentcdodds:
🦋

PR #3571, opened 2022-06-25
fix(remix-dev/cli/init): always remove `remix.init` folder

MichaelDeBoey:
This is important if people run `npx remix init` themselves, like suggested in https://github.com/remix-run/indie-stack/pull/112, https://github.com/remix-run/blues-stack/pull/89 & https://github.com/remix-run/grunge-stack/pull/76


mcansh:
that's fair i think, but can we maybe some kind of `--dry-run` flag to not delete it?

MichaelDeBoey:
@mcansh having a `--dry`/`--dry-run` flag would mean we shouldn't run `remix.init` as well, but log what's going to be changed instead.

I think a `--no-delete`/`--no-remove` flag would be better a better name in this case.

kentcdodds (approved):
Yep 👍 Thanks! Adding a `--no-delete` flag or something could be helpful for testing, but let's get this merged now.

PR #3573, opened 2022-06-25
feat(remix-dev/cli/init): add `--no-delete` flag

MichaelDeBoey:
Follow-up of #3571

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -56,7 +56,13 @@ export async function create({
   spinner.clear();
 }
 
-export async function init(projectDir: string) {
+type InitFlags = {
+  deleteScript?: boolean;
+};
+export async function init(
+  projectDir: string,
+  { deleteScript = true }: InitFlags = {}
+) {
   let initScriptDir = path.join(projectDir, "remix.init");
   let initScript = path.resolve(initScriptDir, "index.js");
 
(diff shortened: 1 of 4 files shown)


packages/remix-dev/cli/commands.ts
```diff
@@ -56,7 +56,13 @@ export async function create({
   spinner.clear();
 }
 
-export async function init(projectDir: string) {
+type InitFlags = {
+  deleteScript?: boolean;
```
machour, line 60: just for clarity, should this be renamed to `deleteInitScript` ?

packages/remix-dev/cli/run.ts
```diff
@@ -376,7 +379,7 @@ export async function run(argv: string[] = process.argv.slice(2)) {
       if (hasInitScript) {
         if (installDeps) {
           console.log("💿 Running remix.init script");
-          await commands.init(projectDir);
+          await commands.init(projectDir, { deleteScript: true });
```
machour, line 382: This isn't really needed as this is the default

machour (approved):

packages/remix-dev/cli/commands.ts
MichaelDeBoey: We're calling the `init` command, so it's obvious that we're talking about the init script I think.

packages/remix-dev/cli/run.ts
MichaelDeBoey: Indeed not necessary, but I like the explicitness.

mcansh (approved):

kentcdodds (approved):
Super! Thanks 👍

kentcdodds:
Rats. I took too long reviewing. Could you handle the merge conflict?

kentcdodds (approved):
Thanks!

PR #3582, opened 2022-06-27
chore(issue-template): Add a section about the OS/Setup/etc.

machour:
When doing triage, I often find myself asking about that kind of things.

I'm sure there's room for improvement, just thought I'd start this and merge your comments/ideas so that we can get all the required bits to understand the issue being reported.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -27,6 +27,12 @@ body:
       required: true
   - type: textarea
     attributes:
+      label: Your environment
+      description: Tell us more about your OS, setup, package manager, etc.
+    validations:
+      required: true
+  - type: textarea
+    attributes:
       label: Steps to Reproduce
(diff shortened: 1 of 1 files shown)


MichaelDeBoey (commented):
What do you think about separate sections with a dropdown for OS & package manager?

What do you mean by "setup" exactly?
This probably needs some clarification.

MichaelDeBoey:
> Do you have a complete list of packages managers ?

Only ones I can think of is the ones we use as well: `npm`, `pnpm` & `yarn`
Probably a good idea to also let people tell us the version of their package manager as well.

> We definitely want a better wording here.

Some explanation or better wording would indeed help to make it clearer I think.

kentcdodds (changes requested):
Thanks! If we're going to add this, let's make what we're asking for more explicit.

machour:
Meh.

mcansh:
we could do something like this, but add the right flags to make it not so verbose https://github.com/tabrindle/envinfo

PR #3588, opened 2022-06-28
fix: do more tree-shaking in dev mode

jacob-ebey:
The type of dead code elimination we want to do depends on the minify syntax property: https://github.com/evanw/esbuild/issues/672#issuecomment-1029682369. Without this we have dev builds that run fine in production, but ship server code / dependencies to the browser, blowing it up due to not optimizing out imports and module level blocks like `if (process.env.NODE_ENV === "test") {`.

This will fix the issue seen by those trying to integrate react-three-fiber into their app, as well as enable a good vitest in-source testing story.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

This is covered by existing integration tests.

### DIFF
--- a/packages/remix-dev/compiler.ts
+++ b/packages/remix-dev/compiler.ts
@@ -442,6 +442,12 @@ function createServerBuild(
       format: config.serverModuleFormat,
       treeShaking: true,
       minify: options.mode === BuildMode.Production && isCloudflareRuntime,
+      // The type of dead code elimination we want to do depends on the
+      // minify syntax property: https://github.com/evanw/esbuild/issues/672#issuecomment-1029682369
+      // without this we have dev builds that run fine in production, but
+      // ship server code to the browser, blowing it up due to not optimizing
+      // out imports and module level blocks like `if (process.env.NODE_ENV === "test") {`.
+      minifySyntax: true,
       mainFields: isCloudflareRuntime
         ? ["browser", "module", "main"]
         : config.serverModuleFormat === "esm"


MichaelDeBoey (approved):

packages/remix-dev/compiler.ts
```diff
@@ -442,6 +442,12 @@ function createServerBuild(
       format: config.serverModuleFormat,
       treeShaking: true,
       minify: options.mode === BuildMode.Production && isCloudflareRuntime,
+      // The type of dead code elimination we want to do depends on the
+      // minify syntax property: https://github.com/evanw/esbuild/issues/672#issuecomment-1029682369
+      // without this we have dev builds that run fine in production, but
+      // ship server code to the browser, blowing it up due to not optimizing
+      // out imports and module level blocks like `if (process.env.NODE_ENV === "test") {`.
```
pcattori, line 449: I took a stab at some alternate wording that I think flows better and makes more sense to me, but feel free to take or leave as much of this wording as you think is best.

```suggestion
      // For dev builds, we need dead code elimination (beyond what's provided by `treeShaking: true`) so that
      // we don't ship server code to the browser and eliminate non-dev blocks (e.g. `if (process.env.NODE_ENV === "test")`).
      // To do so, we enable the `minifySyntax` option, which "does more advanced dead code elimination during syntax compression".
      // See https://github.com/evanw/esbuild/issues/672#issuecomment-1029682369
```

packages/remix-dev/compiler.ts
```diff
@@ -442,6 +442,12 @@ function createServerBuild(
...
+      minifySyntax: true,
```
brophdawg11, line 450: Total nitpick, but I might suggest putting `minifySyntax: true` right below `treeShaking: true` since both are targeting the same end goal (removing code) and both are non-conditional.  Then `minify` comes third and builds on top of `minifySyntax`

brophdawg11:
@jacob-ebey Do you have any quick and easy test cases that we could run locally to see the difference here?  Something hopefully less involved than making a `react-three-fiber` sample app :)

chaance (approved):

jacob-ebey:
@brophdawg11 Throw a `if (process.env === "test")` block in a route module scope and inspect the bundle. Without this you will have `if (false)` and the body of the block left in the bundle.

PR #3587, opened 2022-06-28
fix: pin bahmutov/npm-install version

jenseng:
v1.8.16 breaks cache miss installs on windows, which blocks any PRs/commits that change dependencies. Opened [this issue](https://github.com/bahmutov/npm-install/issues/146) to address the underlying bug, but in the meantime we can pin to the prior version to get things working again.

Making this PR against `main` instead `dev`, since the jobs explicitly use `remix-run/remix/.github/workflows/reusable-test.yml@main`, but I can switch it up if y'all like. Just trying to get my 4-month-old PR unblocked yet again 😆😭 

Additional context:
- Discord: https://discord.com/channels/770287896669978684/940264701785423992/991085961972707338
- Initial PR affected: https://github.com/remix-run/remix/pull/2027
- Repro/Debug PR: https://github.com/remix-run/remix/pull/3584

Testing Strategy:
* Validated that v1.8.15 fixes cache miss installs on windows [here](https://github.com/remix-run/remix/pull/3584)

PR #3593, opened 2022-06-28
fix: remove bahmutov/npm-install to fix windows builds

jenseng:
We already use `actions/setup-node@v3` which has equivalent caching functionality that isn't broken for yarn+windows (or yarn 2+ across the board). So just use that wherever possible. In addition to fixing builds, this has the added benefit of slightly speeding them up, since we won't double-cache things (in some cases we had both caches enabled).

Making this PR against `main` instead `dev`, since the jobs explicitly use `remix-run/remix/.github/workflows/reusable-test.yml@main`, but I can switch it up if y'all like. Just trying to get my 4-month-old PR unblocked yet again 😆😭

For context, see https://github.com/bahmutov/npm-install/issues/146, in particular [this comment](https://github.com/bahmutov/npm-install/issues/146#issuecomment-1169226551) and linked test output.

This is an alternative to #3587 or otherwise waiting for `bahmutov/npm-install` to get fixed.

### DIFF
--- a/.github/workflows/deployments.yml
+++ b/.github/workflows/deployments.yml
@@ -44,6 +44,8 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version-file: ".nvmrc"
+          cache: npm
+          cache-dependency-path: ./scripts/deployment-test/package.json # no lockfile, key caching off package.json
 
       # some deployment targets require the latest version of npm
       # TODO: remove this eventually when the default version we get
@@ -53,10 +55,8 @@ jobs:
         working-directory: ./scripts/deployment-test
 
       - name: 📥 Install deployment-test deps
-        uses: bahmutov/npm-install@v1
-        with:
-          working-directory: ./scripts/deployment-test
-          useLockFile: false
+        run: npm install
+        working-directory: ./scripts/deployment-test
 
       - name: 🚀 Deploy to Arc
         run: node ./arc.mjs
@@ -81,6 +81,8 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version-file: ".nvmrc"
+          cache: npm
+          cache-dependency-path: ./scripts/deployment-test/package.json # no lockfile, key caching off package.json
 
       # some deployment targets require the latest version of npm
       # TODO: remove this eventually when the default version we get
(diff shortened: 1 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -17,6 +17,6 @@ on:
 jobs:
   test:
     if: github.repository == 'remix-run/remix'
-    uses: remix-run/remix/.github/workflows/reusable-test.yml@main
+    uses: jenseng/remix/.github/workflows/reusable-test.yml@fix-yarn-caching
```
jenseng, line 20: will remove, this is just to help validate the PR :)

.github/workflows/test.yml
```diff
@@ -17,6 +17,6 @@ on:
 jobs:
   test:
     if: github.repository == 'remix-run/remix'
-    uses: remix-run/remix/.github/workflows/reusable-test.yml@main
+    uses: jenseng/remix/.github/workflows/reusable-test.yml@pin-npm-install-workflow-step
```
jenseng, line 20: converted to a draft due to this testing chicken/egg problem ... temporarily setting this so the change actually gets tested :)

.github/workflows/test.yml
jenseng, line 20: converted to a draft due to this testing chicken/egg problem ... temporarily setting this so the change actually gets tested :)

jenseng: tests passed using this workflow definition: https://github.com/remix-run/remix/actions/runs/2579571547, i've gone ahead and removed this tweak

chaance (approved):

MichaelDeBoey (commented):
@jenseng Do we want to revert this once https://github.com/bahmutov/npm-install/issues/146 is fixed?

If not, can we also remove it in the stacks workflow?
Or is there another reason why we can't replicate the more fine grained control ourselves?

jenseng:
> @jenseng Do we want to revert this once [bahmutov/npm-install#146](https://github.com/bahmutov/npm-install/issues/146) is fixed?
> 
> If not, can we also remove it in the stacks workflow? Or is there another reason why we can't replicate the more fine grained control ourselves?

My preference would be not to revert, as this commit generally improves things IMO (fewer 3rd-party action steps, no redundant caching, more explicit about when yarn vs npm is used, etc)

I'll take a closer look to see if we can remove it from the stacks workflow. The challenge is that we want a reliable key for busting the cache, and the ideal thing for that is either a lockfile or package.json, which we don't have until later in those jobs (i.e. once an archive is unzipped or create-remix is run). We could key of something else, but then the risk is we bust the cache too often (meaning more frequent downloads of all dependencies) or not often enough (meaning more frequent downloads of changing dependencies). Not a dealbreaker, just makes the caching less effective.

MichaelDeBoey:
@jenseng Maybe we should have a look at how `bahmutov/npm-install` is handling this?

jenseng:
@MichaelDeBoey this does the trick: https://github.com/remix-run/remix/pull/3610

PR #2417, opened 2022-03-21
chore(docs,examples,templates): quote glob run params in `build` command

CanRau:
Just ran into https://github.com/mysticatea/npm-run-all/issues/200 using Yarn 3

Quoting it "dev": "run-p 'dev:*'", like mentioned in the linked issue fixed it for me with Yarn 3 and also tested Yarn 1.22.10 and npm 8.1.0, just to make sure.

So I figured as it shouldn't break anything why not always quote it?

Ref https://github.com/mysticatea/npm-run-all/issues/200

---
As @machour  requested that I fix the conflicts in https://github.com/remix-run/remix/pull/2033 but I felt it was to much mental overhead right now to get my head around fixing the other PR 🤯  I created a new one, renaming my branch and PR and also realized that I missed `run-s` commands containing globs as well 🤦🏻‍♂️

PR #3595, opened 2022-06-28
fix(docs,examples,templates): make `npm-run-all` calls work when using Yarn

lensbart:
`run-p dev:*` [doesn’t](https://github.com/mysticatea/npm-run-all/issues/196) (but probably should) work with Yarn. Thankfully, this can be fixed [quite easily](https://github.com/mysticatea/npm-run-all/issues/196#issuecomment-893470240) by wrapping the argument in quotes.

(As an aside, I find single quotes a bit cleaner than escaped double quotes, but I saw that escaped quotes were already used elsewhere so I stuck to the convention.)


MichaelDeBoey (changes requested):
We already had #2417 to start working on this.
Since it didn't get any response in the last 25 days, I'd rather get this one merged instead.

1 thing I want to do here too though is also update all `run-s` calls as well.

I also prefer using single quotes tbh, as that makes it a bit more easy on the eyes (and has a lower cognitive load as well, since you don't need to think about where you are in the script & if you have escaped all of the correctly)

lensbart:
@MichaelDeBoey consider it done!

MichaelDeBoey (approved):

kentcdodds (changes requested):
I'm pretty sure single quotes make this not work on windows. I've never heard of needing to quote these npm-run-all usages for Windows support 🤔

lensbart:
@kentcdodds fixed

MichaelDeBoey (approved):

kentcdodds (approved):
Thanks!

PR #3597, opened 2022-06-28
chore(remix-dev): improve compiler with lodash

gervasiocaj:
Improves usage of the lodash library on the remix-dev compiler.
I believe we can start applying some functional programming ideas to improve the testability of this package. Please comment if there are any ongoing efforts on this.
Tested this with the following script:

```sh
# yarn
# npx playwright install
yarn test:primary --selectProjects dev
yarn test:integration
```


machour:
Hi @gervasiocaj and thank you for this PR. 
Can you please sign the CLA by adding your name to the contributors?

MichaelDeBoey:
@gervasiocaj Can you please explain a bit why this change would be beneficial?

This seems just like a personal preference to me & I don't think we should merge this PR then.
Otherwise we could have PRs that keep moving this back & forth.

packages/remix-dev/package.json
```diff
@@ -43,7 +43,7 @@
     "jsesc": "3.0.2",
     "json5": "^2.2.1",
     "lodash": "^4.17.21",
-    "lodash.debounce": "^4.0.8",
+    "meow": "^7.1.1",
```
machour, line 46: This change doesn't look right

gervasiocaj: Yes, this was a mishap

```suggestion
```

gervasiocaj:
@MichaelDeBoey the main issue at first was the `ImportsCache` implementation, which was not clear enough to me. Since `lodash `was already a dependency and had a tested implementation of `memoize`, my reasoning was to use it.

I also took the liberty of removing the `lodash.debounce` dependency as it is duplicate code, and changing the function call to the more readable FP style instead of having the trailing delay argument.

As for performance, I considered [this article](https://bretcameron.medium.com/how-javascript-maps-can-make-your-code-faster-90f56bf61d9d) as a reference, but as of now, [this response](https://stackoverflow.com/a/49164774/2282532) and yours made me realize that I have to consider the nature of the keys (pathnames) and how impactful it will be.

Please let me know what you think, and thank you for the feedback so far.

packages/remix-dev/compiler/assets.ts
```diff
@@ -112,47 +113,35 @@ export async function createAssetsManifest(
...
+    let route = routes[routeId];
 
-  if (route.parentId) {
-    parentImports = parentImports.concat(
-      optimizeRouteImports(route.parentId, routes, parentImports, importsCache)
-    );
-  }
+    if (route.parentId) {
+      let optimized = optimizeRouteImports(route.parentId, routes, parentImports);
+      parentImports = parentImports.concat(optimized);
```
MichaelDeBoey, line 131: Let's keep this as it was
```suggestion
      parentImports = parentImports.concat(
        optimizeRouteImports(route.parentId, routes, parentImports)
      );
```

packages/remix-dev/compiler/assets.ts
```diff
@@ -112,47 +113,35 @@ export async function createAssetsManifest(
...
+    }
 
-  let routeImports = (route.imports || []).filter(
-    (url) => !parentImports.includes(url)
-  );
+    let urlMissingInParentImports = (url: string) => !parentImports.includes(url);
+    let routeImports = (route.imports || []).filter(urlMissingInParentImports);
```
MichaelDeBoey, line 135: Same
```suggestion
    let routeImports = (route.imports || []).filter(
      (url) => !parentImports.includes(url)
    );
```

packages/remix-dev/compiler/assets.ts
```diff
@@ -112,47 +113,35 @@ export async function createAssetsManifest(
...
-function optimizeRouteImports(
-  routeId: string,
-  routes: AssetsManifest["routes"],
-  parentImports: string[],
-  importsCache: ImportsCache
-): string[] {
-  if (importsCache[routeId]) return importsCache[routeId];
-
-  let route = routes[routeId];
+  let optimizeRouteImports = memoize(function (
```
MichaelDeBoey, line 122: Let's keep this function outside of `optimizeRoutes`

MichaelDeBoey (changes requested):
@gervasiocaj I'm in favor of removing the `@types/lodash.debounce` & `lodash.debounce` dependencies.
I do however think we should keep the normal `lodash` functions instead of the FP functions.

I also think we should keep the implementation of `optimizeRoutes` & `optimizeRouteImports` as close to the original as possible

kentcdodds (commented):
Thanks for this contribution, but I think we're happy with how it is. It's nice and simple as-is. Thanks anyway.

PR #3601, opened 2022-06-29
docs(blog-tutorial): Fix link to posts in Blog Admin route

natomato:
While working through the developer-blog tutorial, the post links in the /posts/admin route were broken. They worked when the route was changed to not be relative.

Screen Shot 2022-06-28 at 8 40 59 PM

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -272,6 +272,7 @@
 - mush159
 - na2hiro
 - nareshbhatia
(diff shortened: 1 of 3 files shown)


kentcdodds (changes requested):
Hi @natomato,

Thanks for this. The extra credit creates those routes. It's correct as-is.

Issue #3600, opened 2022-06-29
Node v18.3 Network Regression


justsml:
### What version of Remix are you using?

latest

### Steps to Reproduce

1. Install Node v18.3.0
2. Run `npm run dev`

### Expected Behavior

Server starts.

### Actual Behavior

Your server fails to start with `Could not find an IPv4 address.`

This is due to a regression in Node v18.3 [`os.networkInterfaces()`](https://github.com/nodejs/node/blob/49362efd5b39f31a14260612422109e130068d08/lib/os.js#L251).

While it has already been patched in v18.4, version `18.3` consistently causes this.

**References:** 

- [Broken network `family` detection](https://github.com/nodejs/node/blob/49362efd5b39f31a14260612422109e130068d08/lib/os.js#L251)
- [v18.4 patch from 1-2 weeks ago](https://github.com/nodejs/node/commit/70b516e4dbdfdfea2183995ec3b368dacaa38183#diff-f52f7c022afbe0147ec297ea168db37d7e5aa3a611f18d952c5ddb85ae689059R131)

PR #3602, opened 2022-06-29
fix(remix-dev,remix-serve): fix network `family` detection

justsml:
Closes: #3600

This patch addresses a specific regression in Node v18.3 by making the detection of IPv4 addresses more durable.

**References:** 

- [Broken network `family` detection](https://github.com/nodejs/node/blob/49362efd5b39f31a14260612422109e130068d08/lib/os.js#L251)
- [v18.4 patch from 1-2 weeks ago](https://github.com/nodejs/node/commit/70b516e4dbdfdfea2183995ec3b368dacaa38183#diff-f52f7c022afbe0147ec297ea168db37d7e5aa3a611f18d952c5ddb85ae689059R131)

- [x] Docs - n/a
- [x] Tests - n/a

Testing Strategy:

See reproduction steps in #3600

### DIFF
--- a/packages/remix-serve/cli.ts
+++ b/packages/remix-serve/cli.ts
@@ -24,7 +24,7 @@ let onListen = () => {
     process.env.HOST ||
     Object.values(os.networkInterfaces())
       .flat()
-      .find((ip) => `${ip?.family}`.includes("4") && !ip.internal)?.address;
+      .find((ip) => String(ip?.family).includes("4") && !ip?.internal)?.address;
 
   if (!address) {
     console.log(`Remix App Server started at http://localhost:${port}`);
--- a/packages/remix-serve/cli.ts
+++ b/packages/remix-serve/cli.ts
@@ -24,7 +24,7 @@ let onListen = () => {
     process.env.HOST ||
     Object.values(os.networkInterfaces())
       .flat()
-      .find((ip) => ip?.family === "IPv4" && !ip.internal)?.address;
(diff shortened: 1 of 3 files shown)


packages/remix-serve/cli.ts
```diff
@@ -24,7 +24,7 @@ let onListen = () => {
     process.env.HOST ||
     Object.values(os.networkInterfaces())
       .flat()
-      .find((ip) => ip?.family === "IPv4" && !ip.internal)?.address;
+      .find((ip) => `${ip?.family}`.includes("4") && !ip.internal)?.address;
```
MichaelDeBoey, line 27: I think this is a bit cleaner

```suggestion
      .find((ip) => String(ip?.family).includes("4") && !ip.internal)?.address;
```

justsml: Thanks for the suggestion @MichaelDeBoey - Done ✅ 🤘

mcansh (changes requested):
LGTM, can we make this change to https://github.com/remix-run/remix/blob/dev/packages/remix-dev/cli/commands.ts#L294 as well?

justsml:
Done @mcansh - good catch!

mcansh (approved):

MichaelDeBoey (approved):

justsml:
Awesome! Thanks for the review and merge @mcansh @MichaelDeBoey @machour ✨ ❤️ 🚀

PR #3611, opened 2022-06-29
fix(remix-react): prevent form data loss by appending submitter's value in `useSubmit` (`<Form>`)

nrako:
Closes:  (the following issue)

> Currently, remix-react's `<Form>` through `useSubmit` (`useSubmitImpl`) 
    implementation, includes the submitter value by setting it on the
    `formData`. Unfortunetly, this may override any existing inputs having the
    same name than the submitter; resulting in data loss.

- [x] ~~Docs~~ N/A
- [x] Tests

Testing Strategy:

```jsx
<Form>
  <input name="tasks" value="first" />
  <button name="tasks value="" type="submit">Add Task</button>
</Form>
// should submit `?tasks=first&tasks=` but submit `?tasks=`
```

See integration [Bug-Report-Test](https://github.com/remix-run/remix/blob/41870303f049c1fd8b7eec6d14777e31b44587b6/integration/bug-report-test.ts) via commit: https://github.com/remix-run/remix/commit/41870303f049c1fd8b7eec6d14777e31b44587b6

---

**test: failing test with submitter value not appended to form data**
41870303f &lt;Nicholas Rakoto&gt; Wed, 29 Jun 2022 22:30:59 +0200

    Currently, remix-react's `<Form>` through `useSubmit` (`useSubmitImpl`) 
    implementation, includes the submitter value by setting it on the
    `formData`. Unfortunetly, this may override any existing inputs having the
    same name than the submitter; resulting in data loss.

    Instead, the submitter's value should be appended to the `formData`.

**fix(remix-react): prevent form data loss by appending submitter's value**
2a7025077 &lt;Nicholas Rakoto&gt; Wed, 29 Jun 2022 22:34:39 +0200

    When submitting a form the submitter's value must be appended to the
    `formData` and not have it's value "set" or pre-existing data under the same
    name may be overidden resulting in data loss.

    This commit fixes the `useSubmit() hook (`useSubmitImpl`) and consequently
    it's indirect usage through the `<Form>` component.

### DIFF
--- a/integration/hook-useSubmit-test.ts
+++ b/integration/hook-useSubmit-test.ts
@@ -1,0 +1,64 @@
+import { test, expect } from "@playwright/test";
+
+import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
+import type { Fixture, AppFixture } from "./helpers/create-fixture";
+import { PlaywrightFixture } from "./helpers/playwright-fixture";
+
+test.describe("`useSubmit()` returned function", () => {
+  let fixture: Fixture;
+  let appFixture: AppFixture;
+
+  test.beforeAll(async () => {
+    fixture = await createFixture({
+      files: {
+        "app/routes/index.jsx": js`
+        import { useLoaderData, useSubmit } from "@remix-run/react";
+
+        export function loader({ request }) {
+          let url = new URL(request.url);
+          return url.searchParams.toString()
+        }
+
+        export default function Index() {
+          let submit = useSubmit();
+          let handleClick = event => {
+            event.preventDefault()
+            submit(event.nativeEvent.submitter || event.currentTarget)
+          }
+          let data = useLoaderData();
+          return (
+            <form>
+              <input type="text" name="tasks" defaultValue="first" />
+              <input type="text" name="tasks" defaultValue="second" />
+
+              <button onClick={handleClick} name="tasks" value="third">
+                Prepare Third Task
+              </button>
+
+              <pre>{data}</pre>
+            </form>
+          )
+        }
+      `,
+      },
+    });
+
+    appFixture = await createAppFixture(fixture);
+  });
+
+  test.afterAll(async () => {
+    await appFixture.close();
+  });
+
+  test("submits the submitter's value appended to the form data", async ({
+    page,
+  }) => {
+    let app = new PlaywrightFixture(appFixture, page);
+    await app.goto("/");
+    await app.clickElement("text=Prepare Third Task");
+    await page.waitForLoadState("load");
+    expect(await app.getHtml("pre")).toBe(
+      `<pre>tasks=first&amp;tasks=second&amp;tasks=third</pre>`
+    );
+  });
+});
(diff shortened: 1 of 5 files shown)


nrako:
May somebody point me to a doc/thread or let me know what I can expect for the review process?  (cc @MichaelDeBoey ?)

It seems this PR has been "seen", but not reviewed. CI hasn't been ran though... And unfortunately nobody seems to be assigned for reviews. If that's encouraged I can specifically mention maintainers which would seem most appropriate for review by looking at the git history.

I've also opened another PR #3613 fixing issues also touching the `<Form>` which will end up with conflicting code on the tests that I can of course resolve once any of them is hopefully merged – or of course a maintainers can do it too, the necessary permissions have been granted.

I'd be eager to help, both fixes are imho quite pertinent for the `<Form>` (or `useSubmit` in this case).

packages/remix-react/components.tsx
```diff
@@ -1138,7 +1138,7 @@ export function useSubmitImpl(key?: string): SubmitFunction {
 
         // Include name + value from a <button>
         if (target.name) {
-          formData.set(target.name, target.value);
+          formData.append(target.name, target.value);
```
kentcdodds, line 1141: I'm going to swap this temporarily just to make sure the tests fail without this change. Thanks for your patience :)

```suggestion
          formData.set(target.name, target.value);
```

integration/hook-useSubmit-test.ts
```diff
@@ -0,0 +1,61 @@
...
+        import { useLoaderData, useSubmit } from "@remix-run/react";
+
+        export function loader({ request }) {
+          let url = new URL(request.url);
+          return url.searchParams.toString()
+        }
+
+        export default function Index() {
+          let submit = useSubmit();
+          let handleClick = event => submit(nativeEvent.submitter || e.currentTarget)
```
kentcdodds, line 24: Where does `nativeEvent` come from?

kentcdodds:
Looking into this further, this is a behavioral change that I think @ryanflorence will want to look at. I'll let him know.

integration/hook-useSubmit-test.ts
nrako: Oups, thanks. I've fixed that. The test was passing because the `<form>` was natively submitting with the GET method.

nrako: Amended again to call `event.preventDefault()` to only have the behavior of `submit()` being tested. I've double-checked the test and it now test what it meant to.

nrako:
@kentcdodds, @ryanflorence – any change to have a follow up on this? To be honest I don't think this really represent a  behavioral change. IMHO the current implementation is the one deviating from what would be the expected behavior.

kiliman:
I agree, if Remix deviates from default browser behavior, then it's a bug, and this PR resolves it. I would be surprised if anyone is dependent on the buggy behavior.

I think it would be rare to see somebody using the same *name* in a button as for an existing input. But since the browser support this, I am +1 on the fix.

PR #3626, opened 2022-06-30
fix(remix-express,remix-vercel): fix `request.signal`

brophdawg11:
This PR fixes the signal we assign to our Fetch `Request` in Express + Vercel.  We used to abort the signal on the `close` event on the Express/Vercel `Request`, but in `POST`'s this happens to be as soon as we've consumed the body, so any async flow in an `action` causes the signal to be aborted:

```js
export async function action({ request }) {
    await new Promise(r => setTimeout(r, 0));
    console.log(request.signal.aborted); // This is true every time
    ...
}
```

Instead we should we aborting our server-side processing of the request as soon as we are no longer able to send a `Response`, so we can listen for the Express/Vercel `Response` `close` event instead.

**Open Questions:** 
1. I added an integration test to assert that the `signal` is not immediately aborted, but curious if we have a way to assert that it's correctly aborted if the `response` closes?  
   * Jacob and I manually tested this by sending a `fetch` request from the browser and manually triggering its `AbortController` - and then checking the console logs on the Remix server.  Unsure if we can get at that level of info via our integration harness?
2. Other runtimes
   * CF + Deno provide a `signal` on the incoming `request` so we rely on them to trigger abort the signal properly on their end
   * Arc/Netlify don't seem to have any `signal` available, and we do not create one so `request.signal == null` inside our Remix code.  This will potentially be problematic when we pull in the new router since it will be relying on the existing signal.  We can either just make our references defensive via `request.signal?.aborted`, but it would be better if we could find some way to wire up a legit `signal`?  
     * I chose to create a `signal` (which will never be aborted) for these and added an `invariant(request.signal)` in `createStaticHandler.query` to better align with the `Request` typings


packages/remix-express/server.ts
```diff
@@ -89,20 +89,21 @@ export function createRemixHeaders(
...
-  req.on("close", () => {
-    controller.abort();
-  });
+  res.on("close", () => controller.abort());
 
   let init: NodeRequestInit = {
     method: req.method,
     headers: createRemixHeaders(req.headers),
-    signal: controller.signal,
+    signal: controller.signal as AbortSignal,
```
brophdawg11, line 106: It seems our the `abort-controller` polyfill is missing `reason` and `throwIfAborted` props :(

https://github.com/mysticatea/abort-controller/issues/33
https://github.com/mysticatea/abort-controller/issues/36

I wonder if there's an alternative we could look for?  We don't reference either of those fields at the moment .... but this feels like a ticking time bomb since our types tell us we _could_

integration/abort-signal-test.ts
```diff
@@ -0,0 +1,61 @@
...
+test.afterAll(async () => appFixture.close());
+
+test("should not abort the request in a new event loop", async ({ page }) => {
+  let app = new PlaywrightFixture(appFixture, page);
+  await app.goto("/");
+  expect(await app.getHtml(".action")).toMatch("empty");
+  expect(await app.getHtml(".loader")).toMatch("false");
+
+  await app.clickElement('button[type="submit"]');
+  expect(await app.getHtml(".action")).toMatch("false");
```
brophdawg11, line 59: Prior to the fix this would be `true`

brophdawg11:
Hm - seems to be failing CI on a `yarn install` - will try kicking off a new build tomorrow:

```
error An unexpected error occurred: "https://registry.yarnpkg.com/@types/eslint/-/eslint-8.4.4.tgz: Request failed \"404 Not Found\"".
```

jacob-ebey (approved):

PR #3632, opened 2022-07-01
chore: fix internal watch command (upgrading rollup)

alisd23:
Closes: (No issue yet)

- [ ] Docs
- [ ] Tests

Testing Strategy:
**For maintainers/contributers only**

When running the top level `watch` script in the remix monorepo, currently the watch does not work correctly. The `tsc` part of the build pipeline does not re-run when files change, so the watch command is pointless at the moment.

This is because the `--watch.onEnd` (and others) arguments only work from rollup version [2.72.0]([v2.72.0](https://github.com/rollup/rollup/releases/tag/v2.72.0)), and the current version is set to `2.57.0` in the lockfile.

Simply upgrading to the latest version fixes this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "remark": "14.0.2",
     "remark-frontmatter": "4.0.1",
     "remark-gfm": "3.0.1",
-    "rollup": "^2.36.1",
+    "rollup": "^2.75.7",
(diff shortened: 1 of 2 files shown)


ronnylt:
https://github.com/remix-run/remix/pull/3376 already upgraded rollup to the latest version.

Issue #209, opened 2021-06-25
Modules with .server.js becomes different empty files on client build


sergiodxa:
When importing a module with the `.server` prefix they are replaced with something like this:

```js
import {
  __commonJS,
  init_react
} from "/build/_shared/chunk-DOM5QJLA.js";

// empty-module:../services/auth0.server
var require_auth0 = __commonJS((exports, module) => {
  init_react();
  module.exports = {};
});

export {
  require_auth0
};
```

If I import more than one server-only module and go to the browser I see the same amount of different JS files with an almost identical content (specially on prod builds) which are all being loaded in parallel.

**Ideal**:

If the import can't be removed at least they are replaced with a single placeholder empty module that's only loaded once and shared by all modules importing a server-only module.

chaance:
@sergiodxa Combing through old issues, is this one still relevant?

rphlmr:
Hi,
Coming here by chance, looking for info about these empty modules.
I can conform that every empty module is a downloaded chunk (in production, too). 

![image](https://user-images.githubusercontent.com/20722140/184331896-432e4cef-2a18-4665-a192-77a30efc6574.png)
![image](https://user-images.githubusercontent.com/20722140/184331952-5a73caf1-3332-46d9-a483-0119943f20fa.png)

Is it expected ?

PR #3638, opened 2022-07-02
feat(remix-dev/compiler): remove need for `"sideEffects": false`

jacob-ebey:
This PR removes the need for `"sideEffect": false` in your package.json by resolving all bare modules not handled by another plugin as `sideEffect: false` automatically. I also updated it so that we don't get empty chunks sent to the browser when using the `xxx.server.ts` convention.

With these changes, this app with zero code changes, just a bump to the experimental goes from 112kb down to 66.2kb and from 10 .js files down to 7.

Screen Shot 2022-07-01 at 7 47 36 PM

Screen Shot 2022-07-01 at 7 48 45 PM

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

Runs a compilation against a module with a side effect and checks to see if it's in the output bundle. The new test was failing before applying this PR.


ctrlplusb:
I experienced some issues where the `"sideEffects": false` flag removed a polyfill from my `server.ts`. Why is there a requirement for this to be there in the first place? Sorry I am unfamiliar with the compiler. Is there any documentation around this need / decision?

machour:
@jacob-ebey is this work still relevant? It would close https://github.com/remix-run/remix/issues/209

PR #3639, opened 2022-07-02
chore(remix-dev): fix flaky tests, make CLI more testable

jenseng:
Rather than `process.exit(1)`, throw an error so we don't kill the jest worker. Also increase the timeout for two tests that often go more than 10 seconds (especially on Windows). The CLI entrypoint already exits appropriately if there's an error, it just needs a little tweak to better handle "expected" errors.

Because these tests are slow (running yarn, etc.), they would often time out while jscodeshift was still running, causing the afterEach callbacks to delete things underneath it and transforms to fail. `failOnError: true` would then cause jscodeshift to call `process.exit(1)`, which would then kill the current process (the jest worker). Depending on the exact timing, you'd see one or more of the following in the test output:

* A bunch of lines like:
    ```
    ERR C:/Users/RUNNER~1/AppData/Local/Temp/remix-tests-1179c2p2ev8/replace-remix-imports/app/db.server.ts File error: Error: ENOENT: no such file or directory, open 'C:\Users\RUNNER~1\AppData\Local\Temp\remix-tests-1179c2p2ev8\replace-remix-imports\app\db.server.ts'
    ERR C:/Users/RUNNER~1/AppData/Local/Temp/remix-tests-1179c2p2ev8/replace-remix-imports/app/entry.client.tsx File error: Error: ENOENT: no such file or directory, open 'C:\Users\RUNNER~1\AppData\Local\Temp\remix-tests-1179c2p2ev8\replace-remix-imports\app\entry.client.tsx'
    ERR C:/Users/RUNNER~1/AppData/Local/Temp/remix-tests-1179c2p2ev8/replace-remix-imports/app/entry.server.tsx File error: Error: ENOENT: no such file or directory, open 'C:\Users\RUNNER~1\AppData\Local\Temp\remix-tests-1179c2p2ev8\replace-remix-imports\app\entry.server.tsx'
    ```
* A test failure like:
    ```
        thrown: "Exceeded timeout of 10000 ms for a test.
        Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
    ```
* Jest aborting altogether:
    ```
      ●  process.exit called with "1"
          45 |
          46 |   try {
        > 47 |     let { error } = await jscodeshift(transformPath, files, options);
    ```

This PR focuses solely on the migration code/tests, but it's probably worth moving away from `process.exit(1)` elsewhere in the CLI.

Testing Strategy:
* Broken behavior was diagnosed/debugged and the fix was validated here:
  https://github.com/jenseng/remix/actions?query=branch%3Afix-jscodeshift-tests
  In particular, test run # 13 shows the directory structure as it gets mutated underneath jscodeshift by the afterEach
* Manually validated CLI still works normally and a CliError behaves as expected

### DIFF
--- a/packages/remix-dev/__tests__/migrations/convert-to-javascript-test.ts
+++ b/packages/remix-dev/__tests__/migrations/convert-to-javascript-test.ts
@@ -43,6 +43,7 @@ const TEMP_DIR = join(
   `remix-tests-${Math.random().toString(32).slice(2)}`
 );
 
+jest.setTimeout(30_000);
 beforeEach(() => {
   output = "";
   console.log = mockLog;
--- a/packages/remix-dev/__tests__/replace-remix-imports-test.ts
+++ b/packages/remix-dev/__tests__/replace-remix-imports-test.ts
@@ -38,6 +38,7 @@ const TEMP_DIR = path.join(
   `remix-tests-${Math.random().toString(32).slice(2)}`
 );
 
+jest.setTimeout(30_000);
 beforeEach(async () => {
   output = "";
   console.log = mockLog;
--- a/packages/remix-dev/cli.ts
+++ b/packages/remix-dev/cli.ts
@@ -1,11 +1,13 @@
-import { cli } from "./index";
+import { cli, CliError } from "./index";
 
 cli.run().then(
   () => {
     process.exit(0);
   },
-  (error: Error) => {
-    console.error(error);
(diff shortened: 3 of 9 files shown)


machour:
@jenseng could you update this PR to fix conflicts? 🙏🏼

chaance (approved):

PR #3642, opened 2022-07-02
fix: Update the installation step for the jokes app

shashankboosi:
Update the docs

Closes: #

- [x] Docs
- [ ] Tests

Testing Strategy: It is a documentation update, so tested it by checking the preview

### DIFF
--- a/docs/tutorials/jokes.md
+++ b/docs/tutorials/jokes.md
@@ -88,23 +88,12 @@ If you're planning on using CodeSandbox, you can use [the Basic example][the-bas
 npx create-remix@latest
 ```
 
-<docs-info>
-
-This may ask you whether you want to install `create-remix` to run the command. Enter `y`. It will only be installed temporarily to run the setup script.
-
-</docs-info>
-
-When the fun Remix animation is finished, it'll ask you a few questions. We'll call our app "remix-jokes", choose "Just the basics", then the "Remix App Server" deploy target, use TypeScript, and have it run the installation for us:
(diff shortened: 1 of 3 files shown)


docs/tutorials/jokes.md
```diff
@@ -88,23 +88,12 @@ If you're planning on using CodeSandbox, you can use [the Basic example][the-bas
 npx create-remix@latest
 ```
 
-<docs-info>
```
machour, line 91: I think this info block needs to be kept, as the user will be prompted as described.

shashankatencompass: I was checking it on Windows and it wasn't happening to me but changed it to Linux and looks like it will be prompted the first time a dev tries to install/run the command `npx create-remix@latest` and the following installs doesn't have the prompt. Will update it now, thanks for raising it

docs/tutorials/jokes.md
```diff
@@ -90,21 +90,16 @@ npx create-remix@latest
 
 <docs-info>
 
-This may ask you whether you want to install `create-remix` to run the command. Enter `y`. It will only be installed temporarily to run the setup script.
+This may ask you whether you want to install the following packages `create-remix@latest`. To run the command, enter `y`. It will only be installed the first time to run the setup script.
```
kentcdodds, line 93:
```suggestion
This may ask you whether you want to install `create-remix@latest`. Enter `y`. It will only be installed the first time to run the setup script.
```

kentcdodds (approved):
There we go :) Thanks!

PR #3656, opened 2022-07-05
fix(serverBareModulesPlugin): update require.resolve to use full import, update warning

mcansh:

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

### DIFF
--- a/.changeset/yellow-flowers-trade.md
+++ b/.changeset/yellow-flowers-trade.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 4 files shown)


kentcdodds (changes requested):
Thanks! I think we should probably have a test for this functionality (not just the warning, but that too). Could you implement that?

mcansh:
> Thanks! I think we should probably have a test for this functionality (not just the warning, but that too). Could you implement that?

i agree, in my initial PR i wasn't sure about the best way to test for that being that a missing dependency would fail the build, but i'll look into how others are doing that detection

Issue #3652, opened 2022-07-04
 The engine "node" is incompatible with this module when `npm install` on node `14.5.4`


cliffordfajardo:
### What version of Remix are you using?
1.6.3

### Steps to Reproduce
1. Run `npm install` (or `yarn install`) on a remix project using node version `14.15.4`
![CleanShot 2022-07-04 at 17 25 35@2x](https://user-images.githubusercontent.com/6743796/177228097-4ab85aaf-4dd7-4e03-8900-279474862273.png)

Output
```
error @remix-run/eslint-config@1.6.3: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "14.15.4"
error Found incompatible module.
```

### Expected Behavior
installing dependencies (npm install..) should not fail as the minimum node version recommended in remix docs is node 14
![CleanShot 2022-07-04 at 17 02 35@2x](https://user-images.githubusercontent.com/6743796/177226847-ec714291-0e08-4a8b-8a08-a45ac4e57ab6.png)
The

cliffordfajardo:
EDIT:
Eslint has `"node": "^12.22.0 || ^14.17.0 || >=16.0.0"` hard coded in their package.json 🤔 😢 
- https://github.com/eslint/eslint/blob/main/package.json#L164

cliffordfajardo:
Sure - Ill specify that you need to be on either of those 3 node versions at this time since its a hard requirement of eslint

![CleanShot 2022-07-05 at 16 55 02@2x](https://user-images.githubusercontent.com/6743796/177435446-e610e732-5eec-4e22-803a-d5b153152d46.png)

morad-labrid:
upgrade Node by the commande line (bow upgrade node)
its work for me !

PR #3658, opened 2022-07-06
fix(docs/tutorials): update pre-requisite Node version

cliffordfajardo:
## Summary of Changes
- update pre-requisite node version requirements for docs.
- `eslint` version 8 requires the following node version (^12.22.0, ^14.17.0, or >=16.0.0)

Closes: #3652

## Motivation for Changes
- `npm install / yarn install` failure 
  image

- `eslint` v8 requires the 3 versions specified above
   image

- [X] Docs
- [ ] Tests

Testing Strategy:
- N/A docs changes only

### DIFF
--- a/docs/tutorials/blog.md
+++ b/docs/tutorials/blog.md
@@ -19,7 +19,7 @@ Click this button to create a [Gitpod][gitpod] workspace with the project set up
 
 If you want to follow this tutorial locally on your own computer, it is important for you to have these things installed:
 
-- [Node.js][node-js] 14 or greater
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (commented):
I don't think this is necessary, as the errors you're referring to come from using ESLint.

If somebody would remove ESLint usage, Remix would still work fine.

MichaelDeBoey:
@kentcdodds I think we should revert this PR _or_ at least remove the Node v12 occurance

PR #3661, opened 2022-07-06
ci: explicitly pass `github.token` to actions

mcansh:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey (commented):
I don't think this is necessary, as you're passing the same value as the default value now.

mcansh:
> I don't think this is necessary, as you're passing the same value as the default value now.

same, i'm just looking for anything that would be causing us to hit the api rate limit during releases, which sounds like for github.token (and custom tokens) is 1000 per hour

PR #3662, opened 2022-07-06
ci(deploy/vercel): pass token via `--token`

mcansh:
i don't think vercel deploy allows for the environment variable, instead requiring the flag to be passed, but other vercel commands allow it 🤷‍♂️ https://github.com/vercel/vercel/blob/695bfbdd60a97030457ea1861c5bb89bcfe92353/packages/cli/src/index.ts#L439, so I'll keep in spawn's `env`.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey (approved):

kentcdodds (approved):
Thanks!

PR #3665, opened 2022-07-06
fix(remix-react): set fetch referrer to document.referrer

### LINKED ISSUE #3612: Disparity in HTTP referer header between <form> and <Form>

1.6.3

### Steps to Reproduce

Create a <form> whose action submits to a resource route. Something like 

```html
<form action="/resource/route">
  <button>Submit</button>
</form>
```

In the resource route, read the `referer` header:

```js
export async function action({ request }) {
  console.log(request.headers.get('referer'));
  return null
}
```

Try submitting the form and reading the header.

Then change the <form> to <Form> and submit again.

The referer changes from the page containing the form to the page the form is submitting to.

The default form behavior is helpful for something like this: `return redirect(request.headers.get('referer'), 303);`

### Expected Behavior

The referer should be the same as the URL containing the form, as that is there the request originated from.

### Actual Behavior

The referer (when using <Form>) is set to the form's action.

mcansh:
though you really shouldn't use `referer` willynilly when redirecting without verifying it's value, but at least it's consistent with the browser now 😆

~~doesnt really explain why `fetcher.Form` was working though...~~

i think this is due to to `handleActionFetchSubmission` using the actual match, whereas `handleActionSubmissionNavigation` uses the leaf match 

Closes #2110
Closes #3612

- [ ] Docs
- [x] Tests

Testing Strategy:

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -152,6 +152,7 @@
 - isaacrmoreno
 - ishan-me
 - IshanKBG
+- itsMapleLeaf
 - jacob-ebey
 - JacobParis
 - jakewtaylor
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -70,6 +70,10 @@ test.beforeAll(async () => {
         export default function Index() {
           return <div>cheeseburger</div>;
         }
+        
+        export default function Render() {
+            return <span />;
+        }
       `,
     },
   });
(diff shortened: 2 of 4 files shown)


machour (approved):
:chefkiss:

MichaelDeBoey (approved):

mcansh:
closing as this will be fixed by the upcoming React Router 6.4 changes to fetching :) https://github.com/remix-run/remix/pull/2110#issuecomment-1180806310

PR #3631, opened 2022-07-01
docs(api/conventions): Fix broken routing guide link

manan30:
Closes: #3630

- [x] Docs
- [ ] Tests

Testing Strategy: Local testing for docs

PR #3667, opened 2022-07-07
Fix routing guide link

ronnylt:


machour:
Thank you @ronnylt for your contribution, however this was already fixed in `main` by #3631

Contributions to the documentation must always be targeting `main`.

PR #3674, opened 2022-07-07
fix(remix-react): An error in `MetaFunction` or `LinkFunction` should render `ErrorBoundary`

### LINKED ISSUE #3140: An error in `MetaFunction` or `LinkFunction` is not caught by `ErrorBoundary`

1.4.1

### Steps to Reproduce

Create a MetaFunction with an error inside:

```ts
export let meta: MetaFunction = () => {
  throw new Error("boom")
}
```

Or create a LinkFunction:

```ts
export let links: LinkFunction = () => {
  throw new Error("boom")
}
```

This can happen if, for example, you access a property of data that doesn't exists.

### Expected Behavior

It should show the ErrorBoundary from the route or parent route.

### Actual Behavior

It shows a `Unexpected Server Error` white screen with black text and the error message below.

Screen Shot 2022-05-09 at 11 30 19

mcansh:

Closes #3140

- [ ] Docs
- [ ] Tests

Testing Strategy:

- these tests cover the changed code: [error-boundary-test.ts](https://github.com/remix-run/remix/blob/logan/rem-1298-meta-link-error-boundary/integration/error-boundary-test.ts)
- I also opened up a new playground to make sure it all worked there as well.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -385,11 +385,12 @@ export function Links() {
     if (root.ErrorBoundary) {
       renderableMatches.push({
         params: {},
-        pathname: "",
+        pathname: "/",
         pathnameBase: "/",
         route: {
           id: "root",
-          hasErrorBoundary: !!root.ErrorBoundary,
+          path: "/",
+          hasErrorBoundary: true,
         },
       });
     }
@@ -611,11 +612,12 @@ function V1Meta() {
     if (root.ErrorBoundary) {
       renderableMatches.push({
         params: {},
-        pathname: "",
+        pathname: "/",
         pathnameBase: "/",
         route: {
           id: "root",
-          hasErrorBoundary: !!root.ErrorBoundary,
+          path: "/",
+          hasErrorBoundary: true,
         },
       });
     }
@@ -729,11 +731,12 @@ function V2Meta() {
     if (root.ErrorBoundary) {
       renderableMatches.push({
         params: {},
-        pathname: "",
+        pathname: "/",
         pathnameBase: "/",
         route: {
           id: "root",
-          hasErrorBoundary: !!root.ErrorBoundary,
+          path: "/",
+          hasErrorBoundary: true,
         },
       });
     }
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -390,7 +390,9 @@ export function Links() {
         route: {
           id: "root",
           path: "/",
-          hasErrorBoundary: true,
+          hasErrorBoundary: !!root.ErrorBoundary,
+          handle: root.handle,
+          shouldRevalidate: root.shouldRevalidate,
         },
       });
     }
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -22,8 +22,7 @@
   ],
   "scripts": {
     "bug-report-test": "playwright test --config ./integration/playwright.config.ts integration/bug-report-test.ts",
-    "build": "rollup -c && tsc -b",
-    "postbuild": "node scripts/copy-build-to-dist.mjs",
```
kentcdodds, line 26: Was this change intentional?

packages/remix-react/links.ts
```diff
@@ -244,7 +244,12 @@ export async function prefetchStyleLinks(
   routeModule: RouteModule
 ): Promise<void> {
   if (!routeModule.links) return;
-  let descriptors = routeModule.links();
+  let descriptors;
+  try {
+    descriptors = routeModule.links();
+  } catch (error) {
+    throw error;
+  }
```
kentcdodds, line 252: Same here... What am I missing?

kentcdodds (changes requested):
I think I must be missing something here, but I don't see any behavioral changes here. I'm pretty sure the code does exactly the same thing before and after.

package.json
mcansh: yes, otherwise type checking could fail due to tsc using outdated files - but i can move that to a dedicated PR

packages/remix-react/components.tsx
```diff
@@ -696,10 +696,17 @@ export function Meta() {
...
+      if (typeof routeModule.meta === "function") {
+        try {
+          routeMeta = routeModule.meta({ data, parentsData, params, location });
+        } catch (error) {
+          throw error;
+        }
+      } else {
+        routeMeta = routeModule.meta;
+      }
+
```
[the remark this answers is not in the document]
mcansh: indeed, trying the playground again after `yarn clean` on `dev` and it's working as expected, but when i initially did this it was experiencing the same issue that the issue has 🤨

mcansh:
little more digging, looks like if you have a custom ErrorBoundary we render the generic `Unexpected Server Error` 
image


but if you don't have one, it renders the default RemixErrorBoundary
image

packages/remix-server-runtime/server.ts
```diff
@@ -470,6 +477,32 @@ async function handleDocumentRequest({
   } catch (error: any) {
     responseStatusCode = 500;
 
+    // next up, we'll verify that `meta` returned without error
+    try {
+      await callRouteMeta({
```
brophdawg11, line 482: I wonder if there's any use in distinguishing errors thrown when calling a route `meta()` function here from errors thrown when rendering the `<meta>` tags.  The rendering is far more in our control and we should be able to be pretty confident it won't throw render errors.  The exported `meta` function contents are out of our control - but we _do_ control the execution of it in our own `<Meta>` component.

Spitballing here but what if we did something like this in `<Meta>` 

```js
for (let match of matches) {
  let routeId = match.route.id;
  try {
    let routeMeta = routeModule.meta({ data, parentsData, params, location })
    ...
  } catch (error) {
    // Re-throw the error and associate it with the routeId that threw it
    throw new MetaError(routeId, error);
  }
}
```

Then we could identify the error in this `handleDocumentRequest` `catch` block as a meta-specific error (via `instanceof MetaError`) and know the `routeId` that caused it and look for the nearest ancestor boundary (i.,e., prevent re-calling the problematic `meta` on the second pass - but attempt to call the successful ones above it).

This is effectively what we're doing with the mutable boundaryIds (identifying how deep we got) - but we have no choice there since it's React calling them, not us.  `meta` isn't subject to that same restriction so we could be a bit more direct in the identification of the problematic `routeId`.

brophdawg11: Actually we'd still need some logic changes in `Meta` to do that too I think.  It doesn't have the concept of `renderableMatches` that we have for `getDocumentHeaders` which I think is what it would need to not call child's meta function on the second pass if we had identified parent as the renderable boundary.

brophdawg11:
Closing for now, can re-open if/when we come back to this.

PR #3675, opened 2022-07-07
feat(remix-dev): add suppport for `.mjs` and `.cjs` configs

jacob-ebey:
Allows `remix.config.js` to be written as ESM or CJS following package.json `type` field conventions for resolution.

I.e:

If package.json `"type": "module"`
- `.js` loaded as ESM
- `.mjs` loaded as ESM
- `.cjs` loaded as CJS

If package.json is _not_ `"type": "module"`
- `.js` loaded as CJS
- `.mjs` loaded as ESM
- `.cjs` loaded as CJS

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

Existing functionality covered by existing tests.


chaance (approved):

kentcdodds:
Something weird going on with the tests?

kentcdodds:
Looks like we maybe should include a test or two for this and fix the tests that are broken in this PR

MichaelDeBoey (commented):
What about supporting `.ts` config files?

chaance (approved):

jimniels:
For anybody wondering, I was on remix `1.6.0` and came here after upgrading to `1.7.4` because I was getting this error message:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module … not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/jimnielsen/Sites/jim-nielsen.com/readlists/remix/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

I had `type: "module"` in my parent `package.json` but my `remix.config.js` file was a common JS file (it had `module.exports = {}`). Everything was working in my `1.6.0` setup (which was based off the template generator for a remix project) but it stopped working somewhere jumping to `1.7.4`

I had two choices, both of which worked:

1. Rename `remix.config.js` to `remix.config.cjs`
2. Change my config from `module.exports = {…}` to `export default {…}`

Personally, I went with no. 2 and my project started to work again.

PR #3676, opened 2022-07-07
feat(remix-react): add magic comment to allow `webpack` bundling

jacob-ebey:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-react/routeModules.ts
+++ b/packages/remix-react/routeModules.ts
@@ -125,7 +125,7 @@ export async function loadRouteModule(
   }
 
   try {
-    let routeModule = await import(route.module);
+    let routeModule = await import(/* webpackIgnore: true */ route.module);
     routeModulesCache[route.id] = routeModule;
     return routeModule;
   } catch (error) {


kentcdodds (approved):
I hate we have to do this.

kentcdodds:
To make sure I understand, we know that the route module will be there because we make our own manifest rather than relying on webpack's built-in code splitting mechanism?

jacob-ebey:
> To make sure I understand, we know that the route module will be there because we make our own manifest rather than relying on webpack's built-in code splitting mechanism?

Without this, webpack attempts to bundle / create dynamically loadable chunks for the import. Since it can't resolve the variable to a path on disk, webpack ultimately will just throw an error at runtime saying something like "this chunk wasn't able to be bundled".

With this comment webpack doesn't try to bundle that import at all and just resolves whatever string is passed to it at runtime using dynamic import instead of the webpack module graph resolution algorithm.

kentcdodds:
Right. What I was trying to make sure I understood is why it works to leave the dynamic import in place :) I think I get it. Thanks!

Issue #1898, opened 2022-02-11
Defining multiple routes with same route module


gijsroge:
### What version of Remix are you using?

1.1.3

### Steps to Reproduce

As explained in react router v6 https://reactrouter.com/docs/en/v6/faq#what-happened-to-regexp-routes-paths to create optional parameters you can define multiple paths with the same route module file.

But this doesn't work in Remix. The last defined route overrides the previous one.

run `npx create-remix@latest`

remix.config.js
```
module.exports = {
  appDirectory: "app",
  assetsBuildDirectory: "public/build",
  devServerPort: 8002,
  publicPath: "/build/",
  serverBuildDirectory: "build",
  ignoredRouteFiles: [".*"],
  routes(defineRoutes) {
    return defineRoutes(route => {
      route("/user/:id", "routes/index.tsx");
      route("/user", "routes/index.tsx");
    });
  }
};
```

### Expected Behavior

Both `user/:id` & `user/` should render `routes/index.tsx`

### Actual Behavior

only `user/` resolves, `user/:id` returns a 404.

jacob-ebey:
This is something I'd like as well, the issue currently is we derive the route ID base on the file-path of the module, meaning the same module can't be shared for multiple route definitions. One thing I've considered is allowing another param to the `route()` method that accepts a unique user provided ID instead of deriving it from the module filepath.

yomeshgupta:
I had somewhat similar issue while working on [https://devtools.tech](https://devtools.tech/?ref=rr-mrsrmi). I wanted to show the same file/component on two different routes. As @jacob-ebey mentioned the route ID is based on file path so you cannot include the same file twice.

One solution I could think was abstract out the contents into a third component and import it into two different routes.

```js
// Component.js
...

export const links;
export const meta;
export default Component;
```

```js
// route /blog

import Component, { links as importedLinks, meta as importedMeta } from './Component';

export const links = importedLinks;
export const meta = importedMeta;

// we can directly export the component or wrap it if we want to pass or do anything extra.
export default Component;
```

```js
// route /resource

import Component, { links as importedLinks, meta as importedMeta } from './Component';

export const links = importedLinks;
export const meta = importedMeta;

// we can directly export the component or wrap it if we want to pass or do anything extra.
export default Component;
```

I even tried defining a route in the routes folder like `[blog|resource].jsx` but as far as I remember that didn't work. 😅

gijsroge:
[...]

PR #3684, opened 2022-07-08
test(remix-dev): add failing test for 'multiple routes with the same route module' #1898

AndrewLeedham:
Closes: #1898

- [ ] Docs
- [x] Tests

Testing Strategy:

I added a failing test to show that multiple routes pointing at a single route module get overwritten.


machour:
@AndrewLeedham can you please sign the CLA so that we can run the failing test?
Also consider removing the draft state if you're done :)

AndrewLeedham:
@MichaelDeBoey I tried fixing this, but it seems like changing IDs would be a breaking change. Where would be best to discuss this problem?

PR #3697, opened 2022-07-11
fix(remix-react): Preserve search string in form action when `action` prop is omitted

chaance:
When a `Form` action is omitted, the default value is `"."` which resolves to the current route without search params. This is intended behavior when `"."` is passed to `useResolvedPath`, but Remix docs specify that when an action is omitted that the default action is the current location. Users would expect that to include the search string.

Closes: #927
Closes: #931

- [ ] Docs
- [x] Tests

Testing Strategy:

Added tests to the existing form integration test file.


MichaelDeBoey (approved):

mcansh (approved):

ryanflorence:
> Remix docs specify that when an action is omitted that the default action is the current location

The docs are wrong then 🙈. If `action` is ommitted, the default action is `"."` which means the closest route in context, not the current location (just like `<Link to=".">`. Nested routing adds a new concept here (at the cost of some complexity).

Preserving search params seems like the right thing to do, but I'm trying to think about the symmetry here:


**Default Browser/HTML Behavior**

| element | preserves search |
|---------|-------------------------|
| `<a href=".">` | ❌ |
| `<form method="get" action=".">` | ❌ |
| `<form method="get">` | ✅ (technically "yes", the `form.action` includes the search but the form fields wipe it out when submitted, so practically "no") |
| `<form method="post" action=".">` | ❌ |
| `<form method="post">` | ✅ |

https://rv02u9.csb.app/foo/?bar#baz

So:

- `.` means `location.pathname` only
- omitted `action` means `location.pathname + location.search + location.hash`


**What `Form` should do**

The new concept with nested routes is that `.` means "closes route in context` instead of `location.pathname`.

The problem seems to be that we are defaulting `action` to `.`, which is inconsistent with browser behavior. We should default it to the closest route in context (the new concept) but then be consistent with the browser from there and use `route.path + location.search + location.hash`.

The changes on this PR treat `.` and an omitted action the same, which I don't think is what we want to do. Instead we should do this:

| element | preserves search |
|---------|-------------------------|
| `<Link to=".">` | ❌ |
| `<Form method="get" action=".">` | ❌ |
| `<Form method="get">` | ✅ |
| `<form method="post" action=".">` | ❌ |
| `<form method="post">` | ✅ |

I expect the change to be pretty simple.

- Make `action` optional, don't include a default "." anymore.
- If there is no action, default to the `resolvedRoute.path + location.search + location.hash`

integration/form-test.ts
```diff
@@ -534,6 +666,26 @@ test.describe("Forms", () => {
...
 
     test.describe("in a splat route", () => {
+      test("no action resolves relative to the closest route", async ({
+        page,
+      }) => {
+        let app = new PlaywrightFixture(appFixture, page);
+        await app.goto("/projects/blarg");
+        let html = await app.getHtml();
+        let el = getElement(html, `#${SPLAT_ROUTE_NO_ACTION}`);
+        expect(el.attr("action")).toMatch("/projects");
```
brophdawg11, line 676: I think given Ryan's explanation of the splat behavior with "." - these 2 tests should assert the path `/projects/blargh` now

packages/remix-react/components.tsx
```diff
@@ -999,20 +1000,31 @@ type HTMLFormSubmitter = HTMLButtonElement | HTMLInputElement;
...
+  let location = useLocation();
+  let { search, hash } = resolvedPath;
+  if (action === undefined) {
+    search = location.search;
+    hash = location.hash;
+  }
 
-  if (action === "." && isIndexRoute) {
+  let isIndexRoute = id.endsWith("/index");
+  if ((action === undefined || action === ".") && isIndexRoute) {
```
brophdawg11, line 1023: super edge case, but would it make sense to use `(action ?? ".") === "."` here to ensure it's identical logic as the above?  not expected but I think passing `action=null` would result in weird behavior currently since it would match above but not here

brophdawg11: Kinda hate the way this looks though haha.  Maybe `let resolvedAction = action ?? "."` above?

```suggestion
  if (((action ?? ".") === ".") && isIndexRoute) {
```

chaance: @brophdawg11 haven't you heard? Everyone uses TS now so that would never happen 😛

brophdawg11: all your types are belong to us

brophdawg11 (approved):

PR #3698, opened 2022-07-11
docs(guides/typescript): fix code highlighting

vmosyaykin:
I noticed that the line highlight in the Typescript guide is misplaced:

image

The `package.json` template was updated in e9692e11, but the hightlight stayed on the old line number, this PR fixes that.

Other snippets of `package.json` and `tsconfig.json` in the docs are correct,

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -378,6 +378,7 @@
 - visormatt
 - vkrol
 - vlindhol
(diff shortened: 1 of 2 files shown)


machour (approved):
Thank you @vmosyaykin!

PR #3704, opened 2022-07-11
docs(decisions-0003): infer types for `useLoaderData` from `loader` type using generic

pcattori:


decisions/0003-infer-types-for-useloaderdata-from-loader-type-using-generic.md
```diff
@@ -0,0 +1,102 @@
...
+  return { hello: "world" }
+
+  // YES
+  return json({ hello: "world" })
+}
+```
+
+### Users can still provide non-inferred types
+
+Users can still provide their own non-inferred types for `useLoaderData` either with `as` or with a generic:
```
MichaelDeBoey, line 70: "Using `as`" is actually called "type casting"

```suggestion
Users can still provide their own non-inferred types for `useLoaderData` either by type casting (using `as`) or with a generic:
```

decisions/0003-infer-types-for-useloaderdata-from-loader-type-using-generic.md
```diff
@@ -0,0 +1,102 @@
...
+  let option2 = useLoaderData<MyLoaderData>() // TS error: Type 'string' is not comparable to type 'Date'.
+}
+```
+
+While fixing this footgun should be desirable in the vast majority of use-cases, users can always cast to `unknown` first as an escape hatch:
+
+```ts
+export default function Route() {
+  // NOT RECOMMENDED
+  let option1 = useLoaderData() as unknown as MyLoaderData
```
kentcdodds, line 100:
```suggestion
  let data = useLoaderData() as unknown as MyLoaderData
```

decisions/0003-infer-types-for-useloaderdata-and-useactiondata-from-loader-and-action-via-generics.md
```diff
@@ -0,0 +1,211 @@
...
+}
+```
+
+Again, the same goes for `useActionData`.
+
+### Solution criteria
+
+- Return type of `useLoaderData` and `useActionData` should somehow be inferred from `loader` and `action`, not blindly type cast
+- Return type of `loader` and `action` should be inferred
+  - Necessarily, return type of `json` should be inferred from its input
```
kentcdodds, line 110: Just want to add my criteria: https://github.com/remix-run/remix/pull/1254#issuecomment-1141421601

```suggestion
- Return type of `useLoaderData` and `useActionData` should somehow be inferred from `loader` and `action`, not blindly type cast
- Return type of `loader` and `action` should be inferred
  - Necessarily, return type of `json` should be inferred from its input
- No module side-effects (so higher-order functions like `makeLoader` is definitely a no).
- `json` should allow anything that `JSON.stringify` allows.
- `useLoaderData` should not return anything that `JSON.parse` can't return.
```

decisions/0003-infer-types-for-useloaderdata-and-useactiondata-from-loader-and-action-via-generics.md
```diff
@@ -0,0 +1,211 @@
...
+
+Passing in a non-inferred type for `useLoaderData` is hiding an unsafe type cast.
+Using the `useLoaderData` in this way will be deprecated in favor of an explicit type cast that clearly communicates the assumptions being made:
+
+```ts
+type MyLoaderData = {/*...*/}
+
+export default function Route() {
+  let dataGeneric = useLoaderData<MyLoaderData>() // <-- will be deprecated
+  let dataCast = useLoaderData() as MyLoaderData // <- use this instead
```
kentcdodds, line 185: This won't work with our v2 plans, so let's go ahead and recommend what would be needed in that eventuality?

```suggestion
  let dataCast = useLoaderData() as any as MyLoaderData // <- use this instead
```

kentcdodds (commented):
Solid 👍 This is good to merge with my suggestions IMO.

decisions/0003-infer-types-for-useloaderdata-and-useactiondata-from-loader-and-action-via-generics.md
pcattori: Our v2 plans are to return `unknown` in this case. You can type cast from `unknown` to any type without TS getting mad. So `useLoaderData() as MyLoaderData` will work for v2.

kentcdodds: Oh? That's great!

PR #3719, opened 2022-07-13
feat(remix-server-runtime): add `name` to `SerializedError`

jnicklas:
The name of the error is an important piece of information, which helps
in determining the cause of failures. Usually in Node, the name is also
part of the stack trace, so it may not be seen as completely necessary.
However, some tooling expects the name to be meaningful for diagnostics.
This change serializes the error name in the app state, making it
accessible for tooling which wants to introspect it.

- [ ] Docs (not necessary, I think?)
- [x] Tests

Testing Strategy:

Extending the existing tests to check the proper serialization of the error name, and that is properly obfuscated in production mode.

### DIFF
--- a/packages/remix-server-runtime/__tests__/server-test.ts
+++ b/packages/remix-server-runtime/__tests__/server-test.ts
@@ -3,6 +3,10 @@ import { ServerMode } from "../mode";
 import type { ServerBuild } from "../build";
 import { mockServerBuild } from "./utils";
 
+class CustomError extends Error {
+  name = "CustomError"
+}
+
 function spyConsole() {
   // https://github.com/facebook/react/issues/7047
   let spy: any = {};
@@ -421,7 +425,7 @@ describe("shared server runtime", () => {
 
     test("data request calls loader and responds with generic message and error header", async () => {
       let rootLoader = jest.fn(() => {
-        throw new Error("test");
+        throw new CustomError("test");
       });
       let testAction = jest.fn(() => {
         return "root";
@@ -444,8 +448,10 @@ describe("shared server runtime", () => {
       });
 
       let result = await handler(request);
+      let data = await result.json();
       expect(result.status).toBe(500);
-      expect((await result.json()).message).toBe("Unexpected Server Error");
+      expect(data.name).toBe("Error");
+      expect(data.message).toBe("Unexpected Server Error");
       expect(result.headers.get("X-Remix-Error")).toBe("yes");
       expect(rootLoader.mock.calls.length).toBe(1);
       expect(testAction.mock.calls.length).toBe(0);
@@ -455,7 +461,7 @@ describe("shared server runtime", () => {
       let message =
         "data request loader error logged to console once in dev mode";
       let rootLoader = jest.fn(() => {
-        throw new Error(message);
+        throw new CustomError(message);
       });
       let testAction = jest.fn(() => {
         return "root";
@@ -478,8 +484,10 @@ describe("shared server runtime", () => {
       });
 
       let result = await handler(request);
+      let data = await result.json();
       expect(result.status).toBe(500);
-      expect((await result.json()).message).toBe(message);
+      expect(data.name).toBe('CustomError');
+      expect(data.message).toBe(message);
       expect(result.headers.get("X-Remix-Error")).toBe("yes");
       expect(rootLoader.mock.calls.length).toBe(1);
       expect(testAction.mock.calls.length).toBe(0);
@@ -554,7 +562,7 @@ describe("shared server runtime", () => {
         return "root";
       });
       let testAction = jest.fn(() => {
-        throw new Error("test");
+        throw new CustomError("test");
       });
       let build = mockServerBuild({
         root: {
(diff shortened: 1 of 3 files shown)


machour:
Hi @jnicklas and thank you for this PR

Please target the `dev` branch instead of `main` as this is a code-change

jnicklas:
@machour no problem, I was not aware of the branching policy. I've changed it to use the `dev` branch.

jnicklas:
Anything I can do to move this forward?

packages/remix-server-runtime/__tests__/server-test.ts
```diff
@@ -444,8 +448,10 @@ describe("shared server runtime", () => {
       });
 
       let result = await handler(request);
+      let data = await result.json();
       expect(result.status).toBe(500);
-      expect((await result.json()).message).toBe("Unexpected Server Error");
+      expect(data.name).toBe("Error");
```
abenhamdine, line 453: shouldn't it be "CustomError" here ?

packages/remix-server-runtime/__tests__/server-test.ts
```diff
@@ -574,8 +582,10 @@ describe("shared server runtime", () => {
...
```
abenhamdine, line 587: same

packages/remix-server-runtime/__tests__/server-test.ts
jnicklas: These tests are checking that the error name is redacted in production mode, I think they are correct as is.

abenhamdine: ah yes, you're right,  I've been fooled by the name of the config `ServerMode.Test`

abenhamdine (approved):

packages/remix-server-runtime/__tests__/server-test.ts
MichaelDeBoey: Really strange that all tests are passing although this isn't correct 🤔

machour: @MichaelDeBoey do you still see something wrong here? 🤔

MichaelDeBoey: We're still checking for `Error` instead of `customError`, so yes I guess there's still something wrong

machour: It seems that `Error` is expected here, see comment above:

> These tests are checking that the error name is redacted in production mode, I think they are correct as is.

MichaelDeBoey: @jnicklas So why are we using `CustomError` in our tests if we're never checking anything from them? 🤔

alireza-bonab: There is `deserializeError` method too, if you don't handle that one it still will return Error instead of CustomError.

https://github.com/remix-run/remix/blob/fd00bc657ef7f2b55c53e5f9dbd77185fa7d5734/packages/remix-react/components.tsx#L224

alireza-bonab:
The issue is since it serializes and deserializes to Error, it is not possible to keep the typing and custom error format along the data flow from server to client.
The Error handling and ErrorBoundry handling require serious refactoring to be able to handle Custom Error types.

alireza-bonab:
you can use the [serialize-error](https://github.com/sindresorhus/serialize-error) library to handle the custom error serialize/deserialization.

jnicklas:
@alireza-bonab maybe a different way of serializing the error would be better, but I don't think we need to let perfect be the enemy of good. This is a concrete improvement on the status quo, that makes no breaking changes, and requires no deep refactoring, simply adding information which is currently missing. I think this is worth including until someone else takes a closer look at this. I personally don't have the time to spend on implementing a refactored serialization here.

alireza-bonab:
> @alireza-bonab maybe a different way of serializing the error would be better, but I don't think we need to let perfect be the enemy of good. This is a concrete improvement on the status quo, that makes no breaking changes, and requires no deep refactoring, simply adding information which is currently missing. I think this is worth including until someone else takes a closer look at this. I personally don't have the time to spend on implementing a refactored serialization here.

I understand but if you don't handle `deserializeError`, it won't work as expected at least for the `name`.

jnicklas:
@alireza-bonab I don't think there is anything in Remix that actually deserializes this error? I could be wrong, but I think this is just there for introspection? If you're saying that I need to take care of this on my end as a user, then that's fine, I only want to send this to an error reporting tool and for that purpose this solution works ok.

alireza-bonab:
> @alireza-bonab I don't think there is anything in Remix that actually deserializes this error? I could be wrong, but I think this is just there for introspection? If you're saying that I need to take care of this on my end as a user, then that's fine, I only want to send this to an error reporting tool and for that purpose this solution works ok.

https://github.com/remix-run/remix/blob/fd00bc657ef7f2b55c53e5f9dbd77185fa7d5734/packages/remix-react/components.tsx#L224

MichaelDeBoey:
Hi @jnicklas!

Issue #3712, opened 2022-07-12
unstable_shouldReload not working with deferred in some cases


rhefner:
### What version of Remix are you using?

0.0.0-experimental-d28d77b8

### Steps to Reproduce

[Codesandbox showing the issue](https://codesandbox.io/s/remix-deferred-unstable-shouldreload-issue-tf12d5)

Steps to see the potential issue:
- Go to "root" route
- Click on "Deferred" link
- Click the button in the `<nav />` to toggle and save the cookie
- Notice two fetches to /deferred:
```
/deferred?_data=root
/deferred?_data=routes%2Fdeferred
```

Steps to see `unstable_shouldReload` working (I think):
- Go to "/deferred" route directly
- Click on the cookie setter button
- Notice one fetch to /deferred:
```
/deferred?_data=root
```

### Expected Behavior

`unstable_shouldReload` should be respected in all cases - Whether or not it's the first route the user visits or subsequent ones

### Actual Behavior

`unstable_shouldReload` is not respected if the user doesn't go directly to the route

MichaelDeBoey:
Closed by #3721

brophdawg11:
Really digging the symmetry of `3721` closing `3712` - I thought it was a typo at first 😂

PR #3721, opened 2022-07-13
fix(remix-react): fix `shouldReload` bug for deferred cancellations on fetcher submissions

brophdawg11:
Closes: #3712

- [x] Tests

Testing Strategy: Unit test + confirmed this fixes the issue in the reproduction provided in the issue

### DIFF
--- a/packages/remix-react/__tests__/transition-test.tsx
+++ b/packages/remix-react/__tests__/transition-test.tsx
@@ -2401,7 +2401,7 @@ describe("deferred", () => {
           lazy: rootDfds[count].promise,
         });
       },
-      rootShouldReload: shouldReloadSpy,
+      shouldReloads: { root: shouldReloadSpy },
     });
 
     let A = t.navigate.get("/foo?key=value");
@@ -2607,7 +2607,7 @@ describe("deferred", () => {
           lazy: rootDfds[count].promise,
         });
       },
-      rootShouldReload: shouldReloadSpy,
+      shouldReloads: { root: shouldReloadSpy },
     });
 
     let A = t.navigate.get("/foo?a=1");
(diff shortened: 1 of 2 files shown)


packages/remix-react/__tests__/transition-test.tsx
```diff
@@ -2687,6 +2687,38 @@ describe("deferred", () => {
...
+    // Navigate to the same URL forcing a revalidation, which foo can opt out of
+    let B = t.fetch.post("/bar");
+    await B.action.resolve("FETCH ACTION");
+    expect(t.getState().loaderData).toEqual({
+      root: "ROOT",
+      foo: {
+        value: "A",
+      },
+    });
+    expect(shouldReloadSpy).toHaveBeenCalledTimes(1);
```
brophdawg11, line 2720: This is the main aspect here - `shouldReload` was never getting called on a fetcher submission _after_ a resolved deferred

packages/remix-react/transition.ts
```diff
@@ -626,17 +626,21 @@ export function createTransitionManager(init: TransitionManagerInit) {
...
         ]
-          .filter(
-            ([routeId]) =>
-              !cancelledDeferredRouteIds.has(routeId) ||
-              pendingNavigationDeferredControllers.has(routeId)
-          )
+          .filter(([routeId]) => {
+            let isActive = (updates.matches || state.matches).some(
+              (m) => m.route.id === routeId
+            );
```
brophdawg11, line 639: Added this check here to make sure that we only preserve `deferredLoaderData` for actively matched routes.  We were sneaking by ok previously because we were not removing from `pendingNavigationDeferredControllers` upon resolution, which meant that on subsequent abort's we would put it into `cancelledDeferredRouteIds` (even though it wasn't actually cancelled) and that would prevent us from carrying forward the `deferredLoaderData`.  

For example:
 - navigate `/foo` which returns a `deferred`
 - foo `deferred` completely resolved, but remains in `pendingNavigationDeferredControllers`
 - navigate `/bar`
 - foo is "cancelled" because it no longer matches
 - when /bar completes we do not retain foo because it was cancelled

Instead, we should not be marking foo as cancelled and should instead mark it as "not pending".  Then we need to identify that it's not active here and drop it's `deferredLoaderData`

packages/remix-react/transition.ts
```diff
@@ -2077,11 +2081,17 @@ async function makeDeferredLoaderData(
...
             update({
               deferredLoaderData: newDeferredLoaderData,
               loaderData: newLoaderData,
             });
+
+            if (
+              Object.keys(newDeferredLoaderData[match.route.id]).length === 0
+            ) {
+              // Mark this deferred as no longer pending once we've updated state
+              pendingNavigationDeferredControllers.delete(match.route.id);
```
brophdawg11, line 2093: Once a deferred has completely resolved - and we have no pending promises left (and therefore nothing to abort) we can remove it from `pendingNavigationDeferredControllers` so it doesn't think it got cancelled on subsequent actions or navigations

PR #3742, opened 2022-07-14
Defer manifest

ryanflorence:
w/o `defer` loading this script is a blocking resource, but it doesn't need to be. The scripts that need it are already `<script type=module>` which are automatically deferred. Deferred scripts are executed in order, so it's guaranteed to be available before the following scripts that need it are executed.

Testing Strategy: It's just `<script defer>` so we don't need to test that chromium defers correctly. Since all existing tests pass, we know it doesn't break apps.

### DIFF
--- a/.changeset/witty-emus-grab.md
+++ b/.changeset/witty-emus-grab.md
@@ -1,0 +1,8 @@
+---
+"remix": patch
+"@remix-run/deno": patch
+"@remix-run/dev": patch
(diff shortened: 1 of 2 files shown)


chaance:
@ryanflorence Since this PR was added, the script was updated with `async` instead of defer. Is that not correct behavior? Do we really need to defer execution or is simply not blocking sufficient here?

PR #3744, opened 2022-07-14
ci: move comments, deploy and stacks test to different action

mcansh:
currently they're running as part of the changeset action which runs on `push` which causes the ref to be the branch and not the actual release (also causes them to run more frequently)

- moved "postrelease" tasks like deployment tests, stack tests, and commenting to dedicated `postrelease.yml` workflow
- added a `packageVersionToFollow` property so we only follow `remix` for stable releases (nightlies only publish a single git tag, so we don't need this for that) (though maybe we could something a little bit differently so this continues to work when packages are not published in lock-step - group each pr into each affected package, then comment each affected packages name and version, maybe for another time though)
- when a release is stable, we'll now get all merged PRs for both the main and dev branches (missed this when we switched to using tags, but now we'll recomment on things that go stable)

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

https://github.com/mcansh/nightly-release-test


.github/workflows/release.yml
```diff
@@ -1,65 +1,12 @@
...
-    branches:
-      - release
-      - "release-*"
-      - "!release-experimental"
-      - "!release-experimental-*"
-      - "!release-manual"
-      - "!release-manual-*"
+    tags:
+      # only run on `remix` tags
+      - "remix@*"
```
mcansh, line 7: based on how branches work ("release-*"), this should work as the remix package's tag is "remix@1.6.5")

.github/workflows/changeset.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: 🚀 PR / Publish
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          version: yarn run changeset:version
+          commit: "chore: Update version for release"
+          title: "chore: Update version for release"
+          publish: yarn run changeset:release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mcansh, line 60: we're gonna need a custom GitHub token so that our "new" release workflow will run as by default a workflow will not cause another workflow to run

chaance (approved):

.github/workflows/release-comments.yml
```diff
@@ -1,11 +1,14 @@
-name: 📝 Comment on Release
+name: Comment on Release
```
mcansh, line 1:
```suggestion
name: 📝 Comment on Release
```

.github/workflows/release-comments.yml
```diff
@@ -1,11 +1,14 @@
...
 
 on:
   workflow_call:
     inputs:
       ref:
         required: true
         type: string
+      packageVersionToFollow:
+        required: false
```
mcansh, line 10: nightlies publish a single git tag, so we don't need to follow any specific package

.github/workflows/release-comments.yml
```diff
@@ -27,7 +32,7 @@ jobs:
         run: npm ci
         working-directory: ./scripts/release
 
-      - name: 📝 Comment on issues
+      - name: Comment on issues
```
mcansh, line 35:
```suggestion
- name: 📝 Comment on issues
```

.github/workflows/postrelease.yml
```diff
@@ -0,0 +1,45 @@
+name: 🕊 Release
```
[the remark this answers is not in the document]
mcansh: 🫡

PR #3745, opened 2022-07-14
docs(changelogs): Update verbiage on loader/action type changes

chaance:
This PR updates the description of our changes to the loader/action types. I'll cherry-pick this one into `main` since it's technically docs (sort of?), but it's probably best kept in sync on both branches.


kentcdodds (approved):
Stellar 👍

PR #3751, opened 2022-07-15
docs(index): Add a reference to /examples

machour:
I've noticed that we weren't pointing at the examples folder for new comers, so I added a link in the `Community` section.


kentcdodds (approved):
Great idea

PR #3766, opened 2022-07-16
fix(remix-react): support undefined unions as optional keys in `UseDataFunctionReturn` type

itsMapleLeaf:
Previously, objects with possibly undefined values became required. This PR makes it so they become optional instead, and preserves other behavior around undefined.
- Old: `{ a: number | undefined }` -> `{ a: number }`
- New: `{ a: number | undefined }` -> `{ a?: number }`

- [ ] Docs
- [x] Tests

Testing Strategy:

Added a new test to `packages/remix-react/__tests__/hook-types-test.tsx`, and ensured present type tests still pass.

### DIFF
--- a/packages/remix-react/__tests__/hook-types-test.tsx
+++ b/packages/remix-react/__tests__/hook-types-test.tsx
@@ -100,4 +100,14 @@ describe("type serializer", () => {
     type response = UseDataFunctionReturn<Loader>;
     isEqual<response, { arg: string }>(true);
   });
+
+  it("makes keys optional if the value is undefined", () => {
+    type AppData = {
+      arg1: string;
+      arg2: number | undefined;
+      arg3: undefined;
+    };
+    type response = UseDataFunctionReturn<AppData>;
+    isEqual<response, { arg1: string; arg2?: number }>(true);
+  });
 });
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1336,6 +1336,7 @@ type JsonPrimitives =
   | Boolean
   | null;
 type NonJsonPrimitives = undefined | Function | symbol;
+
 type SerializeType<T> = T extends JsonPrimitives
   ? T
   : T extends NonJsonPrimitives
@@ -1353,12 +1354,20 @@ type SerializeType<T> = T extends JsonPrimitives
   : T extends (infer U)[]
   ? (U extends NonJsonPrimitives ? null : SerializeType<U>)[]
   : T extends object
-  ? {
-      [k in keyof T as T[k] extends NonJsonPrimitives
-        ? never
-        : k]: SerializeType<T[k]>;
-    }
+  ? SerializeObject<UndefinedOptionals<T>>
   : never;
+
+type SerializeObject<T> = {
+  [k in keyof T as T[k] extends NonJsonPrimitives ? never : k]: SerializeType<
+    T[k]
+  >;
+};
+
+type UndefinedOptionals<T> = {
+  [k in keyof T as undefined extends T[k] ? never : k]: NonNullable<T[k]>;
+} & {
+  [k in keyof T as undefined extends T[k] ? k : never]?: NonNullable<T[k]>;
+};
 
 export type UseDataFunctionReturn<T extends DataOrFunction> = T extends (
   ...args: any[]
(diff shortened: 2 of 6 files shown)


packages/remix-react/__tests__/hook-types-test.tsx
```diff
@@ -100,4 +100,14 @@ describe("type serializer", () => {
     type response = UseDataFunctionReturn<Loader>;
     isEqual<response, { arg: string }>(true);
   });
+
+  it("makes keys optional if the value is undefined", () => {
+    type AppData = {
+      arg1: string;
+      arg2: number | undefined;
+      arg3: undefined;
```
pcattori, line 108: The alternative would be to have users who want optional properties to use Typescript optionals:

```suggestion
      arg2?: number;
      arg3?: undefined; // probably would just delete this param altogether
```

My concern is that Typescript _does_ distinguish between optional properties and properties that are unions with `undefined`, so if any users want to rely on standard Typescript behavior, we'd be taking that away from them.

Is this feature to help out more novice users who aren't aware of Typescript optionals? Or do you also see upside for users aware of Typescript optionals?

itsMapleLeaf: The `AppData` type in this test isn't something the user would write themselves. It represents the use case of passing something possibly undefined to the browser. Here's an example:

```ts
declare function getUser(): User | undefined

export function loader() {
  return json({ user: getUser() })
}

	const data = useLoaderData()
}
```

Before this PR, the type of `data` would become `{ user: User }`, which is incorrect. It should be `{ user?: User }`

The `arg3: undefined` is only here for completeness. If something is undefined, it can't be JSON-serialized, and doesn't show up in the object at all.

pcattori: I meant the difference between `{a?: T}` and `{a: T | undefined}` (note that `a` is not optional in the second type).

But yes, the `getUser` example is compelling. Especially since `undefined` is invalid JSON and gets dropped in `JSON.stringify`. That convinces me!

packages/remix-react/components.tsx
```diff
@@ -1353,13 +1354,21 @@ type SerializeType<T> = T extends JsonPrimitives
...
+  [k in keyof T as T[k] extends NonJsonPrimitives ? never : k]: SerializeType<
+    T[k]
+  >;
+};
+
+type UndefinedOptionals<T> = {
+  [k in keyof T as undefined extends T[k] ? never : k]: NonNullable<T[k]>;
+} & {
+  [k in keyof T as undefined extends T[k] ? k : never]?: NonNullable<T[k]>;
+};
```
pcattori, line 1370: Instead of an intersection (which IIRC shows up in intellisense as `{a: A} & {b: B}`, not as `{a: A, b: B}`), would we merge the two mapped objects:

```suggestion
type UndefinedOptionals<T> = Merge<{
  [k in keyof T as undefined extends T[k] ? never : k]: NonNullable<T[k]>;
},{
  [k in keyof T as undefined extends T[k] ? k : never]?: NonNullable<T[k]>;
}>;
```

You can use [`Merge` from `ts-toolbelt`](https://millsp.github.io/ts-toolbelt/modules/object_p_merge.html)

pcattori (changes requested):

packages/remix-react/components.tsx
itsMapleLeaf: Wow, actually really neat that it turns into a clean type. Although this seems like a separate general improvement we could apply to all of the complexly-merged types

pcattori: > Wow, actually really neat that it turns into a clean type. Although this seems like a separate general improvement we could apply to all of the complexly-merged types

True, but we won't switch them all over at once. It more like as we come across `&`s, then we switch them over. So let's start with this one!

Unless you see a good reason not to, of course.

MichaelDeBoey: I think it would be better to use [`type-fest`](https://github.com/sindresorhus/type-fest)'s [`Merge` type](https://github.com/sindresorhus/type-fest/blob/main/source/merge.d.ts) instead, as that's already one of our dependencies

itsMapleLeaf: Yeah, I was hesitant to add a new dependency :+1:

pcattori (approved):

Issue #3769, opened 2022-07-16
Instruct to restart the dev server after marked install in blog tutorial


Drew-Daniels:
### What version of Remix are you using?

1.6.5

### Steps to Reproduce

Follow the [Quickstart](https://remix.run/docs/en/v1/tutorials/blog) documentation, and after the step covering installing `marked` on line 576 of `docs/tutorials/blog.md`, validate that the markdown is not rendered into HTML onto the page when you are on a post-specific route (i.e., `http://localhost:3000/posts/90s-mixtape`).

### Expected Behavior

The tutorial makes it appear that just by installing the `marked` npm package, your development server will automatically then rebuild and show the markdown on the page (if you are currently navigated to a post slug page).

### Actual Behavior

It was not until shutting down and restarting my development server with `npm run dev` that I saw the markdown rendered into HTML upon navigating to a post route.

I think including a sentence explaining that users will need to restart their development server would help prevent any confusion around this step. It was not until a few manual page refreshes that I felt I might need to restart my dev server to see these changes.

kiliman:
True. But in defense, it is always advisable to restart processes when adding new packages via npm/yarn.

PR #3770, opened 2022-07-16
docs(tutorial/blog): Instruct to restart the dev server after marked install

Drew-Daniels:
Closes: #3769

- [X] Docs

### DIFF
--- a/docs/tutorials/blog.md
+++ b/docs/tutorials/blog.md
@@ -573,7 +573,7 @@ Quick note on that `invariant` for the params. Because `params` comes from the U
 
 We also have an invariant for the post. We'll handle the `404` case better later. Keep going!
 
-Now let's get that markdown parsed and rendered to HTML to the page. There are a lot of markdown parsers, we'll use "marked" for this tutorial because it's really easy to get working.
+Now let's get that markdown parsed and rendered to HTML to the page. There are a lot of markdown parsers, we'll use "marked" for this tutorial because it's really easy to get working. Please note that you will need to restart your server after installing.
 
 💿 Parse the markdown into HTML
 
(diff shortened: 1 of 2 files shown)


docs/tutorials/blog.md
```diff
@@ -573,7 +573,7 @@ Quick note on that `invariant` for the params. Because `params` comes from the U
 
 We also have an invariant for the post. We'll handle the `404` case better later. Keep going!
 
-Now let's get that markdown parsed and rendered to HTML to the page. There are a lot of markdown parsers, we'll use "marked" for this tutorial because it's really easy to get working.
+Now let's get that markdown parsed and rendered to HTML to the page. There are a lot of markdown parsers, we'll use "marked" for this tutorial because it's really easy to get working. Please note that you will need to restart your server after installing.
 
 💿 Parse the markdown into HTML
```
dca123, line 579: Could we leave the following message (or similar)  around line 585 ? This message is similar to the the once above when updating prisma. Leaving it after the installation also makes it easier to follow along !


>Now that marked has been installed, we will need to restart our server. So stop the dev server and start it back up again with `npm run dev`

dca123 (changes requested):

docs/tutorials/blog.md
Drew-Daniels: Ah, I totally agree! Thanks for the suggestion. Definitely makes more sense to have it placed there instead. I also like the idea of using a similar instruction to the prisma one for consistency. I'll send up those updates soon 👍🏼

machour (approved):
Great addition since this is the first tutorial for Remix, and we want to be as clear as possible.

kentcdodds (approved):
Great, thanks!

Issue #3762, opened 2022-07-16
Type error when inferring the type of `ReadonlyArray`s returned from `useLoaderData`


JNaftali:
### What version of Remix are you using?

1.6.5

### Steps to Reproduce

I've reproduced it as a minimal fork of the stackblitz remix starter - https://stackblitz.com/edit/node-d35hmw

1. return a `ReadonlyArray` from a loader
2. attempt to access that array using `useLoaderData<typeof loader>()`
3. observe the type error when attempting to use array methods

### Expected Behavior

The array should be typed correctly and methods such as `array.map` should be allowed. Either the array should come through typed as `Array` or as `ReadonlyArray`

### Actual Behavior

the object is typed as a bare object with numeric properties and lacking all array methods.

JNaftali:
cc @colinhacks since you might be able to quickly clarify if this is indeed a bug or just things working as intended - seems like not ideal behavior to me

kiliman:
If you use my `superjson` version, it is correctly typed:
<img width="783" alt="image" src="https://user-images.githubusercontent.com/47168/179362227-9d78aa65-963b-4545-bda3-877e454d1f0a.png">

https://stackblitz.com/edit/node-cx46b4?file=app%2Froutes%2Findex.tsx

machour:
Fixed in https://github.com/remix-run/remix/pull/3774 and released in 1.6.6

PR #3774, opened 2022-07-16
Allow readonly arrays in SerializeType

JNaftali:
Closes: #3762

- [ ] Docs
- [ ] Tests

Testing Strategy:

Patched `node_modules` in the package in the original repo and confirmed that it fixed the issue.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1350,7 +1350,7 @@ type SerializeType<T> = T extends JsonPrimitives
         ? null
         : SerializeType<T[k]>;
     }
-  : T extends (infer U)[]
+  : T extends ReadonlyArray<infer U>
   ? (U extends NonJsonPrimitives ? null : SerializeType<U>)[]
   : T extends object
   ? {


MichaelDeBoey (approved):

packages/remix-react/components.tsx
```diff
@@ -1350,7 +1350,7 @@ type SerializeType<T> = T extends JsonPrimitives
         ? null
         : SerializeType<T[k]>;
     }
-  : T extends (infer U)[]
+  : T extends ReadonlyArray<infer U>
```
kentcdodds, line 1353: Does this allow regular arrays as well?

JNaftali: yes, because `Array<unknown>` extends `ReadonlyArray<unknown>`

MichaelDeBoey: @JNaftali Can you point us to the docs of this please?
Been looking into it, but can't find that.

Maybe @mattpocock can also help out on this one?

kentcdodds (approved):
Thanks!

PR #3780, opened 2022-07-17
fix(templates): put `cross-env` in `devDependencies`

lpsinger:
Like npm-run-all, it is only needed for development.


MichaelDeBoey:
Since `cross-env` is used in the `start` script, we should keep it in `dependencies`.

lpsinger:
Why is npm-run-all a dev dependency?

MichaelDeBoey:
@lpsinger `npm-run-all` is never used in `start` scripts, so that's why it's not necessary to have it in `dependencies` & we can just have it in `devDependencies`

The `start` script will be run by you host & will run the production build, so that's why we need to have `cross-env` in `dependencies` as well

PR #3773, opened 2022-07-16
docs(tutorials/blog): Pass `LoaderData` type as generic to `useLoaderData`

dca123:
Closes: #3772

- [x] Docs
- [ ] Tests

Testing Strategy:

Typescript doesn't show any errors !

PR #3784, opened 2022-07-17
doc: fix code example to prevent type error

chiubaca:
Fixes a type error I was getting when following the tutorial:

image


machour:

We're going to be updating all the documentation with the new type inference very soon, so I'm closing this PR.

Duplicate of #3773

PR #3788, opened 2022-07-18
docs(faq): Fix the mismatch between the button value and the expected value in the switch

jasonadelia:
While reviewing the FAQs around multiple forms in one route I noticed this mismatch in the values. The case in the switch is looking for `update` while the button value was `create`.
- [x] Docs


kentcdodds (approved):
Thanks!

jasonadelia:
> Thanks!

Thank you all for Remix 🥳

Issue #3724, opened 2022-07-13
<PrefetchPageLinks /> crashes the page when there is a query param in the URL


samstanding:
### What version of Remix are you using?

1.6.1

### Steps to Reproduce

1. Add a `<PrefetchPageLinks />` to an active route.
2. Add some query params to your URL ie `?query` and hit return to reload the page
3. Observe: 
<img width="1716" alt="image" src="https://user-images.githubusercontent.com/33382227/178798464-aeb1e49b-5cbd-442e-86ac-a7269ff08e49.png">

### Expected Behavior

I would expect the page to load, unless there's some behavior with PrefetchPageLinks I was not aware of where it can't be rendered on a page with query params. 

### Actual Behavior

The page crashes and we see the following `window is not defined error`
<img width="1716" alt="image" src="https://user-images.githubusercontent.com/33382227/178799060-b7f20d39-4449-45df-b818-b1580d2c6b91.png">

mcansh:
tracking this internally now, and I plan to look into it today!

samstanding:
Hey @mcansh any updates? Is it possible to use my PR as a stopgap so that we can use PrefetchPageLinks if there's a query param?

mcansh:
@samstanding not currently, do you have a use case where you can't conditionally render `<PrefetchPageLinks>`?

samstanding:
> @samstanding not currently, do you have a use case where you can't conditionally render `<PrefetchPageLinks>`?
> 
> skipping them during the server render and applying them during hydration would more than likely cause a hydration warning

@mcansh I can conditionally render it but we're using it to help a slow endpoint so it would be preferred to use it across the board. Right I'm doing something like `{!location.search && <PrefetchPageLinks />}`

danielgangl:
This is still broken for me. Any updates?

danielgangl:
using 'typeof document !== "undefined"' now to conditionally render the PrefetchPageLinks component as a workaround

brophdawg11:
I don't think this is an issue anymore since `1.10.0` and the incorporation of `RouterProvider` from `react-router-dom`

PR #3791, opened 2022-07-18
fix(remix-react): <PrefetchPageLinks /> crashes the page when there is a query param in the URL 

samstanding:
Closes: https://github.com/remix-run/remix/issues/3724

- [ ] Docs
- [ ] Tests

Testing Strategy:

- To reproduce this bug I only need to add a `<PrefetchPageLinks />` to a page and then modify the URL of the route it was on to add a query param ie `?query`. I tested my fix against that scenario and also ensured the `<PrefetchPageLinks />` behavior still worked as expected when there was not a query param.

### DIFF
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -16,16 +16,11 @@ type LiteralUnion<LiteralType, BaseType extends Primitive> =
   | LiteralType
   | (BaseType & Record<never, never>);
 
-/**
- * Represents a `<link>` element.
- *
- * WHATWG Specification: https://html.spec.whatwg.org/multipage/semantics.html#the-link-element
- */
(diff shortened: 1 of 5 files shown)


packages/remix-react/links.ts
```diff
@@ -385,8 +385,7 @@ export function getNewMatchesForLinks(
           if (isNew(match, index) || matchPathChanged(match, index)) {
             return true;
           }
-
-          if (match.route.shouldReload) {
+          if (match.route.shouldReload && typeof window !== "undefined") {
```
mcansh, line 388:
```suggestion
          if (match.route.shouldReload && typeof document !== "undefined") {
```

machour:
@samstanding can you please rebase and fix the conflicts? 🙏🏼

PR #3796, opened 2022-07-19
chore: Add script for faster builds

chaance:
This is a simple script that makes builds faster by default, making `tsc` and copying build artifacts to package `dist` directories opt-in via either `process.env.CI` or command line args.


kentcdodds (approved):
Yep, this works great 👍

Issue #3758, opened 2022-07-15
The inferred type of 'loader' cannot be named without a reference


ZiiMakc:
### What version of Remix are you using?

1.6.5

### Steps to Reproduce

pnpm@7.5.1
typescript@4.7.3

```
import { json } from '@remix-run/node';
import { useLoaderData } from '@remix-run/react';

export async function loader() {
  return json({ greeting: 'Hello!' });
}

export default function App() {
  const data = useLoaderData<typeof loader>();
  return <div>{data.greeting}</div>;
}
```

tsconfig:
```
{
  "exclude": ["**/node_modules"],
  "compilerOptions": {
    /* Visit https://aka.ms/tsconfig.json to read more about this file */

    /* Projects */
    // "incremental": true,                              /* Enable incremental compilation */
    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */

    /* Language and Environment */
    "target": "ES2021" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
    // "lib": ["ES2022"],                                /* Specify a set of bundled library declaration files that describe the target runtime environment. */
    // "jsx": "preserve",                                /* Specify what JSX code is generated. */
    // "experimentalDecorators": true,                   /* Enable experimental support for TC39 stage 2 draft decorators. */
    // "emitDecoratorMetadata": true,                    /* Emit design-type metadata for decorated declarations in source files. */
    // "jsxFactory": "",                                 /* Specify the JSX factory function used when targeting React JSX emit, e.g. 'React.createElement' or 'h' */
    // "jsxFragmentFactory": "",                         /* Specify the JSX Fragment reference used for fragments when targeting React JSX emit e.g. 'React.Fragment' or 'Fragment'. */
    // "jsxImportSource": "",                            /* Specify module specifier used to import the JSX factory functions when using `jsx: react-jsx*`.` */
    // "reactNamespace": "",                             /* Specify the object invoked for `createElement`. This only applies when targeting `react` JSX emit. */
[...]

PR #3797, opened 2022-07-19
Fix `json()` when inferring the return type

esamattis:
Closes: #3758

The explanation is in this comment:
https://github.com/remix-run/remix/issues/3758#issuecomment-1188882909

- [ ] Docs
- [ ] Tests

Testing Strategy:

Manual. This fixes a bug that only appears with build packages as it is basically a packaging issue.


mcansh (approved):

kentcdodds (approved):
Thanks!

PR #3798, opened 2022-07-19
docs(guide): Clarify how to handle 404s

runmoore:
When discussing this topic with colleagues the wording 'you don't have to do anything' was interpreted in an ambiguous way.

This re-phrasing link's back to splat routes as a good option for handling 404s

### DIFF
--- a/docs/guides/not-found.md
+++ b/docs/guides/not-found.md
@@ -11,7 +11,7 @@ There are two primary cases where a Remix site should send a 404:
 - The URL doesn't match any routes in the app
 - Your loader didn't find any data
 
-The first case is already handled by Remix, you don't have to do anything. It knows your routes so it knows if nothing matched. The second case is up to you, but it's really easy.
+The first case is already handled by Remix, you don't have to throw a response yourself. It knows your routes so it knows if nothing matched (_consider using a [Splat Route][splat-route] to handle this case_). The second case is up to you, but it's really easy.
 
 ## How to Send a 404
 
@@ -117,3 +117,4 @@ As you can probably tell, this mechanism isn't just limited to 404s. You can thr
 [catch-boundary]: ../api/conventions#catchboundary
 [errors]: errors
 [404-status-code]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
+[splat-route]: ./routing#splats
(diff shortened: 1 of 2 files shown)


machour (approved):

docs/guides/not-found.md
```diff
@@ -11,7 +11,7 @@ There are two primary cases where a Remix site should send a 404:
 - The URL doesn't match any routes in the app
 - Your loader didn't find any data
 
-The first case is already handled by Remix, you don't have to do anything. It knows your routes so it knows if nothing matched. The second case is up to you, but it's really easy.
+The first case is already handled by Remix, you don't have to throw an error yourself. It knows your routes so it knows if nothing matched (_consider using a [Splat Route][splat-route] to handle this case_). The second case is up to you, but it's really easy.
```
tat-ss, line 14: "you don't have to **throw a response**" is correct here. Error and response are totally different thing in this context.

runmoore: yes good point, I'll clarify it to your suggestion

morwoen (approved):

tat-ss:
I really keen to know if it is intentional to have 2 kind of 404 handlings (spat route and catch boundary) being used together to catch simple route mismatch and response throws from loader/action function. Don't you believe they should be unified into one?

kiliman:
@tat-ss There are actually 3 ways to handle 404 responses.

## Standard 404
1) If you don't want to do anything but return 404 to the client, remove the `<CatchBoundary>` from *root*, and add this to `handleRequest` in *entry.server*
```js
  if (responseStatusCode === 404) {
    return new Response("Not Found", { status: 404 });
  }
```
image

## `<CatchBoundary>` for custom 404 page
2) Add a `<CatchBoundary>` to *root* to render a custom 404 page.
```js
export const CatchBoundary = () => {
  const caught = useCatch();
  const location = useLocation();
  return (
    <div>
      <h1>
        Catch {caught.status} {caught.statusText}
      </h1>
      <h2>Location: {`${location.pathname}${location.search}`}</h2>
    </div>
  );
};
```
image

## Splat route for dynamic 404 handling
3) Add a `splat` route. This gives you the ability to do something like check database if route has been moved and redirect.
```ts
// routes/$.tsx
export const loader: LoaderFunction = async ({ request }) => {
  if (url.pathname === "/not-found") {
    return redirect("/found");
  }
  throw new Response("Not Found", { status: 404 });
};
```
image

kentcdodds (approved):
I like this improved wording 👍

PR #3800, opened 2022-07-19
fix(mdx): remove overriding of the loader

kentcdodds:
Ref #3763


kentcdodds:
Nevermind. Just going to revert the whole PR and add docs instead.

PR #3802, opened 2022-07-19
docs(examples): convert `sanity` example to TypeScript

MichaelDeBoey:


examples/sanity/app/lib/sanity/usePreviewSubscription.ts
```diff
@@ -27,29 +30,36 @@ export function usePreviewSubscription(query, subscriptionOptions) {
...
 
     if (!store) {
       createStore();
     }
 
     return () => {
-      if (sub?.unsubscribe()) sub.unsubscribe();
-      if (store) store.close();
+      if (sub?.unsubscribe()) {
+        sub.unsubscribe();
```
machour, line 53: Is it wanted that `unsubscribe()` is called twice?

MichaelDeBoey: It was in the original example, but I don't think it's wanted, so I'll remove the condition call 👍

machour (approved):

PR #3804, opened 2022-07-20
fix: update docs, since CF Functions use a different file for local env variables.

awthwathje:


awthwathje (commented):
Not only this cat trick doesn't work in PowerShell, there is another official way to achieve that.

kentcdodds (approved):
Thanks!

awthwathje:
Wow, Kent C. Dodds himself merged my little PR. That's pretty cool! 🙂

PR #3813, opened 2022-07-21
Fix docs broken route filename convention link

alextes:
Simple docs change.

I believe this should do it. Took a while to figure out with the magical remark processing of the docs files. Hard to be sure.

This is the page the link should go to: https://remix.run/docs/en/v1/api/conventions#route-file-conventions


alextes:
fine CLA bot 😩

machour (approved):

kentcdodds (approved):
Thank you!

PR #3817, opened 2022-07-21
fix: write other server build output files 

mcansh:


mcansh:
this does have one downfall where css files from tailwind are written by tailwind and then again by us here

integration/resource-routes-test.ts
```diff
@@ -85,5 +103,11 @@ test.describe("loader in an app", () => {
       await app.goto("/data.json");
       expect(await page.content()).toContain('{"hello":"world"}');
     });
+
+    test.only("should include imported asset in build", async ({ page }) => {
```
kentcdodds, line 107:
```suggestion
    test("should include imported asset in build", async ({ page }) => {
```

kentcdodds (approved):
Solid 👍

kentcdodds:
This is good to merge when the tests are green 👍

wKovacs64:
While this does produce the files in the build output on the file system, for some reason they don't seem to be picked up by `express.static` middleware and the request for the asset falls through to the Remix request handler.

Note: my current workaround of re-importing the assets in a UI route still works, in case that provides any additional insight.

kentcdodds:
@wKovacs64 can you create a reproduction and open an issue?

kentcdodds:
Ah, sorry, I didn't notice those. @mcansh can you look into why these changes didn't address the original issue?

mcansh:
> Ah, sorry, I didn't notice those. @mcansh can you look into why these changes didn't address the original issue?

i think i misread the initial bug report, but the reason for them not being in the public/build is due to them only being server assets - i'll take a look, but i'm not 100% sure why you'd want to import something in a recorded route and use it elsewhere

mcansh:
alrighty, should be resolved here: https://github.com/remix-run/remix/pull/3841

i'm writing files to both the `serverBuildPath` (esbuild default) and `assetsDirectory`

PR #3830, opened 2022-07-24
docs: Fix unneeded `await` in External API docs

JeffersonBledsoe:


JeffersonBledsoe:
Ignore this, I was using `getLoaderData` function and didn't notice I was already calling `json()` 🤦

PR #3837, opened 2022-07-26
docs(blog-tutorial): Remove json function call and the useLoaderData typescript error

nisalVD:
Remove json function call in blog doc and fixed the typescript error for type assertion on useLoaderData

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -405,3 +405,4 @@
 - zachdtaylor
 - zainfathoni
 - zhe
+- nisalvd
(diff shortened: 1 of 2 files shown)


kiliman:
Hmm.. not sure why you removed the `json` calls. Future versions of Remix will probably deprecate returning naked objects. Loaders and actions should return a Response. The `json` helper, like `redirect` creates a new Response.

NOTE: You also need to use the `json` helper if you want to get the new type inference released in v1.6.5.

```js
function loader({request}: LoaderArgs) {
  return json({ greeting: 'hello' })
}

const data = useLoaderData<typeof loader>()
// data will be fully typed (as the Serialized type)
```

machour:
Thank you for your contribution @nisalVD, but as @kiliman mentioned, you shouldn't be getting rid of `json()` calls, and we will be (internally) updating the documentation soon to reflect the new type inference soon. Stay tuned!

PR #3415, opened 2022-06-08
Bug: Assets only imported in resource routes not included in build output

wKovacs64:
Reproduces #3414

- [ ] Docs
- [x] Tests

Testing Strategy:

Followed the instructions in `bug-report-test.ts`

PR #3841, opened 2022-07-26
fix(remix-dev): write other server build output files to `assetsBuildDirectory`

### LINKED ISSUE #3414: Assets only imported in resource routes not included in build output

1.5.1

### Steps to Reproduce

1. Create a resource route
2. Import an asset (e.g.,  an image)
3. Observe the resulting URL appears correct (includes a hash) 👍 
4. Observe the resulting URL 404s as the asset does not make it into the build directory ☹️ 

Bonus steps:

5. Import the asset into a UI route
6. Observe the asset is now included in the build output

### Expected Behavior

I would expect an asset imported (only) in a resource route to be included in the build output.

### Actual Behavior

The asset is not included in the build output unless it is also imported in a UI route.

mcansh:
(cherry picked from commit 6ae5cdfd6b8ecdc76d40ffc24b208eabf144c27c)

closes #3414
closes #3415

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


kentcdodds:
Ok, to make sure I understand, we're now writing static files to both the public directory as well as the server build directory because those static files can be accessed by both sides of the app right?

I wonder if there's a way for us to determine which side of the app needs it. I'm worried that we'd have issues in serverless situations where you're limited on the amount of content you can upload 🤔

mcansh:
> Ok, to make sure I understand, we're now writing static files to both the public directory as well as the server build directory because those static files can be accessed by both sides of the app right?

correct

> I wonder if there's a way for us to determine which side of the app needs it. I'm worried that we'd have issues in serverless situations where you're limited on the amount of content you can upload 🤔

totally a valid concern, I'll check it out

kentcdodds:
Let's ask @jacob-ebey about this. I have a feeling he'd know the implications of us copying these files over better.

ni554n:
Hi, is this going fix the extraneous tailwind .css files generating in `functions/_assets` directory?

mcansh:
> Hi, is this going fix the extraneous tailwind .css files generating in `functions/_assets` directory?

it will, yes, sorry about that 😬

MichaelDeBoey (approved):

houmark:
Just adding my concern with this one. Cloudflare workers are very limited in size, so a few assets copied to the server build directory could easily block building I think.

mcansh:
> Just adding my concern with this one. Cloudflare workers are very limited in size, so a few assets copied to the server build directory could easily block building I think.

the only extra files that would be written to the build directory would be those imported in resource routes

houmark:
> valid concern, the only extra files that should be written to the build directory would be those imported in resource routes

Thanks @mcansh, but not 100% sure I understand that reply. If I have image files, for example, inside a resource route, will those be imported into the build directory, and will those be part of the cloud function itself, meaning it will count to a potential limit of that cloud service (ie. currently 1MB at Cloudflare)?

New to Remix and new to Cloudflare Pages, so just want to understand what one can expect here. I know best practice is to use a static asset storage, so one would not want to add many image files as imports, but there can be some situations where that is needed.

mcansh:
> If I have image files, for example, inside a resource route, will those be imported into the build directory, and will those be part of the cloud function itself, meaning it will count to a potential limit of that cloud service (ie. currently 1MB at Cloudflare)?

they will end up in the `public/build` directory, yes, but for CF, they'll be uploaded to KV just like your other assets so it shouldn't affect your function size, currently if you import a file only in a resource route, that file is completely ignored.

PR #3851, opened 2022-07-27
docs: First pass at defer API design doc

brophdawg11:
Design doc for the upcoming `defer()` + `<Await>` APIs

[Preview link](https://github.com/remix-run/remix/blob/brophdawg11/deferred-design-doc/decisions/0003-streaming-apis.md)

[🔥 Preview link using github-md.com](https://github-md.com/remix-run/remix/blob/brophdawg11/deferred-design-doc/decisions/0003-streaming-apis.md)

### DIFF
--- a/decisions/0003-streaming-apis.md
+++ b/decisions/0003-streaming-apis.md
@@ -18,7 +18,7 @@ It's also worth nothing that even in a single-page-application without SSR-strea
 
 Streaming in Remix can be thought of as having 3 touch points with corresponding APIs:
 
-1. _Initiating_ a streamed response in your `loader` can be done by returning a `defer(object, responseInit?)` call from your `loader` in which some of the keys on `object` are `Promise` objects
+1. _Initiating_ a streamed response in your `loader` can be done by returning a `defer(object)` call from your `loader` in which some of the keys on `object` are `Promise` objects
 2. _Accessing_ a streamed response from `useLoaderData`
    1. No new APIs here - when you return a `defer()` response from your loader, you'll get `Promise` values inside your `useLoaderData` object 👌
 3. _Rendering_ a streamed value (with fallback and error handling) in your component
--- a/decisions/0003-streaming-apis.md
+++ b/decisions/0003-streaming-apis.md
@@ -22,15 +22,15 @@ Streaming in Remix can be thought of as having 3 touch points with corresponding
    1. Return a naked Javascript object with one or more `Promise` values
    2. Use the `defer(object, responseInit)` utility if you need to customize the `Response`
 2. _Accessing_ a streamed response from `useLoaderData`
-   1. No new APIs here - when you return `Promise` from your loader, you'll get a `Promise` from `useLoaderData` 👌
+   1. No new APIs here - when you return an object with `Promise` values or a `defer()` response from your loader, you'll get `Promise` values inside your `useLoaderData` object 👌
 3. _Rendering_ a streamed value (with fallback and error handling) in your component
-   1. You can render a `Promise` with the `<Await resolve={promise}>`
+   1. You can render a `Promise` from `useLoaderData()` with the `<Await resolve={data.promise}>` component
    2. `<Await>` accepts an `errorElement` prop to handle error UI
-   3. `<Await>` should be wrapped with a `<React.Suspense fallback={}>` component to handle your loading UI
+   3. `<Await>` should be wrapped with a `<React.Suspense>` component to handle your loading UI
 
 ## Details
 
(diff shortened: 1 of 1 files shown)


kiliman:
I like the new naming conventions, but I'm surprised you guys are still going with naked objects.

Right now if you return a naked Promise, Remix will await the Promise, then return a JSON response. How will this work with the new defer API? 

Because with the new API, the data will be streamed, since you're returning a Promise, and the result from `useLoaderData` will be a Promise, so accessing the data (without await) will fail. So essentially, once they update Remix, existing code may break.

By forcing everyone to use either `json` or `defer`, they're explicitly opting into the desired functionality and existing code will continue to function (perhaps deprecated).

```ts
async function getData() {
  return Promise.resolve([1, 2, 3]);
}
export const loader = async ({ request }: LoaderArgs) => {
  // return naked Promise
  // current Remix awaits and converts to JSON
  return getData();
};

export default function () {
  // data is currently [1,2,3] but defer API will make it a Promise
  const data = useLoaderData();

  // will break when defer API is released
  console.log(data.length)
  ...
}
```

brophdawg11:
@kiliman I might need to clarify in the doc but a single naked `Promise` being returned from the `loader` won't automatically be streamed.  It'll only be root-level Promises on objects returned from the `loader`.  So in your case above data would still be an array.

```
async function getData() {
  return Promise.resolve([1, 2, 3]);
}
export const loader = async ({ request }: LoaderArgs) => {
  return {
    critical: await getData(),
    // Lazy will get streamed since it's still a Promise after await loader()
    lazy: getData()
  };
}
```

brophdawg11:
@kiliman I see your confusion now - I was a bit too liberal with the "return a promise from a loader" wording.  I'm changing those to be more explicit now, it's more "return an object with Promise values"

kiliman:
Ok, but that's kind of a _minor_ distinction with _huge_ implications. Anyway, I think it's going to be very confusing to a lot of people, because they won't read the documentation. You're going to have a bunch of "I'm returning a Promise, but it's not streaming" questions.

Me and @sergiodxa handle a lot of support questions on Discussions and Discord. So I'm just giving you feedback on my experience with the level of questions we get.

sergiodxa:
I like the new `Await` API and that `useLoaderData` returns a Promise instead of a custom object, this mean I could do
```ts
let { someValue } = useLoaderData<typeof loader>();
async function doSomething() {
  let value = await someValue;
  // use value here with some client-side only code
}
```
This can be interesting to integrate with client-side libs that requires getting some data that could be slow.

kiliman:
I like the API as well. I'm just not a fan of the perpetuation of naked objects. The problem is that Remix is going to have to be smart about what a naked object means. You already have to support the legacy of always converting to JSON (and awaiting if a Promise).

As new APIs are created, you will always have to some documentation in place that the Promise, Stream, Whatever has to be wrapped in an object first. Where as you could always do the following. This is the perfect opportunity to deprecate naked objects. Now anybody looking at your code can immediately see what and how the data is returned.

```js
return json("hello")
return defer(new Promise())
return stream(new Stream())
```

Anyway, just my 2c

sergiodxa:
> Because this promise isn't cancelled

Talking about cancellation of promises, should there be a documented way to cancel them? I know there's `request.signal` but I think it's never marked as cancelled (I remember there's a PR to fix it if it's not merged yet).

It would be really useful to document to you could defer and cancel those promises.

```ts
export async function loader({ request }: LoaderArgs) {
  return defer({ slowThing: getSlowThing() })

  async function getSlowThing() {
    // shows how to use it to abort a fetch to another service
    let res = await fetch(SLOW_API_ENDPOINT, { signal: request.signal }
    return await res.json()
  }
}
```

I think something like this if `request.signal` works.

And maybe another showing how to use it without fetch, probably the simplest can be a setTimeout but the idea could be to show how you could use the abort event or the aborted property.

```ts
export async function loader({ request }: LoaderArgs) {
  return defer({ slowThing: getSlowThing() })

  function getSlowThing() {
    return new Promise(resolve => {
      let timer = setTimeout(() => resolve(data, 1000))
      // show how to subscribe to the abort event
      request.signal.addEventListener("abort", () => clearTimeout(timer), { once: true })
    })
  }
}
```

```ts
export async function loader({ request, params }: LoaderArgs) {
  return defer({ slowThing: getSlowThing() })

  async function getSlowThing() {
    let user = await getUser(params.id)
    // this can show how you could stop a process in the middle
    if (request.aborted) throw new AbortError();
    let friends = await getFriends(user.friendList)
    return { user, friends })
  }
}
```

kiliman:
❤️ Love the changes. Thanks!

PR #3872, opened 2022-07-29
Calm esm warnings

jacob-ebey:
Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

Added integration test that checks console output for warnings.

### DIFF
--- a/.changeset/silent-trains-build.md
+++ b/.changeset/silent-trains-build.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Don't show ESM warnings when consumed via dynamic import.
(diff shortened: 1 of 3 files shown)


kentcdodds (commented):
Sweet! Thank you so much for taking the time to work on this.

One other thing I've noticed (with the current implementation) is that if there's a dep of a dep that's ESM only I'll get a warning about that as well. For example, in mdx-bundler we're using `@mdx-js/esbuild` and `remark-frontmatter` which are both ESM-only, but mdx-bundler itself is CJS so it's using a dynamic import: https://github.com/kentcdodds/mdx-bundler/blob/37f402702bcf8163cb64314a8c43ed9efac04fbc/src/index.js#L39-L42

Does this cover that use case as well? If not, can you expand this PR to silence those as well?

jacob-ebey:
@kentcdodds added a test for the CJS package that does a dynamic import of an ESM only package.

mcansh (approved):

kentcdodds (approved):
Stellar. Thank you!

PR #3873, opened 2022-07-29
ci: use a custom token so that postrelease workflow is triggered

mcansh:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


chaance:
@mcansh Can you explain the issue a bit in this one? Is this because we have a separate token for nightly releases? Where in the postrelease workflow are we looking for this token?

mcansh:
> @mcansh Can you explain the issue a bit in this one? Is this because we have a separate token for nightly releases? Where in the postrelease workflow are we looking for this token?

you bet, the gist is that the token that's automatically added to an action will trigger another workflow to run https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

PR #3875, opened 2022-07-29
docs: add decision doc for remix router + react router 6.4

brophdawg11:
Started a decision doc to capture decisions made in the react-router 6.4.0 work.  Just a rough first pass - comments welcome 😄 

[Preview](https://github.com/remix-run/remix/blob/brophdawg11/router-decision-doc/decisions/0005-remixing-react-router.md)

[Github-md Preview 🔥](https://github-md.com/remix-run/remix/blob/brophdawg11/router-decision-doc/decisions/0005-remixing-react-router.md)


decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,173 @@
...
+
+This is no longer the case in a data-aware landscape. Now, when a user clicks a link, we need to first tell the router "hey the user _intends_ to go to this location." In response to that the router can initiate some data fetches but during these fetches we're still on the old page! The user is still looking at the old content, and the URL should reflect that. This fits with the "browser emulator" concept as well. If you had a non-JS landscape and a user clicked a link from `/a -> /b` and the server took 5 seconds to send back a response for `/b` - during that time the browser URL bar shows the URL and title for `/a` and a little spinner in the tab. This is exactly how we built the router, it first loads data, then it updates state and tells history to update the URL.
+
+There's one caveat here when it comes to back/forward button usage. When the users navigates back/forward in the history stack we get a `popstate` event _but the URL has already been updated_. So the best we can do there is react to the new URL. This is _not_ what the browser would do in a non-JS world but we really have no choice.
```
machour, line 29:
```suggestion
There's one caveat here when it comes to back/forward button usage. When the user navigates back/forward in the history stack we get a `popstate` event _but the URL has already been updated_. So the best we can do there is react to the new URL. This is _not_ what the browser would do in a non-JS world, but we really have no choice.
```

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,173 @@
...
+
+All this being said - history is no longer a simple process of "update the URL then tell the router to re-render". History and routing are far more intertwined and behave slightly differently for PUSH/REPLACE than they do for POP navigations. For PUSH/REPLACE we go `router.navigate -> load data -> update state -> update history`, but for POP we `update history -> router.navigate -> load data -> update state`. So in PUSH, the router informs history. But in POP, history informs the router. These nuances made make sense to keep the router as the public API and make history more of an internal implementation detail.
```
machour, line 31:
```suggestion
All this being said - history is no longer a simple process of "update the URL then tell the router to re-render". History and routing are far more intertwined and behave slightly differently for PUSH/REPLACE than they do for POP navigations. For PUSH/REPLACE we go `router.navigate -> load data -> update state -> update history`, but for POP we `update history -> router.navigate -> load data -> update state`. So in PUSH, the router informs history. But in POP, history informs the router. These nuances made sense to keep the router as the public API and make history more of an internal implementation detail.
```

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,173 @@
...
+
+We plan to export `useNavigation` in Remix and encourage folks to switch, but we will continue to include `useTransition` in a deprecated fashion
+
+### Navigations/Fetchers state structure changes
+
+**`useTransition().type` is removed**
+
+In Remix, the `useTransition` hook returned a Transition object which had a `state` property of `"idle" | "loading" | "submitting"`. It also had a `type` property which represented sort of "sub-states" such as `"normalLoad" | "actionReload" | "loaderRedirect"` etc. In React Router we chose to get rid of the `type` field for 2 reasons:
+
+1. In practice we found that the _vast_ majority of the time al you needed to reference was the `state`
```
machour, line 66:
```suggestion
1. In practice, we found that the _vast_ majority of the time al you needed to reference was the `state`
```

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  // Otherwise, do what we would have done by default
+  return defaultShouldRevalidate;
+}
+```
+
+### `<ScrollRestoration getKey>` prop
+
+In Remix, the `<ScrollRestoration>` component made an assumption that we would always restore scroll position based on `location.key`. If the key was the same as a prior location we knew the scroll position for, then we knew you had been there before and we should restore. This works great for back/forward navigations but it's a bit overly restrictive. You cannot choose tgo restore scroll based on anything other than `key`. Twitter has a great implementation of this as you click around in their left nav bar - your tweet feed is always at the same place when you click back to it - even though its a _new_ location in the history stack. This is because they're restoring by pathname here instead of `location.key`. Or maybe you want to maintain scroll position for all routes under a given pathname and you thus want to use a portion of the pathname as the scroll restoration key.
```
machour, line 127:
```suggestion
In Remix, the `<ScrollRestoration>` component made an assumption that we would always restore scroll position based on `location.key`. If the key was the same as a prior location we knew the scroll position for, then we knew you had been there before and we should restore. This works great for back/forward navigations but it's a bit overly restrictive. You cannot choose tgo restore scroll based on anything other than `key`. Twitter has a great implementation of this as you click around in their left nav bar - your tweet feed is always at the same place when you click back to it - even though it's a _new_ location in the history stack. This is because they're restoring by pathname here instead of `location.key`. Or maybe you want to maintain scroll position for all routes under a given pathname and you thus want to use a portion of the pathname as the scroll restoration key.
```

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,173 @@
...
+}
+```
+
+**Backwards Compatibility**
+
+We're ok here since the new prop is optional and defaults to using `location.key`
+
+### `<Link resetScroll="false">` prop
+
+In addition to `<ScrollRestoration>` handling "restoring" scroll position on previously visited routes. It also handles "resetting" scroll position back to the top on _new_ routes. This is not always desireable if you're clicking around inside a tabbed view or something, so we've introduced a new `<Link resetScroll>` prop that lets you disable the scroll reset behavior _for a given navigation_. Note that this "resetting" logic happens if and only if we cannot restore scroll to a previously known location for that scroll restoration key.
```
machour, line 148:
```suggestion
In addition to `<ScrollRestoration>` handling "restoring" scroll position on previously visited routes. It also handles "resetting" scroll position back to the top on _new_ routes. This is not always desirable if you're clicking around inside a tabbed view or something, so we've introduced a new `<Link resetScroll>` prop that lets you disable the scroll reset behavior _for a given navigation_. Note that this "resetting" logic happens if and only if we cannot restore scroll to a previously known location for that scroll restoration key.
```

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,173 @@
...
+
+The differentiation between error and catch prove to be a bit vague over time and a source of confusion for developers. We chose to go with just a single `errorElement` in the router for simplicity. If you throw anything, it ends up in the error boundary and propagates accordingly.
+
+**Backwards Compatibility**
+
+We will handle this in Remix where it's `errorElement` will decide if it should render via the `ErrorBoundary` or `CatchBoundary`.
+
+### `Request.signal` instead of `signal` param
+
+We dropped the `signal` parameter to loaders and actions because an incoming `Request` already has it's own signal!
```
machour, line 166:
```suggestion
We dropped the `signal` parameter to loaders and actions because an incoming `Request` already has its own signal!
```

machour (commented):
Some nitpicking 🤓

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,173 @@
...
+
+### No distinction between Error and Catch boundaries
+
+TODO: Ryan to fill in some more details here.
+
+The differentiation between error and catch prove to be a bit vague over time and a source of confusion for developers. We chose to go with just a single `errorElement` in the router for simplicity. If you throw anything, it ends up in the error boundary and propagates accordingly.
+
+**Backwards Compatibility**
+
+We will handle this in Remix where it's `errorElement` will decide if it should render via the `ErrorBoundary` or `CatchBoundary`.
```
machour, line 162:
```suggestion
We will handle this in Remix where its `errorElement` will decide if it should render via the `ErrorBoundary` or `CatchBoundary`.
```

machour: About this part, I understand that things will not change in Remix, and we'll keep using `ErrorBoundary` and `CatchBoundary`.

If that's indeed the case, why aren't we clearing up the confusion in Remix too by having a single function for both?

decisions/0004-remixing-react-router.md
brophdawg11:
```suggestion
1. In practice, we found that the _vast_ majority of the time all you needed to reference was the `state`
```

decisions/0004-remixing-react-router.md
brophdawg11: yeah - still some stuff up in the air there.  We definitely plan to drop to a single function in v2, I think initially we were thinking we'd just drop `CatchBoundary` and handle everything in `ErrorBoundary` in v2.  However, I think what you're proposing is introduce a net new (maybe `ErrorElement`) in Remix v1 and deprecating both `ErrorBoundary`/`CatchBoundary` in v1 (and dropping in v2).  I kind of like that idea - I'll run that by the team as well!

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  // Otherwise, do what we would have done by default
+  return defaultShouldRevalidate;
+}
+```
+
+### `<ScrollRestoration getKey>` prop
+
+In Remix, the `<ScrollRestoration>` component made an assumption that we would always restore scroll position based on `location.key`. If the key was the same as a prior location we knew the scroll position for, then we knew you had been there before and we should restore. This works great for back/forward navigations but it's a bit overly restrictive. You cannot choose tgo restore scroll based on anything other than `key`. Twitter has a great implementation of this as you click around in their left nav bar - your tweet feed is always at the same place when you click back to it - even though it's a _new_ location in the history stack. This is because they're restoring by pathname here instead of `location.key`. Or maybe you want to maintain scroll position for all routes under a given pathname and you thus want to use a portion of the pathname as the scroll restoration key.
```
Girish21, line 127:
```suggestion
In Remix, the `<ScrollRestoration>` component made an assumption that we would always restore scroll position based on `location.key`. If the key was the same as a prior location we knew the scroll position for, then we knew you had been there before and we should restore. This works great for back/forward navigations but it's a bit overly restrictive. You cannot choose to restore scroll based on anything other than `key`. Twitter has a great implementation of this as you click around in their left nav bar - your tweet feed is always at the same place when you click back to it - even though it's a _new_ location in the history stack. This is because they're restoring by pathname here instead of `location.key`. Or maybe you want to maintain scroll position for all routes under a given pathname and you thus want to use a portion of the pathname as the scroll restoration key.
```

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,201 @@
...
+  // Otherwise use the key
+  return location.key;
+}
+```
+
+**Backwards Compatibility**
+
+We're ok here since the new prop is optional and defaults to using `location.key`
+
+### `<Link resetScroll="false">` prop
```
sergiodxa, line 170: Why not `<Link preventScrollReset>` or something like that? With resetScroll the default behavior will be true so you will have to do `<Link resetScroll={false} />` if the prop default is false the usage would be just `<Link propName />` which is a little nicer.

decisions/0004-remixing-react-router.md
```diff
@@ -0,0 +1,201 @@
...
+
+**TODO:** Ryan to fill in some more details here.
+
+The differentiation between error and catch prove to be a bit vague over time and a source of confusion for developers. We chose to go with just a single `errorElement` in the router for simplicity. If you throw anything, it ends up in the error boundary and propagates accordingly.
+
+**Backwards Compatibility**
+
+We have a few options here. In all cases, Remix v1 will provide an `errorElement` implementation that will need to do some forking to maintain backwards compatibility.
+
+1. We could introduce a new `ErrorElement` in Remix v1 and deprecate `ErrorBoundary`/`CatchBoundary` (and eventually drop them in v2)
```
sergiodxa, line 188: Proposal: ErrorComponent instead of ErrorElement, because you should export a component in Remix, not an element, like with ErrorBoundary/CatchBoundary/default which are all components.

kiliman: I wonder if they use `element` here to match React Router `<Route path="" element={}/>`

brophdawg11: yeah this was a very quick "let's call it what we call it in react router," but I agree with Sergio that from a typing standpoint you're exporting a Component in Remix, not something that's been through `createElement` 👍

decisions/0004-remixing-react-router.md
brophdawg11: @ryanflorence What do you think?

MichaelDeBoey (approved):

decisions/0005-remixing-react-router.md
```diff
@@ -0,0 +1,202 @@
...
+
+But, in Remix we were considering the latter a "submission" such that `useTransition().state === "submitting"`. In order to ensure our "navigations" reflect the browser behavior we have changed this in the router such that GET Form submissions result in `useNavigation().state === "loading"`.
+
+**Backwards Compatibility**
+
+This will be handled in the deprecated `useTransition` hook along with the backfill of `type` and `submission` properties
+
+### Form automatic replace behavior
+
+When performing POST navigations, you don't want to end up with a duplicate entry in the history stack which makes back-button routing weird when going through the same page twice. This is further complicated in browsers that hang onto the submission info and thus have to prompt you to warn you of re-submitting your data. We'll look at a few examples to demonstrate, but the intention is that the default behavior of the router should ensure that you can't get yourself into this double-history-entry situation when using `<Form method="post">` submission navigations.
```
abenhamdine, line 109: interesting !
may you precise what is the current/future behaviour for PATH/PUT/DELETE methods ?

brophdawg11: They all behave the same as `POST` when JS is enabled: https://reactrouter.com/en/dev/components/form#replace

decisions/0005-remixing-react-router.md
```diff
@@ -0,0 +1,202 @@
...
+  formMethod?: FormMethod;
+  formAction?: string;
+  formEncType?: FormEncType;
+  formData?: FormData;
+}
+```
+
+**Backwards Compatibility**
+
+We plan to remain backwards compatible here in Remix. Very likely we'll expose the `useNavigation` hook and encourage users to move to that. And then `useTransition` will remain in a deprecated state and it will cal `useNavigation` and then backfill the `type` and `submission` properties.
```
machour, line 86:
```suggestion
We plan to remain backwards compatible here in Remix. Very likely we'll expose the `useNavigation` hook and encourage users to move to that. And then `useTransition` will remain in a deprecated state and it will call `useNavigation` and then backfill the `type` and `submission` properties.
```

decisions/0005-remixing-react-router.md
```diff
@@ -0,0 +1,202 @@
...
+```jsx
+<a href="/search?query=matt">Search</a>
+
+<form action="/search">
+  <input name="query" value="matt" />
+  <button type="submit">Search</button>
+</form>
+```
+
+But, in Remix we were considering the latter a "submission" such that `useTransition().state === "submitting"`. In order to ensure our "navigations" reflect the browser behavior we have changed this in the router such that GET Form submissions result in `useNavigation().state === "loading"`.
```
machour, line 101:
```suggestion
But, in Remix we were considering the latter a "submission" such that `useTransition().state === "submitting"`. In order to ensure our "navigations" reflect the browser behavior, we have changed this in the router such that GET Form submissions result in `useNavigation().state === "loading"`.
```

decisions/0005-remixing-react-router.md
```diff
@@ -0,0 +1,202 @@
...
+- Action redirects to `/private`
+  - If we treated this redirect as a REPLACE, we'd be replacing the _initial_ navigation to `/login` since we haven't yet touched history for the POST. This would leave the history stack as `[/, /private]` and we'd lose the fact that we were ever at the login page.
+  - Instead when an action redirects, we'll use a PUSH and in this case the history stack would become `[/, /login, /private]` and the user would be able to navigate back through the login page and to the home page
+
+Note: User's can still be explicit here and use `<Form method="post" replace={shouldReplace}>` and the router will respect the value passed to `replace`.
+
+### `unstable_shouldReload` stabilized as `shouldRevalidate`
+
+We stabilized the API for when a given route loader should re-run, and changed the name to align with the "revalidation" nomenclature and the `useRevalidator` hook. We also leave more control in the hands of the user here. In Remix there were some cases in which you _could not_ opt out of revalidation and if your method did run you had full control and couldn't necessarily handle one edge case and then say "do what you otherwise would have done".
```
machour, line 133:
```suggestion
We stabilized the API for when a given route loader should re-run, and changed the name to align with the "revalidation" nomenclature and the `useRevalidator` hook. We also leave more control in the hands of the user here. In Remix there were some cases in which you _could not_ opt out of revalidation and if your method did run, you had full control and couldn't necessarily handle one edge case and then say "do what you otherwise would have done".
```

decisions/0005-remixing-react-router.md
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  // Otherwise, do what we would have done by default
+  return defaultShouldRevalidate;
+}
+```
+
+### `<ScrollRestoration getKey>` prop
+
+In Remix, the `<ScrollRestoration>` component made an assumption that we would always restore scroll position based on `location.key`. If the key was the same as a prior location we knew the scroll position for, then we knew you had been there before and we should restore. This works great for back/forward navigations but it's a bit overly restrictive. You cannot choose to restore scroll based on anything other than `key`. Twitter has a great implementation of this as you click around in their left nav bar - your tweet feed is always at the same place when you click back to it - even though it's a _new_ location in the history stack. This is because they're restoring by pathname here instead of `location.key`. Or maybe you want to maintain scroll position for all routes under a given pathname and you thus want to use a portion of the pathname as the scroll restoration key.
```
machour, line 151:
```suggestion
In Remix, the `<ScrollRestoration>` component made an assumption that we would always restore scroll position based on `location.key`. If the key was the same as a prior location we knew the scroll position for, then we knew you had been there before and we should restore. This works great for back/forward navigations but it's a bit overly restrictive. You cannot choose to restore scroll based on anything other than `key`. 

Twitter has a great implementation of this as you click around in their left nav bar - your tweet feed is always at the same place when you click back to it - even though it's a _new_ location in the history stack. This is because they're restoring by pathname here instead of `location.key`. Or maybe you want to maintain scroll position for all routes under a given pathname and you thus want to use a portion of the pathname as the scroll restoration key.
```

decisions/0005-remixing-react-router.md
```diff
@@ -0,0 +1,202 @@
...
+
+### `useRevalidator()` hook
+
+This has been a long time coming - see https://github.com/remix-run/remix/discussions/1996 🙂
+
+### No distinction between Error and Catch boundaries
+
+**TODO:** Ryan to fill in some more details here.
+
+The differentiation between error and catch prove to be a bit vague over time and a source of confusion for developers. We chose to go with just a single `errorElement` in the router for simplicity. If you throw anything, it ends up in the error boundary and propagates accordingly.
```
machour, line 182:
```suggestion
The differentiation between error and catch proved to be a bit vague over time and a source of confusion for developers. We chose to go with just a single `errorElement` in the router for simplicity. If you throw anything, it ends up in the error boundary and propagates accordingly.
```

decisions/0005-remixing-react-router.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Decisions
+
+### Move the bulk of logic to a framework-agnostic router
+
+Thankfully this decision was sort of already made by Ryan. Maybe a surprise to some, maybe not, the current transition manager doesn't import or reference `react` or `react-router` a single time. This is by design because the logic being handled has nothing to do with how to render the UI layer. It's all about "what route am I on?", "what route am I going to?", "how do I load data for the next route?", "how do I interrupt ongoing navigations?" etc. None of these decisions actually _care_ about how the route and it's data will eventually be rendered. Instead, the router simply needs to know whether given routes _have_ components and/or error boundaries - but it doesn't need to know about them or how to render them.
```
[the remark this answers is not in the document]
brophdawg11: Thanks @machour!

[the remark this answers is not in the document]
kiliman: Ha I still mix up when to use `its` and `it's`

PR #3889, opened 2022-07-31
perf(remix-server-runtime): use faster alternative to `jsesc`

redabacha:
Closes: #3854 

- [ ] Docs
- [x] Tests

Testing Strategy:

used a faster version of [`jsesc`](https://github.com/mathiasbynens/jsesc) to serialize the server handoff very much inspired by what next.js has done [here](https://github.com/vercel/next.js/blob/97ab23fb6f05c4b44c2a14102b2b141ca3ec2e2e/packages/next/server/htmlescape.ts). created a benchmark [here](https://github.com/redabacha/remix-jsesc-slow) (which i also included some other libraries in the comparison for additional points of reference).

results from my machine:
```
cpu: AMD Ryzen 9 5950X 16-Core Processor
runtime: node v16.16.0 (x64-linux)

benchmark                                       time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------------------------------- -----------------------------
escapeHtml                                   18.69 ms/iter   (17.17 ms … 32.03 ms)  17.95 ms  32.03 ms  32.03 ms
escapeHtml w/ fast-json-stable-stringify      99.6 ms/iter  (86.78 ms … 146.71 ms)  97.41 ms 146.71 ms 146.71 ms
escapeHtml w/ json-stable-stringify         103.18 ms/iter  (75.77 ms … 134.89 ms) 116.19 ms 134.89 ms 134.89 ms
escapeHtml w/ json-stringify-deterministic  105.39 ms/iter  (89.47 ms … 131.08 ms) 115.49 ms 131.08 ms 131.08 ms
escapeHtml w/ superjson.stringify           351.23 ms/iter    (322 ms … 432.56 ms) 362.27 ms 432.56 ms 432.56 ms
jsesc                                       258.64 ms/iter (237.88 ms … 289.99 ms) 263.97 ms 289.99 ms 289.99 ms
js-string-escape                            106.71 ms/iter  (92.13 ms … 132.33 ms) 117.55 ms 132.33 ms 132.33 ms
serialize-javascript                         43.52 ms/iter   (40.56 ms … 73.23 ms)  42.22 ms  73.23 ms  73.23 ms
remix-typedjson serialize                    42.35 ms/iter   (40.97 ms … 50.97 ms)  42.53 ms  50.97 ms  50.97 ms

summary
  escapeHtml
   2.27x faster than remix-typedjson serialize
   2.33x faster than serialize-javascript
   5.33x faster than escapeHtml w/ fast-json-stable-stringify
   5.52x faster than escapeHtml w/ json-stable-stringify
   5.64x faster than escapeHtml w/ json-stringify-deterministic
   5.71x faster than js-string-escape
   13.84x faster than jsesc
   18.79x faster than escapeHtml w/ superjson.stringify
```


packages/remix-server-runtime/markup.ts
```diff
@@ -0,0 +1,16 @@
...
+  "&": "\\u0026",
+  ">": "\\u003e",
+  "<": "\\u003c",
+  "\u2028": "\\u2028",
+  "\u2029": "\\u2029",
+};
+
+const ESCAPE_REGEX = /[&><\u2028\u2029]/g;
+
+export function escapeHtml(html: string) {
```
redabacha, line 14: chose this file name and function name based on https://github.com/remix-run/remix/blob/6e1dc4929dc22fa5d6e021d312a45587c7864031/packages/remix-react/markup.ts

packages/remix-dev/package.json
```diff
@@ -64,6 +64,7 @@
     "@types/gunzip-maybe": "^1.4.0",
     "@types/inquirer": "^8.2.0",
     "@types/jscodeshift": "^0.11.3",
+    "@types/jsesc": "^3.0.1",
```
redabacha, line 67: this should've been here already, also bumped the version to match the actual major version of `jsesc` being used here

MichaelDeBoey (approved):

packages/remix-dev/package.json
MichaelDeBoey: Can't we just delete it here? 🤔

redabacha: it's being used [here](https://github.com/remix-run/remix/blob/c0cf73a1995a39ce064942aca3d5b3d89f6ec404/packages/remix-dev/compiler/plugins/serverAssetsManifestPlugin.ts#L38) in the `serverAssetsManifestPlugin` esbuild plugin. i'm not sure why or what it's needed for though so i've just opted to leave it as is.

redabacha:
the single test failure seems unrelated. i'm also seeing it happen in other prs e.g. https://github.com/remix-run/remix/runs/7671223447?check_suite_focus=true

jacob-ebey:
Going to have to double check, but I believe the serialization method needs to be consistent across machines / runtimes. I.e, `serialize(obj)` should return the same thing on the server and in the browser. `JSON.stringify` alone will not suffice here as it does not take into account key order during serialization leading to possible mismatchs on hydration.

redabacha:
> Going to have to double check, but I believe the serialization method needs to be consistent across machines / runtimes. I.e, `serialize(obj)` should return the same thing on the server and in the browser. `JSON.stringify` alone will not suffice here as it does not take into account key order during serialization leading to possible mismatchs on hydration. Going to need something like [json-stable-stringify](https://github.com/substack/json-stable-stringify)

according to mdn, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description:
> Enumerable properties are visited using the same algorithm as [Object.keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys), which has a well-defined order and is stable across implementations. For example, JSON.stringify on the same object will always produce the same string, and JSON.parse(JSON.stringify(obj)) would produce an object with the same key ordering as the original (assuming the object is completely JSON-serializable).

so my guess is that it would be fine. next.js seem to be getting away with just using [JSON.stringify](https://github.com/vercel/next.js/blob/34f5236f572b9f6a2e3c55a537a6a717ffaae71d/packages/next/pages/_document.tsx#L969-L1002) (and [JSON.parse on the client](https://github.com/vercel/next.js/blob/0f99768f9811fdf027a083471bd2a435f6207f4a/packages/next/client/index.tsx#L197-L199)).

redabacha:
@jacob-ebey i've added additional benchmarks with [`json-stable-stringify`](https://www.npmjs.com/package/json-stable-stringify) and some other stable JSON.stringify packages i've found. according to my understanding though (see prior comment), they shouldn't be necessary in this day and age.

packages/remix-server-runtime/markup.ts
```diff
@@ -0,0 +1,16 @@
+// This escapeHtml utility is based on https://github.com/zertosh/htmlescape
```
kentcdodds, line 1: Could we get a comment here explaining why we're not just using that module?

kentcdodds (changes requested):
Thanks! Just one request.

packages/remix-server-runtime/markup.ts
kentcdodds: Also, I'm curious why that module hasn't simply been updated to improve its own performance 🤔

jacob-ebey: I'm down to inline it personally. It's small and simple enough, and reducing deps in general is good IMO.

kentcdodds: Oh, I'm fine with that too. I just know it's a question I'd have if I was working in this code and I think a comment would help.

kentcdodds (approved):
Looks good to me 👍

packages/remix-server-runtime/markup.ts
redabacha: i've added a comment [here](https://github.com/remix-run/remix/pull/3889/commits/3cc17531ecef87333e3b0611951796b28df538b3) :sweat_smile:

PR #3907, opened 2022-08-02
feat(remix-react): export `SerializeType` & `SerializeObject` types

MichFont:
…synergy

This PR is simply to allow for the export of SerializeType and SerializeObject. After release 1.6.5 I created my own custom Serialize component, however, it closely mirrors the implementation that remix-react has and I was sad that it was not publicly accessible. Thus this PR.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

Reran existing tests, all tests passed.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -276,6 +276,7 @@
 - michaeldeboey
 - michaelfriedman
 - michaseel
+- MichFont
 - mikeybinnswebdesign
 - mirzafaizan
 - mjackson
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1338,7 +1338,7 @@ type JsonPrimitives =
   | null;
 type NonJsonPrimitives = undefined | Function | symbol;
 
-type SerializeType<T> = T extends JsonPrimitives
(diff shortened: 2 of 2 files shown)


MichFont:
@kentcdodds @ryanflorence @mcansh @MichaelDeBoey - Test failed and it does not appear that I have the ability to rerun a test. I believe this is a false positive based on the minor change I made. Please review and confirm

MichaelDeBoey (commented):
I don't think we want this, as we're going to change this implementation to [`type-fest`](https://github.com/sindresorhus/type-fest)'s [`Jsonify` type](https://github.com/sindresorhus/type-fest/blob/main/source/jsonify.d.ts)

kentcdodds:
Yeah, I'm not sure we're ready to expose this as public API. Sorry.

MichFont:
Any idea when remix will convert to use type-fest? @kentcdodds @MichaelDeBoey

kentcdodds:
Nobody's currently working on that, so I suppose it'll happen when type-fest supports our use cases and someone opens a PR to convert us over.

PR #3915, opened 2022-08-03
test: windows testing updates

mcansh:
this PR does the following:

- increase jest timeout on windows
- fixes a cache issue where we'd always use the latest playwright version as the cache key
- allows stacks + deployment tests to continue if one of them fails
- enable playwright test retries locally so we can check for flakey tests
- uses default workers based on machine config
- mark upload test as slow
- update playwright timeout on windows
- enable fullyParallel

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

CI and locally using Git Bash and PowerShell in windows terminal

### DIFF
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -137,7 +137,7 @@ jobs:
         id: playwright-version
         shell: bash
         run: |
-          playwright_version=$(npm info @playwright/test version)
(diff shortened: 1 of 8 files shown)


jacob-ebey (approved):

MichaelDeBoey (approved):

kentcdodds (approved):
Thanks!

mcansh:
we’ll that’s a positive start, first try it all worked

PR #3917, opened 2022-08-03
chore: upgrade core packages to react-18

jacob-ebey:
This upgrades our core packages to react 18 with minimal changes to everything else.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy: Existing tests should cover everything.


mcansh:
do we need to do the same changes to the cloudflare integration set up as the node one?

packages/remix-react/components.tsx
```diff
@@ -719,7 +719,7 @@ export function Meta() {
         }
 
         if (name === "title") {
-          return <title key="title">{value}</title>;
+          return <title key="title">{value as string}</title>;
```
mcansh, line 722: nitpick, but i think i prefer `String(value)` here instead of casting

mcansh, line 722: nitpick, but i think i prefer this here instead of casting

```suggestion
          return <title key="title">{String(value)}</title>;
```

jacob-ebey:
> do we need to do the same changes to the cloudflare integration set up as the node one?

Not at the moment. Nothing that uses that suite does react 18 things.

MichaelDeBoey:
@jacob-ebey We didn't update the Cloudflare & Deno integration test helpers.
Was that intentional?
If so, for what specific reason?

Just trying to understand why those weren't updated (yet).

PR #3918, opened 2022-08-03
feat(remix-react): unblock hydration via async module scripts

jacob-ebey:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.changeset/wild-meals-beg.md
+++ b/.changeset/wild-meals-beg.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Unblock hydration via use of async module scripts.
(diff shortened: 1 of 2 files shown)


ryanflorence:
Jacob and I chatted and decided to make it look like this:

```html
<html>
  <link rel="modulepreload" href="deps" />
  <link rel="modulepreload" href="deps" />
  <link rel="modulepreload" href="deps" />
  <link rel="modulepreload" href="deps" />

  <link rel="modulepreload" href="entry" />
  <link rel="preload" as="script" href="manifest" />

  <script async type=module>
    import "./manifest";
    import route1 from "./route1";
    import route2 from "./route2";
    import(entry)
  </script>
</html>
```

MichaelDeBoey (approved):

PR #3923, opened 2022-08-04
feat(remix-dev): import `.gql` & `.graphql` files as text

MichaelDeBoey:
Follow-up of https://github.com/remix-run/remix/pull/3190#discussion_r873095942

### DIFF
--- a/.changeset/four-poets-stare.md
+++ b/.changeset/four-poets-stare.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": minor
+---
+
+feat(remix-dev): import `.gql` & `.graphql` files as text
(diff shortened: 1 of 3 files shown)


sergiodxa:
I think this should be as text not as file

As file will give you the path to the file, as text will give you the content as a string

I think what users will want is to get the query string as a text so it can be used directly to send a query to the API, if you get a path to a file you will have to read it, and AFAIK the path will be a URL

MichaelDeBoey:
@sergiodxa Oops a bit too quick here 😅🙈

Should be good now

machour:
@MichaelDeBoey shouldn't `packages/remix-dev/modules.ts` be updated too?

MichaelDeBoey:
@machour it wasn't updated in #3190, but it should I think?

CC/ @jacob-ebey

jacob-ebey:
> @machour it wasn't updated in #3190, but it should I think?
> 
> CC/ @jacob-ebey

Yes, it should be updated.

MichaelDeBoey:
@jacob-ebey This should be fine to merge now

machour (approved):

PR #3924, opened 2022-08-04
chore(examples/blog-tutorial): revert update of `@testing-library/react` dependency

MichaelDeBoey:
This shouldn't have been done in #3917, as that example is still using React v17


kentcdodds:
We need to get all the examples and things on React 18

MichaelDeBoey:
@kentcdodds On it 👍

PR #3929, opened 2022-08-05
tests(remix-dev): update to React v18

MichaelDeBoey:
This shouldn't have been done in #3917, as `package.json` is updated to React v18


kentcdodds (approved):
Solid 👍

PR #3930, opened 2022-08-05
feat(templates): update to React v18

MichaelDeBoey:
Follow-up of #3917 

I didn't update Cloudflare & Deno templates as those weren't updated in #3917 by @jacob-ebey as well (see https://github.com/remix-run/remix/pull/3917#issuecomment-1206331413 for more questions about the reasoning behind this)

### DIFF
--- a/templates/arc/app/entry.client.tsx
+++ b/templates/arc/app/entry.client.tsx
@@ -1,4 +1,4 @@
 import { RemixBrowser } from "@remix-run/react";
-import { hydrate } from "react-dom";
+import { hydrateRoot } from "react-dom/client";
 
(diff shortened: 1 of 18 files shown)


kentcdodds (approved):
Looks good 👍

MichaelDeBoey:
@kentcdodds Any idea if we should update Cloudflare/Deno templates + test fixtures as well?

kentcdodds:
For anyone looking at this later and saying: "Wait a second, Netlify, Vercel, and Arc are on AWS which doesn't stream!" I merged this because AFAIK you still need that if you want to support React.lazy.

PR #3936, opened 2022-08-05
feat(remix-dev): add `tsconfigPath` config option to RemixConfig

### LINKED ISSUE #3210: Build fails if there's a broken tsconfig.json file outside the project

latest

### Steps to Reproduce

- npx create-remix@latest (just the basics, typescript)
- cd my-remix-app
- touch ../tsconfig.json
- npm run dev

### Expected Behavior

The build should work fine, and the tsconfig.json file outside my project should be ignored

### Actual Behavior

🍺  ~/my-remix-app npm run dev

> dev
> remix dev

Watching Remix app in development mode...

✘ [ERROR] Unexpected end of file

    ../tsconfig.json:1:0:
      1 │
        ╵ ^

Build failed with 1 error:
../tsconfig.json:1:0: ERROR: Unexpected end of file

jwnx:
This patch addresses an issue[1] in which Remix looks for a tsconfig
outside the root directory.

We fix this behavior by looking for a tsconfig where it should be
(in the root directory itself) and passing its absolute path to
`tsconfig-paths`. `tsconfig-paths` will then treat the given path as a
tsconfig file path and won't crawl further. If no tsconfig file is
found by us in our root directory, we simply won't call `tsconfig-paths`
at all. Note that crawling up is `tsconfig-path`s intended behavior,
but not Remix's.

A secondary bug that causes this same issue is due to esbuild[2]. When
no tsconfig is passed to esbuild, it will crawl the filesystem up looking
for any tsconfig available and will load it. When tsconfig is explicitly
set to `undefined`, it does the crawling anyways. Unfortunately,
this is a fix that has to be done in `esbuild`, but a few alternatives
to manage this unwanted behavior for now can be:

1. Create an empty tsconfig in the build directory and pass it to esbuild
   so it doesn't crawl outside rootDir;
2. Show a warning to let the user know why the crawling is happening.

Closes #3210

[1] https://github.com/remix-run/remix/issues/3210
[2] https://github.com/evanw/esbuild/issues/2440

Signed-off-by: Juliana Oliveira <juliana@uma.ni>


jwnx:
(`dev` is already failing for windows unit and integration tests)

packages/remix-dev/config.ts
```diff
@@ -450,6 +455,15 @@ export async function readConfig(
 
   let serverDependenciesToBundle = appConfig.serverDependenciesToBundle || [];
 
+  // When tsconfigPath is undefined, the default "tsconfig.json" is not
+  // found in the root directory.
+  let tsconfigDefaultFilename = "tsconfig.json"
```
mcansh, line 460: do you happen to know if this all works as expected if you use a jsconfig.json file and pass that to esbuild?

mcansh: it appears that it does, can you update this to also check for `jsconfig.json`? i think something like this should be sufficient https://github.com/remix-run/remix/pull/3012/files#diff-852e3f3cc491a6fcc66706c94417be59f4c12569e0f367354613bfc103df7b38R299-R307

jwnx: Sure. There you go!

mcansh: :chefkiss:

mcansh (approved):

jwnx:
@mcansh I can't merge this. How is this done?

mcansh:
> 

getting this over the line now :)

PR #3949, opened 2022-08-07
chore(examples/theme-ui): fix styling

scottybrown:
We expect the [theme-ui example](https://github.com/remix-run/remix/tree/main/examples/theme-ui)'s pages to have a stark blue background, to demonstrate that the theme-ui theme is being applied.

Instead, it has no styling at all.

The cause is [this PR](https://github.com/remix-run/remix/pull/3664), which removes some lines which are fundamental to [how theme-ui works](https://theme-ui.com/getting-started#classic-jsx-runtime).

I have also removed 2 entries from a `useEffect` dependency array which were also added in the mentioned PR, which were causing repeated `Maximum update depth exceeded` to appear in the browser console. This does not seem to affect this example, but can cause UI hangs on apps with state.

I've been unsuccessful in running linting against this example, so I can't verify whether my changes have brought back lint warnings. But they've certainly made the example more useful :)

Testing Strategy:
- [ ] Test that blue background appears on `/` and `/jokes`/
- [ ] Test that `maximum update depth exceeded` does not repeatedly appear in browser console

### DIFF
--- a/examples/theme-ui/app/root.tsx
+++ b/examples/theme-ui/app/root.tsx
@@ -27,20 +27,12 @@ const Document = withEmotionCache(
   ({ children }: DocumentProps, emotionCache) => {
     const serverStyleData = useContext(ServerStyleContext);
     const clientStyleData = useContext(ClientStyleContext);
+    const resetClientStyleData = clientStyleData?.reset;
 
     // Only executed on client
     useEffect(() => {
-      // re-link sheet container
-      emotionCache.sheet.container = document.head;
-      // re-inject tags
-      const tags = emotionCache.sheet.tags;
-      emotionCache.sheet.flush();
-      tags.forEach((tag) => {
-        (emotionCache.sheet as any)._insertTag(tag);
-      });
-      // reset cache to reapply global styles
-      clientStyleData?.reset();
-    }, []);
+      resetClientStyleData();
+    }, [resetClientStyleData]);
 
     return (
       <html lang="en">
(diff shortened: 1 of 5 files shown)


examples/theme-ui/app/root.tsx
```diff
@@ -40,7 +40,7 @@ const Document = withEmotionCache(
       });
       // reset cache to reapply global styles
       clientStyleData?.reset();
-    }, [clientStyleData, emotionCache.sheet]);
+    }, []);
```
MichaelDeBoey, line 43: This causes an ESLint warning
https://github.com/remix-run/remix/runs/7713555214?check_suite_focus=true#step:6:11

```
Warning:   43:8  warning  React Hook useEffect has missing dependencies: 'clientStyleData' and 'emotionCache.sheet'. Either include them or remove the dependency array  react-hooks/exhaustive-deps
```

Can we make sure to not have an `eact-hooks/exhaustive-deps` warning + not disable it, but make the example work please?

MichaelDeBoey (changes requested):

examples/theme-ui/app/root.tsx
scottybrown: Hmm. I had a go, I've been unable to find a solution that doesn't involve infinite looping. I'm going to keep trying, it might need a bit of refactoring to achieve.
In the meantime, do you think it's worth raising the loop issue separately? So we can fix the styling without it also depending on this tricky change.

MichaelDeBoey: I'd love to either have this fixed _or_ set the dependency array correct again.

kentcdodds (approved):
Thank you!

PR #3953, opened 2022-08-07
fix(remix-react): switch `name` & `content` position in `meta` tags

arnaudambro:
When we visualize the html meta tags on a browswer inspector, I think it's better to see
```html
<meta name="twitter:card" content="summary">
```
than
```html
<meta content="summary" name="twitter:card">
```

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -33,6 +33,7 @@
 - archwebio
 - arganaphangquestian
 - AriGunawan
+- arnaudambro
 - arvigeus
 - arvindell
 - ascorbic
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -729,15 +729,15 @@ export function Meta() {
           if (isOpenGraphTag) {
             return (
               <meta
+                property={name}
                 content={content as string}
                 key={name + content}
-                property={name}
               />
             );
           }
 
           if (typeof content === "string") {
-            return <meta content={content} name={name} key={name + content} />;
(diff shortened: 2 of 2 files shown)


machour (approved):

machour:
@kentcdodds This is purely cosmetics, but always have bothered me too :)

MichaelDeBoey (approved):
I personally wouldn't merge this one as this is indeed purely cosmetic, but I can see a value if a lot of people are bothered by this.

arnaudambro:
> I personally wouldn't merge this one as this is indeed purely cosmetic, but I can see a value if a lot of people are bothered by this.

I didn't get this comment :)

This is not purely cosmetic, this is about better debugging, better DX : when you have to go through 50 `meta` tags and you have `content` first, whereas you are looking for `twitter:image:alt` meta tag, then it takes twice as much time if the order is `content` then `name` than the opposite.

I know it's not the PR of the year, but... it's a PR improving something, not just cosmetic !

MichaelDeBoey:
> whereas you are looking for `twitter:image:alt` meta tag, then it takes twice as much time if the order is `content` then `name` than the opposite.

I would instinctively do a `CMD` + `F` to search for what I'm looking for, but as I said that's my personal opinion.

Going to let the team decide what they want to do with this as I don't want to push my personal opinion/preference for everyone.

kentcdodds (approved):
Thanks for the contribution.

PR #3963, opened 2022-08-08
fix(remix-dev): convert config file path to file url

mcansh:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.changeset/stale-worms-laugh.md
+++ b/.changeset/stale-worms-laugh.md
@@ -1,0 +1,8 @@
+---
+"remix": patch
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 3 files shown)


kentcdodds (approved):
Thanks 👍

kentcdodds:
Whoops, looks like the tests don't like that change 😅

mcansh:
> Whoops, looks like the tests don't like that change 😅

need a toString, fixed it locally for a manual test run, but in an actual test it's not happy still, looking into it :)

mcansh:
seems like there is a jest problem somewhere, the "same" (slight modifications) tests work no problem using `node:test`

chaance (approved):

kentcdodds (approved):
Thanks!

PR #3966, opened 2022-08-08
fix(remix-dev): make MDX builds deterministic

jenseng:
This is a fix along the lines of #2027; because the MDX virtual modules resolve to absolute paths, the manifest/hashes/etc. differ based on the parent directory structure.

Refs: #2024

- [x] Tests

Testing Strategy:
1. Expanded existing integration test to cover this scenario (fails without the fix). Also add other virtual module scenarios that weren't already covered
2. Validated in a real MDX Remix app that has split client / server builds (applied fix via patch-package)

### DIFF
--- a/.changeset/purple-ligers-drive.md
+++ b/.changeset/purple-ligers-drive.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Make MDX builds deterministic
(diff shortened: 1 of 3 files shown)


kentcdodds (approved):
Yup 👍 Thank you as always @jenseng!

PR #3980, opened 2022-08-11
ci: use graphql api for tags 

mcansh:
~~updates how tags are fetched for commenting purposes, it now uses the [github git db endpoint](https://api.github.com/repos/remix-run/remix/git/refs/tags) so we get all the tags at once, but we still need to make 2 requests for each of those as we don't get the commit sha from the git db api and we don't get a date from the until you have a commit sha.~~

~~so i think we can filter out unneeded nightly releases by their date in the tag and unneeded stable releases if they're older than the previous version, which should leave us with only a handful of requests~~

edit: on second thought, we shouldn't need to paginate for tags, as realistically we won't cut 100 tags in between releases.

edit 2: jk, we do need to make changes to account for prefixed tags `remix@`

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


kentcdodds (approved):
Nice 👍 Thanks!

PR #3986, opened 2022-08-12
fix: Avoid path collisions between pathless and index routes

brophdawg11:
This fixes a path collision bug when using pathless layout routes and index routes at the same level currently.

The routes folder structure:
```
app/routes
└── nested
    ├── __pathless
    │   └── foo.jsx
    ├── __pathless.jsx
    └── index.jsx
```

Currently produces a route hierarchy of:
```jsx
<Route file="root.tsx">
  <Route path="nested/" file="routes/nested/__pathless.tsx">
    <Route path="foo" file="routes/nested/__pathless/foo.tsx" />
  </Route>
  <Route path="nested" index file="routes/nested/index.tsx" />
  <Route index file="routes/index.tsx" />
</Route>
```

However, you can see that we're creating _two_ routes with `path="nested"` and we also have a route with both a `path` and an `index`.

Instead, we should be detecting that `routes/nested/index` and `routes/nested/__pathless.tsx` both share the same parent `nested` path and they should be nested together under `<Route path="nested">`.

We don't want to have to start representing the `routes/nested` folder in the asset manifest, since it's not an asset, so we can do this determination as hierarchical route creation time and just insert parent routes as needed.  By inserting the parent route and removing `path` from the two pathless route children, we end up with the proper hierarchy of:

```jsx
<Route file="root.tsx">
  <Route path="nested">
    <Route file="routes/nested/__pathless.tsx">
      <Route path="foo" file="routes/nested/__pathless/foo.tsx" />
    </Route>
    <Route index file="routes/nested/index.tsx" />
  </Route>
  <Route index file="routes/index.tsx" />
</Route>
```

Closes: #3014

Testing Strategy:
* Unit tests for `formatRoute` to ensure we capture this in the CLI
* Integration test brought over from #3052

### DIFF
--- a/packages/remix-dev/config/format.ts
+++ b/packages/remix-dev/config/format.ts
@@ -148,11 +148,11 @@ export function createHierarchicalRoutes<
           otherPathRoutes.push(r);
         }
       });
-      // TODO: Need to figure out this typing error :/
-      // @ts-expect-error
-      let folderRoute: HierarchyRoute = {
-        id: `routes/${path}`,
-        path,
+      let folderRoute = {
+        ...createRoute({
+          id: `routes/${path}`,
+          path,
+        } as ManifestRoute),
         children: dupPathRoutes.map((r) => ({ ...r, path: undefined })),
       };
       children = [...otherPathRoutes, folderRoute];
--- a/packages/remix-dev/config/format.ts
+++ b/packages/remix-dev/config/format.ts
@@ -84,11 +84,11 @@ interface BaseManifestRoute {
   parentId?: string;
 }
 
-interface BaseHierarchyRoute {
+type BaseHierarchyRoute<T> = T & {
   id: string;
   path?: string;
-  children?: BaseHierarchyRoute[];
-}
+  children?: BaseHierarchyRoute<T>[];
+};
 
 /**
  * NOTE: This function is duplicated in remix-dev, remix-react, and
@@ -110,11 +110,11 @@ interface BaseHierarchyRoute {
  */
 export function createHierarchicalRoutes<
   ManifestRoute extends BaseManifestRoute,
-  HierarchyRoute extends BaseHierarchyRoute
+  HierarchyRoute extends Omit<BaseHierarchyRoute<unknown>, "children">
 >(
   manifest: Record<string, ManifestRoute>,
   createRoute: (r: ManifestRoute) => HierarchyRoute
-) {
+): HierarchyRoute[] {
   function recurse(parentId?: string) {
     let routes = Object.values(manifest).filter(
       (route) => route.parentId === parentId
@@ -131,9 +131,10 @@ export function createHierarchicalRoutes<
       if (route.index && route.path) {
         indexRoutesWithPath.push(route.path);
       }
-      let hierarchicalRoute = createRoute(route);
-      hierarchicalRoute.children = recurse(route.id);
-      children.push(hierarchicalRoute);
+      children.push({
+        ...createRoute(route),
+        children: recurse(route.id),
+      });
     }
 
     // For each index route that _also_ had a path, create a new parent route
--- a/packages/remix-dev/config/format.ts
+++ b/packages/remix-dev/config/format.ts
@@ -79,6 +79,7 @@ export function formatRoutesAsJsx(routes: JsonFormattedRoute[]) {
 
 interface BaseManifestRoute {
   id: string;
+  index?: boolean;
   path?: string;
   parentId?: string;
 }
@@ -107,7 +108,7 @@ interface BaseHierarchyRoute {
  *                     ignoring children
  * @returns
  */
-function createHierarchicalRoutes<
+export function createHierarchicalRoutes<
   ManifestRoute extends BaseManifestRoute,
   HierarchyRoute extends BaseHierarchyRoute
 >(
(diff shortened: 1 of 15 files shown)


packages/remix-dev/config/format.ts
```diff
@@ -13,11 +13,22 @@ export function formatRoutes(
   routeManifest: RouteManifest,
   format: RoutesFormat
 ) {
+  let routes = createHierarchicalRoutes<ConfigRoute, JsonFormattedRoute>(
```
brophdawg11, line 16: Create a hierarchy once and then print from the same structure with the right format

packages/remix-dev/config/format.ts
```diff
@@ -90,9 +70,95 @@ export function formatRoutesAsJsx(routeManifest: RouteManifest) {
...
+ *                     ignoring children
+ * @returns
+ */
+function createHierarchicalRoutes<
+  ManifestRoute extends BaseManifestRoute,
+  HierarchyRoute extends BaseHierarchyRoute
+>(
+  manifest: Record<string, ManifestRoute>,
+  createRoute: (r: ManifestRoute) => HierarchyRoute
+) {
```
brophdawg11, line 116: Generic function to convert a flat manifest to a hierarchical structure, accepting a function to convert the manifest route into the hierarchical route.  Ideally this would be shared amongst remix-dev/remix-server-runtime/remix-react but is duped for now.  This contains the logic to insert parent routes as needed to disambiguate pathless routes

packages/remix-dev/config/routesConvention.ts
```diff
@@ -188,6 +188,8 @@ export function createRoutePath(partialRouteId: string): string | undefined {
 
   if (rawSegmentBuffer === "index" && result.endsWith("index")) {
     result = result.replace(/\/?index$/, "");
+  } else {
+    result = result.replace(/\/$/, "");
```
brophdawg11, line 192: I don't _think_ this has ripple effects but would love confirmation.  This normalizes the `path` value for index versus pathless layout routes.  Currently, index routes lose the trailing slash and pathless layout routes keep it.  I figure `<Route path="nested">` seems more idiomatic (user's don't usually include the trailing slash) so I opted to go that route.

Making sure they match is currently required for the logic to detect collisions in `createHierarchicalRoutes`.

packages/remix-react/components.tsx
```diff
@@ -707,7 +707,7 @@ export function Meta() {
 
     let routeModule = routeModules[routeId];
 
-    if (routeModule.meta) {
+    if (routeModule?.meta) {
```
brophdawg11, line 710: Now that we have "Routes" in our application that do not have a module, we do have to be defensive when accessing module information

packages/remix-dev/config/format.ts
```diff
@@ -90,9 +70,95 @@ export function formatRoutesAsJsx(routeManifest: RouteManifest) {
...
+        let otherPathRoutes: HierarchyRoute[] = [];
+        let dupPathRoutes: HierarchyRoute[] = [];
+        children.forEach((r) => {
+          if (r.path === path) {
+            dupPathRoutes.push(r);
+          } else {
+            otherPathRoutes.push(r);
+          }
+        });
+        // TODO: Need to figure out this typing error :/
```
brophdawg11, line 150: @pcattori May need your eyes on this guy 😬

packages/remix-server-runtime/server.ts
```diff
@@ -176,14 +180,14 @@ async function handleDocumentRequest({
...
-  routes,
+  rootRoute,
   serverMode,
 }: {
   build: ServerBuild;
   loadContext?: AppLoadContext;
   matches: RouteMatch<ServerRoute>[] | null;
   request: Request;
-  routes: ServerRoute[];
+  rootRoute?: ServerRoute;
```
brophdawg11, line 190: We only ever use the root route inside here so I updated for clarity and to reduce the scope of the server-runtime usage of hierarchical `routes`

alex-ketch:
In case you hadn't seen it yet @brophdawg11, I created a related issue on react-router side: https://github.com/remix-run/react-router/issues/9145.
There I was able to get things working for my needs by adjusting the route scores.
Sharing in case it helps, or you'd like to close the issue once you're done.
Thanks for working on this :)

brophdawg11:
@alex-ketch I did!  That's what sent me down this investigation path :)  Indeed as per [https://github.com/remix-run/remix/discussions/3014#discussioncomment-2666749] it is due to a bug in the way remix generates the routes it hands off to react-router.  So we won't need to change the scoring after all.  Thanks for reporting!  I'll close that out and link over to here

packages/remix-react/components.tsx
brophdawg11: I don't like how easy this is to miss - TS doesn't yell at us when accessing `routeModule.meta` non-defensively, so I may want to look a bit closer at the types there and see what options we have

brophdawg11: Update `RouteModules` type to include `| undefined`

tony-lb:
Hi @brophdawg11  it's there a pre-release that contains this change? I encountered a similar problem not locally in dev, but on a deployed app. Just wanted to make sure this pr also solves my case.

corywhite999:
> Hi @brophdawg11 it's there a pre-release that contains this change? I encountered a similar problem not locally in dev, but on a deployed app. Just wanted to make sure this pr also solves my case.

Hey @brophdawg11, I ran into this today. Any idea when this might be merged? Thanks.

brophdawg11:
Hey folks - sorry this fell behind a few other things on my plate with the 6.4 release.  There's actually a much simpler fix for this based on some recent discussions over in react-router, so I'm going to close this PR out in favor of #4355 and will try to get that into a release soon!

PR #3987, opened 2022-08-12
fix(remix-dev/cli/migrate): fix `convert-to-javascript` migration

MichaelDeBoey:
This will partially resolve https://github.com/remix-run/indie-stack/issues/134

### DIFF
--- a/packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createVariableDeclarationIdentifier.ts
+++ b/packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createVariableDeclarationIdentifier.ts
@@ -1,10 +1,20 @@
 import type { ImportDeclaration, JSCodeshift } from "jscodeshift";
 
+/**
+ * import foo from "foo"
+ * import * as foo from "foo"
+ * =>
+ * const foo = require("foo").default
+ * const foo = require("foo")
+ */
 export const createVariableDeclarationIdentifier = (
   j: JSCodeshift,
   { source, specifiers }: ImportDeclaration
 ) => {
   let callExpression = j.callExpression(j.identifier("require"), [source]);
+  let isDefaultImport = (specifiers || []).some(
+    ({ type }) => type === "ImportDefaultSpecifier"
+  );
 
   return j.variableDeclaration("const", [
     j.variableDeclarator(
@@ -24,7 +34,9 @@ export const createVariableDeclarationIdentifier = (
               : []
           )[0].local?.name || ""
       ),
-      callExpression
+      isDefaultImport
+        ? j.memberExpression(callExpression, j.identifier("default"))
+        : callExpression
     ),
   ]);
 };
--- a/packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/index.ts
+++ b/packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/index.ts
@@ -1,7 +1,9 @@
 import type { Transform } from "jscodeshift";
 
 import { checkNoDifferentImportTypesCombined } from "./checkNoDifferentImportTypesCombined";
-import { createExpressionStatement } from "./createExpressionStatement";
+import { createExportExpressionStatementFromExportDefaultDeclaration } from "./createExportExpressionStatementFromExportDefaultDeclaration";
+import { createExportExpressionStatementFromExportNamedDeclaration } from "./createExportExpressionStatementFromExportNamedDeclaration";
+import { createImportExpressionStatement } from "./createImportExpressionStatement";
 import { createVariableDeclarationIdentifier } from "./createVariableDeclarationIdentifier";
 import { createVariableDeclarationObjectPattern } from "./createVariableDeclarationObjectPattern";
 
@@ -9,9 +11,15 @@ const transform: Transform = (file, api, options) => {
   let j = api.jscodeshift;
   let root = j(file.source);
 
-  let allESImportDeclarations = root.find(j.ImportDeclaration);
-  if (allESImportDeclarations.length === 0) {
-    // This transform doesn't need to run if there are no ES imports
+  let allImportDeclarations = root.find(j.ImportDeclaration);
+  let allExportDefaultDeclarations = root.find(j.ExportDefaultDeclaration);
+  let allExportNamedDeclarations = root.find(j.ExportNamedDeclaration);
+  if (
+    allImportDeclarations.length === 0 &&
+    allExportDefaultDeclarations.length === 0 &&
+    allExportNamedDeclarations.length === 0
+  ) {
+    // This transform doesn't need to run if there are no ES imports/exports
     return null;
   }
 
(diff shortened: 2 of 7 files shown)


packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createExportExpressionStatement.ts
```diff
@@ -0,0 +1,71 @@
+import type { ExportDefaultDeclaration, JSCodeshift } from "jscodeshift";
+
+export const createExportExpressionStatement = (
+  j: JSCodeshift,
+  exportDefaultDeclaration: ExportDefaultDeclaration
+) => {
+  /**
+   * HACK: Can't use casts nor type guards in a `jscodeshift` transform
```
MichaelDeBoey, line 8: We know for sure that `exportDefaultDeclaration.declaration` isn't of `DeclarationKind`, but there's unfortunately no other way than checking each possible union value separately, since we can't cast in a `jscodeshift` transform

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createVariableDeclarationIdentifier.ts
```diff
@@ -24,7 +27,9 @@ export const createVariableDeclarationIdentifier = (
               : []
           )[0].local?.name || ""
       ),
-      callExpression
+      isDefaultImport
+        ? j.memberExpression(callExpression, j.identifier("default"))
+        : callExpression
```
MichaelDeBoey, line 32:
```ts
import Foo from 'foo';
import * as Bar from 'bar';

// =>

const Foo = require('foo').default;
const Bar = require('bar');
```

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/index.ts
```diff
@@ -48,6 +53,12 @@ const transform: Transform = (file, api, options) => {
     );
   });
 
+  allExportDefaultDeclarations.forEach((exportDefaultDeclaration) => {
+    j(exportDefaultDeclaration).replaceWith(
+      createExportExpressionStatement(j, exportDefaultDeclaration.node)
+    );
```
MichaelDeBoey, line 59:
```ts
export default foo;

// =>

module.exports = foo;
```

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createExportExpressionStatement.ts
kentcdodds: Is this just to make TypeScript happy with this code or is it actually needed at runtime?

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createVariableDeclarationIdentifier.ts
kentcdodds: This won't always be correct 😬 I think we've made a big mistake by building this into our CLI. There are too many edge cases 😬

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createExportExpressionStatement.ts
MichaelDeBoey: The whole hack is to make TS happy 😢

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createVariableDeclarationIdentifier.ts
MichaelDeBoey: > This won't always be correct 😬

Can you give some examples where this won't be correct?
I'll try to add them.

> I think we've made a big mistake by building this into our CLI.

We wanted to support JS & TS & (as you know) having this migration was the cleanest way of converting all files in a template/stack to JS.
Otherwise each stack had to do the conversion on its own.

I think the decision for doing it in our CLI was a good one, but it's a pain to think about every possible thing. 🙈

> There are too many edge cases 😬

If only there was a package/library that could convert to ES5 without making the file a mess 🤔😅

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createExportExpressionStatement.ts
MichaelDeBoey: Found a way to convert these as well

packages/remix-dev/cli/migrate/migrations/convert-to-javascript/transform/createVariableDeclarationIdentifier.ts
pcattori: I do think that the TS->JS conversion is better suited as a standalone tool, in its own repo, provided as-is without covering all edge cases, rather than being included in our migrations. I think "migrations" concept in our CLI should be reserved for helping automate upgrading through breaking changes.

pcattori: That said, we currently offer the `--no-typescript` option during create... so we'd have to deprecate / remove that option to get out of the business of doing the TS->JS conversion.

MichaelDeBoey: We supported both JS & TS version of our templates/stacks from v1.0.0, so that's why I added the new `convert-to-javascript` migration.

Now that we come across these extra changes that need to be made, I can see a separate package working for stacks.
We would still have the "problem" that our templates will need to use it as well.

We could definitely stop officially supporting JS, but that would indeed be a breaking change, so needs to wait for v2.

If we want, we can already create a separate package that's solely responsible for the JS conversion (or we could put it in a `scripts` package if that's preferred) & let the CLI use this new package.

kentcdodds: I do like the idea of moving this all to a separate package so it can be quickly iterated upon and remix can depend on that package until v2 when it can be removed.

MichaelDeBoey: A separate Remix package or like a separate personal package?

kentcdodds: Probably a separate Remix package I think. But anyway, I think we should probably just merge this and move forward.

kentcdodds (approved):
Thanks for your work on this 👍

kentcdodds:
Forgot to add a changeset. Added here: https://github.com/remix-run/remix/blob/dev/.changeset/red-apples-stew.md

chaance:
@MichaelDeBoey Looks like this PR failed our checks on Windows, so we probably should have held off on merging. @mcansh is looking into it, but @MichaelDeBoey if you have any instincts as to why (since you wrote the code), let me know!

MichaelDeBoey:
@chaance @mcansh The only thing I can think of is that it's just taking too long on Windows, since we now do extra transforms 🤔

So increasing the timeout would help I guess.

mcansh:
it doesn't appear to be a timeout, it's set to 20s but we're failing at around 9.5s

this is the stack trace:

MichaelDeBoey:
Then it's strange, as I don't do anything that can't be done on Windows I think 🤔
It's just JavaScript + `jscodeshift` usage

mcansh:
> Then it's strange, as I don't do anything that can't be done on Windows I think 🤔 It's just JavaScript + `jscodeshift` usage
> 
> The error seems to be about creating a `FunctionExpression` from a `FunctionDeclaration`. I would assume that would just work 🤔
> 
> ```ts
>       ? j.functionExpression.from(exportDefaultDeclaration.declaration)
> ```

yeah it's weird, macOS has no troubles with it, but ast-types throws on Windows [when comparing it's internal type](https://github.com/benjamn/ast-types/blob/master/lib/types.ts#L656-L658) as it has the "built" type is `FunctionDeclaration`, but it's actual type is `FunctionExpression`, and to top it off with the same log in place macOS doesn't log anything 👻

MichaelDeBoey:
@mcansh That's really weird 🤔😅

I'll resubmit this PR with this `FunctionDeclaration` conversion excluded.
We can always add it again once it's fixed on `ast-types`' side (which I doubt since it's unmaintained for a long time already) or `jscodeshift` (which is more possible, since @Daniel15 @ElonVolo & @trivikr are working on bringing the package up to date)

Issue #3988, opened 2022-08-12
AppLoadContext type augmentation requires undefined check


justinwaite:
### What version of Remix are you using?

1.6.8

### Steps to Reproduce

Augment the server-runtime module and declare properties on the AppLoadContext interface

```typescript
declare module "@remix-run/server-runtime" {
  interface AppLoadContext {
    userId: number;
  }
}
```

### Expected Behavior

I would expect that to be able to use `context` without doing any null/undefined checks, since I've written the "guarantee" that I indeed have this context in my app.

```typescript
export async function loader({context}: LoaderArgs) {
  getSomeData(context.userId); // I would like to be able to do this
}
```

### Actual Behavior

Because the type for context is `AppLoadContext | undefined` in the loader and action args, theres' no way that I know of to tell typescript that it will never be undefined. This means that whenever you wish to use context, you must either assert its existence with `!` , or you must actually check that it exists to make typescript happy. 

```typescript
export async function loader({context}: LoaderArgs) {
  getSomeData(context!.userId); // but instead I have to do this
}
```

```typescript
export async function loader({context}: LoaderArgs) {
  invariant(context); // or this
  getSomeData(context.userId);
}
```

justinwaite:
I had a hard time trying to decide if this was an issue or a discussion, so I chose issue. Let me know if I'm wrong!

PR #3989, opened 2022-08-13
types(remix-server-runtime): make `context` mandatory in `DataFunctionArgs`

justinwaite:
Closes: #3988

My strategy for addressing this issue is to always pass an empty object for the context, rather than passing nothing (or undefined). Since we've already declared the AppLoadContext interface to be `{[key: string]: unknown}`, I feel like this would be a reasonable approach. 

- [ ] Docs
- [ ] Tests

Testing Strategy:

I created an app in the playground and (manually had to) copy over the built remix-server-runtime directory into it. I then augmented the AppLoadContext interface with a property called `notes`. I then tried to access both a known and unknown property to make sure that typescript knew the difference.

```typescript
declare module '@remix-run/server-runtime' {
  export interface AppLoadContext {
    notes: string[]
  }
}

export const loader: LoaderFunction = async ({ request, context }) => {
  console.log(context.asdf.toString());
  
  console.log(context.notes.map(n => n.toLowerCase())); // Look Ma! No undefined check for context!

  let userId = await requireUserId(request);
  let noteListItems = await getNoteListItems({ userId });
  return json<LoaderData>({ noteListItems });
};
```

Here you can see the difference. Since `asdf` isn't part of the AppLoadContext augmented interface, Typescript correctly infers it as `unknown`. However, since `notes` _was_ added to the interface, it correctly infers it as a `string[]`. Note in either case that we didn't need to add an undefined check for the context, since it will always be at least an empty object (`{}`).
Screen Shot 2022-08-12 at 6 17 05 PM
Screen Shot 2022-08-12 at 6 17 30 PM

### DIFF
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -31,7 +31,7 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
   let routes = createRoutes(build.routes);
   let serverMode = isServerMode(mode) ? mode : ServerMode.Production;
 
-  return async function requestHandler(request, loadContext) {
+  return async function requestHandler(request, loadContext = {}) {
     let url = new URL(request.url);
     let matches = matchServerRoutes(routes, url.pathname);
 
@@ -82,7 +82,7 @@ async function handleDataRequest({
   serverMode,
 }: {
   handleDataRequest?: HandleDataRequestFunction;
-  loadContext?: AppLoadContext;
+  loadContext: AppLoadContext;
   matches: RouteMatch<ServerRoute>[];
   request: Request;
   serverMode: ServerMode;
@@ -110,7 +110,7 @@ async function handleDataRequest({
       match = getRequestMatch(url, matches);
 
       response = await callRouteAction({
-        loadContext: loadContext ?? {},
+        loadContext,
         match,
         request: request,
       });
@@ -129,7 +129,7 @@ async function handleDataRequest({
       }
       match = tempMatch;
 
-      response = await callRouteLoader({ loadContext: loadContext ?? {}, match, request });
+      response = await callRouteLoader({ loadContext, match, request });
     }
 
     if (isRedirectResponse(response)) {
@@ -151,7 +151,7 @@ async function handleDataRequest({
 
     if (handleDataRequest) {
       response = await handleDataRequest(response, {
-        context: loadContext ?? {},
+        context: loadContext,
         params: match.params,
         request,
       });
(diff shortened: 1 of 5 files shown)


packages/remix-server-runtime/server.ts
```diff
@@ -110,7 +110,7 @@ async function handleDataRequest({
       match = getRequestMatch(url, matches);
 
       response = await callRouteAction({
-        loadContext,
+        loadContext: loadContext ?? {},
```
marwan38, line 113: I would default this as high up as possible. So, in the args definition  line 79) or even higher up. I'm assuming that the consumer of this function doesn't really or shouldn't be instantiating the empty object.

pcattori (changes requested):
Looks like this is on the right track!

Could you make all user-facing APIs (i.e. things exported from the main `index.ts`) be optional (`context?: AppLoadContext`) while making all the internal APIs required (`context: AppLoadContext`)?

For the user-facing optional ones, I agree with @marwan38 that you could use a default value in the function signature to simplify. For example, within the `createRequestHandler` function:

```ts
  return async function requestHandler(request, loadContext = {}) {
```

^I _think_ that might be the only place you'll need the default value of `= {}`.

justinwaite:
> Looks like this is on the right track!
> 
> We'd like to make all user-facing APIs (i.e. things exported from the main `index.ts`) optional (`context?: AppLoadContext`) while making all the internal APIs required (`context: AppLoadContext`).
> 
> The only user-facing API should be the `requestHandler` function within the `createRequestHandler` function. I think all that's needed is to change that to:
> 
> ```ts
>   return async function requestHandler(request, loadContext = {}) {
> ```
> 
> For all other places, you should be able to make the load context required.

Hey @pcattori, what exactly would you like me to change here? The interface for `RequestHandler` remains the same:
```typescript
export type RequestHandler = (
  request: Request,
  loadContext?: AppLoadContext
) => Promise<Response>;
```

The only place I'm making it _required_ is in `callRouteAction` and `callRouteLoader`. The other changes I'm just using some nullish coalescing to ensure that _something_ is passed in, i.e. an empty object, in the event that the user passes null or undefined into the `requestHandler` for the `context`.

packages/remix-server-runtime/server.ts
justinwaite: Good call. I moved it up to the `requestHandler` to be the default value for the loadContext arg. This definitely did clean it up

pcattori (approved):

MichaelDeBoey (approved):

pcattori:
> Hey @pcattori, what exactly would you like me to change here?

I think @marwan38 said _what_ should change more clearly than I did 😅 was just trying to emphasize that we want that change not only because its tidier, but because it correctly models that users should be able to omit `loadContext` arg and `requestHandler` is the only place users would do that.

pcattori:
Waiting for #4010 to fix the Windows CI errors

marwan38:
Nice! Thank you

Issue #15, opened 2020-10-29
Google App Engine deployment example


NickFoden:
[UPDATE] : see the more recent comments below for how simple this is today and all routing working great etc out of box

I deployed to google app engine and routing is working great once I started using the routes configuration. Before that was going in circle for routing in production. Was trying to update handlers in `app.yaml` and looking for something like that familiar `_redirects` step for netlify deployments etc. 

Anyhow i have a basic example linked below showing the necessary code changes for google app engine deployment. I can list out more steps re: app engine setup if that helps for an example in the docs. It's pretty easy (compared to other cloud providers), but still some stuff missing that could be listed in a clear 1,2,3 steps format for the docs. 

https://github.com/NickFoden/starter-express-app-engine/pull/1

MUCHAS GRACIAS! 

Between Remix launching and testing out subscriptions/streaming data in faunaDB this week is 11/10.

chaance:
Hey @NickFoden I know it's been a while, but I'd love to get this in our examples directory. Have you tried running with more recent versions by chance?

NickFoden:
@chaance I have not, but have been meaning to update a project. Will circle back here in near future after doing a deployment and see what is required. I think it's going to be much simpler today.

NickFoden:
@chaance as expected this is way simpler today and confirmed works great when selecting the `Remix App Server` I may try again with express. But at moment this is very cool news. 

https://github.com/NickFoden/remix-app-server-gae/pull/1
- see the `Notes.md` file, I will clean up with proper bullets and formatting with the `gcloud` cli commands and push a commit after do some things today. 

### Deployed here:
- https://remix-app-server.ue.r.appspot.com/
- routing working well and hitting a 3rd party api in the loader etc

MichaelDeBoey:
@NickFoden It would be awesome to have a PR that adds an example to our repo.

NickFoden:
> @NickFoden It would be awesome to have a PR that adds an example to our repo.

Sure thing. Will add this week. Have few urgent items first.

chaance:
Closing the issue but feel free to add a comment here when/if you're able to get the PR in!

PR #3995, opened 2022-08-14
docs(example): add an example of Google App Engine deployment

himorishige:
Add an example of Google App Engine deployment(standard environment).

Closes: #15

- [x] Docs
- [ ] Tests


machour:
@himorishige it seems that we totally missed this PR when migrating to the new [examples][https://github.com/remix-run/examples] repository 😓 

Would you mind reopening it there? The [base template](https://github.com/remix-run/examples/tree/main/__template) changed a bit as well, make sure to update your code to reflect that.

Thank you! 🙏🏼

Issue #3998, opened 2022-08-14
Action adds `index` query param indefinitely  -> ?index&index&index&index


ZiiMakc:
### What version of Remix are you using?

1.6.8

### Steps to Reproduce

A simple `<Form method='post'>` on index route adds  `index` query param for each action taken resulting in `?index&index&index&index...` after [3697](https://github.com/remix-run/remix/pull/3697) was merged.

This can be mitigated by adding `action='.'`, but then query params are not saved.

### Expected Behavior

No `index` param added or added only once.

### Actual Behavior

`index` param added for each action call.

p-m-p:
I'm seeing the same behaviour with `useSubmit` in latest version update `1.6.8` https://stackblitz.com/edit/node-v3byq1?file=app/routes/index.tsx

brophdawg11:
Should be fixed by #4000, which should go out in 1.6.9 soon 👍

PR #4000, opened 2022-08-15
fix: avoid duplicating index param on submissions

brophdawg11:
Closes: #3998


packages/remix-react/components.tsx
```diff
@@ -1022,12 +1022,22 @@ export function useFormAction(
...
+      search = params.toString() ? `?${params.toString()}` : "";
+    } else {
+      search = location.search;
+    }
     hash = location.hash;
   }
 
-  let isIndexRoute = id.endsWith("/index");
   if ((action == null || action === ".") && isIndexRoute) {
     search = search ? search.replace(/^\?/, "?index&") : "?index";
```
brophdawg11, line 1042: This logic relies on search _not_ having the index param in it, which is how `useResolvedPath` returns it.  Now that we pull directly from `useLocation`, it _does_ have it so this logic appends a duplicate one.  I chose to remove it from the new code path above and leave this untouched, but we could also likely just use separate logic for the `action == null` and `action === "."` cases here if that would be clearer?

chaance (approved):

PR #4001, opened 2022-08-15
fix(rollup): use `copySync` instead of async `copy`

n8agrin:
`yarn playground:new` was generating random errors related to unlinking/deleting files that did not exist. File copying was using an asynchronous method to move files into playground directories.

This change makes the copy/move operation synchronous and eliminates race conditions, ensuring `yarn playground:new` always succeeds.


rollup.utils.js
```diff
@@ -120,7 +120,7 @@ function copyToPlaygrounds() {
             path.join(REPO_ROOT_DIR, "build"),
             playgroundDir
           );
-          await fse.copy(writtenDir, destDir);
+          await fse.copySync(writtenDir, destDir);
```
MichaelDeBoey, line 123: No need for `await` anymore
```suggestion
          fse.copySync(writtenDir, destDir);
```

MichaelDeBoey (changes requested):
I don't think this changes anything tbh, as we're awaiting for the sync to be finished before we continue (just like we would use the synchronous function) 🤔

rollup.utils.js
n8agrin: Good call.

n8agrin:
@MichaelDeBoey regarding this not changing anything, a trivial repro is to checkout a new clone of remix and run `yarn playground:new`. On my machine that frequently produces errors like the following:

```
packages/remix-dev/server-build.ts → build/node_modules/@remix-run/dev, build/node_modules/@remix-run/dev/dist...
[!] (plugin copy-to-remix-playground) Error: ENOENT: no such file or directory, unlink '/Users/nate/Projects/remix-two/playground/playground-1660456168247/node_modules/@remix-run/dev/dist/cli/migrate/migrations/replace-remix-imports/transform/index.js'
Error: ENOENT: no such file or directory, unlink '/Users/nate/Projects/remix-two/playground/playground-1660456168247/node_modules/@remix-run/dev/dist/cli/migrate/migrations/replace-remix-imports/transform/index.js'

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
node:internal/errors:841
  const err = new Error(message);
              ^

Error: Command failed: yarn rollup -c
    at checkExecSyncError (node:child_process:828:11)
    at execSync (node:child_process:899:15)
    at createNewProject (/Users/nate/Projects/remix-two/scripts/playground/new.js:46:3) {
  status: 1,
  signal: null,
  output: [ null, null, null ],
  pid: 9794,
  stdout: null,
  stderr: null
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I agree that in theory the `await fse.copy` gesture should prevent this, and I didn't dig into the internals of `fse`, but changing `copy` to `copySync` completely eliminates the issue for me.

jacob-ebey:
@MichaelDeBoey I said to put this PR up. It's something I've ran into locally quite a bit and I address it with the above solution. I'm honestly not sure why async ops are breaking here, but they are.

MichaelDeBoey (approved):

kentcdodds (approved):
Seems fine 👍

kentcdodds:
Just sitting like this forever. Even after a refresh 🤦‍♂️

PR #4003, opened 2022-08-15
Use fetched item count for simple examples

rkulinski:
In order to prevent bugs related to scrolling with scroll bar use fetched items count instead of all available.


kentcdodds (approved):
Yep. This makes sense. For the best UX, people should be using the advanced example anyway 👍

Thanks a bunch!

PR #4013, opened 2022-08-17
feat(@remix-run/cloudflare,@remix-run/deno,@remix-run/node): `SerializeFrom` utility for loader and action type inference

pcattori:
Renamed internal `UseDataFunctionReturn` type to `SerializeFrom` since its can be used without `useLoaderData` and `useActionData`. A bunch of small type cleanups + bug fixes. Moved serialize type utilities to `@remix-run/server-runtime` and re-exporting `SerializeFrom` and `TypedResponse` from each runtime pkg.

---

- [ ] Docs
- [x] Tests
- [x] Changeset

Testing Strategy:
- updated existing hooks types tests in `@remix-run/react`
- locally created an app with indie stack, installed local `@remix-run/server-runtime` and `@remix-run/node`, imported `SerializedFrom` from `@remix-run/node`, and got the correct type in VS Code


packages/remix-react/__tests__/hook-types-test.tsx
```diff
@@ -1,78 +1,83 @@
...
+import type { TypedResponse } from "../serialize";
+import type { useLoaderData } from "../components";
 
 function isEqual<A, B>(
   arg: A extends B ? (B extends A ? true : false) : false
 ): void {}
 
+// not sure why `eslint` thinks the `T` generic is not used...
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+type LoaderData<T> = ReturnType<typeof useLoaderData<T>>;
```
pcattori, line 10: note: this test is for testing `useLoaderData`'s types, so we actually grab types from `useLoaderData` instead of assuming its using `UseDataFunctionReturn` or `SerializeFrom`.

packages/remix-react/index.tsx
```diff
@@ -58,6 +58,8 @@ export type { ShouldReloadFunction, HtmlMetaDescriptor } from "./routeModules";
 
 export { ScrollRestoration } from "./scroll-restoration";
 
+export type { SerializeFrom } from "./serialize";
```
pcattori, line 61: Here's where we make `SerializeFrom` into a public API importable from `@remix-run/react`.

packages/remix-cloudflare/index.ts
```diff
@@ -32,7 +32,6 @@ export type {
   CookieParseOptions,
   CookieSerializeOptions,
   CookieSignatureOptions,
-  CreateRequestHandlerFunction,
```
pcattori, line 35: `CreateRequestHandlerFunction` is a utility for authoring Remix runtime packages and was incorrectly re-exported before.

packages/remix-deno/index.ts
```diff
@@ -35,7 +35,6 @@ export type {
   CookieParseOptions,
   CookieSerializeOptions,
   CookieSignatureOptions,
-  CreateRequestHandlerFunction,
```
pcattori, line 38: https://github.com/remix-run/remix/pull/4013/files#r949560392

packages/remix-node/index.ts
```diff
@@ -60,7 +59,6 @@ export type {
   CookieParseOptions,
   CookieSerializeOptions,
   CookieSignatureOptions,
-  CreateRequestHandlerFunction,
```
pcattori, line 63: https://github.com/remix-run/remix/pull/4013/files#r949560392

packages/remix-server-runtime/index.ts
```diff
@@ -28,7 +28,6 @@ export type {
   IsSessionFunction,
   JsonFunction,
   RedirectFunction,
-  TypedResponse,
```
pcattori, line 31: The `interface` section of exports is for types related to values/functions that should be exported by every server runtime pkg. We use the fetch API for request/response, so we don't export our own request/response implementations from server runtime pkgs.

Therefore, `TypedResponse` should not be considered part of the server runtime interface.

---

TBH the distinction between "interface" and "re-export" is subtle and I'd much prefer to not need re-exports at all once we have a pkg for exporting runtime-agnostic stuff.

.changeset/neat-beds-unite.md
```diff
@@ -0,0 +1,22 @@
...
+"remix": minor
+"@remix-run/cloudflare": minor
+"@remix-run/deno": minor
+"@remix-run/node": minor
+"@remix-run/react": minor
+"@remix-run/serve": minor
+"@remix-run/server-runtime": minor
+---
+
+feat(@remix-run/cloudflare,@remix-run/deno,@remix-run/node): `SerializeFrom` utility for loader and action type inference
```
chaance, line 11: We don't need the conventional commit format here. Affected packages and change type are already included in the changesets.

chaance (changes requested):
Just one note on the changeset, otherwise looks good!

PR #4017, opened 2022-08-18
feat(templates/express): support `.env` durning local development

lili21:
Closes: #

- [x] Docs
- [ ] Tests

Testing Strategy:


contributors.yml
```diff
@@ -418,3 +418,4 @@
 - zachdtaylor
 - zainfathoni
 - zhe
+- lili.21
```
machour, line 421: This much match your GitHub username exactly, or it won't work.

templates/express/server.js
```diff
@@ -1,3 +1,7 @@
+if (process.env.NODE_ENV === "development") {
+  require("dotenv").config();
```
machour, line 2: `dotenv` is not a dependency in the package.json file, so it would need to be added. 

This change also looks like a specific use case for you, I'm not sure this will be accepted.

machour (changes requested):
CLA needs to be signed correctly before this PR is reviewed by the team.

contributors.yml
MichaelDeBoey:
```suggestion
- lili21
```

templates/express/server.js
MichaelDeBoey: We also already have support for `dotenv` in local development in `@remix-run/dev`

MichaelDeBoey (commented):
I don't think we want to merge this, as this seems very specific and people can do this in their own project if they want to.

Will let the team (@kentcdodds & @mcansh) make the final decision.

templates/express/server.js
mcansh: you can get this same behavior using `node --require dotenv/config server.js` without having to have it as a runtime check, which is what we have set up for the stacks (indie and blues anyway) https://github.com/remix-run/blues-stack/blob/main/package.json#L11

contributors.yml
MichaelDeBoey:
```suggestion
- lili21
```

templates/express/server.js
MichaelDeBoey: We also already have support for `dotenv` in local development in `@remix-run/dev`

lili21:
the express template is created by the remix team. I guess the developer would expect that the DX is the same as the default template. 
things like local environments and live reload should be supported out of the box, right?

templates/express/package.json
```diff
@@ -4,7 +4,7 @@
   "scripts": {
     "build": "remix build",
     "dev": "remix build && run-p \"dev:*\"",
-    "dev:node": "cross-env NODE_ENV=development nodemon ./server.js --watch ./server.js",
+    "dev:node": "cross-env NODE_ENV=development nodemon -r dotenv/config ./server.js --watch ./server.js",
```
MichaelDeBoey, line 7:
```suggestion
    "dev:node": "cross-env NODE_ENV=development nodemon --require dotenv/config ./server.js --watch ./server.js",
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

templates/express/package.json
```diff
@@ -4,7 +4,7 @@
   "scripts": {
     "build": "remix build",
     "dev": "remix build && run-p \"dev:*\"",
-    "dev:node": "cross-env NODE_ENV=development nodemon ./server.js --watch ./server.js",
+    "dev:node": "cross-env NODE_ENV=development nodemon --require dotenv/config ./server.js --watch ./server.js",
```
machour, line 7: If we depend on `dotenv`, it should be explicitly listed in our `devDependencies`

machour (dismissed):

MichaelDeBoey (approved):

machour (commented):
I'm not fan of this change, but I'll leave it to the team to decide since it's not my place.

My 2cts: The express adapter is meant as "do what you want if you feel Remix App Server doesn't meet your needs", and it should be as bare-bone as possible.

mcansh:
The express adapter is meant as "do what you want if you feel Remix App Server doesn't meet your needs", and it should be as bare-bone as possible.

I agree 100% - templates are made to be as basic as possible, however i do think this should be an "example" which you could then create with `npx create-remix@latest --template examples/<name>`

kiliman:
Yes, if you need `dotenv`, it'll be easy to update the _server.js_ file manually after setup.

Also, if/when PR #4123 is merged, you'll get the DX of Remix App Server automatically. With the customization of Express.

SokratisVidros:
I came across the same issue today and after some digging I patched `server.js` to load the `.env` file. However, it is expected to have the same experience across all official remix stacks. +1 on merging this.

mcansh:
the more i think about this, the more i think we should include it. does it work fine if there is no `.env` as well?

SokratisVidros:
> the more i think about this, the more i think we should include it. does it work fine if there is no `.env` as well?

Yes

mcansh (approved):

machour (approved):

Issue #4023, opened 2022-08-19
Duplicated button's name/value pair in FormData on Safari (ONLY)


raskyer:
### What version of Remix are you using?

1.6.8

### Steps to Reproduce

- Create a remix project
- Have a Form component looking like =>
```
<Form method="post" replace>
  <input type="hidden" name="id" value="1" />

  <button type="submit" name="_action" value="delete">
    x
  </button>
</Form>
```
- Open the url with the form in Safari
- Click the button
- FormData sent to action looks like : 
id => 1
_action => delete
_action => delete

This does not occur on : Chrome / Firefox.

### Expected Behavior

Form Data only contains the button name / value pair one time.

### Actual Behavior

Form Data contains the button name / value pair two time (duplicate).

raskyer:
Probably caused by : https://github.com/remix-run/remix/blob/remix%401.6.8/packages/remix-react/components.tsx#L1162

Maybe on Safari the button key value pair is already in the form submission data ? Adding a little if to check if forms already contains the key / value pair might be a solid workaround ?

raskyer:
Pull request : https://github.com/remix-run/remix/commit/5894598c75d0e542651dc92bc400320c108eae8d

mcansh:
> Maybe on Safari the button key value pair is already in the form submission data?

yeah exactly this, safari already includes the "submitter" in the FormData, in other browsers, we check the `event.target.submitter` and add it manually

mcansh:
however, i just pulled this up locally and i'm only getting a single "_action" value sent in Safari 16.

![image](https://user-images.githubusercontent.com/11698668/185960446-f22d0bb6-6f0b-42ea-a383-7b4aabb89222.jpeg)

raskyer:
Have you test on your mobile with recent remix version : https://github.com/remix-run/remix/pull/4024#issuecomment-1222504413

mcansh:
sorry, spoke too soon, the request data only shows one, but logging `formData.getAll('_action')` returns two

mcansh:
also, didn't realize you opened a PR 😅, thanks! I'll check that out

brophdawg11:
This looks like it's a bug in Safari that's been fixed in Safari 16: https://bugs.webkit.org/show_bug.cgi?id=239070

PR #4024, opened 2022-08-19
fix(remix-react): prevent form data duplicated value

raskyer:
After : https://github.com/remix-run/remix/commit/5894598c75d0e542651dc92bc400320c108eae8d
A bug was introduced on Safari. Duplicated value of the form button was appended to the form data.

In Chrome, Firefox (and maybe other's) `new FormData(form)`, where form is a form dom element, will parse the form's input + value into the form data without getting the button element. On Safari, on the other hand, the button's name + value pair are added. With https://github.com/remix-run/remix/commit/5894598c75d0e542651dc92bc400320c108eae8d we don't `set` the button's value anymore but we `append` it. It solves @nrako use case where inputs and button have the same name and that you wan't to keep those. On the other hand, on Safari, it cause a duplicated entry of your button's name + value.

In an attempt to fix this behaviour without affecting @nrako effort, we introduce a check before appending to see if the button's name + value pair is already in the form data.

It's not a perfect solution because some use case would probably like to add also the button value even if it's already in the form data. But since it's a less common use case than just using "Safari" it might be better.

Closes: #4023 

- [ ] Docs
- [x] Tests

Testing Strategy:

copy / past the tests already made by @nrako (thank you) but just checked if the value was already submitted.

### DIFF
--- a/integration/hook-useSubmit-test.ts
+++ b/integration/hook-useSubmit-test.ts
@@ -40,6 +40,36 @@ test.describe("`useSubmit()` returned function", () => {
           )
         }
       `,
+        
+        "app/routes/duplicate.jsx": js`
+        import { useLoaderData, useSubmit } from "@remix-run/react";
+
+        export function loader({ request }) {
+          let url = new URL(request.url);
+          return url.searchParams.toString()
+        }
+
+        export default function Index() {
+          let submit = useSubmit();
+          let handleClick = event => {
+            event.preventDefault()
+            submit(event.nativeEvent.submitter || event.currentTarget)
+          }
+          let data = useLoaderData();
+          return (
+            <form>
+              <input type="text" name="tasks" defaultValue="first" />
+              <input type="text" name="tasks" defaultValue="second" />
+
+              <button onClick={handleClick} name="tasks" value="second">
+                Prepare Third Task
+              </button>
+
+              <pre>{data}</pre>
+            </form>
+          )
+        }
+      `,
       },
     });
 
@@ -59,6 +89,19 @@ test.describe("`useSubmit()` returned function", () => {
     await page.waitForLoadState("load");
     expect(await app.getHtml("pre")).toBe(
       `<pre>tasks=first&amp;tasks=second&amp;tasks=third</pre>`
+    );
+  });
+
+  // IN ORDER TO AVOID DUPLICATED VALUE ON SAFARI
+  test("doesn't submits the submitter's value duplicated to the form data", async ({
+    page,
+  }) => {
+    let app = new PlaywrightFixture(appFixture, page);
+    await app.goto("/");
+    await app.clickElement("text=Prepare Third Task");
+    await page.waitForLoadState("load");
+    expect(await app.getHtml("pre")).toBe(
+      `<pre>tasks=first&amp;tasks=second</pre>`
     );
   });
 });
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1161,7 +1161,9 @@ export function useSubmitImpl(key?: string): SubmitFunction {
 
         // Include name + value from a <button>
         if (target.name) {
-          formData.append(target.name, target.value);
+          if (!formData.getAll(target.name).includes(target.value)) {
+            formData.append(target.name, target.value);
+          }
         }
       } else {
         if (isHtmlElement(target)) {
--- a/contributors.yml
+++ b/contributors.yml
@@ -418,3 +418,4 @@
 - zachdtaylor
 - zainfathoni
 - zhe
+- raskyer
--- a/integration/form-test.ts
+++ b/integration/form-test.ts
@@ -320,6 +320,29 @@ test.describe("Forms", () => {
             )
           }
         `,
+
+        "app/routes/submitter-duplicate.jsx": js`
+          import { useLoaderData, Form } from "@remix-run/react";
+
+          export function loader({ request }) {
+            let url = new URL(request.url);
+            return url.searchParams.toString()
+          }
+
+          export default function() {
+            let data = useLoaderData();
+            return (
+              <Form>
+                <input type="text" name="tasks" defaultValue="first" />
+                <input type="text" name="tasks" defaultValue="second" />
+                <button type="submit" name="tasks" value="second">
+                  Add Task
+                </button>
+                <pre>{data}</pre>
+              </Form>
+            )
+          }
+        `
       },
     });
 
@@ -768,6 +791,19 @@ test.describe("Forms", () => {
     await page.waitForLoadState("load");
     expect(await app.getHtml("pre")).toBe(
       `<pre>tasks=first&amp;tasks=second&amp;tasks=</pre>`
+    );
+  });
+
+  // IN ORDER TO AVOID DUPLICATED VALUE ON SAFARI
+  test("<Form> doesn't submits the submitter's duplicated value to the form data", async ({
+    page,
+  }) => {
+    let app = new PlaywrightFixture(appFixture, page);
+    await app.goto("/submitter-duplicate");
+    await app.clickElement("text=Add Task");
+    await page.waitForLoadState("load");
+    expect(await app.getHtml("pre")).toBe(
+      `<pre>tasks=first&amp;tasks=second</pre>`
     );
   });
 });


packages/remix-react/components.tsx
```diff
@@ -1161,7 +1161,9 @@ export function useSubmitImpl(key?: string): SubmitFunction {
 
         // Include name + value from a <button>
         if (target.name) {
-          formData.append(target.name, target.value);
+          if (!formData.getAll(target.name).includes(target.value)) {
```
herecydev, line 1164:
```suggestion
          if (!formData.has(target.name)) {
```

That should work right?

raskyer: Nop, because it would reset https://github.com/remix-run/remix/commit/5894598c75d0e542651dc92bc400320c108eae8d effort.

```
<Form ...>
  <input type="hidden" name="somethingIWantToKeep" value="1" />
  <button type="submit" name="somethingIWantToKeep" value="2" />
</Form>
```

I don't know if there's a lot of use case for that, but @nrako has fixed it for a reason.

raskyer: Like `!formData.has` + `formData.append` is almost equal to `formData.set`, in the sens that, if your form already have an entry with your form button's name then, (in Chrome + Firefox and maybe others) your button's name + value will not be added. In Safari it's different because the button with it's name + value is already added in the form's data.

So the easiest work around is to check if there's already a similar entry in your form data. Since we can't check from what input comes the entry in the form data class.

It's not perfect, and maybe we don't want to support this use case (having a button with the exact same name of inputs) but in this case we should revert the pr from @nrako.

herecydev: Right gotcha, honestly the use case feels weird for duping name attribute, would be good to better understand that.

raskyer: There's a legitimate use case of having the button name attribute in the form data. Such as having multiple actions possibles in the remix's action.

```
<Form ...>
  <input type="hidden" name="id" value="1" />
  <button type="submit" name="intent" value="delete">Delete Todo</button>
</Form>

<Form ...>
  <input type="text" name="todo" />
  <button type="submit" name="intent" value="add">Add Todo</button>
</Form>
```

For the @nrako use case, I think we should refer to its commit but I can understand that erasing former inputs at the `set` stage could be a little bit annoying, even if I arguably don't see any benefits over hidden input or just changing the name attribute.

So we have a clear choice, either : 
- Revert  https://github.com/remix-run/remix/commit/5894598c75d0e542651dc92bc400320c108eae8d commit in order to avoid duplicated form data entry
- Merge this fix in order to fix duplicated entry on Safari

herecydev: In that case you would only get either set of formData right? Never both at the same time?

brophdawg11:
> On Safari, on the other hand, the button's name + value pair are added

@raskyer What version of Safari are you using?  I'm unable to reproduce this behavior on 15.3 using the following HTML.  So I also can't reproduce the duplicative behavior you're referencing - could you possibly provide a reproduction?

```
<form>
  <input name="x" value="1" />
  <button type="submit" name="y" value="2">Submit</button>
</form>
```

And in the console:

Screen Shot 2022-08-22 at 10 41 21 AM

I'm not sure the tests in the PR are the behavior we want.  If you have an input name/value and a button with the same name/value in raw HTML - those would be duplicated by the browser on form submission by default, so I think Remix is doing the correct thing here.

I confirmed this by rendering a simple HTML form without any JS and submitting and checking the url in both Chrome and Safari: 

```html
<form>
  <input name="x" value="1" />
  <input name="y" value="2" />
  <button type="submit" name="y" value="2">Submit</button>
</form>
```

Chrome: 
Screen Shot 2022-08-22 at 10 51 32 AM

Safari:
Screen Shot 2022-08-22 at 10 51 59 AM

raskyer:
https://la-plateforme-web.azurewebsites.net/test

Version  safari : Last on iPhone 

Code: 
```
import type { ActionArgs } from '@remix-run/node';
import { Form, useActionData } from '@remix-run/react';

export async function action({ request }: ActionArgs) {
  const formData = await request.formData();

  return {
    id: formData.getAll('id'),
    intent: formData.getAll('intent'),
  };
}

export default function Test() {
  const data = useActionData<typeof action>();

  return (
    <Form method="post" replace>
      <input type="hidden" name="id" value="1" />

      <button type="submit" name="intent" value="add">
        Add
      </button>

      {data !== undefined ? (
        <pre>
          id : {data.id.join(', ')}
          <br />
          intent : {data.intent.join(', ')}
        </pre>
      ) : null}
    </Form>
  );
}
```

raskyer:
5E273AF2-FD3E-4A7E-921C-B2E44CCF19CC

raskyer:
My iOS version : 15.6.1

integration/hook-useSubmit-test.ts
```diff
@@ -40,6 +40,36 @@ test.describe("`useSubmit()` returned function", () => {
...
+            event.preventDefault()
+            submit(event.nativeEvent.submitter || event.currentTarget)
+          }
+          let data = useLoaderData();
+          return (
+            <form>
+              <input type="text" name="tasks" defaultValue="first" />
+              <input type="text" name="tasks" defaultValue="second" />
+
+              <button onClick={handleClick} name="tasks" value="second">
```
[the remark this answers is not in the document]
mcansh: similarly to this test https://github.com/remix-run/remix/blob/a06f7d46fccb07cc192e3a11395c0d3ed503436d/integration/form-test.ts#L317-L320 - but that test does have the bug in safari as we end up with two "tasks=" params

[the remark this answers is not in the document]
raskyer: It is expected but as mention in the description in my pr, i believe it’s the easiest fix. Idk if we do specific browser based fix on remix ? But if it’s the case I’m curious to see where and how

packages/remix-react/components.tsx
raskyer: I’ m not sure to understand. 
On the example I gave, it was just to demonstrate the need of having the button name / value in form data.

in this example, on safari, the intent has two intent / delete or intent / add (depending on where you click.

For the commit of @nrako I let you refer to its commit to understand his need.

brophdawg11:
Huh, this is indeed weird, and I think could argue is a bug in Safari.  At first I thought maybe it was new as of 15.4 when safari added the [`event.submitter`](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter#browser_compatibility) functionality.  But I confirmed the behavior below happens in both mobile safari 15.6.1, and desktop safari 15.3.

***tl;dr;**
Safari includes submit button name/value in `new FormData(formEl)` _if you execute it during a form submit event handler event loop_.  It seems to implicitly track on the form the button that kicked off the submit event.

**details**
Per my reading of the [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-the-form-data-set), all buttons should be skipped unless they are the "submitter":

> For each element field in controls, in tree order:
> If any of the following is true:
> ...
> The field element is a button but it is not _submitter_
> ...
> Then continue

So, to me, when we do new `FormData(formElement)` we have not specified a "submitter" in any way so buttons should be excluded.

However, if I load the example page above and run this in the console, I get no values - as I would expect.

```js
new FormData(document.querySelector('form')).get('intent')`
```

However, if I put a breakpoint inside the code that creates the `FormData` and run the same command _during the submit event handling_ - it _does_ include the button value.  So it seems Safari is being stateful in some way such that if you create `FormData` _during the submit event event loop_, it knows _what_ button submitted and will include that.  So if they're implicitly tracking "submitter" you could argue it complies with the spec.  But I wouldn't necessarily expect that implementation to leak out to external/separate `new FormData(formEl)` calls in code on the submit path 😕 

Here's a very simple example reproducing the behavior.  We only get the button value in our `new FormData` call in the submit event handler, not in the `click` handler and not in the `setTimeout` from the `submit` handler.

```jsx
<form>
  <input type="hidden" name="id" value="1">
  <button type="submit" name="intent" value="add">Add</button>
</form>
<script>
  document.querySelector('form').addEventListener('submit', () => {
    event.preventDefault();

    // WILL include the button name/value
    let fd = new FormData(document.querySelector('form'));
    console.log('submit', fd.getAll('intent'));

    setTimeout(() => {
      // WILL NOT include the button name/value
      let fd = new FormData(document.querySelector('form'));
      console.log('submit async', fd.getAll('intent'));
    }, 0)
  });

  document.querySelector('button').addEventListener('click', () =>  {
    // WILL NOT include the button name/value
    let fd = new FormData(document.querySelector('form'));
    console.log('click', fd.getAll('intent'));
  });
</script>
```

Code: https://codesandbox.io/s/safari-dup-submit-button-value-xmxl6l
Live app (open console to see output): https://xmxl6l.sse.codesandbox.io/

raskyer:
> Huh, this is indeed weird, and I think could argue is a bug in Safari. At first I thought maybe it was new as of 15.4 when safari added the [`event.submitter`](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter#browser_compatibility) functionality. But I confirmed the behavior below happens in both mobile safari 15.6.1, and desktop safari 15.3.
> 
> ***tl;dr;** Safari includes submit button name/value in `new FormData(formEl)` _if you execute it during a form submit event handler event loop_. It seems to implicitly track on the form the button that kicked off the submit event.
> 
> **details** Per my reading of the [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-the-form-data-set), all buttons should be skipped unless they are the "submitter":
> 
> > For each element field in controls, in tree order:
> > If any of the following is true:
> > ...
> > The field element is a button but it is not _submitter_
> > ...
> > Then continue
> 
> So, to me, when we do new `FormData(formElement)` we have not specified a "submitter" in any way so buttons should be excluded.
> 
> However, if I load the example page above and run this in the console, I get no values - as I would expect.
> 
> ```js
> new FormData(document.querySelector('form')).get('intent')`
> ```
> 
> However, if I put a breakpoint inside the code that creates the `FormData` and run the same command _during the submit event handling_ - it _does_ include the button value. So it seems Safari is being stateful in some way such that if you create `FormData` _during the submit event event loop_, it knows _what_ button submitted and will include that. So if they're implicitly tracking "submitter" you could argue it complies with the spec. But I wouldn't necessarily expect that implementation to leak out to external/separate `new FormData(formEl)` calls in code on the submit path 😕
> 
> Here's a very simple example reproducing the behavior. We only get the button value in our `new FormData` call in the submit event handler, not in the `click` handler and not in the `setTimeout` from the `submit` handler.
> 
> ```js
> <form>
>   <input type="hidden" name="id" value="1">
>   <button type="submit" name="intent" value="add">Add</button>
> </form>
> <script>
>   document.querySelector('form').addEventListener('submit', () => {
>     event.preventDefault();
> 
>     // WILL include the button name/value
>     let fd = new FormData(document.querySelector('form'));
>     console.log('submit', fd.getAll('intent'));
> 
>     setTimeout(() => {
>       // WILL NOT include the button name/value
>       let fd = new FormData(document.querySelector('form'));
>       console.log('submit async', fd.getAll('intent'));
>     }, 0)
>   });
> 
>   document.querySelector('button').addEventListener('click', () =>  {
>     // WILL NOT include the button name/value
>     let fd = new FormData(document.querySelector('form'));
>     console.log('click', fd.getAll('intent'));
>   });
> </script>
> ```
> 
> Code: https://codesandbox.io/s/safari-dup-submit-button-value-xmxl6l Live app (open console to see output): https://xmxl6l.sse.codesandbox.io/

So you believe we should not try to fix it in remix ? 
It could make sens since it’s only Safari that is impacted but I am afraid to not see a fix anytime soon from Apple and this issue could break production app (especially when you check your form data against Zod or yup schemas

brophdawg11:
I'm leaning towards not fixing since it seems like Safari is not following the spec here, but I'd like to get @ryanflorence/@mjackson's thoughts too.

The fix proposed seems potentially worse, since it fixes this at the expense of introducing a different bug that impacts all browsers, not just Safari.  The following should include duplicate values, but this PR would stop doing that:

```html
<form>
  <input name="x" value="1">
  <button type="submit" name="x" value="1">
</form>
```

raskyer:
> I'm leaning towards not fixing since it seems like Safari is not following the spec here, but I'd like to get @ryanflorence / @mjackson's thoughts too.
> 
> The fix proposed seems potentially worse, since it fixes this at the expense of introducing a different bug that impacts all browsers, not just Safari. The following should include duplicate values, but this PR would stop doing that:
> 
> ```
> <form>
>   <input name="x" value="1">
>   <button type="submit" name="x" value="1">
> </form>
> ```

It was the case before remix 1.6.8 though. So not really "worst", just came back to a pre 1.6.8 situation.

mcansh:
> Here's a very simple example reproducing the behavior. We only get the button value in our `new FormData` call in the submit event handler, not in the `click` handler and not in the `setTimeout` from the `submit` handler.

i get an empty array for all 3 of these in Safari 16, using `new URLSearchParams(new FormData(document.forms[0])).getAll('id')` i get ['1'] for all

brophdawg11:
Thanks to @mcansh for finding this bug report in Webkit!  https://bugs.webkit.org/show_bug.cgi?id=239070

This looks like it's fixed in Safari 16, so I think we're doing the right thing currently

PR #4029, opened 2022-08-20
chore(templates): fix ESLint errors

MichaelDeBoey:
Both `format` & `lint` workflows are failing since ba5c8486

https://github.com/remix-run/remix/runs/7908428299?check_suite_focus=true
https://github.com/remix-run/remix/runs/7908428662?check_suite_focus=true

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,7 @@
+const OFF = 0;
+const WARN = 1;
+
+/** @type {import('eslint').Linter.Config} */
 module.exports = {
   root: true,
   extends: [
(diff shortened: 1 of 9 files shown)


kentcdodds (changes requested):
These are templates. We do not want to generate projects for people with `let`

MichaelDeBoey:
@kentcdodds We've always done this since the refactor of the internal ESLint config.
If you want to, I can fix it so we have `const` + the warnings/errors are fixed

kentcdodds:
Yep. That's sounds good. Thank you.

PR #4036, opened 2022-08-22
feat(remix-react): Add timeout prop for <LiveReload /> component

joelazar:
`LiveReload` component's reconnect attempt was hard coded to 1 sec. However, this breaks the feature if the page which attempts to be loaded has a loader that takes more than 1 sec to load (which communicates with :snail: 3rd-party APIs :cry:). In that case, the `LiveReload` component gets into an infinity loop.

This change allows the consumer to configure the timeout manually.

Also, you could even argue that the default timeout limit should be increased too.

Closes: #

- [ ] Docs
- [x] Tests

### DIFF
--- a/packages/remix-react/__tests__/components-test.tsx
+++ b/packages/remix-react/__tests__/components-test.tsx
@@ -65,6 +65,14 @@ describe("<LiveReload />", () => {
         `1234 + "/socket"`
       );
     });
+
+    it("timeout of reload is set to 200ms", () => {
+      LiveReload = require("../components").LiveReload;
+      let { container } = render(<LiveReload timeout={200} />);
+      expect(container.querySelector("script")).toHaveTextContent(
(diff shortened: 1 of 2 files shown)


machour:
Hi @joelazar, thank you for your submission (and sorry for the delay!)

Can you target `dev` instead of `main` as per our [contributing guide](https://github.com/remix-run/remix/blob/main/docs/pages/contributing.md#making-a-pull-request) requirement?

joelazar:
> Hi @joelazar, thank you for your submission (and sorry for the delay!)
> 
> Can you target `dev` instead of `main` as per our [contributing guide](https://github.com/remix-run/remix/blob/main/docs/pages/contributing.md#making-a-pull-request) requirement?

Ah, sorry  :facepalm: , I've changed the target to `dev` now.

machour (approved):

pcattori (approved):

pcattori:
Thanks @joelazar for the good work! 🙏

PR #4039, opened 2022-08-22
fix(remix-react): don't use nullish coalescing for fallback path

mcansh:
storybook doesn't like it. https://canary.discord.com/channels/770287896669978684/1010847352426352682/1010847352426352682

actually, it appears you can customize the babel config that storybook uses https://storybook.js.org/docs/react/configure/babel

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1013,7 +1013,7 @@ export function useFormAction(
   method: FormMethod = "get"
 ): string {
   let { id } = useRemixRouteContext();
-  let resolvedPath = useResolvedPath(action ?? ".");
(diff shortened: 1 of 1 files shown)


MichaelDeBoey (commented):
Since we can configure Babel for Storybook, I think we should go that way as nullish coalescing is perfectly fine for Node 14

mcansh:
yeah same, we've been having a bit of an internal discussion about browser support via browserslist, i think we're going to use this or something similar, but haven't super into it yet https://github.com/marcofugaro/browserslist-to-esbuild

Issue #4040, opened 2022-08-22
[bug]: architect binary types typo


kamtugeza:
### What version of Remix are you using?

1.6.8

### Steps to Reproduce

Open the [remix-architect/binaryTypes.ts](https://github.com/remix-run/remix/tree/main/packages/remix-architect) and take a look at line 36.

### Expected Behavior

There should be two types: `audio/wav` and `audio/webm`.

### Actual Behavior

There are concatenated types "audio/wavaudio/webm"

MichaelDeBoey:
Fixed by #4041

PR #4041, opened 2022-08-22
fix: separate two binary types which were concatenated

kamtugeza:
Closes: #4040

That's a tiny typo in the [remix-architect/binaryTypes.ts](https://github.com/remix-run/remix/tree/main/packages/remix-architect) at line 36. I suppose there should be two types `audio/wav` and `audio/webm`.


kamtugeza:
I'm not sure should that goes to the discussion section or that could be done here, but I've compared types from that file with types from [mime-types](https://www.npmjs.com/package/mime-types) (thanks to @wingleung for his note in [his fork of Architect Handler](https://github.com/wingleung/remix-aws/blob/main/src/binaryTypes.ts)). The library has all needed types except:
- audio/x-midi
- application/x-gzip
- application/x-java-archive
- application/x-zip

I've not found reliable info about those types. I'm not sure if they are valid, and it seems okay to check binary types by the `mime-types` library instead of the custom method. What do you think about that?

MichaelDeBoey (approved):

mcansh:
> I'm not sure should that goes to the discussion section or that could be done here, but I've compared types from that file with types from [mime-types](https://www.npmjs.com/package/mime-types) (thanks to @wingleung for his note in [his fork of Remix Architect](https://github.com/wingleung/remix-aws/blob/main/src/binaryTypes.ts)). The library has all needed types except:
> 
> - audio/x-midi
> 
> - application/x-gzip
> 
> - application/x-java-archive
> 
> - application/x-zip
> 
> 
> 
> I've not found reliable info about those types. I'm not sure if they are valid, and it seems okay to check binary types by the `mime-types` library instead of the custom method. What do you think about that?

this list started as a form of the list that architect keeps, but i'm down to looking into using mime-types if we can

packages/remix-architect/binaryTypes.ts
```diff
@@ -34,7 +34,8 @@ const binaryTypes = [
   "audio/basic",
   "audio/mpeg",
   "audio/ogg",
-  "audio/wavaudio/webm",
+  "audio/wav",
+  "audio/webm",
```
wingleung, line 38: there is a binaryTypes.ts in the netlify adapter as well, might need to change it there too 👉 https://github.com/remix-run/remix/blob/main/packages/remix-netlify/binaryTypes.ts#L36

wingleung:
> I'm not sure should that goes to the discussion section or that could be done here, but I've compared types from that file with types from [mime-types](https://www.npmjs.com/package/mime-types) (thanks to @wingleung for his note in [his fork of Remix Architect](https://github.com/wingleung/remix-aws/blob/main/src/binaryTypes.ts)). The library has all needed types except:
> 
> * audio/x-midi
> * application/x-gzip
> * application/x-java-archive
> * application/x-zip
> 
> I've not found reliable info about those types. I'm not sure if they are valid, and it seems okay to check binary types by the `mime-types` library instead of the custom method. What do you think about that?

@kamtugeza 👍  yes, I wrote a todo about it because I couldn't find a `isBinaryMimeType()` kind of helper in that library. I had a closer look now and I see the library is more of a tool to figure out mime-types instead of identifying if the mime-type is binary or not.

I see now in the AWS docs that ALB and both api gateway versions already pass a property called `isBase64Encoded` so I think, for the aws integration, we should use that property instead.

And for the other adapters like netlify, we need to check if they have a similar property to this.

- [ALB doc](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#receive-event-from-load-balancer)
- [APIGW doc](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)

kamtugeza:
Here is my silly question: do we have a situation where we should have an empty body, and a response should be "binary"? 🙃

I ask that because we check a response encoding by the `Content-Type` header and if it's `base64` then we pass `nodeResponse.body` property into the [readableStreamToString](https://github.com/remix-run/remix/blob/973cd68528c8c58679a5b2d974ae8cefde0db455/packages/remix-node/stream.ts#L46)  method where we expect body is `ReadableStream<Uint8Array>` and we don’t perform here any type checks or transformation (e.g. string to stream).

So why don’t we do a simple check, does the `nodeResponse.body` has an instance of `ReadableStream`?

PR #4049, opened 2022-08-23
test: enable webkit testing for playwright

mcansh:
to limit testing to chrome or webkit you can use 
- `yarn test:integration --project=webkit`
- `yarn test:integration --project=chromium`

currently the only failing test is due to a bug in Safari, so i've added browser check for the bugged behavior https://github.com/remix-run/remix/pull/4049/files#diff-6ba8f3349b2b29f4668d3cdbcc4df75c8182edad8eaea07c5a6d7f227309acf7R794-R803

related: #4023 / #4024 

edit: looks like `@playwright/test@next` is shipping with webkit 16 👀 

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -14,6 +14,6 @@ on:
 jobs:
   test:
     if: github.repository == 'remix-run/remix'
-    uses: remix-run/remix/.github/workflows/reusable-test.yml@main
+    uses: ./.github/workflows/reusable-test.yml
     with:
       node_version: "[14, 16, 18]"
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -110,13 +110,6 @@ jobs:
           - chromium
           # - firefox
           - webkit
-        include:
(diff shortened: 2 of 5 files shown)


.github/workflows/release-test.yml
```diff
@@ -14,6 +14,6 @@ on:
 jobs:
   test:
     if: github.repository == 'remix-run/remix'
-    uses: remix-run/remix/.github/workflows/reusable-test.yml@main
+    uses: ./.github/workflows/reusable-test.yml
```
mcansh, line 17: keeping the other workflows as is for now as it explicitly states where a reusable workflow is running

brophdawg11 (approved):

brophdawg11:
LGTM!

mcansh:
weird, updated the cache key here [`6f56f35`](https://github.com/remix-run/remix/pull/4049/commits/6f56f35672a336ce2c1771bc7ff936fad1518f20), but runs after it don't have webkit installed 🤔

PR #4052, opened 2022-08-23
docs(example/msw): run msw along with app based on env variable

simjes:


machour:
@simjes  it seems that we totally missed this PR when migrating to the new [examples][https://github.com/remix-run/examples] repository 😓 

Would you mind reopening it there? The [base template](https://github.com/remix-run/examples/tree/main/__template) changed a bit as well, make sure to update your code to reflect that.

Thank you! 🙏🏼

PR #4060, opened 2022-08-24
feat(templates): bring React 18 templates in line with `indie-stack` updates

MichaelDeBoey:
Mainly includes https://github.com/remix-run/indie-stack/commit/e8d39cf00d92d2d16abab93025e4b4e47f8c808c, https://github.com/remix-run/indie-stack/commit/a1a1c05dcc2612e20fe35bdff2567df82719b85b & https://github.com/remix-run/indie-stack/pull/125

### DIFF
--- a/templates/arc/app/entry.client.tsx
+++ b/templates/arc/app/entry.client.tsx
@@ -1,4 +1,22 @@
 import { RemixBrowser } from "@remix-run/react";
+import { startTransition, StrictMode } from "react";
 import { hydrateRoot } from "react-dom/client";
 
-hydrateRoot(document, <RemixBrowser />);
+function hydrate() {
+  startTransition(() => {
+    hydrateRoot(
+      document,
+      <StrictMode>
+        <RemixBrowser />
+      </StrictMode>
+    );
+  });
+}
+
+if (window.requestIdleCallback) {
+  window.requestIdleCallback(hydrate);
+} else {
+  // Safari doesn't support requestIdleCallback
+  // https://caniuse.com/requestidlecallback
+  window.setTimeout(hydrate, 1);
+}
--- a/templates/express/app/entry.server.tsx
+++ b/templates/express/app/entry.server.tsx
@@ -2,11 +2,33 @@ import { PassThrough } from "stream";
 import type { EntryContext } from "@remix-run/node";
 import { Response } from "@remix-run/node";
 import { RemixServer } from "@remix-run/react";
+import isbot from "isbot";
 import { renderToPipeableStream } from "react-dom/server";
 
 const ABORT_DELAY = 5000;
 
 export default function handleRequest(
+  request: Request,
+  responseStatusCode: number,
+  responseHeaders: Headers,
+  remixContext: EntryContext
+) {
+  return isbot(request.headers.get("user-agent"))
+    ? handleBotRequest(
+        request,
+        responseStatusCode,
+        responseHeaders,
+        remixContext
+      )
+    : handleBrowserRequest(
+        request,
+        responseStatusCode,
+        responseHeaders,
+        remixContext
+      );
+}
+
+function handleBotRequest(
   request: Request,
   responseStatusCode: number,
   responseHeaders: Headers,
@@ -18,7 +40,7 @@ export default function handleRequest(
     const { pipe, abort } = renderToPipeableStream(
       <RemixServer context={remixContext} url={request.url} />,
       {
-        onShellReady: () => {
+        onAllReady() {
           const body = new PassThrough();
 
           responseHeaders.set("Content-Type", "text/html");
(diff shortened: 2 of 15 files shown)


mcansh (approved):
LGTM, but I'll defer to @jacob-ebey

templates/express/app/entry.server.tsx
```diff
@@ -12,13 +13,17 @@ export default function handleRequest(
   responseHeaders: Headers,
   remixContext: EntryContext
 ) {
+  const callbackName = isbot(request.headers.get("user-agent"))
```
jacob-ebey, line 16: MJ wants these to be two explicit code-paths, something like:

```js
if (isBot) {
  return renderForBot(...);
}

return renderForBrowser(...);
```

templates/fly/app/entry.server.tsx
```diff
@@ -12,13 +13,17 @@ export default function handleRequest(
   responseHeaders: Headers,
   remixContext: EntryContext
 ) {
+  const callbackName = isbot(request.headers.get("user-agent"))
```
jacob-ebey, line 16: MJ wants these to be two explicit code-paths, something like:

```js
if (isBot) {
  return renderForBot(...);
}

return renderForBrowser(...);
```

templates/remix/app/entry.server.tsx
```diff
@@ -12,13 +13,17 @@ export default function handleRequest(
   responseHeaders: Headers,
   remixContext: EntryContext
 ) {
+  const callbackName = isbot(request.headers.get("user-agent"))
```
jacob-ebey, line 16: MJ wants these to be two explicit code-paths, something like:

```js
if (isBot) {
  return renderForBot(...);
}

return renderForBrowser(...);
```

jacob-ebey (changes requested):

mcansh (approved):

templates/express/app/entry.server.tsx
```diff
@@ -2,6 +2,7 @@ import { PassThrough } from "stream";
 import type { EntryContext } from "@remix-run/node";
 import { Response } from "@remix-run/node";
 import { RemixServer } from "@remix-run/react";
+import isbot from "isbot";
```
mjackson, line 5: This isn't a huge deal, but TS marks this import as missing in dev. I've always tried to avoid this kind of thing in the past. Not sure the best way to fix it here though since this is in our templates.

Maybe we just add `isbot` to our monorepo deps to avoid the TS warning? Just a thought. Open to your suggestions.

Screen Shot 2022-09-29 at 3 49 27 PM

templates/arc/app/entry.client.tsx
```diff
@@ -1,4 +1,22 @@
 import { RemixBrowser } from "@remix-run/react";
+import { startTransition, StrictMode } from "react";
 import { hydrateRoot } from "react-dom/client";
 
-hydrateRoot(document, <RemixBrowser />);
+const hydrate = () => {
```
mjackson, line 5: Small nit: we prefer to use `function` for top-level module scope functions.

```ts
function hydrate() {
  // ...
}
```

templates/express/app/entry.server.tsx
```diff
@@ -12,7 +13,68 @@ export default function handleRequest(
...
+  responseHeaders: Headers,
+  remixContext: EntryContext
+) =>
+  new Promise((resolve, reject) => {
+    let didError = false;
+
+    const { pipe, abort } = renderToPipeableStream(
+      <RemixServer context={remixContext} url={request.url} />,
+      {
+        onAllReady: () => {
```
mjackson, line 43: Could also use a regular function here to save a few chars:

```suggestion
        onAllReady() {
```

templates/express/app/entry.server.tsx
```diff
@@ -12,7 +13,68 @@ export default function handleRequest(
...
+      )
+    : handleBrowserRequest(
+        request,
+        responseStatusCode,
+        responseHeaders,
+        remixContext
+      );
+}
+
+const handleBotRequest = (
```
mjackson, line 31: Please use a regular `function` here to remain consistent with the rest of the file.

templates/arc/app/entry.client.tsx
```diff
@@ -1,4 +1,22 @@
...
+  startTransition(() => {
```
mjackson, line 6: 👍  for using this pattern by default in our templates. Very cool 😎

mjackson (changes requested):
Hey @MichaelDeBoey this looks great, thank you. I'd just like a few small syntax changes to remain consistent with the rest of the codebase.

I know I said I could show you how to dedup the code, but I think I prefer it how it is now. It's clear and concise. If people want ternaries, they can make that refactoring themselves.

Thank you! ❤️

templates/express/app/entry.server.tsx
MichaelDeBoey: #4061 will add `isbot` to the monorepo deps, so these errors should be gone once that one's merged I think.

templates/express/app/entry.server.tsx
MichaelDeBoey: These were already arrow functions, so I just kept it like they were.

Will revert

mcansh (approved):

PR #4061, opened 2022-08-24
feat(integration): bring React 18 fixtures in line with `indie-stack` updates

MichaelDeBoey:
Follow-up of #4060

### DIFF
--- a/integration/helpers/node-template/app/entry.client.tsx
+++ b/integration/helpers/node-template/app/entry.client.tsx
@@ -1,4 +1,22 @@
 import { RemixBrowser } from "@remix-run/react";
+import { startTransition, StrictMode } from "react";
 import { hydrateRoot } from "react-dom/client";
 
-hydrateRoot(document, <RemixBrowser />);
+function hydrate() {
+  startTransition(() => {
+    hydrateRoot(
+      document,
+      <StrictMode>
+        <RemixBrowser />
+      </StrictMode>
+    );
+  });
+}
+
+if (window.requestIdleCallback) {
+  window.requestIdleCallback(hydrate);
+} else {
+  // Safari doesn't support requestIdleCallback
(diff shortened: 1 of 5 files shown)


integration/helpers/node-template/app/entry.client.tsx
```diff
@@ -1,4 +1,22 @@
...
+  });
+};
+
+if (window.requestIdleCallback) {
+  window.requestIdleCallback(hydrate);
+} else {
+  // Safari doesn't support requestIdleCallback
+  // https://caniuse.com/requestidlecallback
+  window.setTimeout(hydrate, 1);
+}
```
MichaelDeBoey, line 22: This isn't really necessary for our integration tests, but I think it's a good idea to keep these consistent with our templates

package.json
```diff
@@ -121,5 +121,8 @@
   },
   "engines": {
     "node": ">=14"
+  },
+  "devDependencies": {
+    "isbot": "^3.5.1"
```
mcansh, line 126: all dependencies in the workspace root are "dev dependencies" so we just put them all in "dependencies"

mcansh (approved):
LGTM, but I'll defer to @jacob-ebey

integration/helpers/node-template/app/entry.server.tsx
```diff
@@ -12,13 +13,17 @@ export default function handleRequest(
   responseHeaders: Headers,
   remixContext: EntryContext
 ) {
+  let callbackName = isbot(request.headers.get("user-agent"))
```
jacob-ebey, line 16: MJ wants these to be two explicit code-paths, something like:
```js
if (isBot) {
  return renderForBot(...);
}

return renderForBrowser(...);
```

jacob-ebey (changes requested):

Issue #1884, opened 2022-02-11
[Bug] Weird behavior of Remix in dev mode


lagz0ne:
### What version of Remix are you using?

1.1.3

### Steps to Reproduce

Environment
Windows 11
WSL2

Reproducible code https://github.com/lagz0ne/remix-bug

The code doesn't really work in Dev mode but works well in Production mode
The action part seems to be called but didn't persist the change to the content. As such, the loader takes the content from the original stage, not sure this has anything to do with the model of rebuilding (please have a look at the index route in the repository).

### Expected Behavior

Works in Dev mode

### Actual Behavior

Doesn't function in dev mode

cmd-johnson:
See #1818 and #1832.
In dev mode, Remix currently purges the `require` cache on every single request, meaning your `cache` object will be initialized to `["1"]` every time you navigate to your Index route.

You can work around this by attaching your cache to the `global` object. This is how I did it:
```ts
const cache = (() => {
  const globalCache = global as { cache?: string[] };
  if (!globalCache.cache) {
    globalCache.cache = ["1"];
  }
  return globalCache.cache;
})();
```

This might be a bit verbose, but it's probably the best you can do for now. Just remember that when using the `global` object, the cache will only be re-created when you restart your dev server. So every time you change your cache initialization code, you'll have to manually restart your app.

balzdur:
It may be similar to [this issue](https://github.com/remix-run/remix/issues/1555).

A promising solution is proposed [here](https://github.com/remix-run/remix/pull/4294).

In the meantime, you can use `patch-package` as proposed [here](https://github.com/remix-run/remix/issues/1555#issuecomment-1259976507).

pcattori:
Fixed by https://github.com/remix-run/remix/pull/5133 which will make the new dev server available in the next minor release under the `unstable_dev` future flag.

PR #4066, opened 2022-08-25
docs(other-api/serve): add preserve cache in development example

jmorel88:
Update docs with an example workaround for preserving cache in development. Example from #1884.


docs/other-api/serve.md
```diff
@@ -37,6 +37,25 @@ In development, `remix-serve` will ensure the latest code is run by purging the
   }
   ```
 
+  If you need a work around for preserving cache in development, you can store it in the global variable.
+
+  ```ts [1-9]
```
MichaelDeBoey, line 42:
```suggestion
  ```ts lines=[1-9] 
```

docs/other-api/serve.md
```diff
@@ -37,6 +37,25 @@ In development, `remix-serve` will ensure the latest code is run by purging the
   }
   ```
 
+  If you need a work around for preserving cache in development, you can store it in the global variable.
```
MichaelDeBoey, line 40:
```suggestion
  If you need a workaround for preserving cache in development, you can store it in the global variable.
```

MichaelDeBoey (changes requested):

docs/other-api/serve.md
```diff
@@ -37,6 +37,25 @@ In development, `remix-serve` will ensure the latest code is run by purging the
...
+    if (!global.uniqueCacheName) {
+      global.uniqueCacheName = new Map()
+    }
+
+    return global.uniqueCacheName
+  })
+
+
+  export async function loader({ params }) {
+    // ...
```
cliffordfajardo, line 55: I think it would be good if we added few more lines of code in the `loader` just like in the example from like 29-37 in this file 😄 

At a first glance, it may not be clear to people if the need to do use the cache like this:

```
export async function loader({ params }) {
    if (global.uniqueCacheName.has(params.foo)) {
```
or

```
export async function loader({ params }) {
   if (cache.has(params.foo)) {
```
What do you think?

jmorel88: Yeah I see how that can be confusing. I've added some more lines of example code to the loader function.

docs/other-api/serve.md
```diff
@@ -32,7 +32,32 @@ In development, `remix-serve` will ensure the latest code is run by purging the
...
+  ```
+
+  If you need a workaround for preserving cache in development, you can store it in the global variable.
+
+  ```ts lines=[1-9] 
+  // since the cache is stored in global it will only 
+  // be recreated when you restart your dev server.
+  const cache = (() => {
+    if (!global.uniqueCacheName) {
+      global.uniqueCacheName = new Map()
```
machour, line 47:
```suggestion
      global.uniqueCacheName = new Map();
```

docs/other-api/serve.md
```diff
@@ -32,7 +32,32 @@ In development, `remix-serve` will ensure the latest code is run by purging the
...
+    }
+
+    return global.uniqueCacheName
```
machour, line 50:
```suggestion
    return global.uniqueCacheName;
```

machour (changes requested):
Just a small update to respect the code style, otherwise looks great!

chaance (approved):

PR #4067, opened 2022-08-25
docs(other-api/react-router): update `react-router` links

Hawk94:
# Update broken links to React Router documentation

This change updates links to the React Router documentation for APIs that might be useful when developing Remix apps. Points to documentation on the main branch so that link stays up to date as project is updated.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -145,6 +145,7 @@
 - gunners6518
 - hadizz
 - hardingmatt
+- hawk94
 - helderburato
 - HenryVogt
 - hicksy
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (commented):
I think this is a bug on the `react-router` website and should be fixed there.

`https://reactrouter.com/docs/components/outlet` is redirecting to `https://reactrouter.com/en/v6.3.0/components/outlet`, which gives a 404 but should be the same as `https://reactrouter.com/en/main/components/outlet`

CC/ @mcansh

mcansh:
> 

i believe we've stopped the "v6" semver range mapping to "main" with the new site to simplify things and these docs don't exist on v6.3.0 - I'll ping @ryanflorence and see what he wants to do here :)

machour:
This have been fixed on the website, and all the urls now redirect where needed. 
Thank you for this PR!

PR #4083, opened 2022-08-27
chore: enable Dependabot for GitHub Actions

MichaelDeBoey:
This way we don't have to update them manually anymore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: daily


sergiodxa (approved):

machour (commented):
Wouldn't daily generate too much noise ?
I'd personally go with weekly

MichaelDeBoey:
@machour It's a best practice to use major version number on GitHub Actions, so it will only be very periodically that we're getting a PR

machour (approved):
Oh, these are GH actions not node dependencies ☕️

mcansh:
i really wish dependabot had grouped updates 😕

PR #4085, opened 2022-08-27
chore(examples): set remix packages' version to `'*'`

MichaelDeBoey:
This way we don't have to manually update the version anymore + it's in line with our templates to be used with `create-remix`


examples/basic/package.json
```diff
@@ -7,15 +7,15 @@
     "start": "remix-serve build"
   },
   "dependencies": {
-    "@remix-run/node": "1.6.4",
```
MichaelDeBoey, line 10: @mcansh This should have been at `1.7.0` btw, so it seems like the version script didn't run for some reason 🤔

machour (approved):
Heh, was thinking of proposing the same thing this morning while running an example locally.

MichaelDeBoey:
@machour Great minds think alike 😉

mcansh:
does this break "open in code sandbox"?

MichaelDeBoey:
@mcansh No it doesn't

- https://codesandbox.io/s/github/remix-run/remix/tree/main/examples/basic
- https://codesandbox.io/s/github/MichaelDeBoey/remix/tree/set-remix-packages-version/examples/basic

examples/basic/package.json
mcansh: yeah changesets does the bumping now, but before we would always use `scripts/version.js` which would bump the examples

mcansh (approved):

PR #4086, opened 2022-08-27
`Form` component doesn't redirect like native component

kevlened:
- [ ] Docs
- [ ] Tests

Issue:

Redirecting after a native form post works successfully.

Redirecting after using Remix's `Form` component does not.

Testing Strategy:

1) I reproduced it locally
2) [I reproduced it in stack blitz](https://stackblitz.com/edit/node-ekrxhg?file=app/routes/index.tsx)
3) I wrote a test per the bug report process, but the test didn't fail

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -192,6 +192,7 @@
 - KenanYusuf
 - kentcdodds
 - kevinrambaud
+- kevlened
 - kgregory
 - kilian
 - kiliman
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -7,68 +7,38 @@ import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
 let fixture: Fixture;
 let appFixture: AppFixture;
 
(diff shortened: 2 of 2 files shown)


machour:
Hi @kevlened 👋
We need you to sign the CLA in order for the test suite to run, could you do that?

kevlened:
@machour done!

machour:
@kevlened thanks! I tested the stack blitz you provided, and looking at the network tab, I see the #remix post stays pending forever with errors in the console. Can you check again locally and inspect thoroughly the network & console tabs? Something else might be going on here, like one of your extensions causing React to trip & fail.

machour:
Update, it just completed with a `Failed to fetch` after 5 minutes 😅 My network seems to be doing fine tho.

Screen Shot 2022-08-28 at 10 32 16 PM

kevlened:
Interesting. My local testing was in my app, but after creating a fresh remix project, I was unable to repro the behavior. I'll close this for now.

Thanks!

PR #4091, opened 2022-08-28
docs(examples/blog): Remove `_` as separator in port number to avoid confusion

grantnorwood:
There is an underscore `_` in the `internal_port` number in blog tutorial's `fly.toml`, which is inconsistent from other references to that port number, and I've simply removed that.  No other code or configuration changes have been made.

As this change is located within an example template, I've branched from `main` instead of `dev`.  _(If this PR should instead be branched from `dev`, please let me know!)_

Signed CLA by including `grantnorwood` in `contributors.yml`.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -140,6 +140,7 @@
 - goncy
 - gonzoscript
 - graham42
+- grantnorwood
 - GregBrimble
 - GSt4r
 - guatedude2
--- a/examples/blog-tutorial/fly.toml
+++ b/examples/blog-tutorial/fly.toml
@@ -19,7 +19,7 @@ source = "data"
 destination = "/data"
 
 [[services]]
-internal_port = 8_080
+internal_port = 8080
 processes = [ "app" ]
 protocol = "tcp"
 script_checks = [ ]


machour:
Thank you for opening the PR. _ is valid here, and helps readability, see https://toml.io/en/v1.0.0#integer

Same thing is possible with JavaScript too, if you ever need it!

grantnorwood:
So I disagree that this PR should've been closed so quickly, especially on the merits of readability.

I understand that Toml allows underscores in integers, and in some cases that can certainly be more readable, especially with large numbers.  However, the underscore in this PR is **absolutely not** more readable because `8080` is not a large number.  It's also not familiar to many new web developers who haven't used a `.toml` file – which is likely most of the devs making use of these tutorial files – and so again less readable/understandable.

Nor is it consistent with other uses of `internal_port` in other `.toml` files in the repo:

- https://github.com/remix-run/remix/blob/main/examples/jokes/fly.toml#L24

And in fact, even within the same file, `8080` is used in strings _(where obviously underscores aren't allowed)_, so inconsistencies in that value make changing the port with a find/replace more difficult if `internal_port` contains an underscore.

Could we please get some more input from other reviewers to determine whether my suggestion/PR is an improvement, as I genuinely believe it is?

machour:
@grantnorwood To be fair, you opened the PR stating the underscore was **accidental** (meaning a typo). I closed it because it wasn't true.

> However, the underscore in this PR is absolutely not more readable because 8080 is not a large number.

This is a pure matter of taste. Yours in this case :)

Reopening to let the team decide if they want to remove the underscore or not. Cheers !

grantnorwood:
Your point is valid!  Thanks for updating the title.  I updated my original description to match.

Thank you 🙂

PR #4094, opened 2022-08-28
feat(remix-dev): allow importing `.psd` files

NikkiDelRosso:
Adds support for importing photoshop image files as assets.

Not sure if there's any need to add the MIME type to the `binaryTypes.ts` files in remix-architect and remix-netlify, but fwiw the MIME is `image/vnd.adobe.photoshop`

## Testing Strategy:

I was able to import a `.psd` file as an asset as expected:

```typescript
import sourceArt from "./assets/art.psd"

export default function PsdLink() {
  return (
    <a href={sourceArt}>Download PSD</a>
  )
}
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -306,6 +306,7 @@
 - nickytonline
 - niconiahi
 - nielsdb97
+- NikkiDelRosso
 - ninjaPixel
 - niwsa
 - nobeeakon
--- a/packages/remix-dev/compiler/loaders.ts
+++ b/packages/remix-dev/compiler/loaders.ts
@@ -25,6 +25,7 @@ export const loaders: { [ext: string]: esbuild.Loader } = {
   ".ogg": "file",
   ".otf": "file",
   ".png": "file",
+  ".psd": "file",
   ".sql": "text",
   ".svg": "file",
   ".ts": "ts",
--- a/packages/remix-dev/modules.ts
+++ b/packages/remix-dev/modules.ts
@@ -72,6 +72,10 @@ declare module "*.png" {
   let asset: string;
   export default asset;
 }
+declare module "*.psd" {
+  let asset: string;
+  export default asset;
+}
 declare module "*.sql" {
   let asset: string;
   export default asset;


kentcdodds:
Yeah, we should probably make it configurable or something

kiliman:
Perhaps we can just add support for a `file:` prefix and maybe even a `content:` prefix to  return the content as a string. Not sure how doable this is, although I remember Remix used to have a `css:` handler at one point.

```ts
import psdFile from 'file:~/app/files/example.psd'
import txtFile from 'content:~/app/files/data.txt'
```

sergiodxa:
I like @kiliman idea, instead of a config file allow any import to have file: or content:, that could even allow things like content:path/to/file.ts and get the TS file as a string instead of doing an actual import

machour:
We would also need to be able to serve the file with the accurate mime-type when it's server from assets. Any ideas on how to implement this as well?

NikkiDelRosso:
Hiya @machour, thanks for taking a look at this! I'm glad to have started a conversation  - I was surprised after reading through the Remix docs on asset files that I had to apply a patch rather than update a configuration file somewhere to be able to do this. 😅 

Let me know if there's any way I can assist.

sergiodxa:
> 

🤔 I think either Remix lets you configure this in remix.config.js (like an extensions option) or Remix trust the server to send the correct mime types on files from the public folder including public/build

kiliman:
I've created a patch to allow you to specify the server file to use with Remix App Server. It even supports TS files. You just need to export a function named `createApp`. I want to add an `eject` command that will generate the *server* file for you. The cool thing about this is you now get to customize the server all you want but still get the DX benefit of using `npm run dev`. No need to spawn a separate express server.

BTW: I had to create a new config option. The `server` option actually bundles the server code with your app, which is not what I wanted. I plan to submit a PR soon.

```ts
//server.ts
import morgan from "morgan";

export function createApp(
  buildPath: string,
  mode = "production",
  publicPath = "/build/",
  assetsBuildDirectory = "public/build/"
) {
  let app = express();

  app.disable("x-powered-by");
  app.use(function (req, res, next) {
    res.setHeader("X-Powered-By", "TS Remix App Server");
    next();
  });
  app.use(compression());

  app.use(
    publicPath,
    express.static(assetsBuildDirectory, { immutable: true, maxAge: "1y" })
  );

  app.use(morgan("tiny"));
  app.all(
    "*",
    mode === "production"
      ? createRequestHandler({ build: require(buildPath), mode })
      : (req, res, next) => {
          // require cache is purged in @remix-run/dev where the file watcher is
          let build = require(buildPath);
          return createRequestHandler({ build, mode })(req, res, next);
        }
  );

  return app;
}
```
```js
// remix.config.js
module.exports = {
  ignoredRouteFiles: ["**/.*"],
  serverEntryFile: "./server.ts",
  // appDirectory: "app",
  // assetsBuildDirectory: "public/build",
  // serverBuildPath: "build/index.js",
  // publicPath: "/build/",
};
```
### Notice the X-Powered-By header
image

kiliman:
Sorry, commented on wrong issue.

chaance (approved):

PR #4104, opened 2022-08-30
Make the dark mode example work without JS

silvenon:
@machour encouraged me to show you how I implemented a dark mode switch which works without JavaScript. It's derived from Matt's excellent blog post, just like your example, but it uses regular form submission instead of `useFetcher` in combination with optimistic UI to maintain the same speed (hopefully). Also, I'm aware that the architecture of this feature is largely a matter of taste, I tried to change as little as possible.

The main improvements are:

1. `persistThemeRef` can be removed, one less hack to think about
2. we don't really need to include `setTheme` in the context, so the context will update only when the theme value truly changes, not every time `ThemeProvider` re-renders, and without having to use `useMemo`
3. when the `(prefers-color-scheme: dark)` changes it will no longer be submitted as the specified theme, which was probably not the intended behavior

Is it a problem that the URL bar quickly switches to `/action/set-theme` and back? It probably causes a few extra re-renders, I could use `useFetcher` when JS is supported in the spirit of progressive enhancement, what do you think?

Ideas for further improvements:

- The dark mode toggle on my blog also features a reset button, which allows people to sync dark mode back to the OS settings. I can add it here as well if you want it.
- `ThemeHead` and `ThemeBody` are within `ThemeProvider`, I can make it so that it's not necessary to pass `ssrTheme` to them because `ThemeProvider` already knows about it.

- [x] Docs
- [ ] Tests

No tests yet, however I'd be very happy to write Playwright tests for this if you want because dark mode is a tricky functionality.

Btw, I'll probably make a `remix-dark-mode` library out of this soon, if there isn't one already. 🤔

### DIFF
--- a/examples/dark-mode/app/routes/action/set-theme.tsx
+++ b/examples/dark-mode/app/routes/action/set-theme.tsx
@@ -6,9 +6,9 @@ import { isTheme } from "~/utils/theme-provider";
 
 export const action: ActionFunction = async ({ request }) => {
   const themeSession = await getThemeSession(request);
-  const requestText = await request.text();
-  const form = new URLSearchParams(requestText);
+  const form = await request.formData();
   const theme = form.get("theme");
+  const redirectTo = new URL(request.url).searchParams.get("redirectTo");
 
   if (!isTheme(theme)) {
     return json({
@@ -18,10 +18,14 @@ export const action: ActionFunction = async ({ request }) => {
   }
 
   themeSession.setTheme(theme);
-  return json(
-    { success: true },
-    { headers: { "Set-Cookie": await themeSession.commit() } }
-  );
+
+  const headers = { "Set-Cookie": await themeSession.commit() };
+
+  if (redirectTo) {
+    return redirect(redirectTo, { headers });
+  } else {
+    return json({ success: true }, { headers });
+  }
 };
 
 export const loader: LoaderFunction = () => redirect("/", { status: 404 });
--- a/examples/dark-mode/app/routes/action/set-theme.tsx
+++ b/examples/dark-mode/app/routes/action/set-theme.tsx
@@ -8,12 +8,19 @@ export const action: ActionFunction = async ({ request }) => {
   const themeSession = await getThemeSession(request);
   const form = await request.formData();
   const theme = form.get("theme");
-  const redirectTo = new URL(request.url).searchParams.get("redirectTo");
+  const redirectTo = form.get("redirectTo");
 
   if (!isTheme(theme)) {
     return json({
       success: false,
       message: `theme value of ${theme} is not a valid theme`,
+    });
+  }
+
+  if (redirectTo && typeof redirectTo !== "string") {
+    return json({
+      success: false,
+      message: `redirect URL should be a string`,
     });
   }
 
@@ -23,9 +30,9 @@ export const action: ActionFunction = async ({ request }) => {
 
   if (redirectTo) {
     return redirect(redirectTo, { headers });
-  } else {
-    return json({ success: true }, { headers });
   }
+
+  return json({ success: true }, { headers });
 };
 
 export const loader: LoaderFunction = () => redirect("/", { status: 404 });
--- a/examples/dark-mode/app/utils/theme-provider.tsx
+++ b/examples/dark-mode/app/utils/theme-provider.tsx
@@ -108,14 +108,10 @@ function ThemeToggle({
   className?: string;
 }) {
   const location = useLocation();
-  const searchParams = new URLSearchParams({ redirectTo: location.pathname });
 
   return (
-    <Form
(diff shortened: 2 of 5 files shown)


examples/dark-mode/app/routes/action/set-theme.tsx
```diff
@@ -18,10 +18,14 @@ export const action: ActionFunction = async ({ request }) => {
...
-    { headers: { "Set-Cookie": await themeSession.commit() } }
-  );
+
+  const headers = { "Set-Cookie": await themeSession.commit() };
+
+  if (redirectTo) {
+    return redirect(redirectTo, { headers });
+  } else {
+    return json({ success: true }, { headers });
+  }
```
machour, line 28: You can remove a bit of cognitive complexity here by removing the else branch:

```js
  if (redirectTo) {
    return redirect(redirectTo, { headers });
  }
   
 return json({ success: true }, { headers });
```

examples/dark-mode/app/utils/theme-provider.tsx
```diff
@@ -52,43 +85,43 @@ function ThemeProvider({
...
+      action={`/action/set-theme?${searchParams}`}
+      method="post"
+      className={className}
+    >
+      <Themed
+        dark={<input type="hidden" name="theme" value={Theme.LIGHT} />}
+        light={<input type="hidden" name="theme" value={Theme.DARK} />}
+      />
+      <button type="submit">{children}</button>
+    </Form>
```
machour, line 124: By using a GET parameter for `redirectTo`, you're opening yourself to an "Open redirection XSS" where a malicious user can pretend to be your website and redirect users elsewhere :

`https://yoursite.com/action/set-theme?redirectTo=https://malwarewebsite.com`

How about making `redirectTo` a hidden input in your form, that would get POSTed instead, defeating the attack?

machour (changes requested):
Thank you sooo much @silvenon 🙌🏼 
Just a little nit & a security concern to address before validation

examples/dark-mode/app/utils/theme-provider.tsx
machour: hum 😅, not really an issue since this is handled in your `action`, so it's harder to forge the attack, but still a healthy habit to avoid passing redirects in the path

silvenon: Good point ✅

machour (changes requested):
I tested the PR locally, and there's a couple of things to fix:

* as you noted, with JS on, there's an `/action/set-theme` that's briefly shown in the browser. (The no-js experience is better here 😄).  `useFetcher` would be welcome here indeed.

* with JS disabled, when I open the page without a cookie, both dark & light strings are originally displayed:

Screen Shot 2022-09-05 at 12 09 34 AM

* not your code, but in `set-theme.tsx` the loader does a redirect() with a `404` status. I don't think this is valid, a 302 would be better here

examples/dark-mode/app/routes/action/set-theme.tsx
```diff
@@ -17,11 +17,22 @@ export const action: ActionFunction = async ({ request }) => {
...
+  const headers = { "Set-Cookie": await themeSession.commit() };
+
+  if (redirectTo) {
+    return redirect(redirectTo, { headers });
+  }
+
+  return json({ success: true }, { headers });
 };
 
 export const loader: LoaderFunction = () => redirect("/", { status: 404 });
```
machour, line 38:
```suggestion
export const loader: LoaderFunction = () => redirect("/", { status: 302 });
```

silvenon: Hm, what's the reasoning behind this? This will only trigger when people type this in the URL bar, I thought that it made sense to return 404 considering that there's nothing there to load, only submit, but that's internal and none of their business 😜

silvenon:
> I tested the PR locally, and there's a couple of things to fix

Thanks for catching this, I'll think for a moment what the best way for fixing would be. My plan is to create a `utils/no-js.jsx` helper similar to `utils/theme-provider.jsx` which makes it easy to render stuff differently based on whether JS is supported.

examples/dark-mode/app/routes/action/set-theme.tsx
machour: Not sure either, but a redirect with a 404 code clearly isn't the way 😅

machour:
@silvenon adding another file could make the example a bit more complex to understand.. I'll let you judge that!

examples/dark-mode/app/routes/action/set-theme.tsx
silvenon: That's a good point 😅

chaance:

silvenon:
Okay, will do! 👌 Once I come up with something elegant.

PR #4107, opened 2022-08-31
feat(remix-dev): add `devServerSecureContextOptions` config option

kyleharper:
My team have experienced issues with live reloading once moving over to https locally (the following issue documents the problem: https://github.com/remix-run/remix/issues/2859)

This PR resolves that by adding a new config item called `devServerSecureContextOptions`. This name is derived from the `tls` node package type definitions.

Big shout out for @isaacs for his PR (https://github.com/remix-run/remix/pull/2861), which helped me understand where to find the server implementation.

I'm happy for either PR to be merged in, the other PR is quite old and I didn't want to resolve conflicts on another fork etc. 

One difference is that this PR allows us to pass in more than just a string as `key` and `cert` can be any of the following;

- key: `string | Buffer | Array<Buffer | KeyObject> | undefined`
- cert: `string | Buffer | Array<string | Buffer> | undefined`

For now we're implementing a patch internally to get round this issue, but it'd be great to not require a patch 👍

Closes: https://github.com/remix-run/remix/issues/2859

- [ ] Docs
- [ ] Tests

Testing Strategy:

On the testing front we've covered all failing tests but couldn't find anything around the `watch` command. Perhaps we missed something or is there no tests for the watch command? (due to mocking challenges etc?)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -422,3 +422,4 @@
 - zachdtaylor
 - zainfathoni
 - zhe
+- kyleharper
--- a/packages/remix-dev/__tests__/readConfig-test.ts
+++ b/packages/remix-dev/__tests__/readConfig-test.ts
@@ -30,6 +30,7 @@ describe("readConfig", () => {
         "cacheDirectory": Any<String>,
         "devServerBroadcastDelay": 0,
         "devServerPort": Any<Number>,
+        "devServerSecureContextOptions": undefined,
         "entryClientFile": "entry.client.tsx",
         "entryServerFile": "entry.server.tsx",
         "mdx": undefined,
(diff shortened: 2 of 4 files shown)


isaacs:
Awesome!  Glad to know that old PR was able to help.

kiliman:
I created a PR #4123 so you can specify your own server file for Remix App Server to customize your setup. This way Remix doesn't have to add a bunch of different configuration options. It gives you the flexibility of the Express adapter, but the DX of being able to simply run `remix dev`.

kyleharper:
That sounds great @kiliman 

I'll make some time either today or tomorrow and give it a test. 

I think I'll prefer your approach but need to give it a go first 👍

kyleharper:
Closing. See https://github.com/remix-run/remix/pull/2861 for more details 👍 

Looking forward to using the new dev server ❤️

Issue #3772, opened 2022-07-16
Typecasting useLoaderData doesn't work in blog tutorial


dca123:
### What version of Remix are you using?

1.6.5

### Steps to Reproduce

Follow the [tutorial](https://remix.run/docs/en/v1/tutorials/blog) till the [Dynamic routing section](https://remix.run/docs/en/v1/tutorials/blog#dynamic-route-params). I was able to use generic to resolve the issue. I've attatched a PR to this issue that should fix this problem

P.S, I'm a fresh contributer, so I'd appreciate any tips !
Regards,
Dev

### Expected Behavior

Able to continue as shown in the tutorial

### Actual Behavior

Using `const { posts } = useLoaderData() as LoaderData;`  results in the following error 

```
Conversion of type '{ posts: { slug: string; title: string; markdown: string; createdAt: string; updatedAt: string; }[]; }' to type 'LoaderData' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Types of property 'posts' are incompatible.
    Type '{ slug: string; title: string; markdown: string; createdAt: string; updatedAt: string; }[]' is not comparable to type 'Post[]'.
      Type '{ slug: string; title: string; markdown: string; createdAt: string; updatedAt: string; }' is not comparable to type 'Post'.
        Types of property 'createdAt' are incompatible.
          Type 'string' is not comparable to type 'Date'.
```

dca123:
Here is my PR #3773 https://github.com/remix-run/remix/pull/3773

jtushman:
FYI -- I also ran into this as a +1

PR #4115, opened 2022-09-02
docs(tutorial/blog): update the example to use LoaderArgs and ActionArgs

### LINKED ISSUE #4110: add SerializeFrom to Developer Blog Tutorial

1.7.0

### Steps to Reproduce

I followed the getting started instructions and began the [Developer Blog Tutorial](https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md).

### Expected Behavior

I expected to not have type errors while following the tutorial.

### Actual Behavior

In the "A little refactoring" section, the return value of `useLoaderData` has a type error of "Type 'string' is not assignable to type 'Date'", because of a mismatch between `Date` and `string` types generated by Prisma and the serialized type of the `createdAt` and `updatedAt` properties.

I did a little digging and found the `SerializeFrom` type that is new to `1.7.0`. I was able to fix the type error using the following code. Though because there are no examples of this, I have no idea if this is the proper way to type LoaderData.

```ts
import type {SerializeFrom } from "@remix-run/node";
import { Link, useLoaderData } from "@remix-run/react";

import { getPosts } from "~/models/post.server";

  return json({
    posts: await getPosts(),
  });
};

type LoaderData = SerializeFrom<typeof loader>;

export default function Posts() {
  const { posts } = useLoaderData() as LoaderData;
  return (
    <main>
      <h1>Posts</h1>
      <ul>
        {posts.map((post) => (
          <li key={post.slug}>
            <Link to={post.slug} className="text-blue-600 underline">
              {post.title}
            </Link>
          </li>
        ))}
      </ul>
    </main>
  );
}
```

johnhooks:
Closes #4110

I removed the `LoaderData` type to simplify the tutorial and avoid a type error that happens after adding Prisma in the tutorial.


docs/tutorials/blog.md
```diff
@@ -203,12 +203,8 @@ type Post = {
   title: string;
 };
 
-type LoaderData = {
-  posts: Array<Post>;
-};
-
 export const loader = async () => {
-  return json<LoaderData>({
+  return json<{ posts: Array<Post>; }>({
```
silvenon, line 207: Unless you want to deliberately assert at this point what you expect the loader to return, you don't need this, `useLoaderData` will still be able to infer the type.
```suggestion
  return json({
```
I set the type only when I have something more complex, like disjoint union.

docs/tutorials/blog.md
```diff
@@ -203,12 +203,8 @@ type Post = {
   title: string;
 };
 
-type LoaderData = {
-  posts: Array<Post>;
-};
-
 export const loader = async () => {
```
silvenon, line 206: I would recommend turning this from a `export const` to an `export async function`, also you need to use `LoaderArgs`, but instead of `{}: LoaderArgs` you can use `args: LoaderArgs` or something like that, that way ESLint won't complain.

```ts
export async function loader(args: LoaderArgs) {
  // ...
}
```

Read release notes for [v1.6.5](https://github.com/remix-run/remix/releases/tag/remix%401.6.5) for more info.

silvenon (commented):
I hope you don't mind some outside review 🤞 I just want to help because I love this feature.

EDIT: Sorry, deleting, it's not my place. I'll leave it up to maintainers

machour:
@silvenon feel free to review any PR your want, starting with this one 🙌

docs/tutorials/blog.md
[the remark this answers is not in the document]
johnhooks: I had thought I was getting a type error without it, but after removing it everything is working fine. So I will definitely remove it.

docs/tutorials/blog.md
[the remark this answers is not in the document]
johnhooks: I can make that change, though the rest of the tutorial uses the const function definitions in several locations. Should I change all of the examples in the tutorial to make it consistent? It seems a bit beyond the scope of the reason for my PR, I just wanted to make sure others wouldn't be tripped up by a type issue while working through the tutorial. I couldn't find the right solution in the documentation, and didn't figure out the right method to fix the issue until after opening the issue.

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L136

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L296

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L456

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L481

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L519

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L553

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L668

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L750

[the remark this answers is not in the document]
machour: @johnhooks we need to update the docs to use the new inference since the 1.6.5 release, but we didn't find the time to do it. 

I wouldn't mind at all if you extended this PR scope to tackle the whole update. WDYT ?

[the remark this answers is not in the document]
johnhooks: @machour I will definitely take that on. I assume that means most of the function definitions don't need to be typed as `LoaderFunction` but should instead have the args typed as `LoaderArgs`. Is there a better way to make sure the modified code examples pass the type checker, rather than copy and pasting into an actual remix project? Just want to make sure I'm not missing some handy way to type check in Markdown.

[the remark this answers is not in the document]
machour: 🙏
Exactly. Read through the 1.6.5 announcement and you'll see the ideal syntax we're looking for, for both loader and action. 
As @jenseng said, typing using `args` instead of `{}` should satisfy the linter. 
I'd really appreciate if you took the time to make the tutorial run locally and update the docs accordingly (you can even share the repo here so that we give it a look)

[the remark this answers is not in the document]
johnhooks: Also how are the line numbers in all the code snippets being generated? Some of the changes, especially adding `import type { LoaderArgs } from "@remix-run/node";` to the top of the example, will mess all of those up. Is there an app example folder that has the code examples in them? I know the Indie Stack is used, but this tutorial adds the `post.tsx` file.

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L132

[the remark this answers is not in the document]
machour: Line numbers gets added automatically, and the `lines=[1-2,4-17]` syntax allow to highlight the interesting lines (here, line 1 and 2, and line 4 to 17).

You'll have to count lines with your fingers to find the good values (and maybe bring up some nice childhood memories 😅)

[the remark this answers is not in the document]
johnhooks: Sorry for all the questions, but I just want to make sure while making all these changes I keep the appropriate intent of the tutorial. Because of the new type inference some sections about making Typescript happy aren't required. Though I could intentionally leave out the generic argument in the first use of `useLoaderData<typeof loader>()`, let it error and use the situation to explain that `useLoaderData` needs a generic argument in order to resolve the type. What do you think?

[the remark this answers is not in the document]
johnhooks: @machour I am going though everything, making sure the examples pass the type checker, and that the tutorial still makes sense with all the example edits. The repo is https://github.com/remix-run/remix, I'm still working on the changes, right now there isn't anything pushed to it to look at, beyond what's already in the PR. I'll let you know when I get it up.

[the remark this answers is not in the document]
machour: Ask as much questions as you need.

Since you were originally confused about this particular part, the approach you're proposing makes a lot of sense imo. 

We will review what you'll come up with and maybe adjust a thing or two, so don't worry and do as you see fit for now.

docs/tutorials/blog.md
```diff
@@ -192,32 +194,14 @@ export default function Posts() {
...
 
-💿 Add the Post type and generic for `useLoaderData`
+💿 Add the generic argument to `useLoaderData` to help it infer the return type of the loader function.
 
 ```tsx filename=app/routes/posts/index.tsx lines=[4-7,9-11,14,29]
+import type { LoaderArgs } from "@remix-run/node";
 import { Link, useLoaderData } from "@remix-run/react";
 
-type Post = {
```
johnhooks, line 201: No longer a need for a explicit Post type

docs/tutorials/blog.md
```diff
@@ -896,28 +845,24 @@ TypeScript is mad again, let's add some types.
...
-import type { Post } from "@prisma/client";
-export type { Post };
+import type { Prisma } from "@prisma/client"
 
 // ...
 
-export async function createPost(
-  post: Pick<Post, "slug" | "title" | "markdown">
-) {
+export async function createPost(post: Prisma.PostUncheckedCreateInput) {
```
johnhooks, line 852: Prisma generates a type of input, I thought its more flexible to use than a `Pick`

docs/tutorials/blog.md
```diff
@@ -462,20 +448,6 @@ export default function PostSlug() {
 
 The part of the filename attached to the `$` becomes a named key on the `params` object that comes into your loader. This is how we'll look up our blog post.
 
-💿 Let's get some help from TypeScript for the loader function signature.
```
johnhooks, line 474: No longer needed because of inference

docs/tutorials/blog.md
```diff
@@ -485,37 +457,31 @@ Update the `app/models/post.server.ts` file:
...
   return prisma.post.findMany();
 }
 
-export async function getPost(slug: string) {
+export async function getPost(slug?: string) {
   return prisma.post.findUnique({ where: { slug } });
 }
 ```
 
-If you see a TypeScript warning, such as `TS2305: Module '"@prisma/client"' has no exported member 'Post'.`, you may need to restart your editor.
```
johnhooks, line 508: This could be used later when actually using an imported Prisma type in the post create action

johnhooks:
I've updated everything and given it two proof reads. I haven't updated the highlighted line suggestions for the code examples, I thought that would be easiest to do after being able to preview the actual line numbers Github will give the examples, and I also don't want to waste time editing the line highlights if there need to be more edits.

docs/tutorials/blog.md
```diff
@@ -485,37 +457,31 @@ Update the `app/models/post.server.ts` file:
...
 import { prisma } from "~/db.server";
 
-export type { Post } from "@prisma/client";
-
 export async function getPosts() {
   return prisma.post.findMany();
 }
 
-export async function getPost(slug: string) {
+export async function getPost(slug?: string) {
```
machour, line 464: ℹ️ The `slug` parameter should not be made optional here.

If the problem is that `params.slug` is typed as `string|undefined`, you should instead do something along those lines:

```js
export async function loader({ params }: LoaderArgs) {
  if (typeof params.slug === "undefined") {
      throw new Response("Missing slug", 400);
  }

  const post = await getPost(params.slug); // <= from now on TS knows that slug is a `string`

```

johnhooks: Got it, I was curious whether I did that right.

machour: Rule of thumb is: never change a parameter signature to satisfy TypeScript. 

Use the type that makes sense (here, we do need a slug as a string) and then, make typescript understand that you're calling your function correctly.

johnhooks: I took a look at the example right after completing the `getPost` code, it explains about TS showing the error, the next section is about making TS happy and adding the code below. I will remove the `?`. The rest is already explained and the possibility of slug being undefined is handled.

https://github.com/remix-run/remix/blob/493b6a06d483562840ffc5e41311f8c7d5bb5c7c/docs/tutorials/blog.md?plain=1#L546-L556

docs/tutorials/blog.md
```diff
@@ -855,35 +855,13 @@ export async function createPost(post: Prisma.PostUncheckedCreateInput) {
 
 ```
 
-```tsx filename=app/routes/posts/admin/new.tsx lines=[1,7]
-import type { ActionArgs } from "@remix-run/node";
```
johnhooks: Section not necessary after implementing changes to the tutorial for 1.6.5 release. The types are already inferred.

johnhooks:
I updated the highlighted lines in the code examples. I believe the changes to implement 1.6.5 is complete.

johnhooks:
This is the Indie Stack filled with the new code examples:

https://github.com/johnhooks/blog-tutorial

The only issue I see, the links to the posts on the admin pages don't work because they are relative to the `/posts/admin` route and that route isn't found. Shouldn't it look like this?

```ts
<Link to={`/posts/${post.slug}`} className="text-blue-600 underline">
```

johnhooks:
@machour just out of curiosity would you like me to start in on the Jokes Tutorial and wrap that together with this or keep them separate PRs? I just peeked into that one and it needs to be updated too.

machour:
@johnhooks let's leave that for another PR.

I'd recommend waiting for feedback on this one before starting converting the Jokes tutorial, since some adjustments may be required.

runofthemill:
noting this covers #3772 / #3773 for anyone looking at that issue/PR

chaance:
Hi! We've migrated our examples to a [dedicated repo](https://github.com/remix-run/examples). If you can get rid of those changes to reduce merge conflicts, that'd be great. You're welcome to make a separate PR over there if you'd still like the examples updated.

johnhooks:
@machour if you guys are still interested in this, I can remove the example updates, though what would be the recommended git strategy to do so?

machour:
@johnhooks should go something like this for you inside your own repository checkout:

```sh
git fetch upstream                # get latest informations from the remix-run/remix repo (it's named upstream, and your fork is named origin)
git merge upstream main           # merge upstream into your own repo main (you'll get conflicts printed to the console)
git rm -rf examples/blog-tutorial # recursively force remove your local examples/blog-tutorial folder
git commit -m "remove examples folder"
git push
```

as for your changes in examples, I extracted and adapted a patch that you can apply against the new examples repository here: https://gist.github.com/machour/01a1f1aada08f29ad547b4ff22d679fe

johnhooks:
Thanks, there is a lot I need to learn about git

johnhooks:
@machour I think I got it resolve

PR #4123, opened 2022-09-02
feat(remix-dev): add `serverEntryFile` config option

kiliman:
Currently if a dev wants to extend the configuration of Remix App Server, they
must eject from RAS to the Express adapter. This PR adds a new config option 
`serverEntryFile` that specifies the file that will be used by RAS instead of
its default configuration. If the file is TypeScript, it will be transpiled
automatically.

```js
// remix.config.js
module.exports = {
  serverEntryFile: './server.ts'
}
```

Running `npm run dev` will launch Remix App Server and import the server file.

For production builds, the dev should update the `start` script to specify the
server file for `remix-serve`. NOTE: If the server file is TypeScript, this must
be transpiled manually during the build process.

```json
"build": "remix build && npm run build:server",
"build:server": "esbuild --format=cjs --platform=node --outfile=build/server.js server.ts",
"start": "remix-serve build build/server.js",
```
    
The server file should export the following functions:

```ts
// REQUIRED: create the Express app and return it
export function createApp(
  buildPath: string,
  mode = "production",
  publicPath = "/build/",
  assetsBuildDirectory = "public/build/"
): Express

// OPTIONAL: create an HTTP/HTTPS server and return it
function createServer(app: Express, port: number): Server

// OPTIONAL: create a WebSocket server and return it
function createSocketServer(port: number): WebSocket.Server
```

Here's a sample server file that includes all the exports.

https://github.com/kiliman/remix-ras-server-example/blob/main/server.ts

### DIFF
--- a/packages/remix-dev/__tests__/readConfig-test.ts
+++ b/packages/remix-dev/__tests__/readConfig-test.ts
@@ -22,6 +22,7 @@ describe("readConfig", () => {
         assetsBuildDirectory: expect.any(String),
         relativeAssetsBuildDirectory: expect.any(String),
         tsconfigPath: expect.any(String),
+        serverEntryFile: expect.any(String),
       },
       `
       Object {
@@ -43,6 +44,7 @@ describe("readConfig", () => {
             "path": "",
           },
         },
+        "serverBuildFile": undefined,
         "serverBuildPath": Any<String>,
         "serverBuildTarget": undefined,
         "serverBuildTargetEntryModule": "export * from \\"@remix-run/dev/server-build\\";",
@@ -53,6 +55,7 @@ describe("readConfig", () => {
         "serverPlatform": "node",
         "tsconfigPath": Any<String>,
         "watchPaths": Array [],
+        "serverEntryFile": undefined,
       }
     `
     );
--- a/packages/remix-dev/__tests__/readConfig-test.ts
+++ b/packages/remix-dev/__tests__/readConfig-test.ts
@@ -22,7 +22,7 @@ describe("readConfig", () => {
         assetsBuildDirectory: expect.any(String),
         relativeAssetsBuildDirectory: expect.any(String),
         tsconfigPath: expect.any(String),
-        serverEntryFile: expect.any(String),
+        serverEntryFile: undefined,
       },
       `
       Object {
@@ -44,18 +44,17 @@ describe("readConfig", () => {
             "path": "",
           },
         },
-        "serverBuildFile": undefined,
         "serverBuildPath": Any<String>,
         "serverBuildTarget": undefined,
         "serverBuildTargetEntryModule": "export * from \\"@remix-run/dev/server-build\\";",
         "serverDependenciesToBundle": Array [],
+        "serverEntryFile": undefined,
         "serverEntryPoint": undefined,
         "serverMode": "production",
         "serverModuleFormat": "cjs",
         "serverPlatform": "node",
         "tsconfigPath": Any<String>,
         "watchPaths": Array [],
-        "serverEntryFile": undefined,
       }
     `
     );
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -7,8 +7,7 @@ import ora from "ora";
 import prettyMs from "pretty-ms";
 import WebSocket from "ws";
 import type { Server } from "http";
-import type * as Express from "express";
-import type { createApp as createAppType } from "@remix-run/serve";
+import express, { type Express } from "express";
 import getPort, { makeRange } from "get-port";
 import * as esbuild from "esbuild";
 
@@ -202,7 +201,8 @@ export async function watch(
       ? remixRootOrConfig
       : await readConfig(remixRootOrConfig);
 
-  let wss = new WebSocket.Server({ port: config.devServerPort });
+  let { createSocketServer } = getServerEntry(config);
+  let wss = createSocketServer(config.devServerPort);
   function broadcast(event: { type: string; [key: string]: any }) {
     setTimeout(() => {
       wss.clients.forEach((client) => {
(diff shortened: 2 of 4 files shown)


kiliman:
Aside from the readConfig test, I haven't added or modified any tests. I'm not really sure the best approach to testing this as it requires verifying `remix dev`, `remix watch` and `remix-serve` is processing the server file correctly.

Please advise.

cliffordfajardo:
@remix-run/dev has a `config.ts` file which expose a [`server` property](https://github.com/remix-run/remix/blob/b1144d4b694b6853d6c9d9b8de3f13c0fdf9ebde/packages/remix-dev/config.ts#L130-L136), which one can use to define a server entry point
 frankly I haven't used the aforementioned `server` property but I did recall seeing @mcansh using it in here in this the [examples folder](https://github.com/mcansh/remix-fastify/blob/main/example/remix.config.js) of his `remix-fastify repo

@kiliman - is there an overlap with your new `serverEntryFile` config property and the [`server` config property](https://github.com/remix-run/remix/blob/b1144d4b694b6853d6c9d9b8de3f13c0fdf9ebde/packages/remix-dev/config.ts#L130-L136) defined here?

kiliman:
Yes there is a difference. The `server` option compiles your server file AND bundles it with your app build (including node_modules) into a single file.

The new option is just an extension of Remix App Server. I suppose I could have added an argument to `remix dev` instead.

cliffordfajardo:
@kiliman - do you have any recommendations on how to test this change out?
I'd love to provide feedback and trial out the API and report my findings back in this thread

kiliman:
@cliffordfajardo I created a sample showing how to use a custom server file for RAS. It uses SSL and includes `getLoadContext` function.

https://github.com/kiliman/remix-ras-server-example

jamesarosen:
It might be nice to show how consumers can use this to get the default `@remix-run/serve` server _plus_ extra middleware:
```ts
// server/index.ts
import { createApp } from '@remix-run/serve'
import express from 'express'
import path from 'path'
import myCustomMiddleware from './myCustomMiddleware'

const app = express()
app.use(myCustomerMiddleware)
app.use(createApp(path.join(process.cwd(), 'build'));
export default app
```
That way, if folks follow the docs, they'll still be as close to `@remix-run/serve` as possible. They can always customize further if they want.

kiliman:
@jamesarosen That's basically what this PR does. The issue is that the current Remix App Server doesn't expose the Express server, nor provide a way to override it.

This PR provides that extension point.

jamesarosen:
> That's basically what this PR does

I understood. I wasn't clear, but I was saying the docs should prefer importing the default server and extending it:
```ts
import { createApp } from '@remix-run/serve'
…
app.use(myCustomMiddleware)
app.use(createApp(…))
```
over replacing it altogether:
```ts
const { createRequestHandler } = require("@remix-run/express");
…
app.use(compression())
app.disable("x-powered-by")
app.use("/build", express.static("public/build", { immutable: true, maxAge: "1y" }))
app.use(express.static("public", { maxAge: "1h" }))
app.use(morgan("tiny"))
app.all(createRequestHandler(…))
```
That way, people who follow the docs will get the benefit of the default `createApp` with its caching, logging, and any new features the community adds over time.

kiliman:
Ah, I think I see. So you're saying that in _addition_ to my PR, make the default `createApp` an export so it can be used and extended in the *server.ts* file.

I agree. Sometimes you only want to make a simple change to the default, so may not want to write it from scratch (even if it is boilerplate).

My only concern is that Express app configuration is order dependent, so that my have unintended consequences, especially if the default is a black box.

jamesarosen:
> make the default `createApp` an export so it can be used and extended in the `server.ts` file.
Exactly!

> My only concern is that Express app configuration is order dependent, so that my have unintended consequences, especially if the default is a black box.

There are definitely pros and cons to both approaches. I don't think there's a _clear_ winner. I spent a lot of time in the Ember community, which places a lot of value on preferring the framework, but offering a series of escalating escape hatches. In this case, that series would be

1. use `@remix-run/serve`. A basic Remix-friendly HTTP server with TypeScript and live-reloading.
2. `serverEntryFile` with `createApp()`. Add features to the basic server. Opt in to updates from the community.
3. `serverEntryFile` with custom `createRequestHandler()`. You still get TypeScript and live-reloading, but opt out of updates from the community.
4. `server` (I think; I'm not actually sure why someone would use this if `serverEntryFile` exists)
5. replace `@remix-run/serve` with `node myServer.js`. Totally custom. No TypeScript or live-reloading unless you build it (e.g. with `tsc` and `nodemon`)

kiliman:
4. `server` is for hosts that don't allow you to run `npm i` for node_modules. This way you bundle everything and copy a single file to your host. Probably pretty rare considering how many providers there are.

danditomaso:
This change is exactly what I needed within Remix, I have lightly modified my express config but am now stuck using nodemon/tsc as described above. Using the framework in this case with an extended server.{t|j}s file would be such an benefit. Thanks for the hard work! Is there a timeline of when this would be merged into the core codebase?

MichaelDeBoey:
@kiliman I think we can close this as we now have `entryServerFilePath` config option since @mcansh's #4600?

mcansh:
> I think we can close this as we now have `entryServerFilePath` config option since @mcansh's #4600?

thats on the resolved config for entry.server

MichaelDeBoey:
Oh yeah I missed that it's not in the user's config, but the resolved config 🙈

kiliman:
Yeah, I haven't been keeping this updated with the latest, as so many changes were happening with the new dev server.

MichaelDeBoey:
Hi @kiliman!

kiliman:
Hmm.. not sure. I think the new `unstable_dev` feature obviates the need for some of the items in this PR. I personally don't use it. I'll close it for now. I'll reconsider if somebody wants me to update it to the latest version.

PR #4148, opened 2022-09-06
fix(remix-node): convert `fetch`'s `WebResponse` to `NodeResponse` so it matches the global

### LINKED ISSUE #4395: fetch() doesn't return correct Response type

1.7.2

### Steps to Reproduce

Inside the remix project try

```javascript
const a = await fetch('https://google.com')
console.log('a instanceof Response: ', a instanceof Response) // -> false
```

This probably has to do with this commit: 
https://github.com/remix-run/remix/commit/25b4c83b10c5bc4d84ea81407e7de1a0e4f43df4#diff-87ec1c3c130fe3481b9f1c542b1d8a53548882d33c73c68b894fd442d606afc0

A current workaround I found thanks to @milamer is to include 
```javascript
import { Response } from '@remix-run/web-fetch'
```
in the file.

### Expected Behavior

instanceof Response should be true

### Actual Behavior

instanceof Response is false

mcansh:
closes #4395, #3480


.changeset/late-worms-impress.md
```diff
@@ -0,0 +1,6 @@
+---
+"remix": patch
```
MichaelDeBoey, line 2: This one's not needed I think?

```suggestion
```

MichaelDeBoey (approved):

.changeset/late-worms-impress.md
mcansh: technically no its not needed, but we release everything in lock step currently 🤷

MichaelDeBoey: Yeah I know, that's why I approved it

Just wanted to let you know, as I thought you might have missed it somehow 😉

MichaelDeBoey:
@acoreyj It seems like this is doing the same thing as your implementation in #5095 or am I missing some things?

acoreyj:
> @acoreyj It seems like this is doing the same thing as your implementation in #5095 or am I missing some things?

@MichaelDeBoey Ah wow I apparently did a bad job searching PRs, o well now I know a lot more about fetch. 

Yes this is the same, and looking through the webfetch code it always resolves a `webResponse` and never rejects one so my extra checks and catch aren't necessary.

Plus the test strategy is definitely better

I think you are good to merge this and close mine.

cliffordfajardo:
@mcansh on this since thus PR seems this PR in the auth.js repo seems to be dependent on this PR getting merged 🙏 

https://github.com/nextauthjs/next-auth/pull/6270

MichaelDeBoey:
@mcansh I think this one can be merged as-is or is there anything that we're still waiting for?

MichaelDeBoey:
@mcansh I don't think anything is holding us back from merging this one?

blazinskip:
Any chances that this one would be merged in the near future 🙏🏻 . Would be great to be able to use `@auth/core` together with the remix.

packages/remix-node/fetch.ts
```diff
@@ -72,5 +72,7 @@ export const fetch: typeof webFetch = (
     ...init,
   };
 
-  return webFetch(info, init as RequestInit);
+  let webResponse = await webFetch(info, init as RequestInit);
```
kiliman, line 75: Doesn't this change the expectation of `fetch`, since now you `await` the response before returning? What if the original response is streaming?

jacob-ebey:
Should be closed by https://github.com/remix-run/remix/pull/7109

PR #4152, opened 2022-09-06
docs: Decision doc for Remix on top of Remix Router/RR 6.4

brophdawg11:
Documenting the plan for layering Remix on top of the new Remix Router and React Router 6.4

[Preview link](https://github.com/remix-run/remix/blob/brophdawg11/rrr-decision-doc/decisions/0007-remix-on-react-router-6-4-0.md)


decisions/0007-remix-on-react-router-6-4-0.md
```diff
@@ -0,0 +1,166 @@
...
+      3. See `differentiateCatchVersusErrorBoundaries` in the `brophdawg11/rrr` branch
+   3. New `RemixContext`
+      1. `manifest`, `routeModules`, `staticHandlerContext`, `serverHandoffString`
+      2. Create this alongside `EntryContext` assert the values match
+   4. If we catch an error during render, we'll have tracked the boundaries on `staticHandlerContext` and can use `getStaticContextFromError` to get a new context for the second pass (note the need to re-call `differentiateCatchVersusErrorBoundaries`)
+
+### Do the UI rendering layer second
+
+The rendering layer in `@remix-run/react` is a bit more of a whole-sale replacement and comes with backwards-compatibility concerns, so it makes sense to do second. However, we can still do this iteratively, we just can't deploy iteratively since the SSR and client HTML need to stay synced (and associated hooks need to ead crom the same contexts). First, we can focus on getting the SSR document rendered properly without `<Scripts/>`. Then second we'll add in client-side hydration.
```
MichaelDeBoey, line 142:
```suggestion
The rendering layer in `@remix-run/react` is a bit more of a whole-sale replacement and comes with backwards-compatibility concerns, so it makes sense to do second. However, we can still do this iteratively, we just can't deploy iteratively since the SSR and client HTML need to stay synced (and associated hooks need to read from the same contexts). First, we can focus on getting the SSR document rendered properly without `<Scripts/>`. Then second we'll add in client-side hydration.
```

MichaelDeBoey (changes requested):

kiliman:
I like the "strangler fig" pattern (first I've heard of it). However, in your pseudo implementation, you mention processing the request twice, once for the current flow and again for the new. My concern is that although it should be fine for GET requests, this plan may cause issues for POST as these requests are rarely idempotent.

Anyway, I'm really excited about the merge. Great job team for getting this far!

brophdawg11:
@kiliman yeah for sure.  My thought is that this "run both flows" flag would only ever be enabled during local development and maybe in a CI branch so we can ensure that all of our test suites pass on both flows.  There's (hopefully) nothing stateful in those test suites so I don't anticipate issues there 🤞.  We may even test on some local applications with the flag enabled - but obviously it would be with little test apps where double mutations wouldn't be a concern.  

I have already run into some issues where we assert that a loader was called once - and with the flag the test fails because it got called twice, so in some cases there may need to be some flag-driven assertions as well.

But certainly not envisioning ever releasing this flag in an "on" state.  Once we're comfortable, we can delete the current implementation and remove the flag, and then release with only the new implementation

Issue #3159, opened 2022-05-11
create-remix fails behind proxy


danilofes:
### What version of Remix are you using?

1.4.3

### Steps to Reproduce

Run `npx create-remix` when you are behind a corporate proxy.

### Expected Behavior

`create-remix` should work, either getting proxy config from environment variables or allowing users to pass it as a parameter.

### Actual Behavior

After answering the questions, the installation fails with the following message:

```
>npx create-remix
Need to install the following packages:
  create-remix
Ok to proceed? (y) y
? Where would you like to create your app? ./my-remix-app
? What type of app do you want to create? Just the basics
? Where do you want to deploy? Choose Remix if you're unsure; it's easy to change deployment targets. Remix App Server
? Do you want me to run `npm install`? No
| Creating your app…request to https://codeload.github.com/remix-run/remix/tar.gz/main failed, reason: connect ETIMEDOUT 20.201.28.149:443
```

I confirmed the problem is actually proxy config by modifying the function `downloadAndExtractTarball` in `node_modules\@remix-run\dev\cli\create.js`. I just added my proxy config hard-coded with the help of `https-proxy-agent` package. With the following change create-remix succeeded.

```js
var HttpsProxyAgent = require('https-proxy-agent');

async function downloadAndExtractTarball(projectDir, url, {
  token,
  filePath
}) {
  const proxyAgent = new HttpsProxyAgent('http://my.proxy.server:3128');
  let response = await fetch__default["default"](url, token ? {
    agent: proxyAgent,
    headers: {
      Authorization: `token ${token}`
    }
  } : {
    agent: proxyAgent
  });
  // rest of the method
}
```

rrgomide:
Please, guys. Fix this. It also happened to me. I think this is extremely important to be fixed since we're trying to incorporate Remix on enterprise-level projects. 

There are a lot of companies that use corporate proxies for security and we just can't disable it.

My solution was to scaffold a Remix project at home and bring all files to the company, which is really annoying.

kiliman:
Yuck. It's 2022. I can't believe we're still dealing with proxy issues like it's 1999.

Shouldn't the O/S be responsible for proxy config at this point? Why does every network-capable software need to deal with proxies?

machour:
> Please, guys. Fix this.

@rrgomide Feel free to open a PR to fix this. I don't think this is on the team priority list, having a PR will definitely help speed things up.

brendonco:
[...]

PR #4159, opened 2022-09-07
feat(create-remix): Add proxy support in create cli

freeman:
Closes: #3159 

- [ ] Docs
- [x] Tests

Testing Strategy:

I tested the built cli behind a proxy and it worked seamlessly.


freeman:
slightly simplified the code by having a single agent for the whole file

freeman:
Let me know if I should do something to get this merged.

This is adding friction to get remix adopted at work :)

mcansh:
thanks for taking this on @freeman! i've had this on my backlog for far too long 😅. looks good to me on first glance, but i'll give it a proper review when i'm back from vacation (Monday) if it's not merged by then

mcansh (approved):

freeman:
Working out the test failures

freeman:
Issues appears only in test with `msw` because `agent-base` has a pretty fragile logic for determining the agent protocol 

https://github.com/TooTallNate/node-agent-base/blob/c011b2b8efc8de6d57600c459d49baa4c608b86b/src/index.ts#L15-L19

This logic breaks under `msw` because it overrides the normal node request processing (for good reasons !).

I saw 2 paths to fix this:
- add some logic to ensure the agent protocol is set correctly based on the target url (even if in theory this is not needed outside of the test harness)
- find a way to bypass using the agent at all if under test harness

I went with the first option as some people noted that the logic of relying on stack trace could be broken in future node versions (https://github.com/TooTallNate/node-agent-base/issues/63).

freeman:
@mcansh could you approve the workflow run if you're ok with the changes to allow tests to pass ?

freeman:
@mcansh I'll take a look but this seems complex to test. I'd need to spin up a temporary proxy server (this seems doable, there is a `proxy` package) and try to ensure it has been used without breaking the encapsulation...

freeman:
@mcansh added a basic test that verifies that proxy env variable is honored.

PR #4170, opened 2022-09-09
docs(examples/usematches): Fix incorrect link in README

Deanmv:
This change fixes the link from Readme to the `useUser` file in the example project


machour (approved):
Thank you!

PR #4198, opened 2022-09-15
feat(templates/cloudflare-workers): update to ESM

jacob-ebey:
Closes: #764

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/templates/cloudflare-workers/README.md
+++ b/templates/cloudflare-workers/README.md
@@ -6,7 +6,7 @@
 
 You will be running two processes during development:
 
-- The Miniflare server (miniflare is a local environment for Cloudflare Workers)
+- The Wrangler (wrangler is a local environment for Cloudflare Workers)
 - The Remix development server
 
 Both are started with one command:
@@ -33,5 +33,42 @@ If you don't already have an account, then [create a cloudflare account here](ht
 Once that's done, you should be able to deploy your app:
 
 ```sh
-npm run deploy
+npx wrangler publish
 ```
+
+### Via GitHub Actions
+
+If you want to deploy automatically on change to the main branch, you can create a `.github/workflows/deploy.yml` with the following contents:
+
+```yml
+name: 🕊 Deploy
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    name: 🕊 Deploy
+    steps:
+      - name: 🛑 Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.9.1
+
+      - name: ⬇️ Checkout repo
+        uses: actions/checkout@v3
+
+      - name: 📥 Install deps
+        uses: bahmutov/npm-install@v1
+
+      - name: 📦 Build
+        run: npm run build
+
+      - name: 🚀 Publish
+        uses: cloudflare/wrangler-action@2.0.0
+        with:
+          apiToken: ${{ secrets.CF_API_TOKEN }}
+          command: publish
+```
+
+You will also need to add an actions secrets called `CF_API_TOKEN` that can be retrieved from https://dash.cloudflare.com/profile/api-tokens.
--- a/templates/cloudflare-workers/app/entry.server.tsx
+++ b/templates/cloudflare-workers/app/entry.server.tsx
@@ -1,4 +1,4 @@
-import type { EntryContext } from "@remix-run/cloudflare";
+import { type EntryContext } from "@remix-run/cloudflare";
 import { RemixServer } from "@remix-run/react";
 import { renderToString } from "react-dom/server";
 
--- a/templates/cloudflare-workers/package.json
+++ b/templates/cloudflare-workers/package.json
@@ -3,31 +3,30 @@
   "sideEffects": false,
   "scripts": {
     "build": "remix build",
-    "deploy": "wrangler publish",
-    "dev:remix": "remix watch",
-    "dev:miniflare": "cross-env NODE_ENV=development miniflare ./build/index.js --watch",
-    "dev": "remix build && run-p \"dev:*\"",
-    "start": "cross-env NODE_ENV=production miniflare ./build/index.js"
+    "dev": "npm run build && concurrently \"npm:dev:remix\" \"npm:dev:cloudflare\"",
+    "dev:remix": "cross-env NODE_ENV=development remix watch",
+    "dev:cloudflare": "cross-env NODE_ENV=development wrangler dev --local --config wrangler.dev.toml",
+    "start": "npm run dev:cloudflare"
   },
   "dependencies": {
+    "@cloudflare/kv-asset-handler": "^0.2.0",
     "@remix-run/cloudflare": "*",
-    "@remix-run/cloudflare-workers": "*",
     "@remix-run/react": "*",
-    "cross-env": "^7.0.3",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
   },
   "devDependencies": {
     "@cloudflare/workers-types": "^3.14.1",
     "@remix-run/dev": "*",
     "@remix-run/eslint-config": "*",
-    "@types/react": "^17.0.47",
-    "@types/react-dom": "^17.0.17",
+    "@types/react": "^18.0.20",
+    "@types/react-dom": "^18.0.6",
+    "concurrently": "^7.4.0",
+    "cross-env": "^7.0.3",
     "eslint": "^8.20.0",
     "miniflare": "^2.6.0",
-    "npm-run-all": "^4.1.5",
-    "typescript": "^4.7.4",
-    "wrangler": "^2.0.22"
+    "typescript": "^4.8.3",
+    "wrangler": "2.1.3"
   },
   "engines": {
     "node": ">=16.13"
--- a/templates/cloudflare-workers/remix.config.js
+++ b/templates/cloudflare-workers/remix.config.js
@@ -2,7 +2,6 @@
 module.exports = {
   serverModuleFormat: "esm",
   serverBuildPath: "build/index.mjs",
-  server: "worker/index.ts",
   devServerBroadcastDelay: 1000,
   devServerPort: 8002,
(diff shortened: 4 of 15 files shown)


templates/cloudflare-workers/README.md
```diff
@@ -33,5 +33,42 @@ If you don't already have an account, then [create a cloudflare account here](ht
...
+    branches:
+      - main
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    name: 🕊 Deploy
+    steps:
+      - name: 🛑 Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.9.1
```
MichaelDeBoey, line 56: Latest version is `0.10.0`

```suggestion
        uses: styfle/cancel-workflow-action@0.10.0
```

templates/cloudflare-workers/app/entry.server.tsx
```diff
@@ -1,4 +1,4 @@
-import type { EntryContext } from "@remix-run/cloudflare";
+import { type EntryContext } from "@remix-run/cloudflare";
```
MichaelDeBoey, line 1: Let's keep the type at the beginning of the import to keep it consistent with the rest of the codebase

```suggestion
import type { EntryContext } from "@remix-run/cloudflare";
```

templates/cloudflare-workers/types/remix-build.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module "remix-build" {
+  import { type ServerBuild } from "@remix-run/cloudflare";
```
MichaelDeBoey, line 2:
```suggestion
  import type { ServerBuild } from "@remix-run/cloudflare";
```

templates/cloudflare-workers/package.json
```diff
@@ -3,31 +3,30 @@
   "sideEffects": false,
   "scripts": {
     "build": "remix build",
-    "deploy": "wrangler publish",
-    "dev:remix": "remix watch",
-    "dev:miniflare": "cross-env NODE_ENV=development miniflare ./build/index.js --watch",
-    "dev": "remix build && run-p \"dev:*\"",
-    "start": "cross-env NODE_ENV=production miniflare ./build/index.js"
+    "dev": "npm run build && concurrently \"npm:dev:remix\" \"npm:dev:cloudflare\"",
```
MichaelDeBoey, line 6: We're using `npm-run-all` in all other templates/examples, so maybe we should use that here as well to keep it consistent?

templates/cloudflare-workers/package.json
```diff
@@ -3,31 +3,30 @@
...
+    "dev:remix": "cross-env NODE_ENV=development remix watch",
+    "dev:cloudflare": "cross-env NODE_ENV=development wrangler dev --local --config wrangler.dev.toml",
+    "start": "npm run dev:cloudflare"
   },
   "dependencies": {
+    "@cloudflare/kv-asset-handler": "^0.2.0",
     "@remix-run/cloudflare": "*",
-    "@remix-run/cloudflare-workers": "*",
```
MichaelDeBoey, line 14: Is it intentional that we're not downloading `@remix-run/cloudflare-workers` by default anymore?

templates/cloudflare-workers/package.json
```diff
@@ -3,31 +3,30 @@
...
     "@remix-run/react": "*",
-    "cross-env": "^7.0.3",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
```
MichaelDeBoey, line 16: Does this mean, we can update the `cloudflare-pages` template to React 18 as well?

templates/cloudflare-workers/remix.config.js
```diff
@@ -1,11 +1,8 @@
...
-  server: "./server.js",
+  serverModuleFormat: "esm",
+  serverBuildPath: "build/index.mjs",
   devServerBroadcastDelay: 1000,
+  devServerPort: 8002,
   ignoredRouteFiles: ["**/.*"],
-  // appDirectory: "app",
-  // assetsBuildDirectory: "public/build",
-  // serverBuildPath: "build/index.js",
-  // publicPath: "/build/",
```
MichaelDeBoey, line 10: We keep this commented out for convenience in our other templates, so maybe we should keep them here as well?

templates/cloudflare-workers/remix.init/index.js
```diff
@@ -0,0 +1,10 @@
+const fs = require("fs");
+const path = require("path");
+
+function main({ rootDirectory }) {
+  const examplePath = path.resolve(rootDirectory, "wrangler.example.toml");
```
MichaelDeBoey, line 5: Is it intentional to keep this file in the root?

We keep `remix.init` files in the same folder in our stacks.

templates/cloudflare-workers/remix.init/package.json
```diff
@@ -0,0 +1,5 @@
+{
```
MichaelDeBoey, line 1: We can just delete this file if we don't have any `dependencies`

templates/cloudflare-workers/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "include": ["remix.env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["**/*.d.ts", "**/*.ts", "**/*.tsx"],
```
MichaelDeBoey, line 2: @mcansh Does this have implications for the `convert-to-javascript` migration?

templates/cloudflare-workers/types/remix.env.d.ts
```diff
@@ -1,3 +1,2 @@
 /// <reference types="@remix-run/dev" />
```
MichaelDeBoey, line 1: Should we move this file to a `types` folder as well in our other templates, so this is consistent?

MichaelDeBoey (changes requested):
@jacob-ebey Do we need these updates for the `cloudflare-pages` template as well?

huw:
This adds a lot of boilerplate. Has the Remix team considered updating `createEventHandler` or adding a new exported function to `@remix-run/cloudflare-workers`?

MichaelDeBoey:
@huw Do you mean something like @ekosz is doing in #1207?

huw:
Yeah, that seems closer to what I’d expect & it feels (very subjectively) more idiomatic.

The reason this PR removes the dependency on `@remix-run/cloudflare-workers` is because it completely supersedes it by inlining the functionality into the template. It would require a few changes to the docs to explain that `@remix-run/cloudflare-workers` is only appropriate for the old service-worker syntax, and that ESM users should copy from this template. My guess is that this isn’t what the Remix team want, unless you’re planning to move away from adapters in the long run?

I suspect it would be easier for end users to understand & upgrade if they only had to make a small change to their `server.js` file—@ekosz’s example is simply:

```js
 import { createEventHandler } from "@remix-run/cloudflare-workers";

 import * as build from "../build";

 const handler = createEventHandler({ build })

 export default {
   fetch(request, env, context) {
     const event = { request, env, waitUntil: context.waitUntil.bind(waitUntil) };
     return handler(event);
   }
 }
 ```

dario-piotrowicz:
I just wanted to mention that I just pulled down this branch and checked locally on a wrangler project and this PR does seem to work in its current state (compared to #1207 which has a few bits not fully working)

@MichaelDeBoey @jacob-ebey could we get this PR merged soon even if some small things can be improved and iterate over those details afterwards? (so that devs using Cloudflare workers can at least start using the ESM version sooner)

huw:
@dario-piotrowicz FWIW this is just a change to the Workers template, not any of Remix's actual code. Remix is fully compatible with Workers today, users just need to copy the code in this template into their setups.

However, that requires a pretty specific setup. Users need to be wary of where their `build` is and update the reference in code if they reconfigure Remix, for one. And they have to use Wrangler's built-in bundler, or rebundle the Remix output themselves (since this code does not bundle dependencies into the final server code as-is). It's also using a deprecated config flag in `remix.config.js` to hack around the Remix compiler.

IMHO, as I said above, this is a very brittle approach to solving this problem & doesn't feel in line with the high standards both Remix & Cloudflare tend to aspire to for developer experience. And it would leave significant fragmentation in the Remix DX for users, because the instructions for integrating with ESM vs CJS would end up being very different--much more different than changing between Pages & Workers Sites, which is as simple as updating `remix.config.js` and a handful of lines in the server index file.

Again, the preferable solution would be to update the `@remix-run/cloudflare-workers` package. I gave this a crack today, but because the KV asset handler requires that `__STATIC_CONTENT_MANIFEST` import, getting interop with CJS is much trickier. I think the right approach might be to:

1. Create a new package (`@remix-run/cloudflare-workers-esm` or just `cloudflare-esm`) with ESM-specific code (i.e. tweaked #1207 code).
2. Update the Remix compiler to compile a new `cloudflare-esm` target as ESM with the `neutral` platform, while still bundling browser code.
3. Update the docs.

I really want to stress how unusable this PR would be in its current state & how much additional support load it will create for Cloudflare & Remix. I'm in a busy period rn but would gladly do this properly once I'm through that.

dario-piotrowicz:
@huw thanks so very much for the nice clarification, yeah I understand that it is only template changes, but this is not at all straightforward to get done without checking out this PR, that's why I was asking to include it so to reduce the barrier to entry (maybe it could be added in a README, docs or something?)

Anyways I had no idea that this was brittle/using a deprecated flag, I thought that the only downside of this solution was the amount of boilerplate needed. Based on what you said I agree that adding something half-baked in at this stage would make things worse rather than better....

Thanks for taking the time to look at this, I'll patiently wait for you to have more time to look into this then 🙂

Also, I don't have much knowledge on how Remix works under the hook, but if you think I could be of any help here please let me know 🙏

MichaelDeBoey:
@jacob-ebey @huw @dario-piotrowicz I'm wondering if you could obtain the same result with the new config properties that were exposed by #4841 🤔

MichaelDeBoey:
@jacob-evey Closing this as @pcattori  updated the `cloudflare-workers` template to ESM in #6650

PR #4214, opened 2022-09-16
feat: add `remix-eslint-flat-config` package

mikeybinns:
This PR is created after I created this discussion: #4201
I ended up doing the work anyway to learn about the new ESLint config standard. I accept that this might just be discarded as this work wasn't approved before the work was completed.

Tasks completed:
- [x] Convert the old require config to modern import exports
- [x] Document the new import config strategy in README
- [x] Convert env keys to imported globals package
- [x] Move ecmaVersion and sourceType to languageOptions
- [x] Move and modify parser and parserOptions to languageOptions
- [x] Convert plugin imports to new standard
- [x] Remove all extends references in favour of new ESLint flat merge
- [x] Check for `--rulesdir` usage and refactor (none found)
- [x] Check for use of `noInlineConfig` and `reportUnusedDisableDirectives` (none found)
- [x] Remove deprecated `jest.js` file
- [x] Remove multiple config files in favour of exporting separate configs from a single source
- [ ] Test new config against old config to ensure it generates the same errors (HELP WANTED)
- [ ] Confirm package.json contents with Remix team to ensure it matches package strategy

Testing Strategy:
Not yet tested, work in progress.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -284,6 +284,7 @@
 - michaeldeboey
 - michaelfriedman
 - michaseel
+- mikeybinns
 - mikeybinnswebdesign
 - mirzafaizan
 - mjackson
--- a/packages/remix-eslint-flat-config/README.md
+++ b/packages/remix-eslint-flat-config/README.md
@@ -1,0 +1,52 @@
+# `@remix-run/eslint-config`
+
+This package includes a shareable ESLint config for Remix projects.
+
+If you create your app with `create-remix` no additional configuration is necessary.
+
+## Installation
+
+First, install this package along with ESLint in your project. **This package requires at least version 8.1 of ESLint**
+
+```sh
+npm install -D eslint @remix-run/eslint-config
+```
+
+Then create a file named `eslint.config.js` in the root of your project:
+
+```js
+import { coreConfig } from "@remix-run/eslint-flat-config";
+
+export default [...coreConfig];
+```
+
+### Additional configs
+
+This packages also ships with optional configuration objects for projects that use the following:
+
+- [Node](https://nodejs.org/en/). To use this config, import `nodeConfig`.
+- [Jest](https://jestjs.io/) with [Testing Library](https://testing-library.com). To use this config, import `testingLibraryConfig`.
+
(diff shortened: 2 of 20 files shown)


mikeybinns:
Just to note down my thought process for the system for posterity:

Each file under "Rules" exports 1 or more named JS objects to be imported into other packages. These files are the only files which should define rules, and therefore the only files that should have the following code pattern:
```js
const OFF = 0;
const WARN = 1;
const ERROR = 2;
```
The only exception is the packageExports file, which doesn't export rules, but helpers for the core rules.

Each file under "Settings" exports a named settings object.

Each file under "configs" exports a named configuration file. Each of these files should export a valid ESLint config array on their own, so each could be used independently of each other if that's what the user wants to do. Even if there is only one object, this should still export an array so the config variable can be spread correctly into another config array, we don't want to mix arrays and objects as this will lead to unexpected errors and confusion.

The `index.js` file provides a single export source so the user can export multiple configs in a single import command. This is not exporting a default export config like the original config due to how the new system handles merging configs.

The `internal.js` file exports a named config only for internal use, so this doesn't necessarily follow the above structure/process to keep all these rules in a single file. You still need to import the config into an `eslint.config.js` file as with the normal rules.

MichaelDeBoey (commented):
I'll leave this up to the team, but I don't think we currently want to do this.

If you want to use a flat config yourself, you should still be able to use existing configs/plugins.
That's one of the goals of the flat config: having backwards compatibility

chaance:
@mikeybinns Appreciate the effort here, but I'd probably hold on additional work at the moment. The team hasn't had an opportunity to review the proposal, and we've got a few other priorities to focus on.

If you prefer this style, you can always publish this as a community package yourself and promote it in our Discord. Once the changes in ESLint are a bit more stable and settled we can revisit this on our end directly in the existing `@remix-run/eslint-config` package.

mikeybinns:
No problem, I knew this was a possibility and honestly it wasn't too much work putting this together, so not much was lost, and I got to learn about the new format in the process.

If you want to use the PR as a jumping off point later down the line, that's completely fine with me, save you having to start from scratch. 🙂

PR #4219, opened 2022-09-18
feat(remix-react): preload critical style-sheets

sanjaiyan-dev:
This is inspired from `Next.js`.
[Preload critical stylesheets to improve loading speed](https://web.dev/preload-critical-assets/).

_Sorry if I made any mistakes :(_

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -364,6 +364,7 @@
 - ryanflorence
 - ryankshaw
 - sandulat
+- sanjaiyan-dev
 - sbernheim4
 - schpet
 - scottybrown
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -682,7 +682,10 @@ function PrefetchPageLinksImpl({
       {styleLinks.map((link) => (
         // these don't spread `linkProps` because they are full link descriptors
         // already with their own props
-        <link key={link.href} {...link} />
+        <React.Fragment key={link.href}>
+          <link rel="preload" href={link?.href} as="style" />
+          <link {...link} />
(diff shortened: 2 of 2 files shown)


machour (changes requested):
Hi @sanjaiyan-dev 

Please target the `dev` branch instead of `main`. 
We also need tests, and an explanation of what you're doing.

sanjaiyan-dev:
> Hi @sanjaiyan-dev 
> 
> Please target the `dev` branch instead of `main`. 
> We also need tests, and an explanation of what you're doing. 

Ok,

Sorry, actually I thought to preload the critical style sheets as recommended by Google.

jacob-ebey:
This actually won't really do anything. `<link>` tags are parsed out of the document and initialized as soon as they are seen anyways. This is a non-blocking step of the document parse, so adding a preload at the literal same location as the link doesn't do anything for ya.

chaance:
@sanjaiyan-dev Thanks for the PR! I'll echo what @jacob-ebey said and add one additional point:

Our design is intentionally different from Next in a lot of ways. Our `links` API provides you with a simple interface for declaring *exactly* which link tags you want to inject on a given route. **We don't inject any stylesheet links on your page that you don't explicitly tell us to.** Per our philosophy page, we try not to over-abstract the underlying web API, so if users want to add additional link tags as you've done, they can do so:

```ts
export function links() {
  return [
    { rel: "preload", href: stylesheetUrl, as: "style"  },
    { rel: "stylesheet", href: stylesheetUrl  },
  ];
}
```

Hope that makes sense!

PR #4232, opened 2022-09-19
fix(remix-react): `FormMethod`'s `type` can only be `'get'` or `'post'`

hollandThomas:
Fixes the `FormMethod` type to only allow `'get' | 'post'` to reflect both the HTML spec and the actual implementation.

Since the `<FormImpl />` component falls back to `'post'` as soon as `method` is anything other than `'get'`
```typescript
 let formMethod: FormMethod =
    method.toLowerCase() === "get" ? "get" : "post";
```
 _and_ since [the spec only allows get or post](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method) (or dialog) anyway, it's best to adapt the `FormMethod` type accordingly.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -809,11 +809,7 @@ function dedupe(array: any[]) {
 
 export interface FormProps extends FormHTMLAttributes<HTMLFormElement> {
   /**
-   * The HTTP verb to use when the form is submit. Supports "get", "post",
-   * "put", "delete", "patch".
-   *
-   * Note: If JavaScript is disabled, you'll need to implement your own "method
-   * override" to support more than just GET and POST.
+   * The HTTP verb to use when the form is submitted. Supports "get" and "post".   
    */
   method?: FormMethod;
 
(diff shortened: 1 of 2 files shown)


rossipedia:
I'm pretty sure this isn't a bug. It's intentional that Remix form's support additional methods when JS is enabled. You'll notice the original, user-supplied `method` gets passed along unaltered here:

https://github.com/remix-run/remix/blob/473330f90399fd0dc8bc81a70a12c408437df90f/packages/remix-react/components.tsx#L986

From [the Docs](https://remix.run/docs/en/v1/api/remix#form-method) (emphasis mine):

> This determines the [HTTP verb](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) to be used: get, post, put, patch, delete. The default is "get".
>
>     <Form method="post" />
>
> Native `<form>` only supports get and post, so if you want your form to work with JavaScript on or off the page you'll need to stick with those two.
>
> Without JavaScript, Remix will turn non-get requests into "post", but you'll still need to instruct your server with a hidden input like `<input type="hidden" name="_method" value="delete" />`. **If you always include JavaScript, you don't need to worry about this**.

PR #4245, opened 2022-09-20
feat(remix-server-runtime): RRR 1.1 - handleResourceRequest

jacob-ebey:
[decision doc](https://github-md.com/remix-run/remix/blob/brophdawg11/rrr-decision-doc/decisions/0007-remix-on-react-router-6-4-0.md#decision)

Part 1 of step 1.1 in layering Remix back on top of React Router now that we have released `@remix-run/router`.

Changes:

- Introduce build time transform for server-runtime that replaces `process.env.EXPERIMENTAL_BUILD`
- more aggressive tree-shaking for server-runtime to allow unused deps to be optimized away

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:


packages/remix-server-runtime/server.ts
```diff
@@ -1,3 +1,7 @@
+/// <reference lib="dom.iterable" />
```
jacob-ebey, line 1: I do not like this and will explore more tomorrow. I think this is a byproduct of how we have configured our root tsconfig.

packages/remix-server-runtime/server.ts
```diff
@@ -509,6 +557,41 @@ async function handleDocumentRequest({
...
 }
 
+async function handleResourceRequestRR(
+  serverMode: ServerMode,
+  staticHandler: StaticHandler,
+  routeId: string,
+  request: Request
+) {
+  try {
+    let response = await staticHandler.queryRoute(request, routeId);
```
brophdawg11, line 567: At the moment, we need to build the `dev` branch of `@remix-run/router` into the `remix/node_modules` directory for this to work, so we can pick up the changes in https://github.com/remix-run/react-router/pull/9353.

To do this (assumes your `remix` and `react-router` repos are sibling directories):

```bash
> cd remix
> yarn
> cd ../react-router/packages/router
> LOCAL_BUILD_DIRECTORY=../../../remix npx rollup -c
> cd ../../../remix
> ENABLE_REACT_ROUTER=1 yarn test:integration resource-route
```

brophdawg11: @jacob-ebey Curious if you have ideas on how to smooth this out.  The `dev` changes will get picked up in the next `react-router` release which I figure we'll do ~weekly or so. But in-between I'd hate to hold up this RRR work. I wonder if changesets would let us do a prerelease for a single package, or maybe we do router-only pre-releases with changes to static handler stuff

packages/remix-server-runtime/server.ts
```diff
@@ -74,24 +75,18 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
           createStaticHandlerDataRoutes(build.routes, loadContext)
         );
 
-        let [response, remixRouterResponse] = await Promise.all([
```
brophdawg11: Let's add an integration test here to post data to a resource route action to prove the bug, and then we can go back to the `Promise.all`.

packages/remix-server-runtime/server.ts
brophdawg11: Copied files into `./router` for the duration of the partial implementation

packages/remix-server-runtime/server.ts
brophdawg11: We have other tests outside of resource routes that fail

PR #4246, opened 2022-09-21
fix(@remix-run/react): remove unused `type-fest` dependency

pcattori:
should have been removed when serialize type helpers were moved into server-runtime pkg


jacob-ebey (approved):

MichaelDeBoey (approved):
`type-fest` should be added to `@remix-run/server-runtime`'s `dependencies`

This can be done in a separate PR if wanted

PR #4269, opened 2022-09-23
chore: update tsconfigs to use composite

jacob-ebey:
Speeds up build a bit by allowing tsc to quickly find references: https://www.typescriptlang.org/docs/handbook/project-references.html#composite

Allows projects to build in more isolation and avoids wired x-project types to leak. For example I have a PR open right now that does not recognize `headers.entries()` as a method in the server-runtime, but with these changes it works fine.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /packages/*/LICENSE.md
 node_modules/
 yarn-error.log
+tsconfig.tsbuildinfo
 
 /jacob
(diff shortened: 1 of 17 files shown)


MichaelDeBoey (approved):

pcattori (approved):
LGTM

Surprised things were working at all before we added `composite: true`!

PR #4271, opened 2022-09-25
Add .env line to blog tutorial

someshkar:
When I was trying out the blog tutorial for the first time, I got stuck at the step where I followed every step very cautiously but I ran into an error where no database URL was defined in the prisma schema, which gets it from an env variable called `DATABASE_URL`.

This was an easy fix (copying `.env.example` to `.env`), but would definitely hinder first time JavaScript users in what is supposed to be a begginer friendly tutorial.

This is what the error looks like:

image

While this is an easy fix with a bit of googling, it might be fairly intimidating to the target audience.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -393,6 +393,7 @@
 - SufianBabri
 - supachaidev
 - Synvox
+- someshkar
 - tagraves
 - tascord
 - TheRealAstoo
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
It seems like the `remix.init` didn't run (this mostly is because you didn't install all dependencies).

If you run `npx remix init`, the `.env` file should be created.

someshkar:
I see, thanks!

PR #4284, opened 2022-09-27
fix(remix-eslint-config): update `internal` ESLint config

MichaelDeBoey:
Follow-up of https://github.com/remix-run/examples/pull/14

### DIFF
--- a/packages/remix-eslint-config/package.json
+++ b/packages/remix-eslint-config/package.json
@@ -43,14 +43,13 @@
   "devDependencies": {
     "@types/eslint": "^8.4.6",
     "eslint": "^8.23.1",
+    "jest": "^27.5.1",
     "react": "^18.2.0",
-    "react-dom": "^18.2.0",
     "typescript": "^4.7.4"
   },
   "peerDependencies": {
     "eslint": "^8.0.0",
     "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
     "typescript": "^4.0.0"
   },
   "peerDependenciesMeta": {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,9 +19,8 @@ module.exports = {
     {
       files: ["templates/**/*.*"],
       rules: {
-        "prefer-const": WARN,
-
         "prefer-let/prefer-let": OFF,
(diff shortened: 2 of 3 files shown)


packages/remix-eslint-config/package.json
```diff
@@ -43,17 +43,20 @@
...
     "@types/eslint": "^8.4.6",
     "eslint": "^8.23.1",
+    "jest": "^27.5.1",
     "react": "^18.2.0",
-    "react-dom": "^18.2.0",
     "typescript": "^4.7.4"
   },
   "peerDependencies": {
     "eslint": "^8.0.0",
+    "jest": "^27.0.0 || ^28.0.0 || ^29.0.0",
```
chaance, line 52:
```suggestion
```

I don't think we should add this, even if marked as optional. You can use Jest's rules without Jest if your test runner share's Jest's API (I use them with projects that use Vitest).

packages/remix-eslint-config/package.json
```diff
@@ -43,17 +43,20 @@
...
     "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
     "typescript": "^4.0.0"
   },
   "peerDependenciesMeta": {
+    "jest": {
+      "optional": true
+    },
```
chaance, line 59:
```suggestion
```

chaance (changes requested):

packages/remix-eslint-config/package.json
MichaelDeBoey: I added this because `jest-testing-library` config didn't work without `jest`
Didn't test with `vitest` though.

Since the config has `jest` in it's name, I think we should at least add it to `peerDependencies` imo.
What do you think about adding `vitest` (& making it optional) as well?

chaance: I still say remove it. Many of our stacks use Vitest and these rules should work well since they share most of the same API.

chaance: I still say remove it. Many of our stacks use Vitest and these rules should work well since they share most of the same API.

chaance (changes requested):

chaance (approved):

PR #4285, opened 2022-09-27
ci: use GitHub Actions account to commit format changes

MichaelDeBoey:
Just like we do in https://github.com/remix-run/examples

https://github.com/remix-run/examples/blob/8c122f535a345c35400e06b77de0628b53342200/.github/workflows/format.yml#L42-L43


mcansh (approved):
nice!

Issue #231, opened 2021-08-02
How to set multiple `Set-Cookie` headers in a single response?


kentcdodds:
I tried copying `concatSetCookieHeaders` from remix but that... eh, doesn't work... Doing that I end up with a single set-cookie header.

Here's the relevant code: 

```ts
const sessionIdCookie = await getUserSessionCookieFromMagicLink(request)
if (sessionIdCookie) {
  const destroyCookie = await loginInfoSession.destroy()
  return redirect('/me', {
    headers: {
      'Set-Cookie': concatSetCookieHeaders(destroyCookie, sessionIdCookie),
    },
  })
}
```

What I need is two `Set-Cookie` headers, and this gives me a single `Set-Cookie` header that doesn't end up updating either one.

If I use either of these cookie values by itself then they each work as expected. But I need to set them both in the same response.

shripadk:
This doesn't seem to be working when you have a parent loader and nested loader both trying to `Set-Cookie`. Since according to Remix docs, the nested loader headers override the parent loader headers, it can only be merged in the `headers` function that is exported. However, there is no way to merge both `Set-Cookie` headers in there either as the result expected by the function is a Javascript object or instance of `Headers` Fetch API.

How do I handle this situation? Or am I missing something?

shripadk:
@kiliman Wasn't sure if this is a bug or a discussion so I posted both. A discussion in Discord as well as a report here. But I am now inclined to think this is definitely a bug.

Tested it out in the root loader (the Response headers contains only 1 `Set-Cookie` that is concatenating all values instead of setting a `Set-Cookie` for each cookie):

<img width="440" alt="image" src="https://user-images.githubusercontent.com/189344/185213886-9f6403af-b350-45d3-b097-95ae080f0cd6.png">

Network tab:

<img width="937" alt="image" src="https://user-images.githubusercontent.com/189344/185214185-cf1ce502-86fa-4c7a-8c7a-c39618b2fcfb.png">

shripadk:
@kiliman  Interestingly it only works with `redirect`. This only tells me that this is definitely a bug as the user of the framework would expect same consistent behavior with respect to headers in both situations (when using `json` and when using `redirect`) as well as within the exported `header` function. I feel this issue should be re-opened as it only tackles one use case. You can see the behavior of `redirect` (which is the right behavior) and contrast it with behavior of `json` above:

[...]

PR #4286, opened 2022-09-27
bug: Cookies incorrectly set via return json on LoaderFunction when also set on root LoaderFunction

davidmytton:
### Remix version

1.7.2

### Description

A CSRF token is generated in the root `LoaderFunction` and then set on the default app context in `root.tsx`. This is used in a form input field and checked as part of the `ActionFunction`.  The `ActionFunction` also sets a `flash` session value which is then read by the `LoaderFunction`. The call to `json()` in the `LoaderFunction` returns the `Set-Cookie` headers to remove the flash session value ([as per the Remix docs](https://remix.run/docs/en/v1/api/remix#sessionflashkey-value)).

This appears to be related to https://github.com/remix-run/remix/issues/231#issuecomment-1218371334

An example project with this reproduced is also at https://github.com/davidmytton/remix-cookie-bug-example

### Expected result

Every time the form is submitted, the CSRF token is validated and the form submit happens correctly.

### What actually happens

The first form submit succeeds, but the regenerated CSRF token is not correctly stored in the session so a subsequent submit fails with a CSRF error.

This is "fixed" by changing the `LoaderFunction` return from:

```typescript
return json(
    { message },
    {
      headers: {
        "Set-Cookie": await commitSession(session),
      },
    }
  );
```

to remove the header:

```typescript
return json({ message });
```

However, this means the flash session value doesn't get removed.

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -108,7 +108,6 @@ test.beforeAll(async () => {
             // a Cookie from createCookie or the CookieOptions to create one
             cookie: {
               name: "__session",
-
               httpOnly: true,
               maxAge: 60,
               path: "/",
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -188,10 +188,9 @@ test.beforeAll(async () => {
           const session = await getSession(request.headers.get("Cookie"));
 
           const message = session.get("globalMessage") || null;
-          const csrf = session.get("csrf") || null;
 
           return json(
-            { message, csrf },
+            { message },
             {
               headers: {
                 "Set-Cookie": await commitSession(session),
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -47,28 +47,214 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/index.jsx": js`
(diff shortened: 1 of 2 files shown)


integration/bug-report-test.ts
```diff
@@ -47,28 +47,213 @@ test.beforeAll(async () => {
...
-          let data = useLoaderData();
+      "app/sessions.ts": js`
+        import { createCookieSessionStorage } from "@remix-run/node"; // or cloudflare/deno
+
+        const { getSession, commitSession, destroySession } =
+          createCookieSessionStorage({
+            // a Cookie from createCookie or the CookieOptions to create one
+            cookie: {
+              name: "__session",
+
```
machour, line 111:
```suggestion
```

brophdawg11:
This is a use-case that middleware will eventually handle, but I believe the underlying issue is the creation of 2 session instances in 2 parallel loaders.  If you're using an express server, you may be able to create the session in `getLoadContext` and pass a single instance in that way.

Otherwise, for now, I think the best solution is to manage the session in one loader to avoid parallel session instance creations.  You can lift the session message reading up into the root loader as well and then access it from the index component via `let { message } = useRouteLoaderData("root");`.

brophdawg11:
Closing this out as a known limitation of Remix v1.  If you need to use different values in different parallel loaders, using 2 sessions is probably the way to go. Otherwise make the changes in one loader.  We're considering how to make this better in the future.

PR #4287, opened 2022-09-28
fix(templates/deno): fix type re-export from `@remix-run/server-runtime`

wangbinyq:
fix(remix-deno): deno fix type reexport from server-runtime not working

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -421,6 +421,7 @@
 - vlindhol
 - vmosyaykin
 - vorcigernix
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (commented):
@wangbinyq can you tell where you see this error?

Import directly from `@remix-run/server-runtime` is a bad practice, you should import from `@remix-run/deno` instead

MichaelDeBoey (approved):

Issue #1555, opened 2022-01-19
[Bug]: JavaScript heap out of memory


JClackett:
### Which Remix packages are impacted?

- `remix` (Remix core)
- `@remix-run/dev`
- `@remix-run/node`
- `@remix-run/react`
- `@remix-run/serve`
- `@remix-run/vercel`

### What version of Remix are you using?

v1.1.1

### What version of Node are you using? Minimum supported version is 14.

v16.13.1

### Steps to Reproduce

It's just a regular remix app and sometimes when saving something in the codebase the server crashes and gives me the error pasted below, relating to JS memory issues. Looks like something to do with esbuild but I'm not really sure, just wondering if anyone else ever has this?

### Expected Behavior

Not crash

### Actual Behavior

```
<--- Last few GCs --->

[37477:0x130008000]  7075356 ms: Mark-sweep 4025.7 (4134.4) -> 4015.9 (4140.4) MB, 3782.7 / 0.0 ms  (average mu = 0.694, current mu = 0.007) allocation failure scavenge might not succeed
[37477:0x130008000]  7095099 ms: Mark-sweep 4034.0 (4142.4) -> 4023.3 (4147.9) MB, 4517.8 / 0.0 ms  (average mu = 0.742, current mu = 0.771) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0x102d1846c node::Abort() [/Users/jclackett/.fnm/node-versions/v16.13.1/installation/bin/node]
 2: 0x102d185f4 node::errors::TryCatchScope::~TryCatchScope() [/Users/jclackett/.fnm/node-versions/v16.13.1/installation/bin/node]
 3: 0x102e667b4 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/Users/jclackett/.fnm/node-versions/v16.13.1/installation/bin/node]
 4: 0x102e66748 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/Users/jclackett/.fnm/node-versions/v16.13.1/installation/bin/node]
 5: 0x102fe9c08 v8::internal::Heap::GarbageCollectionReasonToString(v8::internal::GarbageCollectionReason) [/Users/jclackett/.fnm/node-versions/v16.13.1/installation/bin/node]
[...]

PR #4293, opened 2022-09-29
fix(remix-dev/cli): JavaScript heap out of memory

arnaudambro:
closing #1555 

I'm not sure of the fix though because I don't understand the whole logic there

but the bug was a pain in the *ss before that "fix", and I have no more bug after !


arnaudambro:
closing it, I forked from `main` instead of `dev` and I don't know how to fix that

I'll repoen another one

PR #4294, opened 2022-09-29
fix(remix-dev/cli): fix heap out of memory error

arnaudambro:
closing #1555

I'm not sure of the fix though because I don't understand the whole logic there

but the bug was a pain in the *ss before that "fix", and I have no more bug after !

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -228,6 +228,7 @@ export async function watch(
       log("Rebuilding...");
     },
     onRebuildFinish() {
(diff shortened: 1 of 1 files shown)


balzdur:
It would be awesome if someone from the core team could look at this 🙏🏽

pcattori:
Superceded by https://github.com/remix-run/remix/pull/5133

Note that we needed to rearchitect the dev server as moving the require cache purging to `onRebuildFinished` in the current dev server [caused race conditions](https://github.com/remix-run/remix/pull/4453#discussion_r1009847407)

PR #4299, opened 2022-09-29
bug: server only built-ins polyfilled in client bundle

p-m-p:
Hi, I've been unable to update `@azure/msal-node` library since version `1.12.1` due to the introduction of the loopback listener which imports `createServer` from the `http` builtin as per below.

```js
import http, { createServer } from 'http';
```
From what I can tell from the error message it is trying to resolve from the rollup http polyfill plugin that doesn't implement this method. I tried adding the library to `serverDependenciesToBundle` but that didn't seem to solve anything and if I edit the imported file in node_modules to just `import http from 'http';` everything builds and runs as expected.

Appreciate any help or direction to getting this resolved. Thanks.

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -47,39 +47,35 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/api/client.js": js`
-        import { ConfidentialClientApplication } from '@azure/msal-node';
+      // XXX Remove .server from file for missing polyfill method error
+      "app/api/client.server.js": js`
+        import { AccountInfo } from '@azure/msal-node';
         
-        export function getGreeting() {
-          return 'hola'
+        export function getAccount() {
+          return { name: 'Dora' }
         }
       `,
       "app/routes/index.jsx": js`
         import { json } from "@remix-run/node";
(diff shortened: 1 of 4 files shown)


p-m-p:
I've updated the example test to highlight what I believe to be a wider issue with the polyfill approach in the code splitting and tree shaking.

You can see in the test that by importing from a built in directly in a route and using that in the loader only (no need to polyfill) you end up with the entire polyfill in the client bundle for that route.

MichaelDeBoey:
Closed by #5274

integration/bug-report-test.ts
```diff
@@ -47,30 +47,37 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
+      // XXX Remove .server from file for missing polyfill method error
+      "app/api/client.server.js": js`
+        import { AccountInfo } from '@azure/msal-node';
```
jacob-ebey, line 52: This lib is not meant to be utilized in the browser and is expected to require / use a bunch of node polyfills.

p-m-p: The issue I faced here was with using from a loader in a route which perhaps is considered client code but that wasn't clear to me.  I did solve it as per my previous comment by renaming the included file with `.server.ts` to it wasn't bundled but left this here due to the issue with the polyfills that looks to be closed now so is likely no longer an issue.

PR #4302, opened 2022-09-29
chore(templates): convert all `.eslintrc` files to JS

MichaelDeBoey:
ESLint has a new flat config feature (which is going to be the default in v9 or v10), which [requires a JS config file](https://eslint.org/blog/2022/08/new-config-system-part-2/#the-new-config-file%3A-eslint.config.js)

https://eslint.org/blog/2022/08/new-config-system-part-1
https://eslint.org/blog/2022/08/new-config-system-part-2
https://eslint.org/blog/2022/08/new-config-system-part-3

### DIFF
--- a/packages/remix-server-runtime/__tests__/.eslintrc
+++ b/packages/remix-server-runtime/__tests__/.eslintrc
@@ -1,6 +1,0 @@
-{
-  "rules": {
-    "no-restricted-globals": "off",
-    "import/no-nodejs-modules": "off"
-  }
-}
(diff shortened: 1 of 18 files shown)


mcansh:
it also renames the config to eslint.config.js, so i think we can wait until we fully support the flat config

MichaelDeBoey:
@mcansh This change makes that people only need to change the name of the file, no need for extra formatting anymore (as automatic formatting isn't included in these templates)

mcansh (approved):

PR #4307, opened 2022-10-02
feat(cli): update a touchfile upon successful (re)build in watch mode

andrewbrey:
When in `watch` mode with the `remix` cli we do incremental rebuilds of changed files, but unfortunately there is no way for external processes (e.g. a dev server) to _know_ that a (re)build has successfully finished other than by watching the build output directories, which is a lot of open file descriptors just to know something that the `remix` cli already knows.

This PR adds a very simple signal, in the form of a well-known file saved to the configured cache directory, that external processes can use as a simple mechanism to know concretely that Remix is done rebuilding files.

My ideal solution for this use case would be something more like callbacks supplied via the `remix.config.js`, e.g.

```js
// remix.config.js

module.exports = {
	// ... other configuration
	watchFirstBuildCallback: () => {
		// do stuff on first build, e.g. start a nodemon process for my `express` server...
	},
	watchRebuildCallback: () => {
		// do stuff on successful rebuilds, e.g. purge my express server `require` cache, or just reboot it...
	}
}
```

but that is a bit bigger in scope, so I thought it would be useful to just start small with this :smile:

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -219,6 +219,19 @@ export async function watch(
     broadcast({ type: "LOG", message });
   }
 
+  // Whenever a (re)build finishes in "watch" mode we update a "touchfile"
+  // which external processes can use as a signal of successful builds
+  // without having to watch all of the same files that remix is
+  // already watching - e.g. nodemon watching just this touchfile can
+  // know when to restart a dev server
+  function touch() {
+    fse.writeFileSync(
+      path.join(config.cacheDirectory, ".watch-touch"),
+      `last successful watch build: ${Date.now()}`,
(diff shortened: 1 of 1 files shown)


pcattori:
One workaround could be to only watch the assets manifest and see if the hash in there changes. If you need to watch for server builds, then you only need to watch `build/index.js`. So that would be an effective way to detect changes by only watching for changes in two files.

pcattori:
I'm also looking into a Node API for the Remix compiler so that you could call `watch` yourself with the appropriate callbacks (E.g. `onRebuildSuccess` or something). I think that would ultimately be better for most use-cases than relying on sentinel file watching, though we can certainly make that doable too.

pcattori:
Related to #5133

PR #4319, opened 2022-10-04
feat(revalidate): add special response to send revalidation data on mutations

kentcdodds:
Hey, I was looking at this:

image

And thought it would be better if it could look like this:

image

The idea is simple, have the mutation response include revalidation data. This could speed up revalidation by a *lot* for most use cases.

Here's a quick video demonstrating what this does:

https://user-images.githubusercontent.com/1500684/193799965-34b03519-6293-485e-8cd6-46242b9e2269.mp4

The API for this is simple:

```tsx
export async function action({ request }: ActionArgs) {
  const formData = await request.formData();
  const title = formData.get("title");
  invariant(typeof title === "string", "title must be a string");
  await db.createTodo({ title });
  return revalidate(); // <-- this is all actions need to do to trigger revalidation of all applicable loaders
}
```

`revalidate` accepts `data` which will be used for the fetcher's `data` property and ResponseInit.

My implementation is missing a bunch of stuff. I just spent a few minutes working on this to play around with the idea and make sure it would work. Feel free to close the PR and re-implement it correctly if you'd like.

Let me know what you think!

### DIFF
--- a/packages/remix-node/index.ts
+++ b/packages/remix-node/index.ts
@@ -41,6 +41,7 @@ export {
   isCookie,
   isSession,
   json,
+  revalidate,
   JsonFunction,
   MaxPartSizeExceededError,
   redirect,
--- a/packages/remix-react/data.ts
+++ b/packages/remix-react/data.ts
@@ -30,6 +30,13 @@ export function isRedirectResponse(response: any): boolean {
   );
 }
 
+export function isRevalidatedResponse(response: any): boolean {
+  return (
+    response instanceof Response &&
+    response.headers.get("X-Remix-Revalidated") != null
+  );
+}
+
 export async function fetchData(
   url: URL,
   routeId: string,
--- a/packages/remix-react/routes.tsx
+++ b/packages/remix-react/routes.tsx
@@ -9,8 +9,10 @@ import {
   fetchData,
   isCatchResponse,
   isRedirectResponse,
+  isRevalidatedResponse,
 } from "./data";
 import type { Submission } from "./transition";
+import { Revalidated } from "./transition";
 import { CatchValue, TransitionRedirect } from "./transition";
 import { prefetchStyleLinks } from "./links";
 import invariant from "./invariant";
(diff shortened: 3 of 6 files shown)


cliffordfajardo:
Some thoughts
- like that it cleans up the network tab
- API is explicit (`return revalidate()`) so it doesn't feel magical

Question / Concerns
Before, the revalidation happened with additional independent `GET` calls., if we have a page with lots of nested routing and we need to make lets say 2-3 API calls for the current route, I sense we might run into issues where this POST is taking longer to resolve than expected slowing down the UI. 

How will this effect streaming responses from some loaders higher up in hierachy?

afoures:
Love the explicit API but what about browser cache?
If you have a loader with some sort of cache control for the browser, this revalidation will bypass it, and resend the data, so potentially you will send more than what is really needed.

kentcdodds:
Caching comments here makes sense. I wonder if in addition to this we also add rendering link tags to prefetch the individual loader URLs as well. I understand that feels wasteful, but having a much faster revalidation feels worth it in some scenarios.

As for loaders with cache headers, on mutations you probably want to bypass the cache anyway because that data may have actually changed.

To address @cliffordfajardo's question about how long it takes, this would always be faster because the transition manager will keep things in a pending state until revalidation is finished anyway.

I'm not sure how this would affect streaming.

kiliman:
I think this will be just another one of those "levers" the developer can pull. As with most tech decisions, it will have some tradeoffs (like caching), but for specific use cases, may be a net positive. 

When I first started with Remix, I also wanted to minimize the number of requests by pipelining (my implementation added multiple route-ids in a header).

MichaelDeBoey:
Hi @kentcdodds!

kentcdodds:
This was a demo of a feature the Remix team will need to evaluate. I'm curious what @ryanflorence thinks

machour:
@kentcdodds wouldn't reloading the page trigger the same (duplicate) POST again?

kentcdodds:
I don't believe so

Issue #4328, opened 2022-10-05
Private GitHub stack templates don't work with the `releases/latest/download/{filename}` url structure


johnmberger:
### What version of Remix are you using?

1.7.2

### Steps to Reproduce

Instantiate a stack using a template from a private GitHub repo, an access token, and the "latest" release url:

```
npx create-remix@latest ./my-new-project --template https://github.com/owner/repo/releases/latest/download/stack.tar.gz --token myaccesstoken
```

### Expected Behavior

Since the `https://github.com/owner/repo/releases/latest/download/stack.tar.gz` url should always link to the latest release tag, I would expect the stack validation to work successfully and the remix project to initialize successfully. 

### Actual Behavior

The command fails with the following error: `🚨 The template file could not be verified. Please double check the URL and try again.`

However, if you do know the latest release tag, manually include it, and try again:
```
npx create-remix@latest ./my-new-project --template https://github.com/owner/repo/releases/download/v.1.1/stack.tar.gz --token myaccesstoken
```
Then it works just fine.

I suspect this is because there's a quirk in the Github URL structure between the latest tag and a specified version (`releases/latest/download` vs. `releases/download/version`) and the Remix code isn't handling the former.

My org maintains a private Github monorepo that includes a stack tarball and it would be great to allow users to link directly to the latest release instead of having to know the exact version number.

machour:
Thank you @johnmberger!

PR #4329, opened 2022-10-05
fix(create-remix): Allow private GitHub repos to use the latest release url for tarballs

johnmberger:
Closes: #4328

- [x] Docs
- [ ] Tests

Testing Strategy:

My main mode of testing was to try this out against a private repo with a stack tarball by running these three permutations:

```sh
# 🟢  new functionality
npx create-remix ./my-project --template https://github.com/<owner>/<repo>/releases/download/v.1.1/stack.tar.gz --token <my-token>

# 🟢  existing functionality
npx create-remix ./my-project --template https://github.com/<owner>/<repo>/releases/download/<version>/stack.tar.gz --token <my-token>

# 🔴 using a non-existent version (should fail)
npx create-remix ./my-project --template https://github.com/<owner>/<repo>/releases/download/<bad version>/stack.tar.gz --token <my-token>

# Verify all existing tests pass
yarn test
```

I attempted unsuccessfully to unit test this, but without spending much more time mocking out github api calls, I wasn't able to write anything worthwhile... happy to revisit if you disagree! I also cleaned up a tertiary test to make more sense.

### DIFF
--- a/packages/remix-dev/__tests__/create-test.ts
+++ b/packages/remix-dev/__tests__/create-test.ts
@@ -260,38 +260,13 @@ describe("the create command", () => {
     );
   });
 
-  it("succeeds for links to latest private github release tarballs when including token", async () => {
-    let projectDir = await getProjectDir(
-      "latest-private-release-tarball-with-token"
-    );
-    await run([
-      "create",
-      projectDir,
-      "--template",
-      "https://github.com/private-org/private-repo/releases/latest/download/stack.tar.gz",
-      "--no-install",
-      "--typescript",
-      "--token",
-      "valid-token",
-    ]);
-    expect(output.trim()).toBe(
-      getOptOutOfInstallMessage() +
-        "\n\n" +
-        getSuccessMessage(
-          path.join("<TEMP_DIR>", "latest-private-release-tarball-with-token")
-        )
-    );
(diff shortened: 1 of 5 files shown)


packages/remix-dev/__tests__/create-test.ts
```diff
@@ -266,7 +266,7 @@ describe("the create command", () => {
       "create",
       projectDir,
       "--template",
-      "https://example.com/remix-stack.tar.gz",
+      "https://github.com/private-org/private-repo/releases/download/v0.0.1/stack.tar.gz",
```
machour, line 269: Are you sure about this change?

johnmberger: Yeah, I think so. This test is for `succeeds for private github release tarballs when including token`, which should be a github url instead of `example.com`:
https://github.com/remix-run/remix/blob/6ccac77e14d240cd8eb5b213021f08eb9cbb32a0/packages/remix-dev/__tests__/create-test.ts#L263-L284

We test `example.com` (correctly) here:
https://github.com/remix-run/remix/blob/6ccac77e14d240cd8eb5b213021f08eb9cbb32a0/packages/remix-dev/__tests__/create-test.ts#L228-L245

mcansh (approved):

PR #4332, opened 2022-10-05
fix: ensure dev script runs Remix in development

mdingena:
## Context

We're building a single Remix codebase to host multiple websites. They share a lot of code, such as components. We launch multiple instances of Remix and instruct each instance to run in a certain "mode". This will make Remix switch between different `appDirectory`.

```js
const invariant = require('tiny-invariant');

invariant(typeof process.env.APP_MODE !== 'undefined', 'APP_MODE environment variable must be set.');

module.exports = {
  ignoredRouteFiles: ['**/.*'],
  appDirectory: `app-${process.env.APP_MODE}`
};
```

During testing of this code I noticed that the error message was not printed. `tiny-invariant` only prints error messages in non-production environments.

So I examined the `dev` script, and noticed that `cross-env` is only called on a subscript. I examined console output when running this command and found:

```console
Error: Error loading Remix config at /path/to/project/remix.config.js
Error: Invariant failed
```

Re-running the command with `APP_MODE` set:

```console
Built in 422ms

Watching Remix app in development mode...
💿 Built in 702ms
```

Note that the first build is `production` and the second is `development`.

I've asked around on Discord but I haven't seen an explanation of why the initial build must be done in production mode. It seems this is an oversight, perhaps.

## Changes

- Moved the `cross-env` command into the parent script.

## Consequences

Console output of `npm run dev` now contains the error message as expected in development mode:

```console
Building Remix app in development mode...
Error: Error loading Remix config at /path/to/project/remix.config.js
Error: Invariant failed: APP_MODE environment variable must be set.
```

Console output of `APP_MODE="something" npm run dev` is now:

```console
Building Remix app in development mode...
Built in 474ms

Watching Remix app in development mode...
💿 Built in 388ms
```

Note `development` is used both times, as expected.
```

### DIFF
--- a/templates/express/package.json
+++ b/templates/express/package.json
@@ -3,8 +3,8 @@
   "sideEffects": false,
   "scripts": {
     "build": "remix build",
-    "dev": "remix build && run-p \"dev:*\"",
-    "dev:node": "cross-env NODE_ENV=development nodemon ./server.js --watch ./server.js",
+    "dev": "cross-env NODE_ENV=development remix build && run-p \"dev:*\"",
+    "dev:node": "nodemon ./server.js --watch ./server.js",
     "dev:remix": "remix watch",
     "start": "cross-env NODE_ENV=production node ./server.js"
   },

### CODE REVIEW

**Comment by mdingena** [comment]
https://github.com/remix-run/remix/pull/4333

**Comment by kiliman** [comment]
The issue here is that with the Express adapter, you are responsible for 2 processes. One is your node Express server, and the other is the `remix watch` process that rebuilds your app when it changes. The `run-p dev:*` means to run both processes in parallel.

I believe the initial `remix build` is there because the Express app is requiring the app build `require(BUILD_DIR)`, so it will fail if the build hasn't been done yet. So that initial build was simply "priming the pump".

However, in the current release of the Express template, the `require(BUILD_DIR)` is only called inside a request, not during initialization. So you can simply remove the initial build, as `remix watch` will have done the build before your first request. You can still move `cross-env` up, but `remix watch` defaults to `development`.

```json
"scripts": {
  "dev": "run-p dev:*",
  "dev:node": "cross-env NODE_ENV=development nodemon ./server.js --watch ./server.js",
  "dev:remix": "remix watch"
},
```

mdingena:
> ```json
> "scripts": {
>   "dev": "run-p dev:*",
>   "dev:node": "cross-env NODE_ENV=development nodemon ./server.js --watch ./server.js",
>   "dev:remix": "remix watch"
> },
> ```

I tried this after deleting `/build/` and `/public/build/`, and it won't be able to start. So I think my PR is still valid?

mdingena:
For some reason the change this PR proposes breaks `remix watch` behaviour (changes are watched and rebuilds occur but do not appear in the frontend, even after a hard page refresh).

Requires you to type `rs` to restart `nodemon` or just `^C` out of Remix and restart the dev command.

Regardless, this PR is not very useful that way.

PR #4333, opened 2022-10-06
chore: add mdingena to contributors.yml

mdingena:
Signed the CLA.


mdingena:
I realise this should just be done inside the first PR that I open.

Issue #4348, opened 2022-10-10
Server Side Top Level Await does not work


joshocalico:
### What version of Remix are you using?

Latest (1.7.1)

### Steps to Reproduce

Use this export in any loader or server file.
```js
// My awesome computed function
const myPromise = new Promise((resolve) => {
    resolve("Hello Async");
});

const myResult = await myPromise;

export { myResult };
```

Note: this doesn't apply to client side Javascript/Typescript as Remix rightfully respects the tsconfig / jsconfig.

### Expected Behavior

It will evaluate the promise inside the module while it is being imported.

### Actual Behavior

The app doesn't build.

```
> remix build

Building Remix app in production mode...

✘ [ERROR] Top-level await is not available in the configured target environment ("node14")

    app/webapp.config.server.ts:91:22:
      91 │ export const colors = await getColors();
         ╵                       ~~~~~


Build failed with 1 error:
app/webapp.config.server.ts:91:22: ERROR: Top-level await is not available in the configured target environment ("node14")
Error
    at Object.onBuildFailure (/Users/joshuasteele/fluidity/fluidity-app/web/app.fluidity.money/node_modules/@remix-run/dev/dist/cli/commands.js:183:13)
    at buildEverything (/Users/joshuasteele/fluidity/fluidity-app/web/app.fluidity.money/node_modules/@remix-run/dev/dist/compiler.js:282:13)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.build (/Users/joshuasteele/fluidity/fluidity-app/web/app.fluidity.money/node_modules/@remix-run/dev/dist/compiler.js:102:3)
    at async Object.build (/Users/joshuasteele/fluidity/fluidity-app/web/app.fluidity.money/node_modules/@remix-run/dev/dist/cli/commands.js:178:3)
    at async Object.run (/Users/joshuasteele/fluidity/fluidity-app/web/app.fluidity.money/node_modules/@remix-run/dev/dist/cli/run.js:496:7)
ERROR: "build:remix" exited with 1.
```

machour:
The error message seems quite self-explanatory: top level await are not available for node14.
Can you try with node16 or maybe node14.14 (see https://github.com/evanw/esbuild/issues/253#issuecomment-1042859888)?

joshocalico:
Problem is largely here, remix build doesn't allow you to configure any other target.
For my other non-remix related server things it works fine - I'm using target node16 in my own esbuild builds.

[...]

PR #4349, opened 2022-10-10
Top Level Await Support on Server Side

joshocalico:
This bumps the base required Node version to 14.8 as to support top level await in Remix build.

Closes: #4348 

- [ ] Docs
- [ ] Tests

Testing Strategy: TBA

### DIFF
--- a/templates/deno/app/entry.client.tsx
+++ b/templates/deno/app/entry.client.tsx
@@ -8,7 +8,7 @@ function hydrate() {
       document,
       <StrictMode>
         <RemixBrowser />
-      </StrictMode>
+      </StrictMode>,
     );
   });
 }


sergiodxa:
I think the reason to target 14 instead of a newer version was to support more deployment platforms, some Serverless platforms still don't support newer Node.js versions. It would be great to review this first.

I checked and Vercel and Netlify both support v16, if the rest also support v16 maybe Remix could upgrade the required Node version to v16.

Another thing to consider is that v14 is still under maintenance mode until the 2023-04-30, so this could be another reason to keep support for v14 until that version reach end of life.

joshocalico:
@sergiodxa Granted that we're on security support I think it's be sensible to assume greater than 14.8 for SaaS platforms too. This is only a Node semver minor bump.
All _secure_ or even _supported_ installations of Node 14 would support top-level await. 
Technically the only secure version is really Node 14.20.1 currently anyway.

mcansh:
esbuild does not appear to support top level await https://github.com/evanw/esbuild/issues/253

PR #4355, opened 2022-10-11
fix: avoid collisions on pathless routes with indices

brophdawg11:
This is a drastically simpler approach to fix #3014 than the one I originally proposed in #3986, since it turns out `index` routes with `path`'s _are_ valid in react-router's matching algorithm (see https://github.com/remix-run/react-router/pull/8151).  They effectively serve as a shorthand to avoid adding the nested `index` under the `path` route.

Therefore, it turns out that the underlying issue here was just the inadvertent trailing slash generated by `defineConventionalRoutes`.

For the following `routes/` folder:

```
routes/
  nested/
    __pathless/
      foo.jsx
    __pathless.jsx
   index.jsx
```

We were generating a route tree of:

```jsx
<Routes>
  <Route file="root.jsx">
    <Route path="nested/" file="routes/nested/__pathless.jsx">
      <Route path="foo" file="routes/nested/__pathless/foo.jsx" />
    </Route>
    <Route path="nested" index file="routes/nested/index.jsx" />
    <Route index file="routes/index.jsx" />
  </Route>
</Routes>
```

Notice the trailing slash on the pathless layout route, which was causing it to score higher due to the segment score since `path.split('/')` produced an empty segment at the end.  We just need to also trim trailing slashes from our generated non-index paths and get the following route tree which scores correctly and ranks the index route higher:

```jsx
<Routes>
  <Route file="root.jsx">
    <Route path="nested" file="routes/nested/__pathless.jsx">
      <Route path="foo" file="routes/nested/__pathless/foo.jsx" />
    </Route>
    <Route path="nested" index file="routes/nested/index.jsx" />
    <Route index file="routes/index.jsx" />
  </Route>
</Routes>
```

Closes: #3014

- [x] Tests

Testing Strategy: `yarn test:integration conventional-routes`

### DIFF
--- a/packages/remix-dev/config/routesConvention.ts
+++ b/packages/remix-dev/config/routesConvention.ts
@@ -188,6 +188,8 @@ export function createRoutePath(partialRouteId: string): string | undefined {
 
   if (rawSegmentBuffer === "index" && result.endsWith("index")) {
     result = result.replace(/\/?index$/, "");
+  } else {
+    result = result.replace(/\/$/, "");
   }
 
   return result || undefined;
--- a/.changeset/stale-spoons-tie.md
+++ b/.changeset/stale-spoons-tie.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Avoid collisions between pathless layout routes and their nested index routes
--- a/integration/conventional-routes-test.ts
+++ b/integration/conventional-routes-test.ts
@@ -50,8 +50,6 @@ test.beforeAll(async () => {
         }
       `,
       "app/routes/nested/__pathless/foo.jsx": js`
-        import { useMatches } from '@remix-run/react';
-
         export default function Foo() {
           return <h1>Foo</h1>;
         }
--- a/integration/conventional-routes-test.ts
+++ b/integration/conventional-routes-test.ts
@@ -1,0 +1,107 @@
+import { test, expect } from "@playwright/test";
+
+import { PlaywrightFixture } from "./helpers/playwright-fixture";
+import type { Fixture, AppFixture } from "./helpers/create-fixture";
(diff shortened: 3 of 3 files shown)


packages/remix-dev/config/routesConvention.ts
```diff
@@ -188,6 +188,8 @@ export function createRoutePath(partialRouteId: string): string | undefined {
 
   if (rawSegmentBuffer === "index" && result.endsWith("index")) {
     result = result.replace(/\/?index$/, "");
+  } else {
+    result = result.replace(/\/$/, "");
```
brophdawg11, line 192: We should remove trailing slashes from all routes, not just `index` routes

kiliman:
Weird. I always make sure to add a trailing slash to all my index routes. This is what the browser expects.

`/parent/child/ <- index`
Then all relative URLs are relative to *child*
` <- /parent/child/photo.png`

If you don't add the trailing slash, then all relative URLs are relative to _parent_
`/parent/child`
` <- /parent/photo.png`

brophdawg11:
@kiliman I agree on that approach - this change shouldn't have an impact the URLs you use in your app as react router/remix don't really care about trailing slashes and will work with or without them.  This just normalizes an inconsistency in our conventional route `path` definitions which was causing  an inconsistency when scoring routes.

kiliman:
Yeah, I'm more complaining about how the hosting/framework ecosystem got away from the web standards. Unfortunately, hard to put that genie back in the bottle. At least Remix/RR is striving for some form of consistency. 👍

jacob-ebey:
Looks good to me but you might need to re-run the tests.

jacob-ebey (approved):

mcansh (approved):

PR #4359, opened 2022-10-12
feat(remix-server-runtime): RRR 1.2 - handleDataRequest

jacob-ebey:
[decision doc](https://github-md.com/remix-run/remix/blob/dev/decisions/0007-remix-on-react-router-6-4-0.md#decision)

Part 2 of step 1 in layering Remix back on top of React Router now that we have released `@remix-run/router`.

Changes:

- Implement handleDataRequest on top of RR alongside existing implementation

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -11,3 +11,6 @@
 integration/helpers/deno-template
 packages/remix-deno
 templates/deno
+
+# IDK What's up with this file...
+decisions/0007-remix-on-react-router-6-4-0.md
--- a/.eslintignore
+++ b/.eslintignore
@@ -11,6 +11,3 @@
 integration/helpers/deno-template
 packages/remix-deno
 templates/deno
-
-# IDK What's up with this file...
-decisions/0007-remix-on-react-router-6-4-0.md
--- a/integration/helpers/create-fixture.ts
+++ b/integration/helpers/create-fixture.ts
@@ -42,7 +42,7 @@ export async function createFixture(init: FixtureInit) {
     let url = new URL(href, "test://test");
     let request = new Request(url.toString(), {
       ...init,
-      signal: new AbortController().signal,
+      signal: init?.signal || new AbortController().signal,
     });
     return handler(request);
   };
@@ -52,6 +52,8 @@ export async function createFixture(init: FixtureInit) {
     routeId: string,
     init?: RequestInit
   ) => {
+    init = init || {};
+    init.signal = init.signal || new AbortController().signal;
     let url = new URL(href, "test://test");
     url.searchParams.set("_data", routeId);
     let request = new Request(url.toString(), init);
--- a/packages/remix-server-runtime/routes.ts
+++ b/packages/remix-server-runtime/routes.ts
@@ -75,15 +75,13 @@ export function createStaticHandlerDataRoutes(
               loadContext,
             })
         : undefined,
-      action: route.module.action
-        ? (args) =>
-            callRouteAction({
-              ...args,
-              routeId: route.id,
-              action: route.module.action,
-              loadContext,
-            })
-        : undefined,
+      action: (args) =>
+        callRouteAction({
+          ...args,
+          routeId: route.id,
+          action: route.module.action,
+          loadContext,
+        }),
       handle: route.module.handle,
       // TODO: RRR - Implement!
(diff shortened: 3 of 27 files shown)


.eslintignore
```diff
@@ -11,3 +11,6 @@
 integration/helpers/deno-template
 packages/remix-deno
 templates/deno
+
+# IDK What's up with this file...
+decisions/0007-remix-on-react-router-6-4-0.md
```
brophdawg11, line 16: This should be able to go away now if you merge latest `dev` as the issues are fixed

integration/helpers/create-fixture.ts
```diff
@@ -42,7 +42,7 @@ export async function createFixture(init: FixtureInit) {
     let url = new URL(href, "test://test");
     let request = new Request(url.toString(), {
       ...init,
-      signal: new AbortController().signal,
+      signal: init?.signal || new AbortController().signal,
```
brophdawg11, line 45: Chatted about the requirement of a signal and we want to stick with it to mimic the browser behavior, but we should make our `web-std-io` repo provide a signal automatically if not provided, and then remove the useless signals from vercel/netlify/architect adapters.

packages/remix-server-runtime/routes.ts
```diff
@@ -75,15 +75,13 @@ export function createStaticHandlerDataRoutes(
...
-              loadContext,
-            })
-        : undefined,
+      action: (args) =>
+        callRouteAction({
+          ...args,
+          routeId: route.id,
+          action: route.module.action,
+          loadContext,
+        }),
```
brophdawg11, line 84: Chatted through this and we want to keep toe consistency of a route not having an action (in the react router route tree) if it didn't export an action.

But we have some issues around the 405's making it to the catch boundary currently due to bad assumptions in `queryRoute`.  We lose the visibility into a "thrown" versus "returned" Response in some cases, most notably the 405 "no action found" scenario.  User-thrown or user-returned responses are already handled by `callRouteAction` so we can ignore those.  But we need to surface out router-generated responses so we can add an `X-Remix-Catch` header to them.

Desired Remix API is as follows:

```js
async function handleDataRequestRR(...) {
  try {
    // Requirements here:
    //  - unable to find a route -> throw a 404 response
    //  - unable to find a loader -> wont happen in remix due to route chunk loads
    //  - unable to find an action -> throw a 405 response
    //  - user throws a response -> captured in callRouteAction 
    //  - user returns a response -> we return that response
    let response = await staticHandler.queryRoute(request, routeId);      

    // Process redirects

    return response;
} catch (error) { 
  // All thrown responses to this point are automatically catch boundary errors,
  // they came from the router not user code
  if (error instanceof Response) {
    error.headers.set('X-Remix-Catch', 'yes');        
    return error;
  }

  // Handle 500 error boundary errors
} 
```
```

Comment (brophdawg11) [reply packages/remix-server-runtime/routes.ts:84]: Matt to look into the changes for createStaticHandler to avoid the above ambiguity.

**Comment by brophdawg11** [comment]
Note: `handleResourceRouteRR` should not pass a `routeId` to `queryRoute` so we get proper `?index` handling

File: packages/remix-server-runtime/routes.ts [hunk 4]
```diff
@@ -55,35 +59,51 @@ export function createStaticHandlerDataRoutes(
...
-          loadContext,
-        }),
-      handle: route.module.handle,
-      // TODO: RRR - Implement!
-      shouldRevalidate: () => true,
-    }));
+              route.id
+            ),
+            path: route.path,
+            ...commonRoute,
```
brophdawg11, line 106: Can move path into common route because index routes can have paths as a shorthand for

```jsx
<Route path="foo">
  <Route index>
</Route>

<Route index path="foo" />
```

integration/helpers/create-fixture.ts
jacob-ebey: Removed the unnecessary ones and allowed passthrough from config where necessary.

jacob-ebey:
> Note: `handleResourceRouteRR` should not pass a `routeId` to `queryRoute` so we get proper `?index` handling

Let's revisit this outside of this PR. I believe there is a behavior mismatch here we need to resolve.

Issue #4342, opened 2022-10-07
<Form> submitter is serialized out of tree order


jenseng:
### What version of Remix are you using?

1.7.2

### Steps to Reproduce

1. Make a `<Form>` like so:

```jsx
<Form>
  <button type="submit" name="foo" value="override">Override</button>
  <input type="hidden" name="foo" value="default" />
</Form>
```

2. Click the "Override" button

### Expected Behavior

The form fields should be serialized in tree order, i.e. `foo=override&foo=default` (as is the case with a vanilla `<form>`).

See [Constructing the form data set](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set) in the HTML spec. Submitted fields should be sent in the order they appear in the DOM, including the `submitter` (i.e. the submit button that was clicked).

### Actual Behavior

The form is serialized as `foo=default&foo=override`.

This is problematic on the server-side if you are relying on the order to indicate precedence. `formData.get("foo")` will yield different values, depending on whether the form was submitted by Remix or native browser behavior.

See also https://github.com/remix-run/remix/pull/3611; prior to that fix it was broken in a different way (it would only serialize `foo=default`)

jenseng:
On a related note, `<Form>` serializes image submit buttons in the wrong order **and** with the wrong key/values.

For example, given: `<input type="image" name="image" src="..." />`. If you click it, it gets serialized as `image=` at the very end, rather than the expected `image.x=X&image.y=Y` in tree order. See [Step 3.3](https://www.w3.org/TR/2014/REC-html5-20141028/forms.html#constructing-form-data-set).

Probably worth lumping in to this bug report, as the root cause is the same, and the fix should be too 🤞

jenseng:
I found some more differences between how `<Form>` and `<form>` behave:

1. In `<Form>` submissions, Safari includes the submit button value twice (once in tree order, once at the end) due to [this bug](https://bugs.webkit.org/show_bug.cgi?id=239070)
2. If you have file inputs and the `encType` is `application/x-www-form-urlencoded` (explicit or implicit):
   * `<form>` will send each file entry's `name` as the `value`, as per [the spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#convert-to-a-list-of-name-value-pairs)
   * `<Form>` will throw an error `File inputs are not supported with encType "application/x-www-form-urlencoded" ...`; if this `invariant` check is removed, no `value` gets submitted.

[...]

PR #4365, opened 2022-10-13
test: expand test coverage around forms

jenseng:
Expand test coverage around forms:

* Rework all form tests to run both with and without JavaScript. This way we can better detect/prevent regressions or inconsistencies between how `<form>` and `<Form>` work.
* Dedup and expand form method tests to ensure we cover all supported Form `method`s and submitter `formMethod`s
* Expand coverage around form serialization (tree order, image submit buttons, files in URL-encoded payloads)
* Conditionally mark tests as failing (via [`test.fail`](https://playwright.dev/docs/api/class-test#test-fail-1)):
   * Non-get/post `<Form>` methods with JavaScript disabled (#4420)
   * Form serialization problems with JavaScript enabled (#4342)

References: #4342

- [x] Tests

### DIFF
--- a/integration/form-test.ts
+++ b/integration/form-test.ts
@@ -308,9 +308,9 @@ test.describe("Forms", () => {
           }
         `,
 
-        "app/routes/submitter-formmethod.jsx": js`
-          import { useActionData, useLoaderData, Form } from "@remix-run/react";
-          import { json } from '@remix-run/node'
+        "app/routes/form-method.jsx": js`
+          import { Form, useActionData, useLoaderData, useSearchParams } from "@remix-run/react";
+          import { json } from "@remix-run/node";
 
           export function action({ request }) {
             return json(request.method)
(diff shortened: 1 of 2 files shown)


machour (commented):
Thank you for that js/no-js refactoring 🙌🏼 

Shouldn't the tests related to #4342 fail?

jenseng:
@machour the new tests **do** fail with JavaScript enabled, but they are annotated with [`test.fail(...)`](https://playwright.dev/docs/api/class-test#test-fail-2) to indicate that they are supposed to fail, which makes them pass if they fail and fail if they pass 🙃. See [here](https://github.com/remix-run/remix/pull/4365/commits/4944d26c94304561862e8fd9191e008871fb7dd9#diff-6ba8f3349b2b29f4668d3cdbcc4df75c8182edad8eaea07c5a6d7f227309acf7R627-R630) and [here](https://github.com/remix-run/remix/pull/4365/commits/4944d26c94304561862e8fd9191e008871fb7dd9#diff-6ba8f3349b2b29f4668d3cdbcc4df75c8182edad8eaea07c5a6d7f227309acf7R669-R672)...  it's the playwright equivalent of `@ts-expect-error`

So when we fix the behavior, we'll need to remove those annotations as well, otherwise the tests will start failing because they are passing 😆

jenseng:
I'm thinking it might be good to merge this PR independently of the actual fixes, since it expands coverage generally and gives us a little more flexibility in how/when we fix ... e.g. file handling fix could be separate from tree order fixes, and I have two proposals on how to fix the latter (implement the spec vs. do some temporary form DOM shenanigans)

mcansh:
thank you @jenseng! 🙌 

didn't realize these we're were only running these with js enabled 😱

jenseng:
In light of [this imminent fix](https://github.com/remix-run/remix/pull/4413) I'll rebase and amend the tests, as this is one instance where `<form>` and `<Form>` should behave differently

mcansh (approved):

MichaelDeBoey (approved):

machour (approved):

brophdawg11 (approved):
This looks great @jenseng  - thank you!

chaance (approved):

chaance:
@jenseng Have you already started working on fixes to these issues or should we prepare for that internally? I'd like to prioritize these bugs for the next release if possible.

jenseng:
@chaance I've got fixes/PRs basically ready, was just holding off till this gets merged... IMO the url-encoded-files fix should probably be a separate PR from the submitter fixes, and I have two proposals/approaches for the latter

brophdawg11:
@jenseng Would you mind tagging me on those as well?  We'll want the same changes in `react-router-dom`'s `<Form>`/`useSubmit` so we don't lose them when RR 6.4 is migrated into Remix.  I can copy your changes over there once they're good to go in Remix 👍

Issue #2498, opened 2022-03-25
Action in root.tsx not executed when pathless index route present


styxlab:
### What version of Remix are you using?

1.3.3

### Steps to Reproduce

# Install
git clone https://github.com/styxlab/remix-issue-root-action
cd remix-issue-root-action
yarn
yarn dev

# Reproduce
Enter a text into the text field and press enter => Error: 405 Method Not Allowed
```
Route "routes/__layout" does not have an action, but you are trying to submit to it. To fix this, please add an `action` function to the route
```

### Expected Behavior

No error. The action in `root.tsx` should be executed.

### Actual Behavior

Error: 405 Method Not Allowed

styxlab:
Note:
 1. the action in `root.tsx` is triggered, when you remove the pathless layout route structure
 2. the console error message is misleading because actions in pathless routes are not allowed.

kiliman:
Yeah, when Remix posts to a route, it looks for an action on that route. If you're on an index route, it will look at the parent layout for an action unless you include `?index` in the form action. Looks like Remix isn't skipping pathless layouts as it looks for the "parent" layout route and just stops at the first parent.

styxlab:
Thanks for explaining what's going wrong here. From a DX point of view I would expect that the action is always triggered in the same location where I define the <Form> unless it is changed in the form action attribute.

styxlab:
Quote from the docs:

> Also note that forms without an action prop (<Form method="post">) will automatically post to the same route within which they are rendered ...

Thus I hope this issue will be fixed.

numaga94:
please fix this +1

maranomynet:
> when Remix posts to a route, it looks for an action on that route. If you're on an index route, it will look at the parent layout for an action unless you include `?index` in the form action.

This breaks the principle of least surprise.

IMO, matched routes should have their `action` method invoked regardless of whether the filename is `index.tsx` or something else.

brophdawg11:
I think this is the same bug as [this issue](https://github.com/remix-run/react-router/issues/9284) over in react router.  Pathless routes should be ignored when matching for actions, so `/parent?index` should hit the index route, and `/parent` should use the `parent.jsx` match, not a potential `parent/__pathless.jsx` file.  Otherwise there's no way of differentiating between the parent route and it's child pathless routes.

machour:
Should be fixed now that Remix sits on top of RR.

PR #4376, opened 2022-10-17
fix: Ignore pathless layout routes when matching actions

brophdawg11:
Same fix as https://github.com/remix-run/react-router/pull/9455 in react router

If you have a route tree of:

```jsx
<Route path="parent" action={...} element={<Parent />}>
  <Route element={<Pathless />}>
    <Route index element={<Index />} />
  </Route>
</Route>
```

And a `<Form method="post">` in `<Parent />` submits, then it should ignore the pathless layout route in the middle and walk upwards to the last path-contributing matches action.

Closes: #2498

- [ ] Docs
- [x] Tests

Testing Strategy: Added integration tests

### DIFF
--- a/packages/remix-react/transition.ts
+++ b/packages/remix-react/transition.ts
@@ -576,7 +576,7 @@ export function createTransitionManager(init: TransitionManagerInit) {
         invariant(matches, "No matches found");
         if (fetchControllers.has(key)) abortFetcher(key);
 
-        let match = getFetcherRequestMatch(
+        let match = getRequestMatch(
           new URL(href, window.location.href),
           matches
         );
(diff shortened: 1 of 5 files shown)


packages/remix-react/transition.ts
```diff
@@ -630,17 +630,22 @@ export function createTransitionManager(init: TransitionManagerInit) {
...
-    url: URL,
-    matches: RouteMatch<ClientRoute>[]
-  ) {
+  function getRequestMatch(url: URL, matches: ClientMatch[]) {
     let match = matches.slice(-1)[0];
 
-    if (!isIndexRequestUrl(url) && match.route.index) {
-      return matches.slice(-2)[0];
+    if (isIndexRequestUrl(url) && match.route.index) {
+      return match;
```
brophdawg11, line 637: Return the leaf for index requests, otherwise return the lowest path contributing match

packages/remix-server-runtime/server.ts
```diff
@@ -633,11 +633,19 @@ function isIndexRequestUrl(url: URL) {
...
   let match = matches.slice(-1)[0];
 
-  if (!isIndexRequestUrl(url) && match.route.id.endsWith("/index")) {
-    return matches.slice(-2)[0];
+  if (isIndexRequestUrl(url) && match.route.id.endsWith("/index")) {
+    return match;
   }
 
-  return match;
+  return getPathContributingMatches(matches).slice(-1)[0];
```
brophdawg11, line 640: Same change on the server

packages/remix-react/transition.ts
```diff
@@ -590,7 +590,7 @@ export function createTransitionManager(init: TransitionManagerInit) {
         invariant(matches, "No matches found");
         if (fetchControllers.has(key)) abortFetcher(key);
 
-        let match = getFetcherRequestMatch(
+        let match = getRequestMatch(
```
brophdawg11, line 593: We now use the same util for action `submissions` and fetchers - so we only do this `?index`/`pathless` logic once

chaance (approved):

PR #4378, opened 2022-10-17
fix(remix-react): update server-side check to use window.document instead of AbortController

brophdawg11:
This prevents a false negative on `node >= 15` where `AbortController` exists on the server:

Screen Shot 2022-10-17 at 4 10 57 PM

I wasn't able to add an integration test for this since if you actually do fetch in SSR, it kills the server since it's an uncaught promise rejection _after_ the render pass has completed

### DIFF
--- a/packages/remix-react/transition.ts
+++ b/packages/remix-react/transition.ts
@@ -411,6 +411,12 @@ export const IDLE_FETCHER: FetcherStates["Idle"] = {
   data: undefined,
   submission: undefined,
 };
+
+const isBrowser =
+  typeof window !== "undefined" &&
+  typeof window.document !== "undefined" &&
+  typeof window.document.createElement !== "undefined";
+const isServer = !isBrowser;
 //#endregion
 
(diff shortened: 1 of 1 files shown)


packages/remix-react/transition.ts
```diff
@@ -567,6 +573,14 @@ export function createTransitionManager(init: TransitionManagerInit) {
       }
 
       case "fetcher": {
+        if (isServer) {
+          throw new Error(
+            "a fetcher was called during the server render, but it shouldn't be. " +
+              "You are likely calling useFetcher.load() or useFetcher.submit() in " +
+              "the body of your component. Try moving it to a useEffect or a callback."
+          );
+        }
```
brophdawg11, line 582: Lifted this check so it applies to any fetcher calls, not just `fetcher.load`

jacob-ebey (approved):

MichaelDeBoey (approved):

machour (approved):

PR #4385, opened 2022-10-18
feat(remix-server-runtime): RRR 1.3 / 1.4 - handleDocumentRequest

jacob-ebey:
[decision doc](https://github-md.com/remix-run/remix/blob/dev/decisions/0007-remix-on-react-router-6-4-0.md#decision)

Part 3 and 4 of step 1 in layering Remix back on top of React Router now that we have released `@remix-run/router`.

Changes:

- Implement handleDocumentRequest on top of RR alongside existing implementation
- Disambiguate existing internal errors surfacing to CatchBoundary

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:


packages/remix-server-runtime/server.ts
```diff
@@ -286,19 +286,7 @@ async function handleDataRequestRR(
...
-      // this point will ever have a body. This contrasts with
-      // the new router implementation that has a body under
-      // some conditions.
-      return new Response(null, {
-        status: error.status,
-        statusText: error.statusText,
-        headers: {
-          "X-Remix-Catch": "yes",
-        },
-      });
```
brophdawg11, line 282: @jacob-ebey This might break integration tests - unsure what prompted it in the first place but the new Response creation here was causing mismatches in response `content-type` headers in the `server.ts` unit tests :/

If having a body is problematic we can also look into removing the body from the router Response.  I _think_ the only one with a body now might be the the 405 no action found response but we may not be calling that anymore now that `callRouteAction` is conditional?

packages/remix-server-runtime/data.ts
```diff
@@ -34,7 +34,10 @@ export async function callRouteAction({
   isRemixRouterRequest?: boolean;
 }) {
   if (!action) {
-    let response = new Response(null, { status: 405 });
+    let response = new Response(null, {
+      status: 405,
+      statusText: "Method Not Allowed",
```
brophdawg11, line 39: Add `statusText` to this Remix-generated Response to align with 405's from elsewhere (including the router)

packages/remix-server-runtime/router/router.ts
```diff
@@ -1803,7 +1806,25 @@ export function unstable_createStaticHandler(
     let location = createLocation("", createPath(url), null, "default");
     let matches = matchRoutes(dataRoutes, location);
 
-    if (!matches) {
+    if (!validRequestMethods.has(request.method)) {
```
brophdawg11, line 1809: Short circuit on OPTIONS requests in the router

packages/remix-server-runtime/router/router.ts
```diff
@@ -1856,7 +1877,12 @@ export function unstable_createStaticHandler(
...
+      throw createRouterErrorResponse(null, {
+        status: 405,
+        statusText: "Method Not Allowed",
+      });
```
brophdawg11, line 1884: short circuit on OPTIONS requests in `queryRoute`

packages/remix-server-runtime/router/router.ts
```diff
@@ -1952,7 +1978,7 @@ export function unstable_createStaticHandler(
     if (!actionMatch.route.action) {
       let href = createHref(new URL(request.url));
       if (isRouteRequest) {
-        throw createRouterErrorResponse(`No action found for [${href}]`, {
+        throw createRouterErrorResponse(null, {
```
brophdawg11, line 1981: Align with remix response structure - we have the URL in the request so this isn't actually providing much extra info

packages/remix-server-runtime/server.ts
```diff
@@ -287,6 +287,16 @@ async function handleDataRequestRR(
...
   } catch (error) {
     if (error instanceof Response) {
+      // Our callRouteLoader/callRouteAction methods ensure we always have
+      // X-Remix-Catch on responses thrown from loaders and actions.  However,
+      // if we never make it to the loader (such as a 404 or 405) then we
+      // Remix Router puts it's own custom header on, which we swap here for
+      // X-Remix-Catch
+      if (error.headers.get("X-Remix-Router-Error") === "yes") {
+        error.headers.delete("X-Remix-Router-Error");
+        error.headers.set("X-Remix-Catch", "yes");
```
brophdawg11, line 297: Handle router-thrown responses in data requests (as opposed to user-thrown)

packages/remix-server-runtime/server.ts
```diff
@@ -307,14 +317,18 @@ function findParentBoundary(
...
+  // has default error/catch boundary handling.  This handles the case where
+  // react-router doesn't have a matching "root" route to assign the error to
+  // so it returns context.errors = { __shim-404-route__: ErrorResponse }
+  let route = routes[routeId] || routes["root"];
   let isCatch = isRouteErrorResponse(error);
   if (
-    (isCatch && route.module?.CatchBoundary) ||
-    (!isCatch && route.module?.ErrorBoundary) ||
+    (isCatch && route.module.CatchBoundary) ||
+    (!isCatch && route.module.ErrorBoundary) ||
```
brophdawg11, line 328: This was due to the `__shim-404-route_` stuff.  We now find the root route, but also made it defensive to fall back on it if we don't find one

packages/remix-server-runtime/server.ts
```diff
@@ -400,7 +400,15 @@ async function handleDocumentRequestRR(
...
-      appState.loaderBoundaryRouteId = id;
+      // Only assign the boundary id if this module has a boundary.  This
+      // should be true in almost all cases, except for cases in which no
+      // boundaries exist and the router "assigns" the error to the root route
+      // for lack of a better place to put it.  If the root doesn't have an
+      // error boundary, then we leave loaderBoundaryId null to bubble to the
+      // default boundary at render time
+      if (build.routes[id]?.module.ErrorBoundary) {
+        appState.loaderBoundaryRouteId = id;
+      }
```
brophdawg11, line 411: @jacob-ebey I wouldn't mind talking through the current router approach with you tomorrow for not-founds and error bubbling up _through _ the root to the Remix-provided defaults.  Might be something we can clean up.

packages/remix-server-runtime/server.ts
```diff
@@ -1010,7 +1018,23 @@ async function assertResponsesMatch(_a: Response, _b: Response) {
...
+        if (
+          text.includes(">Application Error</h1>") &&
+          text.includes("💿 Hey developer👋")
+        ) {
+          return stripStackTraceFromDefaultErrorBoundary(text);
+        }
+
+        if (text.includes("<script>window.__remixContext")) {
+          return stripStackTraceFromRemixContext(text);
+        }
```
brophdawg11, line 1034: Short of the above fix stack traces were the rest of the failures, so stripping them out of both `window.__remixConext` and the `<pre>` before comparing

packages/remix-server-runtime/server.ts
```diff
@@ -1019,6 +1043,39 @@ async function assertResponsesMatch(_a: Response, _b: Response) {
...
 
+function stripStackTraceFromDefaultErrorBoundary(text: string) {
+  let openPreStart = text.indexOf("<pre ");
+  let openPreEnd = text.indexOf('">', openPreStart + 1);
+  let closePre = text.indexOf("</pre>", openPreEnd + 1);
+  let error = text.substring(openPreEnd + 2, closePre);
+  let stackStart = error.indexOf("\n");
+  let errorWithoutStack = error.substring(0, stackStart);
+  return (
+    text.replace(error, "ERROR REMOVED") + "\n\nError:\n" + errorWithoutStack
```
brophdawg11, line 1054: Here we end up asserting the full HTML without `<script>window.__remixContext = {...}</script>` concatenated with the stringifed `remixContext` with `stack` replaced

packages/remix-server-runtime/server.ts
```diff
@@ -1019,6 +1043,39 @@ async function assertResponsesMatch(_a: Response, _b: Response) {
...
+  // If we didn't have an error - assert the full document
+  if (!remixContext?.appState?.error?.stack) {
+    return text;
+  }
+
+  // Otherwise remove the stack trace and assert the text/JSON as a
+  // concatenated string
+  remixContext.appState.error.stack = "yes";
+  text = text.replace(scriptContents, "CONTEXT REMOVED");
+  return text + "\n\nContext:\n" + JSON.stringify(remixContext);
```
brophdawg11, line 1076: Here we end up asserting the full HTML without `<script>window.__remixContext = {...}</script>` concatenated with the stringifed `remixContext` with `stack` replaced

packages/remix-server-runtime/server.ts
brophdawg11, line 1054: Assert the HTML without `<pre>` concatenated with the first line of the error stack - which is just the message and removes all the filenames

jacob-ebey:
> Just curious, what's this going to do? Remix already serves document requests for SSR so I suppose is another thing

All of the data loading / error handling logic is no longer going to be owned in the server-runtime, but rather by the router package. Essentially this now calls `router.query(request)` and that's it™️

machour:
@brophdawg11 is there a decision doc for this work? Would help me follow along

brophdawg11:
@machour Yep!  https://github.com/remix-run/remix/blob/main/decisions/0007-remix-on-react-router-6-4-0.md - this PR is steps `1.iii`/`1.iv` under `Decision` - steps `1.i`/`1.ii` were already merged in #4245 and #4359 (behind the compile-time flag)

packages/remix-server-runtime/data.ts
```diff
@@ -79,11 +82,11 @@ export async function callRouteLoader({
...
-    throw new Error(
-      `You made a ${request.method} request to ${request.url} but did not provide ` +
-        `a default component or \`loader\` for route "${routeId}", ` +
-        `so there is no way to handle the request.`
-    );
+    let pathname = new URL(request.url).pathname;
+    let msg =
+      `You made a ${request.method} request to "${pathname}" but did not provide ` +
+      `a \`loader\` for route "${routeId}", so there is no way to handle the request.`;
+    throw new ErrorResponse(405, "Method Not Allowed", new Error(msg), true);
```
brophdawg11, line 89: Jacob and Matt to think a bit more on 405 or 404 here

brophdawg11:
Need to add a changeset to describe the changes in error/catch boundary behaviors in Remix.  Merging this PR will enable those even without the flag enabled.

packages/remix-server-runtime/data.ts
brophdawg11: Landed on a 400 since it seemed the most accurate: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

> The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

brophdawg11 (approved):

brophdawg11:
I think this is good to go now?  Let's touch base with @chaance and make sure that we're good to merge this (since he already cut the `1.7.6` release branch).  This code in dev, without the feature flag enabled, would just result in the minor fixes to catch/error boundary handling and could be part of a `1.7.7` if needed.  But the plan is to enable the feature flag for a `1.8.0-pre.0` later this week or early next week.

PR #4386, opened 2022-10-18
docs(api/remix): fix infinite loop in `useFetcher` example

bjacobso:
I believe this causes an infinite loop because the fetcher is changing, this should instead be on initial mount

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -48,6 +48,7 @@
 - BenMcH
 - bgschiller
 - binajmen
+- bjacobso
 - bmarvinb
 - bmontalvo
 - bogas04
--- a/docs/api/remix.md
+++ b/docs/api/remix.md
@@ -766,7 +766,7 @@ function SomeComponent() {
   useEffect(() => {
     fetcher.submit(data, options);
     fetcher.load(href);
-  }, [fetcher]);
+  }, []);
 
   // build UI with these
   fetcher.state;


jacob-ebey:
> Don't know if this is the right approach, as this will be against the `react-hooks/exhaustive-deps` ESLint rule
> [facebook/react#14920](https://github.com/facebook/react/issues/14920)

This is a valid use of excluding deps and one of the reasons I don't like that lint rule.

machour (approved):

MichaelDeBoey:
@jacob-ebey Shouldn't we include `fetcher.load` & `fetcher.submit` into the dependencies array instead? 🤔

bjacobso:
@MichaelDeBoey I'm still getting the lint error with `[fetcher.load]`. 

This seems to work, but looks a little clunky:
```
  const fetcherLoad = fetcher.load

  useEffect(() => {
    fetcherLoad('/example')
  }, [fetcherLoad])
```

jonschlinkert:
Wait, isn't that an abstract example, to demonstrate the methods that can be used `useEffect`? 

In real usage, removing `fetcher` from the deps array would most likely be incorrect in almost all circumstances. You can fix the infinite loop by checking `fetcher.type` and `fetcher.state` inside `useEffect`.

machour:
That was indeed an abstract example.
Docs have been reworked and they now include several examples avoiding the infinite loop, so I'm going to close this.
Thanks again for opening this PR @bjacobso 🙏🏼

mbrowne:
It seems like most of the examples from the previous version of the docs have been removed. @machour were you referring to this as the recommended method?

```
  useEffect(() => {
    if (fetcher.state === "idle" && fetcher.data == null) {
      fetcher.load("/some/route");
    }
  }, [fetcher]);
```

(found at https://remix.run/docs/en/1.19.3/hooks/use-fetcher)

PR #4391, opened 2022-10-19
chore: ensure AbortController installed globally

jacob-ebey:
Make sure AbortController is installed globally for node version less than 15.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

Existing deployment tests will validate the change.

### DIFF
--- a/packages/remix-node/globals.ts
+++ b/packages/remix-node/globals.ts
@@ -2,6 +2,7 @@ import {
   ReadableStream as NodeReadableStream,
   WritableStream as NodeWritableStream,
 } from "@remix-run/web-stream";
+import { AbortController as NodeAbortController } from "abort-controller";
 
 import { atob, btoa } from "./base64";
 import {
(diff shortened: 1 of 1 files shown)


mcansh (approved):

packages/remix-node/globals.ts
```diff
@@ -35,6 +36,8 @@ declare global {
 
       ReadableStream: typeof ReadableStream;
       WritableStream: typeof WritableStream;
+
+      AbortController: typeof AbortController;
```
chaance, line 40: @jacob-ebey Looking at the `abort-controller` repo, it looks like there may be a few issues with its typings that could potentially cause issues when passing signals. 

https://github.com/mysticatea/abort-controller/issues/16#issuecomment-592464104
https://github.com/mysticatea/abort-controller/issues/36

jacob-ebey: It's not an issue at runtime as we are currently running tests using it. As long as the build passes and doesn't have type issues it's not an issue.

chaance (approved):

PR #3034, opened 2022-04-30
feat(remix-node): add `getFilePath`& `remove` methods to `NodeOnDiskFile`

cstrnt:
Closes: #3021

- [ ] Docs
- [x] Tests

Testing Strategy:

- Added a `test` script in the `remix-node` package (remix-node tests never ran before)
- Added a simple unit test for the `createFileUploadHandler` function

PR #4408, opened 2022-10-23
feat(remix-node): add `getFilePath`& `remove` methods to `NodeOnDiskFile`

julienmonnard:
Closes: #3021

- [ ] Docs
- [x] Tests

Testing Strategy:

One unit test for each new function, added in the remix-node package.

This is basically the same PR as #3034, but it got closed by its author before being merged. I believe this is a useful feature and would very much like to see it added.

### DIFF
--- a/.changeset/flat-cobras-beam.md
+++ b/.changeset/flat-cobras-beam.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/node": minor
+---
+
+Add a "remove" and a "getFilePath" methods to fileUploadHandler
--- a/.changeset/flat-cobras-beam.md
+++ b/.changeset/flat-cobras-beam.md
@@ -2,4 +2,4 @@
 "@remix-run/node": minor
 ---
 
-Add a "remove" and a "getFilePath" methods to fileUploadHandler
+Add "remove" and "getFilePath" methods to fileUploadHandler
--- a/.changeset/flat-cobras-beam.md
+++ b/.changeset/flat-cobras-beam.md
@@ -2,4 +2,4 @@
 "@remix-run/node": minor
 ---
 
-Add "remove" and "getFilePath" methods to fileUploadHandler
+Add `remove` and `getFilePath` methods to `NodeOnDiskFile`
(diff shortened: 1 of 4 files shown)


.changeset/flat-cobras-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/node": minor
+---
+
+Add a "remove" and a "getFilePath" methods to fileUploadHandler
```
machour, line 5:
```suggestion
Add "remove" and "getFilePath" methods to fileUploadHandler
```

machour (approved):
Merci Julien for picking up this! 🙌

Just a small update to the changesetand we should be good to go.

.changeset/flat-cobras-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/node": minor
+---
+
+Add "remove" and "getFilePath" methods to fileUploadHandler
```
MichaelDeBoey, line 5: The exported class is called `NodeOnDiskFile`, so I think we should reference that name instead.

```suggestion
Add `getFilePath`& `remove` methods to `NodeOnDiskFile`
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

cstrnt:
Nice copy pasta :D

julienmonnard:
Indeed. I'm not trying to hide that fact, it's actually mentioned in the description above.

The main difference with your PR is that this one is still open. I need that feature in my project and would like to see it merged. I was disappointed when you closed yours, and as I cannot reopen a PR, I recreated it.

mcansh (approved):

julienmonnard:
Is there something more that should be done for this to be able to make it in one of the next releases?
I guess with the whole react-router/remix stuff happening it was quite tricky, but now that 1.10 wraps it up and has landed, maybe there could be a little slot for this one :)

mcansh (approved):

PR #4410, opened 2022-10-24
refactor(dev): simplify compiler

pcattori:
Refactored the compiler to be architected around `BrowserCompiler`, `ServerCompiler`. Notably, each of these have a `compile` function that can be called once for the initial build, but can be called again subsequently for rebuilds. This greatly simplifies the mental model for the Remix compiler and cuts down the code in `watch`. Other changes include:

- Previously, we used a a ref to the assets manifest promise to coordinate the browser and server builds. The ref was also reused across rebuilds. There were checks in `watch` and within the `serverAssetsManifestPlugin` to make sure the ref was set before being accessed. Overall, this approach was error prone and hard to reason about. This has now been replaced with a Golang-style channel so that the browser build can explicitly communicate the assets manifest to the server build.

- Previously, writing the assets manifest and the server build to the file system needed to be done anywhere in the code where we built the browser and server bundles. This was easy to forget and resulted in duplication. Now, the `compile` function for the browser and server compilers managed writing the results (including the assets manifest) to the filesystem.

Note that the `build` and `watch` APIs for the compiler stayed the same (with the minor exception of renaming `onBuildFailure` to `onCompilerFailure`).


packages/remix-dev/compiler/compile-server.ts
```diff
@@ -147,3 +146,27 @@ export async function writeServerBuildResult(
     }
   }
 }
+
+export const createServerCompiler = (
+  remixConfig: RemixConfig,
+  options: BuildOptions
+): ServerCompiler => {
+  let compile = async (manifestChannel: ReadChannel<AssetsManifest>) => {
+    console.log("started server build");
```
pcattori, line 154: TODO: remove console.log

packages/remix-dev/compiler/assets.ts
```diff
@@ -87,7 +87,7 @@ export async function createAssetsManifest(
       );
       let route = routesByFile.get(entryPointFile);
       invariant(route, `Cannot get route for entry point ${output.entryPoint}`);
-      let sourceExports = await getRouteModuleExportsCached(config, route.id);
+      let sourceExports = await getRouteModuleExports(config, route.id);
```
jacob-ebey, line 90: why aren't we using the cached one?

jacob-ebey (approved):

packages/remix-dev/compiler/assets.ts
pcattori: We are. I renamed `getRouteModuleExportsCached` to just `getRouteModuleExports` since we only ever use the cached version.

PR #4413, opened 2022-10-25
fix(remix-react): `Form`'s `method` shouldn't be ignored

maxschwarzmueller:
Fixes #4412


integration/playwright-report/index.html
```diff
@@ -0,0 +1,3 @@
...bject.assign(r,{versionMadeBy:w,msDosCompatible:p,compressedSize:0,uncompressedSize:0,commentLength:ft(u,l+32),directory:p&&16==(16&ut(u,l+38)),offset:ht(u,l+42)+c,internalFileAttribute:ht(u,l+34),externalFileAttribute:ht(u,l+38),rawFilename:d.subarray(s,h),filenameUTF8:a,commentUTF8:a,rawExtraField:d.subarray(h,_)});const b=_+r.commentLength;r.rawComment=d.subarray(_,b);const g=ct(t,e,"filenameEncoding"),y=ct(t,e,"commentEncoding"),[x,m]=await Promise.all([Pe(r.rawFilename,r.filenameUTF8?tt:g||nt),Pe(r.rawComment,r.commentUTF8?tt:y||nt)]);r.filename=x,r.comment=m,!r.directory&&r.filename.endsWith("/")&&(r.directory=!0),await st(r,r,u,l+6);const k=new Ve(r);if(k.getData=(e,t)=>r.getData(e,k,t),f.push(k),l=b,e.onprogress)try{e.onprogress(i+1,o,new Ve(r))}catch(e){}}return f}async close(){}},e.configure=m,e.getMimeType=function(){return"application/octet-stream"},Object.defineProperty(e,"__esModule",{value:!0})},"object"==typeof exports&&"undefined"!=typeof module?t(exports):"function"==typeof define&&define.amd?define(["exports"],t):t((e="undefined"!=typeof globalThis?globalThis:e||self).zip={})})();</script><script>/*! For license information please see app.bundle.js.LICENSE.txt */
```
machour, line 1: This needs to be removed

machour (changes requested):
Thank you so much for the reproduction and fix @maxschwarzmueller !
Could you remove the generated file?
I'll get eyes on this PR today hopefully

integration/playwright-report/index.html
maxschwarzmueller: Done

maxschwarzmueller:
I'm aware of the `<form>` default and spec. But Remix documentation clearly states that Remix `<Form>` supports `DELETE`, `PATCH` etc. in addition to `GET`& `POST`: https://remix.run/docs/en/v1/api/remix#form-method

chaance:
We definitely shouldn't break existing behavior, so limiting `<Form>` to `GET` and `POST` would need to be considered in a future major release. Probably worth opening a discussion for that if that's something you feel strongly about, as we have no plans for that at the moment. This is an intentional deviation from the basic `<form>` as noted in the docs. The no-JS tests should reflect those differences for now.

MichaelDeBoey (approved):

chaance:
> As for other methods, those are great points on the docs and breaking changes. I wonder if we should more strongly discourage their usage (e.g. in dev mode do a console.warn)? One of Remix' big selling points is that it works before/without JavaScript, and this violates that.

Would you like to start a new discussion on that? I think there's merit and I'd be happy to run it up the flag pole (aka get @ryanflorence to weigh in since his GH notifications are a tire fire 😂)

jenseng:
Sure thing, will do 👍

jenseng:
Opened a discussion [here](https://github.com/remix-run/remix/discussions/4420)

chaance (approved):

chaance:
Should be fixed in `1.7.4-pre.0`. We'll get a stable release out tomorrow.

PR #4417, opened 2022-10-25
docs: use `intent` instead of `action` or hidden input

MichaelDeBoey:
@kentcdodds switched to `intent` in his learning materials now, as this reduces overloading of the term `action`

### DIFF
--- a/docs/pages/faq.md
+++ b/docs/pages/faq.md
@@ -84,8 +84,8 @@ HTML buttons can send a value, so it's the easiest way to implement this:
 ```jsx filename=app/routes/projects/$id.jsx lines=[3-4,33,39]
 export async function action({ request }) {
   let formData = await request.formData();
-  let action = formData.get("action");
-  switch (action) {
+  let intent = formData.get("intent");
+  switch (intent) {
     case "update": {
       // do your update
       return updateProjectName(formData.get("name"));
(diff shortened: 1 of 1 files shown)


kentcdodds (approved):
LGTM

docs/pages/faq.md
```diff
@@ -129,15 +129,6 @@ export default function Projects() {
...
 
-You can also use a hidden input field:
-
-```jsx lines=[2]
-<Form method="post">
-  <input type="hidden" name="action" value="create" />
-  <button type="submit">Create</button>
-</Form>
-```
-
```
machour, line 140: I think this should be kept (and changed to intent), for educational purpose

MichaelDeBoey: People could still use hidden inputs for this if they want to, but I think we should point people to using this best practice (using `button`'s `name`) instead

machour: Educational purpose: some devs may not know about the hidden input, and this is a chance to show it to them.

IMHO, using a button with name, is not a better practice than using a hidden input along with a nameless button. They both achieve the same thing, while using the platform (🥔 vs 🥔)

sergiodxa: IMO, if there’s a single button in the form a hidden input it’s simpler

But if there are two buttons adding the name and value to each button is better

kentcdodds: Can you help me understand why you believe a hidden input is simpler? I feel the opposite 😅 the only time I'd use a hidden input is if I need to communicate something that's dynamic like the ID of the item being deleted or something. "Intent" just makes a lot of sense as something you put on a submit button.

MichaelDeBoey: I'm also in favor of only using the `button`'s `name`, as that makes it a consistent way of doing things, so no need to figure out if we have multiple buttons or not
That's more time to think about the important stuff

machour: Ok. Again, the thing for me is that it was interesting to still show the developer how to achieve the same thing with a hidden input, as this doesn't change **anything** and the example was already here in the first place.

There are more important hills to die on, so I'm gonna dismiss this review, and let you do as you like 😅

kentcdodds: I agree we should have an example in the docs showing how to use hidden fields. I'm not sure where it should be, but it's an important concept for sure.

MichaelDeBoey: I agree having a hidden input example in the docs can be beneficial, but I think we should have one with for instance an `id` of the deleted item, not for setting the `intent`

mcansh (approved):

PR #4421, opened 2022-10-25
fix(remix-dev): adjust conventional routes logic for pathless layout routes

brophdawg11:
### tl;dr; 

This fixes 2 issues ion our conventional routes logic:

* Removes the trailing slash from all generated `path` values since # of slash-delimited segments counts towards route ranking so the trailing slash incorrectly increases the scoring for routes.
* Removes pathless layout routes from the unique route path checks since they inherently trigger duplicate paths.  This allows us to support sibling pathless layouts.

Closes #3014 #3327

### Deets

#4355 had an integration test failure once merged to dev due to conflicts with an unrelated PR that merged an integration tests with an index file inside a pathless layout route folder.  This uncovered a bit of a rabbit hole into some details around our path collision detection in `conventionalRoutes.ts`.

Before #4355, index routes inside pathless layout routes were fine because they had a trailing slash hanging around on the "path", so our `uniqueRouteIds` structure looked like:

```
{
  'parent': 'routes/parent.tsx',
  'parent/': 'routes/parent/__pathless.tsx',
  'parent?index': 'routes/parent/__pathless/index.tsx'
}
```

Once we got rid of the trailing slash in #4355 (to fix #3014) we now start failing the build on the above since the "path" (the key in the Map) is identical for both matches.

This turns out to also be the reason that sibling pathless layout routes fail - even though the following is valid in React Router:

```jsx
<Routes>
  <Route file="root.tsx">
    <Route path="parent" file="routes/parent.jsx">
      <Route file="routes/parent/__a.jsx">
        <Route path="a-child" file="routes/parent/__a/a-child.jsx" />
      </Route>
      <Route file="routes/parent/__b.jsx">
        <Route path="b-child" file="routes/parent/__b/b-child.jsx" />
      </Route>
    </Route>
  </Route>
</Routes>
```

With the current logic, we end up reporting a path collision on `/parent` between `routes/parent/__a.jsx` and `routes/parent/__b.jsx` even though they have mutually exclusive children routes.

What I ended up doing in here is just ignoring path collision detection _only for pathless-layout route files_, since by definition they sort of introduce potential path collisions that the user should be aware of and accounting for (we even [warn them in the docs!](https://remix.run/docs/en/v1/api/conventions#pathless-layout-routes)).  We still check for collisions on any children route files inside a `__pathless/` directory - which lets us detect if they include non-mutually-exclusive route paths _beneath_ a pathless layout route folder.

So at the end of the day we get 2 things from this PR:
* #3014 is fixed and we correctly route to index routes next to pathless layout routes ✅
* We also get some _new_ functionality and can support sibling pathless layout routes 😄


packages/remix-dev/config/routesConvention.ts
```diff
@@ -188,6 +240,8 @@ export function createRoutePath(partialRouteId: string): string | undefined {
 
   if (rawSegmentBuffer === "index" && result.endsWith("index")) {
     result = result.replace(/\/?index$/, "");
+  } else {
+    result = result.replace(/\/$/, "");
```
brophdawg11, line 244: Align route paths so we never include trailing slashes on `<Route path>` entries, which messes with ranking since /-delimited segments count in ranking

brophdawg11: Fixes #3014

packages/remix-dev/config/routesConvention.ts
```diff
@@ -69,11 +69,63 @@ export function defineConventionalRoutes(
...
+       *
+       *   routes/parent/__pathless/foo.tsx
+       *   routes/parent/__pathless2/foo.tsx
+       *
+       * and
+       *
+       *   routes/parent/__pathless/index.tsx
+       *   routes/parent/__pathless2/index.tsx
+       */
+      if (uniqueRouteId && !isPathlessLayoutRoute) {
```
brophdawg11, line 123: Fixes #3327

.changeset/stale-spoons-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Avoid collisions between pathless layout routes and their nested index routes
```
chaance, line 5: Could you add some of the context here that you included in the PR description? That'll help a ton when writing release notes.

chaance (dismissed):
This looks awesome. Just a minor request for the changeset. Also, should we add similar tests for flat route conventions?

brophdawg11:
yeah I can look into the same tests/logic for flat routes 👍

sixers:
Hey, any update on this? I have to patch-package and apply this hotfix each time I upgrade Remix. It would be awesome to finally merge this fix.

sixers:
@brophdawg11 Yes, it's needed, the issue also exists in V2 route convention.

chaance:
@brophdawg11 Just so it's on your radar, you probably want to keep the v1 compat package in sync with these kind of fixes. https://github.com/remix-run/v1-compat-utils

brophdawg11:
Added PR for `@remix-run/v1-route-convention` in https://github.com/remix-run/v1-compat-utils/pull/20

jacob-ebey (approved):

PR #4474, opened 2022-11-01
fix(remix-react): fix submitter serialization

jenseng:
**Note: I have two proposals on how we can fix the submitter serialization bugs. The other approach is [here](https://github.com/remix-run/remix/pull/4475).**

Bring `<Form>` submissions in line with the spec with respect to how and where form submitters are serialized within the data set. 

**Solution:** We accomplish this by temporarily tweaking the form during submission to get the right entries.

- [x] Tests

Problems fixed:
1. Serialize submitters in tree order (i.e. where they appear in the DOM)
2. Serialize Image Button submitter correctly (i.e. separate x and y coordinate entries, rather than a single empty entry)
3. Stop sending multiple submitter entries in older WebKit

References: #4342
Spec: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1201,12 +1201,7 @@ export function useSubmitImpl(key?: string): SubmitFunction {
           target.getAttribute("formenctype") ||
           form.getAttribute("enctype") ||
           defaultEncType;
-        formData = new FormData(form);
-
-        // Include name + value from a <button>
-        if (target.name) {
-          formData.append(target.name, target.value);
-        }
+        formData = buildFormData(form, target);
       } else {
         if (isHtmlElement(target)) {
           throw new Error(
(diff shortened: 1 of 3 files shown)


packages/remix-react/components.tsx
```diff
@@ -1298,6 +1293,69 @@ export function useSubmitImpl(key?: string): SubmitFunction {
...
+    }
+  });
+}
+
+/**
+ * Build the form data set
+ *
+ * FormData doesn't (yet) have a submitter-aware constructor -- see https://github.com/whatwg/xhr/issues/262
+ *
+ * In the meantime, we can temporarily tweak the form during to ensure the data set adheres to the spec:
```
jenseng, line 1315: s/during/during submission/

jenseng:
Closing in favor of https://github.com/remix-run/react-router/pull/9865, since the code has moved there

PR #4475, opened 2022-11-01
fix(remix-react): fix submitter serialization (alt approach)

jenseng:
**Note: I have two proposals on how we can fix the submitter serialization bugs. The other approach is [here](https://github.com/remix-run/remix/pull/4474).**

Bring `<Form>` submissions in line with the spec with respect to how and where form submitters are serialized within the data set. 

**Alt solution:** We accomplish this by constructing the form data set ourselves according to the spec.

- [x] Tests

Problems fixed:
1. Serialize submitters in tree order (i.e. where they appear in the DOM)
2. Serialize Image Button submitter correctly (i.e. separate x and y coordinate entries, rather than a single empty entry)
3. Stop sending multiple submitter entries in older WebKit

References: #4342
Spec: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1201,12 +1201,7 @@ export function useSubmitImpl(key?: string): SubmitFunction {
           target.getAttribute("formenctype") ||
           form.getAttribute("enctype") ||
           defaultEncType;
-        formData = new FormData(form);
-
-        // Include name + value from a <button>
-        if (target.name) {
-          formData.append(target.name, target.value);
-        }
+        formData = buildFormData(form, target);
       } else {
         if (isHtmlElement(target)) {
           throw new Error(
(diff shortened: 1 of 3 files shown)


jenseng:
This feels a tad less hacky than the [other approach](https://github.com/remix-run/remix/pull/4474), but it's a lot more code to maintain, and as the spec evolves we'll have to keep tweaking the implementation, increasing the likelihood of bugs/etc.

packages/remix-react/components.tsx
```diff
@@ -1298,6 +1293,182 @@ export function useSubmitImpl(key?: string): SubmitFunction {
...
+      field[SELECTED_COORDINATE]
+    ) {
+      let prefix = field.name ? `${field.name}.` : "";
+      let { x, y } = field[SELECTED_COORDINATE];
+      entryList.push([`${prefix}x`, String(x)]);
+      entryList.push([`${prefix}y`, String(y)]);
+      continue;
+    }
+
+    //  3. If the field is a form-associated custom element, then perform the entry construction algorithm given field and entry list, then continue.
```
jenseng, line 1404: oops, forgot about this line... might need to see if this is even doable, otherwise add an implementation note here

packages/remix-react/components.tsx
```diff
@@ -1298,6 +1293,182 @@ export function useSubmitImpl(key?: string): SubmitFunction {
...
+    else if (isInputElement(field) && field.type === "file") {
+      if (field.files?.length) {
+        for (let file of field.files) {
+          entryList.push([name, file]);
+        }
+      } else {
+        entryList.push([name, ""]);
+      }
+    }
+    //  9. Otherwise, if the field element is an input element whose type attribute is in the Hidden state and name is an ASCII case-insensitive match for "_charset_":
```
jenseng, line 1437: missed this too 😅  ... i'm not actually sure if browser implement this though, should verify

packages/remix-react/components.tsx
jenseng: yeah this doesn't seem doable, added a comment here with more context

packages/remix-react/components.tsx
jenseng: implemented

jenseng:
Closing in favor of https://github.com/remix-run/react-router/pull/9866, since the code has moved there

PR #4484, opened 2022-11-01
docs(tutorials/jokes): Fix Prisma Client v4.5 seed example code

ivanjuric:
Running a command `npx prisma db seed` in Authentication section resulted with Prisma Client error.
The fix is creating the user with jokes using the create() API from Prisma Client actual version 4.5.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -169,6 +169,7 @@
 - ishan-me
 - IshanKBG
 - itsMapleLeaf
+- ivanjuric
 - jacargentina
 - jacob-ebey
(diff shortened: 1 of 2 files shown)


machour:
Could you share the error message you originally got? I'm surprised the original code don't work anymore

ivanjuric:
> Could you share the error message you originally got? I'm surprised the original code don't work anymore

Hmm, it seems I messed something with my database before executing the seed command. Now I tried again, and it works without any problems. Sorry for unnecessary PR.

PR #4496, opened 2022-11-02
feat(remix-react): emulate put/patch/delete when JS is unavailable

jenseng:
References #4420

- [x] Docs
- [x] Tests

Ensure `<Form method=...>` behaves the same with and without JavaScript. Although native forms don't support PUT/PATCH/DELETE, we can emulate them by doing a POST and injecting a `_method` parameter into the action URL. The Remix server runtime then provides action handlers a `Request` with the overridden `method` so they are none the wiser.

This provides a more reliable way to use these methods, which can be more ergonomic to write and handle than hidden inputs or buttons with values.

Note:
The emulation only works for action requests handled by Remix; if the form is submitted to another endpoint, it would need to handle the `_method` URL parameter accordingly.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1028,11 +1028,13 @@ function dedupe(array: any[]) {
 
 export interface FormProps extends FormHTMLAttributes<HTMLFormElement> {
   /**
-   * The HTTP verb to use when the form is submit. Supports "get", "post",
+   * The HTTP verb to use when the form is submitted. Supports "get", "post",
    * "put", "delete", "patch".
    *
-   * Note: If JavaScript is disabled, you'll need to implement your own "method
-   * override" to support more than just GET and POST.
+   * Note: When JavaScript is unavailable, Remix emulates put/patch/delete by
+   * injecting a `_method` parameter into the form action URL and doing a POST.
+   * The server-side handles this transparently, so your actions can always
+   * count on `request.method` matching `<Form method=...>`.
    */
   method?: FormMethod;
 
@@ -1103,7 +1105,7 @@ let FormImpl = React.forwardRef<HTMLFormElement, FormImplProps>(
     let submit = useSubmitImpl(fetchKey);
     let formMethod: FormMethod =
       method.toLowerCase() === "get" ? "get" : "post";
-    let formAction = useFormAction(action);
+    let formAction = useFormAction(action, method);
 
     return (
       <form
@@ -1154,7 +1156,6 @@ type HTMLFormSubmitter = HTMLButtonElement | HTMLInputElement;
  */
 export function useFormAction(
   action?: string,
-  // TODO: Remove method param in v2 as it's no longer needed and is a breaking change
   method: FormMethod = "get"
 ): string {
   let { id } = useRemixRouteContext();
@@ -1168,21 +1169,28 @@ export function useFormAction(
   let location = useLocation();
   let { search, hash } = resolvedPath;
   let isIndexRoute = id.endsWith("/index");
+  let params = new URLSearchParams(search);
 
   if (action == null) {
     search = location.search;
     hash = location.hash;
+    params = new URLSearchParams(search);
 
     // When grabbing search params from the URL, remove the automatically
     // inserted ?index param so we match the useResolvedPath search behavior
     // which would not include ?index
     if (isIndexRoute) {
-      let params = new URLSearchParams(search);
       params.delete("index");
-      search = params.toString() ? `?${params.toString()}` : "";
     }
   }
 
+  // Emulate additional methods when JavaScript is missing/disabled/pending
+  method = method.toLowerCase() as FormMethod;
+  if (["put", "patch", "delete"].includes(method)) {
+    params.set("_method", method);
+  }
+
+  search = params.toString() ? `?${params.toString()}` : "";
   if ((action == null || action === ".") && isIndexRoute) {
     search = search ? search.replace(/^\?/, "?index&") : "?index";
   }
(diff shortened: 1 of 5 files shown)


packages/remix-react/components.tsx
```diff
@@ -1049,21 +1048,28 @@ export function useFormAction(
...
     }
   }
 
+  // Emulate additional methods when JavaScript is missing/disabled/pending
+  method = method.toLowerCase() as FormMethod;
+  if (["put", "patch", "delete"].includes(method)) {
+    params.set("_method", method);
+  }
+
+  search = params.toString() ? `?${params.toString()}` : "";
```
sergiodxa, line 1072: Is this also affecting the fetcher.Form component? AFAIK the fetcher.Form works without JS too since it renders a normal form tag

jenseng: Yes, `fetcher.Form` and `Form` are backed by the same `FormImpl`+`useFormAction`

sergiodxa (approved):
Love this, it makes other HTTP methods more reliable to use if you want to support a no-JS scenario

machour (approved):
Awesome!

(Although generally speaking, maybe we should start name-spacing remix special parameters like `_remix_method`, `_remix_data` ..)

packages/remix-server-runtime/server.ts
```diff
@@ -49,6 +49,15 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
       );
     }
 
+    let overrideMethod = url.searchParams.get("_method") ?? "";
+    if (["put", "patch", "delete"].includes(overrideMethod)) {
+      url.searchParams.delete("_method");
+      request = new Request(url.toString(), {
+        ...request,
+        method: overrideMethod,
+      });
```
edmundhung, line 58: I like the ideas in general. But this might not work as expected.
Check this [StackBlitz](https://stackblitz.com/edit/js-rfla3u?devToolsHeight=33&file=index.js). 

What if we keep the `_method` on the request url similar to how remix handles `_data` currently and just override the method?

```ts
request = new Request(request, {
    method: overrideMethod,
});
```

jenseng: Oh yikes, good catch! I'll tweak this 👍

jenseng: Fixed ... expanded the tests in the process, which uncovered another issue 😅. Also moved the server logic to live alongside related stuff in data.ts

znycheporuk:
Is it possible to make it work with `<button formmethod="put | delete" />`?

jenseng:
@znycheporuk there are a couple big hurdles:

1. Browsers don't support setting the other form methods on a button. When you do `<button ...`, React is essentially running `document.createElement("button")` and then setting attributes on it. If you set a `formmethod` value other than `get` or `put`, the browser silently ignores it, which means we have no way of getting it later.
2. Even if we **could** get the other form methods to stick on the button element somehow, when the form is submitted without/before JavaScript, it would not get used because browsers don't support that at submission time.

It may be possible to emulate it a different way, but that would require more plumbing and not be worth the trouble. One idea might be for remix to provide a `Button` component along these lines:

```js
function Button({ formmethod, name, value, ...props }) {
  if (["put", "patch", "delete"].includes(formmethod)) {
    name = "_method"
    value = formmethod
  }
  return <button name={name} value={value} {...props} />
}
```

A downside is that you couldn't use an explicit name+value in conjunction with such a `formmethod`, since they would be overridden. Additionally, the server-side `_method` logic would need to be adjusted to also look in `formData` (multipart or url-encoded), which would make the code more complex and may have performance implications.

I suspect it's probably not worth the trouble, and there may be a much simpler alternative if you want to associate a custom form method with a specific button: use multiple forms :) ... for example, instead of this:

```jsx
<Form>
  <input name="todoItem" defaultValue={myTodo} />
  <button formmethod="put">Update</button>
  <button formmethod="delete">Delete</button>
</Form>
```

You could structure your elements like so:

```jsx
<Form method="put">
  <input name="todoItem" defaultValue={myTodo} />
  <button>Update</button>
<Form>
<Form method="delete">
  <button>Delete</button>
</Form>
```

Or if for some reason you really want the buttons in the same form (e.g. to facilitate styling), you can associate it with another form:

```jsx
<Form method="put">
  <input name="todoItem" defaultValue={myTodo} />
  <button>Update</button>
  <button form="delete-form">Delete</button>
<Form>
<Form method="delete" style={{ display: "none" }} id="delete-form"></Form>
```

If neither of those approaches work or your use case (e.g. you need all buttons to submit all the form data), then I'd suggest instead using buttons with a name+value like `intent="update"`, and keying off of that on the server side.

MichaelDeBoey:
As explained in https://github.com/remix-run/remix/discussions/4420#discussioncomment-4412645, I don't think it's a good idea to make non-JS & JS work the same in this case

> I'm personally in favor of dropping non-GET/POST tbh, as that makes us work closer to the spec.
> Although it's maybe not that much discouraged in [the current `Form.method` docs](https://remix.run/components/form#method), we did make sure all other docs use GET/POST & set the submit `button`'s `name` to `'intent'` if necessary not that long ago.
> 
> Not discouraging using non-GET/POST would make people also want to use `<button formmethod="delete|patch|put" />`, which is something we'll **never** be able to support without a custom `Button` component on our end & some crazy logic on our end (as perfectly explained by @jenseng above & in https://github.com/remix-run/remix/pull/4496#issuecomment-1320222238)
> We have `<button name="intent" value="update">...</button>` to support these kind of things anyways, so I would rather make sure the docs are more pushing towards that instead tbh.
> 
> So what I think we should do is:
> - Keep the current support for non-GET/POST, but add a deprecation warning when people are using them
> - Deprecate non-GET/POST in v2
> 
> I agree that this would mean non-JS isn't working the same as when JS is enabled, which is something we should fix in most cases.
> I do however think we shouldn't fix it in this case, as that would encourage using non-GET/POST more, which is (as I explained above) not a good thing imo.

jenseng:
Happy to rebase/revisit this if we can get consensus from the Remix team... do we 1. want to move forward with emulation or 2. just want to deprecate those methods in v2 (or later?)

I think there are good arguments to be made for and against either approach. My personal preference is emulation, but either option would be preferable to the inconsistent status quo.

PR #4515, opened 2022-11-04
fix(remix-react): improve fetcher error message

jmlee2k:
Adding `href` to the "no matches" fetcher error message.

Testing Strategy:

built locally and tested against my remix app


jmlee2k:
Sorry, I made a mess of this, I'll open a new PR.

Issue #4236, opened 2022-09-20
Uncaught Error: Dynamic require of "fs/promises" is not supported - when importing json from @remix-run/node


meglio:
### What version of Remix are you using?

1.7.1

### Steps to Reproduce

```jsx
import { json } from "@remix-run/node";

export default function SomePage(props) {
    return <div>
        Page
    </div>
}
```

Then `yarn dev`.

### Expected Behavior

No errors in developer console.

### Actual Behavior

Error in developer console:

![image](https://user-images.githubusercontent.com/282177/191274209-88932318-4bf0-4f9d-a9d3-b3ae7365e3d7.png)

### package.json

```json
{
  "private": true,
  "sideEffects": false,
  "scripts": {
    "build": "remix build",
    "dev": "remix dev",
    "start": "remix-serve build"
  },
  "dependencies": {
    "@remix-run/node": "^1.7.1",
    "@remix-run/react": "^1.7.1",
    "@remix-run/serve": "^1.7.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0"
  },
  "devDependencies": {
    "@remix-run/dev": "^1.7.1",
    "@remix-run/eslint-config": "^1.7.1",
    "eslint": "^8.20.0"
  },
  "engines": {
    "node": ">=14"
  },
  "packageManager": "yarn@3.2.3"
}


```

### Yarn and Node

* node version: 18.9.0
* Yarn: 3.x

NB. I tried creating the project with using Yarn 1.x and it worked fine. The problem appears after migrating to yarn 3.x

Might be related:

https://github.com/yarnpkg/berry/issues/638#issuecomment-1124629341

Looks like Yarn 3.x supports ESM module loading.  So I tried the following in my `.yarnrc.yml` file, following the recommendations from the link above:

```yml
yarnPath: .yarn/releases/yarn-3.2.3.cjs
pnpEnableEsmLoader: true
nodeLinker: "pnp"
```

But this did not help.

So, the thread above makes me think that yarn 3.x should work with Remix. But it does not. I'm wondering if this is Remix or Yarn issue?

The only workaround that works for me is to rollback to using `nodeLinker: node-modules`.

mwood23:
Just hit this. You need to import things from `@remix-run/cloudflare` instead

meglio:
@mwood23 I am not using cloudflare.

mkikets99:
> @mwood23 I am not using cloudflare.

I hit this too...
[...]

PR #4524, opened 2022-11-05
fix(remix-node): refactor fs promises import

NRodriguezcuellar:
Problem described here https://github.com/remix-run/remix/issues/4236

closes https://github.com/remix-run/remix/issues/4236

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -450,3 +450,4 @@
 - zachdtaylor
 - zainfathoni
 - zhe
+- NRodriguezcuellar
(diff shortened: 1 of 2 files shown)


sergeyram (approved):

pcattori:
As I [mentioned here](https://github.com/remix-run/remix/issues/4236#issuecomment-1332512178), Remix requires Node v14 or higher so we should be able to use `fs/promises` internally.

The root cause of #4236 is that server code is erroneously bundled into the browser bundle. So the fix should address that directly.

Issue #4480, opened 2022-11-01
Delayed startup due to slow route generation


dmarkow:
### What version of Remix are you using?

1.7.4

### Steps to Reproduce

See [sample repo](https://github.com/dmarkow/remix-route-performance-example). This sample app has a patch on the `routesConvention.js` file to display run time for route generation, as well as a `run.js` script that generates a bunch of routes and then runs `remix routes`. Pass an optional number for the number of route sets to create (each set has 111 routes between index/child routes). For example, `node run.js 5` will profile with 5 sets of routes (556 routes including the root route). I nested the sample routes under a couple pathless routes as that seems to also decrease the performance.

```
# Install and run
npm i
npx patch-package
node run.js 5 | head
```

### Expected Behavior

Ideally, performance would decrease linearly as the number of routes increases.

### Actual Behavior

Exponential increase in the time it takes to run. On a fairly powerful Mac, 1 set (112 routes) takes 0.042 seconds to compile, 5 sets (556 routes) takes 3.413 seconds, and 10 sets (1,111 routes) takes 26.797 seconds.

```
$ node run.js 1 | head
Creating sample routes...
defineConventionalRoutes: 0.042s for 112 routes (0.38ms per file)
<Routes>
  <Route file="root.tsx">
    <Route path="0/10/0/0" index file="routes/__a/__b/__c/0/10/0/0/index.js" />
    <Route path="0/10/0/1" index file="routes/__a/__b/__c/0/10/0/1/index.js" />
    <Route path="0/10/1/0" index file="routes/__a/__b/__c/0/10/1/0/index.js" />
    <Route path="0/10/1/1" index file="routes/__a/__b/__c/0/10/1/1/index.js" />
    <Route path="0/10/2/0" index file="routes/__a/__b/__c/0/10/2/0/index.js" />
    <Route path="0/10/2/1" index file="routes/__a/__b/__c/0/10/2/1/index.js" />

$ node run.js 5 | head
Creating sample routes...
defineConventionalRoutes: 3.413s for 556 routes (6.14ms per file)
<Routes>
  <Route file="root.tsx">
    <Route path="0/10/0/0" index file="routes/__a/__b/__c/0/10/0/0/index.js" />
    <Route path="0/10/0/1" index file="routes/__a/__b/__c/0/10/0/1/index.js" />
    <Route path="0/10/1/0" index file="routes/__a/__b/__c/0/10/1/0/index.js" />
[...]

PR #4538, opened 2022-11-07
perf(remix-dev): Optimize `parentRouteId` lookup in `defineConventionalRoutes`

dmarkow:
Closes: #4480. `defineConventionalRoutes` was repeatedly running `.filter` and `.find` against the list of routeIds to find parents, which was causing an exponential slowdown in larger projects. This update first generates a `parentRouteIds` object and then filters against that instead.

- [ ] Docs
- [x] Tests

Testing Strategy: There weren't any direct tests for `defineConventionalRoutes` as far as I could tell, so I added one to make sure the necessary number of routes (including parent route ids) are generated. I used an existing fixture for the routes folders (`remix-dev/__tests__/fixtures/replace-remix-imports`) but I can add one specific to the `routesConventions` tests if you prefer this test remain uncoupled from the replace imports test.

I ran this against my production app with nearly 700 routes and the RouteManifest object was was identical before and after the patch. Before this update, my production app was taking around 10-15 seconds before "Watching Remix app in development mode..." showed up, now it's less than a second.

I also ran the [example test app from my original issue](https://github.com/dmarkow/remix-route-performance-example) against this new code and the speed up is drastic on large projects (almost 27 seconds before for 1,111 routes, now it's 0.104 seconds).

```
$ node run.js 1 | head
Creating sample routes...
defineConventionalRoutes: 0.008s for 112 routes (0.07ms per file)
<Routes>
  <Route file="root.tsx">
    <Route path="0/10/0/0" index file="routes/__a/__b/__c/0/10/0/0/index.js" />
    <Route path="0/10/0/1" index file="routes/__a/__b/__c/0/10/0/1/index.js" />
    <Route path="0/10/1/0" index file="routes/__a/__b/__c/0/10/1/0/index.js" />
    <Route path="0/10/1/1" index file="routes/__a/__b/__c/0/10/1/1/index.js" />
    <Route path="0/10/2/0" index file="routes/__a/__b/__c/0/10/2/0/index.js" />
    <Route path="0/10/2/1" index file="routes/__a/__b/__c/0/10/2/1/index.js" />

$ node run.js 5 | head
Creating sample routes...
defineConventionalRoutes: 0.044s for 556 routes (0.08ms per file)
<Routes>
  <Route file="root.tsx">
    <Route path="0/10/0/0" index file="routes/__a/__b/__c/0/10/0/0/index.js" />
    <Route path="0/10/0/1" index file="routes/__a/__b/__c/0/10/0/1/index.js" />
    <Route path="0/10/1/0" index file="routes/__a/__b/__c/0/10/1/0/index.js" />
    <Route path="0/10/1/1" index file="routes/__a/__b/__c/0/10/1/1/index.js" />
    <Route path="0/10/2/0" index file="routes/__a/__b/__c/0/10/2/0/index.js" />
    <Route path="0/10/2/1" index file="routes/__a/__b/__c/0/10/2/1/index.js" />

$ node run.js 10 | head
Creating sample routes...
defineConventionalRoutes: 0.104s for 1111 routes (0.09ms per file)
<Routes>
  <Route file="root.tsx">
    <Route path="0/10/0/0" index file="routes/__a/__b/__c/0/10/0/0/index.js" />
    <Route path="0/10/0/1" index file="routes/__a/__b/__c/0/10/0/1/index.js" />
    <Route path="0/10/1/0" index file="routes/__a/__b/__c/0/10/1/0/index.js" />
    <Route path="0/10/1/1" index file="routes/__a/__b/__c/0/10/1/1/index.js" />
    <Route path="0/10/2/0" index file="routes/__a/__b/__c/0/10/2/0/index.js" />
    <Route path="0/10/2/1" index file="routes/__a/__b/__c/0/10/2/1/index.js" />
```

### DIFF
--- a/packages/remix-dev/config/routesConvention.ts
+++ b/packages/remix-dev/config/routesConvention.ts
@@ -49,6 +49,10 @@ export function defineConventionalRoutes(
   });
 
   let routeIds = Object.keys(files).sort(byLongestFirst);
+  let parentRouteIds: { [routeId: string]: string | undefined } = {};
+  for (let id of routeIds) {
+    parentRouteIds[id] = findParentRouteId(routeIds, id);
+  }
 
   let uniqueRoutes = new Map<string, string>();
 
@@ -58,7 +62,7 @@ export function defineConventionalRoutes(
     parentId?: string
   ): void {
     let childRouteIds = routeIds.filter(
-      (id) => findParentRouteId(routeIds, id) === parentId
+      (id) => parentRouteIds[id] === parentId
     );
 
     for (let routeId of childRouteIds) {
@@ -86,7 +90,7 @@ export function defineConventionalRoutes(
 
       if (isIndexRoute) {
         let invalidChildRoutes = routeIds.filter(
-          (id) => findParentRouteId(routeIds, id) === routeId
+          (id) => parentRouteIds[id] === routeId
         );
 
         if (invalidChildRoutes.length > 0) {
--- a/packages/remix-dev/config/routesConvention.ts
+++ b/packages/remix-dev/config/routesConvention.ts
@@ -49,10 +49,7 @@ export function defineConventionalRoutes(
   });
 
   let routeIds = Object.keys(files).sort(byLongestFirst);
-  let parentRouteIds: { [routeId: string]: string | undefined } = {};
-  for (let id of routeIds) {
-    parentRouteIds[id] = findParentRouteId(routeIds, id);
-  }
+  let parentRouteIds = getParentRouteIds(routeIds);
 
   let uniqueRoutes = new Map<string, string>();
 
@@ -197,11 +194,16 @@ export function createRoutePath(partialRouteId: string): string | undefined {
   return result || undefined;
 }
 
-function findParentRouteId(
-  routeIds: string[],
-  childRouteId: string
-): string | undefined {
-  return routeIds.find((id) => childRouteId.startsWith(`${id}/`));
+function getParentRouteIds(
+  routeIds: string[]
+): Record<string, string | undefined> {
+  return routeIds.reduce<Record<string, string | undefined>>(
+    (parentRouteIds, childRouteId) => ({
+      ...parentRouteIds,
+      [childRouteId]: routeIds.find((id) => childRouteId.startsWith(`${id}/`)),
+    }),
+    {}
+  );
 }
 
 function byLongestFirst(a: string, b: string): number {
--- a/contributors.yml
+++ b/contributors.yml
@@ -102,6 +102,7 @@
 - dhargitai
 - dhmacs
 - dima-takoy
+- dmarkow
 - DNLHC
 - dogukanakkaya
 - dokeet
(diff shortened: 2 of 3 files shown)


dmarkow:
@machour My git rebasing skills are apparently trash 😅 Rebasing my local branch against dev is still keeping all of main's recent commits as well, and changing the base branch of this PR causes 11 different commits from main to pull over.

Can you point me in the right direction on this, or should I just start over and push a new PR originating from the dev branch?

dmarkow:
@machour Nevermind, figured out `rebase --onto` right after I replied. It should be ready to go now.

packages/remix-dev/config/routesConvention.ts
```diff
@@ -49,6 +49,10 @@ export function defineConventionalRoutes(
   });
 
   let routeIds = Object.keys(files).sort(byLongestFirst);
+  let parentRouteIds: { [routeId: string]: string | undefined } = {};
+  for (let id of routeIds) {
+    parentRouteIds[id] = findParentRouteId(routeIds, id);
```
pcattori, line 54: This is the only usage of `findParentRouteId` in the codebase, so it would be good to inline it here:

```suggestion
    parentRouteIds[id] = routeIds.find((id) => childRouteId.startsWith(`${id}/`));
```

Optionally, you could then factor out the `parentRouteIds` calculation into a `getParentRouteIds(routeIds)` function.

So this whole block would become:

```ts
let parentRouteIds = getParentRouteIds(routeIds)
```

pcattori (changes requested):

packages/remix-dev/config/routesConvention.ts
```diff
@@ -49,6 +49,10 @@ export function defineConventionalRoutes(
...
+  }
```
MichaelDeBoey, line 55: It would give the same result, but is there a reason you went for mutating an object instead of using a built-in array function like `reduce`?

Also: type could be `Record<string, string | undefined>`

```suggestion
  let parentRouteIds = routeIds.reduce<Record<string, string | undefined>>(
    (parentRouteIds, id) => ({
      ...parentRouteIds,
      [id]: findParentRouteId(routeIds, id),
    }),
    {}
  );
```

MichaelDeBoey (approved):

packages/remix-dev/config/routesConvention.ts
dmarkow: I had seen some other code in remix-dev when working on that PR that used the for..or / push approach so went with that, e.g. https://github.com/remix-run/remix/blob/606bc8229b200b1eb8edbf0f2367fbad41ab82df/packages/remix-dev/config/format.ts#L43-L44

I just updated with reduce.

packages/remix-dev/config/routesConvention.ts
dmarkow: Done, I pulled the whole thing into a separate function.

pcattori (approved):
LGTM

@dmarkow looks like some tests are failing. Could you rebase on latest `dev` and see if that fixes things?

dmarkow:
@pcattori Just fixed the tests (looks like an existing fixture changed paths)

pcattori:
@dmarkow thanks for your work! Should make a HUGE difference to many Remix users ❤️

Issue #4488, opened 2022-11-01
🗺 Testing Helpers


ryanflorence:
Will allow unit testing React components that use Remix APIs like `useFetcher` outside of a running Remix app.

Possibly depends on #4485, we'll see as we get into implementation.

## Progress

- [x] RFC #4541 
- [x] PR Remix: #4539
- [x] PR web-std-io: https://github.com/remix-run/web-std-io/pull/26

## Issues/PRs/Discussions

- Original discussion: https://github.com/remix-run/remix/discussions/2481
- PR: https://github.com/remix-run/remix/pull/4539

## Community implementation

https://github.com/jrestall/remix-stubs

PR #4539, opened 2022-11-07
feat: testing helpers

mcansh:
shout out to [@jrestall](https://github.com/jrestall/remix-stubs) for the initial plumbing, we were both working on this concurrently and came up with similar implementations for todays Remix on React Router 6.3 - I'm currently working on switching this over to use Remix on RR@6.4

Closes: #4488

- [ ] Docs
- [ ] Tests

Testing Strategy:

- https://github.com/ryanflorence/tmp-form-data-error
- https://github.com/mcansh/stubbbbbs


packages/remix-testing/create-remix-stub.tsx
```diff
@@ -0,0 +1,319 @@
...
+      params: match.params,
+      pathname: match.pathname,
+      route: convertToEntryRoute(match.route),
+    };
+  });
+}
+
+// Converts route data from a path based index to a route id index value.
+// e.g. { "/post/:postId": post } to { "0": post }
+function convertRouteData(
```
jrestall, line 307: I've updated this function in remix stubs to [support nested routes](https://github.com/jrestall/remix-stubs/blob/7eca61c202cf1eff29b7060b55d12f6586f437d7/packages/remix-stubs/src/create-remix-stub.tsx#L308). It'll need a similar change here. I also  used `'/'` to refer to route data for the pathless root route which I think is accurate but you guys will know best.

ryanflorence:
This is awesome! I made an experimental release to play with it, especially to see if the types and stuff are working well, etc.

A couple of things to look into before we merge:

### Types

`RouteObject` is requiring `element` but that shouldn't be required, like a resource routes. Pretty much every key on a route object should be optional.

```js
{
  path: "/users",
  loader: () => {
    return []
  },
},
```

^ That should be fine but TS yells at you for not having an `element`. I don't know if this is related to this work, or something in `@remix-run/router` that needs to lighten up.

### FormData problems

When testing in vitest with happydom or jsdom we have problems. I'm not completely sure what's going on but this doesn't work:

```jsx
      <fetcher.Form action="/users" method="get">
        <input
          type="text"
          name="search"
          onChange={(event) => {
            // this works in browsers but not happydom/jsdom
            fetcher.submit(event.target.form);
          }}
        />
      </fetcher.Form>
```

Both happydom and jsdom fail [right here in our fork of web-std-io](https://github.com/remix-run/web-std-io/blob/main/packages/form-data/src/form-data.js#L10-L16)

- `happydom` fails our `isFormElement` check because it returns `[object Object]` instead of `[object HTMLFormElement]`. We should open an issue with them.
- `jsdom` fails because we we don't support passing in a form element to `new FormData(form)`.

Rather than try to support this in our FormData implementation, I think we just need to figure out why the testing environment is even using our backend polyfill for FormData in the first place? It should be using the FormData from happydom/jsdom!

I use a stock Indie Stack, so we can look there to see what's going on.

Here's a repo showing the problem: https://github.com/ryanflorence/tmp-form-data-error

```
git clone
npm i
npm test
💣
```

While I know this problem isn't necessarily part of the work in this PR, I'd like to figure this out and fix the test harness in our stacks to make it work before we release so we have the solution from the start.

mcansh:
> RouteObject is requiring element but that shouldn't be required, like a resource routes. Pretty much every key on a route object should be optional.

that's the types in the stub, AgnosticRoutes don't have `element`, but once RR@6.4 is in here we can re-use that type. for now it's fixed here https://github.com/remix-run/remix/pull/4539/commits/2201c4739eb20673961043f2f6082d71b7ce2721

> happydom fails our isFormElement check because it returns [object Object] instead of [object HTMLFormElement].

strange https://github.com/capricorn86/happy-dom/blob/9c0f9b7a7b1363297cf3f4cb2199a430e2a74d1a/packages/happy-dom/test/nodes/html-form-element/HTMLFormElement.test.ts#L16-L20, but yeah looking into why it's not using jsdom's FormData (happydom doesn't have a complete one)

mcansh:
we're running `installGlobals` in `./test/setup-test-env.ts`

using this works, but i dont think we should need to...?

```ts
import "@testing-library/jest-dom";
import { JSDOM } from "jsdom";

installGlobals()

const jsdom = new JSDOM(`<!doctype html>`);
globalThis.FormData = jsdom.window.FormData;
```

mcansh:
removing the installGlobals causes a `Request is not defined` error, perhaps in `@remix-run/testing` we have our own `installGlobals`?


```ts
import { installGlobals as installNodeGlobals } from "@remix-run/node";

async function loadModule(modulePath: string) {
  try {
    return await import(modulePath)
  } catch {
    throw new Error(`Unable to import module ${modulePath}`)
  }
}

export async function installGlobals() {
  installNodeGlobals()
  let JSDOM = await loadModule('jsdom');
  let jsdom = new JSDOM.JSDOM(`<!doctype html>`);
  globalThis.FormData = jsdom.window.FormData;
}
```

ryanflorence:
@mcansh yeah let's do that, and if we think of anything better in the meantime, we'll change it when we remove the unstable prefix.

```js
import { unstable_installGlobals } from "@remix-run/testing";
unstable_installGlobals("jsdom");
```

mcansh:
sounds good! thats the exact public api i have for that haha

ryanflorence:
Wait, what if you're on node 18/19 with jsdom, do you even need to install any globals?

jrestall:
I've found that when using the jsdom environment on vitest as of version 0.25.3 the FormData workaround is no longer required due to this (https://github.com/vitest-dev/vitest/pull/2290) recent fix. Ryan your repro branch is using the older [0.24.5](https://github.com/ryanflorence/tmp-form-data-error/blob/002684fbb38e7cbc3d2096e08d9fcfe2221c4e71/package.json#L84).

This is great as jsdom setup is quite slow and importing/creating JSDOM again in each test setup after vitest already did it [here](https://github.com/vitest-dev/vitest/blob/a6a94ccc6da7dfbc46412bb007c817c1cb1eb482/packages/vitest/src/integrations/env/jsdom.ts#L27) would likely slow test setup. I haven't tested jest's jsdom environment though.

> Wait, what if you're on node 18/19 with jsdom, do you even need to install any globals?

I've found calling installGlobals in a jsdom environment breaks the dom tests due to overwriting FormData. So I have to run the node and jsdom tests separately which isn't ideal. I'd prefer if vitest passed the current environment to the setup files then I could install globals only when running in a node env for action/loader tests.

```ts
// maybe one day...
if (globals.environment === "node") installGlobals();
```

PR #4560, opened 2022-11-10
fix: check index routes against route config instead of id suffix

kiliman:
Index routes in Remix Flat Routes do not end in `/index` so these routes were not correctly identified.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1038,6 +1038,7 @@ export function useFormAction(
   // TODO: Remove method param in v2 as it's no longer needed and is a breaking change
   method: FormMethod = "get"
 ): string {
+  let { manifest } = useRemixEntryContext();
   let { id } = useRemixRouteContext();
   let resolvedPath = useResolvedPath(action ?? ".");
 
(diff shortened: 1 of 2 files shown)


kiliman:
@kentcdodds Yes, I'll write a test for this. I was just trying to at least get this on the radar.

In the meantime, I think I'll just update Remix Flat Routes to add the `/index` to the id for index routes. That way it's a non-blocker until this PR gets merged.

kiliman:
Ok, v0.4.8 has been published. This updates the route ids for index routes to match what Remix is looking for. At least until Remix merges my PR.

kentcdodds:
The benefit is to:
1. Demonstrate the bug
2. Make sure it never happens again

penspinner:
Hi. I added a PR back in May 7 that fixed this: https://github.com/remix-run/remix/pull/3127.

kiliman:
@penspinner Nice.. I didn't see that one. Cool that you have tests. 

BTW: it looks like there is one other place where it incorrectly checks for _/index_: 

https://github.com/remix-run/remix/pull/4560/files?diff=unified&w=1#diff-67cd998c708c10f278fa0382338e65a4d43e3fd67e4ed5bc984d13d1d1f9725f

penspinner:
@kiliman, thanks. Updated.

kiliman:
I'm going to close this PR in favor of #3127.

PR #4561, opened 2022-11-10
fix(remix-react): Remove nullish coalescing operator

machour:
Closes: #

- [ ] Docs
- [ ] Tests

### Testing Strategy:

I was testing my app with an iPhone 6, and found out the client side was totally broken.

We had a [discussion](https://github.com/remix-run/remix/discussions/3240#discussioncomment-4086437) with @sergiodxa a few days ago where we uncovered that we re-introduced a nullish coalescing operator `??` in https://github.com/remix-run/remix/pull/3697 that could be problematic for old browsers.

I can now confirm that it is indeed problematic. 🙃 
This PR gets rid of it.

### DIFF
--- a/.changeset/curvy-countries-boil.md
+++ b/.changeset/curvy-countries-boil.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Fixed old browsers compatibility by removing a nullish coalescing operator
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1039,7 +1039,7 @@ export function useFormAction(
   method: FormMethod = "get"
 ): string {
   let { id } = useRemixRouteContext();
-  let resolvedPath = useResolvedPath(action ?? ".");
+  let resolvedPath = useResolvedPath(action ? action : ".");
 
   // Previously we set the default action to ".". The problem with this is that
   // `useResolvedPath(".")` excludes search params and the hash of the resolved


MichaelDeBoey (commented):
I don't think we should make this change like this tbh.
We're supporting Node v14+, so nullish coalescing should be possible to write without any problems.

In case of the older browser being broken, this is a matter of broken transpilling, most likely not taking these version into account in `browserslist`

machour:
Will happily update the PR if there's a better way to do it. Could you set me on the right track to fix the transpilling part?

MichaelDeBoey:
@machour I think I once saw a discussion about supporting a `browserslist` config, because (if I correctly remember) we're currently only supporting the defaults provided by `esbuild`

machour:
Gotcha. Leaving this as it is for the meantime. It was working, it's not anymore, so we may want to fix this first, then look at changing the way we do things.

MichaelDeBoey:
If this only changed recently, this can be a bug introduced when @pcattori extracted browser=server compiler out of the massive compiler we had before. 🤔

Could be a totally different thing as well, I just know that we had some work done there 🤷‍♂️

sergiodxa (approved):
Until browserlist is supported I think this one line change it’s simpler and can help a lot of people support older browsers with ESM support

chaance:
I'm going to merge this for now, but we'll prioritize a strategy for dealing with this to prevent regressions over the next week.

pcattori:
> If this only changed recently, this can be a bug introduced when @pcattori extracted browser+server compiler out of the massive compiler we had before. 🤔

The compiler code never makes it into the browser, so nullish coalescing is fine in compiler / server-only code.
In this case, its in `useFormAction`, which _does_ make it into the browser. 

I agree that we should allow Node v14 features like nullish coalescing in our codebase and just transpile them away for the browser build.

PR #4566, opened 2022-11-10
feat(compiler): Add support for remix config feature flags

chaance:
Adds preliminary tests for feature flags in `remix.config`.

While this technically adds compiler support for the `v2_meta` feature flag, the actual implementation for that feature will come in a follow-up PR. I wanted this PR to include a feature to show how we plan to enable opt-in features.

- [ ] Docs
- [x] Tests


packages/remix-dev/__tests__/readConfig-test.ts
```diff
@@ -62,4 +68,13 @@ describe("readConfig", () => {
     let newConfig = await readConfig(remixRoot);
     expect(newConfig.devServerPort).toBe(config.devServerPort);
   });
+
+  it("includes feature flags", async () => {
```
chaance, line 72: This actually isn't necessary since we have the snapshot. Removing...

packages/remix-dev/index.ts
```diff
@@ -1,6 +1,6 @@
 import "./modules";
 
-export type { AppConfig } from "./config";
+export type { AppConfig, RemixConfig as ResolvedRemixConfig } from "./config";
```
chaance, line 3: We probably don't want users importing `RemixConfig` but it'll be useful for us internally as features expand. The `ResolvedRemixConfig` alias clarifies intent and makes it less likely for users to confuse it with `AppConfig`.

packages/remix-dev/compiler/compilerServer.ts
```diff
@@ -108,6 +108,9 @@ const createEsbuildConfig = (
       "process.env.REMIX_DEV_SERVER_WS_PORT": JSON.stringify(
         config.devServerPort
       ),
+      "process.env.REMIX_CONFIG_FEATURES_V2_META": JSON.stringify(
+        config.features.v2_meta
+      ),
```
brophdawg11, line 113: I don't think this covers unbundles CJS use-cases - I threw a `console.log("meta", process.env.REMIX_CONFIG_FEATURES_V2_META);` inside our `Meta` component and it logs `undefined` on the server and `true` on the client with the flag set.

Doing the same thing we do for the dev server port fixed it for my `remix dev` use case: https://github.com/remix-run/remix/blob/main/packages/remix-dev/config.ts#L401

That likely means we'll be mapping feature flag name -> env var name in a few spots so maybe we oughta define them in a new `features.ts` file and import from there wherever needed?

```js
export const features = {
  v2_meta: "REMIX_CONFIG_FEATURES_V2_META"
};
```

brophdawg11 (commented):
One small issue with unbundled server-side usage - but working well otherwise 👍

packages/remix-dev/compiler/compilerServer.ts
chaance: @brophdawg11 You wouldn't look expect a value in `meta` because it's not going to be defined at runtime. Instead we'd do something like:

```ts
function Meta_v1() { /* ... */ }
function Meta_v2() { /* ... */ }
export const Meta = process.env. REMIX_CONFIG_FEATURES_V2_META === "true" ? Meta_v2 : Meta_v1;
```

ryanflorence: Why are we using env vars at all? This is surprising to me. The build and the runtime both have access to the remix config, is there a reason we can't just read from there?

ryanflorence: Seems like we could put this on `window.__remixContext` for the browser instead of hacking it into a fake clientside `process.env`.

Server runtime has access to the config naturally, client gets access through hydration at runtime. Can you explore that instead?

brophdawg11: I assumed we didn't want to ship both old/new code flows in the client bundle, but if we don't mind that then I'm on board with runtime config reading

kiliman: I think it's ok to support both during the transition. Then at some point, remove the old code and focus on new.

Anyway, the RFC already proposes a `objectToMetaArray` and `mergeRouteMeta` to handle the old syntax. I would make it so that the old code simply uses these helpers to convert to the new syntax under the covers. That way there's only a single code path. 

That's basically what my PR did. It converted object to array, then the rest of the code only used the array.

packages/remix-dev/compiler/plugins/serverEntryModulePlugin.ts
```diff
@@ -45,6 +45,7 @@ ${Object.keys(config.routes)
   export const assetsBuildDirectory = ${JSON.stringify(
     config.relativeAssetsBuildDirectory
   )};
+  export const features = ${JSON.stringify(config.features)};
```
jacob-ebey, line 48: How are these going to be surfaced to the browser at runtime? Are they going to be serialized into the EntryContext? This is part of why I suggested the "read at build time and inject into globals" approach.

jacob-ebey: Ah yeah ok. Seems like I still prefer build time for this for a few reasons:
1) no increase in client bundle size as we can optimize out unused code paths
2) no increase in serialized context size. this also exposes internals about your site configuration to the public, something I'm not a fan of

packages/remix-dev/compiler/compilerServer.ts
chaance: We decided on a runtime check for now. This means we need to add some data to the entry context that we can access in both builds. Updated accordingly.

packages/remix-dev/compiler/plugins/serverEntryModulePlugin.ts
chaance: @jacob-ebey We discussed these tradeoffs, but @ryanflorence is afraid of overloading `process.env` when we're pushing towards supporting a zero-build environment. That's a larger conversation, but I deferred to the product manager for now 😅

ryanflorence: @brophdawg11 was on the call too. Sometimes a feature flag might only change one or two lines of behavior so no need to duplicate all of that in the function, but most importantly just want to keep everything "the same". It's a runtime thing in all places we need it: bundling, server, client.

Issue #4530, opened 2022-11-06
Possible Documentation Issue - watchPaths


kevinswarner:
### What version of Remix are you using?

1.7.5

### Steps to Reproduce

I am wondering if there is a documentation issue related to watchPaths. In my effort to get watchPaths working in a Turborepo repo (which I still have not been able to), I came across inconsistent information about watchPaths.

The documentation...

https://remix.run/docs/en/v1/api/conventions#watchpaths

A discussion comment...

https://github.com/remix-run/remix/discussions/3896#discussioncomment-3297069

### Expected Behavior

I just wanted to get clarification.

### Actual Behavior

NA

kiliman:
Like I said in my comment, the paths should be an array of globs relative to the current project root. The docs are wrong in that it shows an absolute path.

I imagine if you need to access paths outside of the project root you can use `..` to create the correct relative path. All Remix is doing is passing these paths to `chokidar` to trigger a rebuild when a file is changed.

kevinswarner:
Thanks. I just wanted to report the bug in the documentation.

jsbmg:
@kevinswarner were you able to get `watchPaths` working?

Like in the discussion you referenced, I have a directory like this:

```
│ remix.config.js
│
└───app
│   │   ...
│   
└───posts
    └───file.md
```
and this in `remix.config.js`:
```
module.export = {
...
  watchPaths: ["./posts/**/*"]
```
}

and nothing happens when I change files in the posts directory. I've tried `./posts`, `./posts/*`, `./posts/**`, etc.

This is on Remix `1.7.5`.

kevinswarner:
@jsbmg No, I have not been able to get it to work. I am using Turborepo and have a dependent package that I want to watch, and have Remix pick up the change. When I make a change in the package, the package rebuilds correctly (it is a Typescript package), and I even see Remix detecting the change in the correct path, but, for some reason Remix will not reflect the change. Here is an example of the console output when I make a change to the package I am "watching" in watchPaths.

```shell
 💿 File changed: ../../node_modules/models/dist/appointment.js
[1] 💿 File changed: ../../node_modules/models/dist/appointment.d.ts
[1] 💿 Rebuilding...
[1] 💿 Rebuilt in 122ms
```

But, when I visit the page where I use the package, it does not reflect the change I made in the package.

Note, if I stop the Remix dev environment, and start it again, it does indeed pick up the change.

My package json scripts look like this...

```json
[...]

PR #4578, opened 2022-11-11
docs(api/conventions.md): Update watchPaths section

jsbmg:
watchPaths documentation:
* Update code example to reflect that watchPaths is an Array, and included in the exported config object
* Clarify that watchPaths directories are relative to the project root

This issue is referenced here: #4530 

And in this discussion: https://github.com/remix-run/remix/discussions/3896


machour (changes requested):
`watchPaths` can be either an async function or an array or a string, see it being used here:
 
https://github.com/remix-run/remix/blob/main/packages/remix-dev/config.ts#L448

and its type definition:

https://github.com/remix-run/remix/blob/main/packages/remix-dev/config.ts#L157

Can you update the first sentence accordingly?

jsbmg:
@machour thanks for the clarification. I updated the section. Would it be better if I made a new pull request instead of this merge?

machour:
@jsbmg totally, since the documentation completely changed yesterday and the file you were editing isn't there anymore. 
Make sure to sync main before doing the new PR.

Issue #4570, opened 2022-11-11
Broken Routing: Remix + React Router integration branch


cephalization:
### What version of Remix are you using?

0.0.0-experimental-5b4bceda6

### Steps to Reproduce

Within a Remix loader, call the redirect function with a fully qualified URL like `https://google.com`

Example loader:

```ts
export const loader = async ({ request }: LoaderArgs) => {
  const session = await requestSession(request);
  const auth = getAuthFromSession(session);

  // don't try to login again if already logged in
  if (auth) {
    return redirect("/");
  }

  // this hits a 3rd party api and returns a 304 redirect response with destination
  const res = await getLogin();

  if (res.redirected) {
    return redirect(res.url);
  }

  return redirect("/");
};
```

### Expected Behavior

Calling redirect with a fully qualified url takes my user to that URL.

### Actual Behavior

Calling redirect with a fully qualified url takes my user to my site url with the redirect url appended to the end

Like `http://localhost:3000/https://google.com`

cephalization:
I found the code causing this issue and wrote an integration test to verify.

I will play around with a few solutions. I assume this issue happens because in react router we mostly assume you are routing within an SPA but in remix that’s not necessarily the case.

brophdawg11:
This should be resolved in `1.8.0-pre.2` - @cephalization do you want to confirm your use-case?

cephalization:
Works great @brophdawg11 , thanks!

brophdawg11:
Released in `1.8.0`

PR #4579, opened 2022-11-11
Respect redirects to external url destinations

cephalization:
`redirect('https://www.google.com/')` should go to `http://www.google.com/`

We should not prepend the request origin to the destination URL if they are external.

Closes: #4570 

- [ ] Docs
- [x] Tests

Testing Strategy:

New and existing integration tests pass. Still need to copy build into my remix app to test there.

### DIFF
--- a/packages/remix-server-runtime/router/utils.ts
+++ b/packages/remix-server-runtime/router/utils.ts
@@ -817,7 +817,8 @@ export function resolvePath(to: To, fromPathname = "/"): Path {
   } = typeof to === "string" ? parsePath(to) : to;
 
   let pathname = toPathname
-    ? toPathname.startsWith("/")
+    ? // we don't want to prepend the fromPathname on root or external toPathnames'
+      toPathname.startsWith("/") || toPathname.startsWith("http")
       ? toPathname
       : resolvePathname(toPathname, fromPathname)
     : fromPathname;
--- a/contributors.yml
+++ b/contributors.yml
@@ -62,6 +62,7 @@
 - camiaei
 - CanRau
 - ccssmnn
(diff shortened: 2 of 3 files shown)


brophdawg11:
Thanks @cephalization!  For ease of ongoing dev during this transition of Remix onto RR 6.4, we are temporarily keeping a duplicate copy of the `@remix-run/router` code inside the Remix repo.  I fixed this over at the source in https://github.com/remix-run/react-router/pull/9590 (in a slightly different manner) but we definitely want to keep this new integration test you added here!  Once we get that PR merged, I'll get this PR synced up with the router changes and get this merged as well 👍

packages/remix-server-runtime/router/utils.ts
```diff
@@ -817,7 +817,8 @@ export function resolvePath(to: To, fromPathname = "/"): Path {
   } = typeof to === "string" ? parsePath(to) : to;
 
   let pathname = toPathname
-    ? toPathname.startsWith("/")
+    ? // we don't want to prepend the fromPathname on root or external toPathnames'
+      toPathname.startsWith("/") || toPathname.startsWith("http")
```
machour, line 821: @brophdawg11 do we want to accept other protocols? (app:// itunes:// ..)

brophdawg11: Yeah definitely.  I did this in RR the way it's checked in Remix today which should account for other protocols: https://github.com/remix-run/react-router/pull/9590/files#diff-c6f085a772081501e6db2af3eee90b15fe7cd7c965d2f220c2a3c5c7772bc0c4R2568

PR #4596, opened 2022-11-14
chore: fix `postrelease` workflow

MichaelDeBoey:
The workflow apparently never ran: https://github.com/remix-run/remix/actions/workflows/postrelease.yml

Found that we had a small bug: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

### DIFF
--- a/.github/workflows/postrelease.yml
+++ b/.github/workflows/postrelease.yml
@@ -4,7 +4,7 @@ on:
   push:
     tags:
       # only run on `remix` tags
-      - "remix@*"
+      - remix@**
 
 jobs:
   comment:


mcansh:
🤯 i wonder why my test repo works https://github.com/mcansh/nightly-release-test/blob/main/.github/workflows/postrelease.yml maybe something to do with not using a scoped package here?

MichaelDeBoey:
@mcansh According to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore, a single `*` should work as well 🤔

mcansh (approved):

duailibe:
> @mcansh It seems like this didn't work for some reason 😢 https://github.com/remix-run/remix/actions/workflows/postrelease.yml

I believe it's because this commit was not included in the tree of the tag that was pushed. Unless it was cherry-picked before the push.

Next release it should probably work.

mcansh:
> @MichaelDeBoey I believe it's because this commit was not included in the tree of the tag that was pushed. Unless it was cherry-picked before the push.
> 
> 
> 
> Next release it should probably work.

good point, nightlies require it to be on the main branch, but stable ones should probably be on the dev (or release-* if it exists) branch

PR #4502, opened 2022-11-03
templates(cloudflare-workers): support streaming SSR

lili21:
Closes: #

- [x] Docs
- [ ] Tests

Testing Strategy:

PR #4600, opened 2022-11-14
feat(remix-dev): make `entry.{client,server}.tsx` optional

mcansh:
made this before making a proposal like a dummy, but it's something we've talked about a bit internally

proposal here 👉 https://github.com/remix-run/remix/discussions/4601

follow PR to make remix config optional 🎉 

edit on the optional remix config.. turns out that already works as we check before we attempt to load it. https://github.com/remix-run/remix/blob/611c68bef0df137e79c99e1c6c111a0eabafd3f9/packages/remix-dev/config.ts#L321-L343

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:


MichaelDeBoey:
What about people using JS instead of TS?
Do they need to have this file in their repo if they never changed it?

Should we also remove these files from our templates/examples?

packages/remix-dev/cli/commands.ts
```diff
@@ -347,3 +348,45 @@ function purgeAppRequireCache(buildPath: string) {
...
+  // 0. valid entry file
+  let entries = new Set([
+    "entry.server.tsx",
+    "entry.server.js",
+    "entry.server.jsx",
+    "entry.client.tsx",
+    "entry.client.js",
+    "entry.client.jsx",
+  ]);
+  if (!entries.has(entry)) {
```
MichaelDeBoey, line 362: `Array.includes` appeared to be faster

```suggestion
  if (!entries.includes(entry)) {
```

packages/remix-dev/cli/commands.ts
```diff
@@ -347,3 +348,45 @@ function purgeAppRequireCache(buildPath: string) {
     }
   }
 }
+
+export async function generateEntry(remixRoot: string, entry: string) {
+  // 0. valid entry file
+  let entries = new Set([
```
MichaelDeBoey, line 354:
```suggestion
  let entries = [
```

packages/remix-dev/cli/commands.ts
```diff
@@ -347,3 +348,45 @@ function purgeAppRequireCache(buildPath: string) {
...
+    "entry.server.tsx",
+    "entry.server.js",
+    "entry.server.jsx",
+    "entry.client.tsx",
+    "entry.client.js",
+    "entry.client.jsx",
+  ]);
```
MichaelDeBoey, line 361:
```suggestion
  ];
```

packages/remix-dev/cli/commands.ts
```diff
@@ -347,3 +348,45 @@ function purgeAppRequireCache(buildPath: string) {
...
+  if (!entries.has(entry)) {
+    console.log(
+      colors.error(
+        `Invalid entry file. Valid entry files are ${Array.from(entries).join(
```
MichaelDeBoey, line 365:
```suggestion
        `Invalid entry file. Valid entry files are ${entries.join(
```

MichaelDeBoey (changes requested):
I think we should also run the JS migration on the file when people choose the `js`/`jsx` extension in the generator command?

mcansh:
> I think we should also run the JS migration on the file when people choose the `js`/`jsx` extension in the generator command?

definitely, this is still very much WIP

packages/remix-dev/cli/commands.ts
```diff
@@ -234,3 +237,103 @@ export async function codemod(
...
+      contents = contents.replace(/@remix-run\/node/g, "@remix-run/cloudflare");
+    } else if (deps["@remix-run/deno"]) {
+      contents = contents.replace(/@remix-run\/node/g, "@remix-run/deno");
+    }
+  }
+
+  // 3. if entry is js/jsx, convert to js
+  // otherwise, copy the entry file from the defaults
+  if (/\.jsx?$/.test(entry)) {
+    contents = contents ||= await fse.readFile(inputFile, "utf-8");
```
MichaelDeBoey, line 319: I think you meant

```suggestion
    contents ||= await fse.readFile(inputFile, "utf-8");
```

mcansh: yes, yes i did 🤪 - gonna need to also update PassThrough in deno/cf too, so may go with something different anyways

mcansh:
> What about people using JS instead of TS? 

if they "eject" we'll transform it to js

> Do they need to have this file in their repo if they never changed it?

doesn't hurt

> Should we also remove these files from our templates/examples?

that's up for debate, templates would get any update faster than it being bundled in the CLI (unless we hit GitHub to get the file and use the bundled one if they're offline)

packages/remix-dev/config/defaults/entry.client.tsx
```diff
@@ -1,3 +1,4 @@
+import * as React from "react";
```
MichaelDeBoey, line 1: I don't think this is necessary?

MichaelDeBoey (approved):

packages/remix-dev/config/defaults/entry.client.tsx
```diff
@@ -13,10 +13,10 @@ function hydrate() {
...
-if (typeof requestIdleCallback === "function") {
-  requestIdleCallback(hydrate);
+if (window.requestIdleCallback) {
+  window.requestIdleCallback(hydrate);
 } else {
   // Safari doesn't support requestIdleCallback
   // https://caniuse.com/requestidlecallback
-  setTimeout(hydrate, 1);
+  window.setTimeout(hydrate, 1);
 }
```
MichaelDeBoey, line 22: I think we should keep these to check if global `requestIdleCallback` is a function (see #4998)

```suggestion
if (typeof requestIdleCallback === "function") {
  requestIdleCallback(hydrate);
} else {
  // Safari doesn't support requestIdleCallback
  // https://caniuse.com/requestidlecallback
  setTimeout(hydrate, 1);
}
```

packages/remix-dev/config/defaults/entry.server.cloudflare.tsx
```diff
@@ -0,0 +1,32 @@
+import { type EntryContext } from "@remix-run/cloudflare";
```
MichaelDeBoey, line 1:
```suggestion
import type { EntryContext } from "@remix-run/cloudflare";
```

packages/remix-dev/config/defaults/entry.server.deno.tsx
```diff
@@ -0,0 +1,32 @@
+import { type EntryContext } from "@remix-run/cloudflare";
```
MichaelDeBoey, line 1:
```suggestion
import type { EntryContext } from "@remix-run/deno";
```

MichaelDeBoey (changes requested):

templates/cloudflare-pages/package.json
```diff
@@ -8,7 +8,7 @@
     "dev": "remix build && run-p \"dev:*\"",
     "start": "cross-env NODE_ENV=production npm run wrangler",
     "typecheck": "tsc -b",
-    "wrangler": "wrangler pages dev ./public"
+    "wrangler": "wrangler pages dev ./public --compatibility-flags streams_enable_constructors"
```
MichaelDeBoey, line 11: I think putting this in `wrangler.toml` would be a better idea, as that makes all config is in 1 place (just like @lili21 did in #4502)

packages/remix-dev/config/defaults/entry.server.deno.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export default async function handleRequest(
+  request: Request,
+  responseStatusCode: number,
+  responseHeaders: Headers,
+  remixContext: EntryContext
+) {
+  const body = await renderToReadableStream(
+    <RemixServer context={remixContext} url={request.url} />,
+    {
```
mcansh, line 14: add `signal: request.signal` to allow aborting

packages/remix-dev/cli/commands.ts
```diff
@@ -246,3 +249,107 @@ export async function codemod(
...
+      `Entry file ${entry} created at ${path.relative(remixRoot, outputFile)}.`
+    )
+  );
+
+  return process.exit(0);
+}
+
+async function checkForEntry(remixRoot: string, entries: string[]) {
+  for (let entryToCheck of entries) {
+    let entryPath = path.resolve(remixRoot, "app", entryToCheck);
```
mcansh, line 332: use `appDirectory` from config

packages/remix-dev/cli/commands.ts
```diff
@@ -246,3 +249,107 @@ export async function codemod(
...
+    let entryExists = await fse.pathExists(entryPath);
+    if (entryExists) {
+      console.error(
+        colors.error(
+          `Entry file ${path.relative(remixRoot, entryToCheck)} already exists.`
+        )
+      );
+      return process.exit(1);
```
mcansh, line 340: don't exit, just warn

packages/remix-dev/compiler/plugins/cssBundleEntryModulePlugin.ts
```diff
@@ -27,7 +27,7 @@ export function cssBundleEntryModulePlugin(config: RemixConfig): Plugin {
           contents: [
             // These need to be exports to avoid tree shaking
             `export * as entryClient from ${JSON.stringify(
-              `./${config.entryClientFile}`
+              config.entryClientFilePath
```
mcansh, line 30: keep it relative to make output hash deterministic

packages/remix-dev/config/defaults/entry.server.cloudflare.tsx
```diff
@@ -0,0 +1,32 @@
+import type { EntryContext } from "@remix-run/cloudflare";
+import { RemixServer } from "@remix-run/react";
+import isbot from "isbot";
```
mcansh, line 3: we should also check and install `isbot` if it's not already, then log that they should commit the change

templates/cloudflare-pages/app/entry.server.tsx
```diff
@@ -1,21 +1,30 @@
...
+  const stream = await renderToReadableStream(
+    <RemixServer context={remixContext} url={request.url} />,
+    {
+      signal: request.signal,
+      onError(error: unknown) {
+        didError = true;
+        console.error(error);
+      },
+    }
   );
```
jacob-ebey, line 23: need an "isbot" check.

templates/deno/app/entry.server.tsx
```diff
@@ -1,22 +1,30 @@
...
+  const stream = await renderToReadableStream(
     <RemixServer context={remixContext} url={request.url} />,
+    {
+      signal: request.signal,
+      onError(error: unknown) {
+        didError = true;
+        console.error(error);
+      },
+    }
   );
```
jacob-ebey, line 23: Also needs "isbot".

jacob-ebey (approved):

Issue #4353, opened 2022-10-11
Occasional error when reading session data from file system


midgleyc:
### What version of Remix are you using?

1.6.5

### Steps to Reproduce

1. Use file session storage (`createFileSessionStorage`)
2. Create a large amount of session data (>1MB), saved and read each page load.
3. Visit lots of pages quickly.

### Expected Behavior

No errors.

### Actual Behavior

Occasionally, the page fails to load:

```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at readData (/srv/app/node_modules/@remix-run/node/dist/sessions/fileStorage.js:91:28)
    at getSession (/srv/app/node_modules/@remix-run/server-runtime/dist/sessions.js:113:25)
    at loader10 (/srv/app/build/index.js:2229:17)
    at Object.callRouteLoader (/srv/app/node_modules/@remix-run/server-runtime/dist/data.js:77:14)
    at handleDataRequest (/srv/app/node_modules/@remix-run/server-runtime/dist/server.js:113:18)
    at requestHandler (/srv/app/node_modules/@remix-run/server-runtime/dist/server.js:34:18)
    at /srv/app/node_modules/@remix-run/express/dist/server.js:39:22
```

kiliman:
The file session storage implementation doesn't work well since loaders can run in parallel. There is no file-locking on the current one so it's very possible to corrupt session data.

kiliman:
I've got a working version with file locks, but still trying to figure out performance issues.

raskyer:
Hello, do you know if any progress could be made around this issue ?

dusty:
I just made one to run synchronously and it doesn't corrupt - we only use this in local dev - we use redis in prod.

```
// https://remix.run/docs/en/v1/api/remix#createsessionstorage
import type { Cookie } from '@remix-run/node'
import { createSessionStorage } from '@remix-run/node'
import crypto from 'crypto'
import fs from 'fs'
import path from 'path'

interface FileSessionStorageOptions {
  cookie: Cookie
  dir: string
}

// sync version so the file isn't corrupted on concurrent writes
export function createFileSessionStorage({ cookie, dir }: FileSessionStorageOptions) {
  return createSessionStorage({
    cookie,
    async createData(data, expires) {
      const content = JSON.stringify({ data, expires })
      // eslint-disable-next-line no-constant-condition
      while (true) {
        const id = crypto.randomUUID({ disableEntropyCache: true })
        try {
[...]

PR #4604, opened 2022-11-15
fix(remix-node): use atomic writes for file storage

midgleyc:
All loaders run in parallel. If multiple loaders that run access the session, there is a potential race condition without atomic file writes:

* Loader 1 begins overwriting the file, but does not finish (`session.set`)
* Loader 2 reads the whole file into a variable and attempts to parse the variable as JSON (`session.get`), but fails because the file was only partially written

As the test needs to reproduce a race condition involving multiple files, it is difficult to write. I think it should be possible with an integration test, but I will need to spend some more time to understand how it should work.

Closes: #4353

- [ ] Docs
- [ ] Tests

Testing Strategy:

We have the problem described in the attached ticket, and could reliably reproduce the original issue by rapidly switching between pages until it crashed. With the changed session storage, we have not seen the error and the original method to reproduce the issue does not crash.

### DIFF
--- a/.changeset/strong-snails-stare.md
+++ b/.changeset/strong-snails-stare.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/node": patch
+---
+
+Fix race condition with non-atomic writes for file storage.
--- a/contributors.yml
+++ b/contributors.yml
@@ -351,6 +351,7 @@
 - michaelgmcd
 - michaelhelvey
 - michaseel
+- midgleyc
 - mikeybinnswebdesign
 - mirzafaizan
 - mitchelldirt
--- a/packages/remix-node/package.json
+++ b/packages/remix-node/package.json
@@ -25,11 +25,13 @@
     "abort-controller": "^3.0.0",
     "cookie-signature": "^1.1.0",
     "source-map-support": "^0.5.21",
-    "stream-slice": "^0.1.2"
+    "stream-slice": "^0.1.2",
+    "write-file-atomic": "^5.0.0"
   },
   "devDependencies": {
     "@types/cookie-signature": "^1.0.3",
-    "@types/source-map-support": "^0.5.4"
+    "@types/source-map-support": "^0.5.4",
+    "@types/write-file-atomic": "^4.0.0"
   },
   "engines": {
     "node": ">=14"
--- a/packages/remix-node/sessions/fileStorage.ts
+++ b/packages/remix-node/sessions/fileStorage.ts
@@ -53,7 +53,7 @@ export function createFileSessionStorage<Data = SessionData, FlashData = Data>({
         try {
           let file = getFile(dir, id);
           await fsp.mkdir(path.dirname(file), { recursive: true });
-          await writeFile(file, content, { encoding: "utf-8" });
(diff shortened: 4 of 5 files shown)


machour:
Hi @midgleyc and thank you so much for tackling this. 
Please rebase this PR on the `dev` branch

midgleyc:
I've made a change to revert the initial creation back to using `fs` to simplify the change.

As the `wx` flag is passed, the file is only written to if it did not exist. In practice, this makes it atomic already: if the file does not exist, it is created; if it does exist nothing happens.

MichaelDeBoey:
Hi @midgleyc!

midgleyc:
Sure, I've rebased.

musou1500:
hello, I facing same problem.

I look around write-file-atomic and have concerns about filename collision.

write-file-atomic decide temporary filename as follows.

> The file is initially named filename + "." + murmurhex(__filename, process.pid, ++invocations)

In case of use multiple application server and share session files over NFS, can `murmurhex(__filename, process.pid, ++invocations)` collide over servers?

especially, in container environment such as docker, there is concern that process.pid will be the same on multiple servers.

if filename collide, write-file-atomic overwrite existing file.

midgleyc:
The error from the integration tests is:
```
✘ [ERROR] This require call is not allowed because the imported file "../../../node_modules/@jspm/core/nodelibs/browser/worker_threads.js" contains a top-level await

    ../../../node_modules/write-file-atomic/lib/index.js:18:34:
      18 │     const workerThreads = require('worker_threads')
         ╵                                   ~~~~~~~~~~~~~~~~

  The top-level await in "../../../node_modules/@jspm/core/nodelibs/browser/worker_threads.js" is here:

    ../../../node_modules/@jspm/core/nodelibs/browser/worker_threads.js:97:49:
      97 │ ...rkerData, environmentData }] = await once(parentPort, 'message'));
```

This comes from [esbuild](https://github.com/evanw/esbuild/blob/7d11ef1e24a3f0981e45e37200957268c4e22619/internal/bundler/bundler.go#L2300), and relates to trying to bundle a (series of) packages with top level await somewhere: https://github.com/evanw/esbuild/issues/253#issuecomment-806500846

`write-file-atomic` uses `require('worker_threads')` inside [an IIFE](https://github.com/npm/write-file-atomic/blob/a0daf642b441f3026de36f8d10dae24e46b34f01/lib/index.js#LL18C1-L18C1) to distinguish between multiple threads running on the same pid.

The integration tests have this import the [`jspm-core` version](https://github.com/jspm/jspm-core/blob/209cd168383d76215445c79001b9f45568dd92df/src-browser/worker_threads.js#L96) of worker_threads, which contains a top-level await, which breaks the bundling.

I don't know what JSPM is used for, but the readme mentions "implementing the optimized browser versions of the Node.js builtins", and I think the [actual node version](https://github.com/nodejs/node/blob/main/lib/worker_threads.js) does not contain the top level await issue.

brophdawg11:
The current remix v1 design of parallel loader execution means that accessing the same sessions from different loaders is always non-deterministic if one or any of them are writing to the session.  The approach now would be to use different sessions to read/write different values in loaders, or only read values from a shared session across loaders (no setting in loaders).

PR #4608, opened 2022-11-15
docs(guides/data-loading & api/conventions): fix type inference & add json import

hakimLyon:


machour:
Justement. Open the next PR against main 🙏

PR #4610, opened 2022-11-15
feat(`react`,`server-runtime`): Support for v2 array syntax for route `meta`

chaance:
This is the implementation for https://github.com/remix-run/remix/discussions/4462.

Closes: https://github.com/remix-run/remix/pull/2598

Note: The types here are not complete and not ideal, but we are running into a wall with our current approach of copy-paste-all-the-types because we lack a shared dependency, and continuing down that path will make it even harder to fix. We are prioritizing a solution to that in a separate proposal and providing minimal (but sufficient) types in the mean time.

Also, docs will come in a separate PR because I'd like us to kick the tires on it a bit before documenting the change. This is fine since it's completely opt-in anyway.

- [ ] Docs
- [x] Tests


integration/meta-test.ts
```diff
@@ -419,7 +419,6 @@ test.describe("v2_meta", () => {
             { name: "description", content: data.description },
             { property: "og:image", content: "https://picsum.photos/200/200" },
             { property: "og:type", content: data.contentType }, // undefined
-            { httpEquiv: "refresh", content: "3;url=https://www.mozilla.org" },
```
machour: why did this one get removed? 😬

chaance: Because it was super annoying to debug with `app.poke` because the browser kept redirecting me to mozilla.org 😂 

I don't mind adding it back, just pulled it out while working through some issues. That said, I think it's a little less necessary to test various individual properties with the new approach since we're no longer doing anything special for `"refresh"`.

machour: Oh, I'm fine with removing a test because it works too well 😅

jacob-ebey (approved):

machour (approved):

Issue #4422, opened 2022-10-26
Adding `//` to the end of the root route throws an Internal Server Error


krall12:
### What version of Remix are you using?

1.7.3

### Steps to Reproduce

Spin up a new remix project from scratch. Just the basics. Remix Server. JavaScript. 

Start the dev server and visit the browser at `localhost:3000//`. 

Also, for fun check out `https://remix.run//`

### Expected Behavior

I would expect this to not error. Ideally it would redirect back to the root route. 

### Actual Behavior

On production: An `Internal Server Error` is thrown that doesn't get caught by Error Boundaries. 

On development: An `Invalid URL` error is thrown that also doesn't get caught by Error Boundaries. 
```
TypeError: Invalid URL
    at new NodeError (node:internal/errors:372:5)
    at URL.onParseError (node:internal/url:553:9)
    at new URL (node:internal/url:629:5)
    at createRemixRequest (/Users/bkrall/content/remix-bug-server-error-on-route/node_modules/@remix-run/express/dist/server.js:67:13)
    at /Users/bkrall/content/remix-bug-server-error-on-route/node_modules/@remix-run/express/dist/server.js:37:21
    at /Users/bkrall/content/remix-bug-server-error-on-route/node_modules/@remix-run/serve/dist/index.js:48:7
    at Layer.handle [as handle_request] (/Users/bkrall/content/remix-bug-server-error-on-route/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/bkrall/content/remix-bug-server-error-on-route/node_modules/express/lib/router/route.js:144:13)
    at next (/Users/bkrall/content/remix-bug-server-error-on-route/node_modules/express/lib/router/route.js:140:7)
    at next (/Users/bkrall/content/remix-bug-server-error-on-route/node_modules/express/lib/router/route.js:140:7)
```

brophdawg11:
This is fixed in Remix 1.13.0 - we aren't doing any redirecting but instead are ensuring our adapters can correctly generate a URL with these trailing slashes.  It would still be up to the app to detect and redirect in a loader (or eventually in [middleware](https://github.com/orgs/remix-run/projects/5?pane=issue&itemId=13430611) once that lands).

Hansenq:
[...]

PR #4622, opened 2022-11-16
Redirects when multiple slashes

ErimTuzcuoglu:
Closes: #4422

- [ ] Docs
- [x] Tests

Testing Strategy: New and existing tests pass. We can try slashed routes manually.

### DIFF
--- a/packages/remix-express/server.ts
+++ b/packages/remix-express/server.ts
@@ -53,7 +53,9 @@ export function createRequestHandler({
   ) => {
     try {
       let cleanUrl = req.url.replace(/\/+/g, '/');
-      if(req.url !== cleanUrl) return res.redirect(cleanUrl);
+      if (req.url !== cleanUrl) {
+        return res.redirect(cleanUrl);
+      }
 
       let request = createRemixRequest(req, res);
       let loadContext = getLoadContext?.(req, res);
--- a/packages/remix-express/server.ts
+++ b/packages/remix-express/server.ts
@@ -52,6 +52,9 @@ export function createRequestHandler({
     next: express.NextFunction
   ) => {
     try {
+      let cleanUrl = req.url.replace(/\/+/g, '/');
+      if(req.url !== cleanUrl) return res.redirect(cleanUrl);
+
       let request = createRemixRequest(req, res);
       let loadContext = getLoadContext?.(req, res);
 
--- a/contributors.yml
+++ b/contributors.yml
@@ -119,6 +119,7 @@
 - eldarshamukhamedov
 - emzoumpo
 - eps1lon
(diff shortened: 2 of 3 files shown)


packages/remix-express/server.ts
```diff
@@ -52,6 +52,9 @@ export function createRequestHandler({
     next: express.NextFunction
   ) => {
     try {
+      let cleanUrl = req.url.replace(/\/+/g, '/');
+      if(req.url !== cleanUrl) return res.redirect(cleanUrl);
```
machour, line 56:
```suggestion
      if (req.url !== cleanUrl) {
        return res.redirect(cleanUrl);
      }
```

ErimTuzcuoglu: Updated

brophdawg11:
Is this something we need to consider for servers/runtimes other than express?

machour (changes requested):
@ErimTuzcuoglu can you please fix the conflict in the contributors files? 🙏🏼 
Protip: put your name in the right spot alphabetically, conflicts will be less likely to happen

brophdawg11:
@ErimTuzcuoglu I didn't dig too deep, but was able to reproduce this using the Architect template as well?  I'm wondering this is something that should be handled in a more central location, or at least should this logic go into each of the adapters.  I will probably defer to one of the Remix team members here with more experience in the alternative runtimes.  @jacob-ebey @mcansh?

mcansh:
> I'm wondering this is something that should be handled in a more central location

i just tried a [github](https://github.com//mcansh//snkrs//tree//main//app) url and it resolved with the multiple slashes without redirecting - perhaps we follow suit and normalize while matching?

ErimTuzcuoglu:
The main problem is actually on URL constructor. It doesn't accept multiple slashes as a url and throws error. So when we removed slashes it works correctly, but for redirection I think we must make different implementations on adapters. As @mcansh said, on vercel we don't need any changes. But if we need I can check also other adapters and fix them too.

image

mcansh:
> The main problem is actually on URL constructor. It doesn't accept multiple slashes as a url and throws error. So when we removed slashes it works correctly, but for redirection I think we must make different implementations on adapters. As @mcansh said, on vercel we don't need any changes. But if we need I can check also other adapters and fix them too.
> 
> 
> 
> image
> 
> 

whats interesting is that `new URL("https://github.com///")` seems to be fine image

machour (changes requested):
@ErimTuzcuoglu Please update other adapters as well and rebase this PR to latest dev 🙏🏼

brophdawg11:
Closing in favor of https://github.com/remix-run/remix/pull/5336

PR #4661, opened 2022-11-21
Fix magic export deprecation messaging

brophdawg11:
This reverts #3284 due to a few unforeseen issues:

* The concatenating of files resulted in duplicate definitions of `const getDeprecatedMessage` and `const warn` which caused build failures
* The import and actual _usage_ of things in the ESM file broke the ability to tree-shake server-only code out of client bundles:

```js
// Tree shakes correctly as long as things aren't used
export { createCookie } from '@remix-run/node`;

// Doesn't tree shake because `createCookie` is imported and referenced
import * as node from '@remix-run/node';
const createCookie = warn(node.createCookie, ...);
```

Instead this goes back to the old direct-export approach for the CJS/ESM modules, but keeps the annotated TS `@deprecated` exports.  Then we added a new `esbuild` plugin that detects imports from `remix` at build time and logs out a warning for the problematic file:

Screenshot 2022-11-21 at 3 07 10 PM

**Testing**
* Start a fresh remix app 
  * `npx create-remix@1.8.0-pre.0`
* Add a deprecated import in one of your routes
  * `import { Link } from "remix"`
* Do a local remix build from this branch and direct the output to your new app:
  * `REMIX_LOCAL_BUILD_DIRECTORY=../playgrounds/my-remix-app yarn build`
* Re-run the setup command in your Remix app directory:
  * `npx remix setup node` 
* `npm run dev` should show you a build time warning for the deprecated import
* TS should show a deprecation error in VScode for the problematic import


packages/remix-dev/cli/setup.ts
```diff
@@ -94,7 +94,16 @@ async function combineFilesInDirs(
...
-  let combined = "";
+  let logDeprecationWarning =
+    `console.warn("WARNING: All \`remix\` exports are considered deprecated as of v1.3.3. ` +
+    `All imports should come from their respective packages (i.e., ` +
+    `\`@remix-run/server-runtime\`, \`@remix-run/react\`, etc.). ` +
+    `Run \`npx @remix-run/dev@latest codemod replace-remix-magic-imports\` to ` +
+    `automatically migrate your code.");\n\n`;
+
+  // Don't include the warning in type declarations
+  let combined = ext === ".d.ts" ? "" : logDeprecationWarning;
```
brophdawg11, line 105: Add the warning once to the top of the CJS and ESM files

mcansh (approved):

MichaelDeBoey:
It would be a shame to revert #3284 imo, as we would lose TS deprecations that way, which are another added benefit imo.

Since this only seems to be impacting ESM builds, can we at least keep CJS & TS?

> * The concatenating of files resulted in duplicate definitions of `const getDeprecatedMessage` and `const warn` which caused build failures

This could be fixed by removing it from the specific `magicExports` file & including it in the concatenated file _or_ we could rename these functions to `getDeprecatedMessage_[package-name]` & `warn_[package-name]` to not have duplicate function names 🤔

> * The import and actual _usage_ of things in the ESM file broke the ability to tree-shake server-only code out of client bundles:
> 
> ```js
> // Tree shakes correctly as long as things aren't used
> export { createCookie } from '@remix-run/node`;
> 
> // Doesn't tree shake because `createCookie` is imported and referenced
> import * as node from '@remix-run/node';
> const createCookie = warn(node.createCookie, ...);
> ```

So if we can update the ESM imports properly, we can keep these I guess?

brophdawg11:
@MichaelDeBoey Definitely open to suggestions, this was just the quick fix towards getting `1.8.0-pre.1` out for community testing.  

I like the idea of keeping the TS deprecation warnings.  

I'm not sure if this would work for ESM, since the import is used and therefore I don't think qualifies for tree-shaking:

```
import { createCookie as node_createCookie } from '@remix-run/node';
/** @deprecated Import `createCookie` from `@remix-run/node` instead. */
const createCookie = warn(node_createCookie, getDeprecatedMessage('createCookie', '@remix-run/node'));
//                        ^ by accessing the imported thing here, I don't think it'll be tree shaken
```

Let me see about keeping the types.  I think I lean towards keeping the CJS/ESM approaches the same to avoid unintended behavioral differences, but curious what others think there too. 

I did the `_{packagename}` suffix locally as well and it fixes the duplicate identifier issue, but we would also need to enhance the `warn` function since the react components aren't functions and instead react element objects, so they need to be exported as is instead of called as a function.

MichaelDeBoey:
> I'm not sure if this would work for ESM, since the import is used and therefore I don't think qualifies for tree-shaking:
> 
> ```
> import { createCookie as node_createCookie } from '@remix-run/node';
> /** @deprecated Import `createCookie` from `@remix-run/node` instead. */
> const createCookie = warn(node_createCookie, getDeprecatedMessage('createCookie', '@remix-run/node'));
> //                        ^ by accessing the imported thing here, I don't think it'll be tree shaken
> ```

What about changing it to the following?

```ts
import node from '@remix-run/node';
/** @deprecated Import `createCookie` from `@remix-run/node` instead. */
const createCookie = warn(node.createCookie, getDeprecatedMessage('createCookie', '@remix-run/node'));
```

> Let me see about keeping the types. I think I lean towards keeping the CJS/ESM approaches the same to avoid unintended behavioral differences, but curious what others think there too.

This would mean we would loose the benefit of the deprecated JSdoc on all these functions

> we would also need to enhance the `warn` function since the react components aren't functions and instead react element objects, so they need to be exported as is instead of called as a function.

Since our React component are function components, it's basically the same as a normal function, no?
Also: the `warn` function is re-exporting the same functionality as the original function was doing.

brophdawg11:
@MichaelDeBoey As soon as we reference anything from `@remix-run/node` in any way in that bundle, other than a direct re-export, it's unable to be tree shaken and ends up in the client bundle - since the calling of `warn` is a side effect which breaks tree-shaking:

Screenshot 2022-11-21 at 2 54 25 PM

I added back in the TS deprecations in 5083780 👍 

Then @jacob-ebey had a better idea of how we can do this warning at build-time via an `esbuild` plugin, rather than litter runtime server/browser logs with these warnings (see 229503e).  This allows us to better point the user at the file with the problematic import.

I think the React component was potentially from `Link` being a `React.forwardRef` - the `fn` coming into `warn` was a react-looking object instead of a function in my testing.

mcansh (approved):

rollup.utils.js
```diff
@@ -223,17 +206,15 @@ function magicExportsPlugin({ packageName, version }) {
...
-          .join("\n");
+        let cjsModule = camelCase(packageName.slice("@remix-run/".length));
+        cjsContents += `var ${cjsModule} = require('${packageName}');\n`;
+        for (let symbol of magicExports.values) {
+          cjsContents +=
+            `Object.defineProperty(exports, '${symbol}', {\n` +
+            "  enumerable: true,\n" +
+            `  get: function () { return ${cjsModule}.${symbol}; }\n` +
+            "});\n";
+        }
```
MichaelDeBoey, line 217: I think we can keep the JSdoc deprecation here as well to have an extra IDE warning?

```suggestion
        cjsContents += `var ${moduleName} = require('${packageName}');\n`;
        cjsContents += magicExports.values
          .map(
            (symbol) =>
              `/** @deprecated Import \`${symbol}\` from \`${packageName}\` instead. */\n` +
              `exports.${symbol} = ${symbol};\n`
          )
          .join("\n");
```

MichaelDeBoey (approved):
Even though we would loose the extra JSdoc deprecation warnings in your IDE, if there's no other reason to make sure we can still three-shake in ESM, I like the idea of `esbuild` throwing a warning 👍

MichaelDeBoey:
@brophdawg11 When doing the following:

- Run
  
  ```sh
  npx create-remix@1.8.0-pre.1 --template remix remix-test-magicExports
  ```
  Choosing all defaults, except for JavaScript

- Run
  ```sh
  npm i -D remix@1.8.0-pre.1
  npx remix setup
  ```
- Disable all ESLint extends
- Go to `node_modules/remix/dist/index.js`
- Replace
  ```ts
  Object.defineProperty(exports, 'createCookie', {
    enumerable: true,
    get: function () { return node.createCookie; }
  });
  ```
  with
  ```ts
  /** @deprecated Import `createCookie` from `@remix-run/node` instead. */
  exports.createCookie = node.createCookie;
  ```
  (which is what my last suggestion was doing)
- Add the following line to `index.jsx`
  ```ts
  import {createCookie} from "remix";
  
  const cookieCreator = createCookie; // this is just so my IDE (WebStorm) isn't complaining about `createCookie` not being used
  ```

I get the deprecated warning:
Schermafbeelding 2022-11-22 om 22 46 42

PR #4668, opened 2022-11-22
Bump remix to react-router-dom@6.4.3

brophdawg11:
* Remove the `react-router-dom` dependency from `@remix-run/server-runtime` since it can pull everything from `./router` now
* Upgrade `@remix-run/react` to use `react-router-dom@~6.4.3`.  Went with a `~` here so we inherit bug fixes automatically but have to manually perform minor version upgrades so we have a chance to ensure full compatibility

Note this is not using `StaticRouterProvider`/`RouterRrovider` yet, still using the old approach (direct `Router` usage), just from the latest deployed version.

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -239,6 +239,10 @@ function Routes() {
   // need this component, we can just `renderMatches` from RemixEntry
   let { clientRoutes } = useRemixEntryContext();
   // fallback to the root if we don't have a match
+
+  // TODO: clientRoutes currently errors here since RR 6.4 dropped signal as a
+  // loader argument.  We'll be adding tat bac in as part of the rendering work.
+  // @ts-expect-error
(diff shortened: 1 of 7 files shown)


jacob-ebey (approved):

packages/remix-react/components.tsx
```diff
@@ -239,6 +239,10 @@ function Routes() {
   // need this component, we can just `renderMatches` from RemixEntry
   let { clientRoutes } = useRemixEntryContext();
   // fallback to the root if we don't have a match
+
+  // TODO: clientRoutes currently errors here since RR 6.4 dropped signal as a
+  // loader argument.  We'll be adding tat bac in as part of the rendering work.
```
machour, line 244:
```suggestion
  // loader argument.  We'll be adding that back in as part of the rendering work.
```

PR #4669, opened 2022-11-22
RRR Step 2 - Server Rendering

brophdawg11:
Initial work towards step 2 of [React Router-ing Remix](https://remix.run/blog/react-routering-remix)

* Server rendering via `unstable_createStaticRouter`/`StaticRouterProvider`
* No hydration handled yet, so only test on apps without `<Scripts />` 😉 
* The duplicate copy of `@remix-run/router` is removed in favor of `1.0.4-pre.0`

For branch integration testing, all SSR integration tests have been temporarily given `.only` and can be run with:

```
yarn test:integration action-test bug-report-test catch-boundary-data-test catch-boundary-test error-boundary-test form-data-test headers-test helpers/create-fixture path-mapping-test prefetch-test rendering-test splat-routes-test form-test resource-routes-test upload-test --project chromium
```

For local app testing:
* Remove the `<Scripts />` component from your local remix app since this branch only handles server-rendering
* Install `1.8.0-pre.2` into your local remix app
* Build this branch into your local remix app:
  * `REMIX_LOCAL_BUILD_DIRECTORY=../path/to/your/app yarn build`
* Remove stale duped dependencies
  * `rm -rf node_modules/@remix-run/server-runtime/node_modules`
  * `rm -rf node_modules/@remix-run/react/node_modules`
  * `rm -rf node_modules/react-router-dom/node_modules`
  * `rm -rf node_modules/react-router/node_modules`
* `npm run dev` 

Look for `✅ Rendering RR 6.4 StaticRouterProvider` in your server log to confirm that the new rendering path is applied


packages/remix-react/components.tsx
```diff
@@ -45,354 +54,130 @@ import {
...
-  matches: BaseRouteMatch<ClientRoute>[];
-  routeData: RouteData;
-  actionData?: RouteData;
-  pendingLocation?: Location;
-  appState: AppState;
-  routeModules: RouteModules;
-  serverHandoffString?: string;
-  clientRoutes: ClientRoute[];
-  transitionManager: ReturnType<typeof createTransitionManager>;
-  future: FutureConfig;
```
brophdawg11, line 81: 🪓 Majority of this is now handled by the router.  A new `RemixContext` keeps the Remix-specific things: `manifest`, `routeModules`, `serverHandoffString` and `future`.  `<RemixServer>` and `<RemixBrowser>` render `<RemixContext.Provider>` directly

packages/remix-react/components.tsx
```diff
@@ -45,354 +54,130 @@ import {
...
+    return <RemixCatchBoundary component={CatchBoundary} catch={error} />;
+  }
 
-    context = maybeServerRenderError
-      ? {
-          id,
-          get data() {
-            console.error("You cannot `useLoaderData` in an error boundary.");
-            return undefined;
-          },
```
brophdawg11, line 372: I think this functionality is probably still missing - should get this added to the router

packages/remix-react/components.tsx
```diff
@@ -45,354 +54,130 @@ import {
...
-        }
-      : { id, data };
-
-    element = (
+  if (!isRouteErrorResponse(error) && ErrorBoundary) {
+    return (
+      // TODO: Handle error type?
       <RemixErrorBoundary
```
brophdawg11, line 171: At the moment we just proxy through to the current `RemixErrorBoundary`/`RemixCatchBounary` componnts, but when we add the `v2_errorBoundary` flag this function will hopefully just short circuit to `return <RouteModuleErrorBoundary />`

packages/remix-react/components.tsx
```diff
@@ -548,7 +333,8 @@ export function composeEventHandlers<
  * @see https://remix.run/api/remix#meta-links-scripts
  */
 export function Links() {
-  let { matches, routeModules, manifest } = useRemixEntryContext();
+  let { manifest, routeModules } = useRemixContext();
+  let { matches } = useDataRouterStateContext();
```
brophdawg11, line 337: Most changes to existing/preserved hooks are just grabbing stuff from the new contexts

packages/remix-react/components.tsx
```diff
@@ -1024,423 +830,6 @@ function dedupe(array: any[]) {
...
-
-/**
- * A Remix-aware `<form>`. It behaves like a normal form except that the
- * interaction with the server is with `fetch` instead of new document
- * requests, allowing components to add nicer UX to the page as the form is
- * submitted and returns with data.
- *
- * @see https://remix.run/api/remix#form
- */
-let Form = React.forwardRef<HTMLFormElement, FormProps>((props, ref) => {
```
brophdawg11, line 1077: `Form` is now re-exported from `react-router-dom`

packages/remix-react/components.tsx
```diff
@@ -1490,40 +881,13 @@ export interface RouteMatch {
...
-          params,
-          data: routeData[match.route.id],
-          // if the module fails to load or an error/response is thrown, the module
-          // won't be defined.
-          handle: routeModules[match.route.id]?.handle,
-        };
-      }),
-    [matches, routeData, routeModules]
-  );
-}
```
brophdawg11, line 1518: This actually comes back since RR doesn't know about `handle` at route definition time so needs to populate from the `routeModulesCache`

packages/remix-react/components.tsx
```diff
@@ -1490,40 +881,13 @@ export interface RouteMatch {
...
-
 /**
  * Returns the JSON parsed data from the current route's `loader`.
  *
  * @see https://remix.run/api/remix#useloaderdata
  */
 export function useLoaderData<T = AppData>(): SerializeFrom<T> {
-  return useRemixRouteContext().data;
+  return useLoaderDataRR() as SerializeFrom<T>;
```
brophdawg11, line 890: This is the current solution until we get proper generics in RR - then we can re-export directly.  It felt like that was a whole undertaking unto itself though :)

packages/remix-react/data.ts
```diff
@@ -1,9 +1,11 @@
+import type { FormMethod as FormMethodRR } from "react-router-dom";
+
 import invariant from "./invariant";
 import type { Submission } from "./transition";
 
 export type AppData = any;
 
-export type FormMethod = "get" | "post" | "put" | "patch" | "delete";
+export type FormMethod = FormMethodRR;
```
brophdawg11, line 8: This is something we need to do a pass on still - probably have _lots_ of types that can either be deleted or just re-export something from RR to keep type-API compatibility

packages/remix-react/routes.tsx
```diff
@@ -78,6 +80,42 @@ export interface ClientRoute extends Route {
...
+          routeModules[route.id].ErrorBoundary != null ||
+          routeModules[route.id].CatchBoundary != null ? (
+            <RemixRouteError id={route.id} />
+          ) : undefined,
+        id: route.id,
+        path: route.path,
+        // Note: we don't need loader/action/shouldRevalidate on these routes
+        // since they're for a static render
+        handle: routeModules[route.id].handle,
+      };
```
brophdawg11, line 105: Server-render-time routes don't have loaders/actions since we already fetched the data!

packages/remix-react/server.tsx
```diff
@@ -21,67 +24,26 @@ export function RemixServer({ context, url }: RemixServerProps): ReactElement {
...
+        serverHandoffString,
+        future: context.future,
+      }}
+    >
+      <StaticRouterProvider
+        router={router}
+        context={context.staticHandlerContext}
+        hydrate={false}
+      />
+    </RemixContext.Provider>
```
brophdawg11, line 47: 👨‍🍳 💋

packages/remix-react/transition.ts
```diff
@@ -1,107 +1,11 @@
...
-import type { ClientRoute } from "./routes";
-import { matchClientRoutes } from "./routeMatching";
-import invariant from "./invariant";
-
-////////////////////////////////////////////////////////////////////////////////
-//#region Types and Utils
-////////////////////////////////////////////////////////////////////////////////
+import type { Location, NavigationType as Action } from "react-router-dom";
 
 export interface CatchData<T = any> {
```
brophdawg11, line 3: This file still exists just for typings and `IDLE_TRANSITION` etc.

packages/remix-server-runtime/server.ts
```diff
@@ -222,135 +223,53 @@ async function handleDocumentRequestRR(
...
-    future: build.future,
-  };
+  let headers = getDocumentHeadersRR(build, context);
 
   let entryContext: EntryContext = {
-    ...serverHandoff,
     manifest: build.assets,
     routeModules: createEntryRouteModules(build.routes),
-    serverHandoffString: createServerHandoffString(serverHandoff),
+    staticHandlerContext: context,
```
brophdawg11, line 231: No more `appState` - just work directly with the `@remix-run/router` `staticHandlerContext` now

packages/remix-react/components.tsx
brophdawg11: Bring this over to router - we will have a larger conversation about this in the future

packages/remix-react/components.tsx
brophdawg11: post-release: Dig into why tests fail if we remve `RemixErrorBoundary` in favor of `<ErrorBoundary error={error}>`

Issue #4639, opened 2022-11-18
Use of TypeScript 4.9 satisfies breaks build process


TenkoSpirit:
### What version of Remix are you using?

1.7.6

### Steps to Reproduce

1. Create Remix app with yarn/your tool, I used blues stack
2. Install TypeScript 4.9.3
3. Create a type, something like 
```typescript
type SomethingCool = {
  foo: number;
}
```
5. Try to create an object and use the new `satisfies` keyword with `SomethingCool`
6. Try to build  

### Expected Behavior

The project builds

### Actual Behavior

The project doesn't build and there's an error being thrown 
```
$ remix build
Building Remix app in production mode...

✘ [ERROR] [plugin browser-route-module] Build failed with 1 error:
app/root.tsx:32:4: ERROR: Expected ";" but found "satisfies"
```

jenseng:
Likely just need to bump esbuild to [0.15.13+](https://github.com/evanw/esbuild/blob/master/CHANGELOG.md#01513)

machour:
Oh perfect, then #4301 should fix this too!

pcattori:
Update: #4301 upgrades esbuild to 0.15.12. We were seeing tests hang or fail on 0.15.13+. We're working on fixing the underlying issues and then will upgrade to 0.15.13+

MichaelDeBoey:
Closed by #4301

pcattori:
@MichaelDeBoey we still haven't upgraded esbuild to 0.15.13+, so #4301 does not actually address this. I've edited the description of #4301 to make that more clear.

MichaelDeBoey:
@pcattori Oh didn't know we reverted it to `0.15.12` again, I thought we were using `0.15.14` 🙈

JClackett:
What's the latest with this? still getting:

`Expected ";" but found "satisfies"`

when running `remix build`

using:
`remix: 1.8.0`
`esbuild: 0.15.16`

pcattori:
We're actively working on it, see https://github.com/remix-run/remix/pull/4754 .

Currently, the esbuild upgrade is causing some integration tests to hang indefinitely.
Once we have that fixed, we'll proceed with the upgrade to esbuild 0.15.13+.

PR #4675, opened 2022-11-23
docs(tutorials/blog): add TS 4.9's `satisfies` in combination with `LoaderFunction`

tom-sherman:
Closes: #

- [x] Docs
- [ ] Tests

Testing Strategy:

I don't think this is needed as it's just a utility type. The interesting part happens in userland when we use the `satisfies` keyword.

---

Right now because of the way you need to add types for loaders you can get into a situation where the return type of `useLoaderData` is `never`:

```ts
const loader = () => "not a response";
//    ^? never
```

ideally we'd like typescript to error when we return the string instead of a Response. We can use the new `satisfies` keyword for this:

```ts
// ERROR: Type 'number' is not assignable to type 'Response'
const loader = (() => "not a response") satisfies LoaderFunction;
```

The loader now has a type error but it still allows for the inference of the return type to work with `useLoaderData`.


packages/remix-server-runtime/router/utils.ts
```diff
@@ -114,6 +114,24 @@ export interface ActionFunction {
...
+ *
+ * ```ts
+ * export const loader = (({ request }) => {
+ *   const url = new URL(request.url);
+ *   return new Response("hello from " + url.pathname);
+ * }) satisfies LoaderFunction;
+ * ```
+ */
+export interface LoaderFunction {
+  (args: LoaderFunctionArgs): Promise<Response> | Response;
```
tom-sherman, line 132: We may need to iterate on the return type a little.

Eg. do we want to allow loader functions to return `null | Response`?

packages/remix-server-runtime/routeModules.ts
```diff
@@ -30,6 +30,24 @@ export interface ActionFunction {
...
+ *
+ * ```ts
+ * export const loader = (({ request }) => {
+ *   const url = new URL(request.url);
+ *   return new Response("hello from " + url.pathname);
+ * }) satisfies LoaderFunction;
+ * ```
+ */
+export interface LoaderFunction {
+  (args: DataFunctionArgs): Promise<Response> | Response;
```
tom-sherman, line 48: We may need to iterate on the return type a little.

Eg. do we want to allow loader functions to return `null | Response`?

brophdawg11: This already exists later on this this file?

tom-sherman: Oh! I don't have this, maybe because I'm on an experimental remix build (deferred channel)

tom-sherman:
Ok so the Typescript change can be reverted, I got confused because the version of remix I have doesn't have a LoaderFunction type.

@brophdawg11 do you think the docs change is valuable at all?

packages/remix-server-runtime/routeModules.ts
brophdawg11: ah that makes sense :)

sergiodxa:
Question, can you use `satisfies` without arrow functions?

brophdawg11:
@tom-sherman actually, since you can technically return anything (`type AppData = any`)  - does satisfies even help in that case?  It might be something where it only helps if you define a more narrowly-typed loader function on your own?

tom-sherman:
@sergiodxa Unfortunately not, it appears the syntax doesn't support it right now  - you can only use `satisfies` with an expression.

@brophdawg11 Oh you're absolutely right 🤦 in which case I don't think this even deserves a place in the docs. It might make sense as a more editorialised blog post I guess (I'm already writing one on my own blog)

tom-sherman:
@sergiodxa A workaround is something like this:

```ts
type Fn = () => number;

function foo() { return "ds" }

foo satisfies Fn; // type error
```

Not recommended tho because compilers will likely output a `foo;` statement.

pcattori:
We're already working on upgrading esbuild to 0.15.13 for `satisfies` support. You can check https://github.com/remix-run/remix/issues/4639 for progress on that.

pcattori:
> @tom-sherman actually, since you can technically return anything (`type AppData = any`) - does satisfies even help in that case? It might be something where it only helps if you define a more narrowly-typed loader function on your own?

There's still value in using `satisfies` even since it will make sure the function params are correct for example.

tom-sherman:
> One `function`-compatible option is:
> 
> ```ts
> (function hello() {
>     return "world"
> }) satisfies () => number;
> ```

@pcattori that's a function expression/anonymous function though. You can't do this for example

```ts
export (function hello() {
    return "world"
}) satisfies () => number;
```

pcattori:
> @pcattori that's a function expression/anonymous function though. You can't do this for example
> 
> ```ts
> export (function hello() {
>     return "world"
> }) satisfies () => number;
> ```

Yea, if you want (1) `export` on the same line as the definition and (2) to use `satisfies`, then I don't think there's a way to do it with a named function.

The only way would be:

```ts
export const hello = (function () {
    return "world"
}) satisfies () => number;
```

but the main benefit of named functions is that they can be referenced before their definition, which you would lose out on with this^.

Personally, I prefer to use arrow functions, but yes its unfortunate for those who use `function` that `satisfies` is not supported better for it AFAIK.

packages/remix-server-runtime/routeModules.ts
pcattori: @tom-sherman if you can remove changes to this file, the other doc changes you've made look good to me!

tom-sherman:
Btw, shameless plug 😅 https://tom-sherman.com/blog/remix-typescript-satisfies

pcattori:
> https://tom-sherman.com/blog/remix-typescript-satisfies

FYI we export a `LoaderFunction` type from `@remix-run/node` (and the other runtime packages too), so you can use that instead of defining `Loader` yourself: `satisfies LoaderFunction`

pcattori (approved):

pcattori:
Looks like tests are failing because eslint can't parse Typescript 4.9 `satisfies` correctly. Probably need to upgrade esbuild or the typescript plugin or something like that.

docs/tutorials/blog.md
```diff
@@ -192,6 +192,33 @@ export default function Posts() {
 
 Hey, that's pretty cool. We get a pretty solid degree of type safety even over a network request because it's all defined in the same file. Unless the network blows up while Remix fetches the data, you've got type safety in this component and its API (remember, the component is already its own API route).
 
+We can take the type safety a step further in TypeScript 4.9.0 using the `satisfies` keyword. This will ensure that we always return a type of `Response` from the loader instead of accidentally returning another type which would result in a runtime error.
+
+```ts filename=app/routes/posts/index.tsx lines=[1,22]
+import { json, LoaderFunction } from "@remix-run/node";
```
MichaelDeBoey, line 198: We put types on a separate line in our docs/examples

```suggestion
import type { LoaderFunction } from "@remix-run/node";
import { json } from "@remix-run/node";
```

docs/tutorials/blog.md
```diff
@@ -192,6 +192,33 @@ export default function Posts() {
 
 Hey, that's pretty cool. We get a pretty solid degree of type safety even over a network request because it's all defined in the same file. Unless the network blows up while Remix fetches the data, you've got type safety in this component and its API (remember, the component is already its own API route).
 
+We can take the type safety a step further in TypeScript 4.9.0 using the `satisfies` keyword. This will ensure that we always return a type of `Response` from the loader instead of accidentally returning another type which would result in a runtime error.
+
+```ts filename=app/routes/posts/index.tsx lines=[1,22]
```
MichaelDeBoey, line 197:
```suggestion
```ts filename=app/routes/posts/index.tsx lines=[1,23]
```

MichaelDeBoey (changes requested):

tom-sherman:
> I guess you already knew that since this PR uses LoaderFunction in the docs 😅

@pcattori Weird one, I wasn't actually aware of that type when I wrote the blob because I'm on an experimental channel (deffered) that doesn't have that type 😆

docs/tutorials/blog.md
```diff
@@ -195,7 +195,8 @@ Hey, that's pretty cool. We get a pretty solid degree of type safety even over a
 We can take the type safety a step further in TypeScript 4.9.0 using the `satisfies` keyword. This will ensure that we always return a type of `Response` from the loader instead of accidentally returning another type which would result in a runtime error.
 
 ```ts filename=app/routes/posts/index.tsx lines=[1,22]
```
machour:
```suggestion
```ts filename=app/routes/posts/index.tsx lines=[1,23]
```

machour: Weird, I was seeing an outdated version

moishinetzer:
Is there any updated best practice/solution to solve this?

I want to be able to take advantage of `useLoaderData<typeof loader>()` AND make sure that my loader functions return the correct types.

tom-sherman:
@moishinetzer The simplest way to do that would be to add a return type annotation to the loader eg.

```ts
export async function loader({ context }: LoaderArgs): Promise<TypedResponse<{ posts: Post[], tags: string[] }>> {
  const blog = new BlogData(context);

  const [posts, tags] = await Promise.all([
    blog.listAllPosts(),
    blog.listAllTags(),
  ]);

  return json({
    posts,
    tags,
  });
}

const { posts, tags } = useLoaderData<typeof loader>();
```

You can use `satisfies` if you wanted to assert a partial data shape at the loader level and have the concrete return type inferred.

machour (changes requested):
@tom-sherman Could you rebase this to target `main`?
Documentation & templates updates go directly to main

docs/tutorials/blog.md
```diff
@@ -195,6 +195,34 @@ export default function Posts() {
 
 Hey, that's pretty cool. We get a pretty solid degree of type safety even over a network request because it's all defined in the same file. Unless the network blows up while Remix fetches the data, you've got type safety in this component and its API (remember, the component is already its own API route).
 
+We can take the type safety a step further in TypeScript 4.9.0 using the `satisfies` keyword. This will ensure that we always return a type of `Response` from the loader instead of accidentally returning another type which would result in a runtime error.
+
+```ts filename=app/routes/posts/index.tsx lines=[1,23]
```
MichaelDeBoey, line 200:
```suggestion
```ts filename=app/routes/posts/index.tsx
```

docs/tutorials/blog.md
```diff
@@ -195,6 +195,34 @@ export default function Posts() {
...
+import type { LoaderFunction } from "@remix-run/node";
+import { json } from "@remix-run/node";
```
MichaelDeBoey, line 202:
```suggestion
import type { LoaderFunction } from "@remix-run/node"; // or cloudflare/deno
import { json } from "@remix-run/node"; // or cloudflare/deno
```

MichaelDeBoey (changes requested):

pcattori:
Looks like the upstream issues with eslint plugins and TS 4.9 have been resolved as `yarn lint` passes locally for me with the suggested edits.

That said, we are no longer using this tutorial in the v2 docs, so closing this PR.

PR #4401, opened 2022-10-21
feat(remix-dev): Add support for Cloudflare Module Workers

buzinas:
## What

Closes: #4399

- [ ] Docs
- [ ] Tests

See https://developers.cloudflare.com/workers/learning/migrating-to-module-workers/ for more info about Module Workers and https://github.com/cloudflare/kv-asset-handler#asset_manifest-required-for-es-modules for how to set it up.

Testing Strategy:

This is a patch I've been using locally in order to build for Cloudflare Module Workers. `__STATIC_CONTENT_MANIFEST` module is needed in order to provide it to `handleAsset`.

PR #4676, opened 2022-11-23
feat(adapters): add `cloudflare-workers-esm` adapter

huw:
Closes: #764
Supersedes: #4401 #4198 #1207

- [x] Docs
- [x] Tests

## Testing Strategy:

- I imported the compiled packages locally into my full-stack Remix project (using npm local file path resolution), then updated my `server/index.ts` file to use ES Modules. I checked the compiled code to ensure it was using the new dependencies as well as the new ESM code. I then ran it in `miniflare@2.11.0` and deployed to Cloudflare, where everything worked.

I wasn’t able to run the deploy test because I don’t have access to Remix’s Cypress server or their production Cloudflare account.

## Additional Notes
The main blocker for past attempts at this is that `@cloudflare/kv-asset-handler` requires the code to import a manifest file (as a JSON blob) from a special URL (`__STATIC_CONTENT_MANIFEST`). We can’t let esbuild try to bundle this, but if we try to add the import to `@remix-run/cloudflare-workers` (in order to try and interop), then esbuild leaves a raw ES6 `import` in the CJS code, which will fail to compile for CJS users!

In order to keep build (and code) complexity at a minimum, I think it makes the most sense to avoid interoperability & build this package separately, as a new adapter. This is similar to how Cloudflare Pages is split up, despite using an extremely similar architecture under the hood.

There are a few more scrutinable choices I made that I want to call out, in case the Remix or Workers teams disagree:

1. In code, I’ve preferred the naming scheme `cloudflare-workers` vs `cloudflare-workers-esm`, because this means we don’t have to edit or deprecate the existing package, and the new one has a relatively short name.
2. In docs, I’ve preferred Cloudflare’s naming scheme of ‘Service Worker syntax’ vs ‘Module Worker syntax’, but also clarified it as ‘Module Worker syntax (ESM)’ where that makes sense, just to reduce confusion as much as possible.
3. I didn’t add/update any tests beyond the deploy test, because `cloudflare-pages` didn’t.
4. I didn’t update the `--platform` error to include this package, since it was never available via that flag. But it might cause some extra confusion if someone knows about the flag and tries to use it?
5. For the same reason, I didn’t update the migration/codemods for the old Remix imports.
6. I listed this as a minor bump in the changeset, but I couldn’t find an example of a new adapter being added in the changelog, so I’m not fully sure what the Remix team wants for this.

Cheers! I’m around as much as you need me 🍻

cc @dario-piotrowicz @petebacondarwin


packages/remix-cloudflare-workers-esm/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# `@remix-run/cloudflare-workers-esm`
```
huw, line 1: Admittedly not sure what to do with `CHANGELOG.md` in a new package, but I assume the release process solves this.

docs/other-api/adapter.md
```diff
@@ -117,7 +118,17 @@ exports.handler =
...
-Here's an example with the simplified Cloudflare Workers API:
+Here's an example with the simplified Cloudflare Workers API (Module Worker / ES Modules syntax). This syntax supports additional Cloudflare Workers features, such as Durable Objects (you can access these bindings in the Remix context using the `getLoadContext(request, env, ctx)` option on `createEventHandler`). For more information on the difference between Module Workers and Service Workers, [see Cloudflare's docs][cloudflare-module-workers]):
+
+```ts
+import { createEventHandler } from "@remix-run/cloudflare-workers-esm";
+
+import * as build from "../build";
+
+export default { fetch: createEventHandler({ build }) };
+```
```
huw, line 129: I’m not sure if this is sufficient. Some users browsing the docs might want to know how to bind a Durable Object with an example. On the other hand, this is a good enough starting point and I think the community can fill it in.

packages/remix-cloudflare-workers-esm/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "include": ["**/*.ts"],
+  "exclude": ["dist", "__tests__", "node_modules"],
+  "compilerOptions": {
+    "lib": ["ES2019"],
+    "target": "ES2019",
+    "types": ["@types/node", "@cloudflare/workers-types"],
```
huw, line 7: Perhaps controversial, but I moved the `/// <reference />` out of `worker.ts` and into `tsconfig.json`. This is [best practice according to Microsoft](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-).

In an ideal world we’d probably just do:

```ts
declare process {
  env: {
    NODE_ENV?: string;
  }
}
```

packages/remix-cloudflare-workers-esm/worker.ts
```diff
@@ -0,0 +1,174 @@
...
+      error instanceof NotFoundError
+    ) {
+      return null;
+    }
+
+    throw error;
+  }
+}
+
+export function createEventHandler<Env = unknown>({
```
huw, line 121: I’ve parameterised `Env` here to allow the developer to supply their own contract as necessary. Better than trying to override it.

packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts
```diff
@@ -98,6 +98,12 @@ export function serverBareModulesPlugin(
           case "cloudflare-workers":
           case "deno":
             return undefined;
+          case "cloudflare-workers-esm": {
+            // If we're using Workers Modules (ESM), bundle everything except the special asset manifest import
+            if (path !== "__STATIC_CONTENT_MANIFEST") {
+              return undefined;
```
huw, line 104: I figured this was preferable to setting `external` in `compilerServer.ts`, but let me know.

packages/remix-cloudflare-workers-esm/worker.ts
```diff
@@ -0,0 +1,174 @@
...
+  getLoadContext?: GetLoadContextFunction<Env>;
+  mode?: string;
+}) {
+  let handleRequest = createRemixRequestHandler(build, mode);
+
+  return (request: Request, env: Env, ctx: ExecutionContext) => {
+    let loadContext = getLoadContext?.(request, env, ctx);
+
+    return handleRequest(request, loadContext);
+  };
```
dario-piotrowicz, line 51: are the `let`s correct? 🤔 
```suggestion
  const handleRequest = createRemixRequestHandler(build, mode);

  return (request: Request, env: Env, ctx: ExecutionContext) => {
    const loadContext = getLoadContext?.(request, env, ctx);

    return handleRequest(request, loadContext);
  };
```

packages/remix-cloudflare-workers-esm/worker.ts
```diff
@@ -0,0 +1,174 @@
...
+
+export async function handleAsset(
+  request: Request,
+  waitUntil: WaitUntilCallback,
+  assetNamespace: KVNamespace<string>,
+  build: ServerBuild,
+  options?: Partial<KvAssetHandlerOptions>
+) {
+  try {
+    let mergedOptions = {
```
dario-piotrowicz, line 62:
```suggestion
    const mergedOptions = {
```

packages/remix-cloudflare-workers-esm/worker.ts
```diff
@@ -0,0 +1,174 @@
...
+          },
+          ...mergedOptions,
+        }
+      );
+    }
+
+    let cacheControl = {};
+    let url = new URL(request.url);
+    let assetpath = build.assets.url.split("/").slice(0, -1).join("/");
+    let requestpath = url.pathname.split("/").slice(0, -1).join("/");
```
dario-piotrowicz, line 83:
```suggestion
    const url = new URL(request.url);
    const assetpath = build.assets.url.split("/").slice(0, -1).join("/");
    const requestpath = url.pathname.split("/").slice(0, -1).join("/");
```

packages/remix-cloudflare-workers-esm/worker.ts
```diff
@@ -0,0 +1,174 @@
...
+  getLoadContext,
+  mode,
+}: {
+  build: ServerBuild;
+  getLoadContext?: GetLoadContextFunction<Env>;
+  mode?: string;
+}): ExportedHandlerFetchHandler<
+  Env & { __STATIC_CONTENT: KVNamespace<string> }
+> {
+  let handleRequest = createRequestHandler({
```
dario-piotrowicz, line 132:
```suggestion
  const handleRequest = createRequestHandler({
```

packages/remix-cloudflare-workers-esm/worker.ts
```diff
@@ -0,0 +1,174 @@
...
+    build,
+    getLoadContext,
+    mode,
+  });
+
+  let handleEvent = async (
```
dario-piotrowicz, line 138:
```suggestion
  const handleEvent = async (
```

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+  getAppDirectory,
+  getAppName,
+  getSpawnOpts,
+  runCypress,
+  validatePackageVersions,
+} from "./_shared.mjs";
+
+let APP_NAME = getAppName("cf-workers-esm");
+let PROJECT_DIR = getAppDirectory(APP_NAME);
+let CYPRESS_DEV_URL = "http://localhost:8787";
```
dario-piotrowicz, line 22:
```suggestion
const APP_NAME = getAppName("cf-workers-esm");
const PROJECT_DIR = getAppDirectory(APP_NAME);
const CYPRESS_DEV_URL = "http://localhost:8787";
```

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+    projectDir: PROJECT_DIR,
+  });
+}
+
+async function createAndDeployApp() {
+  // create a new remix app
+  await createNewApp();
+
+  // validate dependencies are available
+  let [valid, errors] = await validatePackageVersions(PROJECT_DIR);
```
dario-piotrowicz, line 38:
```suggestion
  const [valid, errors] = await validatePackageVersions(PROJECT_DIR);
```

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+
+  // add cypress to the project
+  await Promise.all([
+    fse.copy(CYPRESS_SOURCE_DIR, path.join(PROJECT_DIR, "cypress")),
+    fse.copy(CYPRESS_CONFIG, path.join(PROJECT_DIR, "cypress.json")),
+    addCypress(PROJECT_DIR, CYPRESS_DEV_URL),
+    pkgJson.save(),
+  ]);
+
+  let spawnOpts = getSpawnOpts(PROJECT_DIR, {
```
dario-piotrowicz, line 61:
```suggestion
  const spawnOpts = getSpawnOpts(PROJECT_DIR, {
```

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+  // run cypress against the dev server
+  runCypress(PROJECT_DIR, true, CYPRESS_DEV_URL);
+
+  // we need to update the workers name
+  let wranglerTomlPath = path.join(PROJECT_DIR, "wrangler.toml");
+  let wranglerTomlContent = await fse.readFile(wranglerTomlPath);
+  let wranglerToml = toml.parse(wranglerTomlContent);
+  wranglerToml.name = APP_NAME;
+  await fse.writeFile(wranglerTomlPath, toml.stringify(wranglerToml));
+  let url = `https://${APP_NAME}.remixrun.workers.dev`;
```
dario-piotrowicz, line 82:
```suggestion
  const wranglerTomlPath = path.join(PROJECT_DIR, "wrangler.toml");
  const wranglerTomlContent = await fse.readFile(wranglerTomlPath);
  const wranglerToml = toml.parse(wranglerTomlContent);
  wranglerToml.name = APP_NAME;
  await fse.writeFile(wranglerTomlPath, toml.stringify(wranglerToml));
  const url = `https://${APP_NAME}.remixrun.workers.dev`;
```

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+  console.log(`worker url: ${url}`);
```
dario-piotrowicz, line 83: was this used for debugging and it should be removed? or it is actually here on purpose?

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+
+  spawnSync("npx", ["wrangler", "--version"], spawnOpts);
+
+  // deploy the app
+  let deployCommand = spawnSync("npx", ["wrangler", "publish"], spawnOpts);
```
dario-piotrowicz, line 88:
```suggestion
  const deployCommand = spawnSync("npx", ["wrangler", "publish"], spawnOpts);
```

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+    console.error(deployCommand.error);
+    throw new Error("Cloudflare Module Workers (ESM) deploy failed");
+  }
+
+  // run cypress against the deployed app
+  runCypress(PROJECT_DIR, false, url);
+}
+
+async function destroyApp() {
+  let result = await fetch(
```
dario-piotrowicz, line 99:
```suggestion
  const result = await fetch(
```

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+    `https://api.cloudflare.com/client/v4/accounts/${process.env.CLOUDFLARE_ACCOUNT_ID}/workers/scripts/${APP_NAME}?force=true`,
+    {
+      method: "DELETE",
+      headers: {
+        "X-Auth-Email": process.env.CLOUDFLARE_EMAIL,
+        "X-Auth-Key": process.env.CLOUDFLARE_GLOBAL_API_KEY,
+      },
+    }
+  );
+  let json = await result.json();
```
dario-piotrowicz, line 110:
```suggestion
  const json = await result.json();
```

scripts/deployment-test/cf-workers-esm.mjs
```diff
@@ -0,0 +1,128 @@
...
+  console.log(`[DESTROY_APP]`, json);
```
dario-piotrowicz, line 111: is this `console.log` intentional? 🙂

dario-piotrowicz (approved):
Amazing stuff @huw thanks so very much for the hard work you put into this 🙏 

(you did mention that you were busy so I wasn't expecting at all such quick turnaround! 🤯)

I'm not really familiar with the codebase but I've put some comments here and there, hopefully some could turn out to be useful, feel free to ignore them if they aren't valid

Thanks so very much again! 😍

packages/remix-cloudflare-workers-esm/worker.ts
huw: Yeah, this is what their ESLint config prefers. Not sure about the reasoning but I got autocorrected into it.

scripts/deployment-test/cf-workers-esm.mjs
huw: This is just a copy of the same script for Service Workers. I assume it's intentional so that it shows up in the GitHub Actions logs when it runs.

packages/remix-cloudflare-workers-esm/worker.ts
dario-piotrowicz: ah.... weird :sweat:

machour: Remix code base generally use `let` instead of `const` : https://dev.to/ryanflorence/comment/1k697

dario-piotrowicz: ah... I see :open_mouth: 

Thanks for the link, very informative :slightly_smiling_face:

dario-piotrowicz:
quick ping, is anything blocking this PR? 🙂

MichaelDeBoey:
@huw @dario-piotrowicz @petebacondarwin I'm wondering if you could obtain the same result with the new config properties that were made public by #4841 🤔

huw:
Got it working locally with the new properties--no need to make changes to the core Remix codebase. I'll post info tomorrow and formally close the PR then.

We'll still have to publish a separate adapter package for Cloudflare ESM, otherwise users would have to add a lot of boilerplate to their codebase. My only concern here is that users could get confused between the Remix-owned worker adapters and the community/Cloudflare-owned adapters. I would suggest that either Remix own all of them, or Cloudflare own all of them. But I work for neither org so can't really make that conversation happen beyond politely asking for it :) (cc @MichaelDeBoey @dario-piotrowicz)

dario-piotrowicz:
@huw thanks so much for the hard work :slightly_smiling_face: 

At Cloudflare we aren't really keen on owning adapters ourselves (if we can help it), since we support many different frameworks and owning an maintaining adapters for all of them would be very costly (most frameworks do provide their own Cloudflare adapters).

As I mentioned we might do it when other alternatives are not present and there is a high enough demand for us supporting a framework.

So based on what I just mentioned our preferred way would definitely for Remix to own the Cloudflare adapters (and that's been the case so far, so I am not sure why now the Cloudflare adapters should be externalized), we can anyways contribute and help maintaining them if need be.

If that is not possible we implement and own our own adapters (but that's not a decision for me to make).

I'd be very interested to hear what @MichaelDeBoey has to say about it on behalf of the Remix team, is there a clear need/intention to externalize the Cloudflare adapters?

kiliman:
Perhaps we need a `@remix-community` scope in `npm`. Something that's not official, but endorsed by the Remix team.

MichaelDeBoey:
@huw @dario-piotrowicz I don't work for the Remix team, I only contribute & help triaging things here on GitHub

huw:
Code is:

For `src/server/createEventHandler.ts` (this is pretty well exactly the same as the code in this PR):

```ts
import moduleJSON from "__STATIC_CONTENT_MANIFEST";
import {
  getAssetFromKV,
  MethodNotAllowedError,
  NotFoundError,
  type Options as KvAssetHandlerOptions,
} from "@cloudflare/kv-asset-handler";
import {
  type AppLoadContext,
  createRequestHandler as createRemixRequestHandler,
  type ServerBuild,
} from "@remix-run/cloudflare";

type AssetManifest = Omit<KvAssetHandlerOptions["ASSET_MANIFEST"], string>;
type KvAssetHandlerEvent = Parameters<typeof getAssetFromKV>[0];
type WaitUntilCallback = KvAssetHandlerEvent["waitUntil"];
interface StaticContentEnvironment { __STATIC_CONTENT: KVNamespace<string> }

const module = JSON.parse(moduleJSON) as AssetManifest;

/**
 * A function that returns the value to use as `context` in route `loader` and
 * `action` functions.
 *
 * You can think of this as an escape hatch that allows you to pass
 * environment/platform-specific values through to your loader/action.
 */
type GetLoadContextFunction<Environment = unknown> = (
  request: Request,
  environment: Environment,
  context: ExecutionContext
) => AppLoadContext;

/**
 * Returns a request handler for the Cloudflare runtime that serves the
 * Remix SSR response.
 */
const createRequestHandler = <Environment extends StaticContentEnvironment = StaticContentEnvironment>({
  build,
  getLoadContext,
  mode,
}: {
  build: ServerBuild;
  getLoadContext?: GetLoadContextFunction<Environment>;
  mode?: string;
}) => {
  const handleRequest = createRemixRequestHandler(build, mode);

  return (request: Request, environment: Environment, context: ExecutionContext) => {
    const loadContext = getLoadContext?.(request, environment, context);

    return handleRequest(request, loadContext);
  };
};

const handleAsset = async (
  request: Request,
  waitUntil: WaitUntilCallback,
  assetNamespace: KVNamespace<string>,
  build: ServerBuild,
  options?: Partial<KvAssetHandlerOptions>
) => {
  try {
    const mergedOptions = {
      ASSET_NAMESPACE: assetNamespace,
      ASSET_MANIFEST: module,
      ...options,
    };

    if (process.env.NODE_ENV === "development") {
      return await getAssetFromKV(
        { request, waitUntil },
        {
          cacheControl: {
            bypassCache: true,
          },
          ...mergedOptions,
        }
      );
    }

    let cacheControl = {};
    const assetpath = build.assets.url.split("/").slice(0, -1).join("/");
    const requestpath = url.pathname.split("/").slice(0, -1).join("/");

    if (requestpath.startsWith(assetpath)) {
      // Assets are hashed by Remix so are safe to cache in the browser
      // And they're also hashed in KV storage, so are safe to cache on the edge
      cacheControl = {
        bypassCache: false,
        edgeTTL: 31_536_000,
        browserTTL: 31_536_000,
      };
    } else {
      // Assets are not necessarily hashed in the request URL, so we cannot cache in the browser
      // But they are hashed in KV storage, so we can cache on the edge
      cacheControl = {
        bypassCache: false,
        edgeTTL: 31_536_000,
      };
    }

    return await getAssetFromKV(
      { request, waitUntil },
      {
        cacheControl,
        ...mergedOptions,
      }
    );
  } catch (error) {
    if (error instanceof MethodNotAllowedError || error instanceof NotFoundError) {
      return null;
    }

    throw error;
  }
};

const createEventHandler = <Environment extends StaticContentEnvironment = StaticContentEnvironment>({
  build,
  getLoadContext,
  mode,
}: {
  build: ServerBuild;
  getLoadContext?: GetLoadContextFunction<Environment>;
  mode?: string;
}): ExportedHandlerFetchHandler<Environment> => {
  const handleRequest = createRequestHandler({
    build,
    getLoadContext,
    mode,
  });

  const handleEvent = async (
    request: Request,
    environment: Environment & StaticContentEnvironment,
    context: ExecutionContext
  ) => {
    let response = await handleAsset(request, context.waitUntil, environment.__STATIC_CONTENT, build);

    if (!response) {
      response = await handleRequest(request, environment, context);
    }

    return response;
  };

  return (
    request: Request,
    environment: Environment & { __STATIC_CONTENT: KVNamespace<string> },
    context: ExecutionContext
  ) => {
    try {
      return handleEvent(request, environment, context);
    } catch (error: unknown) {
      if (process.env.NODE_ENV === "development" && error instanceof Error) {
        return new Response(error.message, {
          status: 500,
        });
      }

      return new Response("Internal Error", { status: 500 });
    }
  };
};

export default createEventHandler;
```

For `global.d.ts`/`manifest.d.ts`:

```ts
declare module "__STATIC_CONTENT_MANIFEST" {
  const value: string;
  export default value;
}
```

For `index.ts`:

```ts
import createEventHandler from "./createEventHandler";

interface Environment {
  __STATIC_CONTENT: KVNamespace<string>;
}

const index = {
  fetch: createEventHandler<Environment>({ build })
};

export default index;
```

And finally, `remix.config.js` (in ES Module syntax, since I’m assuming the package is `type: “module”`, but you could equally write this in CJS if you wanted):

```js
/**
 * @type {import('@remix-run/dev').AppConfig}
 */
const config = {
  // Bundles everything _except_ `__STATIC_CONTENT_MANIFEST`. Be careful if editing.
  serverDependenciesToBundle: [/^(?!(__STATIC_CONTENT_MANIFEST)$).*$/u],
  // Optional, but highly recommended for fitting within the Worker bundle size
  serverMinify: true,
  serverModuleFormat: "esm",
  // YMMV with setting to `node`, most likely anything but simple code won’t work with the CF node compat layer
  serverPlatform: "neutral",
  // `workerd` is now the standard Workers main field, then follow the others in order
  serverMainFields: ["workerd", "browser", "module", "main"],
  // Even though we’re not in a Service Worker anymore, this is at least required for React SSR, otherwise it’ll assume we’re in a Node environment
  serverConditions: ["worker"],
};

export default config;
```

You can get this working today in an ESM repo on `v0.0.0-nightly-79859c9-20230203`. My recommendation is that, like with the existing adapters, the `createEventHandler` file is managed in its own package, but it’s not _that much_ boilerplate. I will close this PR while CF & Remix discuss ^_^

petebacondarwin:
Thanks for your work on this @huw. It looks like it would be fairly straightforward to implement this adapter.
Let's push forward with discussions on how it can best be published and maintained.

xanderberkein:
I also took a shot at creating a Cloudflare Module Worker adapter and came across this PR while trying to expose `__STATIC_CONTENT_MANIFEST` to Cloudflare's asset handler.

I managed to get Remix working on Cloudlare's module Workers with even less boilerplate. Take a look at my [`worker.ts` file](https://github.com/xanderberkein/remix-cloudflare-module-workers/blob/main/src/worker.ts); there's no need to reimplement `handleAsset` and `createRequestHandler`.

I published my code as a very lightweight adapter, [`remix-cloudflare-module-workers`](https://github.com/xanderberkein/remix-cloudflare-module-workers), and included some additional docs on integrating Workers APIs with Remix, as there aren't many resources available on this topic yet.

Since the adapter has such a low footprint, I'm happy to maintain it until Cloudflare or Remix decides to release an official one.

PR #4686, opened 2022-11-25
chore(remix-eslint-config): upgrade typescript eslint plugin for satisfies keyword

fredericoo:
This PR simply upgrades `@typescript-eslint/eslint-plugin` to v5.44.0, which is the [first one to offer support for the `satisfies` keyword on Typescript](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v5.44.0).


MichaelDeBoey:
Thanks for taking the time to update this dependency @fredericoo!

However, I'm going to close this as you will already have this version when doing a clean install.

PR #4697, opened 2022-11-27
feat(remix-eslint-config): warn on possible leaked render

MichaelDeBoey:
Not doing this can result in bugs.
More info in @kentcdodds' blogpost: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

### DIFF
--- a/packages/remix-eslint-config/rules/react.js
+++ b/packages/remix-eslint-config/rules/react.js
@@ -7,6 +7,7 @@ module.exports = {
   "react/forbid-foreign-prop-types": [WARN, { allowInPropTypes: true }],
   "react/jsx-key": WARN,
   "react/jsx-no-comment-textnodes": WARN,
+  "react/jsx-no-leaked-render": [WARN, { validStrategies: ["ternary"] }],
   "react/jsx-no-target-blank": WARN,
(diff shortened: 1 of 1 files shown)


machour:
Agreed with @chaance. I already vouched against this in the `examples` repo: https://github.com/remix-run/examples/pull/90#discussion_r1033003957

chaance:
I also disagree with adding lint rules that result in new errors without a major update **unless absolutely necessary** (deprecation warnings, as in the case of magic exports, I think was a valid exception). While it's arguably not breaking, seeing a bunch of errors in a minor or patch release is unexpected -- particularly based on a more opinionated rules.

pcattori:
Most of the rule we currently have are either:
(a) part of the `recommended` configs for `eslint` or `eslint-plugin-react` (like `react/jsx-key`)
(b) necessary for Remix conventions or compilation (like `react/display-name`)

`jsx-no-leaked-render` doesn't fall under either of those criteria.

Since `jsx-no-leaked-render` is a heuristic (since there are use-cases where `&&` is valid for rendering) and not something that is _definitely_ a bug, I don't think we should make an exception for it.

Agreed that we should not be opinionated about this.

PR #4703, opened 2022-11-29
RRR Step 3 - Hydration

brophdawg11:
Initial work towards step 3 of [React Router-ing Remix](https://remix.run/blog/react-routering-remix)

* Client hydration via `createBrowserRouter`/`RouterProvider`

Sibling RR PR: https://github.com/remix-run/react-router/pull/9664

For local app testing:
* Install `latest` into your local remix app
* Remove duped dependencies
  * `rm -rf node_modules/@remix-run/server-runtime/node_modules`
  * `rm -rf node_modules/@remix-run/react/node_modules`
  * `rm -rf node_modules/react-router-dom/node_modules`
  * `rm -rf node_modules/react-router/node_modules`
* Build this Remix branch into your local remix app:
  * `REMIX_LOCAL_BUILD_DIRECTORY=../path/to/your/app yarn build`
* Build the corresponding RR branch from the above PR into your local remix app:
  * `LOCAL_BUILD_DIRECTORY=../path/to/your/app yarn build`
* Run your local Remix app with `npm run dev`


integration/transition-state-test.ts
```diff
@@ -254,9 +256,11 @@ test.describe("rendering", () => {
...
           hash: "",
           state: {
-            isRedirect: true,
-            setCookie: false,
-            type: "loader",
+            _isRedirect: true,
+            // These were private API for transition manager that are no longer
+            // needed with the new router so OK to disappear
+            // setCookie: false,
+            // type: "loader",
```
brophdawg11, line 263: Passing integration tests from `dev` change slightly since we use new private API on location state to detect `type`

packages/remix-react/browser.tsx
```diff
@@ -1,64 +1,61 @@
...
 
-  let entryContext = window.__remixContext;
-  entryContext.manifest = window.__remixManifest;
-  entryContext.routeModules = window.__remixRouteModules;
-  // In the browser, we don't need this because a) in the case of loader
-  // errors we already know the order and b) in the case of render errors
-  // React knows the order and handles error boundaries normally.
-  entryContext.appState.trackBoundaries = false;
-  entryContext.appState.trackCatchBoundaries = false;
+    router = createBrowserRouter(routes, { hydrationData });
```
brophdawg11, line 47: Create our client side `router` singleton

packages/remix-react/browser.tsx
```diff
@@ -1,64 +1,61 @@
...
-    />
+    <RemixContext.Provider
+      value={{
+        manifest: window.__remixManifest,
+        routeModules: window.__remixRouteModules,
+        future: window.__remixContext.future,
+      }}
+    >
+      <RouterProvider router={router} fallbackElement={null} />
+    </RemixContext.Provider>
```
brophdawg11, line 59: Render `<RouterProvider>` directly

packages/remix-react/components.tsx
```diff
@@ -764,6 +796,7 @@ import(${JSON.stringify(manifest.entry.module)});`;
         />
         <script
           {...props}
+          suppressHydrationWarning
```
brophdawg11, line 799: Any reason this wasn't there already in dev?  We added it while debugging some other integration test hydration failures, which I think came down to the lack of a `<head>` in our `root.tsx` but this felt like a good change?

packages/remix-react/components.tsx
```diff
@@ -904,11 +954,203 @@ export function useActionData<T = AppData>(): SerializeFrom<T> | undefined {
...
+  let lastTransitionRef = React.useRef<Transition>();
+
+  if (lastTransitionRef.current && lastNavigationRef.current === navigation) {
+    return lastTransitionRef.current;
+  }
+
+  lastNavigationRef.current = navigation;
+  lastTransitionRef.current = convertNavigationToTransition(navigation);
+
+  return lastTransitionRef.current;
```
brophdawg11, line 976: Concurrent mode prevents `useMemo` here since it re-runs even if the `navigation` doesn't change and creates a new `transition` object identity 😖

packages/remix-react/components.tsx
```diff
@@ -904,11 +954,203 @@ export function useActionData<T = AppData>(): SerializeFrom<T> | undefined {
...
+}
+
+function convertNavigationToTransition(navigation: Navigation): Transition {
```
brophdawg11, line 979: Mega if/else to convert navigations to transitions

packages/remix-react/components.tsx
```diff
@@ -929,16 +1171,177 @@ export function useFetcher<TData = any>(): FetcherWithComponents<
...
+): Fetcher {
+  let { state, formMethod, formAction, formEncType, formData, data } =
+    fetcherRR;
+
+  let isActionSubmission =
+    formMethod != null &&
+    ["POST", "PUT", "PATCH", "DELETE"].includes(formMethod.toUpperCase());
+
+  if (state === "idle") {
+    if (fetcherRR[" _hasFetcherDoneAnything "] === true) {
```
brophdawg11, line 1202: Needed _something_ here to differentiate fetcher `idle`/`init` and `idle`/`done`

brophdawg11: For the `useFetcher` use-case we could do this by tracking some local state (`let didAnything = useState(false)`).  

However, in `useFetchers` we don't have a stable key to track fetchers so we cant do the same there.  

In user application code, they can also derive this "done" state via their own `useState` if needed.

packages/remix-react/components.tsx
```diff
@@ -929,16 +1171,177 @@ export function useFetcher<TData = any>(): FetcherWithComponents<
...
+          // TODO: this is created as a random hash value in useSubmitImpl in
+          // Remix today. We do not have this key in react router as we
+          // flattened submissions down onto the fetcher.  We can't recreate in
+          // this back-compat layer in a stable manner for useFetchers because
+          // we don't have a stable fetcher identity.  So we could:
+          //  - Expose the fetcher key from the router (might make sense if
+          //    we're considering adding useFetcher({ key }) anyway
+          //  - Expose a hidden field with a stable identifier on the fetcher
+          //    like we did for _hasFetcherDoneAnything
+          key: "todo-not-implemented-yet",
```
brophdawg11, line 1242: Router doesn't have submission keys - need to think about the best way to tackle this.

Maybe we just add a navigation key to router?

```js
navigation = {
  state,
  location,  // has a key
  form*,
  key, // populate on _all_ navigations, submisison or not
       // does not change on redirects during a naviagtion
}
```

packages/remix-react/components.tsx
```diff
@@ -929,16 +1171,177 @@ export function useFetcher<TData = any>(): FetcherWithComponents<
...
+        },
+        data: undefined,
+      };
+      return fetcher;
+    } else {
+      invariant(false, "nope");
```
brophdawg11, line 1248: Note to self - make this more descriptive

packages/remix-react/components.tsx
```diff
@@ -929,16 +1171,177 @@ export function useFetcher<TData = any>(): FetcherWithComponents<
...
+        // will fix this bug.
+        let url = new URL(formAction, window.location.origin);
+
+        // This typing override should be safe since this is only running for
+        // GET submissions and over in @remix-run/router we have an invariant
+        // if you have any non-string values in your FormData when we attempt
+        // to convert them to URLSearchParams
+        url.search = new URLSearchParams(
+          formData.entries() as unknown as [string, string][]
+        ).toString();
```
brophdawg11, line 1308: This bug is fixed in `useNavigation` but we wanted to keep the bug behavior in `useTransition`

packages/remix-react/data.ts
```diff
@@ -33,17 +30,28 @@ export function isRedirectResponse(response: any): boolean {
...
-  let init: RequestInit = submission
-    ? getActionInit(submission, signal)
-    : { credentials: "same-origin", signal };
+  let init: RequestInit | undefined;
+
+  // TODO: There's a bug in @remix-run/router here at the moment where the
+  // loader Request keeps method POST after a submission.  Matt has a local
+  // fix but this does the trick for now.  Once the fix is merged to the
+  // router, we can remove the isAction param and use the method here
+  // if (request.method !== "GET") {
```
brophdawg11, line 46: This is no longer the case - Matt to update

packages/remix-react/data.ts
```diff
@@ -53,47 +61,7 @@ export async function fetchData(
...
+  // @remix-run/router detection
   return response;
 }
-
-export async function extractData(response: Response): Promise<AppData> {
-  // This same algorithm is used on the server to interpret load
-  // results when we render the HTML page.
-  let contentType = response.headers.get("Content-Type");
-
-  if (contentType && /\bapplication\/json\b/.test(contentType)) {
```
brophdawg11, line 64: For the above TODO - in the router we use the following check - any concerns?

```
if (contentType && contentType.startsWith("application/json")) {
```

brophdawg11: Any reason to do a case insensitive check?

brophdawg11: Seems we do want to regex due to https://httpwg.org/specs/rfc9110.html#field.content-type

packages/remix-react/routeModules.ts
```diff
@@ -23,7 +27,7 @@ export interface RouteModule {
     | V1_HtmlMetaDescriptor
     | V2_MetaFunction
     | V2_HtmlMetaDescriptor[];
-  unstable_shouldReload?: ShouldReloadFunction;
+  shouldRevalidate?: ShouldRevalidateFunction;
```
brophdawg11, line 30: We are not keeping `unstable_shouldReload` - users will have to change to `shouldRevalidate`

packages/remix-react/routes.tsx
```diff
@@ -87,91 +46,74 @@ export function createServerRoutes(
...
+  return function (arg) {
     let module = routeModules[route.id];
     invariant(module, `Expected route module to be loaded for ${route.id}`);
-    if (module.unstable_shouldReload) {
-      return module.unstable_shouldReload(arg);
+    if (module.shouldRevalidate) {
+      return module.shouldRevalidate(arg);
     }
-    return true;
+    return arg.defaultShouldRevalidate;
```
brophdawg11, line 115: `shouldRevalidate` replaces `unstable_shouldReload`

packages/remix-react/routes.tsx
```diff
@@ -183,82 +125,56 @@ async function loadRouteModuleWithBlockingLinks(
...
-
     try {
-      if (!route.hasAction) {
+      if (isAction && !route.hasAction) {
         console.error(
           `Route "${route.id}" does not have an action, but you are trying ` +
             `to submit to it. To fix this, please add an \`action\` function to the route`
         );
+        // TODO: Should we `return null` here like we do for loaders?  Is there
+        // a benefit to triggering a network request that will error/404?
```
brophdawg11, line 145: Confirm behavior here that we want to make the fetch request that we know is going to fail with a 405.  Easier for debugging?

packages/remix-react/routes.tsx
```diff
@@ -183,82 +125,56 @@ async function loadRouteModuleWithBlockingLinks(
...
+      return result;
     } finally {
       await routeModulePromise;
     }
   };
+}
 
-  return action;
+function getRedirect(response: Response): Response {
+  let status = parseInt(response.headers.get("X-Remix-Status")!, 10) || 302;
```
brophdawg11, line 172: This is new behavior to support the bug fix in the router that respects 307/308s and preserves the method/body on the redirect: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307

Remix was previously not sending the status code along at all because it became a 204

packages/remix-server-runtime/routes.ts
```diff
@@ -88,8 +88,6 @@ export function createStaticHandlerDataRoutes(
               })
           : undefined,
         handle: route.module.handle,
-        // TODO: RRR - Implement!
-        shouldRevalidate: () => true,
```
brophdawg11, line 92: Not needed on the server routes

packages/remix-server-runtime/server.ts
```diff
@@ -251,7 +259,8 @@ async function handleDocumentRequestRR(
     );
   } catch (error: unknown) {
     if (serverMode !== ServerMode.Test) {
-      // TODO Do we want to log this here?
+      // TODO Do we want to log this here?  Can we get it to not log on
+      // integration tests runs?
       console.log("Error in entry.server handleDocumentRequest:", error);
```
brophdawg11, line 264: This is probably fine, can remove the todo

packages/remix-eslint-config/rules/core.js
```diff
@@ -89,7 +89,16 @@ module.exports = {
...
+  "no-restricted-imports": [
+    WARN,
+    ...replaceRemixImportsOptions,
+    {
+      importNames: ["useTransition"],
+      message:
+        "useTransition is deprecated in favor of useNavigation as of v1.9.0.",
+      name: "@remix-run/react",
+    },
+  ],
```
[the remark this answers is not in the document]
brophdawg11: Lets bulk include deprecation warnings in a subsequent release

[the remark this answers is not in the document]
brophdawg11: 3 types of deprecation warnings: `eslint`, type definitions (jsdoc + ts), and runtime `console.warn`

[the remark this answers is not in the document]
brophdawg11: Remove for now, we may add back in for the stable release

packages/remix-react/components.tsx
brophdawg11: Leave this in 👍

packages/remix-react/components.tsx
brophdawg11: Lets convert this back to `useMemo` and update our integration test - but Matt/Jacob to put together a smaller reproduction to discuss separately.

packages/remix-react/components.tsx
brophdawg11: Might be able to solve this with `location.state._submissionKey` which would allow us to track this key across submission redirects

packages/remix-react/routes.tsx
brophdawg11: doesn't break your app - it de-optimizes your app

packages/remix-react/routes.tsx
brophdawg11: No need to call the fetch for the 405 here can short circuit like we do for loaders

PR #4704, opened 2022-11-29
fix: specify esbuild target

### LINKED ISSUE #1901: All client side javascript fails to run on non ES-2020 compliant browsers (PRODUCTION only)

1.1.3

### Steps to Reproduce

npx create-remix@latest (with typescript, remix app server)
remix build (Must be a production build. Development doesn't have the problem)
start

Go to localhost:3000 using a non ES-2020 compliant browser (I'm using Safari 12.5)
Check the js console and you will see errors

### Expected Behavior

Client side JavaScript should run without syntax errors on browsers that don't support ES-2020

### Actual Behavior

**ALL** Client side JavaScript fails to run. Safari 12.5 reports Syntax error: Unexpected token '?'

It appears that the some of the client bundles contain the nullish coalescing operator (??) which is an ES-2020 feature.  ES-2020 is too new. There are still many devices out there that can't support it such as older iPhones.

Again, this only happens in the PRODUCTION build.

There appear to be three places where this is happening, only one of which I've tracked down to source code:

1. ScrollRestoration component inside dangerouslySetInnerHTML (this has been reported and fixed already)
2. Unknown source. Appears in multiple entry.client-XXXXX.js bundles. Looks like this:
```
var t=n.checked;return R({},n,{defaultChecked:void 0,defaultValue:void 0,value:void 0,checked:t??e._wrapperState.initialChecked})}function Po(e,n){var
```
3. Unknown source. Appears in multiple other chunks. Looks like this:
```
:A)}),L=Math.min(Math.max(b??v.length-1,0),v.length-1),B=e.Action.Pop,O=v[
```

lili21:
> Remix only runs in browsers that support [ES Modules](https://caniuse.com/es6-module).

Remix should work in browsers that support ES Modules. But Remix doesn't handle some new syntax which break the statement.

fix #1901

### DIFF
--- a/packages/remix-dev/compiler/compileBrowser.ts
+++ b/packages/remix-dev/compiler/compileBrowser.ts
@@ -154,6 +154,7 @@ const createEsbuildConfig = (
     entryPoints,
     outdir: config.assetsBuildDirectory,
     platform: "browser",
+    target: ['safari12'],
     format: "esm",
     external: [
       // This allows Vanilla Extract to bundle asset imports, e.g. `import href


machour:
I don't think this PR will make it as it is, I know the team have been thinking about having a browserlist setting.
/cc @pcattori

packages/remix-dev/compiler/compileBrowser.ts
```diff
@@ -82,6 +82,7 @@ const createEsbuildConfig = (
     entryPoints,
     outdir: config.assetsBuildDirectory,
     platform: "browser",
+    target: ['safari12'],
```
jfsiii, line 85: If the goal is to support safari as documented, this should be `safari11` as that's the first version to support ES modules
Screenshot 2022-12-03 at 8 14 42 AM

I created a ticket for this https://github.com/remix-run/remix/issues/4762

lili21: the reason I pick `safari12` is because this error message

```
✘ [ERROR] Transforming async generator functions to the configured target environment ("safari11" + 1 override) is not supported yet

    node_modules/.pnpm/@web3-storage+multipart-parser@1.0.0/node_modules/@web3-storage/multipart-parser/esm/src/search.js:253:0:
      253 │ async function* arrayIterator(iter) {
          ╵ ~~~~~
```

brophdawg11:
Closing this out in favor of https://github.com/remix-run/remix/discussions/4769

Issue #4483, opened 2022-11-01
🗺 Flat Routes


ryanflorence:
An improved file system route convention to help route visibility and module co-location.

## Progress

- [x] RFC: https://github.com/remix-run/remix/discussions/4482
- [x] PR: #4880

## Community Implementation

https://github.com/kiliman/remix-flat-routes

ryanflorence:
@kiliman has offered to send a PR :)

kiliman:
Yes I’ll be back from Thanksgiving vacation on Monday.

kiliman:
Initial PR https://github.com/remix-run/remix/pull/4716

mcansh:
Next Steps: @pcattori to review updates https://github.com/remix-run/remix/pull/4880

mcansh:
merged!

kiliman:
Awesome!

![image](https://user-images.githubusercontent.com/47168/210888517-c16121d5-5ef8-4d3f-9f5a-aab4e7146664.png)

machour:
Awesome!
We need to document this before shipping it in a release

PR #4716, opened 2022-11-30
feat: add flat routes convention

kiliman:
Closes: #4483

- [ ] Docs
- [ ] Tests

This is the initial PR to add flat routes convention

Testing Strategy:

### DIFF
--- a/.changeset/eight-forks-collect.md
+++ b/.changeset/eight-forks-collect.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Add flat routes convention
--- a/packages/remix-dev/__tests__/readConfig-test.ts
+++ b/packages/remix-dev/__tests__/readConfig-test.ts
@@ -24,6 +24,7 @@ describe("readConfig", () => {
         tsconfigPath: expect.any(String),
         future: {
           v2_meta: expect.any(Boolean),
+          v2_routeConvention: expect.any(Boolean),
         },
       },
       `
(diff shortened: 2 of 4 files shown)


kiliman:
Ok, here's my initial PR. I simply copied over my flatRoutes function. I will work on integrating with existing config and updating the tests.

ryanflorence:
We'll want it behind a future flag, too. Let's go with:

```js
// remix.config.js
{
  future: {
    v2_routeConvention: true
  }
}
```

Future flags were merged here: https://github.com/remix-run/remix/pull/4566

If you run into questions about future flags, reach out to @chaance :)

machour:
@ryanflorence regarding the flag name, have you reached a conclusion regarding keeping or removing the current convention in V2? (Question raised in the RFC)

We may want to change the flag name to `v2_flatRoutesConvention` if we keep both.

kiliman:
@ryanflorence ok so the plan is if the feature flag is true for flat routes, then this will automatically use the flat routes convention instead of the default (i.e., no need to specify the `routes` config option). Correct?

ryanflorence:
Let's move forward as if this is the only built-in convention for v2, so let's use `v2_routeConvention`.

As folks start updating we can revisit if it maybe shouldn't be the only convention, or even the default. I suspect people will prefer it, though.

kiliman:
Ok, I've pushed the latest version. This also includes the *hybrid* routes.

I think this gives you the best of both worlds. You can still organize with nested folders, but you also get colocation.

machour:
@kiliman Tested hybrid routes on 0.5.0, works like a charm 👌🏼 Thank you so much for this.

mcansh:
closed in favor of #4880 (which includes these commits)

PR #4717, opened 2022-11-30
ci: fix race condition writing `globals.d.ts` shim

ascorbic:
There is currently a race condition, where the globals.d.ts shim is created in a Promise.all with the job that copies the same file. This leads to [intermittent errors](https://github.com/remix-run/remix/actions/runs/3583686955/jobs/6029416947#step:7:116) where it tries to copy the file before it is created. This PR ensures that the fiel is created beofre the jobs to copy it are run.

Additionally, it ensures that types are always generated if "publish" is run. 

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


scripts/build.mjs
```diff
@@ -1,8 +1,8 @@
 import { spawn } from "cross-spawn";
 
 const args = process.argv.slice(2);
-const tsc = process.env.CI || args.includes("--tsc");
 const publish = process.env.CI || args.includes("--publish");
+const tsc = process.env.CI || args.includes("--tsc") || publish;
```
ascorbic, line 5: If using the `--publish` flag the `--tsc` flag is required.

MichaelDeBoey (approved):

chaance (approved):

Issue #2997, opened 2022-04-27
Unexpected behaviour when redirecting in action


liegeandlief:
### What version of Remix are you using?

1.4.1

### Steps to Reproduce

Create a Remix project (with the built in Remix server) with the following routes:

index.tsx:
```tsx
import { memo } from 'react'

const IndexRoute = memo(() => {
  return <h1>Home</h1>
})
IndexRoute.displayName = 'IndexRoute'

export default IndexRoute
```

magic-link.tsx:
```tsx
import { memo, useEffect } from 'react'
import { LoaderFunction, json, useLoaderData, useSubmit } from 'remix'
import { emailFormFieldName, codeFormFieldName } from '~/routes/verify-magic-link'

type LoaderData = {
  email: string,
  code: string
}

const loader: LoaderFunction = async ({ request }) => {
  const url = new URL(request.url)
  const email = url.searchParams.get("email") ?? ''
  const code = url.searchParams.get("code") ?? ''

  return json<LoaderData>({
    email,
    code
  })
}

const MagicLinkRoute = memo(() => {
  const loaderData = useLoaderData<LoaderData>()
  const submit = useSubmit()

  useEffect(() => {
    const formData = new FormData()
    formData.append(emailFormFieldName, loaderData.email)
    formData.append(codeFormFieldName, loaderData.code)
    submit(formData, { method: 'post', action: '/verify-magic-link' })
  }, [loaderData.email, loaderData.code, submit])

  return (
    <p>Some pending UI</p>
  )
})
MagicLinkRoute.displayName = 'MagicLinkRoute'

export default MagicLinkRoute
export { loader }
```

verify-magic-link.tsx:
```tsx
import { memo } from 'react'
import { ActionFunction, json, redirect, useActionData } from 'remix'

export const emailFormFieldName = 'email'
export const codeFormFieldName = 'code'

// fake verifyMagicLink function with forced failure
const verifyMagicLink = ({ email, code } : { email: string, code: string  }) => Promise.resolve(false)

type ActionData = Awaited<ReturnType<typeof verifyMagicLink>>

const action: ActionFunction = async ({ request }) => {
  const body = await request.formData()
  const email = body.get(emailFormFieldName)?.toString() ?? ''
  const code = body.get(codeFormFieldName)?.toString() ?? ''

  const verifyMagicLinkResult = await verifyMagicLink({ email, code })

  if (verifyMagicLinkResult === true) {
    return redirect('/')
  }

[...]

PR #859, opened 2021-12-02
feat(remix-react): add WebSocket reconnect to `LiveReload`

garand:
Currently if the LiveReload socket closes, for instance if the dev server restarts, the page loses connection to the socket and needs to be manually refreshed to reconnect.

This wraps the WebSocket connection script in a function that is immediately called, and adds an `onclose` method letting the user know the connection has closed. A timeout is created (1 second) to reconnect and pass an optional config object with an `onOpen` callback to force a reload when a new connection is opened.

This was born out of my use case of building a component library outside of Remix and setting up a separate script to restart the dev server when my component library changed.

---

Original discussion on Discord.

https://discord.com/channels/770287896669978684/911308489769488485/911314308967841832

PR #4725, opened 2022-11-30
fix: Firefox LiveReload

### LINKED ISSUE #4692: LiveReload - Indefinite loop in Firefox

v1.7.6

### Steps to Reproduce

1. `create-remix` using defaults
2. Add a loader to `app/routes/index.tsx` that sleeps for 1 second
3. Open http://localhost:3000 in Firefox Developer Edition (I'm using `108.0b6 (64-bit)` macOS silicon)
4. Make a change to the file to cause remix rebuild while the firefox window is in view (focus state doesn't seem to matter)
5. Page continually reloads until window is no longer in view (focus state doesn't seem to matter)

```typescript
        await new Promise((resolve) => setTimeout(resolve,1000));
        const data = { key: "value" };
        return json(data);
}
```

### Expected Behavior

The page should only live reload once

### Actual Behavior

The page reloads in a looping manner. Normal behavior can be observed in chrome or by removing the sleeping timeout.

ryanflorence:
Firefox infinitely reloads the page as long as `<LiveReload>` is rendering.

Closes #4692

### DIFF
--- a/.changeset/thin-oranges-boil.md
+++ b/.changeset/thin-oranges-boil.md
@@ -1,0 +1,15 @@
+---
+"@remix-run/react": patch
+---
+
+Fixed a problem with live reload and firefox infinitely reloading the page
+
+The problem is:
+
+1. Firefox is calling `ws.onclose` immediately upon connecting (?!)
+2. Then we’re trying to reconnect, and upon reconnection, we reload the page.
+3. Firefox then calls `ws.onclose` again after reconnecting and the loop starts over
+
+This fix is to check `event.code === 1006` before actually trying to reconnect and the reload the page. 1006 means the connection was closed abnormally (https://www.rfc-editor.org/rfc/rfc6455#section-7.4.1). In our case, that means the server was shut down in local dev and then the socket can reconnect again when the server is back up.
+
+It’s unclear to me why Firefox is calling `onclose` immediately upon connecting to the web socket, but it does.
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1646,7 +1646,6 @@ export const LiveReload =
                   let socketPath = protocol + "//" + host + ":" + ${String(
                     port
                   )} + "/socket";
-
                   let ws = new WebSocket(socketPath);
                   ws.onmessage = (message) => {
                     let event = JSON.parse(message.data);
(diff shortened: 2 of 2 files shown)


packages/remix-react/components.tsx
```diff
@@ -1663,15 +1662,17 @@ export const LiveReload =
                       config.onOpen();
                     }
                   };
-                  ws.onclose = (error) => {
+                  ws.onclose = (event) => {
                     console.log("Remix dev asset server web socket closed. Reconnecting...");
```
chaance, line 1666: Should this log move into the check as well? We aren't actually reconnecting if not, right?

ryanflorence: oops, yep

.changeset/thin-oranges-boil.md
```diff
@@ -0,0 +1,15 @@
...
+
+The problem is:
+
+1. Firefox is calling `ws.onclose` immediately upon connecting (?!)
+2. Then we’re trying to reconnect, and upon reconnection, we reload the page.
+3. Firefox then calls `ws.onclose` again after reconnecting and the loop starts over
+
+This fix is to check `event.code === 1006` before actually trying to reconnect and the reload the page. 1006 means the connection was closed abnormally (https://www.rfc-editor.org/rfc/rfc6455#section-7.4.1). In our case, that means the server was shut down in local dev and then the socket can reconnect again when the server is back up.
+
+It’s unclear to me why Firefox is calling `onclose` immediately upon connecting to the web socket, but it does.
```
danieljb, line 15: Hi. When I first had this issue I did some digging and might have an idea of what’s happening:

1. files change
2. [Chrome, Firefox] websocket receives message `type: RELOAD`, triggers `window.location.reload`
3. [Chrome, Firefox] start reload _Navigated to http://localhost:3000/_
4. [Firefox] closes the websocket connection `code: 1001`
5. [Firefox] sets timeout to call `remixLiveReloadConnect`
6. [Firefox] timeout completes _before_ reload is complete and calls:
  `remixLiveReloadConnect({ onOpen() { /* … */ } })`
7. [Firefox] re-connects to the websocket, calls `onOpen` method that triggers another `window.location.reload` starting the infinite reloading

My conclusion is Firefox does **not** in fact call `onclose` upon connecting. It is the other way around, the connection is closed and tries to reconnect before the browser fully unloads the page. Both Chrome and Safari seem to only send the websocket close event when the dev server is stopped and not when the page reloads.

The reason why the `onOpen` config was introduced was to enable automatic [re-connection when the dev server is stopped and restarted](https://github.com/remix-run/remix/pull/859). When that happens the CloseEvent is sent with code `1006` which is why [both of us](https://github.com/remix-run/remix/issues/2997#issuecomment-1277308621) came up with the `event.code === 1006` check.

This fixes the infinite reloads. But the issue on Firefox is actually a race condition. When the timeout is increased (e.g. to `10000`) the page reload reliably finishes before the timeout re-initializes, re-connects to the websocket and the reloads loop kicks in.

Issue #4199, opened 2022-09-15
TypedResponse allows incompatible types


haines:
### What version of Remix are you using?

1.7.0

### Steps to Reproduce

Use a `TypedResponse<T>` in place of a `TypedResponse<U>`, where `T` and `U` are incompatible.

```typescript
const wat: TypedResponse<string> = json(42);

// or return it from a function
function wat(): TypedResponse<number> {
  return json("oh no");
}
```

### Expected Behavior

Compile errors due to the type mismatch.

### Actual Behavior

No compile errors.

suhasabhishek:
@machour Can I work on this?

machour:
@suhasabhishek of course, help is always welcome!
Make sure to target the `dev` branch of this repo.

dabdine:
Messing around with this, I found that the reason for this behavior is:

- Typescript's `Body` interface (which `Response` extends) defines the json function as `json(): Promise<any>`. 
- The override in `TypedResponse` does not properly handle `Body`'s `any` type. I'm not an expert in Typescript generics, but I'm guessing it's because `any` is more weakly typed than anything else causing the stronger typed generic parameter T to be lost due to erasure.

The fix in my PR addresses this by `Omit`ting the `json` function that is defined on Typescript's `Body` interface and redefining it with the more strongly typed version from `TypedResponse`.

PR #4734, opened 2022-12-01
Fix 4199: TypedResponse allows incompatible types

dabdine:
Closes: #4199: TypedResponse allows incompatible types

NOTE: I believe there's a _small_ chance this can be a breaking change for some people who may have been counting on the previous TypedResponse behavior. To be more explicit, something like this would fail now, as the actual return type is different than the declared return type (which was not enforced prior to this patch):

```typescript
export const action = async (args: ActionArgs): Promise<TypedResponse<string>> => {
    return json(42);
};
```

This should have been the desired behavior though, and would be a benefit as users would either have to update their method declarations or fix their code as we're uncovering errors that were previously hidden.

[Playground link demonstrating the issue](https://www.typescriptlang.org/play?&q=483#code/LAKALgngDgpgBAFWjAJgJRgZygewHaYwA8CcMAHmDHipnAK54DWeOA7nnALwPOscA+bnAzZ8hOADI4Ab1Bw4AK0z4AFAEoAXHAAKAJxwBbAJaESAgDSgAvgG5QoSLDgA5GGySx0WXAWKkKKho6RhZ2Th5Q-jwhHgB5EzAiUV9CCzgAcmV8DKFpORAFbLwNbX0jU39LG3sQUAB6ergASRcAYTi0NABRNoRtMAALUzhMQZx6ABsUOFYwOAAjeFxMTGMFyZhQTfnjPABjHD09GH2wbU9UFPFiPHpDJb1Y2Ws4AEM6S+8xPyJMMD0ewA5gJag0mh0ur1+nAAJoTOBAmDzN5wJxYfaAqDzGDHI5wPZo4Z0fYfLYgHZwQ7HU7nVzuL7XX53B6454yV4fekeZDfVLEf6AvAg2pAA)

- [ ] Docs
- [X] Tests

Testing Strategy:

Added a test case to `responses-test.ts`. Note, it's a compile-time test, so it's not something `yarn test` would normally catch. However, I also ran the following from my command line:

```bash
yarn test
```

### DIFF
--- a/.changeset/light-sheep-give.md
+++ b/.changeset/light-sheep-give.md
@@ -1,0 +1,26 @@
+---
+"remix": patch
+"@remix-run/serve": patch
+"@remix-run/server-runtime": patch
+---
+
+Fix `TypedResponse` so that Typescript correctly shows errors for incompatible types in loaders and actions.
+
+Previously, when the return type of a loader or action was explicitly set to `TypedResponse<SomeType>`,
+Typescript would not show errors when the loader or action returned an incompatible type.
+
+For example:
+
+```ts
+export const action = async (args: ActionArgs): Promise<TypedResponse<string>> => {
+    return json(42);
+};
+```
+
+In this case, Typescript would now show an error even though `42` is clearly not a `string`.
+
+In this case, happens because `json` returns a `TypedResponse<string>`,
+but because `TypedReponse<string>` is just `Response & { json: () => Promise<string> }`
+and `Response` already defines `{ json: () => Promise<any> }`, then type erasure caused `Promise<any>` to be used for `42`.
+
+To fix this, we explicitly omit the `Response`'s `json` property before intersecting with `{ json: () => Promise<T> }`.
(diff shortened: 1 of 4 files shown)


pcattori (approved):

pcattori:
@dabdine thanks for the fix! 🙏

Looks like there are some unrelated errors for our Ubuntu integration tests. We're looking into those now.

.changeset/light-sheep-give.md
```diff
@@ -0,0 +1,26 @@
...
+
+For example:
+
+```ts
+export const action = async (args: ActionArgs): Promise<TypedResponse<string>> => {
+    return json(42);
+};
+```
+
+In this case, Typescript would now show an error even though `42` is clearly not a `string`.
```
dabdine, line 20: I think we want "In this case, Typescript would *not* show an error..." (now -> not)

dabdine: I can make the change real quick

dabdine:
@pcattori Thanks for the review and for putting the changeset together! Still learning how to contribute correctly :)

pcattori:
> @pcattori Thanks for the review and for putting the changeset together! Still learning how to contribute correctly :)

No worries! You did it right 😁 if PR authors write a changeset, that can be nice, but I also think its good for us to write one to capture a maintainer's understanding of the feature/fix. Both to double check our own understanding, but also to put it in context of Remix in general for release notes.

PR #4736, opened 2022-12-01
chore(scripts): Use relaxed peer dependencies to avoid triggering major version bumps

chaance:
Consider this scenario:
1. Run `yarn changeset:version` to bump versions
2. Changesets sees that a package has a minor change
3. Because we release packages in lockstep, all packages get a minor update
4. `@remix-run/dev` has `"peerDependencies": { "@remix-run/serve": "1.8.0" }`
5. This dependency will now be out of range after the update
6. Changesets makes the safe bet and updates `@remix-run/dev` to 2.0.0 because it can't be sure this doesn't result in a breaking change
7. Because we release packages in lockstep, all packages get a major update

In practice, this means any `minor` changeset will result in a major bump, which definitely isn't what we want. This means I end up manually updating all of the package versions which ... kinda sucks.

Instead, we relaxe the peer dependency range for internal packages. That way the update doesn't result in an out-of-range peer dependency, and all packages are bumped to the next minor release instead.

Because changesets doesn't automatically bump peer dependencies with the relaxed range (which makes sense in most cases), I added a script does that for us. This makes the change safer because updating the leading dependency will result in a peer dependency warning if the user doesn't bump the peer dependency for some reason.

Thanks to @Andarist for the guidance here and the original script!

### DIFF
--- a/.changeset/big-spoons-grab.md
+++ b/.changeset/big-spoons-grab.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": minor
+---
+
+allow defining multiple routes for the same route module file
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -19,6 +19,7 @@
       "@remix-run/react",
       "@remix-run/serve",
       "@remix-run/server-runtime",
+      "@remix-run/testing",
       "@remix-run/vercel"
     ]
   ],
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -26,5 +26,8 @@
   "access": "public",
   "baseBranch": "dev",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": [],
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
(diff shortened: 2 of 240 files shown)


Andarist:
> Because changesets doesn't automatically bump peer dependencies with the relaxed range (which makes sense in most cases)

I wouldn't say that this makes sense in most cases though. It's a contextual decision. For example, it's not a good default for Babel plugin (usually they depend on `^7`) or UI kit libraries with a peer dep on a framework (when depending on React you might want to depend on a version with a certain feature, like if you depend on hooks then u stay on the `^16.8` range).

I think though that this strategy makes sense in your case.

chaance:
> I wouldn't say that this makes sense in most cases though. It's a contextual decision. For example, it's not a good default for Babel plugin (usually they depend on `^7`) or UI kit libraries with a peer dep on a framework (when depending on React you might want to depend on a version with a certain feature, like if you depend on hooks then u stay on the `^16.8` range).

Updated the inline comment as not to sully your good name 😀

PR #4744, opened 2022-12-02
fix(remix-react): fix unsafe access to sessionStorage

Jarrku:
Solves: Remix apps crashing in incognito mode when embedded through an iframe in another webpage.

Safeguards accessing `sessionStorage` in cases where it's not available (iframe'd application on another website in incognito mode).

Closes: #

- [ ] Docs
- [ ] Tests

CSB: https://codesandbox.io/s/zealous-knuth-btnb9r?file=/index.html

Testing Strategy:

Visit following csb url in both regular chrome & incognito chrome:  https://btnb9r.csb.app/

Regular chrome: It loads
Incognito chrome: Crashes due to accessing sessionStorage.

Context:
We're developing a plugin that third parties can embed in their own webpage through an iframe. However, remix currently crashes on those webpages in incognito mode because it accesses sessionStorages before checking if it's available.

Initially, we worked around this issue by removing the scroll restoration component, but it seems some of it's code is ran regardless. 

Thanks for your time and feedback in advance!

### DIFF
--- a/.changeset/wicked-bananas-taste.md
+++ b/.changeset/wicked-bananas-taste.md
@@ -1,0 +1,6 @@
+---
+"remix": patch
+"@remix-run/react": patch
+---
+
+fix(remix-react): fix unsafe access to sessionStorage
--- a/contributors.yml
+++ b/contributors.yml
@@ -178,6 +178,7 @@
 - jamiebuilds
 - janhoogeveen
 - Jannis-Morgenstern
+- Jarrku
 - jasonadelia
 - jaydiablo
 - jca41
--- a/packages/remix-react/scroll-restoration.tsx
+++ b/packages/remix-react/scroll-restoration.tsx
@@ -8,8 +8,21 @@ let STORAGE_KEY = "positions";
 
 let positions: { [key: string]: number } = {};
 
+let isSessionStorageAvailable = true;
+try {
+  void sessionStorage;
+} catch {
+  isSessionStorageAvailable = false;
+}
+
+const safeSessionStorage = {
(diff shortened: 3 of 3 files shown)


machour (changes requested):
Hi @Jarrku and thank you so much for opening this PR!

Can you target the `dev` branch instead?

Jarrku:
I'll rebase on top of dev

machour:
LGTM, although we may want to disable the component all together if sessionStorage is not available, or if we want to look for another Storage that would be always available (cookies?). Let's see what other think.

In the mean time, could you add a changeset (see bot comment above)? Should be a patch bump for `remix-react`.

Jarrku:
Wasn't sure either, noop method seemed least invasive to me. Added the changeset, it prompted me for `remix` package as well. Can remove this entry if needed?

machour:
@Jarrku since 1.10.0, `useScrollRestoration` now lives in [react-router-dom](https://github.com/remix-run/react-router/blob/0e4c524b329bf9641ccfc85da95bcde708bc56ca/packages/react-router-dom/index.tsx#L1094), so I'm going to close this now unmergeable PR.

Would you mind reopening it in the react-router repository? 🙏🏼 
/cc @brophdawg11

brophdawg11:
@Jarrku Please also re-test if this is an issue still.  Removing the `<ScrollRestoration>` component should work just fine now since all `sessionStorage` access is within there in the new implementation of this in `react-router-dom`

Jarrku:
CSB example seem to work now, besides logging expected errs ,so I assume all will be good, thanks!

Issue #4733, opened 2022-12-01
createRequestHandler in server-runtime causing slower responses in larger apps


dmarkow:
### What version of Remix are you using?

1.8.0

### Steps to Reproduce

My production app has a large number of routes (around 700). I'm dealing with some slower requests, so I ran some benchmarking to see how long remix takes to process each request, excluding time for my loaders/actions. I narrowed down the bulk of the extra processing time to the use of `unstable_createStaticHandler` in `@remix-run/server-runtime` (which delegates to `@remix-run/router`). Even though this function returns the same thing given the same `loadContext`, it's being called on every request.

### Expected Behavior

In production, the results of `unstable_createStaticHandler` should either be calculated once at build time, or cached and only updated when the `loadContext` changes.

### Actual Behavior

`unstable_createStaticHandler` is called for every request. In my 700-route app, which is adding up to 200ms to my requests.

```typescript
// packages/remix-server-runtime/server.ts

export const createRequestHandler: CreateRequestHandlerFunction = (
  build,
  mode
) => {
  let routes = createRoutes(build.routes);
  let serverMode = isServerMode(mode) ? mode : ServerMode.Production;

  return async function requestHandler(request, loadContext = {}) {
    let url = new URL(request.url);
    let matches = matchServerRoutes(routes, url.pathname);

    // This is being called for every request. In a large app, this single line can easily add
    // 100-200ms of overhead.
    let staticHandler = unstable_createStaticHandler(
      createStaticHandlerDataRoutes(build.routes, loadContext)
    );
```

I just set up a local patch which memoizes `unstable_createStaticHandler`. This immediately dropped my requests from  around 250ms each to around 80ms each (latency to my server is 60ms so this is around 20ms for remix to handle the request instead of almost 200ms).

brophdawg11:
Thanks @dmarkow!  Are you using the `loadContext` in your app?  Do you want to share your patch and we will look into improving the perf there 👍

dmarkow:
I added a cache that only gets updated in production. Digging into the internals, I'm always a little worried I may be breaking something I'm unaware of. I don't personally use `loadContext` but this approach should allow for creating a new handler when the context changes.

```diff
diff --git a/packages/remix-server-runtime/server.ts b/packages/remix-server-runtime/server.ts
index feea61bd..6b77a8ba 100644
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
[...]

PR #4748, opened 2022-12-02
perf(remix-server-runtime): Performance improvements for large apps

dmarkow:
Closes: #4733

- [ ] Docs
- [x] Tests

Testing Strategy: All existing tests are still passing. Since this is only performance-related, not sure of any new tests to add. Also running these patches in production with no issues so far.

In apps with many routes, there were some performance issues causing some remix overhead for every request. In my production app (700+ routes), 200ms on average was being added to *every* request, before my loaders/actions even ran. I narrowed this down to route parsing.

- The `requestHandler` in `createRequestHandler` was calling `createStaticHandlerDataRoutes` on every request, which is an expensive recursive function that creates a nested set of data route objects. In production, `build.routes` shouldn't be changing between requests, so the result is now cached and only recalculated if `loadContext` changes. In development, nothing will be cached since `createRequestHandler` is called on every request.
- Both `createStaticHandlerDataRoutes` and `createRoutes` were recursive functions which repeatedly filtered the entire list of routes. I extracted a `groupRoutesByParentId` function to create an object indexed by `parentId`. The two functions now call this once, and pass the results down recursively avoiding the repeat filtering. In my app, this dropped each of these functions from 50-100ms down to around 1ms each.
- After these fixes, the remix overhead before running loaders/actions dropped from 200ms to around 1ms.
- With those two functions *much* quicker now, I tried removing the caching in `requestHandler`. That increased the overhead back up to 10-20ms when the server is under load. So I'm inclined to leave the caching in place as a cheap way to shave those few ms off the response time.


dmarkow:
Just updated to remove the caching logic in `server.ts`, as @brophdawg11 mentioned an API change will be coming that will not require this in the future.

packages/remix-server-runtime/routes.ts
```diff
@@ -40,27 +40,50 @@ export interface ServerRoute extends Route {
...
-  parentId?: string
+  parentId?: string,
+  routesByParentId?: Record<string, Omit<ServerRoute, "children">[]>
 ): ServerRoute[] {
-  return Object.entries(manifest)
-    .filter(([, route]) => route.parentId === parentId)
-    .map(([id, route]) => ({
+  // Create a map of routes by parentId to use recursively instead of
+  // repeatedly filtering the manifest.
+  routesByParentId ||= groupRoutesByParentId(manifest);
```
cliffordfajardo, line 64: @machour - would the [logical or assignment ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment)(|=) be transpiled correctly for older browsers in Remix's build process?

I recall recently that null coalescing operator was removed:
https://github.com/remix-run/remix/pull/4561

machour: As far as I know, this code will only run on the server, so it shouldn't cause any problem with the browser

brophdawg11: We ran into an issue on node with this on node 14 that should be fixed since we changed the target in our `babelrc`, but that might be worth confirming.

brophdawg11: I ended up removing the `||=` in [df6f364](https://github.com/remix-run/remix/pull/4748/commits/df6f364d92b80eef7747d188a31b99be895b3323) and switched it to use a default param value

brophdawg11 (approved):

mcansh (approved):

PR #4751, opened 2022-12-02
ci: only apply `awaiting release` label to prs related to `./packages`

mcansh:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.github/workflows/merged-pr.yml
+++ b/.github/workflows/merged-pr.yml
@@ -5,6 +5,8 @@ on:
     types: [closed]
     branches:
       - dev
+    paths:
+      - "packages/**"
 
 permissions:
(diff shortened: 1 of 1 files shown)


MichaelDeBoey:
I think we can use the built-in `paths` for this 🤔
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

mcansh:
> I think we can use the built-in `paths` for this 🤔

always keeping me in check haha ❤️

machour (approved):

MichaelDeBoey (approved):

PR #4757, opened 2022-12-02
fix(remix-testing): fix deps

brophdawg11:
The feature branch for the `remix-testing` got a bit stale once we released 1.8.0 and deleted the dup router code

### DIFF
--- a/packages/remix-server-runtime/index.ts
+++ b/packages/remix-server-runtime/index.ts
@@ -74,21 +74,3 @@ export type {
   UploadHandler,
   UploadHandlerPart,
 } from "./reexport";
-
-export type {
-  AgnosticDataRouteObject,
-  AgnosticIndexRouteObject,
-  AgnosticNonIndexRouteObject,
-  AgnosticRouteObject,
-  InitialEntry,
-  Location,
-  MemoryHistory,
-  StaticHandler,
-} from "./router";
-export {
-  createMemoryHistory,
-  matchRoutes,
-  unstable_createStaticHandler,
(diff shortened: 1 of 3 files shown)


packages/remix-server-runtime/index.ts
```diff
@@ -74,21 +74,3 @@ export type {
...
-  MemoryHistory,
-  StaticHandler,
-} from "./router";
-export {
-  createMemoryHistory,
-  matchRoutes,
-  unstable_createStaticHandler,
-} from "./router";
-export type { Update } from "./router/history";
-export type { AgnosticRouteMatch } from "./router/utils";
```
brophdawg11, line 94: No longer have this dup code, can import these right from `@remix-run/router` over in `remix-testing` instead of exporting through here

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -8,23 +8,30 @@ import type {
...
-  json,
   matchRoutes,
   unstable_createStaticHandler as createStaticHandler,
-} from "@remix-run/server-runtime";
+} from "@remix-run/router";
+
+type Update = {
+  action: Action;
+  location: Location;
+};
```
brophdawg11, line 34: This isn't exported through the router since history is an implementation detail now, but it could be if needed.  Just defined it here for now though

mcansh (approved):

machour (approved):

MichaelDeBoey (approved):

PR #4758, opened 2022-12-02
chore(remix-testing): use `@remix-run/router` directly

MichaelDeBoey:
Follow-up of #4539 

Fixes CI failures


MichaelDeBoey:
@brophdawg11 was a bit quicker 🙈

Closing this as duplicate of #4757

brophdawg11:
haha you and logan are both so fast!  Jacob and I came across these during some local RR work as well so I had them locally but just got around to putting them into a PR 👍

PR #4761, opened 2022-12-03
perf(remix-architect,remix-netlify): improve performance of `isBinaryType`

gyx1000:
close #4760

I noticed that the isBinaryType function of the remix-architect adapter has a complexity of O(N*includesComplexity)
This change could improve complexity to O(1)  but could generate braking change for malformed "Content-Type" headers.


gyx1000:
Hi @MichaelDeBoey,

Thank you for your comment.

I don't know what benchmark tools you usually use but I did the test: https://jsben.ch/lCwux
The difference is between 30% and 40%.

Indeed, this change will not have a huge impact on a query but I think it is normal to avoid an iteration for each answer. Especially since remix is mainly intended to return text/html.

This adapter is also intended to be used on the AWS infrastructure which charges by runtime.

mcansh (approved):

MichaelDeBoey (approved):

mcansh:
a similar change should probably be brought upstream to arc itself https://github.com/architect/functions/blob/main/src/http/helpers/binary-types.js

mcansh:
oh oh oh! one more thing @gyx1000 can you make the same change to the netlify file as well? https://github.com/remix-run/remix/blob/dev/packages/remix-netlify/binaryTypes.ts

mcansh (approved):

gyx1000:
@mcansh Yeah perfect.
Sorry I didn't watch the other adapters

MichaelDeBoey (approved):

gyx1000:
As revealed by @ryanblock in https://github.com/architect/functions/pull/539, using `Array.includes` is faster than `Set.has`. Indeed, binaryTypes does not have many entries.

MichaelDeBoey (approved):

Issue #4766, opened 2022-12-04
Property 'future' is missing in type 'typeof import("./node_modules/@remix-run/dev/server-build")'


simonsinclair:
### What version of Remix are you using?

1.8.0

### Steps to Reproduce

remix.config.js
```js
/** @type {import('@remix-run/dev').AppConfig} */
module.exports = {
  devServerBroadcastDelay: 1000,
  ignoredRouteFiles: ['**/.*'],
  server: './server/index.ts',
};
```

server/index.ts
```ts
import express from 'express';
import { createRequestHandler } from '@remix-run/express';
import * as build from '@remix-run/dev/server-build';

const app = express();

app.use('/build', express.static('public/build', { immutable: true, maxAge: '1y' }));

app.use(express.static('public', { maxAge: '1h' }));

app.all(
  '*',
  createRequestHandler({
    build,
    mode: process.env.NODE_ENV,
  })
);

app.listen(3000);
```

### Expected Behavior

The changes to `@remix-run/server-runtime` from `1.7.6` (the version I was using) to `1.8.0` are listed as patch and minor changes, so I expected my existing code to be compatible.

### Actual Behavior

<img width="894" alt="error" src="https://user-images.githubusercontent.com/1430787/205490424-95e6264b-c5c9-4614-813a-442f5864c45a.png">

## Note
I tried adding `future: {}` to `remix.config.js` as per https://github.com/remix-run/remix/issues/4759, but this does not resolve the issue.

machour:
@simonsinclair Are all your remix deps up to date? A gist with both your package.json & package-lock.json would help 🙏🏼 
Can you also provide the @remix-run/dev/server-build.js file?

simonsinclair:
> Are all your remix deps up to date? A gist with both your package.json & package-lock.json would help 🙏🏼 Can you also provide the @remix-run/dev/server-build.js file?

Yes, they should be. Here are the Gists:
- `package.json` https://gist.github.com/simonsinclair/1c74ac2f29b4666d99f631e15bfdf0b6 
- `package-lock.json` https://gist.github.com/simonsinclair/c39f14e66173fbdda548f825628811e6 
- `@remix-run/dev/server-build.js` https://gist.github.com/simonsinclair/20dac22b3f80c97c678022a716e12233

MichaelDeBoey:
It seems like the `grunge-stack` has this problem as well
[...]

Issue #4759, opened 2022-12-03
Error in vercel deploy: TypeError: Cannot read properties of undefined (reading 'v2_meta')


Lucas-Eduardo-Goncalves:
### What version of Remix are you using?

1.6.7

### Steps to Reproduce

clone repository by ssh: `git clone git@github.com:Lucas-Eduardo-Goncalves/pac.app.git`
then add this repository to vercel, and try.

### Expected Behavior

My site should work normally. On localhost I can't find it works normally.

### Actual Behavior

I find the error message: TypeError: Cannot read properties of undefined (reading 'v2_meta')

meglio:
Same error for me in `1.8.0`. Wasn't a case with `1.7.2`

machour:
Can you both provide a gist of you package.lock file?

machour:
Can you add `future: {}` in your `remix.config.js` and see if the issue goes away?

antoninbeaufort:
Had the same problem but adding `future: {}` solved the problem

JClackett:
@machour Adding future doesn't fix, also broken in local for me

ThanhPhat1080:
Seem you are missing edit the remix config while you want to deploy to Vercel.
In `remix.config.js`
```js
/** @type {import('@remix-run/dev').AppConfig} */
module.exports = {
  serverBuildTarget: "vercel",
  // When running locally in development mode, we use the built in remix
  // server. This does not understand the vercel lambda module format,
  // so we default back to the standard build output.
  server: process.env.NODE_ENV === "development" ? undefined : "./server.js",
  ignoredRouteFiles: ["**/.*"],
  // appDirectory: "app",
  // assetsBuildDirectory: "public/build",
  // serverBuildPath: "api/index.js",
  // publicPath: "/build/",
};
```

Create new `server.js` file at your root folder (the same level of `remix.config.js`)

```js
import { createRequestHandler } from "@remix-run/vercel";
import * as build from "@remix-run/dev/server-build";

export default createRequestHandler({ build, mode: process.env.NODE_ENV });

```

sjwinther:
Having the same problem, but adding `future: {}` does not solve the issue. I already have a `server.js` file as described by @ThanhPhat1080.

machour:
@sjwinther are all your remix deps on 1.8.0? Can you share a gist of your package.json depedencies & devDependencies sections?

sjwinther:
[...]

PR #4771, opened 2022-12-05
fix(types): Add missing type def for `future` value in `@remix-run/dev/server-build`

chaance:
This should fix https://github.com/remix-run/remix/issues/4766. Unrelated to https://github.com/remix-run/remix/issues/4759.

Changeset will be added to the release branch @brophdawg11 started. Aiming to cut a hotfix shortly.

### DIFF
--- a/.changeset/blue-tigers-look.md
+++ b/.changeset/blue-tigers-look.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Added a missing type definition for the Remix config `future` option to the `@remix-run/dev/server-build` virtual module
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (approved):

MichaelDeBoey:
I think this should go to `dev` branch instead of `main`? 🤔

chaance:
> I think this should go to `dev` branch instead of `main`? 🤔

Hot-fixes actually go to `main` to avoid releasing other work on `dev` early. We need to prioritize this fix along with some router stuff @brophdawg11 is finishing up.

mcansh (approved):

PR #4778, opened 2022-12-06
fix(server-runtime/session): unexpected behaviour in update of session data

cairin:
### Description

> **TLDR**: `commitSession(session)` will always perform an update when there is a valid cookie. Even if the underlying data no longer exists.

The following usage of `createSessionStorage` is currently producing unexpected results:

```js
export const { getSession, commitSession, destroySession } =
  createSessionStorage({
    cookie,
    async createData(data, expires) {
      const id = v4()
      await redisClient.set(id, JSON.stringify(data), {
        PXAT: expires?.valueOf(),
        NX: true // Only perform set if data doesn't exist
      })
      return id
    },
    async readData(id) {
      const data = await redisClient.get(id)
      if (data == null) return null
      return JSON.parse(data)
    },
    async updateData(id, data, expires) {
      await redisClient.set(id, JSON.stringify(data), {
        PXAT: expires?.valueOf(),
        XX: true // Only perform set if data exists
      })
    },
    async deleteData(id) {
      await redisClient.del(id)
    }
  })

```

It would seem reasonable that we can enforce data create/update in the respective CRUD controllers. However, the current implementation of [commitSession](https://github.com/remix-run/remix/blob/dev/packages/remix-server-runtime/sessions.ts#L242-L252) doesn't check if the session data exists when deciding on whether it should create/update the session. If the database fails and the session data is destroyed (before the cookie expires), the above implementation will silently fail. `updateData` won't update the session because it no longer exists, but `commitSession` will not create the data again either. No errors will be thrown.

One could argue that `updateData` should be able to set data, but then why have a full CRUD interface? Also, a decision like that should be documented.

### Testing

The testing strategy and fix is documented in the relevant commit descriptions.

Tests ran using `yarn test:primary --selectProjects server-runtime` during development and `yarn test:primary` to check it didn't have any unexpected implications.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -61,6 +61,7 @@
 - bsides
 - bustamantedev
 - c43721
+- cairin
 - camiaei
 - CanRau
 - ccssmnn
--- a/packages/remix-server-runtime/__tests__/sessions-test.ts
+++ b/packages/remix-server-runtime/__tests__/sessions-test.ts
@@ -1,5 +1,6 @@
 import { createCookieFactory } from "../cookies";
 import type { SignFunction, UnsignFunction } from "../crypto";
+import type { SessionData } from "../sessions";
 import {
   createSession,
   createSessionStorageFactory,
@@ -28,8 +29,16 @@ const createCookie = createCookieFactory({ sign, unsign });
 const createCookieSessionStorage =
   createCookieSessionStorageFactory(createCookie);
 const createSessionStorage = createSessionStorageFactory(createCookie);
-const createMemorySessionStorage =
-  createMemorySessionStorageFactory(createSessionStorage);
+
+const externalMemoryDataStore = new Map<
+  string,
+  { data: SessionData; expires?: Date }
+>();
+
+const createMemorySessionStorage = createMemorySessionStorageFactory(
+  createSessionStorage,
+  externalMemoryDataStore
+);
 
 describe("Session", () => {
   it("has an empty id by default", () => {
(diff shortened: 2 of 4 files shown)


brophdawg11:
I don't know that we should blindly call `readData` in `commitSession` because then every call to `commitSession` results in 2 calls to your backing store?

> If the database fails and the session data is destroyed (before the cookie expires)

Is this the only use case for this?  Somehow the backing session store gets deleted without you ever calling `destroySession`?


> No errors will be thrown.

This is because of NX/XX flags, no?  If you removed them would it "fix" the issue?  What's the use-case for needing them in the first place?

As you said, this can be handled in the current methods for advanced use-cases (by doing your own checks), so I'm leaning towards that being the path forward here.

cairin:
> I don't know that we should blindly call readData in commitSession because then every call to commitSession results in 2 calls to your backing store?

Yes you probably don't want the duplicate call. We actually just settled on always setting the data, regardless of whether create/update is called. 

> > If the database fails and the session data is destroyed (before the cookie expires)
>
> Is this the only use case for this? Somehow the backing session store gets deleted without you ever calling destroySession?

I can't remember the exact specifics of what was happening, but there was a race condition with the cookie/data expirations, that was causing them to be out of sync. And update/create was being called when it shouldn't. The redis client also didn't throw that error based on the flag, which meant if we tried to do an update on data that was no longer there it would fail silently.

> As you said, this can be handled in the current methods for advanced use-cases (by doing your own checks), so I'm leaning towards that being the path forward here.

Yes that seems reasonable. It might also be a good idea to add to the [createSessionStorage documentation](https://remix.run/docs/en/1.19.3/utils/sessions#createsessionstorage) some more context on when `createData` and `updateData` are called. Basically remix will call `createData` if there isn't a valid session cookie and `updateData` otherwise, and that you need to be aware that remix doesn't check if the underlying data exists.

brophdawg11:
👍 I added some details to the docs in https://github.com/remix-run/remix/commit/fd31ce92d7bbf891a11a00f076211c9366b99405

Issue #4741, opened 2022-12-02
Error: Expected a Response to be returned from queryRoute


alfonsocartes:
### What version of Remix are you using?

1.8.0

### Steps to Reproduce

I'm downloading an image from S3 with a resource route and I get Error: Expected a Response to be returned from queryRoute.

This doesn't happen in 1.7.6.

### Expected Behavior

Get image

### Actual Behavior

```
GET /assets/images/[AWSIMAGEKEY]?w=768&format=webp 500 - - 563.637 ms
The following error is a bug in Remix; please open an issue! https://github.com/remix-run/remix/issues/new
Error: Expected a Response to be returned from queryRoute
    at Object.invariant [as default] (/Users/.../node_modules/@remix-run/server-runtime/dist/invariant.js:18:11)
    at handleResourceRequestRR (/Users/.../node_modules/@remix-run/server-runtime/dist/server.js:267:25)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at requestHandler (/Users/.../node_modules/@remix-run/server-runtime/dist/server.js:46:18)
    at /Users/.../node_modules/@remix-run/express/dist/server.js:39:22
```

brophdawg11:
@alfonsocartes Are you able to provide a minimal reproduction or code sample of what your loader looks like and how you're getting the image from S3?  That would help us dig into the root cause here.  Thanks!

alfonsocartes:
Sure, the loader fetches from an external API through a resource route.

It's based on this example: [https://github.com/remix-run/examples/tree/main/file-and-s3-upload](Remix examples)

`export async function loader({ request, params }: LoaderArgs) {
  try {
    // extract all the parameters from the url
    const { key, width, height, fit, format } = extractParams(params, request);

    if (
      (width && !ALLOWED_IMAGE_SIZES.includes(width)) ||
      (height && !ALLOWED_IMAGE_SIZES.includes(height))
    ) {
      throw new Error(`Image size not allowed`);
    }
    return downloadImage(key, width, height, fit, format);
  } catch (error: unknown) {
    // if the image is not found, or we get any other errors we return different response types
    return handleError(error);
  }
}`

The downloadImage function calls an external API:

` return fetch(
      `${process.env.API_URL}/images/${imageKey}?width=${width}&height=${height}&fit=${fit}&format=${format}`
    );
`

And the external API streams the image:

`
router.get("/:imageKey", async (req, res) => {
  try {
    const { imageKey } = req.params;
[...]

PR #4782, opened 2022-12-06
Remove instanceof Response checks in favor of isResponse

brophdawg11:
Fixes https://github.com/remix-run/remix/issues/4741 by using `isResponse` instead of `instanceof Response`

Note: integration tests will fail until we get `@remix-run/server-runtime` onto the latest `@remix-run/router`

### DIFF
--- a/integration/loader-test.ts
+++ b/integration/loader-test.ts
@@ -69,17 +69,22 @@ test.describe("loader in an app", () => {
   let appFixture: AppFixture;
 
   let HOME_PAGE_TEXT = "hello world";
+  let REDIRECT_TARGET_TEXT = "redirect target";
+  let FETCH_TARGET_TEXT = "fetch target";
 
   test.beforeAll(async () => {
     appFixture = await createAppFixture(
       await createFixture({
         files: {
           "app/root.jsx": js`
+            import { Outlet } from '@remix-run/react'
+
             export default function Root() {
               return (
                 <html>
                   <body>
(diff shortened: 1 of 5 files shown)


integration/loader-test.ts
```diff
@@ -103,5 +124,12 @@ test.describe("loader in an app", () => {
     let app = new PlaywrightFixture(appFixture, page);
     await app.goto("/redirect");
     expect(await app.getHtml()).toMatch(HOME_PAGE_TEXT);
+    expect(await app.getHtml()).toMatch(REDIRECT_TARGET_TEXT);
```
brophdawg11, line 127: Strengthened the existing redirect tests a bit to ensure it renders the right content at the redirected location.

integration/loader-test.ts
```diff
@@ -69,27 +69,48 @@ test.describe("loader in an app", () => {
...
-            export const loader = () => redirect("/");
+            export const loader = () => redirect("/redirect-target");
             export default () => <div>Yo</div>
           `,
+          "app/routes/redirect-target.jsx": js`
+            export default () => <div>${REDIRECT_TARGET_TEXT}</div>
+          `,
+          "app/routes/fetch.jsx": js`
+            export function loader({ request }) {
+              return fetch(new URL(request.url).origin + '/fetch-target');
```
brophdawg11, line 103: This is the case we're fixing here - returning a raw `fetch` `Response` wasn't matching our `instanceof Response` check

PR #4790, opened 2022-12-06
Lift static handler creation and pass context through query/queryRoute

### LINKED ISSUE #4733: createRequestHandler in server-runtime causing slower responses in larger apps

brophdawg11:
Updates to `@remix-run/router@1.0.5-pre.1` with support for `requestContext` in `query`/`queryRoute` and lifts the static handler creation outside of `requestHandler`

Closes #4733

### DIFF
--- a/packages/remix-server-runtime/routes.ts
+++ b/packages/remix-server-runtime/routes.ts
@@ -1,11 +1,9 @@
-// TODO: RRR - Change import to @remix-run/router
 import type {
   AgnosticDataRouteObject,
   ActionFunctionArgs,
   LoaderFunctionArgs,
 } from "@remix-run/router";
 
-import { type AppLoadContext } from "./data";
 import { callRouteActionRR, callRouteLoaderRR } from "./data";
 import type { ServerRouteModule } from "./routeModules";
 
@@ -56,7 +54,6 @@ export function createRoutes(
 // createStaticHandler
 export function createStaticHandlerDataRoutes(
   manifest: ServerRouteManifest,
-  loadContext: AppLoadContext,
   parentId?: string
 ): AgnosticDataRouteObject[] {
   return Object.values(manifest)
@@ -73,8 +70,9 @@ export function createStaticHandlerDataRoutes(
         loader: route.module.loader
           ? (args: LoaderFunctionArgs) =>
               callRouteLoaderRR({
-                ...args,
-                loadContext,
+                request: args.request,
+                params: args.params,
+                loadContext: args.context,
                 loader: route.module.loader!,
                 routeId: route.id,
               })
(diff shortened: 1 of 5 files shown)


packages/remix-server-runtime/routes.ts
```diff
@@ -73,17 +70,19 @@ export function createStaticHandlerDataRoutes(
         loader: route.module.loader
           ? (args: LoaderFunctionArgs) =>
               callRouteLoaderRR({
-                ...args,
-                loadContext,
+                request: args.request,
+                params: args.params,
+                loadContext: args.context,
```
brophdawg11, line 75: No longer comes from the closure scope and instead passes straight through

packages/remix-server-runtime/server.ts
```diff
@@ -99,10 +100,14 @@ async function handleDataRequestRR(
...
-  request: Request
+  request: Request,
+  loadContext: AppLoadContext
 ) {
   try {
-    let response = await staticHandler.queryRoute(request, routeId);
+    let response = await staticHandler.queryRoute(request, {
+      routeId,
+      requestContext: loadContext,
+    });
```
brophdawg11, line 110: Send the context through `query`/`queryRoute`

PR #4731, opened 2022-12-01
refactor(remix-react): upgrade Remix to `react-router-dom@6.4` (non-data-router) and drop `history`

brophdawg11:
This upgrades Remix to use the latest `react-router-dom@6.4.4` release for rendering _without yet using the new `RouterProvider`_.  This still renders with `StaticRouter` on the server and and a raw `Router` on the client.  The difference is that we're now using the new `createBrowserHistory` + `v5Compat` flag from `@remix-run/router` and we drop the `history@5` dependency.

Subsequent releases will switch from `StaticRouter`/`createBrowserHistory`/`Router` to `StaticRouterProvider`/`createBrowserRouter`/`RouterProvider`

PR #4801, opened 2022-12-07
refactor(remix-server-runtime): remove `react` & `react-dom` from `peerDependencies`

MichaelDeBoey:
Follow-up of #4731

### DIFF
--- a/packages/remix-server-runtime/package.json
+++ b/packages/remix-server-runtime/package.json
@@ -18,6 +18,7 @@
   "dependencies": {
     "@remix-run/router": "1.3.0",
     "@types/cookie": "^0.4.0",
+    "@types/react": "^18.0.15",
     "@web3-storage/multipart-parser": "^1.0.0",
     "cookie": "^0.4.1",
     "set-cookie-parser": "^2.4.8",
@@ -25,13 +26,7 @@
   },
   "devDependencies": {
     "@remix-run/web-file": "^3.0.2",
-    "@types/set-cookie-parser": "^2.4.1",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0"
-  },
-  "peerDependencies": {
-    "react": ">=16.8",
-    "react-dom": ">=16.8"
+    "@types/set-cookie-parser": "^2.4.1"
   },
   "engines": {
     "node": ">=14"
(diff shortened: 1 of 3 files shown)


packages/remix-server-runtime/package.json
```diff
@@ -18,20 +18,15 @@
   "dependencies": {
     "@remix-run/router": "1.0.4",
     "@types/cookie": "^0.4.0",
+    "@types/react": "^18.0.15",
```
MichaelDeBoey, line 21: We need this, because we're using `ComponentType` type in `packages/remix-server-runtime/routeModules.ts`.

We should probably move that to `@remix-run/react` and make `@remix-run/server-runtime` more renderer-agnostic in that place.

packages/remix-server-runtime/routeModules.ts
```diff
@@ -1,5 +1,5 @@
+import type { Location, Params } from "@remix-run/router";
 import type { ComponentType } from "react";
-import type { Location, Params } from "react-router-dom";
```
MichaelDeBoey, line 2: `react-router-dom` isn't in `dependencies` anymore, so this would always be invalid

Probably a good idea to add an ESLint rule to catch these kind of things? 🤔

brophdawg11: yeah - I would have actually expected that to fail on the build but I guess the yarn workspace bites us there?  I feel like I recall this coming up in the past and maybe we never found a good solution then.  Do you know of one?  I've used https://www.npmjs.com/package/eslint-import-resolver-webpack before but that won't work here 😄

packages/remix-server-runtime/package.json
MichaelDeBoey: It's fixed now, so no need to rush it 🤷‍♂️

Just wanted to notice you (and the team)

packages/remix-server-runtime/routeModules.ts
brophdawg11: ok yeah that feels like what we need, doesn't need to go so far as the webpack one above and actually resolve since we don't use aliases and such

mcansh: `eslint-plugin-import` should do the trick, specifically the `import/no-extraneous-dependencies` rule

MichaelDeBoey: That's indeed the rule I was talking about!

I'll create a separate PR to enable it, so we're sure this never happens again

MichaelDeBoey:
@brophdawg11 @mcansh Anything that's preventing us from merging this one?

pcattori (approved):

mcansh (approved):

PR #4803, opened 2022-12-08
Add fetcher state/type tests

brophdawg11:
Added fine-grained tests around `fetcher.type` that we can use to ensure we get the "Remix on Router 6.4" back-compat layer implemented correctly

### DIFF
--- a/integration/fetcher-state-test.ts
+++ b/integration/fetcher-state-test.ts
@@ -199,8 +199,7 @@ test.describe("fetcher states", () => {
         type: "loaderSubmission",
         submission: {
           formData: { key: "value" },
-          // TODO: This feels potentially wrong - the formAction is just /page,
-          // the request URL is /page?key=value
+          // TODO: I think this is a bug and will be fixed with RR 6.4
           action: "/page?key=value",
           method: "GET",
           encType: "application/x-www-form-urlencoded",
--- a/integration/fetcher-state-test.ts
+++ b/integration/fetcher-state-test.ts
@@ -199,7 +199,9 @@ test.describe("fetcher states", () => {
         type: "loaderSubmission",
         submission: {
           formData: { key: "value" },
-          // TODO: I think this is a bug and will be fixed with RR 6.4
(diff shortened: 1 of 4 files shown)


packages/remix-react/package.json
```diff
@@ -16,8 +16,8 @@
   "typings": "dist/index.d.ts",
   "module": "dist/esm/index.js",
   "dependencies": {
-    "@remix-run/router": "1.0.4",
-    "react-router-dom": "6.4.4"
+    "@remix-run/router": "1.0.5",
```
brophdawg11, line 19: This was out of sync with the version used in `@remix-run/server-runtime` which caused issues during integration tests.  I don't _think_ this would be an actual problem in apps since server-runtime had it's own copy nested in a local `node_modules`, however we do some tricky stuff with our integration tests builds and I think everything reads from the global repo `node_modules` so it was picking up on the wrong version from `@remix-run/react`. 

Generally speaking we'll upgrade these in lock step going forward so I expect this is largely a non-issue - this mismatch is just a function of the PR to update to `react-router-dom@6.4.4` and the Remix `1.8.2` hotfix release that bumped to `@remix-run/router@1.0.5` merging side by side.

integration/fetcher-state-test.ts
```diff
@@ -0,0 +1,271 @@
...
+        state: "submitting",
+        type: "loaderSubmission",
+        submission: {
+          formData: { key: "value" },
+          // TODO: This feels potentially wrong - the formAction is just /page,
+          // the request URL is /page?key=value
+          action: "/page?key=value",
+          method: "GET",
+          encType: "application/x-www-form-urlencoded",
+        },
```
brophdawg11, line 207: @jacob-ebey What do you think here?  IIRC we specifically tested against this in `@remix-run/router` so the `submission` stayed true to the state of the DOM and only the `request` represented the serialized `URLSearchParams`

brophdawg11: Same behavior happens for `loaderSubmission` transitions - updated the test in `transition-state-test` below with the same logic

PR #4808, opened 2022-12-09
docs(guide/data-writes): add missing json imports

petetnt:
- [x] Docs

This PR fixes missing `json` in `data-writes.md` document. At certain point in the guide we return the validation errors and form values with `json`, but it isn't actually imported anywhere. This PR makes it so that `json` is always imported too wherever the other used action `redirect` is mentioned.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -359,6 +359,7 @@
 - pcattori
 - penspinner
 - penx
+- petetnt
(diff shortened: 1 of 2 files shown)


machour (approved):

machour:
Thank you!

PR #4826, opened 2022-12-11
bug: v1.8 loaders have empty request.headers when called after action

nrako:
This is a bug report: `yarn bug-report-test`

---

Prior to Remix v1.8.0 the loader function when called subsequentially to a (POST) action would have the request's `headers` set. Since v1.8 this is not the case anymore.

This is highlyt problematic. For example, if you have some high level session based cookie which is used to guard GET request in the page action (or a parent layout route); the authentication would fail every time the loaders data is retrieved subsequentially to a POST action.

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -49,26 +49,26 @@ test.beforeAll(async () => {
     files: {
       "app/routes/index.jsx": js`
         import { json } from "@remix-run/node";
-        import { useLoaderData, Link } from "@remix-run/react";
+        import { useLoaderData, useActionData } from "@remix-run/react";
 
-        export function loader() {
-          return json("pizza");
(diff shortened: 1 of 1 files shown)


brophdawg11:
Thanks!  Moved this out of the bug report test and added some additional test in #4829.  The underlying fix for this is over in react-router: https://github.com/remix-run/react-router/pull/9721

brophdawg11:
Hey @nrako - just FYI this fix is available in a `1.9.0-pre.0` release if you wanted to give it a test!  We're hoping to get the stable `1.9.0` out later today or tomorrow.

nrako:
Thanks for letting me know @brophdawg11, I did a quick test with `@remix-run/server-runtime` and that looks great, my tests are all green.

PR #4827, opened 2022-12-11
docs(api/remix): fix link to `PrefetchPageLinks` page

MichaelDeBoey:


machour (commented):
How about moved-68 instead like the rest of the document for consistency?

MichaelDeBoey:
@machour This will be auto-fixed by the format workflow

Don't know if it will shift all links or not, that's why I kept it as a normal link

machour (approved):
Let's give it a try 👀

MichaelDeBoey:
@machour As expected: https://github.com/remix-run/remix/commit/534e1ec071f17654a4db8622e30d6ff70548ce26

machour:
Always learning new things with you 😄

Issue #4791, opened 2022-12-07
`useCatch` in root `CatchBoundary` returns nested route's data if root and nested loader throw


nikosdouvlis:
### What version of Remix are you using?

1.8.1

### Steps to Reproduce

- Create a new Remix app by running `npx create-remix@latest`. The stack does not matter, so feel free to choose the default configuration (TS + Remix App Server).
- Export a `loader` from `root.tsx` that throws a json response
- Export a CatchBoundary from `root.tsx` that renders the data returned by calling `useCatch`
- Create a new route, `routes/withBoundary.tsx`. This route exports a `loader` that throws a json response. It also exports a CatchBoundary.
- Create a new route, `routes/withoutBoundary.tsx`. This route exports a `loader` that throws a json response. It does not export a CatchBoundary.
- Run `npm run dev` to start the app

### Expected Behavior

- Navigating to `/`: the root CatchBoundary should render and useCatch will return the data of the root loader response.
- Navigating to `/withBoundary`: the root CatchBoundary should render and useCatch will return the data of the root loader response.
- Navigating to `/withoutBoundary`: the root CatchBoundary should render and useCatch will return the data of the root loader response.

### Actual Behavior

- Navigating to `/`: the root CatchBoundary renders, and useCatch returns the data of the root loader response.
- Navigating to `/withBoundary`: the root CatchBoundary renders and useCatch returns of the root loader response.
- Navigating to `/withoutBoundary`: the root CatchBoundary renders **but useCatch returns the data of the `/withoutBoundary` loader response instead**.

## More context
Hello everyone! First things first... Thanks for the fantastic work you've been doing on Remix and RR; both are a joy to use. 

[Clerk](https://clerk.dev/) added support for Remix apps last February when we released the `@clerk/remix` package, receiving great feedback from our community. Our auth mechanism is based on auto-refreshing short-lived JWTs. We use the root `loader` and root `CatchBoundary` to get a fresh token if we detect a stale one.

We've been investigating an issue our users reported following the Remix 1.8 release. It looks like the behavior of the `useCatch`/`CatchBoundary` changed between Remix versions 1.7.6 and 1.8, breaking our integration for some specific edge cases.

Maintaining OS projects is hard, so we hope the "Debugging" and "Reproducible examples" section below are helpful. Would you let us know if this is a breaking change or just a bug that slipped in the 1.8 release?

## Debugging
[...]

PR #4829, opened 2022-12-12
Add integration tests for request structures

brophdawg11:
Adds additional tests to the bug report in #4826.  I went back and made sure all these tests pass on `1.7.6` and then with the sibling PR in `react-router` they all pass on 1.8.x Remix now as well.

_Note: CI will fail until we release RR and update `server-runtime`'s `package.json` in this branch_

React Router PR: https://github.com/remix-run/react-router/pull/9721

Can test via:
* Check out this branch in remix
* Check out the corresponding branch in react-router
* In `react-router`, 
  * `LOCAL_BUILD_DIRECTORY=../remix yarn build`
  * Assumes your `remix` and `react-router` repo clones are sibling directories
* In `remix`:
  * `yarn test:integration request-test --project chromium` 

Also, since this PR will update the `@remix-run/router` dependency I also added an integration tests for https://github.com/remix-run/remix/issues/4791 which ids fixed by https://github.com/remix-run/react-router/pull/9702

### DIFF
--- a/integration/catch-boundary-test.ts
+++ b/integration/catch-boundary-test.ts
@@ -67,6 +67,9 @@ test.describe("CatchBoundary", () => {
                 <Link to="${HAS_BOUNDARY_LOADER}">
                   ${HAS_BOUNDARY_LOADER}
                 </Link>
+                <Link to="${HAS_BOUNDARY_LOADER}/child">
+                  ${HAS_BOUNDARY_LOADER}/child
+                </Link>
                 <Link to="${NO_BOUNDARY_LOADER}">
                   ${NO_BOUNDARY_LOADER}
                 </Link>
@@ -111,11 +114,27 @@ test.describe("CatchBoundary", () => {
         `,
 
         [`app/routes${HAS_BOUNDARY_LOADER}.jsx`]: js`
+          import { useCatch } from '@remix-run/react';
           export function loader() {
             throw new Response("", { status: 401 })
           }
           export function CatchBoundary() {
-            return <div id="boundary-loader">${OWN_BOUNDARY_TEXT}</div>
+            let caught = useCatch();
+            return (
+              <>
+                <div id="boundary-loader">${OWN_BOUNDARY_TEXT}</div>
+                <pre id="status">{caught.status}</pre>
+              </>
+            );
+          }
+          export default function Index() {
+            return <div/>
+          }
+        `,
+
+        [`app/routes${HAS_BOUNDARY_LOADER}/child.jsx`]: js`
+          export function loader() {
+            throw new Response("", { status: 404 })
           }
           export default function Index() {
             return <div/>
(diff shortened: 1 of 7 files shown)


integration/request-test.ts
```diff
@@ -0,0 +1,162 @@
...
+  expect(actionData.method).toEqual("POST");
+  expect(actionData.url).toMatch(/^http:\/\/localhost:\d+\/$/);
+  expect(actionData.headers.cookie).toEqual("cookie=nomnom");
+  expect(actionData.text).toEqual("type=ssr");
+
+  loaderData = JSON.parse(await page.locator("#loader-data").innerHTML());
+  expect(loaderData.method).toEqual("POST");
+  expect(loaderData.url).toMatch(/^http:\/\/localhost:\d+\/$/);
+  expect(loaderData.headers.cookie).toEqual("cookie=nomnom");
+  expect(loaderData.text).toEqual("");
```
brophdawg11, line 135: During MPA submissions, the loader `request` proxies the action method (`POST`) but strips the `body`.  This makes sense since there's only a single submission happening.

integration/request-test.ts
```diff
@@ -0,0 +1,162 @@
...
+  expect(actionData.method).toEqual("POST");
+  expect(actionData.url).toMatch(/^http:\/\/localhost:\d+\/$/);
+  expect(actionData.headers.cookie).toEqual("cookie=nomnom");
+  expect(actionData.text).toEqual("type=csr");
+
+  loaderData = JSON.parse(await page.locator("#loader-data").innerHTML());
+  expect(loaderData.method).toEqual("GET");
+  expect(loaderData.url).toMatch(/^http:\/\/localhost:\d+\/$/);
+  expect(loaderData.headers.cookie).toEqual("cookie=nomnom");
+  expect(loaderData.text).toEqual("");
```
brophdawg11, line 161: During SPA submissions, the loader is called through separate revalidation `GET` requests, so the assertions here differ on the method.

I think this all makes sense given how Remix is executing loaders differently in MPA/SPA calls, even if it introduces a slight different between them from a DX standpoint.  Users shouldn't be branching on `request.method` in loaders anyway?

brophdawg11: This logic is implemented in the router here: https://github.com/remix-run/react-router/pull/9721

integration/catch-boundary-test.ts
```diff
@@ -292,4 +311,23 @@ test.describe("CatchBoundary", () => {
     expect(await app.getHtml("#child-catch")).toMatch("400");
     expect(await app.getHtml("#child-catch")).toMatch("Caught!");
   });
+
+  test("prefers parent catch when child loader also bubbles, document request", async () => {
+    let res = await fixture.requestDocument(`${HAS_BOUNDARY_LOADER}/child`);
+    expect(res.status).toBe(401);
```
brophdawg11, line 317: parent throws 401, child throws 404

integration/request-test.ts
```diff
@@ -0,0 +1,162 @@
...
+  await app.clickElement("#submit-post-ssr");
+
+  let actionData = JSON.parse(await page.locator("#action-data").innerHTML());
+  expect(actionData.method).toEqual("POST");
+  expect(actionData.url).toMatch(/^http:\/\/localhost:\d+\/$/);
+  expect(actionData.headers.cookie).toEqual("cookie=nomnom");
+  expect(actionData.text).toEqual("type=ssr");
+
+  loaderData = JSON.parse(await page.locator("#loader-data").innerHTML());
+  expect(loaderData.method).toEqual("GET");
```
brophdawg11, line 132: Loaders receive `method:GET` on SSR "revalidations" now.  This fixes a bug in Remix 1.7.6 where the loader `request` was `POST` but without a `body`, and thus made for a different DX if you did a SSR/CSR submissions since `request.method` would be different in the `loader`.

PR #4841, opened 2022-12-12
feat(remix-dev): deprecate `serverBuildTarget`

MichaelDeBoey:
As discussed in the second roadmap livestream

---

Closes #3556


jenseng (approved):

jenseng:
Between this and #4842 are we missing anything? I see there are some places where we key off of this in the compiler (e.g. [here](https://github.com/remix-run/remix/blob/3832d457b34d7a62633bf86a2606971263e710cb/packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts#L24)), do we need additional config to support this behavior?

MichaelDeBoey:
@jenseng Good call!
That's indeed something we need to figure out in some other way 🤔

Listing all the usages:

- https://github.com/remix-run/remix/blob/6a6fa5911e09667c1788451512956ba636aca043/packages/remix-dev/compiler/compilerServer.ts#L45-L48
- https://github.com/remix-run/remix/blob/6a6fa5911e09667c1788451512956ba636aca043/packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts#L24
- https://github.com/remix-run/remix/blob/6a6fa5911e09667c1788451512956ba636aca043/packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts#L95-L101
- https://github.com/remix-run/remix/blob/6a6fa5911e09667c1788451512956ba636aca043/packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts#L116-L117

MichaelDeBoey:
@Harmony7 mentioned the same problem in #4860 regarding not being able to fully deprecate `serverBuildtarget`

> This PR is necessary because, except for the first two subitems in the list above, these behaviors cannot be set through `remix.config.js`.

mcansh:
bundling everything should be achievable using `serverDependenciesToBundle: /.*/`

MichaelDeBoey:
@mcansh So how do you think that would that look like?
Creating new `serverConditions`, `serverMinify` & `serverMainFields` config options?

mjackson:
> new serverConditions, serverMinify & serverMainFields config options

Yeah, @MichaelDeBoey. I think we will need these.

> If so, I would suggest to go with a server object instead, to not pollute the global config namespace.

No, let's keep the config object as flat as possible.

MichaelDeBoey:
@mjackson I introduced the `serverConditions`, `serverMinify` & `serverMainFields` config options and updated `serverDependenciesToBundle` so that it can also receive `"all"` as value.

The only places where we're now still using `serverBuildTarget` is in `serverBareModulesPlugin`:
- https://github.com/remix-run/remix/blob/6a6fa5911e09667c1788451512956ba636aca043/packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts#L24-L37
- https://github.com/remix-run/remix/blob/6a6fa5911e09667c1788451512956ba636aca043/packages/remix-dev/compiler/plugins/serverBareModulesPlugin.ts#L116-L117

Can't think of a good way to remove usage here though 🤔

packages/remix-dev/__tests__/readConfig-test.ts
```diff
@@ -65,15 +65,11 @@ describe("readConfig", () => {
...
         "serverBuildPath": Any<String>,
-        "serverBuildTarget": undefined,
         "serverBuildTargetEntryModule": "export * from \\"@remix-run/dev/server-build\\";",
         "serverConditions": undefined,
         "serverDependenciesToBundle": Array [],
         "serverEntryPoint": undefined,
-        "serverMainFields": Array [
-          "main",
-          "module",
-        ],
```
MichaelDeBoey: This should be kept, otherwise we're not doing the same thing anymore when compiling

packages/remix-dev/config.ts
```diff
@@ -298,7 +292,7 @@ export interface RemixConfig {
    * The order of main fields to use when resolving server dependencies.
    * Defaults to `["main", "module"]`.
    */
-  serverMainFields: string[];
+  serverMainFields?: string[];
```
MichaelDeBoey, line 295:
```suggestion
  serverMainFields: string[];
```

packages/remix-dev/config.ts
```diff
@@ -375,41 +369,66 @@ export async function readConfig(
...
+  serverConditions = appConfig.serverConditions
+    ? appConfig.serverConditions
+    : serverConditions;
+  serverMainFields = appConfig.serverMainFields
+    ? appConfig.serverMainFields
+    : serverMainFields;
+  serverDependenciesToBundle =
+    (appConfig.serverDependenciesToBundle
+      ? appConfig.serverDependenciesToBundle
+      : serverDependenciesToBundle) || [];
```
MichaelDeBoey, line 422: This can be simplified to

```suggestion
  serverConditions = appConfig.serverConditions ?? serverConditions;
  serverMainFields = appConfig.serverMainFields ?? serverMainFields;
  serverDependenciesToBundle =
    appConfig.serverDependenciesToBundle ?? serverDependenciesToBundle;
```

packages/remix-dev/config.ts
```diff
@@ -375,41 +369,66 @@ export async function readConfig(
...
-  );
-  let isDenoRuntime = appConfig.serverBuildTarget === "deno";
+  // TODO: default to esm for v2
+  let serverModuleFormat: ServerModuleFormat = "cjs";
+  let serverPlatform: ServerPlatform = "node";
+  let serverConditions = undefined;
+  let serverMainFields = undefined;
+  let serverBuildPath = "build/index.js";
+  let defaultPublicPath = "/build/";
+  let serverDependenciesToBundle: "all" | Array<string | RegExp> | undefined;
```
MichaelDeBoey, line 379:
```suggestion
  let serverDependenciesToBundle: "all" | Array<string | RegExp> = [];
```

packages/remix-dev/config.ts
MichaelDeBoey: We also need to override `serverModuleFormat` & `serverPlatform` with config + **always** set `serverMainFields` to `["module", "main"]` if `serverModuleFormat` is set to `"esm"`

packages/remix-dev/config.ts
```diff
@@ -342,21 +369,67 @@ export async function readConfig(
...
+    : serverConditions;
+  serverMainFields = appConfig.serverMainFields
+    ? appConfig.serverMainFields
+    : serverMainFields;
+  serverDependenciesToBundle =
+    (appConfig.serverDependenciesToBundle
+      ? appConfig.serverDependenciesToBundle
+      : serverDependenciesToBundle) || [];
+
+  serverBuildPath = path.resolve(rootDirectory, serverBuildPath);
```
MichaelDeBoey, line 424: You're still missing some things that I extracted into `resolveServerBuildPath` here

```ts
// retain deprecated behavior for now
  if (appConfig.serverBuildDirectory) {
    serverBuildPath = path.join(appConfig.serverBuildDirectory, "index.js");
  }

  if (appConfig.serverBuildPath) {
    serverBuildPath = appConfig.serverBuildPath;
  }
```

packages/remix-dev/config.ts
```diff
@@ -263,12 +263,6 @@ export interface RemixConfig {
    */
   serverBuildPath: string;
 
-  /**
-   * The target of the server build. Defaults to "node-cjs".
-   *
-   * @deprecated Use a combination of `{@link AppConfig.publicPath}`, `{@link AppConfig.serverBuildPath}`, `{@link AppConfig.serverConditions}`, `{@link AppConfig.serverDependenciesToBundle}`, `{@link AppConfig.serverMainFields}`, `{@link AppConfig.serverMinify}`, `{@link AppConfig.serverModuleFormat}` and/or `{@link AppConfig.serverPlatform}` instead.   */
-  serverBuildTarget?: ServerBuildTarget;
-
```
MichaelDeBoey, line 263: Since we're exporting `RemixConfig` type (as `ResolvedRemixConfig`), we should keep this until the next major release, as removing it would be a breaking change otherwise.
https://github.com/remix-run/remix/blob/837517ef13e6e52acaea7adc8d2c1648aed11ae9/packages/remix-dev/index.ts#L3

```suggestion
  /**
   * The target of the server build. Defaults to "node-cjs".
   *
   * @deprecated Use a combination of `{@link AppConfig.publicPath}`, `{@link AppConfig.serverBuildPath}`, `{@link AppConfig.serverConditions}`, `{@link AppConfig.serverDependenciesToBundle}`, `{@link AppConfig.serverMainFields}`, `{@link AppConfig.serverMinify}`, `{@link AppConfig.serverModuleFormat}` and/or `{@link AppConfig.serverPlatform}` instead.
   */
  serverBuildTarget?: ServerBuildTarget;

```

packages/remix-dev/config.ts
```diff
@@ -540,7 +554,6 @@ export async function readConfig(
     rootDirectory,
     routes,
     serverBuildPath,
-    serverBuildTarget,
```
MichaelDeBoey:
```suggestion
    serverBuildTarget,
```

packages/remix-dev/config.ts
```diff
@@ -342,20 +375,41 @@ export async function readConfig(
...
+    serverPlatform = "neutral";
+  }
+  if (isDenoRuntime) {
+    serverConditions ??= ["deno", "worker"];
+    serverDependenciesToBundle = "all";
+    serverMainFields ??= ["module", "main"];
+    serverModuleFormat = "esm";
+    serverPlatform = "neutral";
   }
+  serverMainFields ??=
```
jacob-ebey, line 410: This should be 
```js
serverMainFields = appConfig.serverMainFields ?? serverMainFields ?? serverModuleFormat === "esm" ? ["module", "main"] : ["main", "module"];
```

MichaelDeBoey: Hmm that's weird 🤔

Because in the declaration of the variable, I'm setting it to `appConfig.serverMainFields` & I'm only overriding it in the `if`-statements if it doesn't have a value yet
Same for the `esm` check: only doing that if it doesn't have a value yet

harmony7:
awesome stuff! =)

TrySound:
Client conditions would be useful too
https://github.com/remix-run/remix/discussions/6003

PR #4844, opened 2022-12-12
ScrollRestoration on RR 6.4

brophdawg11:
Update to `ScrollRestoration` to wrap the RR 6.4 internals

Testing:
* ~Needs to be run with changes from https://github.com/remix-run/react-router/pull/9664~
* Confirm client side scroll restoration works on back/forward button routing
* Confirm SSR restoration works on hard reloads
* Confirm `<ScrollRestoration getKey={(location, matches) => ...} />` works
* Confirm `<Link preventScrollReset>` works


packages/remix-react/components.tsx
```diff
@@ -863,29 +863,6 @@ function dedupe(array: any[]) {
...
- *
- * Note: The `callback` argument should be a function created with
- * `React.useCallback()`.
- *
- * @see https://remix.run/api/remix#usebeforeunload
- */
-export function useBeforeUnload(
-  callback: (event: BeforeUnloadEvent) => any
-): void {
-  // TODO: Export from react-router-dom
```
brophdawg11, line 880: ✅

packages/remix-react/scroll-restoration.tsx
```diff
@@ -1,49 +1,62 @@
 import * as React from "react";
-import { useLocation } from "react-router-dom";
+import type { ScrollRestorationProps as ScrollRestorationPropsRR } from "react-router-dom";
+import {
+  useLocation,
+  UNSAFE_useScrollRestoration as useScrollRestoration,
```
brophdawg11, line 5: The hook was private in both Remix and RR so kept it that way via the `UNSAFE_` prefix

packages/remix-react/scroll-restoration.tsx
```diff
@@ -1,49 +1,62 @@
...
+}: ScriptProps & {
+  getKey: ScrollRestorationPropsRR["getKey"];
+}) {
+  let location = useLocation();
+  let matches = useMatches();
+
+  useScrollRestoration({
+    getKey,
+    storageKey: STORAGE_KEY,
+    skip: !hydrated,
```
brophdawg11, line 32: Matt to so some double checking here and see if there's any way `skip` could go away...

brophdawg11: If we assume `hydrationData` means "skip the first restoration" then we could get rid of this

Issue #4632, opened 2022-11-17
🗺️ CSS Modules


markdalgleish:
## Progress

- [x] RFC https://github.com/remix-run/remix/discussions/4631
- [x] PR: https://github.com/remix-run/remix/pull/4852

## Open questions

- What format should we use for scoped class names and should it differ between dev and prod? Should it be configurable? e.g. it's common to have debuggable class names during development, like `.Button-module__root__a4X9gs`, but only emit the hash in production, like `.a4X9gs`.

## Notes

This work will also support node_modules that `import "./blah.css"` and `import {} from "./blah.module.css"` so that folks design systems don't blow up.

ryanflorence:
@jacob-ebey can you find the package that generates types and link it here?

jacob-ebey:
https://www.npmjs.com/package/typescript-plugin-css-modules

narutosstudent:
@markdalgleish Excited for this! 🔥  🙏  😁

machour:
@laurencefass have a look at the [CSS Modules RFC](https://github.com/remix-run/remix/discussions/4631) linked above, and comment there if you see something missing.

markdalgleish:
I noticed that a note had been added to this issue about CSS side effect imports (`import "./styles.css"`) but in my mind this is a separate piece of work that builds on top of this foundation. I've split out a standalone proposal for this: https://github.com/remix-run/remix/discussions/4775

markdalgleish:
**Next steps:**  @chaance to follow up on PR review changes, approve if everything looks good.

PR #4852, opened 2022-12-13
feat: CSS Modules support

markdalgleish:
Closes: https://github.com/remix-run/remix/issues/4632

- [x] Docs
- [x] Tests

Note that docs will definitely some need more holistic changes once we've removed the feature flag and added more bundling features, but this is PR contains a minimal set of updates to document this change.

I ultimately went with [postcss-modules](https://www.npmjs.com/package/postcss-modules) as it's the most commonly used implementation in the wild and we're aiming to support people migrating from other tools. We may use other implementations alongside this in the future if it helps with performance and we can detect that unsupported features aren't being used.

## Testing Strategy

I've added a Playwright test suite that checks each CSS Modules feature individually since they have different levels of support with different implementations. I've ensured that all tests are decoupled from the generated class name format since it can differ between CSS Modules implementations.

I've also added some tests to catch potential issues in the future (duplicated hashes, duplicated rules) if we want to migrate to another implementation or change our CSS minification strategy.

Since there was already a deterministic build output test, I've added some CSS Modules files to the fixture to ensure hashes are stable.

## Kudos

Big shout out to @chaance for his prior work on https://github.com/remix-run/remix/pull/2489 which really helped in this implementation.


kentcdodds:
Amazing work Mark! Thank you 👏

.changeset/little-avocados-fetch.md
```diff
@@ -0,0 +1,9 @@
+---
+"@remix-run/css-bundle": patch
+"@remix-run/dev": patch
```
chaance, line 3: I believe we should bump to `minor` for this one since it's a new feature.

docs/guides/styling.md
```diff
@@ -766,6 +766,76 @@ Other CSS-in-JS libraries will have a similar setup. If you've got a CSS framewo
...
+
+```sh
+npm install @remix-run/css-bundle
+```
+
+Then, import the CSS bundle's `href` and add it to a link descriptor, most likely in `root.tsx` so that it applies to your entire application.
+
+```tsx filename=root.tsx lines=[2,6]
+import type { LinksFunction } from "@remix-run/node"; // or cloudflare/deno
+import cssBundleHref from "@remix-run/css-bundle";
```
chaance, line 802: I'm wondering if we shouldn't use a more specific virtual here for CSS Modules. If we take a similar approach with raw CSS imports and Vanilla Extract, it could be nice to have access to each of those bundles independently (ex: your admin routes are styled w/ CSS Modules, marketing pages use some other bundle)

devongovett:
Hey, anything I can do to convince you to use lightningcss? What caused you to migrate away and what can I do to help?

docs/guides/styling.md
chaance: Perhaps we could achieve both?

```ts
// all bundles in one stylesheet
import cssBundleStylesheet from "@remix-run/css-bundle";

// split up
import cssModulesStylesheet from "@remix-run/css-bundle/css-modules";
import vanillaExtractStylesheet from "@remix-run/css-bundle/vanilla-extract";
```

integration/deterministic-build-output-test.ts
```diff
@@ -15,31 +15,63 @@ test("builds deterministically under different paths", async () => {
...
   //  * mdx (via app/routes/index.mdx)
   //  * serverAssetsManifestPlugin (implicitly tested by build)
   //  * serverEntryModulePlugin (implicitly tested by build)
   //  * serverRouteModulesPlugin (implicitly tested by build)
   let init = {
     files: {
+      "remix.config.js": js`
+        module.exports = {
+          future: {
+            unstable_cssBundle: true
```
chaance, line 30: I wonder if we could also make this config a bit more granular? That way we could continue experimenting with the same API in the future once CSS Modules are stable.

```ts
{
  future: {
    unstable_cssBundle: {
      cssModules: true, 
      vanillaExtract: true,
    },
  },
}
```

integration/deterministic-build-output-test.ts
```diff
@@ -15,31 +15,63 @@ test("builds deterministically under different paths", async () => {
...
         export * from "~/foo/bar.server";
-        export default () => "YAY";
+        import styles from "~/styles/foo.module.css";
+        export default () => <div className={styles.foo}>YAY</div>;
       `,
       "app/foo/bar.server.ts": "export const meta = () => []",
+      "app/styles/foo.module.css": css`
+        .foo {
+          background-image: url(~/images/foo.svg);
+          composes: bar from "~/styles/bar.module.css";
```
chaance, line 44: 😍

chaance (commented):
I'll take a bit more time to look over the compiler implementation, but I left a few comments on the API. **SO STOKED** 🔥

docs/guides/styling.md
markdalgleish: We don't get that level of granularity since we're now hooking into esbuild's CSS loader, which means we'll only get a single file at the end. I'm not convinced this is a big enough use case to warrant implementing our own CSS bundling logic, compared to bundle splitting of CSS for example.

It also sounds like a bit of a fragile pattern to me since nothing enforces that you're only using CSS Modules / Vanilla Extract / etc. within a given section of your application, leading to missing styles if you happen to use these features elsewhere.

markdalgleish:
@devongovett Lightning CSS was great to work with and I really like the subset of CSS Modules features you went with. If we were designing CSS Modules from scratch today, I like your limited feature set better. I just ultimately couldn't convince myself that it's a good idea for us to drop a bunch of CSS Modules features that currently work in basically every other tool and can be found in Stack Overflow answers, blog posts etc. — especially since we're wanting to support people who are migrating existing apps.

I saw that Parcel falls back to PostCSS if legacy features are detected. That's probably the main thing I'd need in order to switch back. I'd love to call `.bundleAsync` and have it transparently use PostCSS if it needs to. If we wanted to do this ourselves it would require a bit more work since we would't be able to use `.bundleAsync` anymore and would need to walk each dependency manually, checking for legacy features each step of the way.

docs/guides/styling.md
chaance: Good points. Perhaps another thing we attempt to solve later by splitting CSS at the route level.

devongovett:
I did the fallback for PostCSS in Parcel so as not to break anything in a minor version, but I would have considered dropping them entirely had it been a major version since people could always configure PostCSS themselves. Since Remix doesn't currently have support for CSS modules, now would be the time for you to do that. But I can understand the point about compatibility, and not sure if Remix lets users configure PostCSS themselves.

I would be open to adding support for more CSS modules features to Lightning CSS if it can help tools be able to adopt it. I didn't do it for the initial release because they seemed somewhat rarely used and we had the fallback on PostCSS so perf benefits would be minimal. I'd really like to see Lightning CSS adoption increase though, so whatever I can do to help, please let me know.

> I'd love to call .bundleAsync and have it transparently use PostCSS if it needs to.

You might be able to do it using a custom resolver, by implementing the `read` method and calling postcss before returning the result to the bundler. This would impact perf more than being implemented in Rust though.

markdalgleish:
> not sure if Remix lets users configure PostCSS themselves

We don't allow for configuration of PostCSS plugins within our compiler. We're only just introducing PostCSS now so this may be something we revisit in the future. For consumers who want to use PostCSS, the docs currently encourage consumers to manually run it as a separate process before passing the resulting files to Remix. However, this approach doesn't transparently work with CSS Modules and you'd need to manually import the generated JSON file as well which is a bit awkward.

> an option to enable legacy postcss features could work potentially

I think we ideally want to keep the number of flags and options to a minimum, especially when it comes to bundling features.

> You might be able to do it using a custom resolver, by implementing the read method and calling postcss before returning the result to the bundler. This would impact perf more than being implemented in Rust though.

Oh, that's an interesting idea. I might look into this.

integration/deterministic-build-output-test.ts
ryanflorence: I'm good with more granular but I'd rather not have nested flags, just keep them all flat.

.changeset/little-avocados-fetch.md
markdalgleish: Fixed.

integration/deterministic-build-output-test.ts
markdalgleish: I've flattened it to `future.unstable_cssModules`.

mjackson:
LGTM! ❤️

markdalgleish:
Once this has been merged, I'll raise some separate proposals for potential DX improvements that weren't seen as release blockers:

- Provide feedback when the feature flag isn't enabled but a `.module.css` file has been imported.
- Provide feedback when the CSS bundle isn't included in the page.
- Optionally generate .d.ts files for `.module.css` files to provide proper type checking, e.g. `styles.foo` is a type error if the style sheet doesn't contain a `.foo` class.

lili21:
is that mean all module stylesheet will be bundle into one css file?

lili21:
> is that mean all module stylesheet will be bundle into one css file?
I noticed that the RFC already mentioned it. Ignore me.

docs/guides/migrating-react-router-app.md
```diff
@@ -492,7 +492,7 @@ The answer to all of these questions is up to your bundler, _not you_. We think
 
 **Note:** Remix does not currently support CSS processing directly. If you use preprocessors like Sass, Less, or PostCSS, you can run those as a separate process in development.
 
-We also do not yet support CSS Modules, as that requires compiler integration and current approaches are not aligned with our design philosophy. We are actively working on a solution and plan to have an API for CSS Modules very soon.
+We also do not yet support [CSS Modules][css-modules] by default, but it is currently [available behind a feature flag.][css-modules]
```
chaance, line 495:
```suggestion
We do process [CSS Modules][css-modules], but support is currently [opt-in behind a feature flag][css-modules].
```

docs/guides/styling.md
```diff
@@ -766,6 +766,76 @@ Other CSS-in-JS libraries will have a similar setup. If you've got a CSS framewo
 
 NOTE: You may run into hydration warnings when using Styled Components. Hopefully [this issue][styled-components-issue] will be fixed soon.
 
+## CSS Modules
+
+<docs-warning>This feature is unstable and currently only available behind a feature flag. We're confident in the use cases it solves but the API and implementation may change in the future.</docs-warning>
+
+<docs-warning>Since all CSS Modules styles end up in a single CSS file at the end of the build, we recommend that you only use this styling approach for smaller applications, or sparingly within larger applications. If you're concerned about bundle size, you should probably look at [Tailwind](#tailwind-css) or [Regular Stylesheets](#regular-stylesheets) instead.
```
chaance, line 773: I'm actually not sure that we need this warning. I think a single module bundle is probably fine for decently large apps since it's easy to cache, and CSS files are rarely a bottleneck before some pretty serious scale. I think we can talk about the tradeoffs without making it look scary.

docs/guides/styling.md
```diff
@@ -766,6 +766,76 @@ Other CSS-in-JS libraries will have a similar setup. If you've got a CSS framewo
...
+/** @type {import('@remix-run/dev').AppConfig} */
+module.exports = {
+  future: {
+    unstable_cssModules: true,
+  },
+  // ...
+};
+```
+
+With this feature flag enabled, the Remix compiler will now generate a single CSS file containing all CSS Modules styles in your application. Note that any [regular stylesheet imports](http://localhost:3000/docs/en/v1/guides/styling#regular-stylesheets) will remain as separate files.
```
chaance, line 788:
```suggestion
With this feature flag enabled, the Remix compiler will now generate a single CSS file containing all CSS Modules styles in your application. Note that any [regular stylesheet imports](#regular-stylesheets) will remain as separate files.
```

docs/guides/styling.md
```diff
@@ -766,6 +766,76 @@ Other CSS-in-JS libraries will have a similar setup. If you've got a CSS framewo
...
+
+Unlike many other tools in the React ecosystem, insertion of this CSS bundle into the page is not handled for you automatically. Instead, we ensure that you always have control over the link tags on your page. This lets you decide where the CSS bundle is inserted and in what order relative to other style sheets.
```
chaance, line 790:
```suggestion
Unlike many other tools in the React ecosystem, we do not insert the CSS bundle into the page automatically. Instead, we ensure that you always have control over the link tags on your page. This lets you decide where the CSS file is loaded relative to other stylesheets in your app.
```

docs/guides/styling.md
```diff
@@ -766,6 +766,76 @@ Other CSS-in-JS libraries will have a similar setup. If you've got a CSS framewo
...
+
+To get access to the CSS bundle, first install the `@remix-run/css-bundle` package.
+
+```sh
+npm install @remix-run/css-bundle
+```
+
+Then, import `cssBundleHref` and add it to a link descriptor, most likely in `root.tsx` so that it applies to your entire application.
```
chaance, line 798:
```suggestion
Then, import `cssBundleHref` and add it to a link descriptor—most likely in `root.tsx` so that it applies to your entire application.
```

docs/guides/styling.md
```diff
@@ -766,6 +766,76 @@ Other CSS-in-JS libraries will have a similar setup. If you've got a CSS framewo
...
+
+export const links: LinksFunction = () => {
+  return [
+    { rel: "stylesheet", href: cssBundleHref },
+    // ...
+  ];
+};
+```
+
+You're all set! You can now opt into CSS Modules via the `.module.css` file name convention, for example:
```
chaance, line 812:
```suggestion
You're all set! You can now opt into CSS Modules via the `.module.css` file name convention. For example:
```

packages/remix-dev/compiler/compileBrowser.ts
```diff
@@ -123,29 +152,108 @@ export const createBrowserCompiler = (
...
+        ? esbuild.build({
+            ...createEsbuildConfig("app", remixConfig, options),
+            metafile: true,
+            incremental: true,
+          })
+        : appCompiler.rebuild());
+
+      invariant(
+        appCompiler.metafile,
+        "Expected app compiler metafile to be defined"
```
chaance, line 170:
```suggestion
        "Expected app compiler metafile to be defined. This is likely a bug in Remix. Please open an issue at https://github.com/remix-run/remix/issues/new"
```

packages/remix-dev/compiler/compileBrowser.ts
```diff
@@ -123,29 +152,108 @@ export const createBrowserCompiler = (
...
+            ...createEsbuildConfig("css", remixConfig, options),
+            metafile: true,
+            incremental: true,
+            write: false,
+          })
+        : cssCompiler.rebuild())) as esbuild.BuildIncremental;
+
+      invariant(
+        cssCompiler.metafile,
+        "Expected CSS compiler metafile to be defined"
```
chaance, line 192:
```suggestion
        "Expected CSS compiler metafile to be defined. This is likely a bug in Remix. Please open an issue at https://github.com/remix-run/remix/issues/new"
```

packages/remix-dev/compiler/compileBrowser.ts
```diff
@@ -123,29 +152,108 @@ export const createBrowserCompiler = (
...
+              // Grab the CSS bundle path so we can use it to generate the manifest
+              cssBundlePath = outputPath;
+              await fse.ensureDir(path.dirname(outputPath));
+
+              let contents =
+                options.mode === "production"
+                  ? await postcss([
+                      // We need to discard duplicate rules since "composes"
+                      // in CSS Modules can result in duplicate styles
+                      postcssDiscardDuplicates(),
```
chaance, line 217: Since we're balking on sourcemaps for the moment, is there any reason not to dedupe in development as well?

packages/remix-dev/compiler/compileBrowser.ts
```diff
@@ -123,29 +152,108 @@ export const createBrowserCompiler = (
...
+      cssBundlePath,
+    });
     manifestChannel.write(manifest);
     await writeAssetsManifest(remixConfig, manifest);
   };
+
   return {
     compile,
-    dispose: () => compiler?.rebuild.dispose(),
+    dispose: () => {
```
chaance, line 254: Can't remember if these are synchronous, so ... just checking that there aren't any potential race conditions here.

packages/remix-css-bundle/README.md
```diff
@@ -0,0 +1,9 @@
+# @remix-run/css-bundle
```
chaance, line 1: Could we maybe call this `@remix-run/css`?

751mnd

chaance (changes requested):
Looks great! Just a few minor suggestions and questions. Really stoked about the approach so far 👍

docs/guides/styling.md
cliffordfajardo: It does sound a little scary if someone were evaluating remix and saw CSS modules
Not discounting that the warning isn't true. Maybe there are strategies for reducing the bundle size like minifying the CSS.

just noticed out production CSS output isn't compressed in our app 😅

chaance: > Remix doesn't minify CSS by default?

No, Remix doesn't process CSS by default at all (aside than hashing the filename for effective versioning/caching). This is explained in the same doc.

That said, CSS Modules is opening the door a bit on this bigger question. We should absolutely be able to minify the production build of your CSS Modules bundle IMO.

packages/remix-css-bundle/README.md
mjackson: Nah, css-bundle is way more descriptive.

docs/guides/styling.md
markdalgleish: Whoooooops.

docs/guides/styling.md
markdalgleish: Can't believe I passed up an opportunity to use an em dash.

packages/remix-dev/compiler/compileBrowser.ts
markdalgleish: This was an existing error message—this PR just reformatted it—but I agree with your improvement 👍

packages/remix-dev/compiler/compileBrowser.ts
markdalgleish: I was thinking the same thing. I'll look into this.

packages/remix-dev/compiler/compileBrowser.ts
markdalgleish: This is a good question, I wasn't sure myself until double checking. It's sync, so we're good 👍

packages/remix-css-bundle/README.md
markdalgleish: Definitely looks cleaner, but I agree with @mjackson, I think it's a bit too vague.

packages/remix-dev/compiler/compileBrowser.ts
markdalgleish: I got this working. I had to rework the `compileBrowser` logic a bit so that we can read esbuild's CSS source map and then pass it to PostCSS. It actually cleaned up the code quite a bit since we're now writing the CSS bundle and its source map in a single step rather than mapping over all output files.

markdalgleish: To be clear, we're still skipping out on the broken CSS Modules source maps, but we're maintaining the CSS source map that we get from esbuild, otherwise every rule is said to come from `css-bundle-xxxxx.css` rather than the individual built CSS files. Still not as good as it could be, but much better than nothing.

docs/guides/styling.md
lili21: > I think a single module bundle is probably fine for decently large apps since it's easy to cache

But change any style will break the cache.

chaance: > But change any style will break the cache.

As it should! I'm not saying there are no potential perf tradeoffs here, just that we shouldn't overstate them because:

> CSS files are rarely a bottleneck before some pretty serious scale

markdalgleish: >We should absolutely be able to minify the production build of your CSS Modules bundle IMO.

The generated CSS is already being minified since it's being bundled by esbuild, honouring the same minification settings as the JS output.

xalechez:
Hey Mark! Great stuff here, glad to see CSS modules finally coming to Remix. My question here is about third-party libraries support. We import a component library that uses CSS modules via `require` syntax as it's transpiled to CJS. Is this also supported with your solution? In the past we had this issue with several esbuild CSS modules plugins and it was a big blocker for us with Remix

markdalgleish:
@xalechez Is the component library open source? I know you'd at least need to add the package to the `serverDependenciesToBundle` array, but other than that I'm not sure what the issue would be.

chaance (approved):

PR #4855, opened 2022-12-13
fix(remix-dev): resolve asset entry full path to support mono-repo import of styles

jacob-ebey:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/compiler/compileBrowser.ts
+++ b/packages/remix-dev/compiler/compileBrowser.ts
@@ -75,7 +75,7 @@ const createEsbuildConfig = (
     deprecatedRemixPackagePlugin(options.onWarning),
     cssFilePlugin({
       mode: options.mode,
-      rootDir: config.rootDirectory,
+      rootDirectory: config.rootDirectory,
     }),
     urlImportsPlugin(),
     mdxPlugin(config),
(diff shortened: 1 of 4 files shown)


mcansh (approved):

packages/remix-dev/compiler/plugins/cssFilePlugin.ts
```diff
@@ -72,9 +72,11 @@ export function cssFilePlugin(options: {
         let entry = Object.keys(outputs).find((out) => outputs[out].entryPoint);
         invariant(entry, "entry point not found");
 
-        let normalizedEntry = normalizePathSlashes(entry);
```
pcattori, line 75: If we want to guarantee that this is correct for all cases, I think its better to resolve the path relative to the project root directory specified in the Remix config. I think we can get that path by looking at `args.path` within this plugin.

jacob-ebey: Ended up just passing through the config value so there is no guessing / accidental miss-slice of a bad path.

pcattori (approved):

packages/remix-dev/compiler/compileBrowser.ts
```diff
@@ -73,7 +73,10 @@ const createEsbuildConfig = (
 
   let plugins: esbuild.Plugin[] = [
     deprecatedRemixPackagePlugin(options.onWarning),
-    cssFilePlugin(options),
+    cssFilePlugin({
+      mode: options.mode,
+      rootDir: config.rootDirectory,
```
machour, line 78: `rootDirectory`

jacob-ebey: Yeah, got distracted and pushed before I finished lolol

PR #4880, opened 2022-12-15
feat(remix-dev): flat routes

mcansh:
~~note: [this test](https://github.com/remix-run/remix/blob/434836ff7eb5afa7436fdabcee7e0d063691f6e1/packages/remix-dev/__tests__/routesConvention-test.ts#L66-L69) is currently failing, but i believe that the "expected" result is actually wrong and should actually be ` ":$dollabills?/.lol?/what?/$?/*"`, so i think we have a bug in `defineConventionalRoutes` where this test originated from.~~

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

tests and `yarn playground:new`


kiliman (commented):
Hmm.. the `.route` suffix is actually in the original spec. Is there a reason this is removed?

kiliman (commented):
Hmm.. a lot of the code there was to make it easier to keep the enhanced version in sync with the core one. I think the enhancements will be a good proving ground for including in future core. By stripping all that out, I think it's going to be more difficult to extend in the future.

pcattori:
> note: [this test](https://github.com/remix-run/remix/blob/434836ff7eb5afa7436fdabcee7e0d063691f6e1/packages/remix-dev/__tests__/routesConvention-test.ts#L66-L69) is currently failing, but i believe that the "expected" result is actually wrong and should actually be  ":$dollabills?/.lol?/what?/$?/*", so i think we have a bug in defineConventionalRoutes where this test originated from.

@mcansh I has the same confusion, but I think its [actually correct](https://github.com/remix-run/remix/pull/4706#discussion_r1041228519)

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,240 @@
...
+      }
+    }
+  }
+  // process remaining segment
+  pushRouteSegment(routeSegment);
+  return [routeSegments, separators];
+}
+
+function isPathSeparator(char: string | undefined) {
+  return char === "/" || char === ".";
```
pcattori, line 239: Does this need to account for windows path separator?

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,240 @@
...
+    pathname = pathname.replace(/_index$/, "");
+  }
+  if (pathname.startsWith("/")) {
+    pathname = pathname.substring(1);
+  }
+  return pathname || undefined;
+}
+
+function isIndexRoute(routeId: string) {
+  return routeId.endsWith("_index");
```
pcattori, line 110: We had [discussed using just `_` as the index route convention](https://github.com/remix-run/remix/discussions/4482#discussioncomment-4052865) and last time I spoke with @ryanflorence we agreed on it.

Could we change this to use `_` for index instead?
Note: this might mean we need more robust checks for pathless/layout routes since `endsWith('_')` would then incorrectly capture index routes.

mcansh: last i spoke with @ryanflorence (last wednesday) we decided `_index`, but happy to adjust if that's since changed

pcattori: Spoke to @ryanflorence today. Going with `_index`.

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,273 @@
...
+  let pushRouteSegment = (routeSegment: string) => {
+    if (routeSegment) {
+      routeSegments.push(routeSegment);
+    }
+  };
+  while (index < routeId.length) {
+    let char = routeId[index];
+    index++; // advance to next character
+    let nextChar = routeId[index + 1];
+    let prevChar = routeId[index - 1];
```
pcattori, line 145: Doesn't `index++` cause indices for `nextChar` and `prevChar` to be calculated incorrectly?

```suggestion
    let nextChar = routeId[index + 1];
    let prevChar = routeId[index - 1];
```

and then we'd need an `index += 1` (or some other increment to index) at the end of the `while` loop.

mcansh: > Doesn't index++ cause indices for nextChar and prevChar to be calculated incorrectly?

you'd think, but i just lifted these values up from their previous locations, where we did it inline - i'll tackle making it clearer though

pcattori: I think `about/.[.*/]` is actually _correct_ since the rule is "As soon as we encounter a `[`, the next `]` encounter is what closes that escape block. So `about.[.[.*].].tsx` is parsed as:
1.`about.`, escape(`.[.*`), `.]`
2. `about/`, literal(`.[.*`), `/]`
3. `about/.[.*/]`

I.e., ` [a[b]c]` is _not_ (a) escape(`a`, escape(`b`), `c`), but rather (b) escape(`a[b`), `]c]`

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,269 @@
...
+  if (parentPath.startsWith("/")) {
+    parentPath = parentPath.substring(1);
+  }
+  let routePath = getRouteSegments(routeId).join("/");
+  if (routePath.startsWith("/")) {
+    routePath = routePath.substring(1);
+  }
+  let pathname = parentPath
+    ? routePath.slice(parentPath.length + 1)
+    : routePath;
```
pcattori, line 109: Suggestion: could use [lodash's `escapeRegExp`](https://lodash.com/docs/4.17.15#escapeRegExp) to be more explicit:

```suggestion
  let pathname = routePath.replace(new RegExp("^" + _.escapeRegExp(parentPath)), '')
```

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,269 @@
...
+}
+
+function pathFromRouteId(routeId: string, parentId: string) {
+  let parentPath = "";
+  if (parentId) {
+    parentPath = getRouteSegments(parentId).join("/");
+  }
+  if (parentPath.startsWith("/")) {
+    parentPath = parentPath.substring(1);
+  }
```
pcattori, line 102: Suggestion:
```suggestion
  parentPath = parentPath.replace(/^\//, '')
```

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,269 @@
...
+    if (routes[routeId]) {
+      throw new Error(`Duplicate route: ${routeId}`);
+    }
+
+    let tmpId = "";
+    let parentId = "";
+    for (let processedId of processedIds) {
+      if (routeId.startsWith(processedId.replace(/\$$/, "*"))) {
+        let [segments, delimiters] = splitByDelimiter(routeId);
+        if (segments.at(1)?.endsWith("_")) {
```
pcattori, line 59: What if the `.` is escaped? E.g. `app.hello_[.]world`? Then this check would get incorrectly triggered, no?

pcattori: Also, don't we have to account for ending `_` for other segments? not just the second one (`at(1)`)?

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,295 @@
...
+  let nameMap = new Map<string, RouteInfo>();
+
+  for (let routePath of routePaths) {
+    let routeInfo = getRouteInfo(appDirectory, prefix, routePath);
+    routeMap.set(routeInfo.id, routeInfo);
+    nameMap.set(routeInfo.name, routeInfo);
+  }
+
+  // update parentIds for all routes
+  Array.from(routeMap.values()).forEach((routeInfo) => {
```
pcattori, line 57: You could use the native `for...of` loop for the `values()` iterator:

```suggestion
  for (let routeInfo of routeMap.values()) {
```

Kinda disappointing that iterators don't implement `forEach`...

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,295 @@
...
+    let routeInfo = getRouteInfo(appDirectory, prefix, routePath);
+    routeMap.set(routeInfo.id, routeInfo);
+    nameMap.set(routeInfo.name, routeInfo);
+  }
+
+  // update parentIds for all routes
+  for (let routeInfo of routeMap.values()) {
+    let parentId = findParentRouteId(routeInfo, nameMap);
+    routeInfo.parentId = parentId;
+  }
```
pcattori, line 60: might be nice to extract this block into a helper function to make it explicit that `nameMap` is not used later on and that `routeMap` is not expected to be mutated later on.

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,295 @@
...
+    defineRoute: DefineRouteFunction,
+    parentId?: string
+  ): void {
+    let childRoutes = Array.from(routeMap.values()).filter(
+      (routeInfo) => routeInfo.parentId === parentId
+    );
+    let parentRoute = parentId ? routeMap.get(parentId) : undefined;
+    let parentRoutePath = parentRoute?.path ?? "/";
+    for (let childRoute of childRoutes) {
+      let routePath = childRoute?.path?.slice(parentRoutePath.length) ?? "";
```
pcattori, line 74: I don't think `childRoute` can be `null` or `undefined`, right?

```suggestion
      let routePath = childRoute.path?.slice(parentRoutePath.length) ?? "";
```

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,295 @@
...
+      // remove leading slash
+      if (routePath.startsWith("/")) {
+        routePath = routePath.slice(1);
+      }
```
pcattori, line 78: alternative:

```suggestion
      routePath = routePath.replace(/^\//, "")
```

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,305 @@
...
+    }
+  }
+
+  let routes = defineRoutes(defineNestedRoutes);
+
+  Object.values(routes).forEach((route) => {
+    if (route.parentId === undefined) {
+      route.parentId = "root";
+    }
+  });
```
pcattori, line 115: I'm surprised that setting `parentId` to `"root"` for missing parents isn't already done by `defineRoutes`.

Is that something that should be added to `defineRoutes` so that this doesn't have to be patched here? Or is there a reason to want to keep this logic outside of `defineRoutes`?

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,303 @@
...
+      // remove leading slash
+      routePath = routePath.replace(/^\//, "");
+      let index = childRoute.index;
+      let fullPath = childRoute.path;
+
+      // add option to check for unique route ids
+      // this is copied from remix default convention
+      // but it is currently breaking some flat routes
+      // so until we can figure out a better way to do this
+      // make it optional to unblock users
```
pcattori, line 71: Is this still something that should be made optional? I don't see anywhere in this code where we make it optional.

If not, could you remove the outdated comment:

```suggestion
```

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,303 @@
...
+        let conflict = uniqueRoutes.get(uniqueRouteId);
+        if (conflict) {
+          throw new Error(
+            `Path ${JSON.stringify(fullPath)} defined by route ${JSON.stringify(
+              childRoute.id
+            )} conflicts with route ${JSON.stringify(conflict)}`
+          );
+        } else {
+          uniqueRoutes.set(uniqueRouteId, childRoute.id);
+        }
```
pcattori, line 83: nit: this block can be unnested since the `if` branch throws:

```suggestion
        uniqueRoutes.set(uniqueRouteId, childRoute.id);
```

packages/remix-dev/config/flat-routes.ts
mcansh: forgot to remove the comment 😬

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,292 @@
...
+  return routeId.endsWith("_index");
+}
+
+type SubState = "NORMAL" | "ESCAPE" | "OPTIONAL" | "OPTIONAL_ESCAPE";
+
+export function getRouteSegments(routeId: string) {
+  let routeSegments: string[] = [];
+  let index = 0;
+  let routeSegment = "";
+  let state: "START" | "PATH" = "START";
```
pcattori, line 116: Could you defined `type State` explicitly above `Substate`? Additionally, it would be nice to include a comment describing what the different states mean, maybe with an example. That would make it much easier to understand what the code is doing later on.

```suggestion
type State = "START" | "PATH"
type SubState = "NORMAL" | "ESCAPE" | "OPTIONAL" | "OPTIONAL_ESCAPE";

export function getRouteSegments(routeId: string) {
  let routeSegments: string[] = [];
  let index = 0;
  let routeSegment = "";
  let state: STATE = "START";
```

packages/remix-dev/config/flat-routes.ts
mcansh: code no longer exists in this form

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,291 @@
...
+  let subState: SubState = "NORMAL";
+  let pushRouteSegment = (routeSegment: string) => {
+    if (routeSegment) {
+      if (routeSegment.includes("/")) {
+        throw new Error(
+          `Route segment cannot contain a slash: ${routeSegment} (in route ${routeId})`
+        );
+      }
+      routeSegments.push(routeSegment);
+    }
```
pcattori, line 126: nit: you can unnest most of this with a precondition / early return:
```suggestion
    if (!routeSegment) return
    if (routeSegment.includes("/")) {
      throw new Error(
        `Route segment cannot contain a slash: ${routeSegment} (in route ${routeId})`
      );
    }
    routeSegments.push(routeSegment);
```

packages/remix-dev/config/flat-routes.ts
```diff
@@ -0,0 +1,291 @@
...
+      state = "PATH";
+      subState = "NORMAL";
+    }
+    if (state == "PATH") {
+      switch (subState) {
+        case "NORMAL": {
+          if (isSegmentSeparator(char)) {
+            state = "START";
+            break;
+          }
```
pcattori, line 159: The `"START"` state can be obsoleted by doing the mutations for `"STATE"` within the `isSegmentSeparator` block. Since that would only leave `"PATH"` as a valid `State`, that can also be removed:

```suggestion
      switch (subState) {
        case "NORMAL": {
          if (isSegmentSeparator(char)) {
            pushRouteSegment(routeSegment);
            routeSegment = "";
            subState = "NORMAL";
            break;
          }
```

Then `subState` and `Substate` should be renamed to `state` and `State` respectively.

Incidentally, this would also supercede the need for a comment explaining the states since the substate values are self-explanatory.

mcansh: i'll keep the comments, just for extra clarity :)

pcattori (approved):

packages/remix-dev/__tests__/flat-routes-test.ts
```diff
@@ -0,0 +1,607 @@
...
+      ],
+      // TODO: talk with @pcattori
+      // [
+      //   "routes/about.[*].tsx",
+      //   {
+      //     id: "routes/about.[*]",
+      //     parentId: "routes/about",
+      //     path: "*",
+      //   },
+      // ],
```
pcattori, line 349: Did you still want to chat about this one?

packages/remix-dev/__tests__/flat-routes-test.ts
```diff
@@ -604,4 +597,26 @@ describe("flatRoutes", () => {
...
+  test("should error when using `:` in an escaped route segment", () => {
+    let files: string[] = ["routes/about.[:name].tsx"];
+
+    expect(() =>
+      flatRoutesUniversal(
+        APP_DIR,
+        files.map((file) => path.join(APP_DIR, file))
+      )
+    ).toThrowError("Escaped route segment cannot contain `:`");
+  });
```
pcattori, line 621: I think we want to error out whenever `:` or `*` appear _anywhere_ in the route, not just being escaped.

Since `:` and `*` are not part of Remix's special route characters like `$` or `()`, they should be treated like literals. But since there's not way to convey `:` and `*` as literals for RR (escaped or not), they are footguns that we should prevent users from running into.

mcansh: ah right right

pcattori (approved):

manzano78:
Hey! I tried the nightly release which includes this PR and I think I found an issue, based on the following file structure:

```
routes/
  app/
      _index.tsx
      primary-nav.tsx
      footer.tsx
```
- `footer.tsx` exports a default component. 

What looks wrong to me is that I can see this footer directly on screen by hitting `/app/footer` in the URL bar.
But if I understand the principles, only `_index.tsx` should be treated as a route module, right?

Before you ask, I did activate the future flag :)

kiliman:
@manzano78 You should probably create a new issue instead of commenting on a closed PR.

Anyway, yes. If you are using *flat-folders* convention (where routes are folder names instead of filenames), then *_index.tsx* should be treated as the route and everything else as colocated files.

Can you run `npx remix routes` and paste the result? Please do so in another issue and reference this comment.

Thanks!

Issue #4872, opened 2022-12-15
Revalidation triggers fetcher.submit to change


kentcdodds:
### What version of Remix are you using?

1.8.2

### Steps to Reproduce

Open this: https://stackblitz.com/edit/node-ikzu57?file=app%2Froutes%2Findex.tsx

Open the browser console

Click "Set searchParams"

Observe the console shows that `fetcher.submit` has changed.

### Expected Behavior

I expect `fetcher.submit` to be stable so I can safely use it in dependency arrays.

### Actual Behavior

I cannot use `fetcher.submit` in a dependency array because any time there's a revalidation on the page, all fetcher.submits are new instances of the submit function.

kentcdodds:
After further investigation, what's trigger `fetcher.submit` to change is the `defaultAction` dependency: https://github.com/remix-run/remix/blob/8aacba3c4cd23e63ef57a8cb7b68c9d6c88bdb6e/packages/remix-react/components.tsx#L1260

So, what's happening, is I make a change to the URLSearchParams, that triggers the `defaultAction` to change which invalidates the `submit` on the fetcher which triggers my `useEffect` again.

In my case, I don't care about the `defaultAction` because I specify one when I call `submit`.

So, in order to ensure referential stability, we need to either:

1. allow the user to configure the `action` when calling `useFetcher` so we can ignore the `defaultAction` in that case (so it won't appear in the dependency array and therefore cannot change)
2. Or we need to use a "latest ref" pattern so we don't have to list the `defaultAction` in the dependency array.

I much prefer option 1 because it would actually be a really nice feature to have anyway. It would allow folks to change their code from something like this:

```tsx
const fetcher = useFetcher()

// ...
<button onClick={() => fetcher.submit({hi: 'bob'}, {method: 'post', action: '/some-path'})}>Submit bob</button>
```

To this:


```tsx
const fetcher = useFetcher({
  method: 'post',
  action: '/some-path',
})

// ...
<button onClick={() => fetcher.submit({hi: 'bob'})}>Submit bob</button>
```


I've seen some folks suggest they would actually prefer this API and I think I agree with them. Also, it would solve the problem of fetcher being referentially stable.

brophdawg11:
This should be resolved by https://github.com/remix-run/react-router/pull/10336 which makes `fetcher.submit` stable across `location` changes

brophdawg11:
Should be available in Remix 1.16

PR #4882, opened 2022-12-15
feat(remix-react): add `submitOptions` argument for `useSubmit` and `useFetcher`

kentcdodds:
This fixes an actual bug I was having when trying to use a fetcher in a `useEffect`. It's also a handy API that I've seen people ask for as well.

Closes: #4872

- [x] Docs
- [x] Tests

Testing Strategy: Automated Integration test

~~This is just pushing the failing test to demonstrate the issue.~~

~~The issue is actually technically correct behavior, but I need a new feature to accomplish what I want so I'll be changing the test once this new API is supported. I'm working on that next.~~

Implementation is finished and documented. Ready to discuss/merge.

Until this is merged, the only way to reliably use a fetcher.submit in a dep array is to do something like this:

```tsx
const fetcher = useFetcher()
const submitRef = useRef(fetcher.submit)
useEffect(() => {
  submitRef.current = fetcher.submit
})
React.useEffect(() => {
  // can safely use submitRef without including in dep array
  submitRef({other, deps}, {action: '/some-path', method: 'post'})
}, [other, deps])
```

The drawback here is that if we actually *do* depend on the `defaultAction` then it's possible we'll get the wrong one when submit is called.

This PR fixes that by allowing users to specify an `action` (among other submit options):

```tsx
const fetcher = useFetcher({action: '/some-path', method: 'post'})
React.useEffect(() => {
  // can safely include fetcher in dep array
  fetcher.submit({other, deps})
}, [other, deps, fetcher])
```

### DIFF
--- a/.changeset/metal-bikes-notice.md
+++ b/.changeset/metal-bikes-notice.md
@@ -3,4 +3,4 @@
 "@remix-run/react": minor
 ---
 
-feat(remix-react): add `submitOptions` argument for `useSubmit` and `useFetcher`
+Add `submitOptions` argument for `useSubmit` and `useFetcher`
--- a/.changeset/metal-bikes-notice.md
+++ b/.changeset/metal-bikes-notice.md
@@ -1,0 +1,6 @@
+---
+"remix": minor
+"@remix-run/react": minor
+---
+
+feat(remix-react): add `submitOptions` argument for `useSubmit` and `useFetcher`
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1264,7 +1264,7 @@ export function useSubmit(options?: SubmitOptions): SubmitFunction {
 const DEFAULT_METHOD = "get";
 const DEFAULT_ENC_TYPE = "application/x-www-form-urlencoded";
 
-// IDEA: remove the export here?
+// TODO: remove the export here?
 export function useSubmitImpl(
   key?: string,
   {
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1271,6 +1271,7 @@ export function useSubmitImpl(
     method: submitMethod = DEFAULT_METHOD,
     action: submitAction,
     encType: submitEncType = DEFAULT_ENC_TYPE,
+    replace,
   }: SubmitOptions = {}
 ): SubmitFunction {
   let navigate = useNavigate();
@@ -1420,7 +1421,10 @@ export function useSubmitImpl(
         });
       } else {
         setNextNavigationSubmission(submission);
-        navigate(url.pathname + url.search, { replace: options.replace });
+        navigate(url.pathname + url.search, {
+          replace:
+            typeof options.replace === "undefined" ? replace : options.replace,
+        });
       }
     },
     [
@@ -1429,6 +1433,7 @@ export function useSubmitImpl(
       defaultMethod,
       key,
       navigate,
+      replace,
       transitionManager,
     ]
   );
@@ -1598,7 +1603,8 @@ export type FetcherWithComponents<TData> = Fetcher<TData> & {
  * @see https://remix.run/api/remix#usefetcher
  */
 export function useFetcher<TData = any>(
-  submitOptions?: SubmitOptions
+  // replace doesn't make sense for fetchers
+  submitOptions?: Omit<SubmitOptions, "replace">
 ): FetcherWithComponents<SerializeFrom<TData>> {
   let { transitionManager } = useRemixEntryContext();
 
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1256,16 +1256,28 @@ export interface SubmitFunction {
  *
  * @see https://remix.run/api/remix#usesubmit
  */
-export function useSubmit(): SubmitFunction {
-  return useSubmitImpl();
+export function useSubmit(options?: SubmitOptions): SubmitFunction {
+  let key = undefined; // no key for global submit
+  return useSubmitImpl(key, options);
 }
 
-let defaultMethod = "get";
-let defaultEncType = "application/x-www-form-urlencoded";
+const DEFAULT_METHOD = "get";
+const DEFAULT_ENC_TYPE = "application/x-www-form-urlencoded";
 
-export function useSubmitImpl(key?: string): SubmitFunction {
+// IDEA: remove the export here?
+export function useSubmitImpl(
+  key?: string,
+  {
+    method: submitMethod = DEFAULT_METHOD,
+    action: submitAction,
+    encType: submitEncType = DEFAULT_ENC_TYPE,
+  }: SubmitOptions = {}
+): SubmitFunction {
   let navigate = useNavigate();
-  let defaultAction = useFormAction();
+  let defaultFormAction = useFormAction();
+  let defaultAction = submitAction || defaultFormAction;
+  let defaultMethod = submitMethod || DEFAULT_METHOD;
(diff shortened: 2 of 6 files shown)


packages/remix-react/components.tsx
```diff
@@ -1248,16 +1248,29 @@ export interface SubmitFunction {
...
 
-let defaultMethod = "get";
-let defaultEncType = "application/x-www-form-urlencoded";
-
-export function useSubmitImpl(key?: string): SubmitFunction {
+const DEFAULT_METHOD = "get";
+const DEFAULT_ENC_TYPE = "application/x-www-form-urlencoded";
+
+// IDEA: remove the export here?
+export function useSubmitImpl(
```
kentcdodds, line 1260: I'm not sure why this is exported, so I added a comment to remind me to ask.

packages/remix-react/components.tsx
```diff
@@ -1248,16 +1248,29 @@ export interface SubmitFunction {
  *
  * @see https://remix.run/api/remix#usesubmit
  */
-export function useSubmit(): SubmitFunction {
-  return useSubmitImpl();
+export function useSubmit(options?: SubmitOptions): SubmitFunction {
+  let key = undefined; // no key for global submit
```
kentcdodds, line 1252: I thought about making `key` a part of the options object in `useSubmitImpl` instead of accepting it as a first param, but decided to not cause a breaking change to `useSubmitImpl` because it's exported and it's possible people are using it. I actually don't hate this though so maybe it's fine?

machour:
Remember to target the `dev` branch 🙏

packages/remix-react/components.tsx
```diff
@@ -1248,16 +1248,29 @@ export interface SubmitFunction {
...
+  return useSubmitImpl(key, options);
 }
 
-let defaultMethod = "get";
-let defaultEncType = "application/x-www-form-urlencoded";
-
-export function useSubmitImpl(key?: string): SubmitFunction {
+const DEFAULT_METHOD = "get";
+const DEFAULT_ENC_TYPE = "application/x-www-form-urlencoded";
```
kentcdodds, line 1257: Switched these to `const` and upper-case because otherwise typescript wasn't happy about default-assigning `submitMethod` to `DEFAULT_METHOD` below. Using `const` here allows typescript to narrow this down from a type of `string` to `"get"` which fits in the `FormMethod` type.

packages/remix-react/components.tsx
```diff
@@ -1248,16 +1248,29 @@ export interface SubmitFunction {
...
+    encType: submitEncType = DEFAULT_ENC_TYPE,
+    replace,
+  }: SubmitOptions = {}
+): SubmitFunction {
   let navigate = useNavigate();
-  let defaultAction = useFormAction();
+  let defaultFormAction = useFormAction();
+  let defaultAction = submitAction || defaultFormAction;
+  let defaultMethod = submitMethod || DEFAULT_METHOD;
+  let defaultEncType = submitEncType || DEFAULT_ENC_TYPE;
```
kentcdodds, line 1273: By putting these calculations all here we can avoid having to pass `defaultFormAction` into the dependency array which is what fixes my original issue.

packages/remix-react/components.tsx
```diff
@@ -1400,10 +1413,21 @@ export function useSubmitImpl(key?: string): SubmitFunction {
         });
       } else {
         setNextNavigationSubmission(submission);
-        navigate(url.pathname + url.search, { replace: options.replace });
+        navigate(url.pathname + url.search, {
+          replace:
+            typeof options.replace === "undefined" ? replace : options.replace,
```
kentcdodds, line 1418: I can't remember for sure, but I believe we can't use `??` in this codebase. If we can, feel free to do this:

```suggestion
            options.replace ?? replace,
```

Can't use `||` like the others do because we're dealing with booleans here.

kentcdodds:
Sorry, totally forgot to target `dev`. Thanks for the tip @machour. I've fixed it up.

packages/remix-react/components.tsx
MichaelDeBoey: I think we can safely remove the export, as it was never intended to be a public API

packages/remix-react/components.tsx
MichaelDeBoey: Since I think we can safely remove the export (as it was never intended to be a public API), I think we can make `key` part of the `submitOptions`

packages/remix-react/components.tsx
MichaelDeBoey: According to node.green, the nullish coalescing operator (`??`) is available as of v14.0, so we can safely use it in our codebase.
https://node.green/#ES2020-features--nullish-coalescing-operator-----

machour: If this ends up in the client, then we shouldn't use `??` for now, until we add [support for .browserslistrc](https://github.com/remix-run/remix/discussions/4769).

MichaelDeBoey: `useFetcher` & `useSubmit` indeed end up in the client, so it's indeed better to wait for that proposal instead 👍

packages/remix-react/components.tsx
machour: But it was public, so we would be breaking BC in a minor release. Let's keep it exported, and remove it in v2.

MichaelDeBoey: We released things that were technically breaking in the past (mostly on `@remix-run/server-runtime`'s side) when they weren't meant to be public API, so I don't think we should care here either

packages/remix-react/components.tsx
kentcdodds: I actually kinda like that `submitOptions` is a separate argument, even though we have this funny `key = undefined` thing. But I don't care enough.

packages/remix-react/components.tsx
MichaelDeBoey: Maybe a good idea to add a TODO for that?

packages/remix-react/components.tsx
```diff
@@ -1256,16 +1256,29 @@ export interface SubmitFunction {
...
 }
 
-let defaultMethod = "get";
-let defaultEncType = "application/x-www-form-urlencoded";
-
-export function useSubmitImpl(key?: string): SubmitFunction {
+const DEFAULT_METHOD = "get";
+const DEFAULT_ENC_TYPE = "application/x-www-form-urlencoded";
+
+// IDEA: remove the export here?
```
kentcdodds, line 1267:
```suggestion
// TODO: remove the export here?
```

packages/remix-react/components.tsx
kentcdodds: Done 👍

.changeset/metal-bikes-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+"remix": minor
+"@remix-run/react": minor
+---
+
+feat(remix-react): add `submitOptions` argument for `useSubmit` and `useFetcher`
```
MichaelDeBoey, line 6: We don't add the first bit in changesets

```suggestion
Add `submitOptions` argument for `useSubmit` and `useFetcher`
```

brophdawg11:
We're about to wipe this code out when we merge the 6.4 work 😬  - so if we're going to adopt this we probably need to do it in `react-router-dom`.  Could we open a proposal for this API enhancement?  Get enough upvotes and it could be accepted as soon as Monday!

kentcdodds:
Yep. I expected as much. Didn't really think this would get merged as-is. Here's the discussion over there: https://github.com/remix-run/react-router/discussions/9737

machour:
Closing this as it will be implemented in RR

PR #4911, opened 2022-12-19
chore: disable prefer-const rule for deno

mcansh:
https://github.com/remix-run/remix/actions/runs/3716871751/jobs/6303686481

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey:
#4913 is the proper fix I think

mcansh:
> #4913 is the proper fix I think

depends, we use let for non user facing code generally

MichaelDeBoey:
True, but Deno doesn't have a rule to enforce that as far as I know
Also: this won't enforce `const` in the deno template

PR #4912, opened 2022-12-19
Fix flakey firefox tests

brophdawg11:
Fix some flakey tests in the `feat/rrr-rendering` branch before merging to `dev`

### DIFF
--- a/integration/error-boundary-test.ts
+++ b/integration/error-boundary-test.ts
@@ -1,4 +1,5 @@
 import { test, expect } from "@playwright/test";
+import { ServerMode } from "@remix-run/server-runtime/mode";
 
 import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
 import type { Fixture, AppFixture } from "./helpers/create-fixture";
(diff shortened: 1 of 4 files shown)


integration/error-boundary-test.ts
```diff
@@ -636,3 +637,241 @@ test.describe("ErrorBoundary", () => {
...
+      expect(consoleErrors).toEqual([]);
+
+      await app.clickSubmitButton("/parent/child-without-boundary");
+      await page.waitForSelector("#parent-error");
+
+      expect(await app.getHtml("#parent-error")).toEqual(
+        '<p id="parent-error">Broken!</p>'
+      );
+      expect(await app.getHtml("#parent-matches-data")).toEqual(
+        '<p id="parent-matches-data"></p>'
```
brophdawg11, line 858: FWIW I think this is a bug in TM today that is fixed in the new router.  When JS is enabled, TM doesn't clear out the prior parent `routeData` so we still get a now-stale `PARENT` here even though the loader was not re-run

PR #4914, opened 2022-12-19
refactor: use type imports instead of type modifiers

MichaelDeBoey:
This way we use type imports consistently throughout the codebase


mcansh:
tbh didnt know `import { type X }` was valid in ts lol

MichaelDeBoey:
@mcansh Type modifiers are valid since TS v4.5, but the whole codebase uses type imports, so wanted to make it consistent
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names

PR #4899, opened 2022-12-16
chore: rename non public exports

mcansh:
exposed a few things that were needed for `@remix-run/testing` but shouldn't be public API #4539

Signed-off-by: Logan McAnsh <logan@mcan.sh>

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

PR #4915, opened 2022-12-19
chore(remix-testing): utilize react-router 6.5+

mcansh:
supersedes #4899 

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy: https://github.com/mcansh/stubbbbbs

### DIFF
--- a/packages/remix-testing/create-remix-stub.tsx
+++ b/packages/remix-testing/create-remix-stub.tsx
@@ -1,6 +1,14 @@
 import * as React from "react";
-import type { UNSAFE_RouteData as RouteData } from "@remix-run/react";
-import type { InitialEntry, Router } from "@remix-run/router";
+import type { HydrationState, InitialEntry, Router } from "@remix-run/router";
+import { UNSAFE_RemixContext as RemixContext } from "@remix-run/react";
+import type {
+  UNSAFE_FutureConfig as FutureConfig,
+  UNSAFE_AssetsManifest as AssetsManifest,
+  UNSAFE_EntryRoute as EntryRoute,
+  UNSAFE_RouteManifest as RouteManifest,
+  UNSAFE_RouteModules as RouteModules,
+  UNSAFE_RemixContextObject as RemixContextObject,
+} from "@remix-run/react";
 import type { RouteObject } from "react-router-dom";
 import { createMemoryRouter, RouterProvider } from "react-router-dom";
 
@@ -14,16 +22,13 @@ type RemixStubOptions = {
   initialEntries?: InitialEntry[];
 
   /**
-   *  Used to set the route's initial loader data.
-   *  e.g. initialLoaderData={{ "/contact": { locale: "en-US" } }}
+   *  Used to set the route's initial loader and action data.
+   *  e.g. hydrationData={{
+   *   loaderData: { "/contact": { locale: "en-US" } },
+   *   actionData: { "/login": { errors: { email: "invalid email" } }}
+   *  }}
    */
-  initialLoaderData?: RouteData;
-
-  /**
-   *  Used to set the route's initial action data.
-   *  e.g. initialActionData={{ "/login": { errors: { email: "invalid email" } }}
-   */
-  initialActionData?: RouteData;
+  hydrationData?: HydrationState;
 
   /**
    * The initial index in the history stack to render. This allows you to start a test at a specific entry.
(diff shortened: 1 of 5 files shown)


packages/remix-testing/create-remix-stub.tsx
```diff
@@ -62,110 +67,63 @@ type RemixStubOptions = {
...
 type RemixConfigFuture = Partial<EntryContext["future"]>;
 
 export function createRemixStub(
-  routes: StubRouteObject[],
+  routes: RouteObject[],
   remixConfigFuture?: RemixConfigFuture
 ) {
   // Setup request handler to handle requests to the mock routes
-  let { dataRoutes, queryRoute } = createStaticHandler(routes);
+  let staticHandler = createStaticHandler(routes);
```
brophdawg11, line 77: Hoping this can go away complete and we can just `createMemoryRouter` here

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -62,110 +67,63 @@ type RemixStubOptions = {
...
-        context={remixContext}
-        action={state.action}
-        location={state.location}
-        navigator={history}
-      />
+      <RemixContext.Provider value={{ manifest, routeModules, future }}>
+        <StaticRouterProvider
+          router={memoryRouter}
+          context={staticHandlerContext}
+        />
```
brophdawg11, line 121: Should be able to switch this over to `RouterProvider` with the memory router

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -46,12 +35,28 @@ type RemixStubOptions = {
    */
   initialLoaderData?: RouteData;
 
+  /**
+   * Used to set the route's initial loader headers.
+   * e.g. initialLoaderHeaders={{ "/contact": { "Content-Type": "application/json" } }}
+   */
+  initialLoaderHeaders?: Record<string, Headers>;
```
brophdawg11, line 42: Hopefully these new input opts can go away too if we ditch the static handler

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -62,245 +35,25 @@ type RemixStubOptions = {
...
-  // Setup request handler to handle requests to the mock routes
-  let { dataRoutes, queryRoute } = createStaticHandler(routes);
+export function createRemixStub(routes: RouteObject[]) {
   return function RemixStub({
     initialEntries,
-    initialLoaderData = {},
-    initialActionData,
     initialIndex,
+    initialActionData,
+    initialLoaderData,
```
brophdawg11, line 43: Should we just accept `hydrationData` here?

MichaelDeBoey (approved):

packages/remix-testing/create-remix-stub.tsx
mcansh: i was thinking that too, im indifferent on it unless you have a usecase where youd test "errors" with it

i was also thinking of omitting some things on the route object (id, caseSensitive, etc)

brophdawg11: actually yeah if we do that we avoid the need to bring in `RouteData`(and get rid of the `UNSAFE` export) and can use `HydrationState` from the router instead

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -62,245 +35,25 @@ type RemixStubOptions = {
...
-    index: route.index,
-    caseSensitive: route.caseSensitive,
-    path: route.path,
-    parentId,
-    hasAction: !!route.action,
-    hasLoader: !!route.loader,
-    module: "",
-    hasCatchBoundary: false,
-    hasErrorBoundary: false,
+    return <RouterProvider router={routerRef.current} />;
```
brophdawg11, line 57: I think I was expecting we'd still have a `RemixContext.Provider` here - I think at the moment if you test a component like `<Link prefetch="render">` it'll break?  We should be able to create those from `router.routes` I think - similar to how we did it before

mcansh: oop, yep you're right

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -33,27 +38,112 @@ type RemixStubOptions = {
...
    */
   initialIndex?: number;
+
+  remixConfigFuture?: Partial<FutureConfig>;
 };
 
-export function createRemixStub(routes: RouteObject[]) {
+type StubRouteObject = Omit<
+  RouteObject,
+  "id" | "handle" | "shouldRevalidate" | "errorElement" | "hasErrorBoundary"
```
brophdawg11, line 47: I could see some of these being useful for more advanced tests/stories.  There's no reason you couldn't force an error to bubble to an `errorElement` on your route, or access something via `useMatches` with `id`/`handle`, or opt out of a `fetcher.load` revalidation etc.  I think if we `Omit`, we'll just get folks asking us to un-omit eventually (or they'll just `@ts-expect-error` and do it anyway 😂 ).  

`hasErrorBoundary` could probably be omitted (but I don't think it needs to be) since it'll get added in `createMemoryRouter`

brophdawg11 (approved):
I think we can delete `RemixEntry` now too :)

mcansh:
> I think we can delete `RemixEntry` now too :)

removed that one here https://github.com/remix-run/remix/commit/2bc3d0cdd1981e82e0a832826f8dc300bfaa6250

PR #4916, opened 2022-12-19
The new dev server

pcattori:
## Approach

Running the new dev server does the following:
1. Builds an initial build
2. Watches for changes to app source code and rebuilds when changes occur
3. Waits for the app server (i.e. `server.js`) to be ready
4. Sends a `RELOAD` message over the websocket

For step (3), the dev server polls `GET <app server>/__REMIX_ASSETS_MANIFEST` and compares the build hash received from the app server with the latest build hash produced by the compiler run within the dev server.

### 1 No more require cache purge

App server is now only restarted when a rebuild happens, not on every request.
That means that in-memory persistence is fixed. So no need to use the `global` trick anymore when dealing with DB connections, in-memory caches, etc..

In fact app server might not need to restart _at all_ if the app server supports some sort of server-side HMR.

### 2 Supports _any_ app server

Previously, dev mode only worked for `@remix-run/serve`, [not for custom servers](https://github.com/remix-run/remix/blob/main/packages/remix-dev/devServer/serve.ts#L34) like `server.{js,ts}`.
Now _any_ custom server is compatible with the new dev server.

### 3 Support ESM mode for Node server?

Previously, we needed fine-grain control of the require cache, which prevented Remix servers from being run in ESM mode.
Since the new dev server does not need to purge the require cache at all, this limitation has been removed.

I _think_ this means we could support ESM-mode Remix servers now if we wanted to.

## Future work

- Stop compiling/bundling custom app servers into `build/index.js`
- Deprecate / remove the virtual module for the server build. Just import from `./build/index.js` manually.
- Give back control to users for how they want to compile/run their server. E.g. if using TS, just run `ts-node ./server.js`

## TODO

- [x] future flag so that `remix dev` uses the new dev server
- [x] add configuration for Remix request handler path (so we can calculate full path for `__REMIX_ASSETS_MANIFEST`)
- [x] add configuration for poll interval and max retries
- [x] Tune the default values for poll interval and max retries/timeout
- [x] Use `http-proxy` directly (not `http-proxy-middleware`)
- [x] Inject websocket port into server build in dev
- [ ] CLI flag for app server port: `remix dev2 --app-server-port=3000`
- [ ] default websocket port preference in v2_dev config
- [ ] change future flag to `unstable_dev`
- [ ] Changeset
- [ ] dev server should hang (not crash) when app server is down. with a warning for timeout
---

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -239,6 +239,6 @@ export async function codemod(
 export async function dev2(projectDir?: string) {
   projectDir ??= process.cwd();
   let config = await readConfig(projectDir);
-  let dispose = await devServer2.serve(config, 3000);
-  await new Promise(() => {})
+  await devServer2.serve(config);
+  await new Promise(() => {});
 }
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -8,6 +8,7 @@ import * as esbuild from "esbuild";
 import * as colors from "../colors";
 import * as compiler from "../compiler";
 import * as devServer from "../devServer";
+import * as devServer2 from "../devServer2";
 import type { RemixConfig } from "../config";
 import { readConfig } from "../config";
 import { formatRoutes, RoutesFormat, isRoutesFormat } from "../config/format";
@@ -233,4 +234,11 @@ export async function codemod(
     }
     throw error;
   }
+}
+
+export async function dev2(projectDir?: string) {
+  projectDir ??= process.cwd();
+  let config = await readConfig(projectDir);
+  let dispose = await devServer2.serve(config, 3000);
+  await new Promise(() => {})
 }
--- a/packages/remix-dev/package.json
+++ b/packages/remix-dev/package.json
@@ -20,9 +20,9 @@
   "dependencies": {
     "@babel/core": "^7.18.6",
     "@babel/generator": "^7.18.6",
+    "@babel/parser": "^7.18.6",
     "@babel/plugin-syntax-jsx": "^7.18.6",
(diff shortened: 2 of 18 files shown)


cliffordfajardo:
> how to detect when app server is ready? look for specific pattern in stdout? wait for proxy port to be taken?
@pcattori - this implementation might be relevant: https://github.com/sindresorhus/wait-for-localhost

packages/remix-dev/package.json
```diff
@@ -52,6 +53,7 @@
     "lru-cache": "^7.14.1",
     "minimatch": "^3.0.4",
     "node-fetch": "^2.6.7",
+    "node-http-proxy": "^0.2.4",
```
pcattori, line 56: TODO remove this

cliffordfajardo:
Probably not relevant for this PR & more of a heads up 
ESBuild releases a few features & breaking API changes in 0.17.0
- release notes: https://github.com/evanw/esbuild/releases/

packages/remix-dev/cli/commands.ts
```diff
@@ -234,3 +235,10 @@ export async function codemod(
     throw error;
   }
 }
+
+export async function dev2(projectDir?: string) {
+  projectDir ??= process.cwd();
+  let config = await readConfig(projectDir);
+  await devServer2.serve(config);
+  await new Promise(() => {});
+}
```
pcattori, line 244: alternatively, we could detect `v2_dev` in future flags and change the behavior of `remix dev`.

pcattori:
Superceded by https://github.com/remix-run/remix/pull/5133

PR #4945, opened 2022-12-22
optional segments changes should trigger loaders

akamfoad:
I'm experiencing that clicking a Link/Navlink that points to a change in optional segments wont trigger the loaders to reload, even though the browser URL changes.

I wanted to open an issue but looks like you folks prefer a failing test, which is an amazing idea btw.

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -47,29 +47,92 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/index.jsx": js`
-        import { json } from "@remix-run/node";
-        import { useLoaderData, Link } from "@remix-run/react";
+      "app/root.jsx": js`
+      import { redirect } from "@remix-run/node";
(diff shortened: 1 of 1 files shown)


brophdawg11:
This doesn't look like a bug to me - the root loader is correctly not being re-called since it's a re-used loader and the `:lang` parameter is deeper in the tree (i.e., no parameters at the root route actually changed).  In this setup I would add a `($lang).tsx` layout file and do the language loader loading in there -- that loader will be recalled when switching languages.

Another option would be to leverage [`unstable_shouldReload`](https://remix.run/docs/en/v1/route/should-reload) in `1.9.x` or the upcoming [`shouldRevalidate`](https://reactrouter.com/en/main/route/should-revalidate) in `1.10.0` to opt-into re-running the root loader on `params.lang` changes.

akamfoad:
Thanks @brophdawg11 I'm not sure about the shouldReload/shouldRevalidate since the root never called either, I have to move the inital content in root further down, which is fine but I don't like it much, I'd rather reloadDocument the NavLink instead, but that was helpful, thanks for clarifying.

PR #4986, opened 2023-01-01
Bug report: weird behavior for `<Link>` without `reloadDocument` that matches both a dynamic route and a public/ asset

TomerAberbach:


brophdawg11:
I don't think this is a bug - it's sort of a fight between Express and Remix for the route `/asset.txt` since express is configured to serve matching filenames form `public` (via `app.use(express.static("public", { maxAge: "1h" }));`) and Remix will match any root-level dynamic parameter via `/:postId`.

What's happening here is that on client side navigation, Remix makes a `fetch('/asset.txt?_data=routes%2F%24postId')` call.  This gets handled by the express static middleware and never makes it to the Remix handler, so the "loader data" that remix gets back is the file itself.

I would recommend updating express to serve these files from a `/static/asset.txt` root path so they don't conflict with `/:postId`:

```js
app.use("/static", express.static("public", { maxAge: "1h" }));
```

PR #4998, opened 2023-01-04
templates: don't use `window` in `entry.client.tsx`

MichaelDeBoey:
This fixes the problem we have on `main` in `templates/deno` + it also complies with our gotcha

- https://github.com/remix-run/remix/actions/runs/3834895046/jobs/6527706087#step:8:11
- https://lint.deno.land/#no-window-prefix
- https://remix.run/docs/en/v1/pages/gotchas#typeof-window-checks

---

Tagging @kentcdodds as he implemented this initially in our stacks, so maybe there was a reason to call these methods on `window` instead of using the global ones

### DIFF
--- a/integration/helpers/node-template/app/entry.client.tsx
+++ b/integration/helpers/node-template/app/entry.client.tsx
@@ -13,10 +13,10 @@ function hydrate() {
   });
 }
 
-if (window.requestIdleCallback) {
-  window.requestIdleCallback(hydrate);
+if (requestIdleCallback) {
+  requestIdleCallback(hydrate);
 } else {
   // Safari doesn't support requestIdleCallback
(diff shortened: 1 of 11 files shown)


machour (approved):

machour:
Fine by me, but let's wait for other approvals 🙏🏼

kentcdodds:
No important reason

mcansh (approved):

mcansh:
looks like `if (requestIdleCallback)` is passing in safari and then erroring...

```log
Unhandled Promise Rejection: ReferenceError: Can't find variable: requestIdleCallback
(anonymous function)
rejectPromise
promiseReactionJob
```

perhaps we should do `typeof requestIdleCallback === 'function'`?

PR #5001, opened 2023-01-04
Add integration test for 404 loader behavior

brophdawg11:
Added integration tests around 404 + root loader behavior to ensure 1.10 matches 1.7.6 behavior.

To run these new tests against `1.7.6` locally:

```
git checkout remix@1.7.6
git checkout -b temp
git cherry-pick 2961255c7
yarn
npx playwright install
yarn test:integration catch-boundary
```


brophdawg11:
Don't merge this yet - I want to double check on the client side flow in the new router.  I'm a bit surprised the loader data is not clearing out there 🤔

PR #5004, opened 2023-01-04
feat(remix-dev): `upgrade` command

mcansh:
closes #4996

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -256,7 +256,8 @@ export async function upgrade({
     upgrade: true,
     packageManager,
     filter(name) {
-      return name.startsWith("@remix-run/") || name === "remix" && name !== "@remix-run/router";
+      if (name === "@remix-run/router") return false;
+      return name.startsWith("@remix-run/") || name === "remix";
     },
     removeRange: tag !== "latest",
   });
--- a/packages/remix-dev/cli/commands.ts
+++ b/packages/remix-dev/cli/commands.ts
@@ -4,7 +4,7 @@ import * as fse from "fs-extra";
 import ora from "ora";
 import prettyMs from "pretty-ms";
 import * as esbuild from "esbuild";
-import ncu from "npm-check-updates";
+import execa from "execa";
 
 import * as colors from "../colors";
 import * as compiler from "../compiler";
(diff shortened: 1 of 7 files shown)


packages/remix-dev/cli/commands.ts
```diff
@@ -234,3 +236,43 @@ export async function codemod(
...
+
+  let check = await ncu({
+    cwd: projectDir,
+    target() {
+      return tag;
+    },
+    upgrade: true,
+    packageManager,
+    filter(name) {
+      return name.startsWith("@remix-run/") || name === "remix";
```
brophdawg11, line 259: Love this new command!  I have a local script I use for this.  One thing I ran into was excluding the router in case that's a dependency since it's not synced with the remix releases.  Here's the filter in my script:

```
        (k.startsWith("@remix-run/") || k === "remix") &&
        k !== "@remix-run/router"

```

packages/remix-dev/cli/run.ts
```diff
@@ -456,32 +466,55 @@ export async function run(argv: string[] = process.argv.slice(2)) {
...
       await commands.codemod(input[1], input[2]);
       break;
     }
-    case "dev":
+    case "upgrade": {
+      let packageManager = getPreferredPackageManager();
+
+      if (flags.prerelease && flags.tag) {
+        throw new Error(
+          `Cannot set both --tag and --prerelease. Use --tag pre to get the latest prereleae`
```
MichaelDeBoey, line 502:
```suggestion
          `Cannot set both --tag and --prerelease. Use --tag pre to get the latest prerelease`
```

MichaelDeBoey (approved):

redonkulus:
@mcansh How could something like this be used with dependabot to only create one PR to update all remix deps instead of N number of PR's?

mcansh:
> @mcansh How could something like this be used with dependabot to only create one PR to update all remix deps instead of N number of PR's?

as far as i know, dependabot will issue a new PR for each package and doesn't support any sort of "grouping", i wish it did 😅

this new command would primarily be for easier manual upgrades, but you could also create a workflow to run every x to run the command in your repo

MichaelDeBoey:
@redonkulus @mcansh I've seen @dependabot update both `react` & `react-dom` in 1 PR
Don't know how it was configured or so though 🤔

mcansh:
> 

looks like they may be close https://github.com/dependabot/dependabot-core/issues/1190#issuecomment-1320301083

machour (approved):

brophdawg11:
Going to close this out for now to keep the CLI surface area small - `npx upgrade-remix` is available to folks who want a quick one-liner

PR #5012, opened 2023-01-05
Add outer RemixErrorBoundary to catch root boundary-thrown errors

brophdawg11:
Needed to add back in a global `RemixErrorBoundary` wrapper around the full application in case the root route threw from it's own boundary.

### DIFF
--- a/integration/error-boundary-test.ts
+++ b/integration/error-boundary-test.ts
@@ -1034,14 +1034,19 @@ test.describe("Default ErrorBoundary", () => {
         await page.waitForSelector("h1#index");
       });
 
-      test("renders default boundary on render errors", async ({ page }) => {
+      test("renders default boundary on render errors", async ({
+        page,
+      }, workerInfo) => {
         let app = new PlaywrightFixture(appFixture, page);
         await app.goto("/");
         await app.clickLink("/render-error");
         await page.waitForSelector("pre");
         let html = await app.getHtml();
         expect(html).toMatch("Application Error");
-        expect(html).toMatch("Render Error");
+        // Chromium seems to be the only one that includes the message in the stack
+        if (workerInfo.project.name === "chromium") {
+          expect(html).toMatch("Render Error");
+        }
         expect(html).not.toMatch("Root Error Boundary");
 
         // Ensure we can click back to our prior page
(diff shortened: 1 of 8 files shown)


integration/error-boundary-test.ts
```diff
@@ -875,3 +875,329 @@ test.describe("loaderData in ErrorBoundary", () => {
     });
   }
 });
+
+test.describe("Default ErrorBoundary", () => {
```
brophdawg11, line 879: Document request + SPA transition tests for:

* no root boundary
* root boundary
* root boundary that throws

packages/remix-server-runtime/server.ts
```diff
@@ -259,7 +259,7 @@ async function handleDocumentRequestRR(
     );
   } catch (error: unknown) {
     if (serverMode !== ServerMode.Test) {
-      console.log("Error in entry.server handleDocumentRequest:", error);
+      console.error("Error in entry.server handleDocumentRequest:", error);
```
brophdawg11, line 262: `console.error` is what we used to use - and explains why we started seeing these in integration tests since we only blew away `console.error`

packages/remix-react/browser.tsx
```diff
@@ -55,7 +71,12 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
         future: window.__remixContext.future,
       }}
     >
-      <RouterProvider router={router} fallbackElement={null} />
+      <RemixErrorBoundary
+        location={location}
+        component={RemixRootDefaultErrorBoundary}
+      >
+        <RouterProvider router={router} fallbackElement={null} />
+      </RemixErrorBoundary>
```
brophdawg11, line 79: Wrap everything in a last-resort error boundary, and send the router location in so users can back-button out of it

packages/remix-react/browser.tsx
```diff
@@ -47,6 +53,16 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
     router = createBrowserRouter(routes, { hydrationData });
   }
 
+  // We need to include a wrapper RemixErrorBoundary here in case the root error
+  // boundary also throws and we need to bubble up outside of the router entirely.
+  // Then we need a stateful location here so the user can back0-button navigate
```
jacob-ebey, line 58:
```suggestion
  // Then we need a stateful location here so the user can back-button navigate
```

PR #5018, opened 2023-01-06
Add file extensions to compiler/loaders

staylor:
Closes: https://github.com/remix-run/remix/issues/5019

These extensions are used on some routes on the Shopify website.

Testing Strategy:

Try importing a `.mov` file - a URL should be returned, instead of a loader error.

### DIFF
--- a/.changeset/little-lobsters-decide.md
+++ b/.changeset/little-lobsters-decide.md
@@ -1,0 +1,6 @@
+---
+"remix": patch
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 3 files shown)


BillBrower-Shopify (approved):

machour (changes requested):

New extensions should also be declared in `packages/remix-dev/modules.ts`

See https://github.com/remix-run/remix/pull/3985/files for an example

machour:
You also need to target the `dev` branch instead of `main` since your updating the source code.

PR #5029, opened 2023-01-09
create global context for remix express

tomer-yechiel:
Hello Remix team,

This is a draft PR just to get feedback.

I'm submitting this pull request to suggest the addition of a new API for setting global context in Remix.
The main motivation for this api is implementing contextual logger.

for example lets say I want to add the user id to all the logs that where created during the request response cycle.
currently there is the GetLoadContextFunction but it require to propagate the logger.

```
function getGlobalContext(request){
  const userId = extractUserId(request);
 return {logger:  createLogger({"userId: 1})};
}

createRequestHandler({getGlobalContext})
...
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -470,3 +470,4 @@
 - zainfathoni
 - zhe
 - mitchelldirt
+- tomer-yechiel
--- a/packages/remix-express/server.ts
+++ b/packages/remix-express/server.ts
@@ -13,6 +13,9 @@ import {
   writeReadableStreamToWritable,
 } from "@remix-run/node";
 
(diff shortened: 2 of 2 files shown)


sergiodxa:
What's the difference with `getLoadContext`?

tomer-yechiel:
> What's the difference with `getLoadContext`?

getLoadContext inject context to loader, but if i have some deep nested code that need logger, the logger will need to be passed from the loader to the nested code.
this api allow to get the logger without passsing it
there is more info here
https://nodejs.org/api/async_context.html#class-asynclocalstorage

tomer-yechiel:
closing this, this can be implemented as an express middleware

PR #5034, opened 2023-01-09
fix: update flat-route module check for file colocation

### LINKED ISSUE #5028: Flatten route folders should not consider colocated files as routes

v0.0.0-nightly-45d4491-20230106

### Steps to Reproduce

- Install the nightly release which contains the new flatten routes convention:
`npx create-remix@v0.0.0-nightly-45d4491-20230106`
- Activate the future flag in the `remix.config.js` file: 
```
future: { 
  v2_routeConvention: true
}
```
- Apply the following routes. Make `footer.tsx` and `primary-nav.tsx` files export a default component. 
```
routes/
   app/
      _index.tsx
      primary-nav.tsx
      footer.tsx
```
- Execute the following command to see what are the resulting routes handled by Remix:
`npx remix routes`

### Expected Behavior

```
<Routes>
  <Route file="root.tsx">
    <Route path="app" index file="routes/app/_index.tsx" />
  </Route>
</Routes>
```

### Actual Behavior

```
<Routes>
  <Route file="root.tsx">
    <Route path="app" index file="routes/app/_index.tsx" />
    <Route path="app/footer" file="routes/app/footer.tsx" />
    <Route path="app/primary-nav" file="routes/app/primary-nav.tsx" />
  </Route>
</Routes>
```

Only `_index.tsx` should be considered as a route. Everything else should be ignored because considered as colocated files, as @kiliman says here:

https://github.com/remix-run/remix/pull/4880#issuecomment-1374909081

> @manzano78 You should probably create a new issue instead of commenting on a closed PR.
> 
> Anyway, yes. If you are using _flat-folders_ convention (where routes are folder names instead of filenames), then __index.tsx_ should be treated as the route and everything else as colocated files.
> 
> Can you run `npx remix routes` and paste the result? Please do so in another issue and reference this comment.
> 
> Thanks!

mcansh:
closes #5028

before:
```jsx
<Routes>
  <Route file="root.tsx">
    <Route index file="routes/_index.tsx" />
    <Route path="help" file="routes/help.tsx">
      <Route index file="routes/help._index.tsx" />
      <Route path="bam" file="routes/help.bam.tsx" />
    </Route>
    <Route path="app" index file="routes/app/_index.tsx" />
    <Route path="app/footer" file="routes/app/footer.tsx" />
    <Route path="app/primary-nav" file="routes/app/primary-nav.tsx" />
    <Route path="something" index file="routes/something/_index.tsx" />
  </Route>
</Routes>
```

after:
```jsx
<Routes>
  <Route file="root.tsx">
    <Route index file="routes/_index.tsx" />
    <Route path="help" file="routes/help.tsx">
      <Route index file="routes/help._index.tsx" />
      <Route path="bam" file="routes/help.bam.tsx" />
    </Route>
    <Route path="app" index file="routes/app/_index.tsx" />
    <Route path="something" index file="routes/something/_index.tsx" />
  </Route>
</Routes>
```

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.changeset/funny-numbers-type.md
+++ b/.changeset/funny-numbers-type.md
@@ -1,0 +1,6 @@
+---
+"remix": patch
+"@remix-run/dev": patch
+---
+
+update flat route module check for file colocation
--- a/packages/remix-dev/config/flat-routes.ts
+++ b/packages/remix-dev/config/flat-routes.ts
@@ -321,7 +321,8 @@ function getRouteMap(
   return routeMap;
 }
 
-let routeRegex = /\/(_index|([a-zA-Z0-9_$.[\]-]))\.(tsx|ts|jsx|js|mdx|md)/;
+let routeRegex =
+  /(([+][/\\][^/\\:?*]+)|[/\\]((_index)|([^/\\:?*])))\.(tsx?|jsx?|mdx?)$$/;
 
 function isRouteModuleFile(filename: string) {
   // flat files only need correct extension
--- a/packages/remix-dev/config/flat-routes.ts
+++ b/packages/remix-dev/config/flat-routes.ts
@@ -303,9 +303,13 @@ function getRouteMap(
   let nameMap = new Map<string, RouteInfo>();
 
   for (let routePath of routePaths) {
-    let routeInfo = getRouteInfo(appDirectory, prefix, routePath);
-    routeMap.set(routeInfo.id, routeInfo);
-    nameMap.set(routeInfo.name, routeInfo);
+    let routesDirectory = path.join(appDirectory, prefix);
+    let pathWithoutAppRoutes = routePath.slice(routesDirectory.length + 1);
+    if (isRouteModuleFile(pathWithoutAppRoutes)) {
+      let routeInfo = getRouteInfo(appDirectory, prefix, routePath);
+      routeMap.set(routeInfo.id, routeInfo);
+      nameMap.set(routeInfo.name, routeInfo);
+    }
   }
 
   // update parentIds for all routes
@@ -315,4 +319,15 @@ function getRouteMap(
   }
 
   return routeMap;
+}
+
+let routeRegex = /\/(_index|([a-zA-Z0-9_$.[\]-]))\.(tsx|ts|jsx|js|mdx|md)/;
+
+function isRouteModuleFile(filename: string) {
+  // flat files only need correct extension
+  let isFlatFile = !filename.includes(path.sep);
+  if (isFlatFile) {
+    return routeModuleExts.includes(path.extname(filename));
+  }
+  return routeRegex.test(filename);
 }
--- a/packages/remix-dev/config/flat-routes.ts
+++ b/packages/remix-dev/config/flat-routes.ts
@@ -324,10 +324,9 @@ function getRouteMap(
 function isRouteModuleFile(filepath: string) {
   // flat files only need correct extension
   let isFlatFile = !filepath.includes(path.sep);
-  let ext = path.extname(filepath);
   if (isFlatFile) {
-    return routeModuleExts.includes(ext);
+    return routeModuleExts.includes(path.extname(filepath));
   }
 
-  return isIndexRoute(filepath.slice(0, -ext.length));
+  return isIndexRoute(createRouteId(filepath));
 }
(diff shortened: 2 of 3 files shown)


.changeset/funny-numbers-type.md
```diff
@@ -0,0 +1,6 @@
+---
+"remix": patch
+"@remix-run/dev": patch
+---
+
+update flat route module check for file colocation
```
machour, line 6: since flat route aren't released yet, do we really need a changeset?

mcansh: probably not, easy enough to clean up when changesets opens a PR though

packages/remix-dev/config/flat-routes.ts
```diff
@@ -316,3 +320,15 @@ function getRouteMap(
 
   return routeMap;
 }
+
+let routeRegex =
+  /(([+][/\\][^/\\:?*]+)|[/\\]((_index)|([^/\\:?*])))\.(tsx?|jsx?|mdx?)$$/;
```
pcattori, line 325: This regex is too complex for me to understand at a glance. Could we break this down into simpler pieces somehow?

For example, this regex matches `/a.js`, but not `/abc.js` nor `a.js`. I'm not really sure if that's intentional or not.
It also seems to only match after the final `/` so `/abc/d.js` is only matched on `/d.js`. If we only want to match after the final path separator, I think it'd be better to use `path` utilities to get the base filename first and then match against regexes.

In any case, I think this deserves at least a comment explaining it, but probably warrants doing it a different way to keep things readable/understandable.

packages/remix-dev/config/flat-routes.ts
```diff
@@ -316,3 +320,15 @@ function getRouteMap(
...
+
+function isRouteModuleFile(filename: string) {
+  // flat files only need correct extension
+  let isFlatFile = !filename.includes(path.sep);
+  if (isFlatFile) {
+    return routeModuleExts.includes(path.extname(filename));
+  }
+  return routeRegex.test(filename);
+}
```
pcattori, line 334: nit: I think `filepath` might be a better name since a file name would never contain a path separator:

```suggestion
function isRouteModuleFile(filepath: string) {
  // flat files only need correct extension
  let isFlatFile = !filepath.includes(path.sep);
  if (isFlatFile) {
    return routeModuleExts.includes(path.extname(filepath));
  }
  return routeRegex.test(filepath);
}
```

pcattori (changes requested):

packages/remix-dev/config/flat-routes.ts
kiliman: Yes, the regex is complex. Even more so in my package as it handles more conventions. Unfortunately, JavaScript doesn't support the `x` flag (ignore whitespace) to make the regex easier to read.

Since routes can be determined as either a filename `a.b.c.tsx` or a folder `a.b.c./_index.tsx`, I don't think you can just get the base filename.

The filepath in `isRouteModuleFile` will be the full path after `routes` folder. So will never start with `/`.

To match `flat-files` convention, this is the regex: `^[^\/\\:?*]+\.(tsx?|jsx?|mdx?)$`

- Start of path `^` 
- Followed by one or more non-path characters `[^\/\\:?*]+`
- Ending with `.tsx`, etc `\.(tsx?|jsx?|mdx?)$`

image

To match `flat-folders` convention, this is the regex: `[\/\\]_index\.(tsx?|jsx?|mdx?)$`

- Path ends with `/_index.tsx`
- So it should ignore colocated files like `users/component.tsx`

image

Put them together and you get: `((^[^\/\\:?*]+)|([\/\\]_index))\.(tsx?|jsx?|mdx?)$`

image

pcattori (approved):

packages/remix-dev/config/flat-routes.ts
mcansh: instead of using a regex would using `isIndexRoute` be sufficient for non flat files?

kiliman: Probably, considering the new core feature is a subset of the `remix-flat-rotues` package. You don't need to handle quite as many cases.

But if you end up expanding the naming conventions, you'll most likely need to use regex at some point.

Issue #4708, opened 2022-11-29
🗺️ Vanilla Extract


ryanflorence:
## Progress

- [x] RFC https://github.com/remix-run/remix/discussions/5015
- [x] PR https://github.com/remix-run/remix/pull/5040

ryanflorence:
@chaance to give it the beans and then if it's good we can merge.

markdalgleish:
Merged!

chaance:
I gave it the beans

PR #5040, opened 2023-01-09
feat: add Vanilla Extract support

markdalgleish:
Closes: #4708 

- [x] Docs
- [x] Tests

I ended up writing our own custom esbuild plugin rather than using the official Vanilla Extract one since we want to have lower level control over the esbuild child compilation. This lets us ensure that assets and asset imports are handled the same as the rest of the app. We're still calling out to the `@vanilla-extract/integration` package which is the basis of all other bundler plugins, so we're still relying on the official implementation of the core logic.

Autoprefixing is not currently supported in this initial release since there's no way to intercept the generated CSS, but will be implemented as part of #5017.

Testing Strategy:

Similar testing strategy to CSS Modules and CSS side-effect imports, using integration tests to compile an app with Vanilla Extract and assert that styles end up on the page correctly. I've also added some Vanilla Extract fixtures to the deterministic build output test.


packages/remix-dev/compiler/plugins/vanillaExtractPlugin.ts
```diff
@@ -0,0 +1,184 @@
...
+    name: "vanilla-extract-transform-plugin",
+    setup(build) {
+      build.onLoad({ filter: cssFileFilter }, async ({ path }) => {
+        let source = await fse.readFile(path, "utf-8");
+
+        let contents = await transform({
+          source,
+          filePath: path,
+          rootPath: rootDirectory,
+          packageName: getPackageName(rootDirectory),
```
chaance, line 147: Curious -- what is `packageName` used for in Vanilla Extract? Is it to prevent hash clashes between libs in a monorepo?

chaance (approved):
Looks good! Just a question but nothing blocking IMO

packages/remix-dev/compiler/plugins/vanillaExtractPlugin.ts
markdalgleish: Good question — it's used to further scope hashes so that you can build library styles ahead of time without clashing, even if the source directory happens to be the same. In our case we probably could just hard code it to "remix-app".

mcansh:
> 🤖 Hello there,
> 
> We just published version `1.10.1` which includes this pull request. If you'd like to take it for a test run please try it out and let us know what you think!
> 
> Thanks!

*small* bug in the commenting workflow 🙃 - will be available in the nightly tonight

AdsonCicilioti:
I've created a React component library and Remix don't work with css files imported from static package bundle.

I'm trying the future flags unstable_cssBundle and unstable_vanillaExtract on remix.config.

PR #5042, opened 2023-01-10
fix: write assets from CSS bundle build

markdalgleish:
This fixes an issue where assets that are _only_ referenced in the CSS bundle build aren't written to disk, which impacts images referenced in CSS Modules files. This wasn't picked up in the test suite because we weren't asserting that requests for the image were successful, so I've updated all CSS bundle tests to catch this. I've confirmed that these updated tests fail against the latest code in dev.

This slipped through during development because the test image I was using was also referenced in the browser app build (via a regular style sheet) which _does_ write files to disk. I was able to reproduce this issue when I updated my test app to only use CSS Modules.

### DIFF
--- a/integration/css-modules-test.ts
+++ b/integration/css-modules-test.ts
@@ -482,12 +482,17 @@ test.describe("CSS Modules", () => {
   });
   test("image URLs", async ({ page }) => {
     let app = new PlaywrightFixture(appFixture, page);
+    let imgStatus: number | null = null;
+    app.page.on("response", (res) => {
+      if (res.url().endsWith(".svg")) imgStatus = res.status();
+    });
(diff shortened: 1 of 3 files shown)


integration/css-modules-test.ts
```diff
@@ -482,12 +482,17 @@ test.describe("CSS Modules", () => {
   });
   test("image URLs", async ({ page }) => {
     let app = new PlaywrightFixture(appFixture, page);
+    let imgStatus: number | null = null;
+    app.page.on("response", (res) => {
+      if (res.url().endsWith(".svg")) imgStatus = res.status();
```
chaance, line 487: Excellent test 👍

chaance (approved):

PR #5065, opened 2023-01-11
Fetchers should persist data through reload/resubmit

### LINKED ISSUE #5064: useFetcher in 1.10.0 no longer returns previous data during reload

1.10.0

### Steps to Reproduce

1.10.0 brings automatic revalidation to `fetcher.load(...)` calls, which is awesome! I'll be excited to get rid of the many `useEffect` calls I have to reload data.

However, it seems that the data handling logic has changed. In 1.9.0, calling `load` on a fetcher that already had data loaded would go through these states:

```
{ state: "idle", data: myOldData }
// Call fetcher.load("/url") again
{ state: "loading", data: myOldData }
{ state: "idle", data: myNewData}
```

In 1.10.0, it's now:
```
{ state: "idle", data: myOldData }
// Call fetcher.load("/url") again, or submit a form which triggers automatic revalidation, and data is undefined
{ state: "loading", data: undefined }
{ state: "idle", data: myNewData}
```

So while it's loading the new data, the old data is no longer being returned. However, the UI often depends on that old data until the new data is available. At the very least, this ends up flashing empty content until the new data is available, but it can actually break the app.

In one of our examples, we load a checklist on-demand in a modal using `useFetcher`. As soon as we check off an item on the list, which submits a form, the checklist disappears because `fetcher.data` is now undefined. Even worse, because the checklist component unmounts since the checklist data is undefined, the child component that made the form submission also gets unmounted. This seems to cancel the revalidation requests, so now the top level checklist modal component's `useFetcher` gets stuck in `state: "loading"`. 

I can add some `useState` calls to hold the most recent version of the data if this is now the expected behavior, but I don't need to do that with `useLoaderData()` so it feels inconsistent.

### Expected Behavior

`useFetcher` should return the old data until the new data is available, just like `useLoaderData()` already does for the route's data.

### Actual Behavior

`useFetcher` returns `data: undefined` any time it's in a loading state, regardless of previously-loaded data.

brophdawg11:
Fixes an issue in the fetcher back-compat layer where fetchers were losing their data during reload/resubmissions

Closes #5064

### DIFF
--- a/packages/remix-react/transition.ts
+++ b/packages/remix-react/transition.ts
@@ -120,7 +120,7 @@ export type FetcherStates<TData = any> = {
     formData: FormData;
     formEncType: string;
     submission: ActionSubmission;
-    data: undefined;
+    data: TData | undefined;
   };
   SubmittingLoader: {
     state: "submitting";
--- a/.changeset/eighty-donkeys-run.md
+++ b/.changeset/eighty-donkeys-run.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Fetchers should persist data through reload/resubmit
--- a/integration/fetcher-test.ts
+++ b/integration/fetcher-test.ts
@@ -1,4 +1,4 @@
-import { test } from "@playwright/test";
+import { expect, test } from "@playwright/test";
 
 import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
 import type { Fixture, AppFixture } from "./helpers/create-fixture";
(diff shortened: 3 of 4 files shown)


packages/remix-react/transition.ts
```diff
@@ -120,7 +120,7 @@ export type FetcherStates<TData = any> = {
     formData: FormData;
     formEncType: string;
     submission: ActionSubmission;
-    data: undefined;
+    data: TData | undefined;
```
brophdawg11, line 123: I think this was just always a type bug?  We absolutely kept data around](https://github.com/remix-run/remix/blob/remix%401.9.0/packages/remix-react/transition.ts#L676) on fetcher action resubmissions - so I'm not even sure how the types didn't yell at us there...

mcansh (approved):

packages/remix-react/transition.ts
brophdawg11: Huh - yeah I'm not sure why this wasn't yelling at us before?

https://www.typescriptlang.org/play?#code/C4TwDgpgBAYhwGMAWEBOBlYBDYEDOUAvFAN4BQUUAkgCYA2EAXKRZVHtrswEQCW9EbgG5WlGjizMArgDsaEAGa8ZEGiMoBfdVHRSARgFtewYMoDmAQQSmA9jObk27Tkyjc8+oyfPDRUcdjScorKqtparAAidq6ObBw4rnwCvk4BklBYMiDhIhFkEAAeYDaowFCgkLDwyGhE1YgoGC54ANoA1hAgNgoNtc2JeAC6ImQM5TQxcI1ozNP99XEJXFAA5PwMqwA0rOk8eDYG8Eg+ZPnjUCoA7vNNczVNmIOt3LqGxqYylta8dtxDi1YyySHne3i+3B2lAA9NCoAB1JAgKC8PAyVblYAnAhZKBoVClAD8lDeXk+31sMigqAgAEcpLwaQQ9rJ5EoVDRCbsJMxJipbmgAHTpKAAH1FUFZIQ5ZxEQA

pcattori: The issue is that `TData` is `any`, which TS treats as compatible with `undefined`. If we more accurately typed `TData` as `unknown` (or whatever it _actually_ is), then the error would show up in TS.

PR #5076, opened 2023-01-12
docs(route/headers): add missing types on `headers()`

irsooti:
I noted it misses the types over the `headers` function. 

Another alternative could be adding a new type `HeadersArgs` using the same conventions as for `loader` and `action`


machour (approved):

machour:
Thank you!

Issue #5073, opened 2023-01-12
[v1.10.0] TypeError: Cannot read properties of undefined (reading 'matches')


juniorforlife:
### What version of Remix are you using?

1.10.0

### Steps to Reproduce

- Upgrade to 1.10.0
- Reload the app
- The whole app crashes with this error 
`
TypeError: Cannot read properties of undefined (reading 'matches')
at Object.createStaticRouter (project-name/node_modules/@remix-run/react/node_modules/react-router-dom/server.js:189:25)
`

### Expected Behavior

No crashes

### Actual Behavior

Crashes with error `TypeError: Cannot read properties of undefined (reading 'matches')`

machour:
Can you provide a reproduction somewhere?
Not much we can do with this report as it is.

juniorforlife:
@machour  I can't share the source code because it's proprietary. Is it useful to upload a video recording?

felixh10r:
@juniorforlife Do you access `EntryContext.matches` in your `entry.server.ts`?
f604197be74588c5bda1315443c47a2487e6e171 moved `EntryContext.matches` to `EntryContext.staticHandlerContext.matches`.

juniorforlife:
I reverted to 1.9.0 and it works perfectly

machour:
@brophdawg11 does this ring any bells?

kiliman:
@juniorforlife v1.10 is the first version where Remix uses the new React Router.

I'm pretty sure if you scroll up in the dev console, you'll see additional errors which is most likely the root cause. Please report back with more info.

kiliman:
Also, when you upgrade, make sure *ALL* your Remix dependencies are the same version. I see your dev dependencies are still at 1.7.6

juniorforlife:
@kiliman @machour I've upgraded the dev dependencies and it works now. Thanks for your time

juniorforlife:
@kiliman do I need to install @remix-run/router for 1.10.0?
Here's my package.json
```
"dependencies": {
    "@remix-run/node": "^1.10.0",
    "@remix-run/react": "^1.10.0",
    "@remix-run/vercel": "^1.10.0",
    "@tailwindcss/line-clamp": "^0.4.2",
    "@vercel/node": "^2.6.2",
    "axios": "^1.2.1",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
    "remix": "^1.10.0"
  },
  "devDependencies": {
[...]

PR #5077, opened 2023-01-12
chore(bug-report): Make sure the reporter have aligned remix dependencies

machour:
We often receive regressions reports where it turns out that the OP only updated a single package, leading to unexpected behaviors.

Example: #5073

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -25,6 +25,12 @@ body:
       label: What version of Remix are you using?
     validations:
       required: true
+  - type: checkboxes
+    attributes:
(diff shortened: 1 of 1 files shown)


machour:
I also feel like this should be mentioned in the documentation, but I'm not sure where.. suggestions?

mcansh:
i've also seen this from time to time and was my main driving factor for https://github.com/remix-run/remix/pull/5004

kiliman:
I have a script I use to update remix versions

```bash
npm outdated | grep -Eo '@remix-run/[-a-z]+' | uniq | xargs npm up
```
https://gist.github.com/kiliman/ba4586bea1acf06fe6cc12e76e8db94a

MichaelDeBoey (approved):

Issue #5016, opened 2023-01-05
🗺 Accept http `to` values in Link, treat like `<a>`


ryanflorence:
### Discussed in https://github.com/remix-run/remix/discussions/4888

mcansh:
PR for React Router support here 👉 https://github.com/remix-run/react-router/pull/9900 - in Remix we're _probably_ going to need to use the same regex there too in order to add `prerender` links

ryanflorence:
@mcansh to get back to it for the Remix PR

mcansh:
so browser prerendering is a little iffy, https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/prerender, but does work - however chrome now has this "speculation rules api", which uses a script tag and a custom `type=speculationrules` thing

PR #5092, opened 2023-01-13
feat(remix-react): allow `<Link>`'s "to" prop to accept absolute urls

mcansh:
external "to" detection should be brought over to React Router (pr: https://github.com/remix-run/react-router/pull/9900)

Closes: #5016

- [ ] Docs
- [ ] Tests

Testing Strategy:

our current testing for Link should be enough, both here and in RR

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -291,7 +291,7 @@ function usePrefetchBehavior(
  */
 let NavLink = React.forwardRef<HTMLAnchorElement, RemixNavLinkProps>(
   ({ to, prefetch = "none", ...props }, forwardedRef) => {
-    let isAbsolute = 
+    let isAbsolute =
       typeof to === "string" &&
       (/^[a-z+]+:\/\//i.test(to) || to.startsWith("//"));
 
@@ -325,12 +325,12 @@ export { NavLink };
  * @see https://remix.run/components/link
  */
 let Link = React.forwardRef<HTMLAnchorElement, RemixLinkProps>(
-  ({ to, prefetch = "none", children, ...props }, forwardedRef) => {
-    let location = typeof to === "string" ? to : createPath(to);
+  ({ to, prefetch = "none", ...props }, forwardedRef) => {
     let isAbsolute =
-      /^[a-z+]+:\/\//i.test(location) || location.startsWith("//");
+      typeof to === "string" &&
+      (/^[a-z+]+:\/\//i.test(to) || to.startsWith("//"));
 
-    let href = useHref(location);
+    let href = useHref(to);
     let [shouldPrefetch, prefetchHandlers] = usePrefetchBehavior(
       prefetch,
       props
@@ -338,9 +338,12 @@ let Link = React.forwardRef<HTMLAnchorElement, RemixLinkProps>(
 
     return (
       <>
-        <RouterLink ref={forwardedRef} to={to} {...props} {...prefetchHandlers}>
-          {children}
-        </RouterLink>
+        <RouterLink
+          ref={forwardedRef}
+          to={to}
+          {...props}
+          {...prefetchHandlers}
+        />
         {shouldPrefetch && !isAbsolute ? (
           <PrefetchPageLinks page={href} />
         ) : null}
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -291,9 +291,9 @@ function usePrefetchBehavior(
  */
 let NavLink = React.forwardRef<HTMLAnchorElement, RemixNavLinkProps>(
   ({ to, prefetch = "none", ...props }, forwardedRef) => {
-    let location = typeof to === "string" ? to : createPath(to);
-    let isAbsolute =
-      /^[a-z+]+:\/\//i.test(location) || location.startsWith("//");
+    let isAbsolute = 
+      typeof to === "string" &&
+      (/^[a-z+]+:\/\//i.test(to) || to.startsWith("//"));
 
     let href = useHref(to);
     let [shouldPrefetch, prefetchHandlers] = usePrefetchBehavior(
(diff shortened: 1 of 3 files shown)


packages/remix-react/components.tsx
```diff
@@ -322,20 +322,34 @@ export { NavLink };
...
     );
+
+    if (isExternal) {
+      return (
+        <>
+          <a ref={forwardedRef} href={to} {...props} {...prefetchHandlers}>
+            {children}
+          </a>
+          {shouldPrefetch ? (
+            <link key={href} rel="prefetch" as="document" href={to} />
```
sergiodxa, line 342: I think you don't need the `as="document"` here, I'm using it on my blog without that and it works https://github.com/sergiodxa/sergiodxa.com/blob/ca1fa0d9c7f8d92322b01f009951bbfdd54c6214/app/root.tsx#L213-L215

Also when I tried to prefetch links to external websites the browser didn't re-used the prefetched response and did a new request.

mcansh: looks like `rel=prerender` might be what we're actually after https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/prerender

sergiodxa: Is there any browser supporting prerender? I tried in Safari and Chromium (Edge) and it did nothing 🤔

mcansh: back working on this, looks like pre-render was removed 🫠 https://developer.chrome.com/blog/prerender-pages/#a-brief-history-of-prerender

mcansh: now it's a script tag apparently and idk how to feel lol 
```html
<script type="speculationrules">
{
  "prerender": [
    {"source": "list",
     "urls": ["/home", "/about"]}
  ],
  "prefetch": [
    {"source": "list",
     "urls": ["https://en.wikipedia.org/wiki/Hamster_racing"],
     "requires": ["anonymous-client-ip-when-cross-origin"]}
  ]
}
</script>
```

mcansh: additional update: pre-render does work still in chrome, confirmed using https://stackoverflow.com/a/64330790
image

brophdawg11: Do we really _need_ to try to optimize external links to apps we don't control, especially if it's potentially not straightforward?

mcansh: _probably_ not, but would've been nice if it was a straightforward api

brophdawg11: yeah definitely

packages/remix-react/components.tsx
```diff
@@ -315,28 +316,34 @@ let NavLink = React.forwardRef<HTMLAnchorElement, RemixNavLinkProps>(
...
  * @see https://remix.run/api/remix#link
  */
 let Link = React.forwardRef<HTMLAnchorElement, RemixLinkProps>(
-  ({ to, prefetch = "none", ...props }, forwardedRef) => {
-    let href = useHref(to);
+  ({ to, prefetch = "none", children, ...props }, forwardedRef) => {
+    let location = typeof to === "string" ? to : createPath(to);
+
+    let isAbsolute =
+      /^[a-z+]+:\/\//i.test(location) || location.startsWith("//");
```
mcansh, line 331: same detection we do over in router land https://github.com/remix-run/react-router/blob/dev/packages/react-router-dom/index.tsx#L421-L424

packages/remix-react/components.tsx
```diff
@@ -295,6 +296,10 @@ function usePrefetchBehavior(
  */
 let NavLink = React.forwardRef<HTMLAnchorElement, RemixNavLinkProps>(
   ({ to, prefetch = "none", ...props }, forwardedRef) => {
+    let location = typeof to === "string" ? to : createPath(to);
+    let isAbsolute =
+      /^[a-z+]+:\/\//i.test(location) || location.startsWith("//");
```
brophdawg11, line 301: I don't think this occurred to me in RR, but we should only need to check the regex if they passed in a string - since you can't specify an absolute URL using the object syntax (`<Link to={{ pathname: 'https://...' }}>`) - so could we avoid the need for `createPath`/`location` (which is already handled in `useHref`) and just do this?

```suggestion
    let isAbsolute = 
      typeof to === "string" &&
      (/^[a-z+]+:\/\//i.test(to) || to.startsWith("//"));
```

packages/remix-react/components.tsx
```diff
@@ -308,35 +313,42 @@ let NavLink = React.forwardRef<HTMLAnchorElement, RemixNavLinkProps>(
           {...props}
           {...prefetchHandlers}
         />
-        {shouldPrefetch ? <PrefetchPageLinks page={href} /> : null}
+        {shouldPrefetch && !isAbsolute ? (
+          <PrefetchPageLinks page={href} />
+        ) : null}
```
brophdawg11, line 318: Love this simplified approach 👌

packages/remix-react/components.tsx
```diff
@@ -308,35 +313,42 @@ let NavLink = React.forwardRef<HTMLAnchorElement, RemixNavLinkProps>(
...
-        <RouterLink
-          ref={forwardedRef}
-          to={to}
-          {...props}
-          {...prefetchHandlers}
-        />
-        {shouldPrefetch ? <PrefetchPageLinks page={href} /> : null}
+        <RouterLink ref={forwardedRef} to={to} {...props} {...prefetchHandlers}>
+          {children}
+        </RouterLink>
```
brophdawg11, line 348: Do we need to pull this out or could it stay in `props` and get spread on?

brophdawg11 (approved):
A few minor comments but looks good to me!

packages/remix-react/components.tsx
mcansh: ha, i was also thinking that myself a few minutes ago 😆

packages/remix-react/components.tsx
mcansh: probably fine to keep in props and spread, i.. i just don't like that lmao

PR #5095, opened 2023-01-13
fix(remix-node): make global `fetch` return an instance of global `Response`

acoreyj:
Closes: #4395 

- [ ] Docs
- [ x] Tests

Testing Strategy:

This test covers this code: packages/remix-node/__tests__/fetch-test.ts  `it("should return an instance of Response and WebResponse"`


acoreyj:
Note that this blocks being able to use @authjs in remix node environments

https://github.com/nextauthjs/next-auth/pull/6270

acoreyj:
Hmm I'm struggling to think of ways to mock this test to work in CI as mocking the fetch obviously ruins the point, any thoughts @MichaelDeBoey  @jacob-ebey

acoreyj:
@MichaelDeBoey I pulled out the logic for creating the response so I could mock the fetch itself and test this which should pass in the CI now

packages/remix-node/fetch.ts
```diff
@@ -72,5 +78,11 @@ export const fetch: typeof webFetch = (
...
   };
 
-  return webFetch(info, init as RequestInit);
+  return webFetch(info, init as RequestInit)
+    .then((res) => {
+      return getReturnableResponse(res);
+    })
+    .catch((res) => {
+      return getReturnableResponse(res);
+    });
```
MichaelDeBoey, line 87: This could have a smaller footprint if we do

```suggestion
    .then(getReturnableResponse)
    .catch(getReturnableResponse);
```

MichaelDeBoey (approved):

MichaelDeBoey:
Closing this as it seems to be a duplicate of @mcansh's #4148

PR #5116, opened 2023-01-17
chore(bug-report): Failing test for issue Redirect with absolute path results in 404s

emilbryggare:
This is a failing test case for the issue Redirect with absolute path results in 404s. 

Link to bug report issue: https://github.com/remix-run/remix/issues/5117


brophdawg11:
This was resolved in 1.11.0 - moved this test into redirects-test for future regression purposes

brophdawg11:
Unsure why CI isn't running - going to close and reopen and see if that jump starts it

PR #5125, opened 2023-01-18
feat: pass a custom config reader to `watch`

frandiox:
Hi! This issue was raised in another channel but here's a summary:

The `watch` function of the Remix compiler currently accepts a `RemixConfig` as its first parameter. The problem is that after adding new files to the project, the `watch` function reads a fresh config from the `remix.config.js` file, which removes any kind of customization to the config object that we passed originally in the parameter.

Instead of reading a fresh config from the filesystem, this PR changes it to accept a config getter function so that we can apply custom default values every time. I've made it optional to avoid breaking changes since it is publicly mentioned in [this discussion](https://github.com/remix-run/remix/discussions/4547) and other devs might be importing `watch` already.

I've tested this in Hydrogen v2 and it fixes the problem we were having.

--

Not sure if I should add a changeset since this is just a change of an internal API 🤔


jacob-ebey (approved):

packages/remix-dev/compiler/watch.ts
```diff
@@ -29,7 +29,7 @@ export type WatchOptions = Partial<CompileOptions> & {
 };
 
 export async function watch(
-  config: RemixConfig,
+  configGetter: RemixConfig | (() => RemixConfig | Promise<RemixConfig>),
```
pcattori, line 32: Could you separate out the initial config from getter for the subsequent configs? That will make the types simpler and won't require any `typeof` checks further on.

```suggestion
  initialConfig: RemixConfig,
  reloadConfig: (rootDirectory: string) => RemixConfig = readConfig,
```

Also note that the root directory _could_ change when the config is reloaded, so to be extra defensive, its nice to accept the current `rootDirectory` as an arg.

pcattori: nit: I also think `reloadConfig` conveys that this is doing some I/O so I prefer that to `getNewConfig`

pcattori (changes requested):

packages/remix-dev/compiler/watch.ts
frandiox: >Could you separate out the initial config from getter for the subsequent configs? That will make the types simpler and won't require any typeof checks further on.

I'm ok with the change but this would mean `watch` now has 2 optional parameters: `watch(initialConfig, reloadConfig?, options?)` so it could potentially go `watch(config, undefined, {...})`. It's also a breaking change for anyone using `watch` already. Should we move `reloadConfig` to the existing options object instead? Just want to double check with you and @jacob-ebey before updating the PR 👍

pcattori: yea including it in `options` sounds good

frandiox: Alright, I've updated the PR, thanks!

pcattori (approved):

pcattori:
@frandiox the upstream lint errors have been fixed so you should be able to rebase off of `dev` branch and get tests passing

PR #5133, opened 2023-01-18
The new dev server

pcattori:
## Approach

Running the new dev server does the following:
1. Builds an initial build
2. Watches for changes to app source code and rebuilds when changes occur
3. Waits for the app server (i.e. `server.js`) to be ready
4. Sends a `RELOAD` message over the websocket

For step (3), the dev server polls `GET <app server>/__REMIX_ASSETS_MANIFEST` and compares the build hash received from the app server with the latest build hash produced by the compiler run within the dev server.

### 1 No more require cache purge

App server is now only restarted when a rebuild happens, not on every request.
That means that in-memory persistence is fixed. So no need to use the `global` trick anymore when dealing with DB connections, in-memory caches, etc..

In fact app server might not need to restart _at all_ if the app server supports some sort of server-side HMR.

### 2 Supports _any_ app server

Previously, dev mode only worked for `@remix-run/serve`, [not for custom servers](https://github.com/remix-run/remix/blob/main/packages/remix-dev/devServer/serve.ts#L34) like `server.{js,ts}`.
Now _any_ custom server is compatible with the new dev server.

### 3 Support ESM mode for Node server?

Previously, we needed fine-grain control of the require cache, which prevented Remix servers from being run in ESM mode.
Since the new dev server does not need to purge the require cache at all, this limitation has been removed.

I _think_ this means we could support ESM-mode Remix servers now if we wanted to.

## Future work

- Stop compiling/bundling custom app servers into `build/index.js`
- Deprecate / remove the virtual module for the server build. Just import from `./build/index.js` manually.
- Give back control to users for how they want to compile/run their server. E.g. if using TS, just run `ts-node ./server.js`

## TODO

- [x] Changeset

---

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

### DIFF
--- a/.changeset/mean-clocks-bow.md
+++ b/.changeset/mean-clocks-bow.md
@@ -1,0 +1,54 @@
+---
+"remix": minor
+"@remix-run/dev": minor
+"@remix-run/react": minor
+"@remix-run/serve": minor
+"@remix-run/server-runtime": minor
+---
+
+# The new dev server
+
+The new dev flow is to spin up the dev server _alongside_ your normal Remix app server:
+
+```sh
+# spin up the new dev server
+remix dev
+
+# spin up your app server in a separate tab or via `concurrently`
+nodemon ./server.js
+```
+
+The dev server will build your app in dev mode and then rebuild whenever any app files change.
+It will also wait for your app server to be "ready" (more on this later) before triggering a live reload in your browser.
+
+## Benefits
+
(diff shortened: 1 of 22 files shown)


pcattori:
@kiliman thanks for trying it out! I just merged a couple fixes and improvements (https://github.com/remix-run/remix/pull/5137) that should fix this for you. Also the expectation is that your app server is running behind `nodemod` or something else that will pick up the changes.

PR #5134, opened 2023-01-18
docs: added meta@v2 docs

ryanflorence:


docs/route/meta.md
```diff
@@ -114,7 +118,217 @@ const meta: MetaFunction<
...
+  return [...parentMeta, { title: "Projects" }];
+};
+```
+
+## `data`
+
+This is the data from your loader.
+
+```tsx
+export function loader({ params }) {
```
MichaelDeBoey, line 204: Let's make the `loader` async, just like in the rest of our docs

```suggestion
export async function loader({ params }: LoaderArgs) {
```

docs/route/meta.md
```diff
@@ -114,7 +118,217 @@ const meta: MetaFunction<
...
+  return [{ title: data.task.name }];
+}
+```
+
+## `parentsData`
+
+Often you'll need the data from a parent route, you can look it up by route ID on `parentsData`
+
+```tsx filename=routes/project/$pid/tasks/$tid.tsx
+export function loader({ params }) {
```
MichaelDeBoey, line 218:
```suggestion
export async function loader({ params }: LoaderArgs) {
```

docs/route/meta.md
```diff
@@ -114,7 +118,217 @@ const meta: MetaFunction<
...
+  return getTask(params.tid);
+}
+
+export function meta({ data, parentsData }) {
```
MichaelDeBoey, line 222:
```suggestion
export const meta: V2_MetaFunction<
  typeof loader,
  { "routes/project/$pid": typeof projectDetailsLoader }
> = ({ data, parentsData }) => {
```

docs/route/meta.md
```diff
@@ -114,7 +118,217 @@ const meta: MetaFunction<
...
+```
+
+```tsx bad filename=app/routes/projects.tsx
+export const meta = () => {
+  return [{ title: "Projects" }];
+};
+```
+
+```tsx bad filename=app/routes/projects/$id.tsx
+export const meta: V2_MetaFunction = ({ data }) => {
```
MichaelDeBoey, line 262:
```suggestion
export const meta: V2_MetaFunction<typeof loader> = ({ data }) => {
```

docs/route/meta.md
```diff
@@ -114,7 +118,217 @@ const meta: MetaFunction<
...
+        <Scripts />
+      </body>
+    </html>
+  );
+}
+```
+
+### Avoid Meta in Parent Routes
+
+You can also avoid the merge problem by simply not exporting meta that you want to override from parent routes. Instead of defining meta on the parent route, use the [index route][index-route]. This way you can avoid complex merge logic for things the title. Otherwise you will need to find the parent title descriptor and replace it with the child's title. It's much easier to simply not need to override by using index routes.
```
MichaelDeBoey, line 304:
```suggestion
You can also avoid the merge problem by simply not exporting meta that you want to override from parent routes. Instead of defining meta on the parent route, use the [index route][index-route]. This way you can avoid complex merge logic for things like the title. Otherwise, you will need to find the parent title descriptor and replace it with the child's title. It's much easier to simply not need to override by using index routes.
```

PR #5149, opened 2023-01-19
chore(eslint): ignore `no-extraneous-deps` in markdown files

chaance:


.eslintrc.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
   plugins: ["markdown"],
   overrides: [
     {
-      files: ["rollup.config.js", "**/*.md"],
+      files: ["rollup.config.js", "**/*.md/**"],
```
brophdawg11, line 14: ughh ok that makes sense, I assume that lets you target language-specific code blocks in a markdown file

brophdawg11 (approved):

.eslintrc.js
brophdawg11: nice find!

mcansh (approved):

.eslintrc.js
mcansh: > ughh ok that makes sense, I assume that lets you target language-specific code blocks in a markdown file

yep! i tried it locally, but didn't try extending it to the other configs

PR #5154, opened 2023-01-20
bug: CSS modules are ignored when component is re-exported using index.js

iamkd:
I know CSS Modules are unstable but hopefully this will help improve and release them faster 😄 

This is happening when the file structure is roughly like this (other combinations could also be prone to this bug):
```
components/
├─ Button/
│  ├─ index.js
│  ├─ Button.js
│  ├─ Button.module.css
```

where `Button.js` is importing the CSS module and using it, while `index.js` is re-exporting it using
`export * from './Button.js'`.

Basically, the bug happens depending on how this component is imported:
* `import { Button } from '~/components/Button'` will ignore the CSS module and will not add it to the bundle. The classname will still be generated but will be useless.
* `import { Button } from '~/components/Button/Button'` will bypass the `index.js` and work as expected.

Using `index.js` to re-export stuff and make imports prettier is not something everybody likes, but this is still valid code and CSS modules should be bundled in this case as well. 

This PR includes a bug report test that checks if the `cssBundleHref` is defined when the component is imported in a way described above. Because the test consists of a single component, it will fail with `cssBundleHref` being undefined. In case of multiple components, only the styles that are imported correctly will be included in the bundle.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -173,6 +173,7 @@
 - Hopsken
 - humphd
 - hzhu
+- iamkd
 - IAmLuisJ
 - ianduvall
 - illright
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -2,7 +2,12 @@ import { test, expect } from "@playwright/test";
 
 import { PlaywrightFixture } from "./helpers/playwright-fixture";
 import type { Fixture, AppFixture } from "./helpers/create-fixture";
-import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
+import {
+  createAppFixture,
+  createFixture,
+  css,
+  js,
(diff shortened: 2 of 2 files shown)


markdalgleish:
Thanks for the detailed bug report and test case! Unfortunately this is actually [an issue with esbuild's CSS tree-shaking](https://github.com/evanw/esbuild/issues/1370) and not something we have control over. We're currently tracking this issue in https://github.com/remix-run/remix/issues/5270.

Until the esbuild issue is fixed, as a workaround you can avoid using `export *` and use named re-exports instead.

```diff
-export * from './Button.js';
+export { Button } from './Button.js';
```

Since we're currently unable to fix this issue within Remix, I'm closing this in favour of #5270.

PR #5157, opened 2023-01-20
feat(remix-react): add types to `useRouteLoaderData`

lpsinger:
Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy: added unit tests


lpsinger:
I didn't add any documentation for this because I could not find the file for https://remix.run/docs/en/v1/hooks/use-route-loader-data on the dev branch.

machour (dismissed):
Hi @lpsinger and thank you for your contribution!

This change needs to target the `dev` branch. Also make sure you don't include other commits than yours in the PR

lpsinger:
> This change needs to target the `dev` branch. Also make sure you don't include other commits than yours in the PR

Fixed.

WesleyYue:
Is there anything blocking this from merging?

lpsinger:
Any news on this?

packages/remix-react/components.tsx
```diff
@@ -1140,6 +1141,15 @@ export function useLoaderData<T = AppData>(): SerializeFrom<T> {
   return useLoaderDataRR() as SerializeFrom<T>;
 }
 
+/**
+ * Returns the loaderData for the given routeId.
+ *
+ * @see https://remix.run/hooks/use-route-loader-data
+ */
+export function useRouteLoaderData<T = AppData>(routeId: string): SerializeFrom<T> {
```
tom-sherman, line 1149: To be safe this should return `SerializeFrom<T> | undefined`, as the `routeId` can technically be any string.

lpsinger: Wouldn't that also be true of `useLoaderData`? Shouldn't `useLoaderData` and `useRouteLoaderData` have the same return type?

tom-sherman: No, because useLoaderData can never return undefined.

useLoaderData essentially only has a single input which is the generic argument. Of course there is room for misconfiguration here but it's small and obvious to spot.

useRouteLoaderData is different. It has a secondary routeId argument that a user can pass any old string to. You now not only have to provide the correct type to the generic (import the correct loader type) but also provide the corresponding ID for that loader.

If users are sure that the route ID is correct, then they can use a non-null assertion but this shouldn't be the default.

I think there is a case for having exceptions for well know route IDs, like the root. That could be special cases with an overload.

KevinMind:
Plus 1 on when can we merge this?

lpsinger:
It's been a bit over a month now since the last correspondence on this PR. Any updates?

MichaelDeBoey (approved):

PR #5160, opened 2023-01-20
fix(remix-dev/flat-routes): use `index` without leading underscore for colocation

mcansh:
before:
```
  app_.projects.$id.roadmap/
    _index.tsx
    chart.tsx
    update-timeline.server.tsx
```

after:
```
  app_.projects.$id.roadmap/
    index.tsx
    chart.tsx
    update-timeline.server.tsx
```

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/__tests__/flat-routes-test.ts
+++ b/packages/remix-dev/__tests__/flat-routes-test.ts
@@ -4,6 +4,7 @@ import {
   createRoutePath,
   flatRoutesUniversal,
   getRouteSegments,
+  isIndexRoute,
 } from "../config/flat-routes";
 import type { ConfigRoute } from "../config/routes";
 
@@ -25,10 +26,10 @@ describe("flatRoutes", () => {
       ["nested/$slug", "/nested/:slug"],
       ["flat.$slug", "/flat/:slug"],
       ["flat.sub", "/flat/sub"],
-      ["nested/_index", "/nested"],
+      ["nested/index", "/nested"],
       ["flat._index", "/flat"],
       ["_index", undefined],
-      ["_layout/_index", undefined],
+      ["_layout/index", undefined],
       ["_layout/test", "/test"],
       ["_layout.test", "/test"],
       ["_layout/$slug", "/:slug"],
@@ -88,7 +89,8 @@ describe("flatRoutes", () => {
     for (let [input, expected] of tests) {
       it(`"${input}" -> "${expected}"`, () => {
         let routeSegments = getRouteSegments(input);
-        expect(createRoutePath(routeSegments)).toBe(expected);
+        let isIndex = isIndexRoute(input);
+        expect(createRoutePath(routeSegments, isIndex)).toBe(expected);
       });
     }
 
@@ -332,9 +334,9 @@ describe("flatRoutes", () => {
         },
       ],
       [
-        "routes/app_.skipall_/_index.tsx",
+        "routes/app_.skipall_/index.tsx",
         {
-          id: "routes/app_.skipall_/_index",
+          id: "routes/app_.skipall_/index",
           index: true,
           parentId: "root",
           path: "app/skipall",
@@ -580,9 +582,9 @@ describe("flatRoutes", () => {
         },
       ],
       [
-        "routes/brand/_index.tsx",
+        "routes/brand/index.tsx",
         {
-          id: "routes/brand/_index",
+          id: "routes/brand/index",
           parentId: "root",
           path: "brand",
           index: true,
--- a/packages/remix-dev/__tests__/flat-routes-test.ts
+++ b/packages/remix-dev/__tests__/flat-routes-test.ts
@@ -289,6 +289,16 @@ describe("flatRoutes", () => {
         },
       ],
 
+      // validate windows path
+      [
+        "routes\\app.windows\\path.tsx",
+        {
+          id: "routes/app.windows/path",
+          parentId: "routes/app",
+          path: "windows/path",
+        },
+      ],
+
       // Opt out of parent layout
       [
         "routes/app_.projects.$id.roadmap[.pdf].tsx",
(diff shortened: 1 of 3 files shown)


packages/remix-dev/__tests__/flat-routes-test.ts
```diff
@@ -289,6 +289,16 @@ describe("flatRoutes", () => {
         },
       ],
 
+      // validate windows path
```
mcansh, line 292: all tests already replace '/' with path.sep https://github.com/remix-run/remix/blob/5f30962cdf2d7642eae24af8a392a1d136bf641d/packages/remix-dev/__tests__/flat-routes-test.ts#L615, but i wanted to explicitly test it :)

pcattori (approved):

packages/remix-dev/config/flat-routes.ts
```diff
@@ -103,8 +103,11 @@ export function flatRoutesUniversal(
   return routes;
 }
 
-function isIndexRoute(routeId: string) {
-  return routeId.endsWith("_index");
+export function isIndexRoute(routeId: string) {
+  let isFlatFile = !routeId.includes(path.posix.sep);
+  return isFlatFile
+    ? routeId.endsWith("_index")
+    : /^(.*)\/index$/.test(routeId);
```
pcattori, line 110: nit: I don't think the `^(.*)` is necessary

```suggestion
    : /\/index$/.test(routeId);
```

pcattori (approved):

packages/remix-dev/config/flat-routes.ts
mcansh: yeah i had it like that originally, but thought better safe than sorry 🤷‍♂️

Xiphe:
Hi 👋 and thanks for the fix 👍

I somehow find it irritating that `routes/index.ts` is still not recognized while `routes/app/index.tsx` is. 
Is there a reasoning behind that or is it a bug?

lennerd:
I believe it‘s a bug. Or at least the ideas and conventions of colocation explained in the RFC/Docs and the implementation do not match yet. See my comment (and the RFC) here: https://github.com/remix-run/remix/discussions/4482#discussioncomment-4741560

A correct implementation (as far as I was able to test it) can be added via @kiliman library https://github.com/kiliman/remix-flat-routes

As @ryanflorence is saying in his [gist](https://gist.github.com/ryanflorence/0dcc52c2332c2f6e1b52925195f87baf) :

> while the example is exclusively files, they are really just "import paths". So you could make a folder for a route instead and the index file will be imported, allowing all of a route's modules to live along side each other.

chaance:
@lennerd If there are still bugs, do you mind opening a new issue? It'll be much easier to get it prioritized that way.

kiliman:
@Xiphe and @lennerd There's a difference between v2 routing and my package.

With v2 routing, it **only** supports `_index` for the _index_ route (`<Route index>`) and `/index.tsx` for the `flat-folder` route file.

**flat-files**
`routes/_index.tsx` => `/` index route
`routes/app._index.tsx` => `/app/` index route

**flat-folders**
`routes/_index/index.tsx` => `/` index route
`routes/app._index/index.tsx` => `/app/` index route

Xiphe:
He, was about to post a reference to your comment here: https://github.com/remix-run/remix/issues/5216#issuecomment-1400482019 and that I seem to have understood things wrong. 

The irritating thing for me was that both routes and folders use "index" but they don't mean the same thing.

kiliman:
Yeah, it is a little confusing, but there is a difference in meaning. That's why v2 routing makes it explicit which type you mean: _index_ route vs colocated index _route file_.  It's simpler to teach it that way.

> So the flat-folder equivalent of `routes/app._index.tsx` is `routes/app._index/index.tsx `and not ~~routes/app/index.tsx~~
> Also `routes/app.tsx` and `routes/app/index.tsx` are equivalent where none of it is an index route.

Yes that is correct.

ryanflorence:
> I somehow find it irritating that routes/index.ts is still not recognized as the index root route while routes/app/index.tsx is considered the index for the app layout.

> Similarly when using dot routes routes/app._index.tsx the underscore is still required whereas in routes/app/index.tsx it isn't.

`app._index.tsx` and `app/index.tsx` are different routes. The first is an index route for "app/", and will be rendered into the outlet of `app/index.tsx`. The second is the parent route itself.

`index.tsx` no longer has any semantic meaning for "index routes", but rather the node module resolution convention of "index modules".

`routes/app.tsx` and `routes/app/index.tsx` *are the same route*. You just moved it to a folder and made an `index.tsx` because that's how node module resolution looks for the module at `routes/app`.

If you want an index route, you use `_index`

| file name | route path | layout |
|-------|-----------|--------|
| `routes/index.tsx` | `/index` | root |
| `routes/_index.tsx` | `/` | root |
| `routes/app.tsx` | `/app` | root |
| `routes/app/index.tsx` | same route as above | root |
| `routes/app._index.tsx` | `/app` | `routes/app.tsx` or `routes/app/index.tsx` |
| `routes/app._index/index.tsx` | same route as above | `routes/app.tsx` or `routes/app/index.tsx` |

I'm expecting this is only confusing temporarily for people coming from the old convention, or people who aren't familiar with node module "index" resolution.

Again, only `_index` has meaning for route paths (it's the index route) while `index` has no meaning at all for route paths but regular module resolution.

ryanflorence:
https://remix.run/docs/en/v1/file-conventions/route-files-v2#folders-for-organization

Xiphe:
Awesome! I like how it's written in that section. Thanks for clarifying!

lennerd:
> @lennerd If there are still bugs, do you mind opening a new issue? It'll be much easier to get it prioritized that way.

@chaance Here you go: https://github.com/remix-run/remix/issues/5303

PR #5164, opened 2023-01-20
Docs: Add missing concerts.tsx layout in example

vegar:
The Nested URLs without Layout Nesting example shows how `concerts_.mine.tsx` opts out from the `concerts.tsx` layout. The url-route-layout table still reference the `concerts.tsx` layout, but the file is no longer part of the file structure. I guess it should be? Like in the previous example?

Closes: #

- [x] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -445,6 +445,7 @@
 - UsamaHameed
 - ValentinH
 - valerie-makes
+- vegar
 - veritem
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
This is already fixed by #5181

Issue #1757, opened 2022-02-01
[Bug]: App crash on fast browser navigation


axel-xitaso:
### What version of Remix are you using?

1.1.3

### What version of Node are you using? Minimum supported version is 14.

16.0.0

### Steps to Reproduce

1. Create a new Remix app with `npx create-remix@latest`.
2. Choose the Remix dev server.
3. Install dependencies using `yarn install`.
4. Next to `app/routes/index.tsx`, add `app/routes/other.tsx` with a component that simply renders `<>Other</>`
5. In `app/routes/index.tsx`, add `<Link to="/other">Other</Link>` somewhere, with `import { Link } from "remix";`.
6. Run your app using `yarn dev` or `yarn start`; the error happens for both dev and prod builds.
7. Open your browser and create a new browser tab (the bug happens with both Firefox and Chrome).
8. Go to your app's URL.
9. Click the link to `Other`.
10. Click the browser back button two times. The tab should now show your browser's default UI for new tabs.
11. Now very quickly click the browser's forward button twice.
12. The app should now crash. If it doesn't you didn't click the forward button fast enough.

### Expected Behavior

The page content should be displayed without any errors.

### Actual Behavior

In the simple repo above, the app crashes and there is the following stack trace in the dev console. In my real app in a production build, this issue then causes #1678, basically making the entire browser tab unresponsive.

```
components.js:470 Uncaught TypeError: Cannot read properties of undefined (reading 'meta')
    at Meta (components.js:470:21)
    at renderWithHooks (react-dom.development.js:14985:18)
    at mountIndeterminateComponent (react-dom.development.js:17811:13)
    at beginWork (react-dom.development.js:19049:16)
    at HTMLUnknownElement.callCallback2 (react-dom.development.js:3945:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:16)
    at invokeGuardedCallback (react-dom.development.js:4056:31)
    at beginWork$1 (react-dom.development.js:23964:7)
    at performUnitOfWork (react-dom.development.js:22776:12)
    at workLoopSync (react-dom.development.js:22707:5)
Meta @ components.js:470
renderWithHooks @ react-dom.development.js:14985
mountIndeterminateComponent @ react-dom.development.js:17811
beginWork @ react-dom.development.js:19049
callCallback2 @ react-dom.development.js:3945
invokeGuardedCallbackDev @ react-dom.development.js:3994
invokeGuardedCallback @ react-dom.development.js:4056
[...]

PR #5166, opened 2023-01-20
[Bug]: App crash on fast browser navigation or shows wrong url

n8agrin:
Reproduction case for issue: https://github.com/remix-run/remix/issues/1757

To run:

```
yarn bug-report-test
```

Failure is either an error where javascript complains about `default` missing from an object, or a url not matching its content. In the case of the test, we expect `/burgers` to contain the string `cheeseburgers`, but often only gets to the first route in history, `/`

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

* this is a test that reproduces the behavior in issue #1757

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -93,149 +93,17 @@ test("[description of what you expect it to do]", async ({ page }) => {
   await app.clickLink("/burgers");
   expect(await app.getHtml()).toMatch("cheeseburger");
 
-  await page.goBack({ waitUntil: "commit" });
-  await page.goBack({ waitUntil: "networkidle" });
-  await page.goForward({ waitUntil: "commit" });
-  await page.goForward({ waitUntil: "commit" });
-  expect(await app.getHtml()).not.toContain(
-    `Cannot destructure property 'default'`
-  );
+  let retry = 100;
 
-  await page.goBack({ waitUntil: "commit" });
-  await page.goBack({ waitUntil: "networkidle" });
-  await page.goForward({ waitUntil: "commit" });
-  await page.goForward({ waitUntil: "commit" });
-  expect(await app.getHtml()).not.toContain(
-    `Cannot destructure property 'default'`
-  );
-
-  await page.goBack({ waitUntil: "commit" });
-  await page.goBack({ waitUntil: "networkidle" });
-  await page.goForward({ waitUntil: "commit" });
-  await page.goForward({ waitUntil: "commit" });
-  expect(await app.getHtml()).not.toContain(
-    `Cannot destructure property 'default'`
-  );
-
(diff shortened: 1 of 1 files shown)


machour:
@n8agrin this PR seems to fail at reproducing the issue. 
Also you may want to use a for loop to help readability 🙏

n8agrin:
@machour thanks for taking a look. This fails locally for me consistently, let me try to get it failing for your ci

n8agrin:
@machour getting this to fail on CI was tricky. I have not been able to reproduce the case where javascript throws an exception, complaining `TypeError: Cannot destructure property 'default' of 'n[e]' as it is undefined.`

The bug requires transitioning from a url that is not the remix app under test, to a url in a remix app which fails to load the correct module for the page. This is typically done either by browsing back and forth quickly, or finding a rare state, where Remix issues a 300 level http redirect to a module that has not loaded yet. This is why I've added a navigation event to `https://remix.run` itself, to ensure the test clears the remix app under test's modules from memory.

To get this to fail on CI I took a different tact, which leverages another issue I've observed locally: when this error manifests, the url does not match the page content. This seems to fail consistently on ci. That said, occasionally the flaky test retries were able to overcome the particular check, requiring I add some retries in to force the failure.

The behavior this test is trying to assert is flaky by nature. It's difficult to capture it in a succinct regression test. I recommend checking out this branch locally and running the tests for yourself. If you'd like to see the TypeError exception, follow the directions I've left in the comments and you should trivially reproduce the issue.

Also happy to jump on a call with you to explain it.

This error is blocking our production app from being able to be tested in CI, causing us and our customers real pain.
`

machour:
Thank you so much for all the efforts trying to reproduce this issue @n8agrin.
Yes, this is definitely not easy to reproduce in CI.

I'm going to defer to @mcansh (our testing expert) or @brophdawg11 (our routing expert) on this one. 
Nothing much I can do on my side.

n8agrin:
@machour thank you! @mcansh @brophdawg11 I've separated the failure cases into two tests. Hopefully you are able to reproduce locally and see the errors described above. Please feel free to reach out and let me know how I can help.

n8agrin:
I still believe this is an issue. I will update the repro case against latest remix version

brophdawg11:
I was about to re-open since this is not yet fixed but since we the issue is open, I think it's OK for this to remain closed since we can still see this reproduction when we get to addressing that bug 👍

n8agrin:
@brophdawg11 coincidentally, yesterday I worked out a patch that fixes the underlying bug. I’ll post the diff tonight when I’m back at my keyboard

n8agrin:
@brophdawg11 PR is here https://github.com/remix-run/remix/pull/6409 (sorry for dup msgs, covering all channels)

PR #5177, opened 2023-01-22
docs: add information about throwing `redirect`s

andrelandgraf:
The current docs don't highlight that you can throw a redirect response. Also, it's not mentioned that throwing a `redirect` is different from throwing any other response (does not trigger `CatchBoundary`). The goal is to emphasize that throwing a redirect is kind of a special case.

### DIFF
--- a/docs/route/loader.md
+++ b/docs/route/loader.md
@@ -190,7 +190,7 @@ See also:
 
 ## Throwing Responses in Loaders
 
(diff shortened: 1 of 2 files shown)


machour (approved):
Thank you!

MichaelDeBoey:
@andrelandgraf Since this is a docs change, can you please rebase your branch onto `main` & set the base branch in this PR to `main` as well?

andrelandgraf:
Done! Thanks for the instructions @MichaelDeBoey!

chaance (approved):

PR #5179, opened 2023-01-22
chore(remix-dev): upgrade `cacache` to remove transitive dependency warnings

JakeGinnivan:
cacache v15 -> deprecated @npmcli/move-file@1.1.2

I didn't open an issue and instead just did the upgrade, the usage is really small and looks to be on the critical path for loading routes and caching them. Any issues with the upgrade i'd imagine would be pretty apparent in the tests.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -186,6 +186,7 @@
 - jacargentina
 - jacob-ebey
 - JacobParis
+- jakeginnivan
 - jakewtaylor
 - jamiebuilds
 - janhoogeveen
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (commented):
We can't update `cacache` as that would be a breaking change, because we support Node v14 and up (so also v15) completely & `cacache`@v16's `engines` field isn't compatible with that

```json
 "engines": {
    "node": "^12.13.0 || ^14.15.0 || >=16"
  },
```

JakeGinnivan:
Fair enough @MichaelDeBoey. Is there a v2 breaking changes backlog this could be added to so it isn't forgotten when v2 is done?

I guess the other alternative is remove the dependency but that's more work than i'd be up for picking up at. Will close

Issue #190, opened 2021-05-28
better message when people import fake built-ins?


ryanflorence:
If you `npm add crypto` you get a build error no matter what:

```
> remix build

Building Remix app in production mode...
 > node_modules/@remix-run/node/sessions/fileStorage.js:6:21: error: Could not resolve "crypto" (use "platform: 'node'" when building for node)
    6 │ var crypto = require('crypto');
      ╵                      ~~~~~~~~

 > node_modules/cookie-signature/index.js:5:21: error: Could not resolve "crypto" (use "platform: 'node'" when building for node)
    5 │ var crypto = require('crypto');
```

I'm not sure why our compiler blows up on this (haven't dug in yet) but I know we look at the app deps and ignore built-ins, but maybe we ought to just not allow using any node modules w/ the same name as a built-in node module?

PR #5222, opened 2023-01-23
fix(dev): exclude `node:`-prefixed built-ins from the browser build

pcattori:
Closes: #4544 

- [ ] Docs
- [x] Tests

## TODO
- [ ] can we get rid of the `fakeBuiltins` check introduced in https://github.com/remix-run/remix/issues/190 ?

### DIFF
--- a/packages/remix-dev/compiler/compileBrowser.ts
+++ b/packages/remix-dev/compiler/compileBrowser.ts
@@ -1,6 +1,6 @@
 import * as path from "path";
 import * as fse from "fs-extra";
-import { builtinModules as nodeBuiltins } from "module";
+import { builtinModules } from "module";
 import * as esbuild from "esbuild";
 import { NodeModulesPolyfillPlugin } from "@esbuild-plugins/node-modules-polyfill";
 import postcss from "postcss";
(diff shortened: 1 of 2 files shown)


sergiodxa (approved):
Will this also allows `import Buffer from "buffer"` to import `buffer` from the node_modules? Or is this just to consider `node:*`  as server-only imports?

MichaelDeBoey (approved):

packages/remix-dev/compiler/compileBrowser.ts
```diff
@@ -35,6 +35,14 @@ export type BrowserCompiler = {
   dispose: () => void;
 };
 
+let nodeBuiltins = Array.from(
+  new Set([
+    ...builtinModules,
+    // account for `node:`-prefixed built-ins like `"node:fs"`
+    ...builtinModules.map((x) => `node:${x}`),
+  ])
+);
```
mcansh, line 44: i'm sort of surprised that `builtInModules` doesn't already account for the `node:*` prefix tbh 🤷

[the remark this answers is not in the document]
mcansh: even better, they have `isBuiltin` now https://github.com/nodejs/node/pull/43396

MichaelDeBoey: It does, but not sure since what version that's included tbh

https://github.com/nodejs/node/blob/762a3a8ad925d56a12b43e0e7f7c811e93097784/lib/internal/modules/cjs/loader.js#L311

MichaelDeBoey: This was also introduced in v16.17 (see nodejs/node#43396), so we should account for values that can or can not have `node:`-prefix

A possible solution would be to map over all values, remove the `node:`-prefix, put that array in a Set so we have unique values, loop over the Set to add the `node:`-prefix again

pcattori:
This is blocked on getting better polyfills that account for `node:` prefixes. Currently, without polyfills for `node:`-prefix the app will have console errors.

MichaelDeBoey:
@pcattori #5274 unblocks this PR again 🎉

pcattori:
Superceded by #5773

Issue #5085, opened 2023-01-13
🗺️ Built-in PostCSS and Tailwind support


markdalgleish:
## Progress

- [x] RFC: https://github.com/remix-run/remix/discussions/5083
- [x] PR: https://github.com/remix-run/remix/pull/5229

ryanflorence:
@chaance to review

chaance:
Reviewed 🚢

PR #5229, opened 2023-01-24
feat: add built-in PostCSS and Tailwind support

markdalgleish:
Closes: #5085 

- [x] Docs
- [x] Tests

Long term these features are designed to be enabled by default, but for now they're behind feature flags.

These two features are designed to be completely independent from a Remix perspective:

- If you enable PostCSS support in isolation, it works as expected.
- If you enable Tailwind support in isolation, we set up PostCSS behind the scenes with the Tailwind plugin. Consumers don't need to know that PostCSS is being used at all. Note that PostCSS was already a dependency of Remix in order to support CSS Modules, so this strategy is an extension of that.
- If you enable both features at the same time, we ensure that the Tailwind PostCSS plugin is included behind the scenes if it's missing in your PostCSS config. This is so that you can use the built-in Tailwind support and then later add PostCSS without inadvertently breaking your Tailwind setup because you forgot to include the plugin.

Testing strategy:

Integration tests have been provided for both PostCSS and Tailwind support, in both cases being used with regular style sheets, CSS Modules, CSS side-effect imports and Vanilla Extract.


docs/guides/styling.md
```diff
@@ -400,7 +400,80 @@ export function links() {
...
+```
+
+Then include the `@tailwind` directives in your CSS. For example, you could create a `tailwind.css` file at the root of your app:
+
+```css filename=app/tailwind.css
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
+```
+
```
MichaelDeBoey, line 455: This step isn't necessary if we're not creating any custom classes & just using the Tailwind CLI.

I think we should remove it by default & add some section to explain that people should add this file if they want to add custom classes

markdalgleish: It's needed when using the built-in support since it's going through PostCSS, not the CLI. It needs to know where to insert the generated Tailwind CSS. Here are Tailwind's PostCSS setup docs for more details: https://tailwindcss.com/docs/installation/using-postcss

packages/remix-dev/compiler/plugins/cssFilePlugin.ts
```diff
@@ -15,31 +18,35 @@ function normalizePathSlashes(p: string) {
...
           write: false,
-          sourcemap: false,
+          sourcemap: Boolean(options.sourcemap && postcssProcessor), // We only need source maps if we're processing the CSS with PostCSS
           incremental: false,
           splitting: false,
           stdin: undefined,
           outfile: undefined,
-          outdir: outfile ? path.dirname(outfile) : outdir,
-          entryNames: assetNames,
+          outdir: config.assetsBuildDirectory,
```
markdalgleish, line 48: This was needed so the build is consistent between browser and server builds. With the original code, the server and client get different source maps which results in multiple, differing copies of the same styles being generated for a single Remix build. This is because the hash of the source map is in the generated CSS file since it contains a comment with the file name. (This was a tough one to track down!)

docs/guides/styling.md
```diff
@@ -400,7 +400,80 @@ export function links() {
...
+];
+```
+
+With this setup in place, you can also use [Tailwind's functions and directives][tailwind-functions-and-directives] anywhere in your CSS.
+
+Note that if you're also using Remix's [built-in PostCSS support](#built-in-postcss-support), the Tailwind PostCSS plugin will be automatically included if it's missing.
+
+### Manual Tailwind Integration
+
+It's also possible to use Tailwind without leveraging the built-in support by using the `tailwindcss` CLI directly.
```
chaance, line 476: Maybe we should also mention that it works as a postcss plugin and link to the section on postcss below?

docs/guides/styling.md
```diff
@@ -527,21 +600,75 @@ export const links: LinksFunction = () => {
...
+       autoprefixer: {},
+     },
+   };
+   ```
+
+If you're using [Vanilla Extract](#vanilla-extract), since it's already playing the role of CSS preprocessor, you may want to apply a different set of PostCSS plugins relative to other styles. To support this, you can export a function from `postcss.config.js` which is given a context object that lets you know when Remix is processing a Vanilla Extract file.
+
+```js filename=postcss.config.js
+module.exports = (ctx) => {
+  return ctx.remix?.vanillaExtract
```
chaance, line 645: `ctx.remix` >> this is awesome

packages/remix-dev/compiler/plugins/cssFilePlugin.ts
```diff
@@ -119,3 +129,33 @@ export function cssFilePlugin(options: {
...
+  postcssProcessor,
+  options,
+}: {
+  postcssProcessor: Processor;
+  options: CompileOptions;
+}): esbuild.Plugin {
+  return {
+    name: "postcss-plugin",
+    async setup(build) {
+      build.onLoad({ filter: /\.css$/, namespace: "file" }, async (args) => {
```
chaance, line 143: I know we're going to look at Sass in a separate proposal, but is there any harm in allowing `.scss` files to get picked up by this plugin? We could catch potential errors from postcss processing and let users know they need to configure postcss to parse and transform Sass.

packages/remix-dev/compiler/plugins/cssModulesPlugin.ts
```diff
@@ -17,14 +19,20 @@ interface PluginData {
...
-  rootDirectory: string;
   outputCss: boolean;
 }): Plugin => {
   return {
     name: pluginName,
     setup: async (build: PluginBuild) => {
+      let postcssPlugins = await loadPostcssPlugins({ config });
+
       build.onResolve(
         { filter: cssModulesFilter, namespace: "file" },
```
chaance, line 37: Same note here about `.scss`, I wonder if supporting sass modules would be this easy...

chaance (approved):
This looks awesome. A couple minor nits on docs and comments otherwise, but no blockers. Stoked for this 👍

docs/guides/styling.md
markdalgleish: Good point, it makes sense to call this out 👍

packages/remix-dev/compiler/plugins/cssFilePlugin.ts
markdalgleish: I think there's a bit of effort involved for this to work since you need a different parser for `.scss` files, and I'd want to have tests for it too. I'd prefer to do this as part of the Sass integration to ensure everything's working together correctly.

docs/guides/styling.md
```diff
@@ -400,7 +400,80 @@ export function links() {
...
+  },
+  // ...
+};
+```
+
+Then install Tailwind:
+
+```sh
+npm install -D tailwind
+```
```
rossipedia, line 427: Just FYI, [`tailwind`](https://www.npmjs.com/package/tailwind) isn't the right module here, it should probably be

```
npm install -D tailwindcss
```

zwily:
> 🤖 Hello there,
> 
> We just published version `1.12.0` which includes this pull request. If you'd like to take it for a test run please try it out and let us know what you think!
> 
> Thanks!

You appear to be mistaken, kind bot. :)

ryanflorence:
Ah, better go check out what's going on in meatspace

docs/guides/styling.md
machour: Got fixed by the man himself: https://github.com/remix-run/remix/pull/5252/files

mcansh:
> You appear to be mistaken, kind bot. 😀

the kind, but sometimes mistaken bot has received a stern talking to and promises to not do it again, i believe him. https://github.com/remix-run/remix/pull/5342

PR #4459, opened 2022-10-31
Feature/basename

yracnet:
I have implemented the feature described on the  https://github.com/remix-run/remix/discussions/2891.

This include the serverBasename attribute on remix.config.js for declare a prefix to URL
> serverBasename default value is ""
> fix deploy on path server (basename)
> fix the navegation in client side or server side components

PR #5236, opened 2023-01-24
feat(remix-dev): add `serverBasename` config option

yracnet:
I have implemented the feature described on the https://github.com/remix-run/remix/discussions/2891.

This include the serverBasename attribute on remix.config.js for declare a prefix to URL

> serverBasename default value is ""
> fix deploy on path server (basename)
> fix the navegation in client side or server side components

**clean branch from from https://github.com/remix-run/remix/pull/4459**

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -471,6 +471,7 @@
 - yomeshgupta
 - youbicode
 - youngvform
+- yracnet
 - zachdtaylor
 - zainfathoni
 - zhe
--- a/packages/remix-dev/__tests__/readConfig-test.ts
+++ b/packages/remix-dev/__tests__/readConfig-test.ts
@@ -60,6 +60,7 @@ describe("readConfig", () => {
             "path": "",
           },
         },
+        "serverBasename": "",
         "serverBuildPath": Any<String>,
         "serverBuildTarget": undefined,
         "serverBuildTargetEntryModule": "export * from \\"@remix-run/dev/server-build\\";",
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -92,6 +92,14 @@ export interface AppConfig {
   serverBuildPath?: string;
 
   /**
+   * Server Basename is a location path for app deploy on server.
+   *
+   * If omitted, the default location path is / (root directory)
+   * {@link ServerConfig.serverBasename}.
+   */
+  serverBasename?: string;
+
+  /**
    * The path to the browser build, relative to `remix.config.js`. Defaults to
    * "public/build".
    */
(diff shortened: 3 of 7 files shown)


kentcdodds:
Could we get an integration test for this please?

justinwaite:
Something I've struggled with Remix apps hosted under the same domain at my company is redirecting from loaders and actions to other apps. For example, say we have the following apps:

App 1: domain.com/app1
App 2: domain.com/app2
... etc.
Launchpad: domain.com/home

Now say that in App 2 I return a response that should redirect to `/home`. What happens currently is that remix tries to find the `/home` route within the App2 remix router manifest, does not find it, then renders a 404. 

This makes it difficult to do something like redirecting to a login page that lives in the "Launchpad" app. This PR should also handle this case if it doesn't. I know that the redirect logic within remix does handle external (non-remix) routes, but only if the domain is different. Now I want remix to check the domain + this serverBasepath.

justinwaite:
I think I've discovered that the redirect issue will need to be resolved in react-router. I'll try and put together a PR over there.

sissifly:
This feature is very urgently needed.

maxprilutskiy:
@justinwaite @kentcdodds @yracnet folks, any chance we can solve this any time soon? Any support needed with anything?

@justinwaite have you put together a PR you talked about or not just yet?

justinwaite:
> @justinwaite @kentcdodds @yracnet folks, any chance we can solve this any time soon? Any support needed with anything?
> 
> 
> 
> @justinwaite have you put together a PR you talked about or not just yet?

The redirect and Link changes should be tied to this bug: https://github.com/remix-run/react-router/issues/10052

Looks like these will be taken care of soon. Then Remix needs to be updated to allow a base name.

19Qingfeng:
I want to know what happened to this pr.

rolandjohann:
We need this feature, too and investigated into implementing it ourself. If this PR is stale, how is the process to take over the work? At least @dmbarry86 mentioned implementing the desired functionality via third party libs, we can invest into it as well.

MichaelDeBoey:
Hi @yracnet!

yracnet:
Hi, I have implemented a plugin in ViteJS that provides support for the server's basename. You can check it out at https://www.npmjs.com/package/vite-plugin-remix

maximilize:
Sad to see this PR gone, would be a helpful feature. What's the reason for not doing it the remix way, but to use vite?

silvenon:
The Vite plugin seems to be using Remix internals to hack around this. I would've love to see Remix support for this because it uses React Router under the hood, which supports the `basename` prop, so I'm a little skeptical that the base path problem can be solved cleanly outside of Remix.

Vite itself has a `base` option, but it only refers to assets. Perhaps now that Remix is moving to Vite that option can be used in the Remix Vite plugin somehow and used accordingly.

justinwaite:
> The Vite plugin seems to be using Remix internals to hack around this. I would've love to see Remix support for this because it uses React Router under the hood, which supports the `basename` prop, so I'm a little skeptical that the base path problem can be solved cleanly outside of Remix.
> 
> 
> 
> Vite itself has a `base` option, but it only refers to assets. Perhaps now that Remix is moving to Vite that option can be used in the Remix Vite plugin somehow and used accordingly.

This is what I would like to see as well, but perhaps we need to wait for vite support to stabilize more. I know that some changes are occurring with the asset paths per #8077

ahmedk743:
Is anyone still interested in getting this PR merged?

PR #5237, opened 2023-01-24
fix(express): Remove dependency on express (use vanilla http instead)

nicksrandall:
Continuation of: #3234

This PR removes express as a dependency from the "express" handler (using only vanilla `http` module). This allows the handler to be used by any "express like" router (like polka) or using `http.createServer` directly! This should improve remix's compatibility with many node based http server frameworks. 

## Example (using http.createServer & sirv)

```ts
import * as http from 'http';
import * as path from 'path';
import sirv from 'sirv';
import { createRequestHandler } from '@remix-run/express';

const remixHandler = createRequestHandler({ build: require(BUILD_DIR) });

const server = http.createServer(sirv("public", {
	  maxAge: 3600,
	  setHeaders(res, pathname) {
		    if (pathname.startsWith("/build/")) {
		      res.setHeader("Cache-Control", "public, max-age=31536000, immutable");
		    }
	  },
	  onNoMatch: remixHandler
}));

server.listen(process.env.PORT || 3000);
```

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

All tests pass.


chaance:
I'm struggling to understand the benefit of removing the `express` dependency from our `express` adapter. This adapter always assumes you want to use `express`, hence the name.

If you want an interop Node server, wouldn't it make more sense to have a separate adapter for that?

nicksrandall:
@chaance I agree that if this change were to land, the name of the module no longer makes sense. Maybe it would be best to introduce this has a net new module but then what happens to `@remix/express` since this this new module is a drop in replacement? Maybe just a re-export for compatibility sake?

The idea here is that nearly every HTTP framework for node is based on the "http" module. With this change, it will be much easier to adapt remix to work with *any* http framework (Fastify, Koa, Polka, ect...) and not just express.

chaance:
@nicksrandall I'll run the idea up the flagpole. I'm not sure if we want to introduce new adapters internally but it'd be a good candidate for community adapters if not 👍

chaance:
FYI I'm going to close the PR for now, only because I don't believe we will merge these changes into the existing Express adapter. The code will be a great reference should we decide to ship a new adapter -- and if not, we'd totally promote it if you wanted to ship it yourself as a standalone package given that it's pretty small!

nicksrandall:
@chaance Just so that my intention is clear... I think it would be really helpful for Remix to support a "vanilla" node adapter that only used node's standard library (i.e. the "http") package. This basic package would do the work to convert a Node Request (IncomingMessage) to a Remix Request and Node Response (ServerResponse) to a Remix Response. With this in place, the community would have everything it needs to create adapters for nearly *every* node http framework/router/ect...

The current `@remix-run/express` module *IS SO CLOSE* to already being this. The only two express only apis it uses is how it gets the request's protocol and how it appends headers. EVERYTHING else is already vanilla node.

This PR only changed those two function calls to be vanilla node (as you can see, it is a very small changeset).

I understand if Remix doesn't want to support new adapters but this would give the community what they need to create their own adapters -- which I think would be a big win for everybody.

nicksrandall:
Also, FWIW, the code changes here are copied directly from the express repo so the functionality is 100% exactly equivalent to what is in place now.

zackify:
I actually need this exact feature because I want to run remix in a NextJS app during migration. Using the plain http server is ideal. Thanks for your work @nicksrandall!

PR #5249, opened 2023-01-25
docs: cleanup docs

MichaelDeBoey:
Mainly removing TOCs when they're unnecessary + some smaller things I came across while looking at all the pages


machour (approved):
Neat 👌🏼

PR #5257, opened 2023-01-25
feat(remix-react): export `react-router-dom`'s `useMatch` hook

nicksrandall:
This is useful if you need to give a component some added functionality if a specific path is "active" or not.

### DIFF
--- a/.changeset/export-use-match.md
+++ b/.changeset/export-use-match.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Add `useMatch` re-export from `react-router-dom`
--- a/contributors.yml
+++ b/contributors.yml
@@ -375,6 +375,7 @@
 - nicksrandall
 - nickytonline
 - niconiahi
+- nicksrandall
 - nielsdb97
 - NikkiDelRosso
 - ninjaPixel
--- a/packages/remix-react/index.tsx
+++ b/packages/remix-react/index.tsx
@@ -24,6 +24,7 @@ export {
   useFormAction,
   useHref,
   useLocation,
+  useMatch,
   useNavigate,
   useNavigation,
   useNavigationType,


nicksrandall:
I'll note that I can import `useMatch from `react-router-dom` directly in my app but it's annoying to keep it up to date with exactly what remix depends on. If there is a version mis-match, my app breaks.

chaance:
We probably want to wrap `useMatch` instead of exporting it directly to include more of the context from `useMatches`. As it stands, you'd get more info from wrapping `useMatches` yourself. I think a more consistent implementation looks something like this:

```ts
export function useMatch<
  ParamKey extends ParamParseKey<Path>,
  Path extends string
>(pattern: PathPattern<Path> | Path): PathMatch<ParamKey> | null {
  let matches = useMatches();
  let location = useLocation();
  return React.useMemo(() => {
    let match = matchPath(pattern, location.pathname); // from react-router
    let found = matches.find((m) => m.pathname === match?.pathname);
    if (!found || !match) return null;
    return {
      ...match,
      ...found,
    };
  }, [matches, pattern, location.pathname]);
}
```

nicksrandall:
@chaance Would you accept this PR if it included the change you mentioned? 

The big reason that I need this is that I don't want to depend on "react-router-dom" or "react-router" directly. I just want to be able to import from "@remix-run/react";

With the change above I still have to add a new dependency because `matchPath` also isn't in `@remix-run/react`

chaance:
> @chaance Would you accept this PR if it included the change you mentioned?

Hang tight on making those changes for now. Let me run this by the team and get some input to make sure it's something we want to expose (and that my implementation makes sense). I understand your concern, but there's no issue with doing what you're doing in the mean time as the React Router dependency isn't going away.

brophdawg11 (approved):

PR #5259, opened 2023-01-25
feat(dev): HMR + Hot Data Revalidation

pcattori:
Exploring ~~the depths of our universe~~ HMR in Remix w/ @jacob-ebey

Depends on https://github.com/remix-run/react-router/pull/9996

## TODO

- [x] `reason` field for HMR updates to support contextual logs and debugging
- [x] Prevent preload from fetching duplicate copy of updated route
- [x] RR: rename new API -> `_internalSetRoutesAndRevalidate()` (naming tbd)
- [x] Gate all HMR stuff behind `unstable_dev` future flag
- [x] Restore dev server ports
- [x] Detect loader function declaration (`export function loader() {...}`) OR loader variable expression (`export let loader = () => {...}`)
- [x] Tests
- [x] move HMR from `<Scripts />` to `<LiveReload />` or dedicated component?
- [x] Resolve conflicts with `dev`
- [x] Rewrite AST transforms with babel
- [x] Handle MDX files in browser routes plugin
- [x] Changelog
- [x] Merge https://github.com/remix-run/react-router/pull/9996
- [x] Point to RR pre-release (not experimental release)

## Follow-on work
- [ ] Scan dependencies and pre-bundle them
- [ ] Optimize revalidations via granular route update APIs in RR
- [ ] Don't send HMR-only data (like loader hashes) to browser
- [ ] Improved error handling in compiler for `esbuild`'s `result.errors`
- [ ] Refactor compiler to have separate subcompilers for browser js and css


packages/remix-dev/package.json
```diff
@@ -44,6 +44,7 @@
     "fs-extra": "^10.0.0",
     "get-port": "^5.1.1",
     "gunzip-maybe": "^1.4.2",
+    "http-proxy": "^1.18.1",
```
pcattori, line 47: I think we can drop this

.changeset/hmr.md
```diff
@@ -0,0 +1,20 @@
...
+
+Features:
+- HMR for component and style changes
+- HDR when loaders for current route change
+
+Known limitations for MVP:
+- Only implemented for React via React Refresh
+- No `import.meta.hot` API exposed yet
+- Revalidates _all_ loaders on route when loader changes are detected
+- Loader changes do not account for imported dependencies changing
```
kentcdodds, line 20: Which of these are short-term limitations and which are really hard problems that will have to be solved longer-term (or never solved at all)?

pcattori: > - Revalidates _all_ loaders on route when loader changes are detected
> - Loader changes do not account for imported dependencies changing

👆 short-term, should be resolved before HMR is stabilized

> - No `import.meta.hot` API exposed yet

👆 we already prototyped a `import.meta.hot` API so my gut is that we can also do this short-term as needed

> - Only implemented for React via React Refresh

👆 If we have `import.meta.hot` API, then HMR should be implementable in user-land for any other rendering frameworks/libraries. Out-of-the-box HMR support for other renderers would come after we support that renderer (e.g. after `@remix-run/vue` exists or something like that).

kentcdodds: Sweet. Thanks!

kentcdodds: Also, I've noticed that HMR can be quite slow (several seconds after save). Is that expected? If you want to play around with a real world project that has this issue, check: https://github.com/epicweb-dev/rocket-rental

kentcdodds: I don't think my server is restarting. I'm doing the purge require cache approach instead

pcattori: > I don't think my server is restarting. I'm doing the purge require cache approach instead

Ah whoops. Saw `tsx watch` and made assumptions, but you are only watching for changes in `./index.js` and not `build/index.js`. Make sense.

Yea so to summarize, we can optimize HMR speed directly, but have limitations on HDR speed depending on how fast your app server responds with an up-to-date build hash via `__REMIX_ASSETS_MANIFEST` endpoint. Should be able to at least speed up HMR part as part of stabilizing this feature.

kentcdodds: That's great to hear! Thanks Pedro!

PR #5274, opened 2023-01-26
refactor(remix-dev): replace `@esbuild-plugins/node-modules-polyfill` with `esbuild-plugin-polyfill-node`

MichaelDeBoey:
As mentioned by @remorses (the maintainer of `@esbuild-plugins/node-modules-polyfill` in https://github.com/remorses/esbuild-plugins/pull/29#issuecomment-1497077035, we should use `esbuild-plugin-polyfill-node` instead

This will also remove some deprecation warnings when installing `@remix-run/dev` & fixes some imports of built-in Node packages

---

Closes #4299


MichaelDeBoey:
@nkrmr `node:` protocol imports are now supported by #5773 (instead of #5222)

@pcattori We do need this PR in order to close #4299 though

Since we have to wait for https://github.com/remorses/esbuild-plugins/pull/29 to be merged before we can continue, I'll make this a draft

jacob-ebey:
Thanks @MichaelDeBoey

TrySound:
@MichaelDeBoey still not good replacement
our cloudflare bundle grew up by 45kB!
All because of unnecessary added global polyfills

image

MichaelDeBoey:
@TrySound If you can come up with a better solution here, we're happy to accept PRs! 🙏

PR #5286, opened 2023-01-27
Added better detection for absolute urls in link component

lkwr:
This merge request uses the same improved detection as in remix-run/react-router#9994.


brophdawg11 (approved):
Thanks!

PR #5298, opened 2023-01-29
chore(templates): update cloudflare templates to support streaming and `defer` in entry.server.tsx

dgurns:
The `cloudflare-pages` and `cloudflare-workers` templates had been updated to React 18, but do not currently support streaming responses and `defer`.

This PR:
- Updates those templates to do so. Now a new project initialized with either of those templates should support `defer` out of the box. 
- Updates the docs for `defer` to include steps on how to upgrade an older Remix app to support `defer`, including necessary changes in `entry.client.tsx` and `entry.server.ts` that enable client-side React 18 hydration and streaming responses.
- Run `cloudflare-pages`/`cloudflare-workers` local dev servers on port 3000 to match the default `remix` server behavior 

I pulled the `entry.server.tsx` code from @jacob-ebey's [example project](https://github.com/jacob-ebey/remix-d1-northwind/blob/main/app/entry.server.tsx) which deploys to Cloudflare and tested that the updated `cloudflare-pages`/`cloudflare-workers` templates run successfully for local dev, including with `defer`.

Thanks for all your great work and very excited to begin playing around with `defer`.

### DIFF
--- a/docs/utils/defer.md
+++ b/docs/utils/defer.md
@@ -42,4 +42,12 @@ export const loader: LoaderFunction = async () => {
 };
 ```
 
+## How to update an older Remix app to use `defer`
+
+- Update all `@remix-run/*` packages to the latest version
+- Update `react` and `react-dom` to the latest version, `18` or above
+- If using TypeScript, update `@types/react` and `@types/react-dom` to the latest version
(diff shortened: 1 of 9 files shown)


MichaelDeBoey:
Closing this as we're already going to have these changes once #4600 is merged

dgurns:
Sounds good, only thought is it might still be worthwhile to have info in the `defer` README about how to upgrade an older Remix app to support it?

Even with the new implicit entry files, an older project would still have them explicitly defined and they likely wouldn't be set up for streaming yet.

Even better would be some kind of codemod that would automatically upgrade deps and entry files to support streaming/defer.

PR #5299, opened 2023-01-29
docs(route/should-revalidate): Fix shouldRevalidate typing

zayenz:
The type for the shouldRevalidate function is ShouldRevalidateFunction and not ShouldReloadFunction.


machour (approved):
Thank you soooo much 🙏🏼

PR #5307, opened 2023-01-30
docs: add CSS bundle tree-shaking issue warnings

markdalgleish:
We've already had a couple of consumers trip over this, so I'm adding some warnings to the docs while the esbuild issue still exists.


machour (approved):
👌🏼

Issue #5309, opened 2023-01-30
getDependenciesToBundle can't find dependencies of ESM package


nimaa77:
### What version of Remix are you using?

1.11.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

in `remix.config.js` file import `getDependenciesToBundle` package and pass an ESM package to it
```ts
const { getDependenciesToBundle } = require("@remix-run/dev");

/**
 * @type {import('@remix-run/dev').AppConfig}
 */
module.exports = {
  cacheDirectory: "./node_modules/.cache/remix",
  ignoredRouteFiles: ["**/.*", "**/*.css", "**/*.test.{js,jsx,ts,tsx}"],
  serverDependenciesToBundle: [...getDependenciesToBundle("@uiw/react-md-editor")],
};
```

run build script

### Expected Behavior

to bundle all ESM dependencies and of the passed in dependency the app and start the app without any errors

### Actual Behavior

Online Playground: https://stackblitz.com/edit/node-cytlj4?file=app/routes/index.tsx

during the build these get printed on the screen:

```
/Users/nima/projects/my-remix-app/node_modules/package.json does not exist
rehype is possibly an ESM only package and should be bundled with "serverDependenciesToBundle in remix.config.js.
rehype-prism-plus is possibly an ESM only package and should be bundled with "serverDependenciesToBundle in remix.config.js.
```

and when start the app using dev command

```
/Users/nima/projects/my-remix-app/node_modules/@uiw/react-markdown-preview/lib/index.js:13
var _reactMarkdown = _interopRequireDefault(require("react-markdown"));
                                            ^
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/nima/projects/my-remix-app/node_modules/react-markdown/index.js from /Users/nima/projects/my-remix-app/node_modules/@uiw/react-markdown-preview/lib/index.js not supported.
Instead change the require of /Users/nima/projects/my-remix-app/node_modules/react-markdown/index.js in /Users/nima/projects/my-remix-app/node_modules/@uiw/react-markdown-preview/lib/index.js to a dynamic import() which is available in all CommonJS modules.
```

lucaashrq:
Did you fix it? I got same issue.

0xtlt:
I still got this issue with the same configuration

lifeiscontent:
[...]

PR #5310, opened 2023-01-30
fix(remix-dev/compiler): fix `getPackageDependenciesRecursive`

nimaa77:
### Description

turns out `getDependenciesToBundle` has an issue that it can not find the package json of the passed dependency so it will skip the dependencies of that package,

that's why when you call it with a dependency you will get `my-remix-app/node_modules/package.json does not exist`

the logic of the function has some issues, if we want to get the package file

```js
  // packages/remix-dev/compiler/dependencies.ts
  
  let pkgPath = require.resolve(pkg);
  let lastIndexOfPackageName = pkgPath.lastIndexOf(pkg);
  if (lastIndexOfPackageName !== -1) {
    pkgPath = pkgPath.substring(0, lastIndexOfPackageName);
  }
  let pkgJson = path.join(pkgPath, "package.json");
  ```
  
`pkgPath.lastIndexOf(pkg)` gives us the index of the first character of the passed pkg name, but what developer intended to to was to get the last index of the pkg name so at next line when we take the substring from the pkgPath and pkgPath contains the package name as well

```js
// expected
"my-remix-app/node_modules/mdx-bundler"

// what we got
"my-remix-app/node_modules"
```

the fix for this simply would be:

```
 if (lastIndexOfPackageName !== -1) {
    pkgPath = pkgPath.substring(0, lastIndexOfPackageName + pkg.length);
  }
```

but changing this alone will not solve the issue, `require.resolve(pkg)` some times will throw

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in my-remix-app/node_modules/@babel/runtime/package.json
```

and all that we want to do is to get the package path, so we can use this package called  `resolve-package-path` to get the package path

playground link: https://stackblitz.com/edit/node-ddywpz?file=remix.config.js

- [ ] Docs
- [X] Tests

Testing Strategy:

This test covers this code: `integration/compiler-test.ts`

the way it works is by adding a new package to fixtures and this new package has another ESM package as its dependency so when the `getPackageDependenciesRecursive` reads the package.json it should the dependency 

this is the script that will run the tests `yarn test:integration integration/compiler-test.ts`

Closes: #5309

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -530,6 +530,4 @@
 - iamzee
 - TimonVS
 - yudai-nkt
-- mitchelldirt
-- krolebord
 - nimaa77
--- a/contributors.yml
+++ b/contributors.yml
@@ -530,3 +530,6 @@
 - iamzee
 - TimonVS
 - yudai-nkt
+- mitchelldirt
+- krolebord
+- nimaa77
--- a/integration/compiler-test.ts
+++ b/integration/compiler-test.ts
@@ -32,6 +32,7 @@ test.describe("compiler", () => {
               "esm-only-pkg",
               "esm-only-single-export",
               ...getDependenciesToBundle("esm-only-exports-pkg"),
+              ...getDependenciesToBundle("esm-only-pkg-with-dep"),
             ],
           };
         `,
@@ -74,6 +75,13 @@ test.describe("compiler", () => {
             return <div id="built-ins-polyfill">{path.join("test", "file.txt")}</div>;
           }
         `,
+        "app/routes/esm-only-pkg-with-dep.jsx": js`
+          import esmOnlyPkgWithDep from "esm-only-pkg-with-dep";
+
+          export default function EsmOnlyPkg() {
+            return <div id="esm-only-pkg-with-dep">{esmOnlyPkgWithDep}</div>;
+          }
+        `,
         "app/routes/esm-only-pkg.jsx": js`
           import esmOnlyPkg from "esm-only-pkg";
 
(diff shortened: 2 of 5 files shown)


nimaa77:
I think I need to return Array of regex patterns instead of array of package names from `getDependenciesToBundle`, something like (`return Array.from(aggregatedDeps).map((dep) => new RegExp('^' + dep + '.*'));`)

not sure tho

lukebro:
Ran into same issue, and too thought that `+ pkg.length` would be an appropriate solution.

Packages that don't have a "main" entry via exports/main is tough... Now what do you build? You have to parse your usage, just becomes a bigger mess.

Another thing to consider: the dependencies of a package need to resolve relative to that dependency, not the root of project. If a package depends on a different non-semver compatible package that ends up not being hoisted aka `node_modules/pkg-a/node_modules/pkg-b` you could tell remix to build `pkg-b`, and I'd assume it would resolve+build `node_modules/pkg-b` if it existed. It's an edge case but should be considered, as I've been bitten by this. 

I could be wrong here, but perhaps just doing the `+ pkg.length` should fix most use cases. If `require.resolve` fails to resolve due to a main entry being missing, you can continue to print the warning in the console so the dev needs to handle it manually.

MichaelDeBoey:
Hi @nimaa77!

nimaa77:
@MichaelDeBoey rebased and resolved

contributors.yml
```diff
@@ -530,3 +530,6 @@
 - iamzee
 - TimonVS
 - yudai-nkt
+- mitchelldirt
+- krolebord
```
MichaelDeBoey, line 534: Seems like something went wrong here when rebasing 🤔

```suggestion
```

MichaelDeBoey (changes requested):

nimaa77:
@MichaelDeBoey I applied your suggestion, thanks for the reveiew

MichaelDeBoey (changes requested):
@nimaa77 It seems like you have some conflicts to resolve
Could you please fix them?

Issue #5322, opened 2023-01-31
1.12 `ENOENT: no such file or directory` referencing `outes` instead of `routes`


sockthedev:
### What version of Remix are you using?

1.12.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

- Create a Remix instance selecting the "Remix App Server" option
- Run `remix dev`

The following error occurs;

```
TypeError: Cannot read properties of undefined (reading 'remixRequestHandlerPath')
    at requestHandler (/Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/@remix-run+server-runtime@1.12.0/node_modules/@remix-run/server-runtime/dist/server.js:38:94)
    at /Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/@remix-run+express@1.12.0_express@4.18.2/node_modules/@remix-run/express/dist/server.js:39:28
    at /Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/@remix-run+serve@1.12.0/node_modules/@remix-run/serve/dist/index.js:47:7
    at Layer.handle [as handle_request] (/Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/route.js:144:13)
    at next (/Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/route.js:140:7)
    at next (/Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/route.js:140:7)
    at next (/Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/route.js:140:7)
    at next (/Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/route.js:140:7)
    at next (/Users/sean/sockthedev/code/hivemind-tales/node_modules/.pnpm/express@4.18.2/node_modules/express/lib/router/route.js:140:7)
```

### Expected Behavior

The dev server executes as before.

### Actual Behavior

An error occurs on dev server load.

machour:
[...]

Issue #5324, opened 2023-01-31
Flat routes are broken on macOS (and probably Linux) in 1.12.0


haines:
### What version of Remix are you using?

1.12.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Set `future.v2_routeConvention = true`, attempt to build app on macOS.

### Expected Behavior

Routes are loaded, build succeeds

### Actual Behavior

Build breaks when loading routes

``` 
ENOENT: no such file or directory, open '<root>/app/outes/<route file>'
```

Note it is trying to load from `outes` not `routes`!

The bug was introduced in https://github.com/remix-run/remix/pull/5266. By splitting and rejoining the paths, it drops the leading `/`, converting `/Users/me/...` to  `Users/me/...`. However `appDirectory` still has a leading `/` so removing directory prefixes with `slice` drops one too many characters.

machour:
Maybe related to #5322 (the `outes` thingy)

haines:
Workaround using patch-package:

`patches/@remix-run+dev+1.12.0.patch`:

```patch
diff --git a/node_modules/@remix-run/dev/dist/config/flat-routes.js b/node_modules/@remix-run/dev/dist/config/flat-routes.js
index e3646a7..52c0688 100644
--- a/node_modules/@remix-run/dev/dist/config/flat-routes.js
+++ b/node_modules/@remix-run/dev/dist/config/flat-routes.js
@@ -33,7 +33,7 @@ function flatRoutes(appDirectory, ignoredFilePatterns) {
   // fast-glob will return posix paths even on windows
   // convert posix to os specific paths
   let routePathsForOS = routePaths.map(routePath => {
-    return path__default["default"].join(...routePath.split(path__default["default"].posix.sep));
+    return path__default["default"].normalize(routePath);
   });
   return flatRoutesUniversal(appDirectory, routePathsForOS);
 }
```

haines:
Duplicate of #5322

PR #5228, opened 2023-01-23
fix(remix-dev): flat route fixes and enhancements

mcansh:
Signed-off-by: Logan McAnsh <logan@mcan.sh>

Closes: #5170
Closes: #5261
Closes: #5303
Closes: #5322
Closes: #5324
Closes: #5327

- [ ] Docs
- [x] Tests

Testing Strategy:

updated current tests

PR #5327, opened 2023-01-31
fix: use `path.normalize` to convert POSIX paths to OS-specific paths

haines:
Closes: #5322

(from #5324): The build currently breaks when loading routes

``` 
ENOENT: no such file or directory, open '<root>/app/outes/<route file>'
```

Note it is trying to load from `outes` not `routes`!

The bug was introduced in https://github.com/remix-run/remix/pull/5266. By splitting and rejoining the paths, it drops the leading `/` on Unix-like systems, converting `/Users/me/...` to  `Users/me/...`. However `appDirectory` still has a leading `/` so removing directory prefixes with `slice` drops one too many characters.

This PR resolves the issue by using [`path.normalize`](https://nodejs.org/docs/latest-v14.x/api/path.html#path_path_normalize_path) to covert `/` to `\` on Windows, rather than splitting and rejoining the paths.

- [x] Tests

Testing Strategy:

This test covers this code (sort of... the test mirrors the production code so does not expose the bug): https://github.com/remix-run/remix/blob/ebfa6458af3f1e4ccf137200da05ac8f31df5a50/packages/remix-dev/__tests__/flat-routes-test.ts#L141

### DIFF
--- a/packages/remix-dev/config/flat-routes.ts
+++ b/packages/remix-dev/config/flat-routes.ts
@@ -28,7 +28,7 @@ export function flatRoutes(
   // fast-glob will return posix paths even on windows
   // convert posix to os specific paths
   let routePathsForOS = routePaths.map((routePath) => {
-    return path.join(...routePath.split(path.posix.sep));
+    return path.normalize(routePath);
   });
 
   return flatRoutesUniversal(appDirectory, routePathsForOS);
--- a/.changeset/slimy-insects-greet.md
+++ b/.changeset/slimy-insects-greet.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Fixed issue with `v2_routeConvention` on Unix-like systems
--- a/contributors.yml
+++ b/contributors.yml
@@ -165,6 +165,7 @@
 - gunners6518
 - gyx1000
 - hadizz
+- haines
 - hardingmatt
 - helderburato
 - HenryVogt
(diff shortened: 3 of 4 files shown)


machour:
Hey @haines, could you target the `dev` branch instead ? 🙏🏼

haines:
@machour done! Could you please re-approve the workflow run?

packages/remix-dev/config/flat-routes.ts
```diff
@@ -28,7 +28,7 @@ export function flatRoutes(
   // fast-glob will return posix paths even on windows
   // convert posix to os specific paths
   let routePathsForOS = routePaths.map((routePath) => {
-    return path.join(...routePath.split(path.posix.sep));
+    return path.normalize(routePath);
```
sockthedev, line 31: TIL 👍

machour:
As far as I see, this PR is modifying a test already passing.
Could you revert that and add a new test case to verify what you're fixing instead? 🤔

haines:
@machour the existing test just mirrors the production code, so the test has the same bug and it needs to be modified. The problem is that the test exercises the internal function `flatRoutesUniversal` (to avoid having to run against a real filesystem), but the bug occurs in `flatRoutes`.

I could rewrite the test to exercise `flatRoutes` instead, but that'd require writing the test files to the filesystem (in a temporary directory). What do you think?

MichaelDeBoey:
Superseded by #5228

PR #5343, opened 2023-02-01
Add server + route middleware functionality

brophdawg11:
Wire up Remix to new React Router middleware

Proposal: https://github.com/remix-run/react-router/discussions/9564
React Router PR: https://github.com/remix-run/react-router/pull/9975

_Note: CI will fail here until we can point to an experimental or prerelease version of the RR branch.  They pass locally if you build the RR branch into the Remix `node_modules` directory._


packages/remix-express/package.json
```diff
@@ -14,7 +14,8 @@
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
   "dependencies": {
-    "@remix-run/node": "1.12.0"
+    "@remix-run/node": "1.12.0",
+    "@remix-run/router": "1.3.1"
```
brophdawg11, line 18: Adapters will now include the router as a dependency so they can re-export  `MiddlewareContext`/`createMiddlewareContext` alongside their own `serverMiddleware` concepts

packages/remix-express/server.ts
```diff
@@ -38,14 +50,52 @@ export type RequestHandler = (
...
+
+      try {
+        await serverMiddleware({ request: req, response: res, context });
+        if (!response) {
+          // User never called next(), so doesn't need to do any post-processing
+          response = await callRemix();
+        }
+        if (!res.headersSent) {
+          await sendRemixResponse(res, response);
+        }
```
brophdawg11, line 90: A few things going on here:

* If the user never calls `context.next()` we automatically call the rest of the pipeline.  You cannot short circuit by skipping the call.  The only way to short circuit is to `throw`
* If the user actually responds via the express `Response` in the middleware (and thus `res.headersSent === true`) then we skip `sendRemixResponse()`.  This allows them to take full control if they want to.

packages/remix-express/server.ts
```diff
@@ -38,14 +50,52 @@ export type RequestHandler = (
...
   let handleRequest = createRemixRequestHandler(build, mode);
 
+  if (build.future.unstable_middleware && serverMiddleware) {
+    return async (
+      req: express.Request,
+      res: express.Response,
+      next: express.NextFunction
+    ) => {
+      let response: NodeResponse | undefined;
+      let context = createMiddlewareStore();
```
brophdawg11, line 70: It's worth noting that at the moment, since our context is keyed by object instances for strong-typing, users can run into cross-boundary CJS/ESM instance issues when using `serverMiddleware`.  This is solveable by either (1) bundling your entire server via the `remix.config.js` `server` option, or (2) potentially using import maps in your application `package.json` (untested)

packages/remix-server-runtime/routeModules.ts
```diff
@@ -68,11 +70,32 @@ export interface LinksFunction {
...
+export interface ActionFunctionWithMiddleware {
+  (args: ActionArgsWithMiddleware): DataFunctionReturnValue;
+}
+
+export interface LoaderFunctionWithMiddleware {
+  (args: LoaderArgsWithMiddleware): DataFunctionReturnValue;
+}
+
+export interface MiddlewareFunction {
+  (args: DataFunctionArgsWithMiddleware): DataFunctionReturnValue;
```
brophdawg11, line 98: Let's bikeshed these names a bit - should they be `LoaderFunction_V2`, etc.?

packages/remix-server-runtime/server.ts
```diff
@@ -75,15 +85,19 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
...
-        loadContext
+        loadContext,
+        unstable_middleware ? middlewareContext : undefined
       );
 
-      if (build.entry.module.handleDataRequest) {
+      if (
+        unstable_middleware !== true &&
+        build.entry.module.handleDataRequest
+      ) {
```
brophdawg11, line 100: Just like we will now ignore `getLoadContext` when you opt into middleware, we will also ignore `handleDataRequest` since you can operate on responses via the `serverMiddleware`.  If folks need to decipher between document and data requests they should be able to do that via looking at `response` headers or the `_data` query param

packages/remix-express/server.ts
```diff
@@ -38,14 +50,59 @@ export type RequestHandler = (
...
+            // User never called next(), so doesn't need to do any post-processing
+            response = await callRemix();
+          }
+          if (!res.headersSent) {
+            await sendRemixResponse(res, response);
+          }
+        } else {
+          let request = createRemixRequest(req, res);
+          response = (await handleRequest(request)) as NodeResponse;
+          await sendRemixResponse(res, response);
```
brophdawg11, line 96: If you turn on `unstable_middleware` and don't provide a `serverMiddleware` we just call through directly without any concept of `getLoadContext`

packages/remix-express/server.ts
```diff
@@ -38,14 +50,59 @@ export type RequestHandler = (
...
+
+          await serverMiddleware({ request: req, response: res, context });
+
+          if (!response) {
+            // User never called next(), so doesn't need to do any post-processing
+            response = await callRemix();
+          }
+          if (!res.headersSent) {
+            await sendRemixResponse(res, response);
+          }
```
brophdawg11, line 92: A few things going on here:

* If the user never calls `context.next()` we automatically call the rest of the pipeline.  You cannot short circuit by skipping the call.  The only way to short circuit is to `throw`
* If the user actually responds via the express `Response` in the middleware (and thus `res.headersSent === true`) then we skip `sendRemixResponse()`.  This allows them to take full control if they want to.

integration/middleware-test.ts
```diff
@@ -0,0 +1,396 @@
...
+    });
+
+    test("Allows middleware to alter loader responses (data requests)", async () => {
+      let response;
+
+      response = await fixture.requestData(
+        "/parent-response",
+        "routes/parent-response"
+      );
+      expect(response.headers.get("x-count")).toBe("2");
```
jamesarosen, line 326: These tests are OK, but I think you could be more explicit than a number. What about something like this?
```
expect(response.headers.get("x-processed-by")).toBe("parent")
…
expect(response.headers.get("x-processed-by")).toBe("parent,child")
…
expect(response.headers.get("x-processed-by")).toBe("parent,child,grandchild")
```

That makes it clearer which middleware are being invoked for each request.

brophdawg11:
Closing temporarily - we'll be revisiting shortly after v2 releases

pspeter3:
Now that V2 is out, is there a place to track this work?

brophdawg11:
Yep!  These are on the [Roadmap](https://github.com/orgs/remix-run/projects/5/views/1) as https://github.com/remix-run/remix/issues/7643 and https://github.com/remix-run/remix/issues/7645

UglyJoevIne (approved):

Issue #5345, opened 2023-02-02
Loader header are null if the loader uses defer


dpobel:
### What version of Remix are you using?

1.11.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

I followed https://remix.run/docs/en/v1/route/headers and https://remix.run/docs/en/v1/guides/streaming but it looks like retrieving the headers sent by `loader` from `headers` does not work when the `loader` uses a `defer` (with `json` it works perfectly)

To reproduce use the following route:

```tsx
import React from "react";
import { defer, json } from "@remix-run/node";

export const loader = async ({ params, context }) => {
  // with json, no problem
  // return json(
  // with defer, in `headers`, `loaderHeaders.get("x-custom-header-from-loader")`
  // returns null.
  return defer(
    { whatever: "stuff" },
    {
      headers: {
        "x-custom-header-from-loader": "value from loader",
      },
    }
  );
};

export const headers = ({ loaderHeaders }) => {
  console.log(
    "Header from loader",
    loaderHeaders.get("x-custom-header-from-loader")
  );
  return {
    "x-customer-header": loaderHeaders.get("x-custom-header-from-loader"),
  };
};

export default function BugHeader() {
  return <p>Bug header?</p>;
}
```

### Expected Behavior

in `headers`, `loaderHeaders.get("myheader")` should return the value set in `loader`

### Actual Behavior

it seems `loaderHeaders.get("myheader")` always return `null`

zackify:
can confirm, just check the browser and no response headers are sent. Switch to json, and they are

brophdawg11:
This should be fixed in `1.14.0`

PR #5346, opened 2023-02-02
BUG: Loader header are null if the loader uses defer

dpobel:
Test case for #5345


brophdawg11:
Thanks! Brought these over into #5627

PR #5358, opened 2023-02-03
fix(remix-dev): update tailwind config detection to also look for cjs

mcansh:
when using `type: "module"` in package.json you need to make any commonjs files have the `.cjs` extension however `require.resolve` won't look for them 🙈

note: tailwindcss doesn't currently support mjs so that extension is intentionally missing.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/compiler/utils/postcss.ts
+++ b/packages/remix-dev/compiler/utils/postcss.ts
@@ -131,7 +131,7 @@ async function loadTailwindPlugin(
 
     // Load Tailwind from the project directory
     tailwindPath = require.resolve("tailwindcss", { paths: [rootDirectory] });
-  } catch (err) {
+  } catch {
     // If they don't have a Tailwind config or Tailwind installed, just ignore it.
     return null;
   }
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (approved):

packages/remix-dev/compiler/utils/postcss.ts
```diff
@@ -119,12 +120,17 @@ async function loadTailwindPlugin(
...
+      tailwindConfigExtensions
+    );
+    if (!tailwindConfig) throw new Error("No Tailwind config found");
 
     // Load Tailwind from the project directory
-    tailwindPath = require.resolve("tailwindcss", {
-      paths: [rootDirectory],
-    });
+    tailwindPath = require.resolve("tailwindcss", { paths: [rootDirectory] });
   } catch (err) {
```
mcansh, line 134: we're not doing anything with the caught error.

```suggestion
  } catch {
```

markdalgleish (approved):

PR #5367, opened 2023-02-05
 docs(file-conventions/route-files-v2): Fix incorrect routes path

not-ivy:
Fixed incorrect route path in file conventions > route files v2 > folders for organization section


machour (approved):
Thank you!

PR #5390, opened 2023-02-07
Added better detection for absolute urls in link component

brophdawg11:
Wrapping up https://github.com/remix-run/remix/pull/5286 + adding changeset

### DIFF
--- a/.changeset/smart-insects-melt.md
+++ b/.changeset/smart-insects-melt.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (approved):

kiliman:
BTW: another way to test local vs external URLs is via the `URL` object itself
```ts
let u = new URL('https://github.com/remix-run', 'http://localhost')
let isExternal = u.origin !== 'http://localhost'
```

brophdawg11:
That works for _external_ URLs, and is what react-router [does today](https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L435) - but this is specifically checking for _absolute_ URL patterns.

PR #5398, opened 2023-02-08
chore: sync `FutureConfig` interface

mcansh:
(cherry picked from commit 2d28768b17a0767c2723625c5acec51a02ec0d22)

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -55,7 +55,7 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
     let { unstable_dev } = build.future;
     if (
       mode === "development" &&
-      unstable_dev !== false &&
+      typeof unstable_dev !== "boolean" &&
       url.pathname ===
         (unstable_dev.remixRequestHandlerPath ?? "") +
           "/__REMIX_ASSETS_MANIFEST"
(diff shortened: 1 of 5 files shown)


packages/remix-server-runtime/server.ts
```diff
@@ -55,7 +55,7 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
     let { unstable_dev } = build.future;
     if (
       mode === "development" &&
-      unstable_dev !== false &&
+      typeof unstable_dev === "object" &&
```
pcattori, line 58: Since `object` is vague (arrays, `Date`s, and a bunch of stuff has `typeof` set to `object`), let's check for `boolean` instead:

```
typeof unstable_dev !== 'boolean' &&
```

packages/remix-dev/config.ts
```diff
@@ -32,12 +32,12 @@ export type ServerBuildTarget =
...
 export type ServerPlatform = "node" | "neutral";
 
-type Dev = {
+interface Dev {
   port?: number;
   appServerPort?: number;
   remixRequestHandlerPath?: string;
   rebuildPollIntervalMs?: number;
-};
+}
```
pcattori, line 40: Can we keep this as a `type` rather than an `interface` so that it can't be accidentally extended?

`interface`s used to be preferred since they were more performant, but that stopped being the case a while ago 😅 so now I think it's nice to prefer `type` and only use `interface` if you need it

pcattori (changes requested):

pcattori (approved):

PR #5425, opened 2023-02-10
feat: support `@cloudflare/workers-types` v4

lkwr:
Update the Cloudflare Workers types to the latest major.

This solve dependency installing warnings when using `@cloudflare/workers-types@^4` with remix.
There was no peer dependency for version 4 in the cloudflare packages.


.changeset/cuddly-cameras-collect.md
```diff
@@ -0,0 +1,7 @@
+---
+"@remix-run/cloudflare": patch
+"@remix-run/cloudflare-pages": patch
+"@remix-run/cloudflare-workers": patch
```
MichaelDeBoey, line 4:
```suggestion
"@remix-run/cloudflare": minor
"@remix-run/cloudflare-pages": minor
"@remix-run/cloudflare-workers": minor
```

.changeset/cuddly-cameras-collect.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+chore: update `@cloudflare/workers-types` to v4
```
MichaelDeBoey, line 7:
```suggestion
feat: support `@cloudflare/workers-types` v4
```

MichaelDeBoey (changes requested):

zhuhaow:
If possible, we should switch to v4 new importable types https://github.com/cloudflare/workerd/tree/main/npm/workers-types#importable-types and drop support of v2/v3.

The current way of referencing `@cloudflare/workers-types` doesn't work at all. Just create a new remix project with `create-remix`, run `pnpm typecheck` and it fails.

So this dependency version bump is probably not a change but a fix.

MichaelDeBoey:
As this is a breaking change, we should do that in the next major version

I created #5732, which targets `v2` branch

MichaelDeBoey:
Implemented by #5732

Issue #5445, opened 2023-02-13
HTTP calls to an invalid URL exposes an stack trace


ronnylt:
### What version of Remix are you using?

v1.12.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

```
http --follow --all https://www.shopify.com/\?_data\=rootwew
HTTP/1.1 403 Forbidden
CF-Cache-Status: BYPASS
CF-RAY: 798e74c86c3c11b3-BCN
Connection: keep-alive
Content-Encoding: gzip
Content-Type: application/json; charset=utf-8
Date: Mon, 13 Feb 2023 15:04:23 GMT
NEL: {"success_fraction":0.01,"report_to":"cf-nel","max_age":604800}
Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=SBSW8rub9EdDe5si331mCz8bi913VPD3DCpBDnmrkUeR%2B1VXboNLlevhnyeC%2Ff996saEKTIlvyASR05GlrO2AIVtF7rYIjgwezqXmqW3wCJceN3KdBxFmEOuXC1WMPe9gQ%3D%3D"}],"group":"cf-nel","max_age":604800}
Server: cloudflare
Server-Timing: cfRequestDuration;dur=236.999989
Strict-Transport-Security: max-age=15552000; includeSubDomains; preload
Transfer-Encoding: chunked
Vary: Accept-Encoding
X-Content-Type-Options: nosniff
X-Dc: gcp-us-east1
X-Request-ID: 5ddd5bc5-80f3-4e5f-a254-b6bfa73488fb
alt-svc: h3=":443"; ma=86400, h3-29=":443"; ma=86400
x-remix-error: yes

{
    "message": "Unexpected Server Error",
    "stack": "Error: Unexpected Server Error\n    at handleDataRequestRR (/app/remix/node_modules/@remix-run/server-runtime/dist/server.js:128:31)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at requestHandler (/app/remix/node_modules/@remix-run/server-runtime/dist/server.js:55:18)\n    at Object.handler (/app/remix/node_modules/@remix-run/express/dist/server.js:39:22)"
}

```

### Expected Behavior

Do not expose an stack trace in production. 

### Actual Behavior

Extract trace dumped in the response. Internal server structure exposed.

[...]

PR #5446, opened 2023-02-13
fix(server-runtime) Data request call to an invalid routes do not catch error and exposes a stack trace

ronnylt:
Closes: https://github.com/remix-run/remix/issues/5445

- [ ] Docs
- [x] Tests

Data request call to an invalid routes do not catch error and exposes a stack trace in production.


brophdawg11:
Handled in #5541

Issue #5466, opened 2023-02-15
Type is not assignable to type `ServerBuild` with `exactOptionalPropertyTypes: true`


aaronadamsCA:
### What version of Remix are you using?

1.12.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

`server.ts`

```ts
import type { RequestHandler } from "@remix-run/cloudflare-pages";
import { createPagesFunctionHandler } from "@remix-run/cloudflare-pages";
import * as build from "@remix-run/dev/server-build";

export const onRequest: RequestHandler = createPagesFunctionHandler({
  build,
  getLoadContext: (context) => context,
  mode: process.env["NODE_ENV"] ?? "production",
});
```

### Expected Behavior

Works normally.

### Actual Behavior

```
Type 'typeof import("node_modules/@remix-run/dev/server-build")' is not assignable to type 'ServerBuild' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  Types of property 'dev' are incompatible.
    Type '{ liveReloadPort: number; } | undefined' is not assignable to type '{ liveReloadPort: number; }'.
      Type 'undefined' is not assignable to type '{ liveReloadPort: number; }'. ts(2375)

worker.d.ts(13, 5): The expected type comes from property 'build' which is declared here on type 'createPagesFunctionHandlerParams<any>'
```

This type mismatch was introduced in #5133, I can submit a PR now.

aaronadamsCA:
Not stale, PR in review.

pcattori:
#6724 removed the need to pass dev-only fields in `ServerBuild`, so that should address this too.

PR #5467, opened 2023-02-15
fix(remix-server-runtime): fix `ServerBuild` compatibility with `exactOptionalPropertyTypes` TS setting

aaronadamsCA:
Closes: #5466

- [ ] Docs
- [ ] Tests

Testing Strategy: None - I don't think there are currently any tests for this, and it's a very minor compat change, whereas new tests would be a much larger change.


.changeset/rude-beers-promise.md
```diff
@@ -0,0 +1,7 @@
+---
+"remix": patch
+"@remix-run/serve": patch
```
MichaelDeBoey, line 3:
```suggestion
```

.changeset/rude-beers-promise.md
```diff
@@ -0,0 +1,7 @@
...
+"@remix-run/server-runtime": patch
+---
+
+Fix `ServerBuild` type for compatibility with TypeScript exact optional property types.
```
MichaelDeBoey, line 7:
```suggestion
Fix `ServerBuild` compatibility with `exactOptionalPropertyTypes` TypeScript setting
```

MichaelDeBoey (changes requested):
@aaronadamsCA I guess this would not be the only place we have an optional property, so we should probably update some other types as well in that case

Don't know if this is something we want to do for all our types though 🤔
I once saw an ESLint enforcing this, but can't remember which one it was though 😕

mcansh (approved):

.changeset/rude-beers-promise.md
mcansh: remix and remix-serve depend on server-runtime iirc, so should be in here imo. on the other hand all packages are fixed on each other for releasing, so it's fine having just one 🤷

aaronadamsCA:
> @aaronadamsCA I guess this would not be the only place we have an optional property, so we should probably update some other types as well in that case
> 
> Don't know if this is something we want to do for all our types though 🤔 I once saw an ESLint enforcing this, but can't remember which one it was though 😕

You really only need it where you have an optional pass-through. IMO they can be adjusted one at a time as/if people run into challenges.

MichaelDeBoey:
Hi @aaronadamsCA!

MichaelDeBoey (approved):

brophdawg11:
I'd like to get @pcattori's eyes on this since it's both TS and dev server related

packages/remix-server-runtime/build.ts
```diff
@@ -15,7 +15,7 @@ export interface ServerBuild {
   publicPath: string;
   assetsBuildDirectory: string;
   future: FutureConfig;
-  dev?: { port: number };
```
pcattori, line 18: The intention is to allow omission of `dev`, not to let it be explicitly set to `undefined`. From what I can tell, including `| undefined` is not the semantics we want here.

Where is `dev` getting set to `undefined` explicitly? I think a better solution would be to make sure its omitted there instead.

pcattori (changes requested):

packages/remix-server-runtime/build.ts
aaronadamsCA: The problem originates here:

https://github.com/remix-run/remix/blob/3e589152bc717d04e2054c31bea5a1056080d4b9/packages/remix-dev/server-build.ts#L14

Because `ServerBuild["dev"]` is an optional property, the type of `dev` must be widened to include `undefined`. You could change this type to `NonNullable<ServerBuild["dev"]>`, but that seems like it might be less correct than the current change.

pcattori:
`dev` field has been removed in #6724 , obsoleting this PR

PR #5472, opened 2023-02-15
templates(cloudflare-pages,netlify): add `immutable` to `Cache-Control` header

appden:
This is a minor improvement to both the Netlify and Cloudflare Pages templates to include the `immutable` directive in the `Cache-Control` header for the hashed build assets. This removes the `s-maxage` for those as well since they are redundant to the `max-age` given they are equal. These now exactly match `Cache-Control` headers for hashes assets in the Deno, Express, and fly.io (via `remix-serve`) templates.

See `immutable` docs here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#immutable

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -33,6 +33,7 @@
 - AntoninBeaufort
 - anubra266
 - apeltop
+- appden
 - Aprillion
 - arange
 - archwebio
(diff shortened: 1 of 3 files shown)


sergiodxa (approved):

appden:
Hi, this was approved a couple weeks ago, so kindly following up to see if it can be merged.

MichaelDeBoey (approved):

mcansh (approved):

PR #5478, opened 2023-02-16
fix(dev): compiler not logging errors on initial compile

marwan38:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

Ensure an error is throwing during the initial compile. In my case, add an import to an export that doesn't exist in `routes/root.tsx`.

```typescript
// routes/root.tsx
import BadImport from './nonExistantFile';
```


marwan38:
This is a duplicate of #5441.

PR #5479, opened 2023-02-16
feat(dev): Add support for `serverBundles` config to output multiple server bundles

TooTallNate:
This adds the ability to split up the server-side build into multiple bundles via a new config property: `serverBundles`.

Example:

```js
  serverBundles: [
    { serverBuildPath: 'build/build1.js', routes: ['routes/index', 'routes/about'] },
    { serverBuildPath: 'build/build2.js', routes: ['routes/generate', 'routes/error'] }
  ],
```

When `serverBundles` is defined, it takes precedence over the top-level `serverBuildPath` property, since each "bundle" entry can define its output file path.

Related to Discussion: [#5405](https://github.com/remix-run/remix/discussions/5405#discussioncomment-4990250)

- [ ] Docs
- [ ] Tests

I will add docs / tests if this proposal is agreed upon.

Testing Strategy:

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -42,6 +42,11 @@ type Dev = {
   rebuildPollIntervalMs?: number;
 };
 
+interface ServerBundle {
+  serverBuildPath: string;
+  routes: string[];
+}
+
 interface FutureConfig {
   unstable_cssModules: boolean;
   unstable_cssSideEffectImports: boolean;
@@ -182,6 +187,12 @@ export interface AppConfig {
    * The platform the server build is targeting. Defaults to "node".
    */
   serverPlatform?: ServerPlatform;
+
+  /**
+   * Configuration of server bundles to produce. If this is defined then the
+   * top-level `serverBuildPath` value is ignored.
+   */
+  serverBundles?: ServerBundle[];
 
   /**
    * A list of filenames or a glob patterns to match files in the `app/routes`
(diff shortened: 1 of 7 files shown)


packages/remix-dev/config.ts
```diff
@@ -329,6 +338,15 @@ export interface RemixConfig {
    */
   serverPlatform: ServerPlatform;
 
+  /**
+   * Configuration of server bundles to produce. If this is defined then the
+   * top-level `serverBuildPath` value is ignored.
+   */
+  serverBundles?: {
+    serverBuildPath: string;
+    routes: RouteManifest;
```
tom-sherman, line 347: Routes should for sure accept some kind of pattern IMO, maybe a blob? Alternatively, for maximum flexibility, we could accept a function here. eg.

```js
serverBundles: (path) => path.startsWith("app/") ? "build/logged-in.js" : "build/logged-out.js",
```

That would also mean Remix wouldn't have to implement the matching internally, just leave it up to the user to return a build path for each route path.

tom-sherman:
Just to check, are the route paths file paths or pathnames? IMO they really should be path names, as this better fits the mental model of "route splitting".

TooTallNate:
They'd be path names since they're "normalized". Otherwise, the user would need to handle all the different file naming conventions.

packages/remix-dev/config.ts
TooTallNate: That's an interesting idea. I think the function would be defined on `routes` instead though, and return a boolean. I imagine there might end up being more properties on the bundle object that would affect the output (`server`, `serverBuildTarget`, etc) which would make your syntax not quite work.

tom-sherman: That could be done by returning an object from the function.

PR #5498, opened 2023-02-17
Link Component — renders an absolute URL as a relative href

thomasverleye:
- [ ] Docs
- [x] Tests

Testing Strategy:

run `yarn bug-report-test `


brophdawg11:
Thanks @thomasverleye!  I confirmed this is fixed if I update Remix to use `react-router-dom@6.8.2-pre.1` which includes your fix over there.  Since we've got it tested over in the router I'm going to close this out 👍

Issue #5509, opened 2023-02-18
Remove updating TSConfig file with suggested compiler options


SeanGroff:
### What version of Remix are you using?

1.13.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Remix "suggests" the following TypeScript compiler options:

```
  allowJs: true,
  forceConsistentCasingInFileNames: true,
  lib: ["DOM", "DOM.Iterable", "ES2019"],
  strict: true,
  target: "ES2019",
```

The Remix dev server will update your project's tsconfig file if any of the above compiler options are missing.

To reproduce, create a tsconfig file in your own Remix project and exclude from the config file any of the following compiler options:

```
  allowJs: true,
  forceConsistentCasingInFileNames: true,
  lib: ["DOM", "DOM.Iterable", "ES2019"],
  strict: true,
  target: "ES2019",
```

Run `npm run dev` and any of these compiler options that are excluded from your Remix project will be added to your project's TSConfig file. You can of course undo the change via git, but every time `npm run dev` is run, the compiler options above will be added back to your config.

### Expected Behavior

Running the Remix dev server should never update my project's tsconfig file with the compiler options that Remix **suggests** .

### Actual Behavior

Remix dev server will update my projects tsconfig file if any of the compiler options Remix **suggests** (see below) are missing from the config.

```
  allowJs: true,
  forceConsistentCasingInFileNames: true,
  lib: ["DOM", "DOM.Iterable", "ES2019"],
  strict: true,
  target: "ES2019",
```

PR #5510, opened 2023-02-18
fix(remix-dev): stop automatically updating `tsconfig.json`

SeanGroff:
Closes: #5509 

- [x] Docs
- [x] Tests

Testing Strategy:

I opened up my mac machine and ran `npm run dev`
I didn't have the tsconfig compiler option `strict` specified. The Remix dev package updates my tsconfig file by adding the compiler option `strict: true`

Here's a discussion I had on Twitter about this issue: https://twitter.com/_SeanGroff/status/1626637132176424963?s=20

Updating the TSConfig file shouldn't be the default behavior. The user should have to pass a `--fix` flag to opt into `$ remix dev` automatically updating the TSConfig file.


garand:
Interesting post by Wes Bos regarding this behavior... he liked it. https://twitter.com/wesbos/status/1627707679840915456

Maybe it would make sense to notify the user and give them the option to update their config?

SeanGroff:
While my current PR is an improvement, I think we can iterate on this solution to strike a happy middle ground between ease of use and being too forceful:

💡We keep the "Mandatory/Required" TSConfig compiler suggestions and "Recommended" compiler suggestions. However, we **no longer update the TSConfig. We display our suggestions in the CLI**. This provides the developer the freedom to configure the TSConfig themselves. If issues arise, maybe consider the suggestions from Remix 😀. Then we add a flag to the Remix Config that will disable the TSConfig suggestions in the terminal if the developer prefers. 

This should also be documented, depending on the direction we decide to go.

It's important to keep in mind the scope of this PR as well. This refers to the Remix Dev server, NOT the Create Remix App generator.

SeanGroff:
We could also prompt the developer with these changes in the CLI and they can provide input `Y/n` to apply the changes to the TSConfig. 

My worry here is in a team environment. A developer may not know the full implications of this change. They submit a PR, the CI fails or a Senior engineer asks them to revert their TSConfig changes. 

On dev teams, updating the tooling should be intentional and planned out. Preferably decoupled from any feature or bugfix work.

chaance (changes requested):
I'd suggest that we keep the warning but allowing users to opt-in to automatic changes with `--fix` or something. That said, the warning should also be dismiss-able. Probably beyond the scope for this PR, but if we can leave the log in place as a warning for now I think that's a step in the right direction.

SeanGroff:
> I'd suggest that we keep the warning but allowing users to opt-in to automatic changes with `--fix` or something. That said, the warning should also be dismiss-able. Probably beyond the scope for this PR, but if we can leave the log in place as a warning for now I think that's a step in the right direction.

@chaance I've implemented the `--fix` flag. Thanks for the feedback.

SeanGroff:
> I'd suggest that we keep the warning but allowing users to opt-in to automatic changes with `--fix` or something. That said, the warning should also be dismiss-able. Probably beyond the scope for this PR, but if we can leave the log in place as a warning for now I think that's a step in the right direction.

Anything else I can do to help progress this?

PR #4789, opened 2022-12-06
Integration test for race condition in <Form /> with Firefox

m5r:
Run `yarn bug-report-test` to experience the bug discussed in #4780

Issue #4780, opened 2022-12-06
Race condition in <Form /> with Firefox


m5r:
### What version of Remix are you using?

1.8.0

### Steps to Reproduce

- clone the minimal repro case: https://github.com/m5r/remix-form-firefox-redirect-bug
- create a [github oauth app](https://github.com/settings/applications/new) - make sure the field Authorization callback URL is set to `http://localhost:3000/sign-in/github/callback`
- generate a new client secret for this oauth app and copy the id & the secret into `the .env` file
- run the app
- sign in with github - this will take you through github's approval screen and then it redirects you back to the app
- sign out
- sign in again with github - this time, it doesnt go through github's approval screen and the app will never authenticate you

At this point, the remix app server will have logged something like this:
```
GET / 200 - - 7.426 ms
GET /sign-in?_data=routes%2Fsign-in 200 - - 4.052 ms
POST /sign-in/github 302 - - 5.879 ms
GET /sign-in/github/callback?code=f1944599380f04a97597&state=90d56e50-81cf-46f3-b392-ba71c359e9f2 - - - - ms
Error: queryRoute() call aborted
    at loadRouteData (/home/mokhtar/app/remix-form-firefox-redirect-bug/node_modules/@remix-run/router/router.ts:2278:13)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at queryImpl (/home/mokhtar/app/remix-form-firefox-redirect-bug/node_modules/@remix-run/router/router.ts:2082:20)
    at Object.queryRoute (/home/mokhtar/app/remix-form-firefox-redirect-bug/node_modules/@remix-run/router/router.ts:2041:18)
    at handleResourceRequestRR (/home/mokhtar/app/remix-form-firefox-redirect-bug/node_modules/@remix-run/server-runtime/dist/server.js:265:20)
    at requestHandler (/home/mokhtar/app/remix-form-firefox-redirect-bug/node_modules/@remix-run/server-runtime/dist/server.js:46:18)
    at /home/mokhtar/app/remix-form-firefox-redirect-bug/node_modules/@remix-run/express/dist/server.js:39:22
GET /sign-in 200 - - 10.210 ms
```

### Expected Behavior

The form submission on `/sign-in/github` should wait for response from `/sign-in/github/callback`.  
It works as expected with Chrome but not with Firefox.

### Actual Behavior

[...]

PR #5511, opened 2023-02-18
 Integration test for race condition in <Form /> with Firefox

m5r:
Run `yarn bug-report-test` to experience the bug discussed in #4780

[Initial PR](#4789) could not be reopened so I created this new one

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -1,8 +1,21 @@
+import http from "node:http";
+import { setTimeout as wait } from "node:timers/promises";
 import { test, expect } from "@playwright/test";
 
 import { PlaywrightFixture } from "./helpers/playwright-fixture";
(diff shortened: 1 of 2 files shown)


m5r:
@machour the bot wants to close the PR but I'm guessing it should stay open to stay on the team's radar

machour:
@m5r yup, adding a comment makes the needs-response label goes away ;)
Could you rebase & fix conflicts please?

m5r:
Done. Is it intentional that the CI only runs tests with chromium now?

On the bright side, I ran this test locally and it now passes. I ran it with remix@1.14.0 where it failed and then re-ran it with remix@1.15.0 where it passed so some changes happened there that fixed the race condition :raised_hands:  
I would guess it's related to the react-router upgrade but that's just my uneducated guess. Anyway, this PR and its related issue can now be closed :)

PR #5537, opened 2023-02-22
feat(vercel): Add Vercel Edge runtime entrypoint

TooTallNate:
These days, the `@remix-run/vercel` runtime adapter is not strictly required for usage of Remix on Vercel, since we are now auto-injecting a server entry point by default (we've even [updated our template](https://github.com/vercel/vercel/pull/9472) to not include a `server.js` file anymore).

_However_, [users may still be relying on a `server.js` file](https://github.com/orgs/vercel/discussions/1596) for the purposes of defining a `getLoadContext()` function (and possibly other edge cases).

So this PR adds a new entrypoint into `@remix-run/vercel` that will be utilized when building Remix for the Vercel Edge runtime to allow for usage of `@remix-run/vercel` for these edge cases. Since the signature of the Edge runtime does not match the Node runtime, there is a separate `getEdgeLoadContext()` function which may be used to define the load context when running on Edge.

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-vercel/edge.ts
+++ b/packages/remix-vercel/edge.ts
@@ -1,0 +1,26 @@
+import type { AppLoadContext, ServerBuild } from "@remix-run/server-runtime";
+import { createRequestHandler as createRemixRequestHandler } from "@remix-run/server-runtime";
+
+export type GetEdgeLoadContextFunction = (request: Request) => AppLoadContext;
+
+/**
+ * Returns a request handler for the Vercel Edge runtime that serves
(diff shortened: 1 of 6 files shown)


MichaelDeBoey:
@TooTallNate Due to your comment on https://github.com/vercel/vercel/pull/9745#issuecomment-1519196324, is this PR still necessary?

TooTallNate:
Not not really, I'll close. This functionality exists now in the `@vercel/remix` package.

PR #5541, opened 2023-02-22
Ensure stack traces are removed from all server side errors

brophdawg11:
Ensure all server-side flows are stripping `Error` stack traces in `production` mode, including document requests, data requests, resource requests, and deferred responses.

Closes: #5445 
Related: #5446, #5199

Testing Strategy: new integration tests


integration/error-sanitization-test.ts
```diff
@@ -0,0 +1,358 @@
...
+    });
+
+    test("does not sanitize defer errors in document requests", async () => {
+      let response = await fixture.requestDocument("/defer?loader");
+      let html = await response.text();
+      expect(html).toMatch("Defer Error");
+      expect(html).not.toMatch("RESOLVED");
+      expect(html).toMatch('{"message":"REJECTED"}');
+      // TODO: I think we should be getting the stack here too?
+      expect(html).toMatch(/stack/i);
```
brophdawg11, line 304: @jacob-ebey Should we get the stack here?  I think it might have something to do with not having `NODE_ENV` set right in the E2E setup?

integration/error-sanitization-test.ts
```diff
@@ -0,0 +1,358 @@
...
+      expect(html).not.toMatch("MESSAGE:");
+      expect(html).not.toMatch(/stack/i);
+    });
+
+    test("sanitizes defer errors in document requests", async () => {
+      let response = await fixture.requestDocument("/defer?loader");
+      let html = await response.text();
+      expect(html).toMatch("Defer Error");
+      expect(html).not.toMatch("RESOLVED");
+      // TODO: is this expected or should we get "Unexpected Server Error" here?
```
brophdawg11, line 194: @jacob-ebey I think this isn't working also because of `NODE_ENV`?

packages/remix-server-runtime/errors.ts
```diff
@@ -50,22 +52,45 @@ export interface ThrownResponse<T = any> {
...
 }
 
+export function sanitizeError<T = unknown>(error: T, serverMode: ServerMode) {
+  if (error instanceof Error && serverMode !== ServerMode.Development) {
+    let sanitized = new Error("Unexpected Server Error");
+    sanitized.stack = undefined;
+    return sanitized;
+  }
+  return error;
+}
```
brophdawg11, line 62: This should be the underlying method that _every_ `Error` runs through

packages/remix-server-runtime/errors.ts
```diff
@@ -50,22 +52,45 @@ export interface ThrownResponse<T = any> {
...
   message: string;
   stack?: string;
 };
 
-export async function serializeError(error: Error): Promise<SerializedError> {
+export function serializeError(
+  error: Error,
+  serverMode: ServerMode
+): SerializedError {
+  let sanitized = sanitizeError(error, serverMode);
```
brophdawg11, line 84: Sanitize all errors before serializing them

packages/remix-server-runtime/server.ts
```diff
@@ -89,7 +89,7 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
         let match = matches!.find((match) => match.route.id == routeId)!;
         response = await build.entry.module.handleDataRequest(response, {
           context: loadContext,
-          params: match.params,
+          params: match ? match.params : {},
```
brophdawg11, line 92: Be defensive in the case of `/valid-route?_data=not-a-valid-route`

packages/remix-server-runtime/server.ts
```diff
@@ -174,26 +178,24 @@ async function handleDataRequestRR(
...
-      return errorBoundaryError(errorInstance, status);
-    }
-
-    return errorBoundaryError(new Error("Unexpected Server Error"), status);
+    return json(serializeError(errorInstance, serverMode), {
+      status,
+      headers: {
+        "X-Remix-Error": "yes",
+      },
+    });
```
brophdawg11, line 198: `errorBoundaryError` was only being used in this one spot so I inlined it

packages/remix-server-runtime/server.ts
```diff
@@ -270,6 +272,11 @@ async function handleDocumentRequestRR(
     return context;
   }
 
+  // Sanitize errors outside of development environments
+  if (context.errors) {
+    context.errors = sanitizeErrors(context.errors, serverMode);
+  }
```
brophdawg11, line 278: Sanitize before we render from this context

MichaelDeBoey (approved):

Issue #5070, opened 2023-01-11
`clientLoader` and `clientAction`


ryanflorence:
Discussed in https://github.com/remix-run/remix/discussions/4950
Related proposal: https://github.com/remix-run/remix/discussions/2853

@ryanflorence to write more details in here for the API

Docs need to be clear the risk of breaking your app with this API is high.

OnurGvnc:
for reference, the relevant section in Roadmap Planning # 4;  https://youtu.be/4Q1P8T27UBo?t=3230

jacob-ebey:
Initial work on this https://github.com/remix-run/remix/pull/5543

ryanflorence:
With RSC I think we have a different path forward here, moving back to the backlog for now.

PR #5543, opened 2023-02-23
initial work on clientLoader & clientAction

jacob-ebey:
Closes: #5070

- [ ] Docs
- [x] Tests

Testing Strategy:

- new integration tests

### DIFF
--- a/packages/remix-react/data.ts
+++ b/packages/remix-react/data.ts
@@ -26,11 +26,17 @@ export function isErrorResponse(response: any): boolean {
   );
 }
 
+const redirectStatusCodes = new Set([301, 302, 303, 307, 308]);
+
+export function isRedirectStatusCode(statusCode: number): boolean {
+  return redirectStatusCodes.has(statusCode);
+}
+
 export function isRedirectResponse(response: any): boolean {
   return (
     response instanceof Response &&
     (response.headers.get("X-Remix-Redirect") != null ||
-      (response.status >= 300 && response.status < 400))
+      isRedirectStatusCode(response.status))
   );
 }
 
--- a/packages/remix-react/data.ts
+++ b/packages/remix-react/data.ts
@@ -29,7 +29,8 @@ export function isErrorResponse(response: any): boolean {
 export function isRedirectResponse(response: any): boolean {
   return (
     response instanceof Response &&
-    response.headers.get("X-Remix-Redirect") != null
+    (response.headers.get("X-Remix-Redirect") != null ||
+      (response.status >= 300 && response.status < 400))
   );
 }
 
--- a/packages/remix-react/routeModules.ts
+++ b/packages/remix-react/routeModules.ts
@@ -37,7 +37,7 @@ export interface RouteModule {
 export interface ClientDataFunctionArgs {
   request: Request;
   params: Params;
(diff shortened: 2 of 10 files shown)


packages/remix-react/data.ts
```diff
@@ -29,7 +29,8 @@ export function isErrorResponse(response: any): boolean {
 export function isRedirectResponse(response: any): boolean {
   return (
     response instanceof Response &&
-    response.headers.get("X-Remix-Redirect") != null
+    (response.headers.get("X-Remix-Redirect") != null ||
+      (response.status >= 300 && response.status < 400))
```
brophdawg11, line 33: We can steal this util from RR (or even export it as `UNSAFE_` and use it directly:

```ts
const redirectStatusCodes = new Set([301, 302, 303, 307, 308]);

export function isRedirectStatusCode(statusCode: number): boolean {
  return redirectStatusCodes.has(statusCode);
}
```

brophdawg11: Is the reason we need this because clientLoader doesn't run through the internal checking?  Maybe we can get rid of the need for this an put _less_ in the `doFetch`...

packages/remix-react/routeModules.ts
```diff
@@ -27,6 +28,38 @@ export interface RouteModule {
...
+  clientAction?: ClientActionFunction;
+  clientLoader?: ClientLoaderFunction;
+}
+/**
+ * The arguments passed to ActionFunction and LoaderFunction.
+ */
+export interface ClientDataFunctionArgs {
+  request: Request;
+  params: Params;
+  next: () => Promise<Response | DeferredData | null>;
```
brophdawg11, line 40: We can bikeshed this as we go - I like `serverFetch` at the moment

packages/remix-react/routes.tsx
```diff
@@ -177,23 +179,59 @@ function createDataFunction(
...
+      let msg =
+        `Route "${route.id}" does not have an action, but you are trying ` +
+        `to submit to it. To fix this, please add an \`action\` function to the route`;
+      console.error(msg);
+      throw new Error(msg);
+    } else if (!isAction && !route.hasLoader) {
+      return null;
+    }
+
+    let result = await fetchData(request, route.id);
```
brophdawg11, line 193: I wonder if `doFetch` could stop here so we get all the same error/redirect/catch handling whether they hit the server or not?

packages/remix-react/routes.tsx
```diff
@@ -177,23 +179,59 @@ function createDataFunction(
...
+        ? route.hasClientAction
+          ? (await routeModulePromise).clientAction
+          : undefined
+        : route.hasClientLoader
+        ? (await routeModulePromise).clientLoader
+        : undefined;
 
-      let result = await fetchData(request, route.id);
+      let result = clientDataFunction
+        ? clientDataFunction({
```
brophdawg11, line 229: Should we `await` here?

brophdawg11 (commented):
So pumped for this 🤩

packages/remix-react/data.ts
jacob-ebey: The reason we need this is because Remix is expecting a 204 response to represent redirects and we don't currently handle *real* redirect responses.

PR #5555, opened 2023-02-23
perf(remix-server-runtime): use json.parse for route manifest

redabacha:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

used JSON.parse in the generated code for adding the route manifest as it's faster than using an object literal, see https://v8.dev/blog/cost-of-javascript-2019#json

created benchmarks here: https://github.com/redabacha/remix-manifest-json-parse. it's anywhere from 1x to 2x faster depending on the number of routes in the manifest (the more routes the larger the performance difference from what i've observed).
```
Creating test manifest with 100 route(s)
cpu: AMD Ryzen 9 5950X 16-Core Processor
runtime: node v18.14.2 (x64-linux)

benchmark               time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------- -----------------------------
with JSON.parse     461.36 µs/iter   (391.85 µs … 3.36 ms)  460.6 µs   1.19 ms   1.84 ms
without JSON.parse  591.14 µs/iter   (536.34 µs … 3.62 ms)  593.3 µs 831.95 µs   1.07 ms

summary
  with JSON.parse
   1.28x faster than without JSON.parse
```


machour:
@redabacha Would you mind rebasing your PR and fixing the conflicts ? 🙏🏼

MichaelDeBoey (approved):

packages/remix-dev/compiler/manifest.ts
```diff
@@ -107,7 +107,7 @@ export const write = async (config: RemixConfig, assetsManifest: Manifest) => {
 
   await writeFileSafe(
     path.join(config.assetsBuildDirectory, filename),
-    `window.__remixManifest=${JSON.stringify(assetsManifest)};`
+    `window.__remixManifest=JSON.parse('${JSON.stringify(assetsManifest)}');`
```
brophdawg11, line 110: Wrapping this in single quote's isn't guaranteed to work if the object contains any single quotes.  We should switch this over to a double `JSON.stringify` like we do in `react-router-dom` `StaticRouterProvider`.

Screenshot 2023-05-02 at 9 27 30 AM

brophdawg11: Updated in [4739fbf](https://github.com/remix-run/remix/pull/5555/commits/4739fbf945415aa3520d24ddee34d4f732801a69)

brophdawg11: The only concern with the double `JSON.stringify` is the added bytes over the wire due to the `\\` escaping though - but the manifest oughta be cached to alleviate that after the first load?

pcattori:
`JSON.parse` vs `JSON.stringify` is not a bottleneck and only makes sub-millisecond improvement, but adds additional indirection. Feel free to reopen if this becomes a bottleneck.

brophdawg11:
Just as a note, I don't think we can fully trust the benchmarks reported here since they're from Node on a 16-core machine.  The thing aiming to be sped up here is the cost of reading the manifest in the browser on the client device, not necessarily a 16-core node CPU.  I would expect to see better results on a slower device/CPU, but would want to see real RUM data before declaring this approach as "better".

PR #5556, opened 2023-02-23
Fix `useMatches` returning  different loader data on sub navigation on client side

wizardlyhel:
Closes: #

- [ ] Docs
- [ ] Tests

**Problem:** `useMatches` returning different route match data on sub navigation. This problem isn't obvious because the last `useMatches` hook call always returns the correct route's loader data

https://stackblitz.com/edit/remix-run-remix-askr4f?file=app/root.tsx

Navigate between home, test and test 2 page.

1. On full page reload, you will see multiple logs from `useMatches` of the same route (This is fine and it is due to React re-renders)
2. On sub navigation, you will see multiple logs from `useMatches` but of different route loader data. It is always the current route and the previous route

Notice that the server side console only logs loader data for the correct route

### DIFF
--- a/.changeset/thirty-cameras-rule.md
+++ b/.changeset/thirty-cameras-rule.md
@@ -1,0 +1,6 @@
+---
+"remix": patch
+"@remix-run/react": patch
+---
+
+Fix `useMatches` causing client side render twice on sub navigation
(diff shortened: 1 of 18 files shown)


brophdawg11:
@wizardlyhel Thanks for this!  Would you mind rebasing onto `dev` and re-pointing there since this contains a code change (docs-only changes can go straight to main)?

wizardlyhel:
I mess up the rebase recreate the PR https://github.com/remix-run/remix/pull/5603

PR #5560, opened 2023-02-24
chore(remix-dev): remove broken param collision check

mcansh:
hit a snag in a demo app as it complained there was a conflict when there wasn't.

```log
⚠️ Route Path Collision: "/:username"

The following routes all define the same URL, only the first one will be used

🟢 routes/_user.$username.tsx
⭕️️ routes/sneakers.$sneakerId.tsx
```

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/__tests__/flat-routes-test.ts
+++ b/packages/remix-dev/__tests__/flat-routes-test.ts
@@ -640,9 +640,7 @@ describe("flatRoutes", () => {
       let routes = Object.values(routeManifest);
 
       expect(routes).toHaveLength(testFiles.length);
-      expect(consoleError).not.toHaveBeenCalledWith(
-        getRouteConflictErrorMessage("/:username", testFiles)
-      );
+      expect(consoleError).not.toHaveBeenCalled();
     });
   });
 
--- a/packages/remix-dev/__tests__/flat-routes-test.ts
+++ b/packages/remix-dev/__tests__/flat-routes-test.ts
@@ -619,6 +619,33 @@ describe("flatRoutes", () => {
     }
   });
 
+  describe("doesn't warn when there's not a route collision", () => {
+    let consoleError = jest
+      .spyOn(global.console, "error")
+      .mockImplementation(() => {});
+
+    afterEach(consoleError.mockReset);
+
+    test("same number of segments and the same dynamic segment index", () => {
+      let testFiles = [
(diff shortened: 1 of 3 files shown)


packages/remix-dev/__tests__/flat-routes-test.ts
```diff
@@ -619,6 +619,33 @@ describe("flatRoutes", () => {
     }
   });
 
+  describe("doesn't warn when there's not a route collision", () => {
```
mcansh, line 622: this test case is here for when param collision is brought back

mcansh:
i think we can skip the changeset here as this was never in a stable release yet

packages/remix-dev/__tests__/flat-routes-test.ts
```diff
@@ -666,7 +692,7 @@ describe("flatRoutes", () => {
       );
     });
 
-    test("same path, different param name", () => {
+    test.skip("same path, different param name", () => {
```
mcansh, line 695: remove skip once new conflict detection is brought in

packages/remix-dev/__tests__/flat-routes-test.ts
```diff
@@ -619,6 +619,33 @@ describe("flatRoutes", () => {
...
+        APP_DIR,
+        testFiles.map((file) => path.join(APP_DIR, normalizePath(file)))
+      );
+
+      let routes = Object.values(routeManifest);
+
+      expect(routes).toHaveLength(testFiles.length);
+      expect(consoleError).not.toHaveBeenCalledWith(
+        getRouteConflictErrorMessage("/:username", testFiles)
+      );
```
brophdawg11, line 645: nit: could probably make this a stricter assertion to ensure no errors got logged

```suggestion
      expect(consoleError).not.toHaveBeenCalled();
```

brophdawg11 (approved):

brophdawg11:
Ah - @mcansh we will need a changeset here since this points to `release-next` - otherwise it won't trigger a new release.  But we can remove the actual line from the final release notes since this never shipped

mcansh:
> Ah - @mcansh we will need a changeset here since this points to `release-next` - otherwise it won't trigger a new release. But we can remove the actual line from the final release notes since this never shipped

👍 this is the fella to remove `.changeset/afraid-tigers-occur.md`

PR #5574, opened 2023-02-26
perf(remix-dev/config): improve `createRoutPath`'s footprint & unit tests

gyx1000:
Hi Remix,

I chose a todo at random in the code and came across the createRoutePath function of remix-dev.

I added the last missing unit test to have 100% coverage but I'm not sure if the case reacts correctly.

Indeed, a routes/index/children path will generate `routeschildren`. I did not change the way it works to avoid a braking change.
I don't know what we want in this case `createRoutePath('routes/index/children')`:

`routeschildren`
`routes/children`
`routes/index/children`

https://github.com/gyx1000/remix/blob/perf/dev-create-route-path/packages/remix-dev/config/routesConvention.ts#L201

By analyzing the function, I also realized that we could improve its performance by about 50% (with the tests I performed).
Several functions are initialized for each iterated character. I simply converted the function to a variable to avoid this. 

```javascript
console.time('createRoutePath');
for(let i=0; i<1_000; ++i) {
    createRoutePath('routes/children/subchildren')
}
console.timeEnd('createRoutePath');

// before change: 22.96ms
// after change: 11.143ms
```

Thanks again for this wonderful project.

### DIFF
--- a/packages/remix-dev/__tests__/routesConvention-test.ts
+++ b/packages/remix-dev/__tests__/routesConvention-test.ts
@@ -14,6 +14,7 @@ describe("createRoutePath", () => {
       ["routes/$slug", "routes/:slug"],
       ["routes/sub/$slug", "routes/sub/:slug"],
       ["routes.sub/$slug", "routes/sub/:slug"],
+      ["routes/index/children", "routeschildren"],
       ["$", "*"],
       ["nested/$", "nested/*"],
       ["flat.$", "flat/*"],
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thanks for the update @gyx1000!  This code is actually about to be deleted when we move our default to the "flat routes" convention for v2 - but we've moved this "nested folders" approach over to a [new package](https://github.com/remix-run/v1-compat-utils) folks can continue to use in v2 if they prefer it.  So I don't think it's worth merging these changes here just to delete the code but if you'd like to make them over in the other repo and open a PR we could merge them for future usages of that package!

As for the `routes/index/children` test - since `index.tsx` is a reserved name, you may need to leverage `routes/[index]/children.tsx` if you actually want to be able to hit a URL such as `https://domain.com/index/children`, so I'm not sure if that is a valid test.  But the routing convention isn't my specialty so I'll defer to @mcansh on that for the new PR.  

Thanks!

PR #5587, opened 2023-02-27
fix(*): fix error message

c43721:
Fixes error message when a deferred value can never be resolved to read more fluently.

I checked other usages in this file and found no other obvious violations, so this is unfortunately One Of Those:tm: prs.

- [x] Docs
- [ ] Tests

Testing Strategy:

N/A


c43721:
@machour I assume changing this error resulted in some failures, thus the revert?

machour:
@c43721 Thank you for this 🙏🏼

PR #3171, opened 2022-05-12
bug(remix-react): add failing test for scroll-restoration

Mattinton:
Page scroll should be at the top on the new page

PR #5590, opened 2023-02-27
Move scroll test to it's own file

brophdawg11:
Reverts an unintentional change to `bug-report-test.ts` (#3171) and moves the passing test to it's own `scroll-test.ts` file

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -319,6 +319,7 @@
 - Matthew-Mallimo
 - MatthewAlbrecht
 - matthova
+- Mattinton
 - mattmazzola
 - mattstobbs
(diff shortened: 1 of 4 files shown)


contributors.yml
```diff
@@ -319,7 +319,6 @@
 - Matthew-Mallimo
 - MatthewAlbrecht
 - matthova
-- Mattinton
```
mcansh, line 322: might want to keep them around :)

brophdawg11: ah good catch

jacob-ebey (approved):

mcansh (approved):

PR #5594, opened 2023-02-27
Bump router versions

brophdawg11:


.changeset/red-kids-look.md
```diff
@@ -1,5 +1,5 @@
 ---
-"remix": patch
+"@remix-run/dev": patch
```
brophdawg11, line 2: Changed this for visibility since it seems `remix`-only changesets don't get picked up by the bot?

mcansh (approved):

PR #5601, opened 2023-02-28
fix: Memoize useMatches in the Remix layer

brophdawg11:
RR memoizes `useMatches` but we missed it in the Remix layer which needs to read the `handle` from `routeModulesCache`


brophdawg11:
Closing in favor of https://github.com/remix-run/remix/pull/5556

PR #5602, opened 2023-02-28
chore: don't log server errors for aborted requests

brophdawg11:
Since aborted requests is an ecpedted flow, we should not log the error from the router (i.e., `new Error("queryRoute() call aborted")`).  We had this check for some flows but not all.

### DIFF
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -182,9 +182,7 @@ async function handleDataRequestRR(
       errorInstance = error.error || errorInstance;
     }
 
-    if (serverMode !== ServerMode.Test && !request.signal.aborted) {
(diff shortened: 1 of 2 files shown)


packages/remix-server-runtime/server.ts
```diff
@@ -383,10 +389,6 @@ async function errorBoundaryError(error: Error, status: number) {
 }
 
 function returnLastResortErrorResponse(error: any, serverMode?: ServerMode) {
-  if (serverMode !== ServerMode.Test) {
-    console.error(error);
-  }
```
brophdawg11, line 388: Lift this to the two places we call this function so we can check the signal

jacob-ebey (approved):

MichaelDeBoey (approved):

PR #5606, opened 2023-02-28
feat(remix-dev): add warning when `future.v2_routeConvention` is not enabled

mcansh:
~~for some reason adding the warning is causing the HMR tests to fail 🤔 https://github.com/remix-run/remix/actions/runs/4317024896/jobs/7533572210~~

edit: it was checking stderr length and the warning goes to stderr 🤦‍♂️

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -731,5 +731,6 @@ let listFormat = new Intl.ListFormat("en", {
   type: "conjunction",
 });
 
-export let serverBuildTargetWarning = `The "serverBuildTarget" config option is deprecated. Use a combination of "publicPath", "serverBuildPath", "serverConditions", "serverDependenciesToBundle", "serverMainFields", "serverMinify", "serverModuleFormat" and/or "serverPlatform" instead.`;
-export let flatRoutesWarning = `The old route convention has been deprecated in favor of "flat routes". Please enable it via your remix.config https://remix.run/docs/en/main/file-conventions/route-files-v2`;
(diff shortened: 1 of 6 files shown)


packages/remix-dev/config.ts
```diff
@@ -727,3 +732,4 @@ let listFormat = new Intl.ListFormat("en", {
 });
 
 export let serverBuildTargetWarning = `The "serverBuildTarget" config option is deprecated. Use a combination of "publicPath", "serverBuildPath", "serverConditions", "serverDependenciesToBundle", "serverMainFields", "serverMinify", "serverModuleFormat" and/or "serverPlatform" instead.`;
+export let flatRoutesWarning = `The old route convention has been deprecated in favor of "flat routes". Please enable it via your remix.config https://remix.run/docs/en/main/file-conventions/route-files-v2`;
```
brophdawg11, line 735: Should we prefix this with `⚠️ DEPRECATED: or something to make it stand out a bit more?

```suggestion
export let flatRoutesWarning = `⚠️ DEPRECATED: The old nested folders route convention has been deprecated in favor of "flat routes.  Please enable the new routing convention via the \`future.v2_routeConvention\` flag in your \`remix.config.js\` file.  For more information, please see https://remix.run/docs/en/main/file-conventions/route-files-v2.`;
```

Eventually I'd love to see it in yellow via `chalk` or something but I know we have future CLI improvements in the pipeline so probably worth waiting for that before doing anything too wild.

brophdawg11: I wonder if we should also make sure we have a section in the new docs on "here's how to continue using the old nested folders version via `defineRoutes`

brophdawg11 (approved):

packages/remix-dev/config.ts
mcansh: all uses of `warnOnce` could benefit from being yellow imo

- sessions not being signed
- cookies have expires setup
- flat routes v2
- serverBuildTarget

jacob-ebey (approved):

PR #5634, opened 2023-03-02
refactor(remix-dev): flat routes perf improvements

mcansh:
- Stop glob matching for routes, instead fs.readdirSync() just the routes directory, and if (isDir) look for a index or route module.
- Move away from define routes helper for something like this with a single iteration over routes sorted by reverse ID length

this is available as `v0.0.0-experimental-699baf49c`

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

### DIFF
--- a/.changeset/flat-routes-perf.md
+++ b/.changeset/flat-routes-perf.md
@@ -1,0 +1,6 @@
+---
+"remix": patch
+"@remix-run/dev": patch
+---
+
+flat routes perf improvements
--- a/integration/flat-routes-test.ts
+++ b/integration/flat-routes-test.ts
@@ -209,7 +209,7 @@ test.describe("warns when v1 routesConvention is used", () => {
     console.error = originalConsoleError;
   });
 
-  test("warns about conflicting routes", () => {
+  test("v2_routeConvention is not enabled", () => {
     console.log(buildOutput);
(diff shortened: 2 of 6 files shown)


templates/arc/remix.config.js
```diff
@@ -3,7 +3,6 @@ module.exports = {
   future: {
     v2_routeConvention: true,
   },
-  ignoredRouteFiles: ["**/.*"],
```
MichaelDeBoey, line 6: We added these as a generic glob, mainly because we wanted to ignore dotfiles (like `.DS_Store`), but also to give people the chance to ignore other files if needed

Does this mean that wouldn't be possible anymore?

mcansh: > Does this mean that wouldn't be possible anymore?

still possible to ignore things (https://github.com/remix-run/remix/pull/5634/commits/fc2090cc0e923e4b5b5f74bc22ed1635240aef37), but in flat-routes only actual route files are picked up https://github.com/remix-run/remix/pull/5634/files#diff-1047822594960dda7f26a890c6fa53668b6294d6349ee53bcba6590653ce077eR106-R122

all in all, in a flat routes world there should be less need to manually ignore things

MichaelDeBoey (approved):

jacob-ebey (approved):

PR #5637, opened 2023-03-02
fix(remix-dev): `require.resolve` isolateChunks for pnpm

### LINKED ISSUE #5811: pnpm doesn't work with HMR/HDR
PR: https://github.com/remix-run/remix/pull/5637

mcansh:
`react-refresh` is a dependency of `@remix-run/dev`, not of your app so it's not "hoisted" to node_modules.

so we can use require.resolve to find it. This fixes an issue where remix-dev would not work with pnpm.

see https://github.com/pnpm/pnpm/issues/2957#issuecomment-719832332 for a similar issue in CRAv4 and how they fixed it there https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/230/files

confirmed working: https://twitter.com/jplhomer/status/1631375544263516180

closes #5811

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/compiler/compileBrowser.ts
+++ b/packages/remix-dev/compiler/compileBrowser.ts
@@ -134,13 +134,13 @@ const createEsbuildConfig = (
   if (mode === "development" && config.future.unstable_dev) {
     // TODO prebundle deps instead of chunking just these ones
     let isolateChunks = [
-      "react",
-      "react/jsx-dev-runtime",
-      "react/jsx-runtime",
-      "react-dom",
-      "react-dom/client",
+      require.resolve("react"),
+      require.resolve("react/jsx-dev-runtime"),
+      require.resolve("react/jsx-runtime"),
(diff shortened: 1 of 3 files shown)


packages/remix-dev/compiler/compileBrowser.ts
```diff
@@ -135,7 +135,7 @@ const createEsbuildConfig = (
       "react/jsx-runtime",
       "react-dom",
       "react-dom/client",
-      "react-refresh/runtime",
+      require.resolve("react-refresh/runtime"),
```
mcansh, line 138: other packages listed are installed in the user's app already, so we shouldn't need to resolve them

mcansh: did it anyway to be consistent and safe

MichaelDeBoey (approved):

Issue #5629, opened 2023-03-02
1.14.0: Build performance regression in browserRouteModulesPlugin


dmarkow:
### What version of Remix are you using?

1.14.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Remix 1.14.0 introduced more complexity to browserRouteModulesPlugin. This has caused up to a 5x increase in build times, which is very noticeable when working in a large app. These times are roughly the same for an initial build and rebuild.

- 1.13.0: 1.9s
- 1.14.0 / old dev server: 4.2s
- 1.14.0 / new dev server: 8.7s

Throwing an [esbuild profile wrapper](https://github.com/evanw/esbuild/issues/1778#issuecomment-977522702) in, it's showing most the time is spent in this plugin. Since esbuild uses every core as much as possible, the actual 41 seconds in the first example is far less in real time. In the second example it's a tiny bit more, but something in there is causing the 2x speed increase. Adding in the new dev server (with HMR), it jumps another 2x on top of that (I'm unsure if that additional increase is in browserRouteModulesPlugin, hmrPlugin, or a combo of both).

Detailed output of each build below. I'll try to dig into a possible PR to improve this, but wanted to raise the issue in case anyone is seeing problems.

Remix 1.13.0 (same results with unstable_dev on or off):

```
± yarn remix dev
yarn run v1.22.19
$ /Users/dylan/dev/pfcs/node_modules/.bin/remix dev
[plugin: deprecated-remix-package]
  total: 649ms
   ├ start: 1ms
   └ onResolve: 648ms

[plugin: css-file]
  total: 122ms
   ├ start: 1ms
   └ onLoad: 121ms

[plugin: url-imports]
  total: 1ms
   └ start: 1ms

[plugin: remix-mdx]
  total: 79ms
   └ start: 79ms

[plugin: browser-route-modules]
  total: 41566ms
   ├ start: 1ms
   ├ onResolve: 14ms
   └ onLoad: 41551ms

[plugin: empty-modules]
  total: 8ms
   ├ start: 0ms
   ├ onResolve: 6ms
   └ onLoad: 2ms

[plugin: node-modules-polyfills]
  total: 224ms
   ├ start: 0ms
   ├ onResolve: 197ms
   └ onLoad: 27ms

💿 Built in 1.9s
```

Remix 1.14.0, unstable_dev turned off:

```
± yarn remix dev
yarn run v1.22.19
[...]

PR #5645, opened 2023-03-03
Use cache to avoid reprocessing unchanged files

dmarkow:
Closes: #5629. Caching file transforming results on unchanged files brings rebuild times back down to pre-1.14.0 levels.

- [ ] Docs
- [ ] Tests

Testing Strategy: HMR integration test is still passing. Used it this week on my primary app and HMR seems to still work as expected, but with much quicker build times.

```bash
$ playwright test --config ./integration/playwright.config.ts -j 1 integration/hmr-test.ts

Running 4 tests using 1 worker
[chromium] › hmr-test.ts:185:1 › HMR
[webkit] › hmr-test.ts:185:1 › HMR
[edge] › hmr-test.ts:185:1 › HMR
[firefox] › hmr-test.ts:185:1 › HMR

  4 passed (33s)
```


ZipBrandon:
@dmarkow Thanks for making this. I patch-packaged it onto 1.14.0 and my rebuild time is around 5.5 seconds compared to 9 or 10.

dmarkow:
@pcattori This is the PR I mentioned on Discord. just rebased to the current dev branch and tests still passing. 700 route app was at 10-12s rebuilds before this, 2-3s after. Still not ideal, but almost an order of magnitude better. 

Uses the caching strategy outlined at https://esbuild.github.io/plugins/#caching-your-plugin to avoid re-transforming/writing files that haven't actually changed.

I tried to find a way to speed up initial builds too, but I think babel's transform function was where most of the time was spent (it's been a little while since I dug into this). We could add a file-based cache, but that adds a lot more complexity and potential for stale cache files.

dmarkow:
Tested my patch on the released 1.16.0 today and it still works. Time savings aren't quite as good now, unfortunately -- I'm at around 4 second rebuilds now with a 12 second clean build, whereas I was getting low 2 second rebuilds on 1.15 with my patch. This may be related to the css plugins though -- I noticed a fair amount of time is spent compiling the same css multiple times which may be part of that slowdown, I'll dig in when/if I get more time.

pcattori:
Thanks @dmarkow! With new dev server architecture done, I'm now laser-focused on bug fixes and perf improvements for compiler / dev server.

packages/remix-dev/compiler/js/plugins/routes_unstable.ts
```diff
@@ -130,28 +132,42 @@ export function browserRouteModulesPlugin(
 
           let sourceCode = fs.readFileSync(routeFile, "utf8");
```
dennishall, line 133: @dmarkow - What do you think of this idea? Maybe it's faster? Check for filechange by using `fs.statSync` -- before reading the file's contents -- and if stats.mtime or stats.size differ from cache, then file has changed? Could save the time of reading the whole file every time to see if it has changed.

dennishall: Surprisingly underwhelming result. Thanks for checking it out!

PR #5658, opened 2023-03-04
docs(guides/not-found): Set statusText correctly in not-found guide

everdimension:
This is a fix to set `statusText` correctly in the [not-found](https://remix.run/docs/en/main/guides/not-found#how-to-send-a-404) guide example.
I assume the example is meant to set "statusText" and not the response body, because the following CatchBoundary displays it:

```jsx
...
<h1>
  {caught.status} {caught.statusText}
</h1>
...
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -142,6 +142,7 @@
 - eps1lon
 - esamattis
 - evanwinter
+- everdimension
 - exegeteio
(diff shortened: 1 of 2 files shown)


machour (approved):
Thank you so much @everdimension 🙌🏼

PR #5660, opened 2023-03-04
Please allow the user to set the session key in cloudflare workers kv.

YankeeTube:
- Cloudflarekv is powerful, but for future session management queries, the user must be able to set up prefixes, etc.
 - But we don't have the option to set the session key.
 - Therefore, you should be able to give the key as an option.


YankeeTube:
Oh, my God, it's only access deno.

Issue #5677, opened 2023-03-06
Duplicate children keys in `getStylesheetPrefetchLinks`


grinkus-adapt:
### What version of Remix are you using?

1.14.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. Open [the sandbox](https://codesandbox.io/p/sandbox/elastic-volhard-pko1zw?file=%2Fapp%2Fstyles%2FComponentOfComponents.css&selection=%5B%7B%22endColumn%22%3A22%2C%22endLineNumber%22%3A2%2C%22startColumn%22%3A22%2C%22startLineNumber%22%3A2%7D%5D) and devtools. Clear the console.
2. Hover over the "Foobar?" link with mouse (or focus with keyboard, etc).
3. See that there's a "Encountered two children with the same key, `/build/_assets/Component-O4OLHRZ6.css`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version." warning in the devtools console.

The issue is that we're importing the stylesheet in a component, that is imported in a few different places.

### Expected Behavior

No errors in the console.

### Actual Behavior

There's an error in the devtools console "Warning: Encountered two children with the same key, `/build/_assets/Component-O4OLHRZ6.css`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version."

grinkus-adapt:
I imagine doing something like the code below in [remix-react/links.ts](https://github.com/remix-run/remix/blob/9a9c955f241c5742d227542e13c7387b99d4628b/packages/remix-react/links.ts#L337) would resolve the issue, but I'm not totally sure if that wouldn't break anything.

```diff
diff --git a/packages/remix-react/links.ts b/packages/remix-react/links.ts
index 6c070d574..dc59eb38e 100644
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -338,6 +338,15 @@ export async function getStylesheetPrefetchLinks(
     .flat(1)
     .filter(isHtmlLinkDescriptor)
     .filter((link) => link.rel === "stylesheet" || link.rel === "preload")
+    .filter(
+      // Dedupe links by rel and href
+      (link, indexToFind, linksArr) => {
+        let foundIndex = linksArr.findIndex(needle => {
[...]

PR #5678, opened 2023-03-06
fix(remix-react): dedupe `link`s in `getStylesheetPrefetchLinks`

grinkus-adapt:
Closes: #5677

- [ ] Docs
- [ ] Tests

Deduplicating links in `getStylesheetPrefetchLinks` avoids inserting `link`s with the same `key` prop, which React dev complains about. It's not an issue in prod build, but this tweak makes the developer experience a bit better in an edge-case where there's a stylesheet that's imported in the sub-page that a `Link` with `prefetch` prop is pointing to.

### DIFF
--- a/.changeset/hip-moose-fold.md
+++ b/.changeset/hip-moose-fold.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Deduplicate links in `getStylesheetPrefetchLinks`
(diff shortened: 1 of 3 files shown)


markdalgleish:
Thanks for the PR! As mentioned in the original issue, I've opened a PR that goes a bit further than your suggested fix because this was actually a symptom of a deeper issue, so I'm closing this in favour of this one: https://github.com/remix-run/remix/pull/7060.

PR #5679, opened 2023-03-06
Add type deprecations for types now in React Router

brophdawg11:
* Where possible, use React Router types directly
* Where we differ slightly in implementation or naming, deprecate the Remix type in favor of the React Router type


packages/remix-react/components.tsx
```diff
@@ -743,6 +730,8 @@ export function Meta() {
   return future?.v2_meta ? <V2Meta /> : <V1Meta />;
 }
 
+// TODO: The RR version is just missing the generic and resolve is
+// `TrackedPromise | any`.  Should we add the generic in RR?
 export interface AwaitProps<Resolve> {
```
brophdawg11, line 735: I need to check on how this generic is used and see if we want this in RR

packages/remix-react/components.tsx
```diff
@@ -1103,6 +1092,8 @@ function dedupe(array: any[]) {
 }
 
 // TODO: Can this be re-exported from RR?
+// Yes, but data -> unknown and handle -> unknown.  Is handle forced to
+// be an object in remix?
```
brophdawg11, line 1096: Can we switch to RR's "stricter" (due to `unknown`) type in v2?

packages/remix-react/data.ts
```diff
@@ -2,16 +2,9 @@ import {
...
-import type { FormMethod as FormMethodRR } from "react-router-dom";
 
 export type AppData = any;
 
-export type FormMethod = FormMethodRR;
-
-export type FormEncType =
-  | "application/x-www-form-urlencoded"
-  | "multipart/form-data";
-
```
brophdawg11, line 14: Exported directly from RR now

packages/remix-react/errors.ts
```diff
@@ -3,6 +3,10 @@ import { ErrorResponse } from "@remix-run/router";
 
 import type { AppData } from "./data";
 
+/**
+ * @deprecated in favor of React Router `ErrorResponse`
+ * TODO: Need to add generics in RR
+ */
 export interface ThrownResponse<
```
brophdawg11, line 10: I'm not even sure we need this or the generics once `v2_errorBoundary` becomes the default since we won't have `useCatch` anymore?

packages/remix-react/routeModules.ts
```diff
@@ -9,7 +10,8 @@ import type {
 import type { AppData } from "./data";
 import type { LinkDescriptor } from "./links";
 import type { EntryRoute } from "./routes";
-import type { RouteData } from "./routeData";
+
+type RouteData = RouterState["loaderData"];
```
brophdawg11, line 14: Grab this right from the router now instead of re-defining the `string->any` map

packages/remix-react/transition.ts
```diff
@@ -81,29 +75,6 @@ export type TransitionStates = {
...
-    type: "loaderSubmission";
-    setCookie: boolean;
-  };
-  FetchAction: {
-    isRedirect: true;
-    type: "fetchAction";
-    setCookie: boolean;
-  };
-};
-
```
brophdawg11, line 106: Removed a few thigs here that were exported but never used elsewhere - leftovers from TM implementation

packages/remix-server-runtime/errors.ts
```diff
@@ -43,7 +43,10 @@ import { isRouteErrorResponse } from "@remix-run/router";
  * line.
  */
 
-// TODO Re-export as ErrorResponse?
+/**
+ * @deprecated in favor of the `ErrorResponse` class in React Router.
+ * TODO: Need to add generics in RR
+ */
 export interface ThrownResponse<T = any> {
```
brophdawg11, line 50: Same as above - may not need this or the generics once `v2_errorBoundary` becomes the default?

packages/remix-server-runtime/responses.ts
```diff
@@ -13,11 +15,15 @@ export type TypedDeferredData<Data extends Record<string, unknown>> = Pick<
...
+// TODO: The RR version of this is the same minus the generic.  Should we add
+// that over there and remove this?
 export type DeferFunction = <Data extends Record<string, unknown>>(
   data: Data,
   init?: number | ResponseInit
 ) => TypedDeferredData<Data>;
 
+// TODO: The RR version of this is identical except it's generic doesn't
+// `extends unknown`.  Should we add that over there and remove this?
 export type JsonFunction = <Data extends unknown>(
```
brophdawg11, line 27: Unsure if these generics make sense in RR if we're moving to an [`element: (data) => <p>{data}</p>`](https://github.com/orgs/remix-run/projects/5?pane=issue&itemId=18313356) type API in the future for better typing

packages/remix-server-runtime/responses.ts
```diff
@@ -39,28 +45,18 @@ export type TypedResponse<T extends unknown = unknown> = Omit<
...
- * to JSON and sets the `Content-Type` header.
+ * This is a shortcut for creating Remix deferred responses
  *
- * @see https://remix.run/api/remix#json
+ * @see https://remix.run/docs/utils/defer
  */
 export const defer: DeferFunction = (data, init = {}) => {
   let responseInit = typeof init === "number" ? { status: init } : init;
 
+  // TODO: Do we need this or is this copy/paste from json()?
```
brophdawg11, line 59: @jacob-ebey `defer` doesn't need these headers does it?

packages/remix-server-runtime/responses.ts
MichaelDeBoey: Although it ca also be inferred, it still makes sense to define the actual type that's passed in `json` imo

packages/remix-server-runtime/routeModules.ts
```diff
@@ -66,6 +78,8 @@ export interface LinksFunction {
 
 /**
  * A function that loads data for a route.
+ *
+ * @deprecated in favor of React Router `LoaderFunction`
```
brophdawg11, line 82: We might not be able to include this deprecation yet since the naming is the same and we don't want folks to start importing from `react-router-dom`?

packages/remix-server-runtime/responses.ts
brophdawg11: Going to split the generic out for a v2 decision - not needed in 1.15.0

packages/remix-server-runtime/responses.ts
```diff
@@ -39,37 +39,16 @@ export type TypedResponse<T extends unknown = unknown> = Omit<
...
- * @see https://remix.run/api/remix#json
+ * @see https://remix.run/docs/utils/defer
  */
 export const defer: DeferFunction = (data, init = {}) => {
-  let responseInit = typeof init === "number" ? { status: init } : init;
-
-  let headers = new Headers(responseInit.headers);
-  if (!headers.has("Content-Type")) {
-    headers.set("Content-Type", "application/json; charset=utf-8");
-  }
```
brophdawg11, line 67: @jacob-ebey I think this was only here from a copy/paste and eventually gets overwritten with `text/remix-deferred` anyway.  So we can just use the router `defer` verbatim (while keeping the return type casting)?

packages/remix-react/components.tsx
```diff
@@ -112,19 +112,6 @@ function useRemixContext(): RemixContextObject {
...
-export function RemixEntry(props: {
-  context: EntryContext;
-  action: NavigationType;
-  location: Location;
-  navigator: Navigator;
-  static?: boolean;
-}) {
-  return <h1>Not Implemented!</h1>;
-}
-
```
brophdawg11, line 127: No longer used

packages/remix-server-runtime/responses.ts
```diff
@@ -39,37 +39,16 @@ export type TypedResponse<T extends unknown = unknown> = Omit<
...
-
-  return new Response(JSON.stringify(data), {
-    ...responseInit,
-    headers,
-  });
+  return routerJson(data, init);
```
brophdawg11, line 42: Use router methods directly, but preserve typecasting on the returned values

packages/remix-server-runtime/responses.ts
jacob-ebey: That should be fine.

Issue #5616, opened 2023-03-01
Remove magic exports for v2


chaance:
- [x] v1 - This may already be done: make sure they get deprecation warnings if they import from "remix", import from the right thing.
- [x] v2 - Remove everything, should be an empty module now (https://github.com/remix-run/remix/pull/5682)
  - [x] Remove ESLint warnings (https://github.com/remix-run/remix/pull/5695)
  - [x] Remove `replace-remix-magic-imports` codemod (https://github.com/remix-run/remix/pull/5750)
  - [x] Remove `deprecatedRemixPackagePlugin` (https://github.com/remix-run/remix/pull/5751)

PR #5682, opened 2023-03-07
Remove magic exports support for v2

chaance:
Closes: #5616

- [ ] Docs
- [x] Tests


MichaelDeBoey (approved):

packages/remix-dev/cli/commands.ts
```diff
@@ -108,25 +107,16 @@ export async function init(
...
-    platform = SetupPlatform.Cloudflare;
-  } else {
-    platform = isSetupPlatform(platformArg) ? platformArg : SetupPlatform.Node;
-  }
-
-  await setupRemix(platform);
-
-  log(`Successfully setup Remix for ${platform}.`);
+  console.warn(
+    "The setup command is no longer necessary as of v2. This is a no-op."
```
brophdawg11, line 118: Should we add something like `please remove this from your dev and CI scripts`?  And maybe a `WARN:` or `DEPRECATED:` prefix to catch the eye?

rollup.utils.js
```diff
@@ -430,7 +223,6 @@ module.exports = {
   getCliConfig,
   getOutputDir,
   isBareModuleId,
-  magicExportsPlugin,
 };
 
 /**
```
brophdawg11, line 228: Can we get rid of line 229 below?

```
 * @typedef {Record<string, { values?: string[]; types?: string[] }>} MagicExports
```

brophdawg11 (approved):

PR #5685, opened 2023-03-07
Remove unstable_shouldReload references

brophdawg11:
Removes references to the removed `unstable_shouldReload`, which was replaced by `shouldRevalidate` in 1.10.10


docs/api/conventions.md
```diff
@@ -190,21 +190,13 @@ title: Conventions
 
 [Moved →][moved-44]
 
-### unstable_shouldReload
+### shouldRevalidate
 
 [Moved →][moved-45]
```
machour, line 195: You're deleting the linked file in this PR, is that wanted?

MichaelDeBoey (approved):
Should we also delete these docs on `main` already?

PR #5691, opened 2023-03-07
Deprecate fetcher.type and fetcher.submission

brophdawg11:
Deprecate `fetcher.type` and `fetcher.submission`

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1361,8 +1361,8 @@ function convertNavigationToTransition(navigation: Navigation): Transition {
  */
 export function useFetchers(): Fetcher[] {
   let fetchers = useFetchersRR();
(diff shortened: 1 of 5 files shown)


packages/remix-react/components.tsx
```diff
@@ -1403,15 +1405,63 @@ export function useFetcher<TData = any>(): FetcherWithComponents<
...
+    },
+    set(value: Fetcher["type"]) {
+      // Devs should *not* be doing this but we don't want to break their
+      // current app if they are
+      type = value;
+    },
+    // These settings should make this behave like a normal object `type` field
+    configurable: true,
+    enumerable: true,
+  });
```
brophdawg11, line 1440: Only warn if they actually access `type` or `submission`

jacob-ebey (approved):

PR #5692, opened 2023-03-07
docs: use `useNavigation` instead of `useTransition`

MichaelDeBoey:
Taking the docs changes from @brophdawg11 in #5689 & already publish them


MichaelDeBoey:
@brophdawg11 already did this in #5687

PR #5161, opened 2023-01-20
fix(remix-react): undeprecate `LiveReload`'s `nonce` prop

kentcdodds:
It was necessary for CSPs before, then it was no longer necessary thanks to https://github.com/remix-run/remix/pull/2783, but then it became necessary again when this PR was merged: https://github.com/remix-run/remix/pull/2874 but the deprecation warning was never removed.

Until this is released, people will get this in their editors:

image

PR #5699, opened 2023-03-07
feat!(remix-react): remove `nonce` prop from `LiveReload` component

MichaelDeBoey:


MichaelDeBoey:
This might be in conflict with @kentcdodds' #5161

kentcdodds (changes requested):
No, this prop is definitely needed! The deprecation message should be removed, but the prop should stay!

MichaelDeBoey:
@brophdawg11 @chaance @jacob-ebey @mcansh We should merge @kentcdodds's #5161 instead

PR #5700, opened 2023-03-07
feat(remix-react): show deprecation warning when `LiveReload`'s `nonce` prop is used

MichaelDeBoey:
This will warn people about #5699

### DIFF
--- a/.changeset/long-keys-explain.md
+++ b/.changeset/long-keys-explain.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": minor
+---
+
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
This might be in conflict with @kentcdodds' #5161

kentcdodds (changes requested):
This should definitely not be deprecated or removed

MichaelDeBoey:
@brophdawg11 @chaance @jacob-ebey @mcansh We should merge @kentcdodds's #5161 instread

PR #5713, opened 2023-03-08
fix(remix-server-runtime): inline `react`'s `ComponentType` type

### LINKED ISSUE #5615: 1.13 - 1.14 : @remix-run/server-runtime EntryContext is not assignable to type @remix-run/react EntryContext 🔴

1.13.0

- [X] Yes

### Steps to Reproduce

Install remix server template.

Typescript version : 4.9.5

### Expected Behavior

RemixServer doesn't complain with remixContext

### Actual Behavior

Type error with remix context.

image
image

MichaelDeBoey:
Follow-up of #4801

Not the most ideal thing to do imo, but it fixes #5615 🤷‍♂️

As I mentioned in https://github.com/remix-run/remix/pull/4801#discussion_r1042617554, I added `@types/react` to `dependencies` because we're using `ComponentType` type in `packages/remix-server-runtime/routeModules.ts`

https://github.com/remix-run/remix/blob/793531210d0ef0635ac076d9aee1d31d3ee82655/packages/remix-server-runtime/routeModules.ts#L40
https://github.com/remix-run/remix/blob/793531210d0ef0635ac076d9aee1d31d3ee82655/packages/remix-server-runtime/routeModules.ts#L45
https://github.com/remix-run/remix/blob/793531210d0ef0635ac076d9aee1d31d3ee82655/packages/remix-server-runtime/routeModules.ts#L218

We should probably move that to `@remix-run/react` and make `@remix-run/server-runtime` more renderer-agnostic in that place.

@brophdawg11 thought that would be the case when moving to RR 6.4 (see https://github.com/remix-run/remix/pull/4801#discussion_r1042733354), but apparently that's not the case 😢

---

Closes #5615


MichaelDeBoey:
> we could even just use `any`

I would personally go with `unknown` instead

> * `@remix-run/server-runtime` would not import anything from `@types/react` and could eliminate the dependency entirely

👍

> * `@remix-run/react` would, but could go back to the original `devDependencies`/`peerDependencies` approach?


React is already in `peerDependencies` of `@remix-run/react`
https://github.com/remix-run/remix/blob/209ac82ea9b874ceb093a5a86b759560ebbf2d23/packages/remix-react/package.json#L30-L33

---

What do you want me to do about `@types/cookie`?
Do I create another PR for adding it back to `dependencies`, so it fixes #6275?

MichaelDeBoey:
> I would personally go with `unknown` instead
> https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any

We should use `any` as otherwise people who use the exported types, will get type errors

brophdawg11:
We should repoint this to the `v2` branch since it's technically a breaking change to the types?

MichaelDeBoey:
@brophdawg11 making it `any` is broadening the current type, so not a breaking change imo?

And like @lpsinger says, this fixes a regression as well

pcattori:
@MichaelDeBoey I don't think we should consider more accurate types as "breaking changes". Types get stripped when building and don't affect prod. More accurate types are surfacing potential bugs that were previously invisible.

MichaelDeBoey:
@pcattori Did you want to tag @brophdawg11?
As I don't think this PR is a breaking change

brophdawg11:
I think this as it stands would be a breaking type change as `ErrorBoundaryComponent`/`RouteComponent` exported from `@remix-run/server-runtime` are becoming significantly _less_ accurate (and being deprecated at the same time in favor of the ones exported from `@remix-run/react`).  If someone was using them in their route files it would no longer provide them any type checking unless they changed their imports to pull from `@remix-run/react`.

```jsx
import { ErrorBoundaryComponent } from '@remix-run/server-runtime';

export const ErrorBoundary: ErrorBoundaryComponent = () => {
  ...
};
```

In my eyes, this PR has evolved into the correct way to handle this _moving forward in v2_ - although we should also stop exporting them from `@remix-run/server-runtime` in v2 and only export from `@remix-run/react`.  

But to @lpsinger's comment I think we still need a way to fix the regression introduced by #4801.  What was the original issue we were fixing there?  Could we revert that original change for v1 and be in an acceptable spot until v2?

pcattori:
> @pcattori Did you want to tag @brophdawg11? As I don't think this PR is a breaking change

Whoops meant to tag both of you 😅 I haven't read through the PR. Just wanted to clearly state that when types change to be more accurate, they aren't "breaking changes", they are bug fixes.

lpsinger:
> But to @lpsinger's comment I think we still need a way to fix the regression introduced by #4801. What was the original issue we were fixing there? Could we revert that original change for v1 and be in an acceptable spot until v2?

Reverting #6051 ought to fix the regression.

brophdawg11:
Talked this over with @jacob-ebey and @pcattori and we decided that the loosening of the types in server-runtime is really just a bug fix since that should never have been exporting react-flavored types.  We've included deprecation warnings pointing to the corresponding types in `@remix-run/react`.

packages/remix-react/package.json
```diff
@@ -23,6 +23,8 @@
     "@remix-run/server-runtime": "1.17.0",
     "@testing-library/jest-dom": "^5.16.2",
     "@testing-library/react": "^13.3.0",
+    "@types/react": "^18.0.15",
+    "@types/react-dom": "^18.0.6",
```
jacob-ebey, line 27: Do we actually need the react-dom types?

MichaelDeBoey: No, we only had `react` before

brophdawg11 (approved):

lpsinger:
Do I understand correctly based on the labels that this won't be fixed until v2? When will that be?

brophdawg11:
@lpsinger Nah sorry for the confusion - we're going to get this merged for the next release

brophdawg11:
Hm, we must have a bug in our bot logic - this is not included in `1.17.1-pre.1` and should land in `1.18.0`

https://github.com/remix-run/remix/compare/remix@1.17.0...remix@1.17.1-pre.1

mcansh:
> Hm, we must have a bug in our bot logic - this is not included in `1.17.1-pre.1` and should land in `1.18.0`
> 
> 
> 
> https://github.com/remix-run/remix/compare/remix@1.17.0...remix@1.17.1-pre.1

yeah that’s odd… it gets the commits via `git log --pretty=format:%H v0.0.0-nightly-eb06147-20230614...remix@1.17.1-pre.1 ./packages`, but seems like github actions got a different list 🤔

indeed not [here](https://github.com/remix-run/remix/compare/v0.0.0-nightly-eb06147-20230614...remix@1.17.1-pre.1), but somehow got into the list in [ci](https://github.com/remix-run/remix/actions/runs/5272477374/jobs/9534845414#step:3:28)

PR #5716, opened 2023-03-08
Remove fetcher.type/fetcher.submission

brophdawg11:
- [x] ~~**⚠️ Repoint to `v2` once https://github.com/remix-run/remix/pull/5689 is merged**~~

Removes `fetcher.type` and `fetcher.submission` for v2.  We still keep a small back-compat layer for now since there's an unintentional mismatch in casing between `formMethod` on Remix `useFetcher`/`useTransition` versus react Router `useFetcher`/`useNavigation`

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1168,17 +1168,7 @@ export function useActionData<T = AppData>(): SerializeFrom<T> | undefined {
  */
 export function useFetchers(): Fetcher[] {
   let fetchers = useFetchersRR();
-  return fetchers.map((f) => {
-    let fetcher = convertRouterFetcherToRemixFetcher({
-      state: f.state,
-      data: f.data,
-      formMethod: f.formMethod,
-      formAction: f.formAction,
-      formData: f.formData,
-      formEncType: f.formEncType,
-    });
-    return fetcher;
-  });
+  return fetchers.map((f) => convertRouterFetcherToRemixFetcher(f));
 }
 
 export type FetcherWithComponents<TData> = Fetcher<TData> & {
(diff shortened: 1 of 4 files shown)


packages/remix-react/components.tsx
```diff
@@ -1178,17 +1168,17 @@ export function useActionData<T = AppData>(): SerializeFrom<T> | undefined {
  */
 export function useFetchers(): Fetcher[] {
   let fetchers = useFetchersRR();
-  return fetchers.map((f) =>
-    convertRouterFetcherToRemixFetcher({
+  return fetchers.map((f) => {
+    let fetcher = convertRouterFetcherToRemixFetcher({
```
hjonasson, line 1172: It looks to me like the interface for `fetchers` has all these props and nothing else. If so then you could just return this

```TS
return fetchers.map(convertRouterFetcherToRemixFetcher);
```

brophdawg11: Good catch!

packages/remix-react/components.tsx
```diff
@@ -1210,129 +1191,36 @@ export function useFetcher<TData = any>(): FetcherWithComponents<
...
-        formData: formData,
-        submission: {
-          method: formMethod.toUpperCase() as ActionSubmission["method"],
-          action: formAction,
-          encType: formEncType,
-          formData: formData,
-          key: "",
-        },
         data,
+        formMethod: formMethod.toUpperCase() as FormMethod,
```
brophdawg11, line 1216: Small back-compat layer kept to preserve the Remix uppercase `FormMethod` on `useFetcher`

MichaelDeBoey (approved):

packages/remix-react/components.tsx
```diff
@@ -1341,93 +1229,24 @@ function convertRouterFetcherToRemixFetcher(
...
-          },
-          data,
-        };
-        return fetcher;
-      }
-    }
+    let fetcher: FetcherStates["Loading"] = {
+      state,
+      data,
+      formMethod: formMethod?.toUpperCase() as FormMethod,
```
brophdawg11, line 1235: We still need this back-compat to align to the old uppercase format of `formMethod`.  Separately we should figure out how to get RR to use uppercase so this can be removed, but that currently would be a breaking change in RR.

mcansh (approved):

MichaelDeBoey (approved):

PR #5732, opened 2023-03-09
feat!(remix-cloudflare,remix-cloudflare-pages,remix-cloudflare-workers): drop `@cloudflare/workers-types` v2 & v3 support

MichaelDeBoey:
@lkwr & @zhuhaow might want to have some remarks as they were active on #5425

---

Closes #5425


jacob-ebey (approved):

brophdawg11:
We should check with @mjackson on his thoughts on the best way to handle versioning of these dev/peer dependencies

mcansh (approved):

Issue #5524, opened 2023-02-21
🗺️ Add caching to Vanilla Extract plugin


markdalgleish:
The build can get quite slow if there are a lot of `.css.ts` files, even on rebuilds, and even when changing non-VE files. We need to add caching to address this.

Vanilla Extract core currently compiles each `.css.ts` file using esbuild before evaluating it. Other bundlers in watch mode don't recompile the file if it or any of its affected files haven't changed. However, esbuild is different in that it leaves caching entirely up to plugins and re-runs everything on each rebuild which is why perf issues are more pronounced for us.

We want to cache the results of each `.css.ts` file but there's no way to generate a cache key just by looking at the `.css.ts` entry point since it imports other modules. To fix this the Vanilla Extract compile + eval step needs to migrate from esbuild + eval to vite-node so that Vite can cache the module graph for us as part of its eval step. We can even share the cache between server and client builds which is really nice.

PR #5735, opened 2023-03-09
feat(remix-dev): add opt-in Vanilla Extract cache

markdalgleish:
Closes: #5524 

- [ ] Docs
- [x] Tests

For now this is opt-in since the cache is powered by a brand new Vanilla Extract compiler that uses vite-node. Since this is a large architectural change, we'll want to ensure it's working correctly in the wild before making it the default. As a result, I've also left this option undocumented until we get some initial feedback.

Testing Strategy:

All tests that exercise the Vanilla Extract integration are now also being run against the new compiler.


packages/remix-dev/compiler/plugins/vanillaExtractPlugin.ts
```diff
@@ -1,219 +1,30 @@
-import { dirname, join, extname } from "path";
-import type { IdentifierOption } from "@vanilla-extract/integration";
```
markdalgleish, line 2: Git didn't track this one very well — this file has been moved to `vanillaExtractPluginUncached.ts`, this file now just switches between the cached and uncached plugins since their implementations are so different.

integration/deterministic-build-output-test.ts
```diff
@@ -3,115 +3,204 @@ import globby from "globby";
...
+      },
     },
-  };
-  let dir1 = await createFixtureProject(init);
-  let dir2 = await createFixtureProject(init);
+  },
+];
 
-  expect(dir1).not.toEqual(dir2);
+testCases.forEach((testCase) => {
```
chaance, line 156: Grouping tests in loops like this makes me a bit uneasy. Makes it tougher to isolate failures IMO. I don't mind using the same files config for each test, but do you think we could run each actual test explicitly instead?

chaance (approved):
Minor feedback on the tests, otherwise 👍

PR #5738, opened 2023-03-09
bug: TypeError for not found routes/data

ledniy:
Found the following TypeErrors after penetration tests of our app.

1. Only happens if we have handleDataRequest, which we need for security headers
2. No special prerequisite need, just a data request from not existing route

I believe it should return 404 instead, as in other cases.


brophdawg11:
Thank you for the reproduction!  This will be handled by #6820.

PR #5764, opened 2023-03-11
Support corepack

TrySound:
See https://nodejs.org/api/corepack.html

Corepack is a great recent feature builtin in node. It allows to use package manager version specified in packageManager field and not have globally installed yarn or pnpm at all. Only need to run `corepack enable`


MichaelDeBoey:
This happened when upgrading to pnpm in #8848

PR #5773, opened 2023-03-12
fix(remix-dev): support `node:` protocol imports

jplhomer:
Cloudflare Workers now support a `nodejs_compat` compatibility flag, which allows imports that include the `node:` protocol, like `node:async_hooks`.

However, the Remix compiler chokes on this protocol import when attempting to build for the Cloudflare target.

This PR adds an ESBuild plugin to ensure that the built-ins which start with `node:` protocols are properly marked as `external`.

- [ ] Docs
- [x] Tests

Testing Strategy: added a failing test which is fixed by the plugin. https://github.com/remix-run/remix/pull/5773/files#diff-e8ec831b2b19bf6531cb11844ca04430025615a96c3c67a83d3d6fc5362393dc


MichaelDeBoey:
This seems to be a duplicate of @pcattori's #5222 I guess?

jasnell:
Great to see. Keep in mind that the current modules supported are node:async_hooks (AsyncLocalStorage and AsyncResource only), node:buffer (Buffer only), node:events (EventEmitter), node:assert, and node:util (types, callbackify, and promisify). More will come incrementally over the coming months.

pcattori:
@jplhomer looking good! 

I think the last thing we need is to add a check in that same integration test to make sure `node:async_hooks` doesn't end up in the browser package. We've used `shelljs`'s `grep` as a platform-agnostic way to check the browser assets, so you could do that here.

integration/cf-compiler-test.ts
```diff
@@ -40,6 +40,13 @@ test.describe("cloudflare compiler", () => {
           import { content as browserPackage } from "browser-pkg";
           import { content as esmOnlyPackage } from "esm-only-pkg";
           import { content as cjsOnlyPackage } from "cjs-only-pkg";
+          import hooks from "node:async_hooks";
```
pcattori, line 43: @jplhomer looking good!

I think the last thing we need is to add a check in that same integration test to make sure node:async_hooks doesn't end up in the browser package. We've used shelljs's grep as a platform-agnostic way to check the browser assets, so you could do that here.

pcattori (changes requested):

integration/cf-compiler-test.ts
jplhomer: Added a test for this! Granted, they aren't included because the import isn't being used in a browser context and thus is being stripped out as server code. Presumably, if the user attempts to import it into e.g. a React component, the bundle would break just like any other node external?

pcattori (approved):

PR #5785, opened 2023-03-13
API updates for v2 `meta` arguments and types

chaance:
This PR does a few things we'll want to include behind the `v2_meta` flag before the major release.

- `V2_HtmlMetaDescriptor` has been renamed to `V2_MetaDescriptor`. This is shorter but also more aligned with the `links` function return value type, which is `LinkDescriptor`. It also makes a bit more sense as we support values for other HTML tags than `<meta>`.
- `V2_HtmlMetaDescriptor` and `V2_HtmlMetaFunction` types should be imported from `@remix-run/react` instead of `@remix-run/<runtime>` since A) the primary responsibility of `meta` is rendering, and B) `matches` data only makes sense in the context of an app calling `useMatches`. I think we should follow suit here with `LinksFunction` in v2, but that's for a separate PR.
- The `meta` function's arguments have been simplified
  - `parentsData` has been removed, as each route's loader data is available on the `data` property of its respective `match` object
    ```tsx
    // before
    export function meta({ parentsData }) {
      return {
        title: parentsData['routes/some-route'].title,
      };
    };
    // after
    export function meta({ matches }) {
      return {
        title: matches.find((match) => match.id === 'routes/some-route').data.title,
      };
    };
    ```
  - The `route` property on route matches has been removed, as relevant match data is attached directly to the match object
    ```tsx
    // before
    export function meta({ parentsData }) {
      let rootModule = matches.find(match => match.route.id === "root");
    };
    // after
    export function meta({ matches }) {
      let rootModule = matches.find(match => match.id === "root");
    };
    ```


docs/route/meta.md
```diff
@@ -150,7 +150,7 @@ export const meta: V2_MetaFunction = () => {
 };
 ```
 
-Meta functions return an array of `V2_HtmlMetaDescriptor` objects. These objects map one-to-one with normal HTML meta tags:
+Meta functions return an array of `V2_MetaDescriptor` objects. These objects map one-to-one with normal HTML meta tags:
```
ryanflorence, line 153: > normal HTML tags

since we have links/script too

packages/remix-react/components.tsx
```diff
@@ -568,12 +571,11 @@ function V1Meta() {
     if (routeModule.meta) {
       let routeMeta =
         typeof routeModule.meta === "function"
-          ? routeModule.meta({
+          ? (routeModule.meta as V1_MetaFunction)({
```
ryanflorence, line 574: shouldn't this type already come from the routeModule type? we don't have to cast the other types of this object, why is this property special?

packages/remix-react/components.tsx
```diff
@@ -651,39 +653,28 @@ function V1Meta() {
...
-
-    let routeMeta: V2_HtmlMetaDescriptor[] | V1_HtmlMetaDescriptor | undefined =
-      [];
+    let routeMeta: V2_MetaDescriptor[] | V1_HtmlMetaDescriptor | undefined = [];
 
     if (routeModule?.meta) {
       routeMeta =
         typeof routeModule.meta === "function"
-          ? routeModule.meta({
+          ? (routeModule.meta as V2_MetaFunction)({
```
ryanflorence, line 673: same question here as above

ryanflorence (approved):
just some questions, nothing blocking though.

packages/remix-react/components.tsx
chaance: Because it's a union type on the route module, TS doesn't know how to disambiguate at the call-site.

Issue #5786, opened 2023-03-14
V2_MetaFunction type incorrect for error boundaries


kentcdodds:
### What version of Remix are you using?

1.14.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. Configure remix for the latest meta and error boundary features
2. Make a route that throws an error or response in the loader
3. Add a meta export that uses the `V2_MetaFunction` generic and pass `<typeof loader>`
4. Use `data.whatever` and see that the types are totally cool with that
5. Hit the route with the thrown error and see the runtime exception.

### Expected Behavior

The types should warn that the `data` property could be `null`.

### Actual Behavior

Meta functions are called for error boundaries so the `data` property that's passed could be `null`. The types don't allow for this currently.

kentcdodds:
I made a PR which may help explain what I need: https://github.com/remix-run/remix/pull/5787

brophdawg11:
Fixed by #6231, will be available in the next release (likely `1.17.0`)

PR #5787, opened 2023-03-14
fix(typescript): Allow meta data and parentsData to be undefined

kentcdodds:
Closes: #5786

~~- [ ] Docs~~
~~- [ ] Tests~~

Testing Strategy:

I tried it locally and it seems to work 👍


john-persson-polestar (approved):

aaronadamsCA:
IMO this would be well worth a patch release, because so far migrating to meta v2 is quite dangerous without it.

machour (changes requested):
@kentcdodds could you fix the conflicts in order to get this PR merged?

I also noticed the same problem in `packages/remix-react/routeModules.ts`, not sure why this definitions are duplicated there, but they'll need fixing too. (I had #6231 open to fix it before noticing your PR)

kentcdodds:
This is untested. Think of this as more of an issue than a PR.

PR #5796, opened 2023-03-14
remove codepaths for v1 meta API

chaance:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


brophdawg11 (approved):
👏 Minor testing comment/question - wonder if we should be removing meta from tests that don't assert against it instead of migrating it v1 -> v2?  

Looks good otherwise - just needs a changeset

PR #5800, opened 2023-03-14
feat!(remix-dev): remove `unstable_cssModules`

markdalgleish:
Removes the `unstable_cssModules` future flag and automatically enables CSS Modules. This means that CSS bundling is no longer conditional within the Remix compiler.


MichaelDeBoey (approved):
Should we also upgrade it to `v2_cssModules` on `dev`? 🤔

PR #5801, opened 2023-03-15
Fix v2 type error by bumping react-router-dom to experimental release

markdalgleish:
This fixes an issue with mismatched types between different versions of react-router-dom.


chaance (approved):

MichaelDeBoey (approved):

markdalgleish:
This PR fixes the type error, but now it looks like we're hitting an unrelated issue with the integration tests that's slipped into `dev`, so I'll merge this even though it's broken so this branch is in a similar state to `dev`.

PR #5813, opened 2023-03-15
docs(pages/technical-explanation): fix wording

pavel-romanov8:
Hi folks! I update documentation for `technical-explanations.md`. It felt like the word was missing while reading it


MichaelDeBoey:
Going to close this as the current wording is how it should be

Nevertheless, thanks for taking the time for this contribution @pavel-romanov8!

PR #5815, opened 2023-03-15
Add future.v2_normalizeFormMethod flag

brophdawg11:
**Note: CI will fail until we point this to an RR experimental/prerelease version with these changes**

Add a future flag to normalize/align `useNavigation().formMethod`.  We want our API to behave like `fetch()` - it can accept lowercase or uppercase, but exposes the method back to you as a normalized uppercase method:

```js
new Request("/", { method: "post" }).method === "POST" // true
new Request("/", { method: "POST" }).method === "POST" // true
```

So this means that both `<Form method="post">` and `<Form method="POST">` should result in `useNavigation().formMethod === "POST"`.

We inadvertently shipped lowercase `useNavigation` values in RR 6.4, and only noticed the inconsistency when we went to deprecate `useTransition` in Remix which used uppercase.  Therefore, we're putting this behind a future flag to avoid breaking apps, and in Remix v2 we'll move to uppercase as the default.


mcansh (approved):

MichaelDeBoey (approved):

packages/remix-react/browser.tsx
```diff
@@ -165,7 +165,18 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+      future: {
+        // Pass through the Remix future flag to avoid a v1 breaking change in
+        // useNavigation() - users can control the casing via the flag in v1.
+        // useFetcher still always uppercases in the back-compat layer in v1.
+        // In v2 we can just always pass true here and remove the back-compat
+        // layer
+        v7_normalizeFormMethod:
+          window.__remixContext.future.v2_normalizeFormMethod,
+      },
+    });
```
brophdawg11, line 179: Note to self add this to `createStaticRouter` as well

brophdawg11:
Cherry-picked the new future flags docs from `main` into here so I could add the new flag to them

PR #5821, opened 2023-03-15
feat(remix-dev): upgrade `unstable_cssModules` config option to `v2_cssModules`

MichaelDeBoey:
Follow-up of @markdalgleish's #5800

### DIFF
--- a/docs/guides/styling.md
+++ b/docs/guides/styling.md
@@ -935,13 +935,13 @@ With this link tag inserted into the page, you're now ready to start using the v
 
 First, ensure you've set up [CSS bundling][css-bundling] in your application.
 
-Then, to enable [CSS Modules], set the `future.unstable_cssModules` feature flag in `remix.config.js`.
+Then, to enable [CSS Modules], set the `future.v2_cssModules` feature flag in `remix.config.js`.
 
 ```js filename=remix.config.js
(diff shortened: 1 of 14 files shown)


markdalgleish:
I'm not sure if we want to introduce churn around this flag just before v2. Do you mind if we close this for now and potentially come back to it?

MichaelDeBoey:
@markdalgleish As pointed out in @brophdawg11's #5829, the way to go is to upgrade `unstable_*` flags to `v2_*` once the API is becoming stable, so that's why I created this PR

chaance:
> @markdalgleish As pointed out in @brophdawg11's #5829, the way to go is to upgrade `unstable_*` flags to `v2_*` once the API is becoming stable, so that's why I created this PR

This is generally the idea, but I think with the CSS stuff it might not be worth bumping to `v2_` for brand new features that don't break or impact current APIs. Matt's comment was more about how we approach new major-version changes to APIs in the future. New experimental/unstable features can stabilize at any point and be merged before a major release so long as nothing is breaking!

MichaelDeBoey:
So basically `unstable_cssModules` should actually be released in `dev` instead of `v2` in future scenarios, but for this time the team opted to do it in `v2`?

PR #5822, opened 2023-03-16
templates: use TS `server` file

MichaelDeBoey:
We use TS in all our docs & examples, so let's make these templates as much TS as possible

We have a TS `server` file in our grunge stack as well


templates/netlify/server.ts
MichaelDeBoey, line 1: Since @anmonteiro created the previous update to this file (see #1835), it would be nice if they can tell us if this is the correct way of doing things 🤔

anmonteiro: LGTM. cc @dwwoelfel and @sgrove

dwwoelfel: @MichaelDeBoey unfortunately Netlify has decided to end support for Netlify Graph and it will go away on May 1st.

It should be fine to still call these methods after the EOL--the token will just stop getting added to the context object and it will always return null. Then in a future version of `@netlify/functions`, the graph helpers will be removed.

Would you like us to submit a PR to remove the netlify graph features?

MichaelDeBoey: @dwwoelfel figuring out the `netlifyGraphToken` was added by @anmonteiro in #1835 to allow easy access to third-party APIs via the Onegraph acquisition.

If it's not necessary anymore, it's fine for me to remove it from the template as well
Or if there's a total new way of doing so, I'm happy to update the template too

Since you all are part of the Netlify team, I think you are best positioned to make the decision on what to do here

dwwoelfel: I would recommend removing it from the template. 

I'm not finding any usage of netlify graph features on remix sites that would give me concern that anyone's site would break if they updated to a newer version of remix.

MichaelDeBoey: I suppose you would also remove the `clientNetlifyGraphAccessToken` & `netlifyGraphSignature` keys?

MichaelDeBoey: @dwwoelfel I suppose you would also remove the `clientNetlifyGraphAccessToken` & `netlifyGraphSignature` keys?

mcansh (approved):

templates/cloudflare-pages/server.ts
```diff
@@ -7,6 +7,6 @@ const handleRequest = createPagesFunctionHandler({
   getLoadContext: (context) => context.env,
 });
 
-export function onRequest(context) {
+export function onRequest(context: EventContext<any, any, any>) {
```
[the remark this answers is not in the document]
MichaelDeBoey: Since `onRequest` just calls `handleRequest`, I'm just going to replace it with the implementation of `onRequest`, as we do that in other `server.ts` files as well

jacob-ebey:
Thanks for this @MichaelDeBoey, been wanting to do this for a while.

rossipedia:
This broke `create-remix` for express :(

lili21:
error message after run `npm run dev` in express template

```bash
[nodemon] starting `cross-env NODE_ENV=production node ./server.js --require dotenv/config ./server.js`
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module '/Users/admin/just-for-fun/try-remix15/server.js'
```

MichaelDeBoey:
@rossipedia @lili21 I'm fixing this in #6145
Feedback is welcome over there

PR #5831, opened 2023-03-16
ci: release commenting overhaul

mcansh:
we now only hit github for the following things:
- asking if a commit is a PR (this will return the PR so we get the number, title, and body which we use)
- getting any linked issues
- commenting on issues and PRs

also took now as an opportunity to break this out to a proper re-usable GitHub Action https://github.com/mcansh/release-comment-action

an example run can be found in the test repo: https://github.com/mcansh/nightly-release-test/pull/180

React Router PR for this -> https://github.com/remix-run/react-router/pull/10265

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


brophdawg11:
> also took now as an opportunity to break this out to a proper re-usable GitHub Action

Dear God, It's Beautiful

.github/workflows/release-comments.yml
```diff
@@ -2,44 +2,22 @@ name: 📝 Comment on Release
...
         env:
+          DIRECTORY_TO_CHECK: "./packages"
+          PACKAGE_VERSION_TO_FOLLOW: "remix"
           GITHUB_REPOSITORY: ${{ github.repository }}
-          GITHUB_TOKEN: ${{ github.token }}
-          VERSION: ${{ inputs.ref }}
-          DEFAULT_BRANCH: "main"
-          RELEASE_BRANCH: ${{ inputs.release_branch }}
-          PACKAGE_VERSION_TO_FOLLOW: ${{ inputs.package_version_to_follow }}
+          GH_TOKEN: ${{ github.token }}
```
MichaelDeBoey, line 23: Same comment as https://github.com/remix-run/react-router/pull/10265#discussion_r1148373603

> What about adding this as input instead of ENV variables?
> 
> This way you can set defaults for at least `GITHUB_REPOSITORY` & `GH_TOKEN` & you don't need to pass them anymore as the user of the Action
> You can see an example of having this for `GH_TOKEN` here:
> https://github.com/actions/labeler/blob/7012d51fe062f0b1e26e224a3c9fb52598ee3302/action.yml#L5-L8

mcansh: good suggestion

mcansh: we'll still need GH_TOKEN to be an environment variable, which i guess we could consume as an input and manually set `process.env.GH_TOKEN = core.getInput("GH_TOKEN")`, but seems icky at first glance

PR #5847, opened 2023-03-20
docs(guides/migrating-react-router-app): add a warning notice

### LINKED ISSUE #4623: Should raise error when app/routes/index.jsx does not exist

1.4.3

### Steps to Reproduce

1. Clone https://github.com/WesleyYue/incremental-react-router-to-remix-upgrade-path
2. `npm install`
3. `npm run dev`

### Expected Behavior

There should be some kind of error that informs the user that they cannot use the catch-all route to render `/`. This is super non-obvious if migrating from react-router and the user will almost certainly have `/` as a path under their old react-router implementation. I waste a few hours trying to figure out why my migration didn't work and it's super frustrating to debug. The [migration guide](https://remix.run/docs/en/v1/guides/migrating-react-router-app) does not say anything about needing this index.[jsx/tsx].

### Actual Behavior

No error anywhere, just empty <body> rendered by the browser if you follow the migration guide step by step.

vcheeze:
Closes #4623

### DIFF
--- a/docs/guides/migrating-react-router-app.md
+++ b/docs/guides/migrating-react-router-app.md
@@ -184,7 +184,7 @@ Inside of your `$.jsx` file, all we need to do is export the code from our old r
 export { default } from "~/old-app/app";
 ```
 
-<docs-warning><strong>Important:</strong> make sure that you still have the `app/routes/index.[jsx|tsx]` file. Remix will not automatically render your `/` React Router fallback route, which means a blank page will be rendered without the index file in the `routes` directory.</docs-warning>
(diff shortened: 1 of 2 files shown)


docs/guides/migrating-react-router-app.md
```diff
@@ -184,6 +184,8 @@ Inside of your `$.jsx` file, all we need to do is export the code from our old r
 export { default } from "~/old-app/app";
 ```
 
+<docs-warning><strong>Important:</strong> make sure `app/routes/index.[jsx|tsx]` exists in your Remix app. Remix will not automatically render your `/` React Router fallback route, which means a blank page will be rendered without the index file in the `routes` directory.</docs-warning>
+
```
brophdawg11, line 188: Instead of using a warning here, could we just change the section to be "Create an Index and a Catch-all route" and export your old app (i.e., `<Routes>` tree) from both?

PR #5854, opened 2023-03-20
refactor(dev): separate subcompilers for brower js, server js, and css

pcattori:
The Remix compiler coordinates 3 separate `esbuild` compilations:
1. The JS assets for the browser (`browserjs`)
2. The CSS assets (`css`)
3. The JS for the server (`serverjs`)

This PR refactors the compiler code to isolate each of these compilations (and the plugins specific to each compilation) to their own namespaces.

Decoupling the `browserjs` and `css` leads to reduced responsibilities for both of those compilations.
Namely, manifest creation can be extricated from `browserjs` and handled by the top-level Remix compiler instead, eliminating the need for asynchronous coordination for the manifest.

This PR intentionally leaves both `browserjs/compiler.ts` and `css/compiler.ts` modules as copies of the old `compileBrowser.ts` module. Now each of these modules can be simplified and optimized independently.


mjackson (approved):
Looking good! 🙌

Issue #5845, opened 2023-03-19
HMR does not work for lazily imported components


rossipedia:
### What version of Remix are you using?

1.14.3

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. Create new Remix project with `npx create-remix@latest`
2. Enable HMR by following the [1.14 instructions](https://github.com/remix-run/remix/releases/tag/remix%401.14.0)
3. Add a new component, and lazily import it from `app/routes/index.tsx`:
  ```tsx
  import { lazy } from 'react';
  const LazyComponent = lazy(() => import('~/app/components/LazyComponent'));

  // add it to the component in `app/routes/index.tsx`:
  return (
    ...
    <LazyComponent /> 
    ...
  );
  ```
   
4. Edit `app/routes/index.tsx` and observer that HMR works correctly and the UI is updated (change a `style` prop or something similar).
5. Edit the lazy component (`~/app/components/LazyComponent.tsx` in this example) 

[Example Repository](https://github.com/rossipedia/remix-hmr-lazy-bug)

### Expected Behavior

HMR works for the lazily imported component and the UI is updated without a page refresh

### Actual Behavior

HMR does not work for the lazy component, and the UI is not updated

pcattori:
Cannot reproduce in 1.19.2. Changing `LazyComponent` (styles, markup, etc..) produces HMR'd content.

PR #5856, opened 2023-03-21
Add HMR support for dynamically imported modules.

rossipedia:
Closes: #5845 

- [ ] Docs
- [X] Tests

Testing Strategy:

I created a playground using `yarn playground:new`, created several dynamically imported components, and started editing them. Diving into how HMR was implemented, I noticed that dynamic imports weren't included in the route's manifest entries, so Remix had no way of knowing that they were dependencies of a given route module. But `esbuild` knows that they are, so when they were edited, esbuild picked up that the route module had changed, and correctly sent an HMR update message to the browser, which then dutifully re-imported the route module.

However, since dynamically imported modules become their own chunk, those were _not_ re-imported, and the browser's module cache still referenced the old code.

The implementation involved adding a `dynamicImports` array to the route manifest entry, and modifying the compiler to add any `kind == "dynamic-import"` nodes to that list. I didn't want to re-use `imports` because static imports end up being bundled into the route module, and so they don't really have their own entry in the manifest and thus no actual URL to re-import. I also wanted to make the distinction between static imports and dynamic imports explicit.

Then I updated the browser-side HMR code to check if the updated route module has any `dynamicImports`, and then re-import those before the route module itself re-imports, so that it can pick up the new changes when the new module is re-executed.

### Integration Tests

I added some code to `integration/hmr-test.ts` to verify that HMR works with dynamic imports.

### DIFF
--- a/packages/remix-dev/hmr.ts
+++ b/packages/remix-dev/hmr.ts
@@ -8,6 +8,7 @@ export type Update = {
   url: string;
   revalidate: boolean;
   reason: string;
+  dynamicImports?: string[];
 };
 
 // route id: filepaths relative to app/ dir without extension
--- a/packages/remix-dev/hmr.ts
+++ b/packages/remix-dev/hmr.ts
@@ -8,7 +8,6 @@ export type Update = {
   url: string;
   revalidate: boolean;
   reason: string;
-  dynamicImports?: string[];
 };
 
 // route id: filepaths relative to app/ dir without extension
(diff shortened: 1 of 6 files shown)


packages/remix-dev/hmr.ts
```diff
@@ -8,6 +8,7 @@ export type Update = {
   url: string;
   revalidate: boolean;
   reason: string;
+  dynamicImports?: string[];
```
rossipedia, line 11:
```suggestion
```

rossipedia:
I need to re-think this. As usual, the problem is more complex than it appears on the surface. I'll close this for now.

PR #5857, opened 2023-03-21
Bug: files in folder are resolved with status code 200 (routing v2)

BenjaVR:
Given these files:
```
├── routes
│   ├── _index
│   │   ├── route.jsx
│   ├── feature
│   │   ├── route.jsx
│   │   └── SomeComponent.jsx
```

I do expect to have only these routes:
- `/` (`_index/route.jsx`)
- `/feature` (`feature/route.jsx`)

When I try to navigate to e.g., `/feature/NonExisting` I receive a 404 as expected.

However, when I navigate to `/feature/SomeComponent` I receive a 200 response with the content of `feature/route.jsx`. As far as I understand this happens because I happen to have a file in my `feature` folder that has the same name as the route.

I think files like these should not impact the routing mechanism and should just return 404, because there is no explicit route defined? Or am I missing something here?

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -51,6 +51,7 @@
 - axel-habermaier
 - barclayd
 - BasixKOR
+- BenjaVR
 - BenMcH
 - benwis
 - bgschiller
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -1,6 +1,5 @@
 import { test, expect } from "@playwright/test";
 
-import { PlaywrightFixture } from "./helpers/playwright-fixture";
 import type { Fixture, AppFixture } from "./helpers/create-fixture";
 import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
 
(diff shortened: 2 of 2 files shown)


mcansh:
the tests appear to be passing as expected?

in any case we did some performance improvements for v2_routeConvention which shipped in a nightly release last night, you can try it out by changing your remix dependency versions to `0.0.0-nightly-4d86674-20230321`

BenjaVR:
@mcansh that's odd. When I run it locally (`yarn bug-report-test`) I get this output:

image

image

BenjaVR:
@mcansh same result. However, the failed test does not match with the snippet it shows 🤔 the test fails on line 98 (test with `SomeComponent`, but the snippet shows the line with `NotExisting`):

image

BenjaVR:
Alright, closing this as issues seem to be resolved in the new nightly (pre-release `1.15.0-pre.0`) by now. Thanks 👏

PR #5863, opened 2023-03-21
Add deprecation warnings for v2_normalizeFormMethod

brophdawg11:


packages/remix-react/browser.tsx
```diff
@@ -151,6 +151,16 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+    if (!window.__remixContext.future.v2_normalizeFormMethod) {
+      warnOnce(
+        false,
+        "⚠️  DEPRECATED: Please enable the `future.v2_normalizeFormMethod` flag to " +
+          "prepare for the Remix v2 release. Lowercase `useNavigation().formMethod`" +
+          "values are being normalized to uppercase in v2 to align with the `fetch()` " +
+          "behavior.  For more information, see https://remix.run/docs/hooks/use-navigation"
+      );
+    }
+
```
mcansh, line 163: ooh i didn't even think to include deprecations in the browser console

mcansh (approved):

packages/remix-react/browser.tsx
brophdawg11: yeah I've only done it for remix-react related flags so far.  Not yet convinced it makes sense to put say a flat routes deprecation log in the console

mcansh: looks like we also let this log in production 🙈 
image

mcansh: PR to fix #5874

PR #5926, opened 2023-03-28
docs(tutorials/blog): use new index routes

sezercik:
change all 'index.tsx' to '_index.tsx' because without underscore you can't use dynamic routes and in new version of remix default name of index files is _index


MichaelDeBoey:
Going to close this PR as there's more to be changed than renaming all `index.tsx` to `_index.tsx`

All the needed changes are done in #5959

Issue #5663, opened 2023-03-05
Infinite loop GET /__REMIX_ASSETS_MANIFEST 200 - - 


arnaudambro:
### What version of Remix are you using?

1.14.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Here is my package.json

```json
{
  "sideEffects": false,
  "scripts": {
    "build": "run-s build:*",
    "build:remix": "remix build",
    "release-sentry": "remix build --sourcemap && node ./node_modules/@sentry/remix/scripts/sentry-upload-sourcemaps.js && rm ./public/**/*.map",
    "dev": "run-p dev:*",
    "dev:build": "cross-env NODE_ENV=development remix dev",
    "dev:serve": "cross-env NODE_ENV=development nodemon --watch build node_modules/.bin/remix-serve build",
    "lint": "eslint --cache --cache-location ./node_modules/.cache/eslint .",
    "start": "remix-serve build",
    "clean-build-and-cache": "rm -rf ./.cache && rm -rf ./build && rm -rf ./public/build",
    "clean": "npm run clean-build-and-cache && npm install",
    "web:dev": "vite",
    "web:build": "vite build --emptyOutDir",
    "web:preview": "vite preview"
  },
  "dependencies": {
    "@remix-run/node": "^1.14.0",
    "@remix-run/react": "^1.14.0",
    "@remix-run/serve": "^1.14.0",
    "react": "^18.2.0",
    "react-dom": "^18.2.0",
  },
  "devDependencies": {
    "@remix-run/dev": "^1.14.0",
    "@remix-run/eslint-config": "^1.14.0",
  },
  "engines": {
    "node": ">=14"
  }
}
```

### Expected Behavior

Everything should be alright

### Actual Behavior

When refresh, there is an infinite call to GET /__REMIX_ASSETS_MANIFEST

machour:
What command are you using to run your project?

arnaudambro:
`npm run dev`

cmd-johnson:
I'm encountering this, too. I'm running the new dev server with the Remix App Server like instructed in the [release notes](https://github.com/remix-run/remix/releases/tag/remix%401.14.0).

I didn't have time yet to get to the bottom of this, but at least it looks like the requests aren't coming from the browser, because the network tab is empty.

[...]

PR #5939, opened 2023-03-28
fix GET /__REMIX_ASSETS_MANIFEST loop #5663

conundrumer:
Closes: #5663

- [ ] Docs
- [ ] Tests

Testing Strategy:

The loop occurs when you run the unstable dev server, make a change, and quickly undo that change. [See this for more details](https://github.com/remix-run/remix/issues/5663#issuecomment-1458390579)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -88,6 +88,7 @@
 - codymjarrett
 - colinhacks
 - confix
+- conundrumer
 - coryhouse
 - craigglennie
 - crismali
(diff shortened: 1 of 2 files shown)


pcattori:
Superceded by #6133 

@conundrumer thanks for your work here! We ended up switching to a different dev server architecture that doesn't rely on polling, so this should be fixed once the tagged PR above lands / is released.

PR #5934, opened 2023-03-28
V2 preparation docs

brophdawg11:
Replaces https://github.com/remix-run/remix/pull/5932 since the `release-next-ryan` branch seemed to be triggering some changesets CI workflows

PR #5976, opened 2023-04-01
docs(pages/v2): remove default values from `serverBuildTarget` section

MichaelDeBoey:
Follow-up of #5934

Since we're talking about replacement, I figured that we don't need to specify the default values in here in order to make it work as expected

### DIFF
--- a/docs/pages/v2.md
+++ b/docs/pages/v2.md
@@ -14,7 +14,7 @@ All v2 APIs and behaviors are available in v1 with [Future Flags][future-flags].
 
 You can keep using the old convention with `@remix-run/v1-route-convention` even after upgrading to v2 if you don't want to make the change right now (or ever, it's just a convention and you can use whatever file organization you prefer).
 
-```shellscript
+```sh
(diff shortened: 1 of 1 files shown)


brophdawg11:
I think the default values can be helpful in this context so you know the full extent of what the old thing was doing for you.  The main docs on `remix.config.js` do include the defaults so folks can figure out if they need all these fields if they want to?

MichaelDeBoey:
@brophdawg11 What about adding them as comments instead then?
I just don't want people to think they need to include _all_ these values in order to make it work.

brophdawg11:
Something like this?  

```js
serverModuleFormat: "cjs", // default value, can be removed
```

@jacob-ebey what do you think?

jacob-ebey:
@MichaelDeBoey, let's leave the values in the config but add the comment next to them for now like @brophdawg11 suggested.

MichaelDeBoey:
@brophdawg11 @jacob-ebey I've updated this PR as requested 👍

mcansh (approved):

PR #5987, opened 2023-04-02
fix(remix-vercel): export `EntryContext`

lifeiscontent:
not sure if this is the right place for it, but based off of this file: https://github.com/remix-run/examples/blob/main/twind/app/entry.server.tsx it looks like EntryContext should be imported from the runtime you're using.

### DIFF
--- a/packages/remix-vercel/index.ts
+++ b/packages/remix-vercel/index.ts
@@ -2,3 +2,4 @@ import "./globals";
 
 export type { GetLoadContextFunction, RequestHandler } from "./server";
 export { createRequestHandler } from "./server";
+export { EntryContext } from "@remix-run/server-runtime";


MichaelDeBoey:
Closing this as all runtimes are exporting `EntryContext`

https://github.com/remix-run/remix/blob/dcff4e81078e27a756655e0b5951b661a0d29959/packages/remix-cloudflare/index.ts#L52
https://github.com/remix-run/remix/blob/dcff4e81078e27a756655e0b5951b661a0d29959/packages/remix-deno/index.ts#L40
https://github.com/remix-run/remix/blob/dcff4e81078e27a756655e0b5951b661a0d29959/packages/remix-node/index.ts#L63

The Vercel adapter is build upon the Node runtime, so you should use it like
```ts
```

Just like the `twind` example you linked to is doing

PR #5991, opened 2023-04-03
docs(pages/v2): fix typo in `useFetcher` section

thomasverleye:
Also see #5968 but this PR is more complete as the import on the second example is also incorrect :-)


dmarkow:
Whoops, good catch -- I closed my PR in favor of yours.

MichaelDeBoey (approved):

PR #5995, opened 2023-04-03
fix(remix-eslint-config/internal): warn on `prefer-const` in code blocks in MD files

MichaelDeBoey:
We're showing `const` on all public code (both in docs & examples).

We already disabled `prefer-let/prefer-let` in code blocks in MD files, but didn't enable `prefer-const` to make sure we're always showing `const`


brophdawg11:
It doesn't look like we're fully consistent one way or another here.  This rule produces a bunch of warnings, but so does the inverse.  I know @ryanflorence is strong team `let` so he may want to see our docs go the other way.  I'm in favor of linter-enforced consistency one way or the other but don't have a strong preference on which way we go :)

MichaelDeBoey:
@brophdawg11 In the past, the consensus was "Use `let` in code & `const` in docs/examples", hence the disabling of `prefer-let/prefer-let` in docs (& before examples folder)
This PR makes that it's actually enforced as it's ment to be

brophdawg11:
Ah ok I wasn't aware there was a consensus.  I forgot we have an auto-formatter too that would fix the warnings I saw when I ran this locally.

brophdawg11 (approved):

PR #6005, opened 2023-04-05
Fix `V2_MetaFunction` type import in templates

TimonVS:
When I bootstrapped a new Remix project a few days ago, I noticed the type `V2_MetaFunction` was imported from the Node adapter `@remix-run/node`, which isn't installed because I chose to deploy to Cloudflare Pages. This is present in all templates. I also noticed that this type can be imported both from the adapter, and from `@remix-run/react`. I wasn't sure which is the correct one, maybe the adapter specific version is more correct, but I couldn't from a brief glance I didn't notice any difference in the types.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -516,3 +516,4 @@
 - zhe
 - arifqys
 - iamzee
+- TimonVS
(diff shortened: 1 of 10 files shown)


mcansh (approved):
thanks!

gonna need you to sign the CLA linked above before merging though :)

PR #6024, opened 2023-04-09
chore(templates/cloudflare-pages): ignore .dev.vars instead of .env

yudai-nkt:
This isn't a simple docs change but is still a small change in template, so I deleted the PR form you prepared. Let me know if that is inconvenient for you.

As is documented [here](https://remix.run/docs/en/1.15.0/guides/envvars), Cloudflare Pages uses `.dev.vars` for storing environment variables instead of `.env`. So ignoring `.env` isn't really helpful for Pages template.

### DIFF
--- a/templates/cloudflare-pages/.gitignore
+++ b/templates/cloudflare-pages/.gitignore
@@ -4,4 +4,4 @@ node_modules
 /functions/\[\[path\]\].js
 /functions/\[\[path\]\].js.map
 /public/build
-.env
+.dev.vars


yudai-nkt:
Hi Remix team, at which line should I sign? Contributors names look alphabetically sorted basically but the last 3 lines are not the cases.

MichaelDeBoey (approved):

MichaelDeBoey:
@yudai-nkt You can just add your name to the end of the file. CI will place them in alphabetical order.

yudai-nkt:
@MichaelDeBoey thanks for the answer and that's cool! Now I've signed.

MichaelDeBoey (approved):

PR #6032, opened 2023-04-10
templates: fix `V2_MetaFunction` import

MichaelDeBoey:
Follow-up of #6005

This import should come from the runtime, not the renderer

### DIFF
--- a/templates/arc/app/routes/_index.tsx
+++ b/templates/arc/app/routes/_index.tsx
@@ -1,4 +1,4 @@
-import type { V2_MetaFunction } from "@remix-run/react";
+import type { V2_MetaFunction } from "@remix-run/node";
 
 export const meta: V2_MetaFunction = () => {
   return [{ title: "New Remix App" }];
--- a/templates/cloudflare-pages/app/routes/_index.tsx
+++ b/templates/cloudflare-pages/app/routes/_index.tsx
@@ -1,4 +1,4 @@
-import type { V2_MetaFunction } from "@remix-run/react";
+import type { V2_MetaFunction } from "@remix-run/cloudflare";
 
 export const meta: V2_MetaFunction = () => {
   return [{ title: "New Remix App" }];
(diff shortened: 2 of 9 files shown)


brophdawg11:
I think importing from react is fine here because this is also a client-side concern.  The thing exported from server-runtime is `V2_ServerRuntimeMetaFunction`.  I'm not sure why it's named differently offhand but it looks close if not identical.  Eventually we want to align some of these duped types so I think I'd vote to hold off on changing this until we do that?

mcansh:
> I think importing from react is fine here because this is also a client-side concern. The thing exported from server-runtime is `V2_ServerRuntimeMetaFunction`. I'm not sure why it's named differently offhand but it looks close if not identical. Eventually we want to align some of these duped types so I think I'd vote to hold off on changing this until we do that?

name difference might be so that it doesn't show up in vscode's autocomplete

MichaelDeBoey:
@brophdawg11 Before #6005, we always imported this type from the runtime & if I remember correctly the plan was to have an extra package that contains more types/functions like this (for example `json` & `redirect`), so `react` is definitely not the correct place. The most correct place we currently have is our runtimes

brophdawg11:
I missed that the adapters rename the export (`V2_ServerRuntimeMetaFunction as V2_MetaFunction`), at first I thought this change wouldn't compile 😬 

I don't think we can say react is definitely not the correct place since `meta` runs both on the server and the client, unlike `json`/`redirect` which _only_ run on the server.  So while we do want to dedupe some types eventually I don't think that necessarily means `json` and `meta` will land in the same place.  

I don't have the background here but as Logan pointed out in #5785 it seems it was a specific choice to import from the react package?  I do know we're considering some API changes to meta/links in v2 (behind a v3 future flag) and these may just be heading in that direction so I'm inclined to leave this as is for now

MichaelDeBoey:
Closed by #6123

PR #6044, opened 2023-04-11
fix(deps): update `node-fetch`

MichaelDeBoey:
This should close https://github.com/remix-run/remix/security/dependabot/20 as now all versions are at `2.6.7` (which is a fixed version)

### DIFF
--- a/packages/remix-vercel/package.json
+++ b/packages/remix-vercel/package.json
@@ -18,8 +18,8 @@
   },
   "devDependencies": {
     "@types/supertest": "^2.0.10",
-    "@vercel/node": "^2.4.2",
-    "@vercel/node-bridge": "^3.0.0",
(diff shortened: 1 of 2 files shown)


mcansh:
this isn't updating `node-fetch` directly so i think the PR title is a bit misleading, but the changes LGTM

MichaelDeBoey:
@mcansh I started with deduping `node-fetch` in `yarn.lock` & then saw the only version of `node-fetch` that was left was because of `@vercel/node` (which has `node-fetch` as an exact dependency version)

So the meaning is still to update `node-fetch`, but I'm happy to give this PR another name if you think it would be better suited

MichaelDeBoey:
@mcansh I guess we should cherry-pick this one onto `main` as well?

PR #6056, opened 2023-04-13
Cookie secrets fix

atesgoral:
Found an edge case where intentionally setting the `secrets` property of `CookieSignatureOptions` to `undefined` causes a runtime error in `isSigned()`, `decodeCookieValue()` and `encodeCookieValue()`.

TypeScript doesn't catch the fact that `secrets` in the scope of `createCookieFactory()` can be `undefined`. TypeScript thinks that it is always a `string[]` because of the default value assigned before the splat, and doesn't take into account that the splatted object might have `secrets` set to `undefined`.

Moving the default value (`[]`) to the destructuring assignment fixes this issue.

(Is this a TypeScript deficiency?)

Testing Strategy:

Try overriding `secrets` to `undefined`:

```ts
const cookie = createCookie('COOKIE', {
  secrets: undefined,
});
```

Then call `cookie.isSigned()`.

Or run the newly added test without the fix. You'll get an error like this:

```
    TypeError: Cannot read properties of undefined (reading 'length')

      101 |       },
      102 |       get isSigned() {
    > 103 |         return secrets.length > 0;
          |                        ^
```


packages/remix-server-runtime/__tests__/cookies-test.ts
```diff
@@ -142,6 +142,18 @@ describe("cookies", () => {
     expect(setCookie).not.toEqual(setCookie2);
   });
 
+  it("makes the default secrets to be an empty array", async () => {
+    let cookie = createCookie("my-cookie");
+
+    expect(cookie.isSigned).toBe(false);
+
+    let cookie2 = createCookie("my-cookie2", {
+      secrets: undefined,
```
atesgoral, line 151: 💥

packages/remix-server-runtime/__tests__/cookies-test.ts
```diff
@@ -142,6 +142,18 @@ describe("cookies", () => {
...
+    });
+
+    expect(cookie2.isSigned).toBe(false);
```
atesgoral, line 154: 💥 This is where it would normally blow up

MichaelDeBoey:
@atesgoral Could you please rebase your branch onto latest `dev` & set the base branch of this PR to `dev` as well?

atesgoral:
@MichaelDeBoey done!

mcansh (approved):

PR #6065, opened 2023-04-13
feat(remix-testing): cast types to use Remix type definitions + allow passing context

### LINKED ISSUE #6054: Typings for createRemixStub should allow passing a regular loader

1.15.0

- [X] Yes

### Steps to Reproduce

Do this:

```tsx
import { json, type DataFunctionArgs } from "@remix-run/node";
import { unstable_createRemixStub as createRemixStub } from "@remix-run/testing";

export async function loader(args: DataFunctionArgs) {
  return json({ hi: "there" });
}

createRemixStub([
  {
    path: "/example",
    loader,
  },
]);
```

Here's [a TypeScript Playground for this](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgKwM4QHYBo4wJ5gCmcAIgIYxkBiArhgMYzCYCCUA5qnAL5wBmUCCDgAiAAJRCIYAA8AtFDoB6DBAAmhEQG4AUKEixEcOqkoAjADaEA+vUkVCAJSmyAyjBpm4ZLncIPnaRl3Tx5+QWFxSSCFZRhCU2AMdm0dHUIZA3gfPAZ+OkZmDDgLCDINKAAKMg5UAC5SCmoCplZagEpEHTg4SQ8oYrRMSqQAC2AGkRhRwkkRHnbdbjS-AJdgjzNKgG1urp6esApRyaUMsnArESw9ntLy2Zue7huAXUWdIA)

### Expected Behavior

The types should be fine with this.

### Actual Behavior

You get the following error:

```
Type '(args: DataFunctionArgs) => Promise<TypedResponse<{ hi: string; }>>' is not assignable to type 'LoaderFunction'.
  Types of parameters 'args' and 'args' are incompatible.
    Type 'LoaderFunctionArgs' is not assignable to type 'DataFunctionArgs'.
      Property 'context' is optional in type 'LoaderFunctionArgs' but required in type 'DataFunctionArgs'.
```

mcansh:
before if you tried to pass a loader/action directly to the route object it would fail as `context` is optional in RR, but required in Remix (types for each are also different, bug would remain if required), so you would have to manually pass `context: {}`

```ts
import { json, type DataFunctionArgs } from "@remix-run/node";
import { unstable_createRemixStub as createRemixStub } from "@remix-run/testing";

export async function loader(args: DataFunctionArgs) {
  return json({ hi: "there" });
}

createRemixStub([
  {
    path: "/example",
    loader: args => loader({ ...args, context: {} }),
  },
]);
```

now we're using the Remix types for loader/action, but we need to cast them on the way out back to RR for use with `createMemoryRouter`

```ts
import { json, type DataFunctionArgs } from "@remix-run/node";
import { unstable_createRemixStub as createRemixStub } from "@remix-run/testing";

export async function loader(args: DataFunctionArgs) {
  return json({ hi: "there" });
}

createRemixStub([
  {
    path: "/example",
    loader,
  },
]);
```

this PR also introduces the ability to pass `context` to your stubbed loaders/actions via a second argument to `createRemixStub`

```ts
unstable_createRemixStub(
  [
    {
      path: "/",
      index: true,
      element: <App />,
      loader({ context }) {
        return json(context);
      },
    },
  ],
  { context: "hello" }
);
```

closes #6054

- [ ] Docs
- [x] Tests

Testing Strategy: new tests

### DIFF
--- a/packages/remix-testing/create-remix-stub.tsx
+++ b/packages/remix-testing/create-remix-stub.tsx
@@ -9,9 +9,18 @@ import type {
   UNSAFE_RouteModules as RouteModules,
   UNSAFE_RemixContextObject as RemixContextObject,
 } from "@remix-run/react";
-import type { DataRouteObject, RouteObject } from "react-router-dom";
+import type {
+  DataRouteObject,
+  IndexRouteObject,
+  NonIndexRouteObject,
+  RouteObject,
+} from "react-router-dom";
 import { createMemoryRouter, RouterProvider } from "react-router-dom";
-import type { AppLoadContext } from "@remix-run/server-runtime";
+import type {
+  ActionFunction,
+  AppLoadContext,
+  LoaderFunction,
+} from "@remix-run/server-runtime";
 
 type RemixStubOptions = {
   /**
@@ -44,10 +53,10 @@ type RemixStubOptions = {
 };
 
 function patchRoutesWithContext(
-  route: (RouteObject | DataRouteObject)[],
+  routes: (StubRouteObject | StubDataRouteObject)[],
   context: AppLoadContext
 ): (RouteObject | DataRouteObject)[] {
-  return route.map((route) => {
+  return routes.map((route) => {
     if (route.children) {
       return {
         ...route,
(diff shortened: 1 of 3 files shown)


packages/remix-testing/create-remix-stub.tsx
```diff
@@ -42,7 +52,55 @@ type RemixStubOptions = {
...
+  routes: (StubRouteObject | StubDataRouteObject)[],
+  context: AppLoadContext
+): (RouteObject | DataRouteObject)[] {
+  return routes.map((route) => {
+    if (route.children) {
+      return {
+        ...route,
+        children: patchRoutesWithContext(route.children, context),
+      };
+    }
```
brophdawg11, line 65: Should this come after the loader/action patch so we don't miss patching the parent's loader/action?

mcansh: yep, had it that way originally, reset and then copilot decided it was to be first 😅

PR #6076, opened 2023-04-15
feat(remix-vercel): remove `@vercel/node` from `peerDependencies`

TrySound:
@vercel/node is a huge and quite overloaded library. It includes stale typescript version, old esbuild and own edge-runtime which is not need for most remix users.

You can see here install size https://packagephobia.com/result?p=%40vercel%2Fnode

And `@remix/vercel` use it only to import two types which are compatible with builtin node types.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy: CI should pass

### DIFF
--- a/packages/remix-vercel/server.ts
+++ b/packages/remix-vercel/server.ts
@@ -1,4 +1,7 @@
-import type { VercelRequest, VercelResponse } from "@vercel/node";
+import type {
+  IncomingMessage as VercelRequest,
+  ServerResponse as VercelResponse,
+} from "node:http";
 import type {
   AppLoadContext,
   ServerBuild,
--- a/contributors.yml
+++ b/contributors.yml
@@ -519,3 +519,4 @@
 - iamzee
 - TimonVS
 - yudai-nkt
+- TrySound
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "packageManager": "yarn@1.22.19",
   "name": "remix-monorepo",
   "private": true,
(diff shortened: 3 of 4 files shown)


packages/remix-vercel/server.ts
```diff
@@ -1,4 +1,7 @@
-import type { VercelRequest, VercelResponse } from "@vercel/node";
+import type {
+  IncomingMessage as VercelRequest,
+  ServerResponse as VercelResponse,
+} from "node:http";
```
MichaelDeBoey, line 4: This would be a breaking change, because TS will throw an error if people are implementing `getLoadContext` and using the added properties

```ts
export declare type VercelRequest = IncomingMessage & {
    query: VercelRequestQuery;
    cookies: VercelRequestCookies;
    body: VercelRequestBody;
};
export declare type VercelResponse = ServerResponse & {
    send: (body: any) => VercelResponse;
    json: (jsonBody: any) => VercelResponse;
    status: (statusCode: number) => VercelResponse;
    redirect: (statusOrUrl: string | number, url?: string) => VercelResponse;
};
```

TrySound: Now vercel users cannot upgrade remix until `@vercel/remix` package is bumped as well because it has fixed `@remix-run/node` dependency. Would be easier for everyone if remix just integrated all the stuff back from vercel fork.

MichaelDeBoey:
Going to close this as `@remix-run/vercel` will be officially deprecated once #5964 is released

PR #6084, opened 2023-04-16
docs(other-api/adapter) Improve Cloudflare Workers examples

xanderberkein:
This PR updates the Cloudflare Workers API examples to include `mode` in the `createRequestHandler`. This is needed with the unstable dev server (and harmless without it).

### DIFF
--- a/docs/other-api/adapter.md
+++ b/docs/other-api/adapter.md
@@ -135,7 +135,10 @@ import { createEventHandler } from "@remix-run/cloudflare-workers";
 
 import * as build from "../build";
 
-addEventListener("fetch", createEventHandler({ build }));
+addEventListener(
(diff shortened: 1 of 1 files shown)


brophdawg11:
Should we add this to all of the examples on this page?  cc @pcattori

xanderberkein:
@brophdawg11 The cloudflare adapters are actually the only adapters where `mode` is not defaulted to `process.env`.

However, the Cloudflare Workers adapter examples here are a bit outdated. Since `1.18.1`, the `cloudflare-workers` template doesn't use the `cloudflare-workers` adapter anymore, but it uses the `cloudflare` adapter directly. So the cloudflare-workers section of this page might need a full revisist.

PR #6102, opened 2023-04-19
docs(tutorials/jokes): Use npx when calling ts-node

dnsbty:
Simple docs change

The jokes tutorial tells the reader to install ts-node to the project dependencies, but then to run the seed script as though it's installed globally. This just adds `npx` before the ts-node seed script command because that's needed for users (like me) who don't have ts-node installed globally. This will probably be especially helpful for those coming from a non-frontend background.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -117,6 +117,7 @@
 - dima-takoy
 - dmarkow
 - DNLHC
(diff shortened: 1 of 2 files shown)


machour (approved):
Thank you!

PR #6038, opened 2023-04-11
chore: bump deps to resolve some security issues

mcansh:
closes #6002
closes https://github.com/remix-run/remix/security/dependabot/20 
closes https://github.com/remix-run/remix/security/dependabot/153

☝️ will automatically close when they end up on `main`, so we can cherry-pick this PR over after-the-merge

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

PR #6108, opened 2023-04-19
chore: dedupe `yarn.lock`

MichaelDeBoey:
This will make PRs like https://github.com/remix-run/remix/pull/6038 or https://github.com/remix-run/remix/pull/6044 less necessary

Did this by running `npx yarn-deduplicate && rm -rf ./node_modules && yarn`

- https://github.com/scinos/yarn-deduplicate
- https://npm.im/yarn-deduplicate

### DIFF
--- a/packages/remix-node/upload/fileUploadHandler.ts
+++ b/packages/remix-node/upload/fileUploadHandler.ts
@@ -150,14 +150,22 @@ export function createFileUploadHandler({
       }
     }
 
-    return new NodeOnDiskFile(filepath, contentType);
+    // TODO: remove this typecast once TS fixed File class regression
+    //  https://github.com/microsoft/TypeScript/issues/52166
+    return new NodeOnDiskFile(filepath, contentType) as unknown as File;
   };
 }
 
-export class NodeOnDiskFile implements File {
+// TODO: remove this `Omit` usage once TS fixed File class regression
+//  https://github.com/microsoft/TypeScript/issues/52166
+export class NodeOnDiskFile implements Omit<File, "constructor"> {
   name: string;
   lastModified: number = 0;
   webkitRelativePath: string = "";
+
+  // TODO: remove this property once TS fixed File class regression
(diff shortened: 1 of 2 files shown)


mcansh:
sweet!

mcansh:
might he nice to have this run on push with changes to ./packages/*/package.json and commit if there's a diff

MichaelDeBoey:
> might he nice to have this run on push with changes to ./packages/*/package.json and commit if there's a diff

I think we just need to check against changes in `yarn.lock` tbh 🤔

We could do that in a follow-up PR I think?

mcansh (approved):

MichaelDeBoey:
@mcansh They seem to be gone if we keep `@types/node@*` at v16 🤔

Should we fix the code or keep it at v16 (for now) until we are fully transitioned to v18 in @jacob-ebey's #5868?

MichaelDeBoey:
@mcansh The problem seemed to be that `NodeOnDiskFile` doesn't have a `prototype` property, but after implementing it, we still get
> Type 'NodeOnDiskFile' provides no match for the signature 'new (blobParts?: BlobPart[] | undefined, options?: BlobPropertyBag | undefined): Blob'.

Since @jacob-ebey is the creator of the class, maybe he can point me/us into the right direction here
I tried a lot, but can't seem to find a solution to make this work tbh 🤔

MichaelDeBoey:
@mcansh It appears @dpobel already has the TS error about the `prototype` property on a fresh install (see https://github.com/remix-run/remix/issues/6119#issuecomment-1521288835), so we should definitely find a solution here 👀

MichaelDeBoey:
@brophdawg11 I managed to (temporarily) "fix" the type errors with `NodeOnDiskFile` that were caused due to the TS regression with `File` class (see https://github.com/microsoft/TypeScript/issues/52166)

These changes are only here to make TS happy again and have no impact on the actual code itself

MichaelDeBoey:
Merging this one because @jacob-ebey approved #7044

Issue #6072, opened 2023-04-15
Infinite loop when using fetcher.load in latest Remix releases


BrycePearce:
### What version of Remix are you using?

1.14.0 (also on 1.15)

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

**1.)** Create a resource-route (or any loader) that returns anything. In this case I'm returning `json({hello: 'world'})`
**2.)** In any route, create a state variable to track to see if it has been invoked. For example

```js
  let fetcher = useFetcher();
  const [shouldCallResourceRoute, setShouldCallResourceRoute] = useState(true);

  useEffect(() => {
    if (shouldCallResourceRoute) {
      setShouldCallResourceRoute(false); // this should cause this fetcher to never load again. But it infinite loops
      fetcher.load("/resource-routes/resource");
    }
  }, [fetcher, shouldCallResourceRoute]);
````
**3.)** Note the infinite loop / crash that happens
![image](https://user-images.githubusercontent.com/16729071/232183037-fd12da86-7379-404d-8138-708bad6f8085.png)

### **-Note-**

 I am upgrading my Remix from an earlier version where my existing code worked fine. This bug only appeared after upgrading and I was able to replicate in a code sandbox here:

https://codesandbox.io/p/sandbox/hopeful-night-z2lsum?file=%2Fapp%2Froutes%2Findex.tsx&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A7%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A7%7D%5D

### Expected Behavior

The fetcher should only load once

### Actual Behavior

An infinite loop occurs, crashing the page.

chuanhhoang:
I face this problem too. Any update on this?

BrycePearce:
> I face this problem too. Any update on this?

I believe this is an unlisted bug/change that only affects the latest version. However there is a work around provided by brocktho in the discord chat

```js
  useEffect(() => {
    if (shouldFetch && fetcher.state === 'idle') {
      fetcher.load('/fetchRoute');
      setShouldFetch(false);
    }
  }, [fetcher, shouldFetch]);
  ```
  https://stackblitz.com/edit/remix-run-remix-ahjkhw?file=app%2Froutes%2F_index.tsx,app%2Fentry.server.tsx
  
  ----

However as an unlisted change that breaks previous code, I still believe this is a bug

pkasarda:
[...]

PR #6121, opened 2023-04-20
fix(remix-react): adopt `react-router`'s removal of `useSyncExternalStore`

brophdawg11:
Sibling PR in React Router that contains the explanation: https://github.com/remix-run/react-router/pull/10377

Closes: #6072


jacob-ebey (approved):

packages/remix-react/routes.tsx
```diff
@@ -151,8 +151,8 @@ export function createClientRoutes(
       // handle gets added in via useMatches since we aren't guaranteed to
       // have the route module available here
       handle: undefined,
-      loader: createDataFunction(route, routeModulesCache, false),
-      action: createDataFunction(route, routeModulesCache, true),
+      loader: createLoaderFunction(route, routeModulesCache),
+      action: createActionFunction(route, routeModulesCache),
```
brophdawg11, line 155: If we do remove `payload` from `action` we shouldn't need these changes

PR #6129, opened 2023-04-21
Add an explicit .file extension loader

ryanwilsonperkin:
Workaround for https://github.com/remix-run/remix/discussions/4549

As a shortcut around the proposal to use custom import assertions (https://github.com/remix-run/remix/discussions/4549) we can use a .file extension in the list of loaders to allow people to load whatever file types they'd like by adding a `.file` suffix on their file name

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -429,6 +429,7 @@
 - ryanflorence
 - ryanjames1729
 - ryankshaw
+- ryanwilsonperkin
 - sandulat
 - sbernheim4
 - schpet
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
As mentioned in the ["Remix Vite is Now Stable" blogpost](https://remix.run/blog/remix-vite-stable#next-steps), we won't add any new compiler features/improvements that aren't targeting Vite

> [...] from now on all new features and improvements that require compiler integrations will only be targeting Vite

PR #6130, opened 2023-04-21
Fix splat route index matching

### LINKED ISSUE #5841: Splat Route doesn't handle root index path.

1.14.3

- [X] Yes

### Steps to Reproduce

1. `npx create-remix@latest`
2. Rename `app/routes/index.tsx` to `app/routes/$.tsx`
3. `npm run dev`
4. Open http://localhost:3000/

### Expected Behavior

The default index route displays.

### Actual Behavior

Nothing is shown.

See the behavior [here](https://stackblitz.com/edit/remix-run-remix-ch5vc2?file=app/routes/$.tsx)

brophdawg11:
Root splat routes should match `/` if an index route is not present.  I.e.:

```
routes/
  $.tsx
```

Closes #5841


ryanflorence (approved):

integration/revalidate-test.ts
```diff
@@ -44,6 +44,11 @@ test.describe("Revalidation", () => {
             }
           `,
 
+          "app/routes/_index.jsx": js`
+            export default function Component() {
+              return <h1>Index</h1>;
+            }
+          `,
```
brophdawg11, line 51: Needed an index route in this existing test.  Putting `path:""` on the root previously meant that `/` would match the root route without it having an `index.jsx` or `$.jsx` file, which seems like a related bug that's also fixed here

PR #6136, opened 2023-04-21
templates: explicit installGlobals call in server entries

jacob-ebey:


sergiodxa (approved):
Likes this, it makes sense and will allow for apps running on modern Node versions to don’t install Remix globals and use Node ones.

MichaelDeBoey (approved):

jacob-ebey:
> Likes this, it makes sense and will allow for apps running on modern Node versions to don’t install Remix globals and use Node ones.

Not until v2 can remove the side-effect in the adapter packages, but it paves the path for it.

PR #6150, opened 2023-04-24
chore: update how `REMIX_LOCAL_BUILD_DIRECTORY` is resolved

mcansh:
```diff
- REMIX_LOCAL_BUILD_DIRECTORY=../../../../some-app yarn build --tsc
+ REMIX_LOCAL_BUILD_DIRECTORY=~/some-app yarn build --tsc
```

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


rollup.utils.js
```diff
@@ -10,7 +10,7 @@ const REPO_ROOT_DIR = __dirname;
 
 let activeOutputDir = "build";
 if (process.env.REMIX_LOCAL_BUILD_DIRECTORY) {
-  let appDir = path.join(
+  let appDir = path.resolve(
     process.cwd(),
```
pcattori, line 14:
```suggestion
```

pcattori (changes requested):

brophdawg11:
It doesn't look like `~` expansion will work quite right, but I think that's fine - I always use relative pathing anyway (https://stackoverflow.com/questions/21077670/expanding-resolving-in-node-js).

I updated this to use just `LOCAL_BUILD_DIRECTORY` instead of `REMIX_LOCAL_BUILD_DIRECTORY` as well to align with React Router (and shorted it by a few chars)

brophdawg11 (approved):

pcattori (approved):

PR #6178, opened 2023-04-26
chore: remove `styfle/cancel-workflow-action` usage

MichaelDeBoey:
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
- https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency

### DIFF
--- a/.github/workflows/deployments.yml
+++ b/.github/workflows/deployments.yml
@@ -28,15 +28,16 @@ on:
       TEST_VERCEL_USER_ID:
         required: true
 
+concurrency:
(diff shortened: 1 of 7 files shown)


mcansh (approved):
thanks! I've been meaning to do this for a bit

MichaelDeBoey:
@mcansh Only new commits on the same branch should cancel previous runs of a specific workflow (just like it did with `styfle/cancel-workflow-action`)

PR #5836, opened 2023-03-17
feat(remix-server-runtime): pass `AppLoadContext` to `handleRequest`

fernandojbf:
This is the same feature change request done [here](https://github.com/remix-run/remix/pull/2045). A new [discussion](https://github.com/remix-run/remix/discussions/5314) was created to this.

My action was mainly check the changes by introduced @axel-habermaier's PR and apply them again. 

This is my first contribution, if i made something wrong please let me know.

Closes: #

- [x] Docs
- [x] Tests

Testing Strategy:

I've added unit tests on server-runtime passing loadContext to the handler.

PR #6190, opened 2023-04-26
templates: have `loadContext` available in `handleRequest`

MichaelDeBoey:
Follow-up of #5836

### DIFF
--- a/docs/guides/migrating-react-router-app.md
+++ b/docs/guides/migrating-react-router-app.md
@@ -48,7 +48,10 @@ Let's start by creating two new files:
 ```tsx filename=app/entry.server.tsx
 import { PassThrough } from "stream";
 
-import type { EntryContext } from "@remix-run/node";
+import type {
+  AppLoadContext,
+  EntryContext,
+} from "@remix-run/node";
 import { Response } from "@remix-run/node";
 import { RemixServer } from "@remix-run/react";
 import isbot from "isbot";
@@ -60,7 +63,8 @@ export default function handleRequest(
   request: Request,
   responseStatusCode: number,
   responseHeaders: Headers,
-  remixContext: EntryContext
+  remixContext: EntryContext,
+  loadContext: AppLoadContext
 ) {
   return isbot(request.headers.get("user-agent"))
     ? handleBotRequest(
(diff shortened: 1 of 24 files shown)


kiliman:
Any chance we can fix the "typo" before v2 launches?

`getLoadContext` should be `getLoaderContext` or, even better, `getAppContext`.

docs/guides/migrating-react-router-app.md
```diff
@@ -60,28 +60,32 @@ export default function handleRequest(
...
       );
 }
 
 function handleBotRequest(
   request: Request,
   responseStatusCode: number,
   responseHeaders: Headers,
-  remixContext: EntryContext
+  remixContext: EntryContext,
+  _loadContext: AppLoadContext
```
brophdawg11, line 88: I don't think these need the underscore since they're not defined inside of `handleRequest`, so they're not in the same scope as the other `loadContext`?

MichaelDeBoey: @brophdawg11 We are passing `loadContext` from `handleRequest` to `handleBotRequest` & `handleBrowserRequest` with this PR, but it's not used in `handleBotRequest` & `handleBrowserRequest`, hence the underscore

brophdawg11: I don't know that we need to bother passing it through to `handleBotRequest`/`handleBrowserRequest` then?  And I don't know that our templates should need to prefix the variable if it's unused.  The intention of having it there is so users can use if if they need/want to - and it feels wrong for them to have to change the name when they do so - or impart a naming preference on their newly created app?

Can we just simplify al templates to to the following and avoid touching the usility funcitons if they don't ever need to touch the contedxt?

```jsx
  request: Request,
  responseStatusCode: number,
  responseHeaders: Headers,
  remixContext: EntryContext,
  loadContext: AppLoadContext
) {
    ? handleBotRequest(
        request,
        responseStatusCode,
        responseHeaders,
        remixContext
      )
    : handleBrowserRequest(
        request,
        responseStatusCode,
        responseHeaders,
        remixContext
      );
}
```

MichaelDeBoey: Should be as requested now 👍

brophdawg11 (approved):

PR #6202, opened 2023-04-26
fix(dev): show first compilation error when build fails

pcattori:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/compiler/compiler.ts
+++ b/packages/remix-dev/compiler/compiler.ts
@@ -44,9 +44,13 @@ export let create = async (ctx: Context): Promise<Compiler> => {
   };
 
   let compile = async () => {
-    let errCancel = (error: unknown) => {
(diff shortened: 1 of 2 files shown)


packages/remix-dev/compiler/compiler.ts
```diff
@@ -69,7 +73,7 @@ export let create = async (ctx: Context): Promise<Compiler> => {
 
     // css compilation
     let css = await tasks.css;
-    if (!css.ok) throw css.error;
+    if (!css.ok) throw error ?? css.error;
```
jacob-ebey, line 76: Can this just be `if (error) throw error`? The error is always assigned, so if the thing you awaited above errors, it will be there, and if anything else errors, also throw. Gets us out of here quicker and avoids future work if we something at all fails.

PR #6213, opened 2023-04-27
feat(remix-cloudflare): support `@cloudflare/workers-types` v4

jacob-ebey:
chore: update CF worker template to be ESM + use new dev server

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey (commented):
As seen in https://github.com/remix-run/remix/pull/5425, this is a breaking change as v4 types aren't compatible with v2 & v3

Hence why I dropped support for v2 & v3 in https://github.com/remix-run/remix/pull/5732 (which will be released in Remix v2)

PR #6229, opened 2023-04-28
template: v2 dev

jacob-ebey:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


templates/cloudflare-workers/remix.config.js
```diff
@@ -1,6 +1,5 @@
 /** @type {import('@remix-run/dev').AppConfig} */
-module.exports = {
-  devServerBroadcastDelay: 1000,
+export default {
   ignoredRouteFiles: ["**/.*"],
   server: "./server.ts",
   serverConditions: ["worker"],
```
huw, line 5: I think this should be `["workerd", "worker", "browser"]`:

- `workerd` is the new standard for Cloudflare Workers and could have slightly different code, usually optimised for edge runtimes. This is being adopted by React [with a different bundle from their browser bundle](https://github.com/facebook/react/blob/559e83aebb2026035d47aa0ebf842f78d4cd6757/packages/react-dom/package.json#L52-L58).
- `worker` should remain, because it’ll usually work and rely on less code than the browser bundle.
- `browser` lets us fall back to browser code in packages that don’t have one of the other builds. Anecdotally, some of my code breaks without adding this condition. I’d argue this is more likely to happen than the inverse (a user imports a code that relies on browser-specific globals that Workers doesn’t have).

templates/cloudflare-workers/server.ts
```diff
@@ -1,7 +1,53 @@
-import { createEventHandler } from "@remix-run/cloudflare-workers";
+import { getAssetFromKV } from "@cloudflare/kv-asset-handler";
+import type { AppLoadContext } from "@remix-run/cloudflare";
+import { createRequestHandler, logDevReady } from "@remix-run/cloudflare";
 import * as build from "@remix-run/dev/server-build";
+import manifestJSON from "__STATIC_CONTENT_MANIFEST";
```
huw, line 5: I don’t think this line will work with `serverDependenciesToBundle: "all"`, unless something has changed recently. The bundler should be complaining about being unable to find this dependency. _If_ that’s the case, the solution is to use `serverDependenciesToBundle: [/^(?!(__STATIC_CONTENT_MANIFEST)$).*$/u]`, which is hacky but at least avoids the issue.

templates/cloudflare-workers/server.ts
```diff
@@ -1,7 +1,52 @@
...
+if (build.dev) {
+  logDevReady(build);
+}
+
+export default {
+  async fetch(
+    request: Request,
+    env: {
+      __STATIC_CONTENT: Fetcher;
+      __STATIC_CONTENT_MANIFEST: string;
```
huw, line 16: Using `__STATIC_CONTENT_MANIFEST` on the environment like this is a Miniflare / Wrangler Local bug. It will break when using Wrangler in live mode or in production. You have to import it [like so](https://github.com/cloudflare/kv-asset-handler#es-module-1), and add `serverDependenciesToBundle: [/^(?!(__STATIC_CONTENT_MANIFEST)$).*$/u]` to your `remix.config.js`.

acusti: i tried using the `server.ts` from this template with the latest versions of remix and wrangler and found that it broke my static assets in production. so i tried reverting this from using `env.__STATIC_CONTENT_MANIFEST` to using 
```ts
import assetManifestAsString from '__STATIC_CONTENT_MANIFEST';
const assetManifest = JSON.parse(assetManifestAsString);
```
and passing that as `ASSET_MANIFEST` to `getAssetFromKV()`, and it seemed to fix the css files that were 404ing but my `/build/root-{hash}.js` and `/build/manifest-{hash}.js` still 404ed in production. so i reverted `server.ts` back to:
```ts
import { createEventHandler } from "@remix-run/cloudflare-workers";

addEventListener(
  "fetch",
  createEventHandler({ build, mode: process.env.NODE_ENV })
);
```
and production is fixed (but no HMR in dev, obviously).

templates/cloudflare-pages/server.ts
```diff
@@ -0,0 +1,13 @@
+import { logDevReady } from "@remix-run/cloudflare";
+import { createPagesFunctionHandler } from "@remix-run/cloudflare-pages";
+import * as build from "@remix-run/dev/server-build";
+
+if (process.env.NODE_ENV === "development") {
+  logDevReady(build);
```
appden, line 6: This logs too soon for me resulting in a failed request from the browser after HMR updates because the server is not yet listening on the port when it's restarted by wrangler. By adding a `setTimeout` things work fine but definitely is hacky (I'm on 100ms to be safe).

huw: Per [my comment on #6415](https://github.com/remix-run/remix/issues/6415#issuecomment-1553969445), I think this is only a problem in Wrangler v2. The appropriate fix for this should probably be to upgrade the dependency to v3.

appden: You're right, it works great without `setTimeout` with Wrangler v3, so definitely recommend updating the dev dependency for the templates. Thanks!

pcattori:
Superceded by #6650

Issue #6210, opened 2023-04-27
False type safety with meta v2


AlemTuzlak:
### What version of Remix are you using?

v1.15

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

- Create a new remix project with the basic config
- Go into the routes/_index file
- Create a loader like the following:
```jsx
export const loader = () => {
  throw new Error();
  return json({
    title: "some title",
  });
};
```
- Change the meta function to the following:
```jsx
export const meta: V2_MetaFunction<typeof loader> = ({ data }) => {
  return [{ title: data.title }];
};
```
- Add an error boundary (eg):
```jsx
export const ErrorBoundary = () => {
  return (
    <div>
      <h1>Oops, something went wrong!</h1> 
    </div>
  );
};
```

### Expected Behavior

When you do ``` V2_MetaFunction<typeof loader> ``` in the meta function the ``` data ``` object  is typed as `T | undefined` where `T` corresponds to the ` typeof loader ` which warns the developer that the data object might not be defined and allows the developer to assure that the application doesn't break instead of it going into the error boundary.

### Actual Behavior

When you do ``` V2_MetaFunction<typeof loader> ``` in the meta function the ``` data ``` object is typed as `T` where `T` corresponds to the return type of ` typeof loader ` which gives you a false sense of security because if anything unexpected in the loader breaks the data object won't be available on the meta function so if you try calling `data.title` instead of `data?.title` it won't go into the error boundary but it will crash the application instead

machour:
This is the same as #5786 and it will be fixed by merging #5787
Closing as a duplicate to keep our issues tracker clean

AlemTuzlak:
@machour sorry for the duplicate! Didn't see that issue

machour:
no worries, I didn't even see that it was already fixed 😅

PR #6231, opened 2023-04-28
fix(v2_meta): hint data as being potentially undefined

machour:
I've been bitten by this in production a while ago, so here's my attempt at fixing it.

Not sure why we have duplicate types definitions in the react package, instead of a re-export from server-runtime.
Also, not sure this is the smartest fix, would definitely accept a better one

Closes: #6210 

- [ ] Docs
- [ ] Tests

Testing Strategy: 

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>
Screenshot 2023-04-28 at 10 24 22 AM

 <td>
Screenshot 2023-04-28 at 10 24 01 AM

</table>


aaronadamsCA:
Note this is a duplicate of #5787.

machour:
Good catch @aaronadamsCA, thank you!

machour:
Reopening for @brophdawg11 consideration.
I think that we both did half the job 😅

brophdawg11:
yeah I like this one better too I think

Issue #6196, opened 2023-04-26
Remix 1.16.0-pre.0  fails to run for CF-Pages. esbuild css-file plugin error


tobiasfoerg:
### What version of Remix are you using?

1.16.0-pre.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

checkout [reproduction repo](https://github.com/tobiasfoerg/cf-pages-remix-css-bug)
or create run `pnpm create-remix@1.16.0-pre.0` select cf pages template.
add a css file and load in root.
start the dev server: `pnpm dev`.

### Expected Behavior

should start the dev server

### Actual Behavior

```console
Building Remix app in production mode...
✘ [ERROR] Build failed with 1 error:
node_modules/.pnpm/esbuild-plugin-polyfill-node@0.2.0_esbuild@0.17.6/node_modules/esbuild-plugin-polyfill-node/polyfills/process.js:1:25: ERROR: Could not resolve "process" [plugin css-file]

    app/root.tsx:10:19:
      10 │ import styles from "./test.css";
         ╵                    ~~~~~~~~~~~~

  This error came from the "onLoad" callback registered here:

    node_modules/.pnpm/@remix-run+dev@1.16.0-pre.0/node_modules/@remix-run/dev/dist/compiler/plugins/cssImports.js:66:12:
      66 │       build.onLoad({
         ╵             ~~~~~~

    at setup (/Users/tobias/github/tobiasfoerg/cf-pages-remix-css-bug/node_modules/.pnpm/@remix-run+dev@1.16.0-pre.0/node_modules/@remix-run/dev/dist/compiler/plugins/cssImports.js:66:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async handlePlugins (/Users/tobias/github/tobiasfoerg/cf-pages-remix-css-bug/node_modules/.pnpm/esbuild@0.17.6/node_modules/esbuild/lib/main.js:1326:9)
```

markdalgleish:
@MichaelDeBoey This seems related to the migration to `esbuild-plugin-polyfill-node` in https://github.com/remix-run/remix/pull/5274. Any ideas?

MichaelDeBoey:
Closed by #6238, which will be available in the next release

PR #6238, opened 2023-04-28
fix: css bundling failing on non-node platforms

jacob-ebey:
Don't forward on injects for CSS compiler as it's never loading any JS code and esbuild seems to have a bug with CSS entries + inject

Closes: #6196

- [ ] Docs
- [ ] Tests

Testing Strategy:


packages/remix-dev/compiler/plugins/cssImports.ts
```diff
@@ -37,6 +37,7 @@ export function cssFilePlugin({
       build.onLoad({ filter: /\.css$/ }, async (args) => {
         let { metafile, outputFiles, warnings, errors } = await esbuild.build({
           ...buildOps,
```
mjackson, line 39: I'd personally prefer to nix this line and be explicit about everything that goes into `esbuild.build` instead of listing an empty array as an override.

pcattori:
Wait to merge this until we go back to `pre` mode for the 1.16.0 release

PR #6247, opened 2023-04-29
feat: expose `HeadersArgs` & `HeadersFunctionArgs` type

mcansh:
add `HeadersFunctionArgs` (and `HeadersArgs`) type to be consistent with `ActionArgs`, `LoaderArgs`, `DataFunctionArgs` and allows for using `function`

 ```tsx
 import type { HeadersFunctionArgs } from '@remix-run/*';
 export function headers({ loaderHeaders }: HeadersFunctionArgs) {
   return {
     "x-my-custom-thing": loaderHeaders.get("x-my-custom-thing") || "fallback"
   }
 }
 ```

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey (approved):

.changeset/headersfunctionargs.md
```diff
@@ -0,0 +1,22 @@
...
+"@remix-run/serve": patch
+"@remix-run/server-runtime": patch
+---
+
+add `HeadersFunctionArgs` type to be consistent with loaders/actions/meta and allows for using `function`
+
+```tsx
+import type { HeadersFunctionArgs } from '@remix-run/*';
+
+export function headers({ loaderHeaders }: HeadersFunctionArgs) {
```
MichaelDeBoey, line 17: Maybe use `HeadersArgs` here, as we're using `ActionArgs`/`LoaderArgs` in other places as well?

```suggestion
import type { HeadersArgs } from '@remix-run/node'; // or cloudflare/deno

export function headers({ loaderHeaders }: HeadersArgs) {
```

jacobparis: He's added HeadersArgs as an alias, while keeping HeadersFunctionArgs for consistency with DataFunctionArgs

MichaelDeBoey: Docs/examples/templates are using `ActionArgs`/`LoaderArgs` though, hence why I think we should use that in the changelog as well

machour (approved):

brophdawg11:
To avoid having 2 aliases of the same thing, I removed `HeadersFunctionArgs` and we only have `HeadersArgs` to be consistent with `LoaderArgs`/`ActionArgs`.  Personally I don't think `DataFunctionArgs` should have ever been exported since it's really an implementation detail.

The agnostic router uses `LoaderFunctionArgs` and `ActionFunctionArgs` so I don't think we need to leak that naming scheme out through Remix.

brophdawg11 (approved):

PR #6248, opened 2023-04-29
feat(remix-eslint-config): extend from typescript-eslint/recommended

JoshuaKGoldberg:
Wipes out most of the existing `typescript.js` lint rules, in favor of directly extending [`plugin:@typescript-eslint/recommended`](https://typescript-eslint.io/linting/configs#recommended). Remaining rule configurations are grouped with comments:

* Rules that Remix explicitly configures differently from the default
* Rules enabled by `rules/core.js` that aren't needed for TypeScript code
* Rules that enforce a style point not agreed upon by Remix <sup>✳️</sup>
* Ones that could eventually be investigated for enabling

Addresses: #6126

- [ ] Docs - ... where would I document this? Should I continue to hold off until this is reviewed?
- ~[ ] Tests~: `yarn lint` still passes, yes?

Testing Strategy: I'd encourage folks to try this out on any large Remix repository written with TypeScript. And let me know what those are so I can test them out too? 😄 

Note that this does not _(yet?)_ tackle two good next steps:

* Enable the ever more powerful rules from [`plugin:@typescript-eslint/recommended-requiring-type-checking`](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking)
    * I would highly recommend this, either as an addition to this PR or as a next step.
* <sup>✳️</sup>[typescript-eslint@v6, currently in beta](https://typescript-eslint.io/blog/announcing-typescript-eslint-v6-beta), separates out any remaining _stylistic_ rules from the _recommended_ configs. So moving to that new major version would make the stylistic disables section unnecessary.

Any code changes are commented inline.


packages/remix-architect/__tests__/server-test.ts
```diff
@@ -3,10 +3,6 @@ import path from "path";
 import lambdaTester from "lambda-tester";
 import type { APIGatewayProxyEventV2 } from "aws-lambda";
 import {
-  // This has been added as a global in node 15+, but we expose it here while we
-  // support Node 14
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  AbortController,
```
JoshuaKGoldberg, line 9: This test doesn't actually use `AbortController` that I can tell. Am I missing something?

packages/remix-dev/server-build.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-unreachable */
```
JoshuaKGoldberg, line 1: This rule is explicitly turned off in typescript-eslint configs because it's covered by TypeScript.

packages/remix-eslint-config/index.js
```diff
@@ -62,15 +62,14 @@ const config = {
...
+        "plugin:@typescript-eslint/recommended",
+      ],
       parser: "@typescript-eslint/parser",
       parserOptions: {
         sourceType: "module",
         ecmaVersion: 2019,
-        ecmaFeatures: {
-          jsx: true,
-        },
-        warnOnUnsupportedTypeScriptVersion: true,
```
JoshuaKGoldberg, line 73: `warnOnUnsupportedTypeScriptVersion` is enabled by default: https://typescript-eslint.io/architecture/parser/#warnonunsupportedtypescriptversion

packages/remix-eslint-config/index.js
```diff
@@ -62,15 +62,14 @@ const config = {
...
+        "plugin:import/typescript",
+        "plugin:@typescript-eslint/recommended",
+      ],
       parser: "@typescript-eslint/parser",
       parserOptions: {
         sourceType: "module",
         ecmaVersion: 2019,
-        ecmaFeatures: {
-          jsx: true,
-        },
```
JoshuaKGoldberg, line 72: `ecmaFeatures.jsx` is enabled by default for `.tsx` files.

packages/remix-eslint-config/rules/typescript.js
```diff
@@ -1,54 +1,49 @@
-const OFF = 0;
-const WARN = 1;
-const ERROR = 2;
```
JoshuaKGoldberg, line 3: I'm intrigued by these constants. Why bother using them? ESLint already allows `"off"`, etc. - and errors if you provide an invalid config value.

packages/remix-eslint-config/rules/typescript.js
```diff
@@ -1,54 +1,49 @@
...
-  "no-duplicate-imports": OFF,
-  // "@typescript-eslint/no-duplicate-imports": WARN,
-
-  "no-redeclare": OFF,
-  "@typescript-eslint/no-redeclare": ERROR,
-  "no-use-before-define": OFF,
+  // These rules are nice and we want to configure over the defaults
   "@typescript-eslint/no-use-before-define": [
-    WARN,
+    "error",
```
JoshuaKGoldberg, line 17: Regarding `"warn"` vs `"error"`: I generally recommend against using warnings except for transitionary code (enabling a new rule with many existing violations). https://github.com/eslint/eslint/issues/16696 -> https://github.com/eslint/eslint/pull/16882 ->  https://eslint.org/docs/latest/use/configure/rules#rule-severities. In typescript-eslint@v5, most preset config rule severities are set to error; in v6, _all_ will be error.

packages/remix-server-runtime/responses.ts
```diff
@@ -21,17 +21,14 @@ export type DeferFunction = <Data extends Record<string, unknown>>(
   init?: number | ResponseInit
 ) => TypedDeferredData<Data>;
 
-export type JsonFunction = <Data extends unknown>(
```
JoshuaKGoldberg, line 24: https://typescript-eslint.io/rules/no-unnecessary-type-constraint - `extends unknown` is redundant for these type parameters.

MichaelDeBoey: The docs are talking about `any` only, but this had `unknown`
Does the same apply?
If so, I guess it would be a good idea to update the docs?

packages/remix-eslint-config/rules/typescript.js
MichaelDeBoey: We have a workflow that runs `npx prettier -w . && npx eslint . --fix` & commits the changes, so I guess that's why the team decided to put these as `warn` instead of `error` 🤔

https://github.com/remix-run/remix/blob/main/.github/workflows/format.yml

packages/remix-server-runtime/responses.ts
JoshuaKGoldberg: Great point 😄 thanks! Filed https://github.com/typescript-eslint/typescript-eslint/issues/6973.

ngbrown:
@JoshuaKGoldberg What would it take to enable [`plugin:@typescript-eslint/recommended-requiring-type-checking`](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking)? I attempted to do that in #6325, but was closed in favor of this pull request. There doesn't seem to be a extensive conflict between the two changes, so could be cribbed for adding to this pull request.

JoshuaKGoldberg:
Yeah enabling [plugin:@typescript-eslint/recommended-requiring-type-checking](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking) would be a very good next step after this. If the powers-that-be are particularly interested, it could certainly be done either as a followup or as part of this PR. 

The main reason I didn't add type checked rules in _this_ PR was [type aware rule performance](https://typescript-eslint.io/linting/typed-linting#how-is-performance).
* `time yarn lint` on `main`: `11.86s user`
* `time yarn lint` on #6325: `58.61s user`

packages/remix-eslint-config/rules/typescript.js
brophdawg11: This has always confused me in the larger ecosystem as well.  I've long used the string values, and always assumed this was a pattern from early days of `eslint` when maybe string values weren't supported 🤷

packages/remix-eslint-config/rules/typescript.js
```diff
@@ -1,54 +1,49 @@
...
-const WARN = 1;
-const ERROR = 2;
-
 module.exports = {
-  "no-dupe-class-members": OFF,
-  "no-undef": OFF,
-
-  // Add TypeScript specific rules (and turn off ESLint equivalents)
-  "@typescript-eslint/consistent-type-assertions": WARN,
-  "@typescript-eslint/consistent-type-imports": WARN,
```
brophdawg11, line 11: Per https://typescript-eslint.io/rules/, these aren't enabled in `@typescript-eslint/recommended` so should we be keeping them?

packages/remix-eslint-config/rules/typescript.js
```diff
@@ -1,54 +1,49 @@
...
+  "@typescript-eslint/no-var-requires": "off",
+  "no-var": "off",
+  "prefer-rest-params": "off",
 
-  // There is a bug w/ @typescript-eslint/no-duplicate-imports triggered
-  // by multiple imports inside of module declarations. We should reenable
-  // this rule when the bug is fixed.
-  // https://github.com/typescript-eslint/typescript-eslint/issues/3071
-  "no-duplicate-imports": OFF,
-  // "@typescript-eslint/no-duplicate-imports": WARN,
```
brophdawg11, line 21: Is this something we could enable now that the referenced bug is closed?  It doesn't look to be enabled in the recommended rules

packages/remix-eslint-config/rules/typescript.js
```diff
@@ -1,54 +1,49 @@
...
-
-  "no-redeclare": OFF,
-  "@typescript-eslint/no-redeclare": ERROR,
```
brophdawg11, line 24: I don't see this on in the recommended rules either?

packages/remix-eslint-config/rules/typescript.js
```diff
@@ -1,54 +1,49 @@
...
   "@typescript-eslint/no-unused-vars": [
-    WARN,
+    "error",
     {
       args: "none",
       ignoreRestSiblings: true,
     },
   ],
-  "no-useless-constructor": OFF,
-  "@typescript-eslint/no-useless-constructor": WARN,
```
brophdawg11, line 53: This looks to only be include din the `strict` ruleset?

brophdawg11 (commented):
Thanks Josh!  Just a few questions on some existing rules and if we're losing them

packages/remix-eslint-config/rules/typescript.js
brophdawg11: It doesn't look like this or the below ones can be autofixed - so error may be correct.  Might fail a few CI builds on updating remix but if it's for a more accurate codebase then folks can always override the rule back down to `warn`.

packages/remix-eslint-config/rules/typescript.js
JoshuaKGoldberg: They can be kept if desired, sure. We removed it from the _recommended_ list because they're really a _stylistic_ concern. In the upcoming v6 of typescript-eslint they're added to the _stylistic_ config: https://v6--typescript-eslint.netlify.app/rules/consistent-type-assertions -> https://v6--typescript-eslint.netlify.app/linting/configs/#stylistic

What we could do is wait until typescript-eslint v6 is published hopefully in a few weeks _(which is something I've been saying for months)_, then extend both the _recommended_ and _stylistic_ configs. Thoughts?

packages/remix-eslint-config/rules/typescript.js
JoshuaKGoldberg: The rule is deprecated in favor of [import/no-duplicates](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-duplicates.md). I'd recommend just using that one instead.

packages/remix-eslint-config/rules/typescript.js
JoshuaKGoldberg: The rule is pretty much redundant when https://eslint.org/docs/latest/rules/no-var is enabled. I've seen exceedingly few projects use `var` over the last few years, and within them anything that would trigger the rule. Is there a reason you'd want to keep it enabled?

packages/remix-eslint-config/rules/typescript.js
JoshuaKGoldberg: Per https://typescript-eslint.io/rules/no-useless-constructor/#caveat it's got a bit too many false flags for us to feel comfortable enabling it. Do you want it despite the caveat?

JoshuaKGoldberg:
Sorry for taking a couple weeks to respond to your comments @brophdawg11! I lost track of this while traveling 😅. But I'm glad you're looking at it!

I responded to each thread inline - for each, let me know and I'm happy to make the change.

MichaelDeBoey:
@JoshuaKGoldberg It seems like you missed something that causes an ESLint error
https://github.com/remix-run/remix/actions/runs/5227221637/jobs/9438647396?pr=6248#step:7:9

Do you want to fix that one please?
If you don't have time, I'm happy to do so as well

JoshuaKGoldberg:
Aha, introduced merging from `main` - just pushed a fix now. Thanks!

packages/remix-eslint-config/rules/typescript.js
brophdawg11: ok, going to keep them in for now with a note that they can be removed once v6 is released and we adopt `@typescript-eslint/stylistic`.  I'm push up some changes shortly

packages/remix-eslint-config/rules/typescript.js
brophdawg11: Nah happy to take the guidance there :)

packages/remix-eslint-config/rules/typescript.js
brophdawg11: Nah - didn't look too deep into these specific rules before, just was looking at things that were net-removed.  Makes sense that it's not useful with `no-var`

brophdawg11:
This should go to `dev` since it changes a deployed package, so going to squash-merge to my branch and then will rebase onto dev there and get it in.  Thanks!

PR #6082, opened 2023-04-16
chore: update `no-response` workflow

MichaelDeBoey:
This will notify people before the PR/issue is actually closed as it now often happens by surprise when people are assuming everything is already fine

PR #6251, opened 2023-04-30
chore(ci/no-response): Don't automatically mark issues/PRs as stale

machour:
Discussed with @ryanflorence on Discord.

Let's leave that to the team discretion. 
Workflow will only close stale issues after 10 days

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -23,19 +23,6 @@ jobs:
           close-pr-message: >
             This PR has been automatically closed because we didn't hear
             anything from the original author after the previous notice.
-          stale-issue-label: needs-response
-          stale-issue-message: >
-            This issue has been automatically marked stale because we haven't
-            received a response from the original author in a while 🙈. This
-            automation helps keep the issue tracker clean from issues that are
-            not actionable. Please reach out if you have more information for us
-            or you think this issue shouldn't be closed! 🙂 If you don't do so
-            within 7 days, this issue will be automatically closed.
-          stale-pr-label: needs-response
-          stale-pr-message: >
-            This PR has been automatically marked stale because we haven't
-            received a response from the original author in a while 🙈. This
-            automation helps keep the issue tracker clean from issues that are
-            not actionable. Please reach out if you have more information for us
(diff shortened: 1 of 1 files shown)


mcansh:
we can probably just revert https://github.com/remix-run/remix/pull/6082 (and related prs for other repos)

machour:
@mcansh fine by me!

machour:
Closed in favor of #6254

MichaelDeBoey:
Reverting all these PRs will bring back the original problem that I was trying to fix by updating all these workflows though:

> This will notify people before the issue/PR is actually closed as it now often happens by surprise when people are assuming everything is already fine

machour:
@MichaelDeBoey as I see it, the workflow is the following:

1 - Someone posts an issue / PR
2 - If no one from the team attend to it, it stays open indefinitely
3 - If someone from the team asks a follow up question, we add a `needs-response` label manually
4 - If OP answers, the label gets removed, back to step 1
5 - If no feedback is received, the bot closes the issue/PR.

In step 3, OP do gets a notification when we comment requesting for more information, so no surprises here.

What am I missing ? 🤔

machour:
What was in place before the reversal

2.1 - Nothing happens, then the bot says "This issue has been automatically marked stale because we haven't received a response from the original author in a while 🙈."

This didn't make sense to me: OP opened an issue, nothing happened, so it will get closed after a while?

MichaelDeBoey:
@machour The concern why I originally update the workflow was because we did ask for changes (and added the `needs-response` label) in https://github.com/remix-run/examples/pull/52, the author updated some things, but did not all of the asked changes.
The author was planning on doing so, but forgot about it.
After a while, the PR was closed automatically.

If there would have been a notice/reminder, the author would have been able to finish all the requested updates & we could have merged the PR.

machour:
@MichaelDeBoey in the linked PR, I'm seeing that the workflow was the one adding the "needs-response" on its own.
This is clearly wrong. Only humans should add that label when requesting a change or asking a question.

Bot should only close issues labeled "needs-response" by a human, if there was indeed no response from the author after a while. The examples workflow was simply missing: `days-before-stale: -1`

PR #6259, opened 2023-05-01
feat(remix-node): only polyfill globals if they're not present

### LINKED ISSUE #6280: v1.16.0 SyntaxError: Identifier 'process' has already been declared

1.16.0

- [X] Yes

### Steps to Reproduce

Not sure the root cause, but this issue is appearing after upgrading Remix from 1.15 to 1.16, no other changes were made.

image

image

**Note**: Going through versions and it appears this issue has been occurring since at least `1.16.0-pre.4`. But from that version until earlier versions have this additional error:
`Cannot destructure property 'basename' of 'React__namespace.useContext(...)' as it is null.`

### Expected Behavior

There is no crash when upgrading to 1.16

### Actual Behavior

There is a crash when upgrading to 1.16

### LINKED ISSUE #6284: Remix v1.16.0 - Unable to build app with crypto dependency

v1.16.0

- [X] Yes

### Steps to Reproduce

1.  Create an app with  `import { createHash } from "crypto"`. 
2. Build the app 

### Expected Behavior

Should build successful - behavior in `1.15.0`

### Actual Behavior

Fails to build
```
✘ [ERROR] No matching export in "node_modules/@remix-run/dev/node_modules/esbuild-plugin-polyfill-node/polyfills/empty.js" for import "createHash"

    node_modules/@taskforcesh/bullmq-pro/node_modules/bullmq/dist/esm/classes/repeat.js:3:9:
      3 │ import { createHash } from 'crypto';
```

### LINKED ISSUE #6290: yjs throws error when using @remix-run/dev 1.16.0

1.16.0

- [X] Yes

### Steps to Reproduce

Version 1.16.0 of `@remix-run/dev` breaks `yjs`. It works fine for 1.15.0 of `@remix-run/dev`.

I didn't know how to add a test for this specific case, but here's a CodeSandbox with reproduction: https://codesandbox.io/p/github/anton-g/remix-yjs-error/main

Steps to recreate from scratch:
```sh
➜ npx create-remix@latest
? Where would you like to create your app? ./yjs-example
? What type of app do you want to create? Just the basics
? Where do you want to deploy? Choose Remix App Server if you're unsure; it's easy to change deployment
targets. Remix App Server
? TypeScript or JavaScript? TypeScript
? Do you want me to run `npm install`? Yes

➜ npm i yjs
```

Add the following to the top of `_index.js`:
```js
import * as Y from 'yjs'

const ydoc = new Y.Doc()
```

Run the app (`npm run dev`) and refresh the page.

### Expected Behavior

New yjs Doc created successfully on page load.

### Actual Behavior

Client crashes with error:
```
caught TypeError: Cannot read properties of undefined (reading 'isTTY')
    at environment.js:125:30
```

### LINKED ISSUE #6348: Remix v1.16.0 - Unnecessary Node polyfills added for Oxygen/Workers.

1.16.0

- [X] Yes

### Steps to Reproduce

I think this was introduced in https://github.com/remix-run/remix/pull/5274 , and is probably related to https://github.com/remix-run/remix/issues/6280 and its linked issues/PRs. I'm still writing this issue to show that it also affects Worker environments

You can reproduce the issue when building a project with a typical config for Oxygen or CFW:

```js
module.exports = {
  appDirectory: 'app',
  ignoredRouteFiles: ['**/.*'],
  server: './server.ts',
  serverMainFields: ['browser', 'module', 'main'],
  serverConditions: ['worker', process.env.NODE_ENV],
  serverDependenciesToBundle: 'all',
  serverModuleFormat: 'esm',
  serverPlatform: 'neutral',
  serverMinify: false, // To check the bundle later
};
```

### Expected Behavior

The resulting server/worker bundle shouldn't contain Node polyfills. Or, at least, they shouldn't make the bundle crash when running on a Worker environment.

### Actual Behavior

Many Node polyfills are added to the JS and Server bundles, thus increasing the size by ~60K non-minified. One of the polyfills is also accessing a global `navigator` variable, making the process exit with `navigator is not defined` in Worker environments:

image

image

---

I'm not sure if the solution is modifying [this conditional](https://github.com/remix-run/remix/blob/bac8b5bf9c9c483a6aaed2db813782a546e40148/packages/remix-dev/compiler/server/compiler.ts#L63-L65) and skip when `serverPlatform === 'neutral'`, or perhaps passing different options to [esbuild-plugin-polyfill-node](https://www.npmjs.com/package/esbuild-plugin-polyfill-node) since it seems to have a different default behavior compared to the old [@esbuild-plugins/node-modules-polyfill](https://www.npmjs.com/package/@esbuild-plugins/node-modules-polyfill).

### LINKED ISSUE #6380: Esbuild inject adds all node globals for non-node environment

1.16

- [X] Yes

### Steps to Reproduce

Install esbuild, add esbuild-plugin-polyfill-node plugin. Import commonjs module from entry point. You'll see all polyfills to be included in resulting bundle.

This might be caused by this behavior. Basically esbuild cannot safely require es module from commonjs. So the entire idea to rely on esbuild to inject globals is not good since all remix code and dependencies is commonjs.

https://github.com/evanw/esbuild/issues/1697#issuecomment-947645782

### Expected Behavior

Unnecessary polyfills should not be included.
Remix should have a way to disable all polyfills (disabling only globals is not supported in plugin).

### Actual Behavior

Our worker bundle increased with 45kB of unused code.

MichaelDeBoey:
Closes #5858
Closes #6280
Closes #6284
Closes #6290
Closes #6348
Closes #6380

### DIFF
--- a/packages/remix-node/globals.ts
+++ b/packages/remix-node/globals.ts
@@ -43,20 +43,20 @@ declare global {
 }
 
 export function installGlobals() {
-  global.atob = atob;
-  global.btoa = btoa;
+  global.atob ||= atob;
+  global.btoa ||= btoa;
 
-  global.Blob = NodeBlob;
-  global.File = NodeFile;
+  global.Blob ||= NodeBlob;
+  global.File ||= NodeFile;
 
-  global.Headers = NodeHeaders as typeof Headers;
-  global.Request = NodeRequest as typeof Request;
-  global.Response = NodeResponse as unknown as typeof Response;
-  global.fetch = nodeFetch as typeof fetch;
-  global.FormData = NodeFormData;
+  global.Headers ||= NodeHeaders as typeof Headers;
+  global.Request ||= NodeRequest as typeof Request;
+  global.Response ||= NodeResponse as unknown as typeof Response;
+  global.fetch ||= nodeFetch as typeof fetch;
+  global.FormData ||= NodeFormData;
 
-  global.ReadableStream = NodeReadableStream;
-  global.WritableStream = NodeWritableStream;
+  global.ReadableStream ||= NodeReadableStream;
+  global.WritableStream ||= NodeWritableStream;
 
   global.AbortController = global.AbortController || NodeAbortController;
 }


isker:
I'm not sure it will be as easy as this.  Before I gave up on using remix for my project, I had to patch other parts of it which were relying on polyfill implementation details in order to get rid of them: https://github.com/isker/neogrok/commit/c47fc1c6557875b6f3d87494a78f23d5c8b7848e#diff-49b936c82e6771ef5a3b352fb8cc1fef5a732a395be7a2bb56ac508bf90e7f79

machour:
Also beware, `||=` is node >= 15. We want to maintain node 14 compatibility.

MichaelDeBoey:
> Also beware, ||= is node >= 15. We want to maintain node 14 compatibility.

Code will be converted to Node 14, no? 🤔

machour:
> Code will be transpiled to Node 14, no? 🤔

You're correct, I totally missed https://github.com/remix-run/remix/pull/5047, thanks!

MichaelDeBoey:
Superseded by #6547

PR #6262, opened 2023-05-01
Revert "chore: remove `styfle/cancel-workflow-action` usage"

mcansh:
Reverts remix-run/remix#6178

it's currently cancelling the comment, deploy tests, and stack tests for nightly and changeset releases https://github.com/remix-run/remix/actions/runs/4843033118, https://github.com/remix-run/remix/actions/runs/4833273314

### DIFF
--- a/.github/workflows/deployments.yml
+++ b/.github/workflows/deployments.yml
@@ -28,16 +28,15 @@ on:
       TEST_VERCEL_USER_ID:
         required: true
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
-
 jobs:
   arc_deploy:
     name: Architect Deploy
     if: github.repository == 'remix-run/remix'
     runs-on: ubuntu-latest
     steps:
+      - name: 🛑 Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.11.0
+
       - name: ⬇️ Checkout repo
         uses: actions/checkout@v3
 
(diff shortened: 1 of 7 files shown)


mcansh:
issue found - `Concurrency ensures that only a single job using the same concurrency group will run at a time.` should be able to just add the concurrency settings to each job

MichaelDeBoey:
@mcansh The cancelling means that multiple runs are happening for the same workflow on the same ref

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

---

> If you have multiple workflows in the same repository, concurrency group names must be unique across workflows to avoid canceling in-progress jobs or runs from other workflows. Otherwise, any previously in-progress or pending job will be canceled, regardless of the workflow.  
> To only cancel in-progress runs of the same workflow, you can use the `github.workflow` property to build the concurrency group:
> ```yml
> concurrency:
>   group: ${{ github.workflow }}-${{ github.ref }}
>   cancel-in-progress: true
> ```

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

---

> |Property name|Type|Description|
> |----|----|---|
> |`github.ref`|`string`|The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by `push`, this is the branch or tag ref that was pushed. For workflows triggered by `pull_request`, this is the pull request merge branch. For workflows triggered by `release`, this is the release tag created. For other triggers, this is the branch or tag ref that triggered the workflow run. This is only set if a branch or tag is available for the event type. The ref given is fully-formed, meaning that for branches the format is `refs/heads/<branch_name>`, for pull requests it is `refs/pull/<pr_number>/merge`, and for tags it is `refs/tags/<tag_name>`. For example, `refs/heads/feature-branch-1`.|
> |`github.workflow`|`string`|The name of the workflow. If the workflow file doesn't specify a `name`, the value of this property is the full path of the workflow file in the repository.|

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

PR #6279, opened 2023-05-02
fix(dev): serve any non-Remix assets from `public/`

### LINKED ISSUE #6271: Remix 1.16 `unstable_dev` - problems with assets

1.16

- [X] Yes

### Steps to Reproduce

1. `git clone git@github.com:ArekBartnik/remix-116-dev-assets.git`
2. `npm i`
3. `npm run dev`
4. go to `http://localhost:3000`

### Expected Behavior

It should show a page using Inter font and an icon after "Welcome to Remix".

| With `unstable_dev: true;` | Without `unstable_dev: true`|
|-|-|
|image| image|

Without new devserver paths to assets are relative but after turning this flag they change to absolute.

### Actual Behavior

There're 2 errors:
- for the fonts -> 404 `http://localhost:3001/fonts/InterV.var.woff2 net::ERR_ABORTED 404 (Not Found)`
- for the svg icon -> `Unsafe attempt to load URL http://localhost:3001/build/_assets/document-QP6N4QKU.svg from frame with URL http://localhost:3000/. Domains, protocols and ports must match.

image

pcattori:
Fixes #6271

### DIFF
--- a/packages/remix-dev/devServer_unstable/index.ts
+++ b/packages/remix-dev/devServer_unstable/index.ts
@@ -196,6 +196,7 @@ export let serve = async (
         maxAge: "1y",
       })
     )
+    .use(express.static("public", { maxAge: "1h" }))
 
     // handle `broadcastDevReady` messages
     .use(express.json())
--- a/packages/remix-dev/devServer_unstable/index.ts
+++ b/packages/remix-dev/devServer_unstable/index.ts
@@ -54,6 +54,7 @@ export let serve = async (
     httpScheme: string;
     httpHost: string;
     httpPort: number;
+    publicDirectory: string;
     websocketPort: number;
     restart: boolean;
   }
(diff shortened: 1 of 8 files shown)


packages/remix-dev/devServer_unstable/index.ts
```diff
@@ -196,6 +196,7 @@ export let serve = async (
         maxAge: "1y",
       })
     )
+    .use("*", express.static("public", { maxAge: "1h" }))
```
kentcdodds, line 199: I believe the public directory is configurable right? Maybe this should be dynamic based on the config?

pcattori: the Remix assets build directory (e.g. `public/build/`) is configurable, but I don't think we have config for `public/` itself.

pcattori: Ok we can make this configurable for dev. Shouldn't need to be set in prod since you already have control of your app server, but in dev we need to make sure we play nicely with any setup.

pcattori: `--public-directory` / `unstable_dev.publicDirectory` option added!

jacob-ebey (approved):

Issue #6286, opened 2023-05-02
Remix 1.16 `unstable_dev` race condition 


thinkafterbefore:
### What version of Remix are you using?

1.16

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. `npx create-remix@latest --template https://github.com/remix-run/remix/tree/templates_v2_dev/templates/cloudflare-pages`

2. Replace `/app/routes/_index.tsx` with 
```ts
import { json } from "@remix-run/cloudflare";

const DELAY = 1000;

export const loader = async () => {
  await new Promise((resolve) => setTimeout(resolve, DELAY));
  return json(null);
};

export default function Test() {
  return <>TESTING</>;
}
```
Play around with `DELAY`, and occasionally get:
```
Application Error
TypeError: Failed to fetch
    at fetchData (http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:6924:24)
    at http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:7172:26
    at runHandler (http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:2480:26)
    at callLoaderOrAction (http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:2519:22)
    at http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:1712:70
    at Array.map (<anonymous>)
    at callLoadersAndMaybeResolveData (http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:1712:55)
    at handleLoaders (http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:1327:15)
    at startNavigation (http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:1169:15)
    at Object.revalidate (http://localhost:3001/build/_shared/chunk-QJS5FCIL.js:1096:7)
```

### Expected Behavior

Should update the text in the route without throwing a fetch error

### Actual Behavior

Application Error - Failed to fetch

----

My actual codebase throws this every time I make a change and the dev server is running. I'm suspecting some race condition with `wrangler` because when I set a breakpoint right before it throws, it always works.

jacob-ebey:
IDK if we can get around this in CF to be honest. There's no "HTTP port is available for connection" hook. Closest we can get is "code is loaded and we are restarting the http server". To get around this I introduced a retry mechanism for fetches caused by HDR in #6287.

PR #6287, opened 2023-05-02
feat: retry revalidation caused by HDR

jacob-ebey:
Closes: #6286

- [ ] Docs
- [ ] Tests

Testing Strategy:


brophdawg11 (approved):
Chatted a bout reducing the # of retires to maybe 3 or 4 and spreading the times out a bit.  If it doesn't work after 3-4 then we're likely screwed anyway

Issue #5247, opened 2023-01-25
V2 routing doesn't detect new routes


xHomu:
### What version of Remix are you using?

v1.11.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. `npx create-remix` and go through all the steps.
2. add `future: {v2_routeConvention: true,}` to`remix.config.js`.
3. `npm run dev` 
4. Create `app/routes/test.tsx` and a basic export default component.
5. Visit `localhost:3000/test`

### Expected Behavior

See the test route.

### Actual Behavior

404 Not Found thrown.

* If we restart the dev server manually, route renders correctly.
* Untoggle v2_routeConvention, route renders correctly.
* Edits to existing route files work as expected.

sample repo:
https://github.com/xHomu/remix-v2-routes-new2/blob/main/app/routes/test.tsx

mcansh:
unfortunately i'm unable to reproduce

both the CLI and browser pick up the new file

```log
Done in 24ms.
💿 File created: app/routes/my-new-route.tsx
💿 Rebuilding...
💿 Rebuilt in 97ms
GET / 200 - - 3.490 ms

Rebuilding...

Done in 33ms.
💿 File changed: app/routes/my-new-route.tsx
💿 Rebuilding...
💿 Rebuilt in 83ms
GET / 200 - - 5.460 ms
GET /my-new-route 200 - - 17.775 ms
```

xHomu:
Mmm, still getting this error. I'll switch up my setup a bit to see if I can narrow down if it's a local issue:
![image](https://user-images.githubusercontent.com/84349818/214685596-dac9ac1a-7338-4c61-a3c9-304bcb37d950.png)

xHomu:
Tested on different environments, seems to be a Windows-specific bug:

✔️ StackBlitz
✔️ Mac
❌ Windows laptop
❌ Windows PC

mcansh:
hm, i'll pull windows up shortly and see what i can come up with

mcansh:
yep, fast glob will return posix style paths `/` instead of windows `\\` - will go out in the next release

trevor-hackett:
I'm still seeing this issue on Windows. Has the fix been released?

stepniaczky:
> I'm still seeing this issue on Windows. Has the fix been released?

I have the same

stepniaczky:
> I'm still seeing this issue on Windows. Has the fix been released?

[...]

PR #6293, opened 2023-05-03
fix(remix-dev): normalize path before testing if it is a route entry

mcansh:
before this, `entryPoints` were normalized to use unix path separators (`/`) where `file` and `appFile` were using windows path separators (`\\`)

commit was done in parallels on windows 11 where i don't have gpg set up, sue me

Closes: #5247 

- [ ] Docs
- [ ] Tests

Testing Strategy: local testing on windows

### DIFF
--- a/.changeset/light-bikes-exercise.md
+++ b/.changeset/light-bikes-exercise.md
@@ -1,0 +1,6 @@
+---
+"remix": patch
+"@remix-run/dev": patch
+---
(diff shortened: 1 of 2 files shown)


pcattori (approved):

xHomu:
Can confirm that new routes are now working in windows with v2 routing flag on!

PR #6294, opened 2023-05-03
fix(dev): dev server listens for updated build hash before writing server files

pcattori:
previously there was a race where the app server could respond with the updated manifest version (aka build hash) _before_ the dev server was listening for that version.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.changeset/poor-nails-warn.md
+++ b/.changeset/poor-nails-warn.md
@@ -1,0 +1,8 @@
+---
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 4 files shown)


brandonpittman:
@pcattori Is this in 1.16.0?

pcattori:
> @pcattori Is this in 1.16.0?

@brandonpittman , no this is not in 1.16. will be in next release.

PR #6306, opened 2023-05-04
refactor(dev): revert serving assets from dev server

pcattori:
When sending HMR updates, the assumption was that serving assets from dev server would allow those updates to be sent immediately, without waiting for the app server to restart. But this isn't the case (e.g. if you navigate to another page that has a loader and the app server is rebooting, that loader fetch will fail).

---

Idea moving forward is:
1) If you don't mind a bit of downtime from your app server during dev, then you can rely on Remix restarts. but you'll need to wait for app server to reboot before you hard refresh or click buttons that trigger actions, or navigate to different pages that need loaders.

2) if that amount of downtime is too much for you, using chokidar and purgeRequireCache w/ --no-restart and that will keep you zipping along with an always-up server.

Note: during rebuild there is no downtime, its only after rebuild but before app server restart that there is some downtime with (1) 
^not serving assets from dev server also means there's less of a difference b/w dev and prod, which is nice. plus we don't have to use dataurls to handle svgs if we do that.

### DIFF
--- a/.changeset/ninety-flies-allow.md
+++ b/.changeset/ninety-flies-allow.md
@@ -1,7 +1,0 @@
----
-"@remix-run/dev": patch
----
-
-- Fix public asset serving for non-Remix assets in the new dev server
-- Add `--public-directory` / `unstable_dev.publicDirectory` option for configuring local directory for serving non-Remix assets like fonts
-- Fix `svg` serving for new dev server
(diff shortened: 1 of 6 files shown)


jacob-ebey (approved):

acusti:
in case this debugging info is useful, i’m using the cloudflare-workers adaptor and found the following when running `yarn dev` (where `"dev": "remix dev --no-restart -c \"wrangler dev --local ./build/index.js\""`):

* using v0.16.0, svg assets 404 (the URL include is `http://localhost:3001/…`)
* using v0.0.0-nightly-6e23fcf-20230505 (with this PR), the build gives me errors for every imported svg file: `✘ [ERROR] No loader is configured for ".svg" files: public/name-of-file.svg`
* using v0.0.0-nightly-0e2763f-20230503 (which has the change in #6279), svg assets work perfectly

pcattori:
@acusti seems like you are also seeing https://github.com/remix-run/remix/issues/6316 . I'm unable to reproduce this locally. If you have a repro, could you post it in #6316

PR #6307, opened 2023-05-04
docs(tutorials/blog): Fix markdown label spacing

Hirochon:
Hi! 

This PR adds a whitespace character to the label of the Markdown input field in the create post form, in line with a similar change made in a later step of the tutorial.

The original code did not have a whitespace character after the colon, which caused a visual inconsistency with the other labels in a later step that was corrected by adding the whitespace character. This PR adds the whitespace character to the earlier step for consistency.

- [x] Docs

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -184,6 +184,7 @@
 - HenryVogt
 - hicksy
 - himorishige
+- Hirochon
 - hkan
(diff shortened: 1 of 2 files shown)


machour (approved):
That is very consistent, thank you!

PR #6327, opened 2023-05-06
docs(guide/styles): Fix anchor for tailwind css

TomVance:
Corrects the fragment for the tailwind css docs, from #tailwind to #tailwind-css

PR #6328, opened 2023-05-06
Add tomvance to contributors

TomVance:
Adding myself to the contributors list so that I can open a docs change here: https://github.com/remix-run/remix/pull/6327


machour:
Hey Tom, let's do it in https://github.com/remix-run/remix/pull/6327 instead

PR #6330, opened 2023-05-06
refactor(remix-server-runtime): expose `SessionStorage`'s methods as arrow functions

ngbrown:
## Summary of changes

This is just the relevant parts of the code changes from #6325 instead of also tackling eslint-config changes.

- Expose methods in the `SessionStorage` interface as arrow functions so destructuring is correctly part of the contract.

## Details

I attempted to add the typescript-eslint rules contained in ["plugin:@typescript-eslint/recommended-requiring-type-checking"](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking) to a Remix based project and came across two common code issues:

```js
const { pipe, abort } = renderToPipeableStream();
```

and then on:

```js
const { getSession, commitSession, destroySession } = createDatabaseSessionStorage();
```

Both of these cause an `unbound-method` error:

> error  Avoid referencing unbound methods which may cause unintentional scoping of `this`.
> 
> If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead @typescript-eslint/unbound-method

This error needed to be taken care of by the type definitions provided by the libraries.

`@types/react-dom` [v18.2.4](https://www.npmjs.com/package/@types/react-dom/v/18.2.4) now includes `this: void` in the functions returned from `renderToPipeableStream()`, so the first half is solved.

This pull request is to address the changes needed in Remix.

This could subtly change the contract for external libraries returning `SessionStorage`, so it would be good to get this in before Remix v2.

Testing Strategy:

> I opened up my windows machine and ran this script:
>
> ```
> yarn test:primary
> npx playwrite install
> yarn test
> yarn build --tsc
> ```

Also tested on a Remix project with eslint and ["plugin:@typescript-eslint/recommended-requiring-type-checking"](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking) enabled.

### DIFF
--- a/.changeset/polite-garlics-invite.md
+++ b/.changeset/polite-garlics-invite.md
@@ -2,4 +2,4 @@
 "@remix-run/server-runtime": patch
 ---
 
-Added `this: void` to functions in the `SessionStorage` and `SessionIdStorageStrategy` interfaces so destructuring is correctly part of the contract.
+Expose methods in the `SessionStorage` interface as arrow functions so destructuring is correctly part of the contract.
--- a/.changeset/polite-garlics-invite.md
+++ b/.changeset/polite-garlics-invite.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/server-runtime": minor
+---
+
+Added `this: void` to functions in the `SessionStorage` and `SessionIdStorageStrategy` interfaces so destructuring is correctly part of the contract.
--- a/.changeset/polite-garlics-invite.md
+++ b/.changeset/polite-garlics-invite.md
@@ -1,5 +1,5 @@
 ---
-"@remix-run/server-runtime": minor
+"@remix-run/server-runtime": patch
 ---
 
 Added `this: void` to functions in the `SessionStorage` and `SessionIdStorageStrategy` interfaces so destructuring is correctly part of the contract.
--- a/packages/remix-server-runtime/sessions.ts
+++ b/packages/remix-server-runtime/sessions.ts
@@ -174,31 +174,28 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
    * Session. If there is no session associated with the cookie, this will
    * return a new Session with no data.
    */
-  getSession(
-    this: void,
+  getSession: (
     cookieHeader?: string | null,
     options?: CookieParseOptions
-  ): Promise<Session<Data, FlashData>>;
+  ) => Promise<Session<Data, FlashData>>;
 
   /**
    * Stores all data in the Session and returns the Set-Cookie header to be
    * used in the HTTP response.
    */
-  commitSession(
-    this: void,
+  commitSession: (
     session: Session<Data, FlashData>,
     options?: CookieSerializeOptions
-  ): Promise<string>;
+  ) => Promise<string>;
 
   /**
    * Deletes all data associated with the Session and returns the Set-Cookie
    * header to be used in the HTTP response.
    */
-  destroySession(
-    this: void,
+  destroySession: (
     session: Session<Data, FlashData>,
     options?: CookieSerializeOptions
-  ): Promise<string>;
+  ) => Promise<string>;
 }
 
 /**
@@ -224,7 +221,6 @@ export interface SessionIdStorageStrategy<
    * Creates a new record with the given data and returns the session id.
    */
   createData: (
-    this: void,
     data: FlashSessionData<Data, FlashData>,
     expires?: Date
   ) => Promise<string>;
@@ -232,16 +228,12 @@ export interface SessionIdStorageStrategy<
   /**
    * Returns data for a given session id, or `null` if there isn't any.
    */
-  readData: (
-    this: void,
-    id: string
-  ) => Promise<FlashSessionData<Data, FlashData> | null>;
+  readData: (id: string) => Promise<FlashSessionData<Data, FlashData> | null>;
 
   /**
    * Updates data for the given session id.
    */
   updateData: (
-    this: void,
     id: string,
     data: FlashSessionData<Data, FlashData>,
     expires?: Date
@@ -250,7 +242,7 @@ export interface SessionIdStorageStrategy<
   /**
    * Deletes data for a given session id from the data store.
    */
-  deleteData: (this: void, id: string) => Promise<void>;
+  deleteData: (id: string) => Promise<void>;
 }
 
 export type CreateSessionStorageFunction = <
--- a/packages/remix-server-runtime/sessions.ts
+++ b/packages/remix-server-runtime/sessions.ts
@@ -175,6 +175,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
    * return a new Session with no data.
    */
   getSession(
+    this: void,
     cookieHeader?: string | null,
     options?: CookieParseOptions
   ): Promise<Session<Data, FlashData>>;
@@ -184,6 +185,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
    * used in the HTTP response.
    */
   commitSession(
+    this: void,
     session: Session<Data, FlashData>,
     options?: CookieSerializeOptions
   ): Promise<string>;
@@ -193,6 +195,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
    * header to be used in the HTTP response.
    */
   destroySession(
+    this: void,
     session: Session<Data, FlashData>,
     options?: CookieSerializeOptions
   ): Promise<string>;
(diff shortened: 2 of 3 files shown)


packages/remix-server-runtime/sessions.ts
```diff
@@ -175,6 +175,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
    * return a new Session with no data.
    */
   getSession(
+    this: void,
```
JoshuaKGoldberg, line 178: FWIW I would generally consider using an arrow function to be more idiomatic. `this: void` is much less common in TypeScript code and much less understood by a lot of TS devs. Arrow functions are pretty commonly understood.

Maybe there's a docs issue to be filed on us in typescript-eslint? Or perhaps even a lint rule around preferring one vs. the other? What do you think? 🤔

packages/remix-server-runtime/sessions.ts
```diff
@@ -221,19 +224,24 @@ export interface SessionIdStorageStrategy<
    * Creates a new record with the given data and returns the session id.
    */
   createData: (
+    this: void,
```
JoshuaKGoldberg, line 227: Is there a real need to `this: void` for an arrow function? I'm surprised it's even allowed - maybe another lint rule request for typescript-eslint...

packages/remix-server-runtime/sessions.ts
MichaelDeBoey, line 178:
```suggestion
  getSession: (
```

packages/remix-server-runtime/sessions.ts
```diff
@@ -175,6 +175,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
    * return a new Session with no data.
    */
   getSession(
+    this: void,
     cookieHeader?: string | null,
     options?: CookieParseOptions
   ): Promise<Session<Data, FlashData>>;
```
MichaelDeBoey, line 181:
```suggestion
  ) => Promise<Session<Data, FlashData>>;
```

packages/remix-server-runtime/sessions.ts
```diff
@@ -184,6 +185,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
    * used in the HTTP response.
    */
   commitSession(
+    this: void,
```
MichaelDeBoey, line 188:
```suggestion
  commitSession: (
```

packages/remix-server-runtime/sessions.ts
```diff
@@ -184,6 +185,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
...
     session: Session<Data, FlashData>,
     options?: CookieSerializeOptions
   ): Promise<string>;
```
MichaelDeBoey, line 191:
```suggestion
  ) => Promise<string>;
```

packages/remix-server-runtime/sessions.ts
```diff
@@ -193,6 +195,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
    * header to be used in the HTTP response.
    */
   destroySession(
+    this: void,
```
MichaelDeBoey, line 198:
```suggestion
  destroySession: (
```

packages/remix-server-runtime/sessions.ts
```diff
@@ -193,6 +195,7 @@ export interface SessionStorage<Data = SessionData, FlashData = Data> {
...
     session: Session<Data, FlashData>,
     options?: CookieSerializeOptions
   ): Promise<string>;
```
MichaelDeBoey, line 201:
```suggestion
  ) => Promise<string>;
```

packages/remix-server-runtime/sessions.ts
MichaelDeBoey, line 227:
```suggestion
```

packages/remix-server-runtime/sessions.ts
```diff
@@ -221,19 +224,24 @@ export interface SessionIdStorageStrategy<
...
   ) => Promise<string>;
 
   /**
    * Returns data for a given session id, or `null` if there isn't any.
    */
-  readData: (id: string) => Promise<FlashSessionData<Data, FlashData> | null>;
+  readData: (
+    this: void,
+    id: string
+  ) => Promise<FlashSessionData<Data, FlashData> | null>;
```
MichaelDeBoey, line 238:
```suggestion
  readData: (id: string) => Promise<FlashSessionData<Data, FlashData> | null>;
```

packages/remix-server-runtime/sessions.ts
```diff
@@ -221,19 +224,24 @@ export interface SessionIdStorageStrategy<
...
 
   /**
    * Updates data for the given session id.
    */
   updateData: (
+    this: void,
```
MichaelDeBoey, line 244:
```suggestion
```

packages/remix-server-runtime/sessions.ts
```diff
@@ -242,7 +250,7 @@ export interface SessionIdStorageStrategy<
   /**
    * Deletes data for a given session id from the data store.
    */
-  deleteData: (id: string) => Promise<void>;
+  deleteData: (this: void, id: string) => Promise<void>;
```
MichaelDeBoey, line 253:
```suggestion
  deleteData: (id: string) => Promise<void>;
```

.changeset/polite-garlics-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/server-runtime": minor
```
MichaelDeBoey, line 2:
```suggestion
"@remix-run/server-runtime": patch
```

MichaelDeBoey (changes requested):
Let's use arrow functions instead, as @JoshuaKGoldberg suggested in https://github.com/remix-run/remix/pull/6330#discussion_r1186723774

packages/remix-server-runtime/sessions.ts
ngbrown: There is a difference between arrow functions with and without the `this: void`. This will affect the conversion of all the functions to arrow functions too.

With the `this: void`, TypeScript enforces the contract on the implementing function that there is no viable `this`. For example, try adding a `this` reference to an implementer, such as in `workerKVStorage.ts`:

```ts
...
    },
    async readData(id) {
      this.deleteData(id); // error when arrow functions in `SessionIdStorageStrategy` has `this: void` no error with just arrow function.
      let session = await kv.get(id);

      if (!session) {
        return null;
      }

      return JSON.parse(session);
    },
...
```

So while changing the `SessionStorage` methods to arrow functions will fix the linting error, it doesn't enforce the contract specified by the documentation.

Thoughts?

packages/remix-server-runtime/sessions.ts
ngbrown: See my other [comment](https://github.com/remix-run/remix/pull/6330#discussion_r1186739382), because just converting to arrow functions won't enforce the documentation contract of those functions being destructured. I'm not sure which way is the right way for Remix, but there is a hidden side-effect of just using arrow functions.

packages/remix-server-runtime/sessions.ts
MichaelDeBoey: If people implement it with an arrow function, the error will be gone on their side as well, no?

ngbrown: It might be that my example of adding the `this.deleteData(id)` into the object creation of `createWorkersKVSessionStorage()` is too contrived.

This would mostly matter if the implementer is creating a class that implements the `SessionIdStorageStrategy` interface and passes that as the strategy to `createSessionStorage`. It appears that a class does not pick up on the difference one way or the other, arrow function or not.

I reverted the changes to `SessionIdStorageStrategy` and converted the `SessionStorage` functions to arrow functions as suggested.

packages/remix-server-runtime/sessions.ts
ngbrown: Having `this: void` in an interface contract will make the IDE's include it in an implementing class, so the implementer can handle the situation correctly, like ensuring it is bound or implemented as an arrow function in the class.

Like you said, `this: void` is not commonly used, but that's not necessarily a good thing. `this` in JavaScript is a fraught subject and there is certainly the need to make it clear whether a method needs to have a specific `this` instance connected to it.

As far as my initial choice, I went with the first suggestion because it seemed that it would result in clearer code when reading to have an annotation rather than change the methods to arrow functions, especially if it could have been a class type implementation.

MichaelDeBoey (approved):

brophdawg11 (approved):

PR #6374, opened 2023-05-11
fix(dev): include `cssBundleHref` in manifest hash

### LINKED ISSUE #6369: CSS bundle gives hydration error on remix 1.16 with css modules

1.16

- [X] Yes

### Steps to Reproduce

- Create a new app using npx create-remix@latest that uses `cssBundleHref` by default
- Create a component that uses a css module. e.g.

```tsx
import css from './button.module.css'

export function Button() {
	return <button className={css.button}>text</button>
}
```

- Serve the app
- Make a change in the css module file
- Observe app being rebuild and page reloaded
- Changes are visible however the app is logging the following error to the browser console:

image

Here's a repo that produces this behaviour: https://github.com/TimoWestland/remix-css-bundle-issue

Also tried with scss, issue is the same for both css and scss.

Tried with an stand alone express server as well as that's what I'm using in my production app, and suspected that might be related, but the error is the same with and without an express server.

### Expected Behavior

App is able to hydrate the css bundle href with the correct fingerprint to the client

### Actual Behavior

Client still uses old fingerprint for the css bundle href.

I'm not to worried about just the error message, although it does make me suspect something else might be broken.

On a production app I'm working on for my client, the issue seems to be bigger. 
When the app has just started, the first couple of reloads are fine, also the error is not logged.

Then once this error is logged, the app does rebuild but no longer reloads after changes, so changes are not visible.

In some cases the css bundle href gives a 404. When I do a hard refresh, all files in /public/build suddenly give 404s. 

When I log `cssBundleHref` from `@remix-run/css-bundle` in this case, I sometimes also see that the fingerprint does not match with the fingerprint on the css file that is in the `/public/build` folder, both on the server and the client.

Haven't been able to pinpoint what exactly is causing this issue in that app, as I cannot reproduce this in the example app that is linked above. But I do expect it might be related somehow.

markdalgleish:
Fixes #6369

The `cssBundleHref` value wasn't being included in the hash in the manifest file name. This means that CSS bundle changes weren't busting the manifest cache. I've also refactored the code to try to avoid this issue coming back when further changes are made to the manifest structure.

### DIFF
--- a/packages/remix-dev/compiler/manifest.ts
+++ b/packages/remix-dev/compiler/manifest.ts
@@ -93,7 +93,14 @@ export async function create({
   invariant(entry, `Missing output for entry point`);
 
   optimizeRoutes(routes, entry.imports);
-  let version = getHash(JSON.stringify({ entry, routes })).slice(0, 8);
+
+  let version = getHash(
+    JSON.stringify({
+      entry,
+      routes,
+      cssBundleHref,
+    })
+  ).slice(0, 8);
 
   return { version, entry, routes, cssBundleHref, hmr };
 }
--- a/packages/remix-dev/compiler/manifest.ts
+++ b/packages/remix-dev/compiler/manifest.ts
@@ -94,15 +94,23 @@ export async function create({
 
   optimizeRoutes(routes, entry.imports);
 
-  let version = getHash(
-    JSON.stringify({
-      entry,
-      routes,
-      cssBundleHref,
-    })
-  ).slice(0, 8);
+  let fingerprintedValues = {
+    entry,
+    routes,
+    cssBundleHref,
+  };
 
-  return { version, entry, routes, cssBundleHref, hmr };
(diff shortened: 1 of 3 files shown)


packages/remix-dev/compiler/manifest.ts
```diff
@@ -93,9 +93,24 @@ export async function create({
   invariant(entry, `Missing output for entry point`);
 
   optimizeRoutes(routes, entry.imports);
-  let version = getHash(JSON.stringify({ entry, routes })).slice(0, 8);
 
-  return { version, entry, routes, cssBundleHref, hmr };
+  let fingerprintedValues = {
```
markdalgleish, line 97: I've refactored the code here to help people avoid the mistake of adding a new value to the manifest but forgetting to include it in the hash.

pcattori (approved):

Issue #3246, opened 2022-05-19
[Tutorial] Suggest `prisma migrate dev` to generate migration file somewhere


donrsh:
### What version of Remix are you using?

irravalent

### Steps to Reproduce

Hi. I just followed the tutoral (https://remix.run/docs/en/v1/tutorials/blog) to deploy an indie app to fly.io.

In the tutorial, there is a section on adding Post model, and it goes with `prisma db push`. After deploying, the post page shows an error since there is no migration file to be run. 

I guess there are some reasons the for the author not do so? Perhaps it would add some complexity to learn, or it gets hard to explain the seed? Anyway, it breaks the deployment, and I think at lease it should be mentioned about using `prisma migrate dev` somewhere.

### Expected Behavior

The post page should works after deployment.
See also: https://github.com/remix-run/indie-stack/issues/77

### Actual Behavior

The post page in  the tutorial (https://remix.run/docs/en/v1/tutorials/blog) breaks after deployment.

kentcdodds:
Agreed we should do this. If anyone would like to open a PR to add something to the tutorial that would be welcome :)

PR #6375, opened 2023-05-11
Update blog.md to create a migration without deleting your database

nickretallack:
The original instructions tell you to generate a migration after you've already done `prisma db push`.  This will prompt you to delete your database, which is not ideal, and then you'd need to seed it again anyway.  It would be better to just generate the migration instead of using `prisma db push` at all, since those commands are largely interchangeable.  I assume the existing migration instructions were hastily added in response to [this issue](https://github.com/remix-run/remix/issues/3246).

Also fixed some punctuation later.

### DIFF
--- a/docs/tutorials/blog.md
+++ b/docs/tutorials/blog.md
@@ -266,10 +266,10 @@ model Post {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


machour (approved):

PR #6382, opened 2023-05-12
docs(tutorials/jokes): use `&&` instead of ternary

te-ce:
Hey, this is just a simple change from a ternary `? `check to a `&&`, to simplify the code and don't have the `:null` at the end

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -478,6 +478,7 @@
 - tagraves
 - takagimeow
 - tascord
+- te-ce
 - TheRealAstoo
 - t-dub
 - therealflyingcoder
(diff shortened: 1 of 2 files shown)


te-ce:
Quite interesing and didn't thought of that.

te-ce:
But wouldn't this all be fixed when you don't use `contacts.length &&` as a condition but rather just `contacts &&` ?

kiliman:
The author of the tutorial has already stated why he uses ternaries, so I doubt this PR will be merged. Anyway it's basically a personal preference and does not impact the understanding of the tutorial.

PR #6383, opened 2023-05-12
Client side navigation ignores defer with nested route and URL param

oscarnewman:
This is a PR as a bug report for a weird issue I noticed with defer:

When navigating (client side) from one route with a deferred loader to it's child route, also with a deferred loader, the navigation is blocking (not deferred) IF there is ANY url param present on the child route.

I've put together a small test to reproduce the specific issue, though I haven't yet validated if there's more general cases where this might be triggered.

@jacob-ebey you might have figured this one out already but here's the repro I mentioned to you

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -535,3 +535,4 @@
 - TimonVS
 - yudai-nkt
 - TomVance
+- oscarnewman
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -48,28 +48,74 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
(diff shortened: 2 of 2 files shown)


brophdawg11:
Unfortunately, this is "expected" at the moment since changing a URL param triggers a revalidation on the parent route, and we don't re-fallback on revalidations of deferred data.  

You can work around this by telling your parent loader not to revalidate on GET navigations: 

```js
export function shouldRevalidate({ formMethod }) {
  return formMethod == null || formMethod === "GET";
}
```

There's some more info [in this comment](https://github.com/remix-run/remix/issues/6637#issuecomment-1600973794) but I think the hope is that eventually when `React.use` stabilizes we just give full control over when to fallback to the developer by just exposing the promise and letting you control it via `Suspense` keys (now that navigations are wrapped in `startTransition`)

PR #6400, opened 2023-05-15
feat(dev): add `vanillaExtract` option for opt-out

### LINKED ISSUE #6314: CSS Side-Effect Imports not working anymore with Cloudscape

1.16.0

- [x] Yes

### Steps to Reproduce

1. Pull https://github.com/joejemmely/remix-cloudscape
2. Run `npm run dev`

### Expected Behavior

Styles should be extracted and applied

### Actual Behavior

Styles are not extracted and applied. It used to work with `unstable_cssSideEffectImports` on 1.14.0

markdalgleish:
Closes #6314.

This allows consumers to explicitly opt-out of Vanilla Extract support, even if the `@vanilla-extract/css` package is requirable. This is important in cases where node_modules has been flattened and consumers inadvertently still have access to this package (since Remix depends on `@vanilla-extract/integration`, which depends on `@vanilla-extract/css`).

### DIFF
--- a/.changeset/vanilla-extract-flag.md
+++ b/.changeset/vanilla-extract-flag.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": minor
+---
+
+Add `vanillaExtract` config option to allow explicit opt-out of Vanilla Extract support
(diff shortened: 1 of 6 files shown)


markdalgleish:
I've opened a PR with a better approach for Vanilla Extract detection which avoids false positives (https://github.com/remix-run/remix/pull/6401), so this PR is no longer required to fix the linked issue.

PR #6401, opened 2023-05-16
fix(dev): detect Vanilla Extract in package.json

### LINKED ISSUE #6314: CSS Side-Effect Imports not working anymore with Cloudscape

markdalgleish:
Closes #6314.

This fixes an issue with our (currently unreleased) Vanilla Extract detection logic which is intended to disable Vanilla Extract support for consumers who aren't using it.

Because we're using `require.resolve` to check for the presence of the `@vanilla-extract/css` package, in the case of a flattened node_modules directory, some consumers can get false positives. This is because Remix depends on `@vanilla-extract/integration` to manage the esbuild integration, but that package depends on `@vanilla-extract/css`. I've seen this myself where `npx create-remix@latest` using npm to install dependencies results in `@vanilla-extract/css` being requirable from my app.

Instead, we now check the app's package.json to ensure that they have a direct dependency on `@vanilla-extract/css`.

### DIFF
--- a/integration/helpers/node-template/package.json
+++ b/integration/helpers/node-template/package.json
@@ -19,6 +19,7 @@
     "@remix-run/dev": "0.0.0-local-version",
     "@types/react": "0.0.0-local-version",
     "@types/react-dom": "0.0.0-local-version",
+    "@vanilla-extract/css": "^1.10.0",
     "typescript": "0.0.0-local-version"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "@types/retry": "^0.12.0",
     "@types/semver": "^7.3.4",
     "@types/ssri": "^7.1.0",
-    "@vanilla-extract/css": "^1.1.0",
+    "@vanilla-extract/css": "^1.10.0",
     "abort-controller": "^3.0.0",
     "abortcontroller-polyfill": "^1.7.3",
     "babel-jest": "^27.5.1",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -84,7 +84,7 @@
     "@types/retry": "^0.12.0",
     "@types/semver": "^7.3.4",
     "@types/ssri": "^7.1.0",
-    "@vanilla-extract/css": "^1.1.0",
+    "@vanilla-extract/css": "^1.10.0",
```
markdalgleish, line 87: This was already being resolved as 1.10.0 in yarn.lock. This is actually the minimum version required so I took the opportunity to tidy this up.

pcattori (approved):

integration/helpers/node-template/package.json
```diff
@@ -19,6 +19,7 @@
     "@remix-run/dev": "0.0.0-local-version",
     "@types/react": "0.0.0-local-version",
     "@types/react-dom": "0.0.0-local-version",
+    "@vanilla-extract/css": "^1.10.0",
```
markdalgleish, line 22: This is needed so our test suite detects Vanilla Extract by default.

PR #6408, opened 2023-05-16
feat(remix-dev): Use cache for HDR loader hashes

dmarkow:
Discussed with @pcattori -- the new approach for identifying loaders has esbuild running on every route for every build to extract loader hashes, adding significant rebuild time on large apps. With the latest nightlies, my rebuild times increased to around 8 seconds.

This is a first pass at adding caching based on route file modification times. This brings my rebuild times back down to 3-4s, similar to 1.16.0.

- [ ] Docs
- [x] Tests

Testing Strategy: Existing tests still pass, and running on my production app HMR/HDR is working as expected. I have *not* done extensive testing to see how this works when multiple overlapping rebuilds happen.

### DIFF
--- a/packages/remix-dev/devServer_unstable/hdr.ts
+++ b/packages/remix-dev/devServer_unstable/hdr.ts
@@ -1,5 +1,6 @@
 import * as path from "node:path";
 import esbuild from "esbuild";
+import * as fs from "node:fs";
 
 import type { Context } from "../compiler/context";
 import { emptyModulesPlugin } from "../compiler/plugins/emptyModules";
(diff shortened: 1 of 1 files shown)


dmarkow:
I didn't realize until reading 1.16.1's release notes that HDR should rerun loaders with changed code *and* loaders affected by changed code in imports (which seems obvious now). This PR's approach caches strictly based on the route files themselves, not any imports. So I'm actually not sure the best approach here other than what's being done now -- fully build all loaders to figure out what changed, and just find a way to do that really quickly.

pcattori:
We can keep track of a dependency tree (which we initialize during initial build). Then if a file is:

- added: imports for existing files were unchanged, so only need to consider HDR if this is a _route_ file
- changed: only check `loader` changes for routes whose descendants include the changed file
- deleted: same as changed^

pcattori:
@dmarkow closing this since we need to preserve the ability to detect changes across modules.

PR #6409, opened 2023-05-16
fix: reload page when routeModules doesn't contain module for current route

n8agrin:
This can happen when entry.client.ts isn't fully loaded before a route change occurs.

The fix is to validate the route's module is loaded, and if not, simply reload the page. Credit to @brophdawg11 who suggested this approach here: https://github.com/remix-run/remix/issues/1757#issuecomment-1401023829

See https://github.com/remix-run/remix/issues/1757 for more info.

Closes: #1757

- [x] Docs
- [x] Tests

Testing Strategy:

- added a new test (included in this PR) named browser-entry-test.ts

### DIFF
--- a/packages/remix-react/browser.tsx
+++ b/packages/remix-react/browser.tsx
@@ -19,6 +19,7 @@ import {
 /* eslint-disable prefer-let/prefer-let */
 declare global {
   var __remixContext: {
+    url: string;
     state: HydrationState;
     future: FutureConfig;
     // The number of active deferred keys rendered on the server
@@ -177,6 +178,21 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
           window.__remixContext.future.v2_normalizeFormMethod,
       },
     });
+
+    // Hard reload if the URL we tried to load is not the current URL.
+    // This is usually the result of 2 rapid backwards/forward clicks from an
+    // external site into a Remix app, where we initially start the load for
+    // one URL and while the JS chunks are loading a second forward click moves
+    // us to a new URL
+    let initialUrl = window.__remixContext.url;
+    let hydratedUrl = window.location.pathname + window.location.search;
+    if (initialUrl !== hydratedUrl) {
+      let errorMsg =
+        `Initial URL (${initialUrl}) does not match URL at time of hydration ` +
+        `(${hydratedUrl}), reloading page...`;
+      console.error(errorMsg);
+      window.location.reload();
+    }
   }
 
   let [location, setLocation] = React.useState(router.state.location);
@@ -188,17 +204,6 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
       }
     });
   }, [location]);
-
-  // Check if __remixRouteModules is missing the module for the current route
-  // and if so, load it.
-  for (let match of router.state.matches) {
-    if (!window.__remixRouteModules[match.route.id]) {
-      window.location.reload();
-      // `reload` executes asynchronously, meaning the error will still flash.
-      // To get around that we return an empty div.
-      return <div/>;
-    }
-  }
 
   // We need to include a wrapper RemixErrorBoundary here in case the root error
   // boundary also throws and we need to bubble up outside of the router entirely.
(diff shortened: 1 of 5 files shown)


n8agrin:
@brophdawg11 This is the PR I referenced here https://github.com/remix-run/remix/pull/5166#issuecomment-1548633436

packages/remix-react/browser.tsx
```diff
@@ -189,6 +189,17 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
   }, [location]);
 
+  // Check if __remixRouteModules is missing the module for the current route
+  // and if so, load it.
+  for (let match of router.state.matches) {
+    if (!window.__remixRouteModules[match.route.id]) {
+      window.location.reload();
+      // `reload` executes asynchronously, meaning the error will still flash.
+      // To get around that we return an empty div here.
+      return <div></div>;
```
n8agrin, line 199: @brophdawg11 this prevents a flash of the error I'm trying to work around. It still produces a set of console errors around problems hydrating the page (no surprise), which seem undesirable. If you have any thoughts on a more holistic way of dealing with this I'd appreciate your feedback. At the same time, this is a rare edge case, so maybe the brief error state resolved by the page reload is acceptable.

lpsinger:
```suggestion
      return <div/>;
```

brophdawg11:
Thanks for the PR @n8agrin - I pulled it down and did some testing - nice work getting an E2E test for this!

I made some minor alterations - let me know what you think:

* I moved the check up into the initialization block of `RemixBrowser` since we don't need it to run on all re-renders, just on the initial hydration render.
* I also changed the approach do to a simple URL comparison.  The `matches` comparison worried me a bit since I think it could potentially put us into a loop if a match for the initial URL was invalid (somehow got deleted from a CDN maybe).  If it had no chance to pass on the subsequent reload, we'd get into a reload loop I think?
  * Instead, now we're just SSR-ing the initial URL and comparing that against `window.location`.  Then if we find a mismatch and reload, that should always break the loop.
* I also removed the `return <div/>` since I don't know that it was much better of a UX.  In my testing I was still seeing a flash of the error - and I think that's fine in these small edge cases.

lpsinger:
I noticed I was tagged to review this, but I'm not a Remix maintainer. Was this a mistake?

brophdawg11:
@lpsinger I think that's only because of your comment above from last month

packages/remix-react/browser.tsx
```diff
@@ -177,6 +178,21 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
       },
     });
+
+    // Hard reload if the URL we tried to load is not the current URL.
+    // This is usually the result of 2 rapid backwards/forward clicks from an
+    // external site into a Remix app, where we initially start the load for
+    // one URL and while the JS chunks are loading a second forward click moves
+    // us to a new URL
+    let initialUrl = window.__remixContext.url;
+    let hydratedUrl = window.location.pathname + window.location.search;
```
n8agrin, line 188: Minor suggestion that this could be formalized into a fn which is used here and in server.ts to construct the string for matching. Would prevent a scenario where a future engineer sees `url` and assumes they could simply `window.location.toString()` this instead of relying only on the path and search params.

brophdawg11 (approved):

packages/remix-react/browser.tsx
brophdawg11: We can't share util functions like that across the package boundaries so we wouldn't be able to DRY it up much.  I think it' ok for now since I don't expect this to change much - and we're pretty familiar with the fact that on the server a URL is only ever `pathname` and `search`, never `hash` 👍

yarbsemaj:
Hello @n8agrin, this PR is causing issue with CDNs. Our CDN is caching page html with query params on from tracking links (e.g.) Google. This is then been compared to the URL another user is visiting and viewing the cached versions of the page, when the mismatch occurs the page gets into an infinite reload loop. I can see this affecting many users who also use CDN's like this.

PR #6425, opened 2023-05-18
Surface thrown response headers to ancestor boundaries

### LINKED ISSUE #5356: [Bug]: throwing a response with headers should preserve the headers
### What version of React Router are you using?

6.4+

### Steps to Reproduce

I put an example with Remix together for this, but I think the fix will be in the router which is why I'm opening this issue here (maybe the repos should be merged afterall?).

Here's the repo: https://github.com/kentcdodds/throw-response-headers

Here's a loom explaining the issue: https://www.loom.com/share/20a07a160bd54ec6bf228dcdc46f1ef3

### Expected Behavior

The short of it is, if you stick this in a loader:

```ts
  throw new Response("", { status: 418, headers: { "x-my-header": "sup" } });
}
```

The `x-my-header` doesn't show up in the response.

But if you do this:

```ts

  throw redirect("/", { headers: { "x-my-header": "sup" } });
}
```

The `x-my-header` does show up.

### Actual Behavior

I would expect throwing a response to preserve the headers of that response (like it does with a redirect).

brophdawg11:
We have a gap in the API currently where you cannot access headers from child-route-thrown-responses because we only run `headers()` from "renderable" matches (which means root down to the boundary).  So if you `throw new Response()` from `/parent/child` and catch it in `/parent` - you have no way to send any child headers back from the `/parent` `headers()` function.

This PR adds a new `errorHeaders` parameter to `headers()` that holds the headers from any thrown Response triggering the error boundary.

If the route that throws the response is also the boundary route, then `errorHeaders` will be the same as what's already passed as `loaderHeaders`/`actionHeaders`.

We also only expose the `errorHeaders` to the leaf renderable match `headers()` function to avoid duplicating the header value through `parentHeaders`.

Closes #5356

### DIFF
--- a/.changeset/expose-error-headers.md
+++ b/.changeset/expose-error-headers.md
@@ -2,4 +2,4 @@
 "@remix-run/server-runtime": minor
 ---
 
-Add `errorHeaders` parameter to `headers()` functions to expose headers from thrown responses that bubble up to ancestor route boundaries.  If the throwing route contains the boundary, then `errorHeaders` will be the same object from `loaderHeaders`/`actionHeaders` for that route.
+Add `errorHeaders` parameter to the leaf `headers()` function to expose headers from thrown responses that bubble up to ancestor route boundaries.  If the throwing route contains the boundary, then `errorHeaders` will be the same object as `loaderHeaders`/`actionHeaders` for that route.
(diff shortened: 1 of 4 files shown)


markdalgleish:
This API makes sense to me 👍

PR #6428, opened 2023-05-19
chore(remix-dev): upgrade `cacache` to v17

belgattitude:
cacache v15 depends on @npmcli/move-file 1.12 which have been deprecated and produce a warning at install.
 
```
@npmcli/move-file@npm:1.1.2 is deprecated: This functionality has been moved to @npmcli/fs
```

There's a similar closed P/R about it https://github.com/remix-run/remix/pull/5179 from January 2023. Reason about closing was about node support (at that time remix supported node 12). 

Situation changed since. Might be interesting to reconsider as it also include upstream fixes

Testing Strategy:

- impact only remix-dev
- checked the [changelog](https://github.com/npm/cacache/blob/main/CHANGELOG.md)
- checked bc of [current usage](https://github.com/remix-run/remix/blob/cb7a4e8fd2f3f6cf6840672905a7bf8e032849c7/packages/remix-dev/cache.ts#L1) (remix-dev depends only on get and put which signatures haven't changed)
- tried to find if re-exported get/put were actually used (which might impact requirements of dependent packages)

image

### DIFF
--- a/packages/remix-dev/package.json
+++ b/packages/remix-dev/package.json
@@ -31,7 +31,7 @@
     "@remix-run/server-runtime": "1.16.1",
     "@vanilla-extract/integration": "^6.2.0",
     "arg": "^5.0.1",
-    "cacache": "^15.0.5",
+    "cacache": "^17.1.3",
     "chalk": "^4.1.2",
     "chokidar": "^3.5.1",
     "dotenv": "^16.0.0",
@@ -100,7 +100,7 @@
     }
   },
   "engines": {
-    "node": ">=14"
+    "node": ">=14.17"
   },
   "files": [
     "dist/",
--- a/contributors.yml
+++ b/contributors.yml
@@ -54,6 +54,7 @@
 - axel-habermaier
 - barclayd
 - BasixKOR
+- belgattitude
 - BenMcH
 - benwis
 - bgschiller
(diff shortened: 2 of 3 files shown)


packages/remix-dev/package.json
```diff
@@ -100,7 +100,7 @@
     }
   },
   "engines": {
-    "node": ">=14"
+    "node": ">=14.17"
```
belgattitude, line 103: To keep aligned with cacache 17 new min requirements: 

https://github.com/npm/cacache/blob/2ae6d2d9dda028700e0bcfc7f0b5f8dc9d9c6e40/package.json#L67

based  on node 14 is deprecated / affect only remix-dev... my current understanding is that it should be safe

MichaelDeBoey:
@belgattitude We still can't update `cacache` as that would be a breaking change, because we support Node v14 and up (so also v15) completely & `cacache` @ v16 + v17's `engines` fields aren't compatible with that

v16:
```json
 "engines": {
    "node": "^12.13.0 || ^14.15.0 || >=16"
  },
```

v17:
```json
"engines": {
  "node": "^14.17.0 || ^16.13.0 || >=18.0.0"
},
```

Once we release v2 (which will drop support for older Node versions than v18), we can do this

belgattitude:
Hey @MichaelDeBoey, remix-dev is already requiring node 14... is there something I missed ?

MichaelDeBoey:
@belgattitude `remix-dev` requires Node `>=14` (which is the same as `>=14.0.0`), `cacache` requires `^14.15.0` (v16) or `^14.17.0` (v17), so this change would be a breaking change

belgattitude:
That I understand better. Thanks

PR #5473, opened 2023-02-15
Proposal: v2_headers flag to inherit parentHeaders

appden:
I've found it very surprising and have been bitten a few times that a parent route's headers are not automatically used by a child route when that child route does not export its own `headers`. This proposes doing just that for v2 since this would be a breaking change.

PR #6431, opened 2023-05-19
feat(server-runtime): v2_headers future flag for inheriting ancestor headers

### LINKED ISSUE #3157: Cookies set in thrown responses are ignored when the CatchBoundary is in a parent route

1.4.3

### Steps to Reproduce

1. Create a new remix app using `npx create-remix` and use the `Just the basics` and `Remix App Server` settings
2. Create the **parent layout** `~/app/routes/parent.tsx`:
```js
export function CatchBoundary() {
  return <main>I'm a catch</main>;
}

export default function Parent() {
  return (
    <main>
      <header>Parent</header>
      <Outlet />
    </main>
  );
}
```
3. Create the **child** `~/app/routes/parent/child.tsx`:
```js
export const loader: LoaderFunction = ({ request }) => {
  if (new URL(request.url).searchParams.has("throw")) {
    throw json(null, { headers: { "Set-Cookie": "thrownCookie=DoesntWork" } });
  }
  return json(null, { headers: { "Set-Cookie": "regularFlowCookie=ThisWorks" } });
};

export default function Child() {
  return <section>Child</section>;
}
```
4. Run the server `npm run dev`
5. `curl --head "localhost:3000/parent/child"`
6. `curl --head "localhost:3000/parent/child?throw"`

### Expected Behavior

The **Set-Cookie** header should be shown in both cases

### Actual Behavior

When the response is thrown in a child of a CatchBoundary, **Set-Cookie** headers are not copied

brophdawg11:
It can be tedious and error-prone to remember to always `export function headers()` from every potential leaf route if you care about response headers on document requests.  Prior to v1 the logic was to use the deepest renderable `headers` function that was found but this changed prior to v1 so we're adding it back behind a future flag for v2 to ensure no unintentional breakages in userland.

We still only call `headers` for "renderable" routes (those at or above the any active error boundaries), but we'll use the deepest discovered `headers` export, and not insist on it being a leaf.

Continuation of #5473 
Closes #3157


integration/headers-test.ts
```diff
@@ -122,15 +126,6 @@ test.describe("headers export", () => {
...
 
         "app/routes/parent.child.jsx": js`
-          export function headers({ actionHeaders, errorHeaders, loaderHeaders, parentHeaders }) {
-            return new Headers([
-              ...(parentHeaders ? Array.from(parentHeaders.entries()) : []),
-              ...(actionHeaders ? Array.from(actionHeaders.entries()) : []),
-              ...(loaderHeaders ? Array.from(loaderHeaders.entries()) : []),
-              ...(errorHeaders ? Array.from(errorHeaders.entries()) : []),
-            ]);
-          }
```
brophdawg11, line 132: This existing test section is now testing the v2 behavior so removed `headers` from the child.  See below for specific tests that continue to assert the v1 behavior.

integration/headers-test.ts
```diff
@@ -256,7 +245,6 @@ test.describe("headers export", () => {
     expect(JSON.stringify(Array.from(response.headers.entries()))).toBe(
       JSON.stringify([
         ["content-type", "text/html"],
-        ["x-child-loader", "success"],
```
brophdawg11, line 259: No more headers from children in these tests since we're using the parent `headers` function

packages/remix-server-runtime/headers.ts
```diff
@@ -37,6 +37,16 @@ export function getDocumentHeadersRR(
...
     let actionHeaders = context.actionHeaders[id] || new Headers();
+
+    // When the future flag is enabled, use the parent headers for any route
+    // that doesn't have a `headers` export
+    if (routeModule.headers == null && build.future.v2_headers) {
+      let headers = parentHeaders;
+      prependCookies(actionHeaders, headers);
+      prependCookies(loaderHeaders, headers);
+      return headers;
+    }
```
brophdawg11, line 48: There's no need to `prependCookies` for `parenHeaders` since that's what we're returning

docs/route/headers.md
```diff
@@ -118,5 +127,25 @@ export default function handleRequest(
...
+
+```
+├── users.tsx
+└── users
+    ├── $userId.tsx
+    └── $userId
+        └── profile.tsx
+```
+
+If we are looking at `/users/123/profile` and `profile.tsx` does not export a `headers` function, then we'll use the return value of `usersId.tsx`'s `headers` function. If that file doesn't export one, then w'ell use the result of the one in `users.tsx`, and so on.
```
brophdawg11, line 146:
```suggestion
If a user is looking at `/users/123/profile` and `profile.tsx` does not export a `headers` function, then Remix will use the return value of `$userId.tsx`'s `headers` function. If that file doesn't export one, then it will use the result of the one in `users.tsx`, and so on.
```

jacob-ebey (approved):

PR #6433, opened 2023-05-19
feat: add `prefetch="viewport"` support to `<Link>`

mcansh:

Closes: https://canary.discord.com/channels/770287896669978684/940264701785423992/1109137291974279239

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -222,7 +222,7 @@ interface PrefetchHandlers {
 function usePrefetchBehavior<T extends HTMLAnchorElement>(
   prefetch: PrefetchBehavior,
   theirElementProps: PrefetchHandlers
-): [boolean, React.RefObject<T> | null, Required<PrefetchHandlers>] {
+): [boolean, React.RefObject<T>, Required<PrefetchHandlers>] {
   let [maybePrefetch, setMaybePrefetch] = React.useState(false);
   let [shouldPrefetch, setShouldPrefetch] = React.useState(false);
   let { onFocus, onBlur, onMouseEnter, onMouseLeave, onTouchStart } =
@@ -276,7 +276,7 @@ function usePrefetchBehavior<T extends HTMLAnchorElement>(
 
   return [
     shouldPrefetch,
-    prefetch === "viewport" ? ref : null,
+    ref,
     {
       onFocus: composeEventHandlers(onFocus, setIntent),
       onBlur: composeEventHandlers(onBlur, cancelIntent),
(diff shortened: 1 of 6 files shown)


packages/remix-react/components.tsx
```diff
@@ -229,19 +229,35 @@ interface PrefetchHandlers {
...
     if (prefetch === "render") {
       setShouldPrefetch(true);
     }
+
+    if (prefetch === "viewport") {
+      let callback: IntersectionObserverCallback = (entries) => {
+        entries.forEach((entry) => {
+          setShouldPrefetch(entry.isIntersecting);
+        });
+      };
```
sergiodxa, line 253: Should the observer be disconnected after prefetch is set? That way the IntersectionObserver doesn't need to keep observing if `ref.current` is in the viewport

mcansh: mostly depends on if we should clean up link preloads when the link leaves the viewport. another thing would be that if it re-enters the vp, should it refetch the loader?

packages/remix-react/components.tsx
```diff
@@ -315,18 +333,18 @@ let Link = React.forwardRef<HTMLAnchorElement, RemixLinkProps>(
...
     );
 
     return (
       <>
         <RouterLink
-          ref={forwardedRef}
-          to={to}
           {...props}
           {...prefetchHandlers}
+          ref={mergeRefs(forwardedRef, ref)}
```
brophdawg11: Is there any internal react overhead in attaching a ref?  This feels potentially troublesome since we now attach a `ref` to _every_ Link, even if the user didn't forward a ref and if they're not using `prefetch="viewport"`.  Does a page with hundreds of links get immediately "slower" with no way for the user to opt out of the refs?

I took a stab at an opt-in ref approach in 44f28f3a2e514e114a818e64c3ffdbca6d5be8f0

brophdawg11 (approved):
Made a small change in https://github.com/remix-run/remix/pull/6433/commits/44f28f3a2e514e114a818e64c3ffdbca6d5be8f0 to only attach the `ref` when prefetch="viewport" is specified so we're not attaching a ton of refs if the viewport option is not used.

packages/remix-react/components.tsx
brophdawg11: There's no perf hit per @jacob-ebey so going to back out that commit

brophdawg11:
Think this should be good to go now.  I'm OOO tomorrow but will see if Jacob can merge once CI passes

Issue #6415, opened 2023-05-18
HMR not working in `cloudflare-pages` template


Jackardios:
### What version of Remix are you using?

1.16.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. `npx create-remix@latest --template https://github.com/remix-run/remix/tree/templates_v2_dev/templates/cloudflare-pages`
2. `npm run dev`
3. change something in the `/app/routes/_index.tsx` file
4. open console and see:
<img width="738" alt="console-error" src="https://github.com/remix-run/remix/assets/24757335/47ac2f61-1ca7-4e06-9013-34148873c4e4">

### Expected Behavior

HMR should work after changing the file

### Actual Behavior

After changing the file, HMR does not work, and an error appears in the console, the changes can only be seen if you manually reload the page. Perhaps this is due to the race condition #6286

pcattori:
Can you manually `GET http://localhost:8788/build/routes/_index-<whatever hash>.js?t=<whatever timestamp>` in a separate browser window or via `curl`? Just to see if the app server can correctly serve that updated asset.

Jackardios:
@pcattori Yes, it exists if you go manually along this path

Jackardios:
@pcattori it looks like it really is due to a race condition. If you set timeout to logDevReady no error occurs:
```ts
// server.ts
import { logDevReady } from "@remix-run/cloudflare";
import { createPagesFunctionHandler } from "@remix-run/cloudflare-pages";
import * as build from "@remix-run/dev/server-build";

if (process.env.NODE_ENV === "development") {
  setTimeout(() => logDevReady(build), 200);
}

export const onRequest = createPagesFunctionHandler({
  build,
  getLoadContext: (context) => ({ env: context.env }),
  mode: process.env.NODE_ENV,
});

```

huw:
Hey, can you clarify whether you’re using Wrangler v2 or Wrangler v3? The reason being is that the new `--local` mode (v3; was `--experimental-local` in v2) doesn’t completely shut down the development server ([see the old behaviour here](https://github.com/cloudflare/workers-sdk/blob/5fd080c88ee7991cde107f8723f06ea2fd2c651d/packages/wrangler/src/dev/local.tsx#L406), which only applies to the old server).

Given that the old server is deprecated, I think the appropriate fix is to upgrade the Wrangler dependency in the v2 templates (I’ll suggest that fix over there)

[...]

PR #6436, opened 2023-05-20
Update wrangler to v3 in `cloudflare-pages` template

Jackardios:
Closes: #6415

In addition to updating wrangler to v3, I also removed `wrangler.toml` (since it is not supported by `wrangler pages dev`) and added the latest compatibility-date as an option to invoke the `wrangler pages dev` command

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -202,6 +202,7 @@
 - itsMapleLeaf
 - izznatsir
 - jacargentina
+- Jackardios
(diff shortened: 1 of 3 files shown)


MichaelDeBoey:
Thanks for taking the time to implement these changes @Jackardios!

Could you please add these changes as suggestions on #6229 instead?
Going to close this as we don't need to have another PR for it.

PR #6441, opened 2023-05-20
feat(dev): add BrowserMeta file for @remix-run/dev

19Qingfeng:
Closes: #6416 

- [ ] Docs
- [x] Tests

I opened up my may machine and ran this script:

1. Create a playgroud project, modify `metafile: true` in remix.config.js, and run `npm run build` to compile remix-run project.
After that, look at `debugDirectory`(`public/.debugger`) to get the `js-metafile.json` file.

2. Create a playgroud project, modify `metafile: true` and `debugDirectory: ".debugger"`  in remix.config.js, and run `npm run build` to compile remix-run project.
After that, look at `debugDirectory`(`.debugger`) to get the `js-metafile.json` file.

3. Create a playground project, execute `remix build --metafile`,then you will get `js-metafile.json`.


pcattori:
@19Qingfeng thank you for the PR!

I think the direction we want to go in is to have a `--metafile` option that produces the metafile(s) into a dedicated directory.
Maybe default to `.debug/` and configurable with `debugDirectory` option in `remix.config.js` or something like that.

19Qingfeng:
> @19Qingfeng谢谢你的公关！
> 
> 我认为我们想要进入的方向是有一个`--metafile`选项可以将图元文件生成到一个专用目录中。 也许默认为`.debug/`并可使用`debugDirectory`选项 in`remix.config.js`或类似的东西进行配置。



> @19Qingfeng thank you for the PR!
> 
> I think the direction we want to go in is to have a `--metafile` option that produces the metafile(s) into a dedicated directory. Maybe default to `.debug/` and configurable with `debugDirectory` option in `remix.config.js` or something like that.

Okay, I will refer to your feedback to modify the parameter name and generation location. I will resubmit it later. Thank you.

19Qingfeng:
> @19Qingfeng thank you for the PR!
> 
> I think the direction we want to go in is to have a `--metafile` option that produces the metafile(s) into a dedicated directory. Maybe default to `.debug/` and configurable with `debugDirectory` option in `remix.config.js` or something like that.

@pcattori hello,pcattori.Now, this time with the metafile and debugDirectory parameters, I will update the test description later.

packages/remix-dev/config.ts
```diff
@@ -268,6 +278,17 @@ export interface RemixConfig {
...
+  /**
+   * The absolute path to the metafile build directory.
+   */
+  debugDirectory: string
+
+  /**
+   * Whether to support generate browser js metafile for esbuild, default to false.
+   * When you set it to true, remix-js-metafile.json will be generated in the assetsBuildDirectory directory.
+   */
+  metafile?: boolean;
```
pcattori, line 290: `metafile` should be specifiable as a config option as you have here, but also as a flag `--metafile` for both `remix build` and `remix dev`.

packages/remix-dev/compiler/compiler.ts
```diff
@@ -108,12 +113,12 @@ export let create = async (ctx: Context): Promise<Compiler> => {
     options.onManifest?.(manifest);
     writes.manifest = writeManifest(ctx.config, manifest);
 
-    // server compilation
```
pcattori, line 111: Could you restore this comment?

packages/remix-dev/compiler/utils/analyzer.ts
```diff
@@ -0,0 +1,44 @@
...
+            fs.mkdirSync(debugDirectory)
+          }
+          return fs.writeFile(normalize(output), JSON.stringify(outputMeta));
+        } catch (e) {
+          invariant(e, `Failed to generate ${JS_META_FILE_NAME} in ${debugDirectory}.`);
+        }
+      }
+    }
+  }
+}
```
pcattori, line 41: Could you rewrite this as a function `writeMetafile( ctx: Context, name: string, metafile: Metafile)`?

packages/remix-dev/compiler/compiler.ts
```diff
@@ -96,6 +100,7 @@ export let create = async (ctx: Context): Promise<Compiler> => {
     let js = await tasks.js;
     if (!js.ok) throw error ?? js.error;
     let { metafile, hmr } = js.value;
+    await createBrowserMetaFile(metafile)
```
pcattori, line 103: Conditional should be visible at this level:

```suggestion
    if (ctx.config.metafile) {
      await writeMetafile( ctx, "js", metafile)
    }
```

pcattori (changes requested):

packages/remix-dev/compiler/compiler.ts
19Qingfeng: Okay, this is really my oversight.

packages/remix-dev/compiler/compiler.ts
19Qingfeng: All right, I'll take care of it later.

packages/remix-dev/compiler/compiler.ts
19Qingfeng: In fact, a better solution is that I think metafile should be turned off by default under dev.

packages/remix-dev/config.ts
19Qingfeng: Currently, metafile in my code will also be enabled according to the configuration when you use remix dev. But actually, I think maybe it's a good idea to shut down generate metafile all the time under remix dev.

19Qingfeng:
@pcattori hi, i updated the code and synchronized the description, please check the changes.

19Qingfeng:
@pcattori Can you provide some feedback? Perhaps I can make further modifications

Doelfke:
What's the status of this?

19Qingfeng:
> What's the status of this?

I'm not sure what the remix team is thinking right now. If I have an idea, Maybe I can resolve the conflict and modify the code.

PR #6461, opened 2023-05-23
fix typo in jokes app tutorial

smeijer:
The session secrets are used to `sign` the cookie, not to `encrypt` it. This mistake can lead to dangerously wrong assumptions 🙂


kentcdodds (approved):
Thank you! That was probably my typo

PR #6488, opened 2023-05-25
fix(remix-dev): loosen `prettier` version

### LINKED ISSUE #5444: Relax version of Prettier in `package.json`

v1.12.0

- [X] Yes

### Steps to Reproduce

I upgraded Prettier in my project to v2.8.4, but I see that v2.7.1 keeps installed in parallel because it's a dependency of `@remix-run/dev`.

### Expected Behavior

Unless there's a good reason not to, Remix should use a `^` version for Prettier:
```diff
-  "prettier": "2.7.1",
+  "prettier": "^2.7.1",
```

### Actual Behavior

It's currently pinned to the specific `2.7.1` version.
https://github.com/remix-run/remix/blob/9adf87c663207e687fbb8e89d2ee4868504fc52c/packages/remix-dev/package.json#L59

MichaelDeBoey:
Closes #5444

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "lodash": "^4.17.21",
     "npm-run-all": "^4.1.5",
     "patch-package": "^6.5.0",
-    "prettier": "2.7.1",
+    "prettier": "^2.7.1",
     "prompt-confirm": "^2.0.4",
     "react": "^18.2.0",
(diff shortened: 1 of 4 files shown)


roninliu:
what's time publish the version in your plan?

PR #6505, opened 2023-05-29
fix: add caching to PostCSS for regular stylesheets

markdalgleish:
For now this is focused on regular stylesheets. Future PRs will extend this to more PostCSS touch points.

This PR introduces a generic file watch cache that can be used to cache async file processing and invalidate cache entries whenever relevant files are changed. I've integrated this into both old and new dev servers, and for consistency I've also integrated it into the build process so that plugins don't need to behave differently between dev and build and can always assume the presence of the cache.

Note that our watcher logic in `watch.ts` needed to be patched so that watch paths are always absolute. I've added a comment explaining why this is the case, but basically we need to _always_ watch absolute paths so that Chokidar provides absolute paths in change events, which is important to ensure consistent values for the cache.

To test this, I've added some more complex PostCSS usage to our HMR test.


jacob-ebey (approved):
I like the cache abstraction, should serve us well.

PR #6508, opened 2023-05-30
test: allow config overrides for Playwright tests

markdalgleish:
This came out of some other compiler work I was doing but was big enough to warrant a PR of its own.

This builds on the same concept we were using for the `future` option on Playwright test fixtures but expands it to support the entire Remix config object. This means it's now super easy to alter the Remix config for a Playwright test without having to manually create a `remix.config.js` file.

Note that in order to support regular expressions and functions, I've switched from `JSON.stringify` to [serialize-javascript](https://www.npmjs.com/package/serialize-javascript).


integration/compiler-mjs-output-test.ts
```diff
@@ -8,8 +8,18 @@ let projectDir: string;
...
+      serverBuildPath: "build/index.mjs",
+      future: { v2_routeConvention: true },
+    },
     files: {
+      // Ensure the config is valid ESM
+      "remix.config.js": js`
+        export default {
+          ...global.INJECTED_FIXTURE_REMIX_CONFIG
+        };
+      `,
```
brophdawg11, line 22: This feels potentially confusing to me having the config come from 2 spots.  Would it be easier to grok if we made these two approaches mutually exclusive?

brophdawg11 (approved):
Love the extension of `future -> config` 👌   

Just one question about potential confusion when both are provided and if it would make sense to just enforce mutual exclusivity.  Doesn't seem like it would be prohibitive to have to put future flags and other configs in `files["remix.config.js"]` in the few cases you need to do a manual file (ESM for example)

integration/compiler-mjs-output-test.ts
markdalgleish: Yep, this was the part I was least sure about. I'll revisit this.

PR #6509, opened 2023-05-31
fix(dev): skip CSS build on missing css-bundle dep

markdalgleish:
Since you can only access the output of the CSS bundle build via the `@remix-run/css-bundle` package, we can detect the _lack_ of this package and use this to avoid running an esbuild pass for no reason.

To make this easier to test (since our monorepo unavoidably has this package available), and also to give consumers more explicit control over this behaviour, I've added a `cssBundle` option to the Remix config, but I'm leaving it undocumented for now.


integration/tailwind-test.ts
```diff
@@ -358,12 +359,10 @@ test.describe("Tailwind enabled via unstable future flag", () => {
         `,
         "app/root.jsx": js`
           import { Links, Outlet } from "@remix-run/react";
-          import { cssBundleHref } from "@remix-run/css-bundle";
```
markdalgleish, line 361: This is just a clean-up since this fixture doesn't use any CSS bundling features.

scripts/copy-build-to-dist.mjs
```diff
@@ -141,9 +141,9 @@ async function getPackageBuildPaths(moduleRootDir) {
       if (path.basename(moduleDir) === "@remix-run") {
         packageBuilds.push(...(await getPackageBuildPaths(moduleDir)));
       } else if (
-        /node_modules\/@remix-run\//.test(moduleDir) ||
-        /node_modules\/create-remix/.test(moduleDir) ||
-        /node_modules\/remix/.test(moduleDir)
+        /node_modules[/\\]@remix-run[/\\]/.test(moduleDir) ||
+        /node_modules[/\\]create-remix/.test(moduleDir) ||
+        /node_modules[/\\]remix/.test(moduleDir)
```
markdalgleish, line 146: This fixes an issue where the built packages weren't being copied to the `dist` directories on Windows. This tripped me up because it meant that the `readConfig` tests passed on Mac/Linux but failed on Windows. This is because `readConfig` detects the presence of the `@remix-run/css-bundle` package, but this package wasn't being detected on Windows.

integration/css-modules-test.ts
```diff
@@ -20,6 +20,7 @@ test.describe("CSS Modules", () => {
   test.beforeAll(async () => {
     fixture = await createFixture({
       config: {
+        cssBundle: true,
```
jacob-ebey, line 23: can we name it `unstable_cssBundle` and type it as just `false`? Make it a hard opt-out, as it needs to be ran if the package is included anyways.

jacob-ebey:
Going to attempt a delayed initialization of the css bundle based on the conditional include of `@remix-run/css-bundle`.

markdalgleish:
Closing for now since we're going to try a different approach.

PR #6551, opened 2023-06-06
chore: use long version numbers

MichaelDeBoey:
This can prevent confusion like https://github.com/remix-run/remix/pull/6428#issuecomment-1579534404


brophdawg11:
Is this a common confusion?  I would assume `>=14` is interpreted as `>=14.0.0`

MichaelDeBoey:
@brophdawg11 I would assume as well, but I've come across people that are confused about this already a couple of times now (all in different repo's), hence the update

PR #6570, opened 2023-06-08
Add support for json/text submissions

brophdawg11:
Make required changes and add tests for the Remix portion of https://github.com/remix-run/react-router/pull/10413 which adds support for `application/json` and `text/plain` submissions

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1456,8 +1456,10 @@ export function useFetchers(): Fetcher[] {
       data: f.data,
       formMethod: f.formMethod,
       formAction: f.formAction,
-      formData: f.formData,
       formEncType: f.formEncType,
+      formData: f.formData,
+      json: f.json,
+      text: f.text,
       " _hasFetcherDoneAnything ": f[" _hasFetcherDoneAnything "],
     });
     addFetcherDeprecationWarnings(fetcher);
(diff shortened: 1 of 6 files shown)


packages/remix-react/data.ts
```diff
@@ -40,12 +40,23 @@ export async function fetchData(
...
+      init.headers = { "Content-Type": contentType };
+      init.body = await request.text();
+    } else if (
+      contentType &&
+      /\bapplication\/x-www-form-urlencoded\b/.test(contentType)
+    ) {
+      init.body = new URLSearchParams(await request.text());
+    } else {
+      init.body = await request.formData();
+    }
```
brophdawg11, line 59: Ensure we proxy the RR request over the wire to Remix with the correct content type and `body`

packages/remix-react/components.tsx
```diff
@@ -1573,22 +1587,29 @@ function convertRouterFetcherToRemixFetcher(
     formMethod &&
     formAction &&
     formEncType &&
-    formData
+    (formData || json || text)
```
jacob-ebey, line 1590: do we want both `json !== undefined` && `text != null` here? This would block null and empty string JSON submission, and empty string text submissions. If the "json" value here is pre-encoded (a json string) that one is fine, but the text case I think would cause issues.

integration/form-test.ts
```diff
@@ -1006,12 +1006,7 @@ test.describe("Forms", () => {
 
     test("submits the submitter's value(s) in tree order in the form data", async ({
       page,
-      javaScriptEnabled,
     }) => {
-      test.fail(
-        Boolean(javaScriptEnabled),
-        "<Form> doesn't serialize submit buttons correctly #4342"
```
brophdawg11, line 1013: It does now!

PR #6597, opened 2023-06-13
fix(dev): switch to JEST_WORKER_ID environment variable

kentcdodds:
Apparently, ESLint sets the NODE_ENV to test and if you are using a eslint plugin that relies on this code you'll end up requiring when you should be importing.

Because this block is for Jest specifically, we'll instead use a known environment variable in a Jest environment.

- ~~[ ] Docs~~
- ~~[ ] Tests~~

Testing Strategy: Run CI. I also verified this solves my use case with a local patch.


kentcdodds:
Related issue: https://github.com/jenseng/eslint-plugin-remix-react-routes/issues/16

kentcdodds:
Closed this because the eslint plugin is where the fix will be, but I'm reopening because I still think this change makes sense...

PR #6600, opened 2023-06-13
templates(vercel): deprecate template

MichaelDeBoey:
Just like in #5964, I added the deprecation warning to the template, but now on `main`


brophdawg11:
Can we wait for the next release and then the changes in `dev` will merge down to `main` automatically?  It seems confusing to have this in the live docs but not showing up against currently released packages.

PR #6609, opened 2023-06-14
Update to react router 6.13.0

brophdawg11:


packages/remix-react/browser.tsx
```diff
@@ -207,7 +207,11 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
         location={location}
         component={RemixRootDefaultErrorBoundary}
       >
-        <RouterProvider router={router} fallbackElement={null} />
+        <RouterProvider
+          router={router}
+          fallbackElement={null}
+          future={{ v7_startTransition: true }}
```
brophdawg11, line 213: This keeps us functionally equivalent to `6.12.0`/`6.12.1`.  We only added the future flag in React Router due to the large number of existing apps using `Suspense` in incompatible ways.  Remix is a much smaller footprint so doesn't warrant the same type of flag and we can just use `startTransition` by default as we're supposed to

jacob-ebey (approved):

brophdawg11:
Cherry-picked 0ab3e9a753761f4688f00e4f259211d4e72964cc into here to see if it helps with the failed E2E tests 🤞

PR #6615, opened 2023-06-15
feat(dev): stabilize v2 dev server

pcattori:
See new docs

### DIFF
--- a/.changeset/clever-ways-love.md
+++ b/.changeset/clever-ways-love.md
@@ -1,0 +1,9 @@
+---
+"@remix-run/dev": minor
+"@remix-run/react": minor
+"@remix-run/serve": minor
+"@remix-run/server-runtime": minor
+"@remix-run/testing": minor
+---
+
+stabilize v2 dev server
--- a/docs/other-api/adapter.md
+++ b/docs/other-api/adapter.md
@@ -1,6 +1,6 @@
 ---
 title: "@remix-run/{adapter}"
-order: 2
+order: 3
 ---
 
 # Server Adapters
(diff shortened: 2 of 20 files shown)


jacob-ebey (approved):

xHomu:
Can confirm about 25% improvement in rebuild time (3.2s -> 2.5s) on an WIndows i7-7700 setup with the latest nightly build!

## Before `v1.17.0`

image

## After `0.0.0-nightly-25620b9-20230618`

image

#6539 @lili21 can you confirm?

lili21:
can confirm, rebuild time goes down from 1.9 to 1.4s。

pcattori:
@xHomu can you `console.log(build)` right before you call `broadcastDevReady`? Curious what that looks like since it should have a `assets.version` set.

xHomu:
> @xHomu can you `console.log(build)` right before you call `broadcastDevReady`? Curious what that looks like since it should have a `assets.version` set.

@pcattori  This is on an CJS app, so we're using require(BUILD_DR) instead: https://github.com/manawiki/starrail/blob/main/core.server.ts

image

xHomu:
@pcattori I'm not sure if it's entirely related, but I noticed another weird bug introduced between `0.0.0-nightly-25620b9-20230618` and `1.18.0-pre.0`. 

Weird .css.map error with chrome devtools on:

image
something wrong

Seems like it's expecting a sourcemap file for .css as well?

This one is reproduceable, you should able build this locally https://github.com/manawiki/starrail/tree/remix_v1.18_pre_bug

Above error shows on Windows with devtools on, switch to `0.0.0-nightly-25620b9-20230618` the error goes away.

pcattori:
@xHomu if you are still encountering these issues, could you create separate GitHub issues?

PR #6641, opened 2023-06-19
docs(jokes.md):  fix code snippet order

femincan:
In the `Mutations` section of the `App Tutorial (Long)` page, I rearranged the order of the code snippets. The code snippet of `app/routes/jokes.new.tsx` now comes after the code snippet of `app/utils/request.server.ts`. Because the first file utilizes the function exported from the second file. Therefore, I reversed the order.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -150,6 +150,7 @@
 - exegeteio
 - F3n67u
 - federicoestevez
+- femincan
 - fergusmeiklejohn
 - fernandojbf
 - fgiuliani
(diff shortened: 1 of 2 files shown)


machour (commented):
The previous paragraph reads:

> Go ahead and validate that the `name` and `content` fields are long enough. I'd say the name should be at least 3 characters long and the content should be at least 10 characters long. Do this validation server-side.

I think it makes more sense to keep the order as it is now, since we're solving the task at hand first. And if the reader wonders what's `badRequest` is all about, the answer is right below.

Thank you for your contribution nonetheless!

PR #6642, opened 2023-06-20
feat(remix-server-runtime): use `type-fest`'s `Jsonify` type in `Serialize`

MichaelDeBoey:
- Extracted `Jsonify` so that it's easier to switch to `type-fest` at a later stage
- Added cases for positive/negative infinity, `Boolean`, `Number`, `String`, `Map` & `Set`
- Added more deep grained type for objects that have a `toJSON` function
- Update `UndefinedToOptional` type
- Renamed some generics so they're more in line with the `type-fest` codebase
  - `NonJsonPrimitive` -> `NotJsonable`
    - Also replaced `Function` with `(...arguments_: any[]) => any`
  - `SerializeTuple` -> `JsonifyList`
    - Also updated this
  - `SerializeObject` -> `JsonifyObject`
    - Also updated this

This is mainly taking over @sachinraja's awesome work from https://github.com/sindresorhus/type-fest/pull/519, thanks for that! 🙏

### DIFF
--- a/packages/remix-server-runtime/package.json
+++ b/packages/remix-server-runtime/package.json
@@ -21,7 +21,8 @@
     "@web3-storage/multipart-parser": "^1.0.0",
     "cookie": "^0.4.1",
     "set-cookie-parser": "^2.4.8",
-    "source-map": "^0.7.3"
+    "source-map": "^0.7.3",
+    "type-fest": "^4.0.0"
   },
   "devDependencies": {
     "@remix-run/web-file": "^3.0.3",
--- a/packages/remix-server-runtime/serialize.ts
+++ b/packages/remix-server-runtime/serialize.ts
@@ -1,76 +1,27 @@
+import type { Jsonify } from "type-fest";
+
 import type { AppData } from "./data";
 import type { TypedDeferredData, TypedResponse } from "./responses";
 
-type JsonPrimitive =
-  | string
-  | number
-  | boolean
-  | String
-  | Number
-  | Boolean
-  | null;
-type NonJsonPrimitive = undefined | Function | symbol;
+// Note: The return value has to be `any` and not `unknown` so it can match `void`.
+type ArbitraryFunction = (...args: any[]) => any;
+type NotJsonable = ArbitraryFunction | undefined | symbol;
 
-/*
- * `any` is the only type that can let you equate `0` with `1`
- * See https://stackoverflow.com/a/49928360/1490091
- */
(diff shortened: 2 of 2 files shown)


pcattori:
@MichaelDeBoey given that we have [abandoned pretty types](https://github.com/remix-run/remix/pull/6746#issuecomment-1619068201) (at least for now), I think we are set to adopt `type-fest` types for JSON serialization

MichaelDeBoey:
@pcattori I also applied all the changes from this repo into `type-fest`, but that does seem to break a test case ( see https://github.com/sindresorhus/type-fest/pull/638)
Any idea what would be the fix?

I'd love to get it fixed upstream & here

pcattori:
@MichaelDeBoey left a [fix for the upstream `type-fest` PR](https://github.com/sindresorhus/type-fest/pull/638/files/6de59840c17c36c3b9cb4c2e6708acae2b33643b#r1255925547)

MichaelDeBoey:
@pcattori I've updated this PR to the latest version of what I've done in https://github.com/sindresorhus/type-fest/pull/638 as well

pcattori:
@MichaelDeBoey I think the goal for this PR should be to replace our types with `type-fest` entirely after the upstream PR lands, rather than just update our types to match theirs.

MichaelDeBoey:
@pcattori Unfortunately that's not possible (yet) as `type-fest` requires Node@>=14.16, which would be a breaking change as all our packages currently require Node@>=14.0.0

`type-fest` also requires Typescript@>=4.7
As we're currently not explicit about which versions we support, I don't think this is a problem though

MichaelDeBoey:
@pcattori It seems like there's still a problem with the `Jsonify` type 🤔
https://github.com/sindresorhus/type-fest/pull/638#issuecomment-1627802167

pcattori:
@MichaelDeBoey since `dev` branch is now for v2 development and requires Node >=18 and TS >= 5.1, can we now switch this to just pull in the types from `type-fest`?

MichaelDeBoey:
@pcattori I now changed this PR to use `type-fest`

pcattori (approved):

smeijer:
Would it make sense to revert this PR till there is a stable solution? The `Jsonify` thing seems to be a serious issue. A lot of my loaders now return `Jsonify({})`, while they were fully typed in Remix v1.9.x.

FYI, I'm typing my loaders using the `DataFunctionArgs` type, like:

```ts
export async function loader({  }: DataFunctionArgs) {
  return defer({ obj });
}
```

PR #6650, opened 2023-06-20
chore(templates): integrate v2_dev

pcattori:
Supercedes #6229

Deno template is not changed as their is an existing bug with that template. Will fix and update Deno template as a follow-on.


templates/cloudflare-workers/remix.config.js
```diff
@@ -1,10 +1,12 @@
 /** @type {import('@remix-run/dev').AppConfig} */
-module.exports = {
-  devServerBroadcastDelay: 1000,
+export default {
   ignoredRouteFiles: ["**/.*"],
   server: "./server.ts",
   serverConditions: ["worker"],
```
huw, line 5: [Same comment I left on the last one](https://github.com/remix-run/remix/pull/6229#discussion_r1181993844)—IMHO this should be `["workerd", "worker", "browser"]`

MichaelDeBoey:
@pcattori Should we get these changes in the stacks as well?

PR #6654, opened 2023-06-21
perf(dev): skip Remix/React packages in CSS build

markdalgleish:
The CSS bundle build ultimately only cares CSS imports and these packages are known not to contain any, so this PR allows esbuild to skip compilation of these packages entirely. In my basic example app using CSS bundling, I've seen builds go from ~170ms to ~120ms.

### DIFF
--- a/.changeset/css-bundle-skip-packages.md
+++ b/.changeset/css-bundle-skip-packages.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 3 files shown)


kiliman:
Hi. Just curious about the decision to exclude packages instead of creating a whitelist in *remix.config*. I'd assume we would just include `app/*` and exclude `node_modules/*` by default. Then let the developer include specific folders to scan for CSS files.

pcattori:
@kiliman this was a way of improving perf without changing API. Though I personally agree that we should explore the approach you suggest as well.

PR #6679, opened 2023-06-23
Avoid re-prefetching stylesheets for active routes during a revalidation

### LINKED ISSUE #5448: UseFetcher causes links to be loaded again

1.2.0

- [X] Yes

### Steps to Reproduce

add link.  
```javascript
/*root.tsx*/
import {cssBundleHref} from '@remix-run/css-bundle';
import iconfontCss from './styles/iconfont/iconfont.css'
import globalCss from './styles/global.css'

export const links = () => {
  return [
    {
      rel: "stylesheet",
      href: iconfontCss,
    },
    {
      rel: "stylesheet",
      href: globalCss,
    },
    ...(cssBundleHref ? [{ rel: "stylesheet", href: cssBundleHref }] : []),
  
  ];
};

```
create  user login api route.  
```javascript
/*routes/apis_.user.login.tsx*/
import { ActionArgs } from "@remix-run/server-runtime";
export async function action({}:ActionArgs){
    return await new Promise((resolve,reject)=>{
        setTimeout(() => {
            resolve("login success");
        }, 3000);
    })
}


```
use `useFetcher` .
```JavaScript
/*routes/login.tsx*/
function Login() {
var fet=useFetcher();
useEffect(()=>{
     fet.submit({},{
      method:'post',
      action:'/apis/user/login',
   
  })},[]);


/*....*/
 }
```

### Expected Behavior

just load one time css link.  like this:
image

### Actual Behavior

after fetch.submit api route. the css link in root.tsx load again.
image

by the way, I used css module.

brophdawg11:
**TODO**
* [ ] See if there's any way to short circuit any earlier in the flow?  Might not be since `links()` could return new stuff
* [ ] Tests

Closes #5448

### DIFF
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -239,6 +239,7 @@ export function getLinksForMatches(
 export async function prefetchStyleLinks(
   routeModule: RouteModule
 ): Promise<void> {
+  debugger;
   if (!routeModule.links) return;
   let descriptors = routeModule.links();
   if (!descriptors) return;
@@ -254,9 +255,14 @@ export async function prefetchStyleLinks(
     }
   }
 
-  // don't block for non-matching media queries
+  // don't block for non-matching media queries, or for stylesheets that are
+  // already in the DOM (active route revalidations)
   let matchingLinks = styleLinks.filter(
-    (link) => !link.media || window.matchMedia(link.media).matches
+    (link) =>
+      (!link.media || window.matchMedia(link.media).matches) &&
+      document.querySelector(
+        `head link[rel="stylesheet"][href="${link.href}"]`
+      ) != null
   );
 
   await Promise.all(matchingLinks.map(prefetchStyleLink));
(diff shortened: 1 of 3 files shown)


packages/remix-react/links.ts
```diff
@@ -254,9 +254,14 @@ export async function prefetchStyleLinks(
...
 
-  // don't block for non-matching media queries
+  // don't block for non-matching media queries, or for stylesheets that are
+  // already in the DOM (active route revalidations)
   let matchingLinks = styleLinks.filter(
-    (link) => !link.media || window.matchMedia(link.media).matches
+    (link) =>
+      (!link.media || window.matchMedia(link.media).matches) &&
+      document.querySelector(
+        `head link[rel="stylesheet"][href="${link.href}"]`
```
jacob-ebey, line 263: Do we care about the head specifically? If a user manually renders a link tag in the root component, but in the body for some reason, should that not also be deduped?

brophdawg11: nah - we can remove it in case they do any manual rendering of their own or the user puts `<Links>` into `body` for some reason

jacob-ebey (approved):

PR #6689, opened 2023-06-26
add support for specifying the live reload url

marshallwalker:
The live reload component does not work in github code-spaces since the port is apart of the URL. it would be nice to have a way to supply an alternate URL to use for the live reloading. My current solution is to copy the live reload component and add it back with this modification. so I figured why not open a PR.

If you want it to be a different name for the env please let me know :)


marshallwalker:
I'm not sure why some of these test are not passing, if any who review this could let me know I will try and fix

giovannibenussi:
I did run the tests on my machine and I got the exact same error, so it may not be related with your changes:

image

I did run the tests with `yarn test` and I got that error, but then I started getting timeout errors so I needed to only run primary tests to replicate the issue with `yarn test:primary`.

pcattori:
#6724 introduced the `REMIX_DEV_ORIGIN` option for telling the app server where to find the dev server. Does that address this?

marshallwalker:
@pcattori I don't believe it would no.

pcattori:
If the port is 4004,  you could do `REMIX_DEV_ORIGIN="https://some_url-4004/blabla"`. The `REMIX_DEV_ORIGIN` env var lets you set the _entire_ origin so you have full control of if the `:<port>` is part of it or not.

marshallwalker:
@pcattori I will test it and get back to you :)

pcattori:
Closing as this should have been fixed by #6724. If issue persists, we can reopen.

PR #6707, opened 2023-06-27
Omit search part of path when detecting if the route changed after downloading asset bundles

### LINKED ISSUE #6703: CDN Causing infinite reload

1.18.0

- [X] Yes

### Steps to Reproduce

1. Place your site behind a CDN, and don't include query params in the cache key, but forward them to the origin 
2. Visit a page with a cache header with query param, then again without
3. Page gets stuck in a infinite reload loop

Issue caused by this PR
https://github.com/remix-run/remix/pull/6409

### Expected Behavior

Page should not get stuck in a loop and should load normally 

### Actual Behavior

Page gets stuck in an infinite reload loop as the url on the cache page does not match the URL of the page you are loading

n8agrin:
Resolves #6703

Previous implementation used the pathname and search parts of a url to determine if route changed, which causes some conflict when CDNs / cache proxies are placed between the Remix app and the client. This change allows remix to reload the page if only the path changes. There may be some issues with how loaders are handled when configured behind a CDN, since Remix expects to trigger loaders when the search part of a url changes, but the total impact here is dependent on how a developer is configuring the network topology on top of a Remix server.

Closes: #6703

- [ ] Docs
- [ ] Tests

Testing Strategy:


n8agrin:
@yarbsemaj @brophdawg11 Please see the note in the description. I am concerned about the impact this might have on loaders which expect specific search params to function correctly. I'm also not 100% sure I understand your use case @yarbsemaj.

Please confirm the issue is something like:
* user visits /foo?a=yes
* cdn caches /foo?a=yes
* user visits /foo
* cdn loads cached assets including the serialized /foo?a=yes string
* remix redirects to /foo
* 💥 the cycle repeats itself

I suspect the better solution here might be to change your cdn cache policy to ignore search params in the url.

BillBrower-Shopify:
This ends up causing an infinite reload cycle for urls without query params when I test it as a patch on top of Remix 1.18.0 in our project.

yarbsemaj:
@n8agrin Unfortunately this is not an option in our setup, and i imagine many other as well

brophdawg11:
> Unfortunately this is not an option in our setup

@yarbsemaj This is a bit ambiguous to me.  Is the pathname-only comparion implemented in this PR not an option for your setup?  

Or were you referring to the solution suggested by @n8agrin above:

> I suspect the better solution here might be to change your cdn cache policy to ignore search params in the url.

From the description of your issue in #6703 it sounds like the root cause of your issue is the inclusion of search params in the hydration check, and that only comparing pathnames would suffice?

brophdawg11:
I'm going to re-open this since I do believe this is the right solution to avoid reloads when CDN's ignore certain search parameters.  I confirmed with @BillBrower-Shopify that this fix resolves the issue they are seeing which is due to the CDN only respecting certain search params.  While this is sort of a non-standard HTTP caching approach, I believe it's probably used pretty frequently.  

Since the underlying issue here is a mismatch in matched routes on the server and the client, and route matching only cares about the `pathname`, I believe this is how the original implementation should have been anyway.

n8agrin:
Thanks for weighing in @brophdawg11

packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -230,7 +230,7 @@ describe("remix CLI", () => {
       await interactWithShell(proc, [
         { question: /Where.*create.*app/i, type: [projectDir, ENTER] },
         { question: /What type of app/i, answer: /basics/i },
-        { question: /Where.*deploy/i, answer: /express/i },
+        { question: /Where.*deploy/i, answer: /remix/i },
```
brophdawg11, line 233: Ignore this unit test fix - this was fixed in `dev` and cherry-picked into `main` and `release-next`.  This branch has the `main` SHA since I merged `main` in - but this code is already in `release-next` so it's not actually changing anything.

packages/remix-server-runtime/server.ts
```diff
@@ -335,7 +335,7 @@ async function handleDocumentRequestRR(
       ...entryContext,
       staticHandlerContext: context,
       serverHandoffString: createServerHandoffString({
-        url: context.location.pathname + context.location.search,
+        pathname: context.location.pathname,
```
jacob-ebey: Do we consider this handoff format public API? If so can we not re-name the property here and re-purpose the "url" name?

packages/remix-server-runtime/server.ts
```diff
@@ -335,7 +335,7 @@ async function handleDocumentRequestRR(
       ...entryContext,
       staticHandlerContext: context,
       serverHandoffString: createServerHandoffString({
-        url: context.location.pathname + context.location.search,
+        url: context.location.pathname,
```
brophdawg11, line 338: Keeping the `url` name in this handoff string since it's technically exposed to users via `EntryContext.serverHandoffString` and `window.__remixContext` so just in case anyone is accessing it.

jacob-ebey (approved):

packages/remix-server-runtime/server.ts
brophdawg11: Updated in [9c6cf2f](https://github.com/remix-run/remix/pull/6707/commits/9c6cf2f5a63a0542883d3484d9daffee5e8fc3f2)

yarbsemaj:
@brophdawg11 
> > Unfortunately this is not an option in our setup
> @yarbsemaj This is a bit ambiguous to me. Is the pathname-only comparion implemented in this PR not an option for your setup?

Sorry for the confusion, i was referring to this comment about updating our CDN Policy, this is due to us needing access to a full range of query parameters on certan, arbitrary routes that can not easily be modeled with cache behaviours.

> I suspect the better solution here might be to change your cdn cache policy to ignore search params in the url.

brophdawg11:
@yarbsemaj No worries!  Can you confirm if `1.18.1-pre.0` fixes your issue?

PR #6736, opened 2023-06-30
fix: avoid circular references and infinitely recursive types

pcattori:
Fixes #6701 , #6693

Reverts "pretty" types for `useLoaderData`, `useActionData`, etc...

Pretty types relied on the ability to eagerly evaluate types, but this can cause infinite recursion or errors with circular referenced types.

[...]

PR #6746, opened 2023-07-01
feat: bring back simplified types for SerializeFrom

afoures:
just saw that #6736 removed pretty types for SerializeFrom because of circular dependencies and infinitely recursive types.

this is an attempt to fix the infinitely recursive issue without loosing type simplification.

i tested it using the same [Typescript playground](https://tsplay.dev/weaQgw).

there is one issue for circular types, where we loose the type name: in the playground, the name of `Stripe.InvoiceLineItem` is lost.

this happen when `[T] extends [U]` in `type PrettyTransform<T, U> = [T] extends [U] ? T : Pretty<U>;` is evaluated to false, and that is the case when `T` is a infinite circular type, at one point `U` will have the circular property evaluated to `never` because max depth is reached, so `T` and `U` are not the same anymore. i'm not sure how we could fix it.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -6,6 +6,7 @@
 - achinchen
 - adamwathan
 - adicuco
+- afoures
 - ahabhgk
 - ahbruns
 - ahmedeldessouki
--- a/packages/remix-server-runtime/serialize.ts
+++ b/packages/remix-server-runtime/serialize.ts
@@ -11,6 +11,10 @@ type JsonPrimitive =
   | null;
 type NonJsonPrimitive = undefined | Function | symbol;
 
+// force Typescript to simplify the type
+type Pretty<T> = { [K in keyof T]: T[K] } & {};
+type PrettyTransform<T, U> = [T] extends [U] ? T : Pretty<U>;
+
 /*
  * `any` is the only type that can let you equate `0` with `1`
  * See https://stackoverflow.com/a/49928360/1490091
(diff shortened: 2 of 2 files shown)


pcattori:
There are a couple issues with the implementation:

---

Screenshot 2023-07-03 at 9 19 37 AM

👆 1. `user` should be optional (`user?:`)

---

Screenshot 2023-07-03 at 9 20 40 AM

👆 2. like you already mentioned, inferred types for circularly referenced like `Stripe.InvoiceListItem` do not preserve the original type names

---

Screenshot 2023-07-03 at 9 19 57 AM

👆another example of (2), this time with `Json` circularly referenced type

---

Screenshot 2023-07-03 at 9 20 10 AM

👆 Haven't checked if this is another example of (2) yet or its own issue, but type names for `Zod` are not wholly preserved either

afoures:
You're right, I did not see that `Json` was broken too. Should I try to fix circular type issues?

pcattori:
@afoures closing this for now, but if you come up with a solution for circular types, I'm happy to reopen.

pcattori:
One more thing is that we are trying to move to `type-fest` for JSON serialize/deserialize types rather than maintaining our own. Adding pretty types in the middle of the type recursion means we won't be able to do that anymore. So maybe a better place to land any pretty types would be there, though the same/similar trade-offs apply whether its here or there.

PR #6749, opened 2023-07-01
feat(remix-server-runtime): add `injectServerEffect`

kettanaito:
## Short description

This adds a new function called `injectServerEffect` that helps third-party packages establish process-wide server-side side effects in Remix. In particular, it implements effect persistence, allowing those side effects to survive HMR and HDR (since, by default, they won't and will inevitably get stale). 

This can be a great addition to Remix API for libraries like MSW, Prisma, and quite a few others. I'd like to use this in MSW examples since it yields a much simpler setup for the end developer, which is a win for everyone. I hope you consider merging this. 

## What's the intended usage of this?

```js
import { injectServerEffect } from '@remix-run/node'
import { init } from 'third-party'
import a from './a'

injectServerEffect('myEffect', (a) => {
  const customSideEffect = init(a)
  return () => {
    customSideEffect.anythingToCleanup()
  }
}, [a])
```

Here's an example of using the proposed API with MSW:

```js
// mocks/node.js
import { setupServer } from 'msw'
import { injectServerEffect } from '@remix-run/node'

export async function initMocks(...handlers) {
  await injectServerEffect('msw', (handlers) => {
    // This is the core of the effect: enabling request interception.
    const server = setupServer(...handlers)
    server.listen()

    // Return the instructions to cleanup.
    return () => {
      server.close()
    }
  }, [handlers])
}
```

Closes: https://github.com/remix-run/remix/discussions/6556

- [ ] Docs
- [x] Tests

Testing Strategy:

Testing this will be tricky since the feature is oriented at the custom effects surviving hot updates. I can at least try to provide this pull request with unit test and can count on the team's best judgement on how to properly test it in the integration scenarios.


kettanaito:
Testing this will be tricky since the feature is oriented at the custom effects surviving hot updates. I can at least try to provide this pull request with unit test and can count on the team's best judgement on how to properly test it in the integration scenarios.

kettanaito:
I'm also considering keeping all the effects in a global `Map` instead of letting Remix pollute the root-level global scope.

MichaelDeBoey:
@kettanaito Would you be willing to export it as in the Deno runtime as well + add tests please?

MichaelDeBoey:
@kettanaito Since this is a code change, could you rebase your branch onto latest `dev` & set target to `dev` as well?

kettanaito:
@MichaelDeBoey, I've rebased and set `dev` as the target. Also added some unit tests and exported the function from `/deno`. Happy to hear some feedback on this feature.

packages/remix-server-runtime/__tests__/injectServerEffect-test.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  injectServerEffect,
+  SERVER_EFFECTS_KEY,
+  ServerEffect,
+  getGlobalEffectById,
+} from "../injectServerEffect";
+
+describe(injectServerEffect, () => {
```
MichaelDeBoey, line 8:
```suggestion
describe("injectServerEffect", () => {
```

packages/remix-server-runtime/injectServerEffect.ts
```diff
@@ -0,0 +1,133 @@
...
+  let effect = new ServerEffect(id, callback);
+  await effect.run(...dependencies);
+
+  // Whenever the process is terminated, remove the effect.
+  process.on("SIGTERM", effect.dispose).on("SIGINT", effect.dispose);
+
+  return effect;
+}
+
+export class ServerEffect<Dependencies extends Array<unknown>> {
```
MichaelDeBoey, line 57: Would it be possible to use normal objects + functions instead?
The only classes we currently use in the codebase are for creating custom `Error`s

packages/remix-server-runtime/injectServerEffect.ts
```diff
@@ -0,0 +1,133 @@
...
+) => Promise<ServerCleanupFunction> | ServerCleanupFunction;
+
+export type ServerCleanupFunction = () => Promise<void> | void;
+
+export type ServerEffectsMap = Map<string, ServerEffect<Array<unknown>>>;
+
+export const SERVER_EFFECTS_KEY = Symbol("remixServerEffects");
+
+/**
+ * Creates a server-side effect persisted on tihs Node.js process.
```
herecydev, line 12:
```suggestion
 * Creates a server-side effect persisted on this Node.js process.
```

MichaelDeBoey, line 12:
```suggestion
 * Creates a server-side effect persisted on this Node.js process.
```

MichaelDeBoey (commented):
@kettanaito I tagged some people from the core Remix team so they can have a look at this and discuss further what direction they want to go with this

packages/remix-server-runtime/__tests__/injectServerEffect-test.ts
kettanaito: Can you elaborate more on why repeat the function name when you can infer it by passing the function reference to the `describe` block? 

The code I pushed yields `describe(injectServerEffect.name)`, which turns into your suggestion exactly.

packages/remix-server-runtime/injectServerEffect.ts
kettanaito: I don't mind. My motivation for using a class was to manage that self-updating `this.cleanup` reference. I can use the functional approach although, imo, it's a bit harder to understand than the class.

packages/remix-server-runtime/__tests__/injectServerEffect-test.ts
MichaelDeBoey: My suggestion is how we do it in the rest of the codebase, so I just want to keep it consistent

packages/remix-server-runtime/injectServerEffect.ts
MichaelDeBoey: Let's remove the class for consistency with the rest of the codebase (where we don't use classes)

kettanaito:
I would really love to hear more opinions from the team before I dedicate more time to this feature. I've been talking with @pcattori, and it seems that Remix is not sure if they wish to add such a function officially.

kettanaito:
While this kind of API may still prove useful in some situations, speaking of MSW in particular, the current state of Remix allows to use initiate MSW server-side following regular Node.js integration rules. No extra steps are necessary anymore. I believe the credit goes to the recent reworks in HMR/HDR area by the Remix team. 

- [Remix + MSW example](https://github.com/mswjs/examples/tree/main/examples/with-remix)

PR #6761, opened 2023-07-03
Fix markdown/html nesting

keul:
Fixing nesting of markdown into `summary` section.

This was leading to raw backticks displayed instead


machour (commented):

There are a bunch more of these in the file, and frankly, it doesn't bother me to leave them as is. Think about those as special quotes to delimit the file name for the reader :)

Issue #6748, opened 2023-07-01
Unable to get the no-restart express app server working using CommonJS


tgds:
### What version of Remix are you using?

1.18.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

I've tried the new Express server template with no-restart dev app server set up, which works nicely. I wanted to give it a go with CJS, as that's what I'm using in my project and followed the [docs](https://remix.run/docs/en/main/other-api/dev-v2#keep-app-server-running-across-rebuilds) to set it up. 

1. Create a new project with the basic `Remix App Server` template
2. Create `server.js`:
```ts
const { createRequestHandler } = require("@remix-run/express");
const { broadcastDevReady, installGlobals } = require("@remix-run/node");
const chokidar = require("chokidar");
const express = require("express");
const morgan = require("morgan");

installGlobals();

const BUILD_PATH = "./build/index.js";
/**
 * @type { import('@remix-run/node').ServerBuild | Promise<import('@remix-run/node').ServerBuild> }
 */
let build = require(BUILD_PATH);

const app = express();

// Remix fingerprints its assets so we can cache forever.
app.use(
  "/build",
  express.static("public/build", { immutable: true, maxAge: "1y" })
);

// Everything else (like favicon.ico) is cached for an hour. You may want to be
// more aggressive with this caching.
app.use(express.static("public", { maxAge: "1h" }));

app.use(morgan("tiny"));

app.all("*", createDevRequestHandler());

const port = process.env.PORT || 3000;
app.listen(port, async () => {
  console.log(`Express server listening on port ${port}`);
  broadcastDevReady(build);
});

function createDevRequestHandler() {
  const watcher = chokidar.watch(BUILD_PATH, { ignoreInitial: true });

  watcher.on("all", async () => {
    // 1. purge require cache && load updated server build
    for (const key in require.cache) {
      if (key.startsWith(BUILD_PATH)) {
        console.log("key", key);
        delete require.cache[key];
      }
    }
    build = require(BUILD_PATH);
    // 2. tell dev server that this app server is now ready
    broadcastDevReady(build);
  });

  return async (req, res, next) => {
    try {
      //
      return createRequestHandler({
        build: build,
        mode: "development",
      })(req, res, next);
    } catch (error) {
      next(error);
    }
  };
}
```
3. run `remix dev --no-restart -c 'node ./server.js'`

### Expected Behavior

[...]

PR #6767, opened 2023-07-04
fix(dev): Wait for the router to be initialized in RemixBrowser before applying HMR changes

defjosiah:
Fixes: #6415 

## Description

With the cloudflare-workers template (and likely pages), there is a race condition. Sometimes the `router` that is used in 
the `import.meta` setup in the `RemixBrowser` component file is undefined.

My guess is that the hmr update is sent before the <RemixBrowser /> component is hydrated (which is where the file global 
let router: Router variable is assigned).

This adds a guard to the `import.meta.hot.accept` function, which waits for the router to be initialized before it is used in 
the rest of the function.

## Testing Strategy

I discovered this in devtools, I have a reproducing example locally (same as the issue this closes).

Without my change, you'll see that the router is undefined, and the error:
```
browser.js:42 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'state')
    at browser.js:42:39
    at Object.emit (hmr-runtime:remix:hmr:64:9)
    at ws.onmessage (login:62:82)
```
is shown. There is also no HMR that works.

Screenshot 2023-07-04 at 4 07 00 PM

Screenshot 2023-07-04 at 4 06 41 PM

With my change, built and applied locally, it properly does HMR and has no errors:

Screenshot 2023-07-04 at 4 04 42 PM

I looked for places to add tests for this, but I don't see any component level tests for the <RemixBrowser /> component, and it looked like a large can of worms to attempt to add the first tests for this component. The hmr integration test will catch if this is an issue with the existing hmr setup.

## Reproduction Steps
1. Start with the cloudflare-workers remix template
2. Try HMR, it may or may not work, if it doesn't work, it'll show that error (or `Uncaught (in promise) TypeError: Failed to fetch dynamically imported module: http://localhost:8787/build/entry.client-EGKEKBAB.js`)


defjosiah:
I'm also running with `--no-restart`, so there's a chance it also fixes:

https://github.com/remix-run/remix/issues/6748

mauricevancooten:
I'm using the cloudflare-workers template and can confirm I'm getting this error too. @defjosiah I'm looking forward to your fix!

sergiodxa (approved):

jacob-ebey:
@defjosiah I have a commit over here that I'd like to see added as part of this fix: https://github.com/remix-run/remix/pull/6774

defjosiah:
@jacob-ebey: sounds good!
I can pull that in here (You can also feel free to push to my branch, but maybe the fork confuses that).

How do you want me to handle that?
I can also just accept your PR

jacob-ebey:
> @jacob-ebey: sounds good! I can pull that in here (You can also feel free to push to my branch, but maybe the fork confuses that).
> 
> How do you want me to handle that? I can also just accept your PR

Feel free to pull it into your fork / this PR if you'd like to do that!

defjosiah:
@jacob-ebey: done! Thanks for the extra error handling

jacob-ebey (approved):

jacob-ebey:
> Do I have to do anything else here, or will y'all handle merging it, etc.?

I'll take care of it. Thanks for the contribution! 🎉

packages/remix-react/browser.tsx
```diff
@@ -180,7 +202,7 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
           window.__remixContext.future.v2_normalizeFormMethod,
       },
     });
-
+
```
pcattori, line 205: @jacob-ebey this looks like unintentional whitespace

packages/remix-react/browser.tsx
```diff
@@ -48,6 +48,19 @@ declare global {
...
+// HMR handler until the router is ready
+let hmrRouterReadyPromise = new Promise<Router>((resolve) => {
+  // body of a promise is executed immediately, so this can be resolved outside
+  // of the promise body
+  hmrRouterReadyResolve = resolve;
+}).catch(() => {
+  // This is a noop catch handler to avoid unhandled promise rejection warnings
+  // in the console. The promise is never rejected.
+  return undefined;
+});
```
pcattori, line 63: You could use a `Channel` here instead so you don't have to manually set `resolve` and check for it later to make TS happy:

```suggestion
let routerReady = Channel.create<Router>()
// ...
let checkRouter = await routerReady.result()
if (!checkRouter.ok) {
  console.error("...")
  return
}
let router = checkRouter.value
// ...

routerReady.ok(router)
```

defjosiah: That's fine with me, however, it looks like that Channel implementation is in remix-dev package?
https://github.com/remix-run/remix/blob/0d22ee92450eb6caf78055d3d9d956b244de6519/packages/remix-dev/channel.ts#L28

And it doesn't look like there's an import in this react package, so not sure how y'all would want to do that.

Also, typescript seems happy with it for me! I'm getting all the proper types out of this.

pcattori: Oh yea good point. Not worth introducing the channel abstraction just for this if its not already in this package.

defjosiah: Ahh, got it! 
Yeah, I really like that Channel abstraction, probably going to shamelessly use it somewhere else 😂

defjosiah:
Hi, should I be keeping this up to date?
Wasn't sure what the merging timeline is for a contribution.

pcattori:
@defjosiah I'll review this with @jacob-ebey on Monday

PR #6772, opened 2023-07-05
feat(dev): output esbuild metafiles for bundle analysis

pcattori:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/compiler/analysis.ts
+++ b/packages/remix-dev/compiler/analysis.ts
@@ -1,0 +1,17 @@
+import fs from "node:fs";
+import path from "node:path";
+import type { Metafile } from "esbuild";
+
+import type { Context } from "./context";
+
+export let writeMetafile = (
+  ctx: Context,
+  filename: string,
+  metafile: Metafile
+) => {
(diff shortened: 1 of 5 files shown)


packages/remix-dev/compiler/analysis.ts
```diff
@@ -0,0 +1,20 @@
...
+  filename: string,
+  metafile: Metafile
+) => {
+  let buildDir = path.dirname(ctx.config.serverBuildPath);
+  if (!fs.existsSync(buildDir)) {
+    fs.mkdirSync(buildDir, { recursive: true });
+  }
+  fs.writeFileSync(
+    path.join(buildDir, filename),
+    JSON.stringify(metafile, undefined, 2)
```
jacob-ebey, line 18: Do we care if the file is pretty?

pcattori: probably not actually. I can remove the `, undefined, 2` part

jacob-ebey (approved):

manafire:
Does this require a special config option or command or something?  I've tried it in a couple of projects and it's not generating the metafiles.

PR #6783, opened 2023-07-06
Properly handle in-flight network errors on _data requests

### LINKED ISSUE #5418: Unexpected HTTP errors on _data calls cause blank routes without any caught errors

1.12.0

- [X] Yes

### Steps to Reproduce

Force a route's _data request to fail with an HTTP error and a response that remix doesn't expect (e.g. some string). This could happen when there is something wrong with e.g. CDN, proxy or load balancer configuration, an error that is generated outside or remix's control. Then make a client side navigation to that route. 
Here's a diff from base create-remix app with express server where I respond with an error from express handler: https://github.com/celebro/remix-http-error-example/commit/3e644df76e4e673a86cbdaecae4e138386ccb682

### Expected Behavior

Remix should trigger ErrorBoundary with some generic error.

### Actual Behavior

Remix renders an empty route, without any indication of an error.

brophdawg11:
If we make a `?_data` `fetch` and it never actually reaches the Remix server, and instead we get a 4xx/5xx response back from a CDN/Proxy/Load Balancer, etc., then we currently incorrectly think it's a returned response from a `loader`/`action` since it's perfectly valid to:

```js
export function loader() {
  return json({ isTeapot: true }, { status: 418 });
}
```

This PR adds an `X-Remix-Response: yes` header to successful responses for `?_data` requests so we can differentiate responses that reached the Remix server from those that did not - and we can treat CDN/Proxy errors as real errors and surface them to the `ErrorBoundary`.

Closes #5418

### DIFF
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -166,10 +166,16 @@ async function handleDataRequestRR(
       let init = deferredData.init || {};
       let headers = new Headers(init.headers);
       headers.set("Content-Type", "text/remix-deferred");
+      // Mark successful responses with a header so we can identify in-flight
+      // network errors that are missing this header
+      headers.set("X-Remix-Response", "yes");
       init.headers = headers;
       return new Response(body, init);
     }
 
+    // Mark all successful responses with a header so we can identify in-flight
+    // network errors that are missing this header
+    response.headers.set("X-Remix-Response", "yes");
     return response;
   } catch (error: unknown) {
     if (isResponse(error)) {
--- a/.changeset/handle-network-errors.md
+++ b/.changeset/handle-network-errors.md
@@ -1,0 +1,6 @@
+---
+"@remix-run/react": patch
+"@remix-run/server-runtime": patch
+---
+
+Properly handle `?_data` HTTP/Network errors that don't reach the Remix server and ensure they bubble to the `ErrorBoundary`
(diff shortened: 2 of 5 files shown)


packages/remix-server-runtime/server.ts
```diff
@@ -166,10 +166,16 @@ async function handleDataRequestRR(
...
+      // Mark successful responses with a header so we can identify in-flight
+      // network errors that are missing this header
+      headers.set("X-Remix-Response", "yes");
       init.headers = headers;
       return new Response(body, init);
     }
 
+    // Mark all successful responses with a header so we can identify in-flight
+    // network errors that are missing this header
+    response.headers.set("X-Remix-Response", "yes");
```
brophdawg11, line 178: Instead of having to remember to do this in each success code path,. should we instead do this higher up?  I kept it here to colocate it with the setting of the error/catch headers but it feels potentially error-prone in the future:

```js
function requestHandler() {
  ...
  if (url.searchParams.has("_data")) {
    response = await handleDataRequestRR(...);

    if (!response.headers.has('X-Remix-Error') && !response.headers.has('X-Remix-Catch')) {
      response.headers.set('X-Remix-Response', 'yes');
    }
    ...
  }
  ...
}
```

brophdawg11:
@jacob-ebey raised a good question around whether a CDN would cache this header or not.  

I think it probably depends on the CDN - I would hope CDN's aren't caching 4xx/5xx responses.  But technically, yeah if a if a CDN cached a user-returned 4xx/5xx response and _did not_ cache the `X-Remix-Response` header along with it - then this change would surface that served cached response to the `ErrorBoundary` 😕

brophdawg11:
Spoke with @mjackson and we agreed that CDN's (1) should not generally be caching 4xx/5xx responses, but (2) if they are configured to - they should be caching the headers along with it, so we do not expect the above scenario to be an issue.

jacob-ebey (approved):

PR #6807, opened 2023-07-10
fix(remix-react): fix `fetcher.formEncType`

AlemTuzlak:
I believe that there is a typing issue with fetchers where the encType is not typed correctly, I noticed this when using useNavigate where the hook was typed but the fetcher wasn't so I tracked it down and fixed it.
 
Testing Strategy:
- Import `useFetcher` or `useFetchers`
- The type of formEncType should be `FormEncType | undefined` instead of `string | undefined`


brophdawg11:
Thank you!  Since this contains code changes, would you mind repointing to `dev` and adding a patch changeset via `yarn changeset`?

AlemTuzlak:
> Thank you!  Since this contains code changes, would you mind repointing to `dev` and adding a patch changeset via `yarn changeset`?

@brophdawg11 will do!

AlemTuzlak:
closing in favor of #6810

PR #6810, opened 2023-07-11
fix(remix-react): fix `fetcher.formEncType`

AlemTuzlak:
I believe that there is a typing issue with fetchers where the encType is not typed correctly, I noticed this when using useNavigate where the hook was typed but the fetcher wasn't so I tracked it down and fixed it.

Testing Strategy:

Import useFetcher or useFetchers
The type of formEncType should be FormEncType | undefined instead of string | undefined

### DIFF
--- a/.changeset/chatty-pears-float.md
+++ b/.changeset/chatty-pears-float.md
@@ -2,4 +2,4 @@
 "@remix-run/react": patch
 ---
 
-Changed the type of formEncType on useFetcher and useFetchers to be FormEncType from react-router-dom instead of strings
+Narrowed the type of `fetcher.formEncType` to use `FormEncType` from `react-router-dom` instead of `string`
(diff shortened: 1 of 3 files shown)


AlemTuzlak:
@brophdawg11 hopfully I did it right this time, last PR was from the main branch and I couldn't change the origin but only the target

MichaelDeBoey (approved):

.changeset/chatty-pears-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Changed the type of formEncType on useFetcher and useFetchers to be FormEncType from react-router-dom instead of strings
```
brophdawg11, line 5:
```suggestion
Narrowed the type of `fetcher.formEncType` to use `FormEncType` from `react-router-dom` instead of `string`
```

brophdawg11 (approved):

brophdawg11:
Thanks!

PR #6818, opened 2023-07-12
chore(dev): warn if not using `v2_dev`

pcattori:
## TODO

- [x] `--manual` + `v2_dev.manual` (undoc/warn for `--no-restart` and `v2_dev.restart`)
- [x] warn if not using `v2_dev`
- [x] changeset

## Follow-on work
- [ ] `remix-serve` compat with CJS `--manual`?
- [ ] update templates
- [ ] update tutorials

### DIFF
--- a/.changeset/hot-ducks-march.md
+++ b/.changeset/hot-ducks-march.md
@@ -1,0 +1,8 @@
+---
+"@remix-run/dev": patch
+---
+
+Warn if not using v2_dev
+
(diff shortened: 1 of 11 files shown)


mjackson:
I don't think we need to undoc or warn if they use `--no-restart`. It's just an alias for the same thing, and could help some people understand what it does from a different angle. I like the fact that we have 2 flags for it.

PR #6820, opened 2023-07-12
Fix bug with ?_data request on non-matching route

brophdawg11:
Closes https://github.com/remix-run/remix/pull/5738

### DIFF
--- a/packages/remix-server-runtime/server.ts
+++ b/packages/remix-server-runtime/server.ts
@@ -82,10 +82,9 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
       );
 
       if (build.entry.module.handleDataRequest) {
-        let match = matches!.find((match) => match.route.id == routeId)!;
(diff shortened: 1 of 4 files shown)


packages/remix-server-runtime/server.ts
```diff
@@ -82,10 +82,9 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
       );
 
       if (build.entry.module.handleDataRequest) {
-        let match = matches!.find((match) => match.route.id == routeId)!;
         response = await build.entry.module.handleDataRequest(response, {
           context: loadContext,
-          params: match ? match.params : {},
+          params: matches?.find((m) => m.route.id == routeId)?.params || {},
```
brophdawg11, line 87: When `handleDataRequest` was defined and a request came in like `/not-a-route?_data=routes/not-a-route` we got a valid 404 from the static handler but then errored out here trying to get the params when no matches exist.

jacob-ebey (approved):

PR #6867, opened 2023-07-18
feat(remix-eslint-config)!: bump typescript-eslint to 6.1.0 and extend stylistic preset

JoshuaKGoldberg:
Closes: #6866

- [ ] Docs - same as before, happy to document if directed to
- [x] Tests - `yarn lint` still passes

Testing Strategy: I'd encourage folks to try this out on any large Remix repository written with TypeScript. And let me know what those are so I can test them out too? 😄

Otherwise, `yarn lint`.

Enables https://typescript-eslint.io/linting/configs#stylistic in the `typescript.js` preset, with any rules that cause a lot of reports internally disabled. 

Any code changes are commented inline.


packages/remix-node/upload/fileUploadHandler.ts
```diff
@@ -228,9 +228,7 @@ export class NodeOnDiskFile implements File {
     return readableStreamToString(this.stream());
   }
 
-  public get [Symbol.toStringTag]() {
-    return "File";
-  }
+  public readonly [Symbol.toStringTag] = "File";
```
JoshuaKGoldberg, line 231: https://typescript-eslint.io/rules/class-literal-property-style: enforces a consistent style for declaring class literals. The default is to prefer a `public readonly` over a `public get` because it's less code. 

Note that the rule provides [_suggestions_ instead _fixes_](https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions) (i.e. its suggestions can only be done manually, not by `--fix`) because they technically change code behavior. But this seems to be equivalent to me, and I couldn't find any issues?

packages/remix-eslint-config/rules/typescript.js
```diff
@@ -1,18 +1,23 @@
...
   "@typescript-eslint/no-inferrable-types": "off",
   "@typescript-eslint/no-namespace": "off",
   "@typescript-eslint/no-non-null-assertion": "off",
   "@typescript-eslint/no-var-requires": "off",
+  "@typescript-eslint/prefer-function-type": "off",
   "no-var": "off",
   "prefer-rest-params": "off",
 
   // These rules are nice and we want to configure over the defaults
+  "@typescript-eslint/consistent-type-imports": "error",
```
JoshuaKGoldberg, line 20: We ended up not putting https://typescript-eslint.io/rules/consistent-type-imports actually in the _`stylistic`_ preset. The TS world hasn't solidified on one preferred style.

brophdawg11:
Hey @JoshuaKGoldberg thanks for putting this together - but we're leaning towards getting out of the eslint/stylistic opinions game at the Remix framework level.  It's none of our busines show app developers want to format their Remix code 😄 .  Our Remix Stacks are really where opinions such as this come into play so we're going to be devising a plan shortly once we release v2 to figure out a path to deprecating `@remix-run/eslint-config` and instead making the sets of "recommended" rules part of the Remix stacks.  At that point we'd definitely be open to including the new stylistic rules in the stacks!

JoshuaKGoldberg:
Makes sense! I agree with your reasoning 🙂. And also have been burned recently by enabling too much of the stylistic rules in another preset: https://github.com/t3-oss/create-t3-app/issues/1520. Happy to hop back in with the stacks if you'd find that helpful!

Just to be clear / for public visibility, is extending from [`recommended-type-checked`](https://typescript-eslint.io/linting/configs#recommended-type-checked) (ref: 36248) still something you'd want?

brophdawg11:
In the stacks, yeah we'll pick one of the base configurations.  Our goal is to provide a _very_ minimal working configuration that simply extends a few opinionated best practice configs (react/recommended, typescript-eslint/recommended-type-checked, etc.) but does little to nothing more.  Then app devs can begin tweaking the base setups to their liking after they create their app from the stack.

brophdawg11:
As for internal Remix framework code linting - I'm not sure what we'll do there just yet...

PR #6885, opened 2023-07-19
Fix typo in manifest regex

lpsinger:
The character class `[A-f0-9]` contains a typo: the manifest versions are upper-case hex strings.

This fixes the following CodeQL warning:
https://codeql.github.com/codeql-query-help/javascript/js-overly-large-range/


pcattori:
I think the `A-f` range is intentional as lowercased hex strings are present in manifest versions:

Screenshot 2023-07-20 at 12 18 33 PM

Plus the integration tests that look for those patterns are now timing out with the proposed changes.

lpsinger:
> I think the `A-f` range is intentional as lowercased hex strings are present in manifest versions:

How very odd. In my Remix 1.19.0 architect project, the filenames have uppercase manifest versions. What hosting target are you using?

packages/remix-dev/devServer_unstable/index.ts
```diff
@@ -136,7 +136,7 @@ export let serve = async (
             transform(chunk, _, callback) {
               let str: string = chunk.toString();
               let matches =
-                str && str.matchAll(/\[REMIX DEV\] ([A-f0-9]+) ready/g);
+                str && str.matchAll(/\[REMIX DEV\] ([A-F0-9]+) ready/g);
```
pcattori, line 139:
```suggestion
                str && str.matchAll(/\[REMIX DEV\] ([A-Fa-f0-9]+) ready/g);
```

pcattori (changes requested):

templates/arc/plugin-remix.js
```diff
@@ -23,7 +23,8 @@ export default {
           path.resolve(process.cwd(), buildPath),
           "utf8"
         );
-        const manifestMatches = contents.matchAll(/manifest-([A-f0-9]+)\.js/g);
+        // FIXME: https://github.com/remix-run/remix/pull/6885
+        const manifestMatches = contents.matchAll(/manifest-([A-F0-9]+)\.js/g);
```
pcattori, line 27:
```suggestion
        const manifestMatches = contents.matchAll(/manifest-([A-Fa-f0-9]+)\.js/g);
```

pcattori (changes requested):

pcattori (approved):

lpsinger:
@pcattori, did we ever find a reason why in my testing the fingerprints were always uppercase but on yours they were lowercase or mixed case? I just have a lingering worry that there might be another bug there, or at least an inconsistency between different hosting targets.

pcattori:
Could be an OS-related thing (case-sensitive vs case-insensitive filesystems?), but not sure. If it comes up, we can create a separate issue.

lpsinger:
> @lpsinger I've tracked this down! Looks like `plugin-remix.js` is grabbing the build hash from `public/build/manifest-<HASH>.js` filename, whereas the rest of the code is looking at the _contents_ of that file, specifically the `assets.version` field.
> 
> Turns out that [`esbuild` uppercases the build hash when including it as part of a filename](https://github.com/evanw/esbuild/issues/1529). By lowercasing that hash before logging it via `logDevReady`, Remix HMR are flowing correctly.

Wouldn't it be more consistent to have `plugin-remix.js` check the file contents?

pcattori:
> Wouldn't it be more consistent to have `plugin-remix.js` check the file contents?

💯 agreed. I already have the change locally but haven't pushed it up as a PR since I've got a couple other fixes I was investigating.

PR #6862, opened 2023-07-18
chore(remix-dev): update `proxy-agent` dependency

roughee:
Fixes #6833

PR #6887, opened 2023-07-20
feat(create-remix): rewrite create-remix CLI

markdalgleish:
This PR makes the following changes to create-remix.

- Moves the logic for `create-remix` out of `@remix-run/dev`. The Remix CLI no longer has a `remix create` command (which was kind of odd in the context of an existing app). This means `create-remix` is now much smaller since it's not downloading the entire Remix CLI and its dependencies up-front.
- Removes the template prompt. This puts official templates and community/third-party templates/stacks on a more equal footing. You need to provide the `--template` option for anything other than the basic template.
- Adds support for directories in GitHub repo shorthands, e.g. `--template remix-run/remix/templates/express`
- Replaces the "Stacks" documentation with more general "Templates" docs
- Displays a link to the new "Templates" docs when the `--template` option isn't provided
- Adds a prompt to automatically run `git init`.
- Adds a prompt to run the `remix.init` script if it's detected. This makes it clear when we're running code from the template/stack rather than `create-remix` itself. This means you can explicitly opt-out too which previously wasn't possible. Instead, you could only implicitly opt out by choosing not to install dependencies.
- Removes support for `index.ts` in `remix.init` scripts to simplify things and remove the esbuild dependency from `create-remix`.
- Removes the option to automatically convert TypeScript to JavaScript.
- Makes the UI a bit nicer.
- Updates all docs to reflect these changes.

Note: This PR replaces #6391 (which was already approved), targeting `dev` instead of the discarded `v2` branch.


packages/create-remix/package.json
```diff
@@ -17,7 +17,28 @@
...
   },
   "dependencies": {
-    "@remix-run/dev": "1.19.0"
+    "arg": "^5.0.1",
+    "chalk": "^4.1.2",
+    "execa": "5.1.1",
+    "gunzip-maybe": "^1.4.2",
+    "log-update": "^5.0.1",
+    "node-fetch": "^2.6.9",
+    "proxy-agent": "^5.0.0",
```
brophdawg11, line 26: Can we update this to `6.3.0` while we're making this overhaul to resolve https://github.com/remix-run/remix/pull/6862?  I think `proxy-agent` can also be removed from remix/dev package.json since (I think) it was only used in `cli/create.ts` which is now deleted.

PR #6911, opened 2023-07-21
feat(dev): remove default server Node polyfills

markdalgleish:
In Remix v2, all polyfills for Node.js built-in modules are now opt-in, regardless of server platform.


package.json
```diff
@@ -110,6 +110,7 @@
     "jest-watch-typeahead": "^0.6.5",
     "jsonfile": "^6.0.1",
     "lodash": "^4.17.21",
+    "mime": "^3.0.0",
```
markdalgleish, line 113: This change was needed so that the tests are running the same version of `mime` as `@remix-run/deno`. Without this change, the Deno tests ended up using a different version of `mime` (hoisted from some of our dependencies) which requires Node's `path` and `fs` modules. Since we're no longer providing Node.js polyfills by default, this breaks the build.

packages/remix-dev/config.ts
```diff
@@ -444,75 +444,7 @@ export async function readConfig(
...
-        "util/types": true, // https://github.com/jspm/jspm-core/blob/main/nodelibs/browser/util/types.js
-        v8: false, // Unimplemented, throws on usage: https://github.com/jspm/jspm-core/blob/main/nodelibs/browser/v8.js
-        vm: true, // https://github.com/jspm/jspm-core/blob/main/nodelibs/browser/vm.js
-        wasi: true, // https://github.com/jspm/jspm-core/blob/main/nodelibs/browser/wasi.js
-        worker_threads: true, // https://github.com/jspm/jspm-core/blob/main/nodelibs/browser/worker_threads.js
-        zlib: true, // https://github.com/jspm/jspm-core/blob/main/nodelibs/browser/zlib.js
-      },
-    };
-  }
+  let serverNodeBuiltinsPolyfill = appConfig.serverNodeBuiltinsPolyfill;
```
brophdawg11, line 447: 😍

.changeset/remove-default-server-node-polyfills.md
```diff
@@ -0,0 +1,69 @@
...
+```js
+exports.serverNodeBuiltinsPolyfill = {
+  modules: {
+    path: true, // Provide a JSPM polyfill
+    fs: "empty", // Provide an empty polyfill
+  },
+};
+```
+
+For reference, the complete set of default polyfills from Remix v1 can be manually specified as follows:
```
brophdawg11, line 18: This feels potentially big for release notes - is there any benefit to putting this in the docs - maybe collapsed in a summary/details section?

brophdawg11 (approved):

.changeset/remove-default-server-node-polyfills.md
markdalgleish: It's in the v2 guide so we could just leave it there for now. I'll drop this from the release notes.

PR #6916, opened 2023-07-21
fix(dev): remove outdated esm import warnings

pcattori:
Supercedes #6822


jacob-ebey (approved):

kentcdodds:
Did this PR not make it into 1.19.2? I'm still seeing warnings with 1.19.2.

pcattori:
@kentcdodds should have landed in 1.19.1 or 1.19.2

We removed Remix-specified warnings that produced false positives, but you might still get _actual_ warnings from `esbuild`. What do the warnings you are seeing look like?

ngbrown:
This change didn't land in `1.19.1` or `1.19.2`. I'm not sure why the bot thought it did.

The errors shown are still the same:

```
> remix build

 info  building... (NODE_ENV=production)
 warn  esm-only package: quick-lru
┃ quick-lru is possibly an ESM-only package.
┃ To bundle it with your server, include it in `serverDependenciesToBundle`
┃ -> https://remix.run/docs/en/main/file-conventions/remix-config#serverdependenciestobundle
┗
```

kentcdodds:
Yeah, I looked in node_modules and the code that was removed in this PR is definitely still there. 🤷‍♂️

kentcdodds:
I just upgraded to 1.19.3 and this is *still* in there: https://unpkg.com/browse/@remix-run/dev@1.19.3/dist/compiler/server/plugins/bareImports.js (line 137)

PR #6926, opened 2023-07-21
fix(dev): websocket heartbeat

pcattori:
Applies changes in #6904 to both dev servers (old and new) and targets `dev` branch

### DIFF
--- a/.changeset/brave-apples-agree.md
+++ b/.changeset/brave-apples-agree.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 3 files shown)


brophdawg11:
We don't need to PR this to `dev`, it'll get there when we merge `release-hotfix` in.  But do we need to make the v2 changes (`packages/remix-dev/devServer_unstable/socket.ts`) in `release-next`?

brophdawg11:
nvm - Jason has a PR to main that has it - https://github.com/remix-run/remix/pull/6927.  I'm going to close this and both updates will make it to dev with the release

PR #6937, opened 2023-07-22
feat(remix-netlify): show deprecation warning on usage

MichaelDeBoey:
Just like we did with the Vercel adapter

We should probably release this as `1.19.2` so people still get this warning before v2 🤔


brophdawg11:
@MichaelDeBoey Agreed - let's get this rebased onto `main` so we can include in a `1.19.2`

MichaelDeBoey:
@brophdawg11 Should be fine now

PR #6939, opened 2023-07-22
feat!: require Node >=18.0.0

MichaelDeBoey:
Due to this, we're also allowed to update major versions of dependencies (which was not always possible before this PR)


packages/create-remix/package.json
```diff
@@ -17,31 +17,10 @@
...
-  },
-  "devDependencies": {
-    "@types/gunzip-maybe": "^1.4.0",
-    "@types/node-fetch": "^2.5.7",
-    "@types/tar-fs": "^2.0.1",
-    "fs-extra": "^10.0.0",
-    "msw": "^0.39.2",
-    "tiny-invariant": "^1.2.0",
-    "esbuild-register": "^3.3.2"
+    "@remix-run/dev": "1.19.1"
```
ngbrown, line 20: Is replacing all of the dependencies intended? The build seems to be failing because of it.

MichaelDeBoey: I must have messed up a rebase 🙈😅

markdalgleish (approved):

Issue #6891, opened 2023-07-20
v1.19.0 removes `build.dev`, breaking `logDevReady()` in Cloudflare


huw:
### What version of Remix are you using?

v1.19.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

(Firstly @pcattori I f**ing love the new dev server and feel like this is as great a place as any to show my appreciation. It’s hard work to keep this compatible with all the different runtimes and other frameworks have just given up on trying—I love that you haven’t!)

This one is fairly simple, I think it’s just an oversight. The Cloudflare template includes:

https://github.com/remix-run/remix/blob/14bdd98c6872762f1c54ae55f852ae2e4c9461e0/templates/cloudflare-workers/server.ts#L10-L12

But as of v1.19.0, `build.dev` no longer exists ([blame](https://github.com/remix-run/remix/commit/d7c026d2e8bcfc6b4dd10a3d9c14d5af643c1354#diff-4c05589aa34d1c57a18782e2120c5a2842258d48fd0bbe90932205aba5e2d559)). As such, my live reloads never trigger and I have a type error when I compile.

In the meantime, I’m gating on `process.env.NODE_ENV`, which I’ve been injecting into my Workers environment anyway.

### Expected Behavior

The CLI should print the build hash to the console, triggering a live reload.

### Actual Behavior

The CLI does not print the build hash to the console and live reloads don’t happen.

huw:
@pcattori I was slightly mistaken; since `server.ts` is part of the Remix build it will automatically have `process.env.NODE_ENV` injected. As such, I’ve submitted #6912 to use that by default in the template.

Ponjimon:
What about `process` not being defined and thus causing TypeScript to error in the IDE?

```
Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.ts(2580)
```

<img width="705" alt="Screenshot 2023-07-23 at 18 23 15" src="https://github.com/remix-run/remix/assets/770632/85c1128d-7220-4ca6-bb92-c4e8e44c1691">

It still runs,  but my IDE shows me this error and I'm not sure if installing `@types/node` is the best way to fix this

pcattori:
@Ponjimon you can force types for `process.env` like so:

```ts
declare var process : {
  env: {
    NODE_ENV: string
  }
}
```

That way you don't need to rely on `@types/node` for non-Node runtimes.
[...]

PR #6964, opened 2023-07-26
feat(dev,server-runtime): detect built mode via `build.mode`

pcattori:
Relates to #6891


packages/remix-dev/__tests__/fixtures/cloudflare/server.ts
```diff
@@ -4,5 +4,5 @@ import * as build from "@remix-run/dev/server-build";
 export const onRequest = createPagesFunctionHandler({
   build,
   getLoadContext: (context) => context.env,
-  mode: process.env.NODE_ENV,
+  mode: build.mode,
```
MichaelDeBoey, line 7: Do we want to simplify `createPagesFunctionHandler` so we don't have to pass `mode` anymore and it gets automatically recognized?

packages/remix-dev/__tests__/fixtures/deno/server.ts
```diff
@@ -5,7 +5,7 @@ import * as build from "@remix-run/dev/server-build";
 
 const remixHandler = createRequestHandlerWithStaticFiles({
   build,
-  mode: Deno.env.get("NODE_ENV"),
+  mode: build.mode,
```
MichaelDeBoey, line 8: Same for `createRequestHandlerWithStaticFiles`?

packages/remix-dev/__tests__/fixtures/cloudflare/server.ts
pcattori: Conceptually, yes, but holding off on that since we are discussing better ways to configure dev-only options than to just lump them all under `NODE_ENV=development`

PR #6991, opened 2023-07-30
docs(pages/gotchas): add 'Minimizing Bundle Size' section

Brocktho:
I created a [discussion](https://github.com/remix-run/remix/discussions/6990#esbuild-bundle) around minimizing build time and bundle size, wanted to add this to gotchas for the Remix docs since it can be easy to include very large packages or polyfills by accident. I tagged along at the end of the gotcha page a section on tree shaking larger packages, and ensuring server only files aren't having polyfills added to your client bundle mistakenly.

- [x] Docs

Please let me know any revisions or what else I should do to help get this through as I believe it could be really useful to developers both as they start to develop with Remix and as their application becomes larger and build times increase.

### DIFF
--- a/docs/pages/gotchas.md
+++ b/docs/pages/gotchas.md
@@ -184,3 +184,58 @@ Note that, even if this issue didn't exist, we'd still recommend using named re-
 [remix-upload-handlers-like-unstable-create-file-upload-handler-and-unstable-create-memory-upload-handler]: ../utils/parse-multipart-form-data#uploadhandler
 [css-bundling]: ../guides/styling#css-bundling
 [esbuild-css-tree-shaking-issue]: https://github.com/evanw/esbuild/issues/1370
+
+
+## Minimizing Bundle Size
+
+As you continue to develop you may start including larger packages and notice an impact on your build time. There are a few things you can do to minimize your build times that'll also end up being great for your end user (less javascript, faster load times!). Often times larger packages will centralize a location to easily import components or hooks into your app, this is great for convenience but can make it difficult for the compiler to tree shake out code that isn't used.
+
+Take for example MUI:
+
+```tsx bad
+import { Button, TextField } from '@mui/material';
+```
+
+Here Button and TextField are imported from the top-level index of the MUI package, this won't have a massive effect on users but can create more work for the compiler giving you longer build times. Instead you can do the following:
+
+```tsx good
+import Button from "@mui/material/Button";
+import TextField from "@mui/material/TextField";
+```
+
+This lessens the work the compiler has to do and guarantees a smaller overall bundle size.
+
+Sometimes you may also import packages that you intend for server use only. Sometimes this will cause immediate errors, but other times the compiler may try adding polyfills that increase your overall bundle size. Take for example core node modules like node:crypto.
+
+```tsx bad
+import crypto from "node:crypto";
+
+import { json } from "@remix-run/node"; 
+
+export async function loader() {
+  const uuid = crypto.randomUUID();
+  return json({uuid});
+}
+```
+
+Importing crypto this way could lead to the bundler including polyfills for crypto into your client build even if it wasn't directly used on the client. An easy way to get around this is to centralize server imports inside a .server file. For example:
+
+```tsx filename=app/utils/Node.server.ts
+import crypto from "node:crypto";
+import fs from "node:fs";
+
+export { fs, crypto };
+```
+
+```tsx good filename=app/routes/some/route.tsx
+import { json } from "@remix-run/node";
+
+import { crypto } from "~/utils/Node.server";
+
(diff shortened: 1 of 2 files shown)


docs/pages/gotchas.md
```diff
@@ -184,3 +184,58 @@ Note that, even if this issue didn't exist, we'd still recommend using named re-
 [remix-upload-handlers-like-unstable-create-file-upload-handler-and-unstable-create-memory-upload-handler]: ../utils/parse-multipart-form-data#uploadhandler
 [css-bundling]: ../guides/styling#css-bundling
 [esbuild-css-tree-shaking-issue]: https://github.com/evanw/esbuild/issues/1370
+
+
+## Minimizing Bundle Size
+
+As you continue to develop you may start including larger packages and notice an impact on your build time. There are a few things you can do to minimize your build times that'll also end up being great for your end user (less javascript, faster load times!). Often times larger packages will centralize a location to easily import components or hooks into your app, this is great for convenience but can make it difficult for the compiler to tree shake out code that isn't used.
```
MichaelDeBoey, line 191:
```suggestion
As you continue to develop you may start including larger packages and notice an impact on your build time. There are a few things you can do to minimize your build times that'll also end up being great for your end user (less JavaScript, faster load times!). Often times larger packages will centralize a location to easily import components or hooks into your app. This is great for convenience, but can make it difficult for the compiler to tree shake out code that isn't used.
```

docs/pages/gotchas.md
```diff
@@ -184,3 +184,58 @@ Note that, even if this issue didn't exist, we'd still recommend using named re-
...
+
+Take for example MUI:
+
+```tsx bad
+import { Button, TextField } from '@mui/material';
+```
+
+Here Button and TextField are imported from the top-level index of the MUI package, this won't have a massive effect on users but can create more work for the compiler giving you longer build times. Instead you can do the following:
```
MichaelDeBoey, line 199:
```suggestion
Here `Button` and `TextField` are imported from the top-level index of the MUI package, this won't have a massive effect on users but can create more work for the compiler giving you longer build times. Instead you can do the following:
```

docs/pages/gotchas.md
```diff
@@ -184,3 +184,58 @@ Note that, even if this issue didn't exist, we'd still recommend using named re-
...
+
+```tsx good
+import Button from "@mui/material/Button";
+import TextField from "@mui/material/TextField";
```
MichaelDeBoey, line 203:
```suggestion
import { Button } from "@mui/material/Button";
import { TextField } from "@mui/material/TextField";
```

docs/pages/gotchas.md
```diff
@@ -184,3 +184,58 @@ Note that, even if this issue didn't exist, we'd still recommend using named re-
...
+```
+
+This lessens the work the compiler has to do and guarantees a smaller overall bundle size.
+
+Sometimes you may also import packages that you intend for server use only. Sometimes this will cause immediate errors, but other times the compiler may try adding polyfills that increase your overall bundle size. Take for example core node modules like node:crypto.
```
MichaelDeBoey, line 208:
```suggestion
Sometimes you may also import packages that you intend for server use only. Sometimes this will cause immediate errors, but other times the compiler may try adding polyfills that increase your overall bundle size. Take for example core Node modules like `node:crypto`.
```

docs/pages/gotchas.md
```diff
@@ -184,3 +184,58 @@ Note that, even if this issue didn't exist, we'd still recommend using named re-
...
+
+import { json } from "@remix-run/node"; 
+
+export async function loader() {
+  const uuid = crypto.randomUUID();
+  return json({uuid});
+}
+```
+
+Importing crypto this way could lead to the bundler including polyfills for crypto into your client build even if it wasn't directly used on the client. An easy way to get around this is to centralize server imports inside a .server file. For example:
```
MichaelDeBoey, line 221:
```suggestion
Importing crypto this way could lead to the bundler including polyfills for crypto into your client build even if it wasn't directly used on the client. An easy way to get around this is to centralize server imports inside a `.server` file. For example:
```

docs/pages/gotchas.md
```diff
@@ -184,3 +184,58 @@ Note that, even if this issue didn't exist, we'd still recommend using named re-
...
+
+```tsx filename=app/utils/Node.server.ts
+import crypto from "node:crypto";
+import fs from "node:fs";
+
+export { fs, crypto };
+```
+
+```tsx good filename=app/routes/some/route.tsx
```
MichaelDeBoey, line 230:
```suggestion
```tsx good filename=app/routes/some-route.tsx
```

MichaelDeBoey (changes requested):

PR #7001, opened 2023-07-31
docs: use TS where possible

MichaelDeBoey:
This keeps things consistent in docs + in tests


brophdawg11:
I'd prefer to wait to merge anything of this scale until v2 is released.  That way if we do end up needing to do more 1.19.x releases we don't find ourselves with large merge conflicts from `main -> dev`.

MichaelDeBoey:
@brophdawg11 Although this is a docs change, I created this against `dev` for the exact reason of possible merge conflicts

brophdawg11 (approved):

PR #7008, opened 2023-07-31
templates: remove `baseUrl` from `tsconfig.json`

MichaelDeBoey:
> As of TypeScript 4.1,` baseUrl` is no longer required to be set when using [`paths`](https://www.typescriptlang.org/tsconfig#paths).

https://www.typescriptlang.org/tsconfig#baseUrl

---

Closes #6997


MichaelDeBoey:
It appears that we need `baseUrl` for some reason and we're throwing an error if it's not present

https://github.com/remix-run/remix/blob/d45f71febf1e1c9000b35fb72a48ea20614329da/packages/remix-dev/compiler/utils/tsconfig.ts#L15-L19

Marking this as draft to further investigate this

brophdawg11:
@pcattori Do you know why we need `baseUrl` in the compiler?

PR #7009, opened 2023-07-31
feat!: remove auto globals install

jacob-ebey:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey (commented):
@jacob-ebey Should we do the docs + templates changes in another PR against `main` as well?

Probably also a good idea to do a patch release for the changes in `@remix-run/serve`

MichaelDeBoey (approved):

PR #7015, opened 2023-08-01
[Proposal] Example PR for augmenting route modules on the client

AlemTuzlak:
A PR to demonstrate a proposal I wrote in the proposals channel (https://github.com/remix-run/remix/discussions/7016)


AlemTuzlak:
this is not required anymore thanks to Vite

PR #7023, opened 2023-08-01
docs(pages/v2): add `devServerPort` deprecation

MichaelDeBoey:


docs/pages/v2.md
```diff
@@ -579,6 +579,26 @@ module.exports = {
...
+  devServerPort: 8002,
+};
+```
+
+```js filename=remix.config.js good lines=[3-5]
+/** @type {import('@remix-run/dev').AppConfig} */
+module.exports = {
+  dev: {
+    port: 8002,
+  },
```
brophdawg11, line 598: I haven't used this but I think I was under the impression these got specified in the future flag?

```
future: {
  v2_dev: {
    port: 8002,
  }
}
```

Then I assume they'll flatten up to a root `dev` object once we release v2 - but this guide should be for folks on v1 preparing for v2?  And then maybe have a note on what the post v2 version will look like?

@pcattori can confirm/deny and provide final guidance though

pcattori: Yep, this guide currently is designed for folks who are trying to get all the v2 flags enabled in v1, so should be `v2_dev`:

```suggestion
  future: {
    v2_dev: {
      port: 8002,
    }
  },
```

pcattori (approved):

PR #7035, opened 2023-08-02
feat(remix-vercel)!: remove Vercel adapter

MichaelDeBoey:
Template was removed by #6784

PR #7036, opened 2023-08-02
templates(netlify): remove Netlify template in favor of official one

MichaelDeBoey:
https://github.com/netlify/remix-template


brophdawg11:
Can we add a major changeset for this?

MichaelDeBoey:
@brophdawg11 Since this is a template (not a package), we can't generate a changeset for it
We also don't have it for Vercel template btw

brophdawg11:
We could still apply a changeset to whatever package we want though right?  This feels like something worth noting against `@remix-run/dev` where `create-remix` currently lives?  

We can also add a separate changeset for the vercel template removal and point to the original PR

brophdawg11:
Yeah I just want to make sure that the auto-generated release notes specify that we deleted the `netlify` and `vercel` templates.  So it can go in the other changelog or a new one - doesn't really matter.  The advantage of doing it in this one is the automatic linking of the PR which did the actual deletion.

Otherwise, folks may think they can use the new `create-remix` CLI with them.  In v1, something like this would work - but it will fail in v2, so it feels like a breaking change worth noting.  Dev's don't really know/care where `create-remix` comes from

```
npx create-remix --template remix-run/remix/templates/netlify
```

MichaelDeBoey:
@brophdawg11 I've updated https://github.com/remix-run/remix/pull/7035 already
Does that look good for you?

brophdawg11:
I think that looks fine for Vercel.  Can we add a similar changeset as `fair-falcons-wink.md` to this PR to mention the removal of the netlify template?  It can be tagged as a major change for the netlify adatper and/or `remix-run/dev`.  I know we have it noted elsewhere, but at the moment the release notes will not say anything about the deletion of the netlify template without a changeset.

brophdawg11:
oh - I think that was just a wrong link above, I see the changeset in https://github.com/remix-run/remix/pull/7058 👍

brophdawg11 (approved):

brophdawg11:
This is good to go once conflicts are resolved 👍

PR #7039, opened 2023-08-02
feat(remix-serve): install `source-map-support`

MichaelDeBoey:
Follow-up of ##7009 & #7010

This should be in a patch release

### DIFF
--- a/.changeset/lucky-kangaroos-develop.md
+++ b/.changeset/lucky-kangaroos-develop.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/serve": patch
+---
+
+Install `source-map-support`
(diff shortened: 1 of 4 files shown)


brophdawg11:
OK cool.  This is fully backwards compatible with v1?  If so - let's repoint to `release-hotfix` and add a changeset to include in `1.19.2`

MichaelDeBoey:
@brophdawg11 It is indeed fully BC as the breaking change is the removal of `source-map-support` from `@remix-run/node`

I'll rebase and add a changeset

MichaelDeBoey:
@brophdawg11 Should be fine now 👍

PR #7044, opened 2023-08-03
chore: remove `abort-controller` dependency

MichaelDeBoey:
`AbortSignal` is stable as of Node v15.4

https://nodejs.org/api/globals.html#class-abortcontroller

---

This PR is relying on #6108, so that one needs to be merged first

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,8 +87,6 @@
     "@types/serialize-javascript": "^5.0.2",
     "@types/ssri": "^7.1.0",
     "@vanilla-extract/css": "^1.10.0",
-    "abort-controller": "^3.0.0",
-    "abortcontroller-polyfill": "^1.7.3",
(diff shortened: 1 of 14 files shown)


MichaelDeBoey:
@brophdawg11 Adding `@types/node` seems to have caused these 😕

Not sure how we should handle them though 🤔

MichaelDeBoey:
@brophdawg11 I managed to (temporarily) "fix" the type error with `NodeOnDiskFile` in #6108

Let's discuss there if this is an approach we like or not
Once merged, I'll rebase this one and this one should be fine again as well (I think)

MichaelDeBoey:
I rebased this branch on #6108 to have passing tests, but that one should be merged first before merging this one

jacob-ebey (approved):

PR #6211, opened 2023-04-27
chore(remix-dev): log when using future flags that have been removed or stablized

mcansh:
given the following `future` config:
```js
  future: {
    unstable_dev: true,
    v2_errorBoundary: true,
    v2_meta: true,
    v2_normalizeFormMethod: true,
    v2_routeConvention: true,
  },
```

image

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

PR #7048, opened 2023-08-03
Warn users about obsolete future flags in remix.config.js

brophdawg11:
Recreation of https://github.com/remix-run/remix/pull/6211 against `dev`

Run against my playground app which has all flags turned on + latest `nightly`:

Screenshot 2023-08-03 at 10 06 16 AM

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -608,7 +608,38 @@ export async function readConfig(
     tsconfigPath = rootJsConfig;
   }
 
+  // Note: When a future flag is removed from here, it should be added to the
+  // list below so we can let folks know if they have obsolete flags in their
+  // config.  If we ever convert remix.config.js to a TS file so we get proper
+  // typings this won't be necessary anymore.
   let future: FutureConfig = {};
+
+  if (appConfig.future) {
+    let userFlags = appConfig.future;
+    let deprecatedFlags = [
+      "unstable_cssModules",
+      "unstable_cssSideEffectImports",
+      "unstable_dev",
+      "unstable_postcss",
+      "unstable_tailwind",
+      "unstable_vanillaExtract",
+      "v2_dev",
+      "v2_errorBoundary",
+      "v2_headers",
+      "v2_meta",
+      "v2_normalizeFormMethod",
+      "v2_routeConvention",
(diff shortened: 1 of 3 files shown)


packages/remix-dev/config.ts
```diff
@@ -608,8 +608,39 @@ export async function readConfig(
...
 
+  if (appConfig.future) {
+    let userFlags = appConfig.future;
+    let deprecatedFlags = [
+      "unstable_cssModules",
+      "unstable_cssSideEffectImports",
+      "unstable_dev",
+      "unstable_postcss",
+      "unstable_tailwind",
+      "unstable_vanillaExtract",
```
MichaelDeBoey, line 625: We already warned about these in v1

```suggestion
```

MichaelDeBoey (changes requested):

packages/remix-dev/config.ts
brophdawg11: I was envisioning this as an ongoing list.  There's no guarantee that the v1 warning got addressed, so if it's still relevant in v2 shouldn't we still let them know?  And if it did get addressed the inclusion of it here is a no-op.

MichaelDeBoey: Than this could become a really long list in time as it will always be "relevant"

I would say that if they didn't address the warnings in v1, they won't in v2 either 🤷‍♂️
Having these flags will also be a no-op though

brophdawg11: I still learn towards including them because it's server-side build-time code, doesn't bloat any bundles, and no-ops if unnecessary.  Some `unstable_` flags only got removed recently ([1.16.0](https://github.com/remix-run/remix/releases/tag/remix%401.16.0)) and there's no guarantee that someone installed a version between 1.16.0 and v2.  Lots of folks only do upgrades once or twice a year so it's not uncommon to upgrade through a bunch of releases (i.e., 1.12 -> v2) and never see the short-lived v1 warning.

It's not to say that we need to keep v2 flags in something like v5, but having them live on through at least one major upgrade seems prudent to me.  Then someone would have to go v1 -> v3 to miss out.

MichaelDeBoey (approved):

PR #7060, opened 2023-08-04
fix(react): dedupe prefetch links

### LINKED ISSUE #5677: Duplicate children keys in `getStylesheetPrefetchLinks`

markdalgleish:
Closes #5677

- [ ] Docs
- [x] Tests

Fixing the linked issue highlighted a number of bugs in the code related to deduping links and generating keys for rendered links. I've added comments below to highlight the notable changes.


packages/remix-react/components.tsx
```diff
@@ -360,7 +360,7 @@ export function Links() {
 
         return (
           <link
-            key={link.rel + (link.href || "") + (imageSrcSet || "")}
+            key={key}
```
markdalgleish, line 363: They key is now the same value used for deduping rather than a custom subset of properties. To enable this, the utilities that return arrays of link descriptors to render now provide `Keyed` variants that expose a `key` property for each link. This ensures we never get any duplicate key errors, whereas the previous logic could result in duplicate keys if `rel`, `href` and/or `imageSrcSet` properties were the same but other properties differed.

packages/remix-react/links.ts
```diff
@@ -321,15 +323,17 @@ export async function getStylesheetPrefetchLinks(
...
-  return links
-    .flat(1)
-    .filter(isHtmlLinkDescriptor)
-    .filter((link) => link.rel === "stylesheet" || link.rel === "preload")
-    .map((link) =>
-      link.rel === "preload"
-        ? ({ ...link, rel: "prefetch" } as HtmlLinkDescriptor)
-        : ({ ...link, rel: "prefetch", as: "style" } as HtmlLinkDescriptor)
-    );
+  return dedupeLinkDescriptors(
```
markdalgleish, line 326: This is the core fix for #5677. The prefetch links weren't deduped as they were in `getLinksForMatches` (now called `getKeyedLinksForMatches`).

packages/remix-react/links.ts
```diff
@@ -481,14 +505,14 @@ export function dedupe(descriptors: LinkDescriptor[], preloads: string[]) {
       return deduped;
     }
 
-    let str = JSON.stringify(descriptor);
-    if (!set.has(str)) {
-      set.add(str);
-      deduped.push(descriptor);
+    let key = JSON.stringify(sortKeys(descriptor));
```
markdalgleish, line 508: This fixes an unreported bug where the deduping didn't work correctly if properties were specified in a different order.

packages/remix-react/links.ts
```diff
@@ -466,12 +470,32 @@ function dedupeHrefs(hrefs: string[]): string[] {
...
+function dedupeLinkDescriptors<Descriptor extends LinkDescriptor>(
+  descriptors: Descriptor[],
+  preloads?: string[]
+): KeyedLinkDescriptor<Descriptor>[] {
   let set = new Set();
   let preloadsSet = new Set(preloads);
 
   return descriptors.reduce((deduped, descriptor) => {
     let alreadyModulePreload =
+      preloads &&
```
markdalgleish, line 498: Since the `preloads` array is now optional, we now bail out early if it hasn't been provided.

packages/remix-react/links.ts
```diff
@@ -321,15 +323,17 @@ export async function getStylesheetPrefetchLinks(
...
-        : ({ ...link, rel: "prefetch", as: "style" } as HtmlLinkDescriptor)
-    );
+  return dedupeLinkDescriptors(
+    links
+      .flat(1)
+      .filter(isHtmlLinkDescriptor)
+      .filter((link) => link.rel === "stylesheet" || link.rel === "preload")
+      .map((link) =>
+        link.rel === "stylesheet"
+          ? ({ ...link, rel: "prefetch", as: "style" } as HtmlLinkDescriptor)
```
markdalgleish, line 333: This was just a slight refactor for legibility since "stylesheet" was checked first in the earlier filter.

packages/remix-react/components.tsx
```diff
@@ -483,10 +485,10 @@ function PrefetchPageLinksImpl({
       {moduleHrefs.map((href) => (
         <link key={href} rel="modulepreload" href={href} {...linkProps} />
       ))}
-      {styleLinks.map((link) => (
+      {keyedPrefetchLinks.map(({ key, link }) => (
         // these don't spread `linkProps` because they are full link descriptors
         // already with their own props
-        <link key={link.href} {...link} />
+        <link key={key} {...link} />
```
markdalgleish, line 491: This fixes an unreported bug where responsive image prefetch links would all have a key of `undefined` since they have `imgSrcSet` and `imgSizes` properties but no `href` property.

packages/remix-react/links.ts
```diff
@@ -306,11 +306,13 @@ function isHtmlLinkDescriptor(object: any): object is HtmlLinkDescriptor {
   return typeof object.rel === "string" && typeof object.href === "string";
 }
 
-export async function getStylesheetPrefetchLinks(
+export type KeyedHtmlLinkDescriptor = { key: string; link: HtmlLinkDescriptor };
+
+export async function getKeyedPrefetchLinks(
```
markdalgleish, line 311: I renamed this from `getStylesheetPrefetchLinks` to `getKeyedPrefetchLinks` because, apart from now being an array of objects with `key` and `link` so consumers don't need to generate keys, it also makes it clearer that this code also handles any link with `rel` set to `"preload"` (as seen on line `330`).

Issue #6676, opened 2023-06-23
npx create-remix fails in empty git repo directory


brophdawg11:
### What version of Remix are you using?

1.17.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

```
> mkdir temp
> cd temp
> git init
> npx create-remix .
🚨 The project directory must be empty to create a new project. Please clear the contents of the directory or choose a different path.
```

I think a directory with nothing but `.git/` can safely be considered empty.  This allows folks to start first with the git repo.

Original issue raised by @jacobparis in discord:

> My dev workflow is repo first 
> I make a new repo, then open it in a cloud dev environment and start mucking with it there
> There’s no time where I have any dev tools installed before the repo is created

CGamesPlay:
My workflow involves using direnv to specify a node version that the project will use. That means I need to set up a `.envrc` in the directory before I even run `npx`. It would be great if there were just an argument like `--overwrite` that would suppress this error and have remix simply overwrite any files which already exist.

jacobparis:
I'd be happy with that solution as well

brophdawg11:
cc @markdalgleish who has been doing some work on the `create-remix` CLI recently - let's sync up on this sometime next week

PR #7062, opened 2023-08-04
feat(create-remix): add `--overwrite` flag

brophdawg11:
We first implemented this as an `--allow-non-empty` flag.  

We pivoted and removed the empty directory requirement and moved to a more granular file-by-file check.  You are now free to write into a non-empty directory.  If no collisions are detected between the template and your directory - then it'll run normally with no additional user input.

If collisions are detected, we prompt for confirmation to overwrite:

Screenshot 2023-08-17 at 4 45 06 PM

Users can skip the confirmation prompt with `--overwrite`.

Closes https://github.com/remix-run/remix/issues/6676


packages/create-remix/__tests__/create-remix-test.ts
```diff
@@ -210,6 +210,25 @@ describe("create-remix CLI", () => {
...
+    let notEmptyDir = getProjectDir("non-interactive-not-empty-dir-allowed");
+    fse.mkdirSync(notEmptyDir);
+    fse.createFileSync(path.join(notEmptyDir, "some-file.txt"));
+
+    let { status, stderr, stdout } = await execCreateRemix({
+      args: [notEmptyDir, "--no-install", "--no-git-init", "--allow-non-empty"],
+      interactive: false,
+    });
+
+    expect(stdout).toMatch(
```
markdalgleish, line 223: This should fix the failing Windows test, similar to other tests in this file.

```suggestion
    expect(
      stdout.trim().replace("<TEMP_DIR>\\", "<TEMP_DIR>/") // Normalize Windows path
    ).toMatch(
```

markdalgleish (approved):
I think `--allow-non-empty` is okay given your point that it doesn't necessarily overwrite anything.

Should we include a note about the `--allow-non-empty` flag in the non-empty directory error message? Otherwise I feel like devs won't know this flag exists in the scenario from the linked issue.

Looks good otherwise, so I'm happy to approve and leave it with you :)

MichaelDeBoey (approved):

brophdawg11:
Chatted with @mjackson and we're going to make this a bit more granular.  We shouldn't care if the directory is empty, instead we should only care if we're actually going to overwrite stuff.  I'm going to update this to:

* No longer require an empty directory
* Check for collisions with existing files, and if found, prompt: "Are you sure you wish to overwrite?"
* Change the flag to `--overwrite-existing-files` to disable the prompt above

packages/create-remix/index.ts
```diff
@@ -43,7 +45,8 @@ async function createRemix(argv: string[]) {
   let steps = [
     introStep,
     projectNameStep,
-    templateStep,
+    copyTemplateToTempDirStep,
+    copyTempDirToAppDirStep,
```
brophdawg11, line 49: In order to do the file level comparison, we now extract  into a temp directory and then compare the temp dir to the destination dir.

packages/create-remix/index.ts
```diff
@@ -204,51 +216,28 @@ async function introStep(ctx: Context) {
 }
 
 async function projectNameStep(ctx: Context) {
-  let cwdIsEmpty = ctx.cwd && isEmpty(ctx.cwd);
```
brophdawg11, line 207: We no longer care if it's empty

packages/create-remix/index.ts
```diff
@@ -285,41 +274,27 @@ async function templateStep(ctx: Context) {
...
+
+      // TODO: Optimization - if template is just a local directory (not a
+      // local tarball etc.), just use that as ctx.tempDir for the rest of
+      // the pipeline and avoid copying it to a temp directory
+
+      await copyTemplate(template, ctx.tempDir, {
         debug: ctx.debug,
         token: ctx.token,
         async onError(err) {
-          let cwd = process.cwd();
```
brophdawg11, line 294: No need to do this cleanup anymore - it's extracting into a temp dir which the OS should cleanup automatically.

packages/create-remix/index.ts
```diff
@@ -329,12 +304,44 @@ async function templateStep(ctx: Context) {
...
+        `             Colliding files: ${files}\n` +
+        `             (You can skip this message with the --overwrite CLI flag)`,
+      initial: false,
+    });
+    if (!overwrite) {
+      throw new Error("Exiting to avoid overwriting files");
+    }
+  }
 
+  await fse.copy(ctx.tempDir, ctx.cwd);
```
brophdawg11, line 343: This is the population of the app directory now

packages/create-remix/utils.ts
```diff
@@ -266,3 +268,26 @@ export function action(key: ActionKey, isSelect: boolean) {
 
   return false;
 }
+
+export async function getDirectoryFilesRecursive(d: string) {
+  // Ignore comparing within these directories - but detect a collision
+  // at the directory level and count it.  These get prepended to strippedFiles
+  // in reverse order below
+  let ignoreDirs = ["node_modules", ".git"];
```
brophdawg11, line 276: We don't check file-by-file inside `.git`/`node_modules` but we treat them as collisions regardless.  There's a part of me that wants to add a hard stop on ever allowing an overwrite of `.git/` though...Thoughts?

MichaelDeBoey (approved):

.changeset/create-remix-override.md
```diff
@@ -0,0 +1,7 @@
+---
+"create-remix": minor
+---
+
+Remove empty directory checking in favor of `overwrite` prompt/flag.
+
+`create-remix` now allows you to write into an existing non-empty directory.  It will perform a file-level comparison and if the template will overwrite any existing files in the destination directory, it will prompt you if it's OK to overwrite the files in the destination directory.  If you answer no (the default) then it will exit without copying any files into the destination directory.  You may skip this prompt with the `--overwrite CLI flag.
```
markdalgleish, line 7:
```suggestion
`create-remix` now allows you to write into an existing non-empty directory.  It will perform a file-level comparison and if the template will overwrite any existing files in the destination directory, it will prompt you if it's OK to overwrite the files in the destination directory.  If you answer no (the default) then it will exit without copying any files into the destination directory.  You may skip this prompt with the `--overwrite` CLI flag.
```

packages/create-remix/utils.ts
markdalgleish: Yeah, I agree with this sentiment, overwriting `.git` feels pretty much guaranteed to be a mistake.

Rather than being an error, maybe we should just never copy the `.git` directory if it exists in a template? Especially since it's not possible to have one when the template is a GitHub repo, which is the primary mechanism for sharing a template.

brophdawg11: ooh, yeah I like that even better

brophdawg11: [048ca90](https://github.com/remix-run/remix/pull/7062/commits/048ca90b4e6fadf495421502ecdeb48fb8a5f6bd)

MichaelDeBoey (approved):

packages/create-remix/__tests__/create-remix-test.ts
```diff
@@ -882,6 +862,282 @@ describe("create-remix CLI", () => {
...
+      ],
+    });
+
+    expect(stderr.trim()).toBeFalsy();
+    expect(status).toBe(0);
+    expect(fse.existsSync(path.join(emptyDir, "package.json"))).toBeTruthy();
+    expect(fse.existsSync(path.join(emptyDir, "app/root.tsx"))).toBeTruthy();
+  });
+
+  it("does not copy .git nor node_modules directories if they exist in the template", async () => {
```
brophdawg11, line 886: @markdalgleish I decided to treat both `.git` and `node_modules` the same and not copy either of them from the template directory.  We probably shouldn't promote bloated templates with committed `node_modules` that we have to download.

packages/create-remix/index.ts
```diff
@@ -285,56 +277,117 @@ async function templateStep(ctx: Context) {
...
+            debug(message);
             await sleep(500);
           }
         },
       });
 
-      await updatePackageJSON(ctx);
+      if (result?.localTemplateDirectory) {
+        ctx.tempDir = path.resolve(result.localTemplateDirectory);
+      }
```
brophdawg11, line 309: Added the optimization that if the template is a local directory on disk we just copy from there instead of "extracting" it to a temp dir.

PR #7063, opened 2023-08-04
feat(remix-dev)!: remove `devServerBroadcastDelay` config option

MichaelDeBoey:


packages/remix-dev/devServer/liveReload.ts
```diff
@@ -26,13 +26,11 @@ export async function liveReload(
...
-        if (client.readyState === WebSocket.OPEN) {
-          client.send(JSON.stringify(event));
-        }
-      });
-    }, config.devServerBroadcastDelay);
+    wss.clients.forEach((client) => {
+      if (client.readyState === WebSocket.OPEN) {
+        client.send(JSON.stringify(event));
+      }
+    });
```
pcattori, line 33: `remix dev` has fixed the race conditions, but `remix watch` has not and still depends on `devServerBroadcastDelay`

MichaelDeBoey: Does this mean we won't be able to remove `devServerBroadcastDelay` at all?
Or should we first update `remix watch` to fix the race condition?

pcattori: `remix watch` architecture is inherently prone to race conditions. The plan is to provide options for `remix dev` that will obsolete `remix watch`.

E.g. `remix dev --mode=watch` or `remix dev --mode=live-reload` or `remix dev --mode=hot` or something like that. still need to nail down the API

For now, `remix watch` exists as a fallback until we land such an API for `remix dev`.

brophdawg11: Should we not have shipped this warning in `1.19.3`?  https://github.com/remix-run/remix/pull/7064

pcattori: @MichaelDeBoey can you set a `500` ms delay to replace `devServerBroadcastDelay`? that way we can remove `devServerBroadcastDelay` but still have a reasonable fallback for `remix watch` that avoids most race conditions. For anyone who is facing race conditions with `500` ms, the fix for them is to switch to `remix dev`.

pcattori (approved):

PR #7069, opened 2023-08-04
feat(remix-node): replace extended classes for `NodeRequest` & `NodeResponse` with interface type casts

### LINKED ISSUE #4395: fetch() doesn't return correct Response type

### LINKED ISSUE #7059: `fetch(...) instanceof Response` fails after installGlobals()

1.19

- [X] Yes

### Steps to Reproduce

Many libraries depend `resp instanceof Response` to return true; https://github.com/panva/oauth4webapi/blob/170f44e2f7c7da60feb8a654871c56fc0a741e6b/src/index.ts#L1513 is one of such a library. However, after Remix patches the global `fetch` instance, this no longer holds.

I've  tested this with the following three scenarios:

1. vanilla environment in node 20 (works)
2. remix-run/web-fetch (works)
3. remix-run/node (doesn't work)

```
{
  // correct behavior
  console.log((await fetch("http://example.com")) instanceof Response);
}
{
  // also correct behavior
  const i = await import("@remix-run/web-fetch");
  console.log((await i.fetch("http://example.com")) instanceof i.Response);
}
{
  // incorrect behavior; fails.
  const j = await import("@remix-run/node");
  console.log((await j.fetch("http://example.com")) instanceof j.Response);
}
```

The last one is important, as this is the version of globals that's installed via `installGlobals`.

### Expected Behavior

`(await fetch(...)) instanceof Response` should return `true` for fetch and Response exposed by `remix-run/node`.

### Actual Behavior

as the exposed types from `remix-run/node` is `NodeResponse`, the `instanceof` fails.

### ❓ Question 

What I don't particularly understand is why we have the `NodeRequest` and `NodeResponse`. This is the reason why it's causing this type mismatch issue; these classes seem to be serving very little purpose. We're actually not instantiating these types at all, most of the (non-test) code is using `... as NodeResponse` to coerce types. Can we just eliminate them, and expose `@remix-run/web-fetch` classes and types directly?

jeeyoungk:
Closes #3480
Closes #4148
Closes #4395
Closes #7059

This is a fix for the referenced issue; The fix is by

* Replace `NodeRequest` and `NodeResponse` classes with interfaces.
* However, class is still needed (as people may want to do `new NodeRequest` or `x instanceof NodeRequest`). For this purpose, `WebRequest` and `WebResponse` are exposed as classes.

This code is quite tricky as we're exporting types and values separately. The correct way of doing this is to define a type and value with the same identifier, `export` handles this correctly.

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.changeset/global-fetch-instanceof.md
+++ b/.changeset/global-fetch-instanceof.md
@@ -1,5 +1,4 @@
 ---
-"remix": patch
 "@remix-run/node": patch
 ---
 
--- a/.changeset/global-fetch-instanceof.md
+++ b/.changeset/global-fetch-instanceof.md
@@ -1,0 +1,6 @@
+---
+"remix": patch
+"@remix-run/node": patch
+---
+
+ensures fetch() return is instanceof global Response by removing extended classes for NodeRequest and NodeResponse in favor of custom interface type cast.
--- a/integration/fetch-globals-test.ts
+++ b/integration/fetch-globals-test.ts
@@ -9,7 +9,7 @@ let appFixture: AppFixture;
 test.beforeAll(async () => {
   fixture = await createFixture({
     files: {
-      "app/routes/index.jsx": js`
+      "app/routes/index.tsx": js`
         import { json } from "@remix-run/node";
         import { useLoaderData } from "@remix-run/react";
         export async function loader() {
--- a/integration/fetch-globals-test.ts
+++ b/integration/fetch-globals-test.ts
@@ -1,0 +1,39 @@
+import { test, expect } from "@playwright/test";
+
+import type { Fixture, AppFixture } from "./helpers/create-fixture";
+import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
+
+let fixture: Fixture;
+let appFixture: AppFixture;
+
+test.beforeAll(async () => {
+  fixture = await createFixture({
+    files: {
+      "app/routes/index.jsx": js`
+        import { json } from "@remix-run/node";
+        import { useLoaderData } from "@remix-run/react";
+        export async function loader() {
+          const resp = await fetch('https://reqres.in/api/users?page=2');
+          return (resp instanceof Response) ? 'is an instance of global Response' : 'is not an instance of global Response';
+        }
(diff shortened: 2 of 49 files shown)


jacob-ebey:
Closes #7059 #4148 #4395 #3480

jacob-ebey (approved):

.changeset/global-fetch-instanceof.md
```diff
@@ -0,0 +1,6 @@
+---
+"remix": patch
```
MichaelDeBoey, line 2:
```suggestion
```

integration/fetch-globals-test.ts
```diff
@@ -0,0 +1,39 @@
...
+import type { Fixture, AppFixture } from "./helpers/create-fixture";
+import { createAppFixture, createFixture, js } from "./helpers/create-fixture";
+
+let fixture: Fixture;
+let appFixture: AppFixture;
+
+test.beforeAll(async () => {
+  fixture = await createFixture({
+    files: {
+      "app/routes/index.jsx": js`
```
MichaelDeBoey, line 12:
```suggestion
      "app/routes/index.tsx": js`
```

MichaelDeBoey (changes requested):

jacob-ebey:
I've re-targeted this PR to the `dev` branch here: https://github.com/remix-run/remix/pull/7109

PR #6922, opened 2023-07-21
templates(express): use `watcher.on("add").on("change")`

xHomu:
Also add in comments on what to change for CJS

Does not close #6919 but the debugger might help with bug tracking.

- [x] Docs

This actually get the express template in line with existing v2 dev doc changes.

Testing Strategy:

PR #7078, opened 2023-08-05
fix(templates/express): dynamically import `chokidar` so it doesn't crash production

mcansh:
prior to this change if you built the remix app, pruned dev only dependencies and attempted to `npm start` the server would bail as it tried to import chokidar

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/templates/express/server.js
+++ b/templates/express/server.js
@@ -4,7 +4,6 @@ import * as url from "node:url";
 
 import { createRequestHandler } from "@remix-run/express";
 import { broadcastDevReady, installGlobals } from "@remix-run/node";
-import chokidar from "chokidar";
 import compression from "compression";
 import express from "express";
 import morgan from "morgan";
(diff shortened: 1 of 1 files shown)


MichaelDeBoey (approved):
I think we should merge this into `main` + we should probably also update dev_v2 docs with this

ngbrown:
This was also solved in https://github.com/remix-run/remix/pull/6922#pullrequestreview-1546618836, along with making CJS easy to switch to and fixing a race condition.

brophdawg11:
As a template change I agree this should go to `main` - but we should hold off until v2 releases since I think the express server is pretty different in `main` (v1 dev server) and `dev` (v2 dev server).

mcansh:
rebased!

i originally opened this PR before v2 went out which is the reason for it being pointed to `dev`, not really in the mood to go through the whole `git rebase --onto main dev (git branch --show-current) --committer-date-is-author-date` shooting match to keep this PR but repoint it to `main` so if that's desired instead of merging + cherrypicking this one, i can close and bring it over to a new PR pointed to `main`

pcattori (approved):

PR #7085, opened 2023-08-06
chore(remix-dev): remove `@babel/preset-env` from dependencies

TrySound:
Looks like preset env was added by accident or was not removed while refactoring. Now it is not referenced anywhere except root config and remix rely on esbuild for transpiling.

Preset env depends on many other packages which makes installing slower especially with bad internet connection.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -573,3 +573,4 @@
 - defjosiah
 - AlemTuzlak
 - ledniy
+- TrySound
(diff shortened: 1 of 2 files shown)


MichaelDeBoey (approved):

MichaelDeBoey:
@TrySound since this is a code-change, please rebase your branch onto latest `dev` and set base branch of this PR to `dev` as well

TrySound:
@MichaelDeBoey Done

PR #7086, opened 2023-08-06
Make remix-express agnostic of express

TrySound:
Remix depends on express without using most of its features and mostly relying on node http server inside of it.

Express is quite big package which may not be necessary in most apps. https://packagephobia.com/result?p=express

Here I replaced express types and covered missing features.
In the future this package could be renamed into remix-http or something like that.


brophdawg11:
Remix doesn't depend on Express - Remix runs on the Web Fetch API - this is just an adapter for Express to convert Express requests into Fetch Requests that can be handled by Remix.  So it doesn't make sense to remove the `express` dependency.  

You are free to use whatever adapter you choose, and even [write your own](https://remix.run/docs/en/main/other-api/adapter) if there's not one available for the HTTP server of your choosing.  There's even a raw node adapter already linked to from the docs: https://remix.run/docs/en/main/other-api/adapter#community-adapters

TrySound:
> Remix doesn't depend on Express

Not entirely true

https://github.com/remix-run/remix/blob/main/packages/remix-serve/index.ts#L1

https://github.com/remix-run/remix/blob/main/packages/remix-dev/devServer_unstable/index.ts#L8

TrySound:
@brophdawg11 the idea of this pr is to reduce install size of remix.
So user do no need to install unnecessary dependencies with bad internet. Im travelling a lot and have this problem all the time.

https://packagephobia.com/result?p=%40remix-run%2Fdev

Express in most cases is a fancy api around native http server. So if this adapter remove dependency on express users could use it with many other frameworks similar to express. Or just use http.createServer.

brophdawg11:
> Not entirely true

Sorry, I should have clarified - Remix at runtime does not rely on Express.  You are free to run with whatever HTTP server you want.

> the idea of this pr is to reduce install size of remix

That's not what this PR is doing?  You've only removed `express` from the depedencies of `@remix-run/express` - which doesn't make any sense since that is an adapter _specifically_ designed for folks who want to use `express`.  This is not a required dependency and therefore will not be included for installs of Remix unless the dev has opted-in.

Similarly, [`@remix-run/serve`](https://remix.run/docs/en/1.19.2/other-api/serve) is also an opt-in dependency, and relies on express by design: _"It's a production-ready, but basic Node.js server built with Express"_

`@remix-run/dev` _does_ currently use express an an implementation detail - and I left https://github.com/remix-run/remix/pull/7087 open because that could be worth considering (deferring to @pcattori there) to reduce the install size.  Please note that as that PR stands though, you've not removed `express` as a dependency so it won't reduce the install size of Remix.

TrySound:
> Similarly, [@remix-run/serve](https://remix.run/docs/en/1.19.2/other-api/serve) is also an opt-in dependency, and relies on express by design: "It's a production-ready, but basic Node.js server built with Express"

Could be built without it

kiliman:
Remix App Server uses Express middleware `morgan` and `compression`, so you can't just use the `http` module directly. There are alternatives to Express that are probably smaller and still compatible with Express middleware, like Fastify or hapi.

RAS is simple and basic, which is what is intended. The nice thing about Remix is that you can create your own server and adapter with whatever framework you want.

PR #7087, opened 2023-08-06
Ping in dev without express

TrySound:
Related to https://github.com/remix-run/remix/pull/7086

Express is very big. Native http server can do the same.


packages/remix-dev/devServer_unstable/index.ts
```diff
@@ -56,20 +55,33 @@ export let serve = async (
...
+  ) {
+    const [pathname] = (req.url ?? "").split("?");
+    if (req.method === "POST" && pathname === "/ping") {
+      const bodyBuffer = await new Promise<Buffer>((resolve, reject) => {
+        let buf: any[] = [];
+        req.on("data", (chunk) => buf.push(chunk));
+        req.on("end", () => resolve(Buffer.concat(buf)));
+        req.on("error", (err) => reject(err));
+      });
+      const body = JSON.parse(bodyBuffer.toString());
```
TrySound, line 71: Interesting thing is this api available since node v16.
```
import { json } from 'node:stream/consumers';

await json(stream)
```

ngbrown:
Can `express` also be removed from `package.json`?

https://github.com/remix-run/remix/blob/4441ff282c2da0b84c92d724fa4db1ba4af95edc/packages/remix-dev/package.json#L42

TrySound:
Not yet. There are other cases it hs used for. Will take time to remove completely.

PR #7099, opened 2023-08-07
feat!: stop passing `isTypeScript` to `remix.init` script

MichaelDeBoey:
Follow-up of @markdalgleish's #6887

### DIFF
--- a/.changeset/kind-grapes-clap.md
+++ b/.changeset/kind-grapes-clap.md
@@ -1,0 +1,6 @@
+---
+"create-remix": major
+"@remix-run/dev": major
+---
+
+Stop passing `isTypeScript` to `remix.init` script
--- a/packages/create-remix/index.ts
+++ b/packages/create-remix/index.ts
@@ -491,7 +491,6 @@ async function runInitScriptStep(ctx: Context) {
 
   let initScriptDir = path.dirname(ctx.initScriptPath);
   let initPackageJson = path.resolve(initScriptDir, "package.json");
-  let isTypeScript = fs.existsSync(path.join(ctx.cwd, "tsconfig.json"));
   let packageManager = ctx.pkgManager;
 
   try {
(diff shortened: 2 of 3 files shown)


markdalgleish:
This was deliberately left in as a backwards compatibility layer for existing templates. If we do this, we might want to fast-follow the v2 release with some PRs to community templates that rely on this flag.

markdalgleish (approved):

MichaelDeBoey:
@markdalgleish I'm happy to create PRs for our stacks + the official Netlify & Vercel templates (if needed)

I can also do the top 10 listed repos with [`remix-stack` tag](https://github.com/topics/remix-stack) if wanted

brophdawg11:
IIRC we would prefer to remove remix init entirely eventually?  Is there a large benefit to removing this flag?  It sounds like it just creates a bunch more work for us (per https://github.com/remix-run/remix/pull/7099#issuecomment-1668676809) so curious if the the juice is worth the squeeze?

MichaelDeBoey:
Imo it makes it more clear to template/stack creators that we're not supporting JS transpilation anymore so that they have to provide a JS version themselves if they want to support that

markdalgleish:
>IIRC we would prefer to remove remix init entirely eventually? Is there a large benefit to removing this flag? It sounds like it just creates a bunch more work for us (per https://github.com/remix-run/remix/pull/7099#issuecomment-1668676809) so curious if the the juice is worth the squeeze?

I looked into removing it as part of the new `create-remix` work but found that it will take us more work to remove it than to keep it at the moment due to the way our stacks heavily rely on it. I ultimately felt that removing `remix init` would be a bunch of work (and churn for community templates) without much payoff. I'm still in favour of removing it if there's a good opportunity it in the future though.

MichaelDeBoey:
I would say: the longer we keep it, more templates/stacks can start to rely on it, so more templates/stacks will need an update to remove it from their codebase

brophdawg11:
I'm on board with whatever you guys think - you've got more context than I do here - just want to avoid creating any unnecessary work :).  I don't think updating the top 10 community stacks is totally necessary if we've got this clearly called out as a breaking change in v2, but if the LOE is small than it could be a nice addition.

markdalgleish:
Yep, let's merge it.

PR #7105, opened 2023-08-08
feat(meta): add thrown `error` to the `meta` params

### LINKED ISSUE #3903: Meta function can't access `loader` data when it throws
### Context

Why is this important? [Screen reader users often rely on `title`'s to infer the content of a page.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title#accessibility_concerns) And the `loader` is used to dynamically set the `title` property in `meta` functions.

The inability to set title's dynamically on errors severely hinders this functionality.

1.6.7

### Steps to Reproduce

You need a `loader `that throws.

```tsx
import type { MetaFunction } from '@remix-run/node';
import { json } from '@remix-run/node';
import { Outlet, useCatch } from '@remix-run/react';

export const loader = () => {
  if (Math.random() > 0.5) {
    throw json(
      { message: 'not found', slug: 'foo', title: 'yolo' },
      { status: 400 },
    );
  }

  return json({ title: 'Not throwing' });
};

export const meta: MetaFunction = parameters => {
  console.log(parameters);
  return {};
};

export default function MyComponent() {
  return <p>Hello</>;
}

export function CatchBoundary() {
  const caught = useCatch();

  return (
    <p>
      {caught.status}: {caught.data.slug}
    </p>
  );
}
```

### Expected Behavior

The `meta` function should have access to the data of `json` that was `thrown`.

### Actual Behavior

The `meta` function doesn't see the `loader`'s data.

## Possible Alternative

In case of errors you can set the title in the higher routes that didn't throw, but that would require all routes always getting all titles, which is bad for performance 😕 

Or maybe there is another way to set the `title` in `CatchBoundary`s?

ryanflorence:
Closes #3903

- [x] Docs
- [x] Tests

Currently there's no access to the error in `meta` so you can't display a decent meta data for the page. This simply takes the error from the error object and passes it into the matching meta functions.


packages/remix-react/routeModules.ts
```diff
@@ -51,6 +51,7 @@ export interface MetaMatch<
   handle?: unknown;
   params: DataRouteMatch["params"];
   meta: MetaDescriptor[];
+  error: unknown;
```
jacob-ebey, line 54: Should this be optional?

packages/remix-react/routeModules.ts
```diff
@@ -74,6 +75,7 @@ export interface MetaArgs<
   params: Params;
   location: Location;
   matches: MetaMatches<MatchLoaders>;
+  error: unknown;
```
jacob-ebey, line 78: Optional?

packages/remix-react/components.tsx
```diff
@@ -516,6 +516,7 @@ export function Meta() {
   let meta: MetaDescriptor[] = [];
   let leafMeta: MetaDescriptor[] | null = null;
   let matches: MetaMatches = [];
+  let error = errors ? Object.values(errors)[0] : null;
```
brophdawg11, line 519: We may also want to test this if 2 different routes throw.  I forget offhand if we can end up with 2 entries in `errors`.  If so we may need this to just walk the matches top down and find the highest error.

ryanflorence: I'm pretty sure we only take the first error in order of parent > child routes even though the `errors` value is an object by route ids, it'll only ever contain one error. If a parent throws and error we can't go render deeper anyway.

ryanflorence: https://github.com/remix-run/react-router/blob/main/packages/router/router.ts#L1266-L1268

ryanflorence: https://github.com/remix-run/react-router/blob/main/packages/router/router.ts#L2050-L2052

ryanflorence: https://github.com/remix-run/react-router/blob/main/packages/router/router.ts#L2267-L2269

ryanflorence: There are a couple more but all the same story, please verify if I'm wrong :D

brophdawg11: ok so we can have 2 errors - but I don't know if we can write a test that causes that since errors are added to the object by the router in top-down order, and key order is preserved in modern JS engines - so this is getting the highest (but it feels risky):

Screenshot 2023-08-08 at 5 52 58 PM

What do you think about:

```suggestion
  let errorMatch = matches.find((m) => errors && errors[m.id] != null);
  let error = errors && errorMatch ? errors[errorMatch.id] : null;
```

brophdawg11: It's not obvious - but here's the buried code that can add multiple keys.  If two routes throw and they bubble to different boundaries (i.e., each has their own) - they both go in `errors` for the nearest ancestor boundary route.  But then downstream we only ever render until the highest one and thus never actually leverage the lower one from `errors`.

https://github.com/remix-run/react-router/blob/main/packages/router/router.ts#L3868

MichaelDeBoey (approved):

PR #7106, opened 2023-08-08
Handle disabled preload config in firefox

### LINKED ISSUE #2619: Firefox `network.preload: false` kills navigation

1.3.4

### Steps to Reproduce

In Firefox:

1. Type `about:config` to edit firefox configuration.
2. Disable `network.preload`
3. Navigate to https://remix.run
4. Click one of the CTAs

Nothing happens

It happens both, using remix, and in the remix website.

I'm not entirely sure if this is a remix or a firefox issue though 🤷 

### Expected Behavior

Navigation works correctly

### Actual Behavior

Links do not navigate, pressing "back" doesn't navigate either.

brophdawg11:
I'm not sure we _want_ this, but it's a POC for a potential fix for https://github.com/remix-run/remix/issues/2619

This adds some client-side detection of preload timeouts (after 3s) and once 3 stylesheet preloads timeout, we try to force a preload `link.onerror`.  If that also times out - we deduce that preload is not supported by the browser (or disabled by the user) and we stop trying to inject preloads for stylesheets on client side navigations.

Closes #2619

### DIFF
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -230,10 +230,7 @@ export async function prefetchStyleLinks(
   if (!routeModule.links) return;
   let descriptors = routeModule.links();
   if (!descriptors) return;
-
-  if (isPreloadDisabled) {
-    return;
-  }
+  if (isPreloadDisabled) return;
 
   // If we've hit our timeout 3 times, we may be in firefox with the
   // `network.preload` config disabled and we'll _never_ get onload/onerror
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -221,12 +221,45 @@ export function getLinksForMatches(
   return dedupe(descriptors, preloads);
 }
 
+let stylesheetPreloadTimeouts = 0;
+let isPreloadDisabled = false;
+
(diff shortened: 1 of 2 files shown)


ryanflorence (approved):
gotta love the look of production code 🤩

packages/remix-react/links.ts
```diff
@@ -221,12 +221,42 @@ export function getLinksForMatches(
...
   if (!routeModule.links) return;
   let descriptors = routeModule.links();
   if (!descriptors) return;
+  if (isPreloadDisabled) return;
+
+  // If we've hit our timeout 3 times, we may be in firefox with the
+  // `network.preload` config disabled and we'll _never_ get onload/onerror
+  // callbacks.  Let's try to confirm this with a totally invalid link preload
+  // which should immediately throw the onerror
+  if (stylesheetPreloadTimeouts >= 3) {
```
emilio, line 239: Commented on the Firefox bug, but you really want `document.createElement("link").relList.supports("preload")` or so.

In any case, I wouldn't bother dealing with it unless you want to also deal with browsers that don't support `<link rel=preload>` (which is fine, but can be simpler as per the above).

PR #7119, opened 2023-08-09
chore: fix `deduplicate-yarn` workflow

MichaelDeBoey:
This should fix the workflow normally 🙏

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

### DIFF
--- a/.github/workflows/deduplicate-yarn.yml
+++ b/.github/workflows/deduplicate-yarn.yml
@@ -11,6 +11,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: write
+
 jobs:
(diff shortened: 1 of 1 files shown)


brophdawg11:
Do we know why https://github.com/remix-run/remix/blob/main/.github/workflows/format.yml doesn't need this?  Could we use the same approach that uses?

I'm not quite sure how that workflow has write permissions - I wouldn't think the `FORMAT_PAT` used in the checkout step would apply to the commit step - but maybe it does?

MichaelDeBoey:
@brophdawg11 It's indeed because that workflow uses a different token that has these permissions
The default GH token doesn't have these permissions by default, hence why you need to set them

I was planning on chaning other workflows to the approach I'm using here as well in a follow-up PR

PR #6331, opened 2023-05-06
chore: let `@typescript-eslint/no-floating-promises` ESLint rule pass

ngbrown:
## Summary of changes

This is just the relevant parts of the code changes from #6325 instead of also tackling eslint-config changes.

- Added [`void` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void) to Promises that are not being awaited. Used `await` on finally retry in test.

## Details

When enabling the typescript-eslint rules contained in ["plugin:@typescript-eslint/recommended-requiring-type-checking"](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking) I tried to improve code quality by addressing some of the issues raised when enabling the type checked linting rules.

This change specifically resolves issues raised with the typescript-eslint error: [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises).

> error  Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises

Testing Strategy:

> I opened up my windows machine and ran this script (with the additional eslint rules enabled from #6325):
>
> ```
> yarn lint
> yarn test:primary
> npx playwrite install
> yarn test
> yarn build --tsc
> ```

The specific rule configuration was:

```js
  "@typescript-eslint/no-floating-promises": [
    "error",
    { ignoreVoid: true, ignoreIIFE: true },
  ],
```

PR #7123, opened 2023-08-10
chore: let `@typescript-eslint/require-await` ESLint rule pass

JoshuaKGoldberg:
Sibling PR to @ngbrown's excellent #6325 -> #6331. I noticed [`@typescript-eslint/require-await`](https://typescript-eslint.io/rules/require-await) wasn't in the list, and it's in the new [`plugin:@typescript-eslint/recommended-type-checked`](https://typescript-eslint.io/linting/configs#recommended-type-checked) preset.

[`@typescript-eslint/require-await`](https://typescript-eslint.io/rules/require-await) disallows async functions which have no await expression. This is nice because `async` adds an extra bit of work to the function (returns a `Promise` instead of a direct value).

Separately, I've also noticed a trend in the React ecosystem of APIs requiring functions return a `Promise` even when they wouldn't always want to. That's in direct conflict with this rule. Enabling the rule in frameworks helps ensure frameworks aren't in conflict: by either making necessary changes to the rule _and/or_ blocking conflict-y things in the frameworks.


packages/remix-server-runtime/utils.ts
```diff
@@ -0,0 +1 @@
+export type PromiseOrSync<T> = Promise<T> | T;
```
JoshuaKGoldberg, line 1: Is there a better place for `PromiseOrSync`?

JoshuaKGoldberg:
Err, I can't tell if the [CI failures on Firefox](https://github.com/remix-run/remix/actions/runs/5843054268/job/15844810827?pr=7123) are a result of the changes in this PR. [#7133 shows similar failures](https://github.com/remix-run/remix/actions/runs/5826527875/job/15800688763?pr=7133). Marking as ready for review on a hunch that they're unrelated.

packages/remix-server-runtime/index.ts
```diff
@@ -77,3 +77,5 @@ export type {
   UploadHandler,
   UploadHandlerPart,
 } from "./reexport";
+
+export type { PromiseOrSync } from "./utils";
```
MichaelDeBoey, line 81: Don't think we need to expose this type 🤔

scripts/playground/template/app/models/user.server.ts
```diff
@@ -5,11 +5,11 @@ import { prisma } from "~/db.server";
 
 export type { User } from "@prisma/client";
 
-export async function getUserById(id: User["id"]) {
```
MichaelDeBoey, line 8: This is copy/pasted from our `indie-stack`, so I would leave it in sync tbh

Don't know if we should change it there though, as I personally like the fact that you can see straight away that a function is async or not (Prisma returns promises)

JoshuaKGoldberg: > in sync

Heh, I would suggest enabling this rule in `indie-stack` then. Then they'd stay in sync!

> is async or not

But the type signature exposed to end users is the same, no? `(id: User["id"]) => Promise<User>`? 

Aside: it is an interesting, subtle difference - being "async" vs. synchronously returning a Promise... I'd posit that it's even a little misleading to see a function being marked as `async` but then not itself including awaiting code. 

<details>
<summary>Details on why <code>return await</code> can improve call stacks, for any curious readers</summary>

If a `try`/`catch` wraps an `await`, only functions that have an `await` inside will be included in the call stack. Running snippet demonstrates different call stacks:

```js
async function inside() {
  await new Promise(setTimeout);
  throw new Error("Oh no!");
}

async function middle(useAwait) {
  return useAwait ? await inside() : inside();
}

async function outer(useAwait) {
  try {
    await middle(useAwait);
  } catch (error) {
    return error;
  }
}

console.log("Using await:", await outer(true));
console.log("Without await:", await outer(false));
```

Output in Node:

```plaintext
$ node temp.mjs
Using await: Error: Oh no!
    at inside (file:///path/to/temp.mjs:3:9)
    at async middle (file:///path/to/temp.mjs:7:21)
    at async outer (file:///path/to/temp.mjs:12:5)
    at async file:///path/to/temp.mjs:18:29
Without await: Error: Oh no!
    at inside (file:///path/to/temp.mjs:3:9)
    at async outer (file:///path/to/temp.mjs:12:5)
    at async file:///path/to/temp.mjs:19:31
```

</details>

Ref: https://typescript-eslint.io/rules/return-await

If you'd really prefer this be `async`, what do you think about having this `return await prisma.user.findUnique`? That way the `getUserById` function is included in call stacks in case of throwing an error.

MichaelDeBoey: The signature will be the same, so there's no problem in that day
It's just that looking at the code you immediately see if a func is async or not without needing to know if Prisma returns a Promise

I don't like awaiting a promise you return either tbh 🙈

But again: this is my personal taste and I'm happy to follow whatever the team decides here and just get used to it 🤷‍♂️

packages/remix-server-runtime/utils.ts
MichaelDeBoey: It's slightly different from the [`Promisable`](https://github.com/sindresorhus/type-fest/blob/HEAD/source/promisable.d.ts) type of @sindresorhus' `type-fest` package, but I guess the end-result is actually the same?

Should we use that one instead? 🤔

MichaelDeBoey (approved):

scripts/playground/template/app/models/user.server.ts
brophdawg11: I think we want to mostly just take whatever the preset/recommended configs include and avoid any type of bikeshedding over eslint rules/prefs.  Since `@typescript-eslint/recommended-type-checked` incudes the rule I vote we make it pass here and (eventually when we re-work our eslint configs) in the stack

packages/remix-server-runtime/crypto.ts
```diff
@@ -1,9 +1,14 @@
...
+export type SignFunction = (
+  value: string,
+  secret: string
+) => Promisable<string>;
 
 export type UnsignFunction = (
   cookie: string,
   secret: string
-) => Promise<string | false>;
+) => Promisable<string | false>;
```
brophdawg11, line 11: Is this a breaking API change since `unsign` is publicly exported?  Won't `unsign` always return a `Promise` today, so `unsign().then()` works - but after this `unsign` just returns a `string | false` (the return type of `cookieSignature.unsign`) so `unsign().then(` would break?

If so, I think we should keep the type API consistent and change the implemtnation:

```
export const unsign: UnsignFunction = (signed: string, secret: string) => {
  return Promise.resolve(cookieSignature.unsign(signed, secret));
};
```

MichaelDeBoey: @brophdawg11 Indeed a good one I didn't think of

Wrapping it in a `Promise` would indeed be the better option in this case I think @JoshuaKGoldberg

JoshuaKGoldberg: Makes sense! I'll change the functions to `Promise.resolve(...)`. Thanks!

MichaelDeBoey (approved):

brophdawg11:
This seems ok to me but I'm going to defer to @pcattori for an approval because it touches a lot of stuff in the dev server

pcattori:
Hey @JoshuaKGoldberg , thanks for your work on this! This PR went up right when we started focusing on shipping v2, so hasn't yet gotten the attention it deserves. But hoping to change that now!

One thing that I don't like about this new lint rule is that it feels like the function signature/interface is now coupled to its implementation. So let's say we find a way to make an `async` function not need any `awaits`, so we remove the `async`. But now that's a breaking change for users. So then I can keep `async`, but this lint rule will make noise about that.

I think there are also times where I want to model something as potentially being async, knowing that my implementation of it may change day-to-day.

But it is on [plugin:@typescript-eslint/recommended-type-checked](https://typescript-eslint.io/linting/configs#recommended-type-checked) and I probably haven't thought through this as deeply as those folks.

tl;dr: If you want to rebase this PR on latest `main`, I'd be ok with merging it, but have some mixed feelings about it.

Issue #1678, opened 2022-01-27
[Bug]: Client-only errors are not handled correctly on initial load. (Infinite loop!)


maulerjan:
### What version of Remix are you using?

1.1.3

### What version of Node are you using? Minimum supported version is 14.

17.3.0

### Steps to Reproduce

1. Clone the simple repro case: https://github.com/maulerjan/remix-issue-1678-repro
2. npm install & npm run dev
3. Click on the /about link and then refresh the page

### Expected Behavior

Renders ErrorBoundary.

### Actual Behavior

ErrorBoundary mounting errors resulting in an infinite loop.

The server correctly renders the page with no error (SSR), but then there's the client-only error during hydration. I'm not sure what exactly happens next, but the whole thing ends up in an infinite loop of mounting ErrorBoundary, which itself throws an error, which again tries to mount the ErrorBoundary and so on.

EDIT:
If the client-only error happens in the root component (`App`), then it's handled correctly, so this only affects nested modules' components.

dvargas92495:
In my case, the initial bug that caused this was having a reference to `process.env.VAR`. This led me to Remix's environment docs which unblocked me: https://remix.run/docs/en/v1/guides/envvars#server-environment-variables

maulerjan:
@dvargas92495 Glad you fixed it. But unfortunately fixing the error doesn't fix the error handling. What if it was your customer who found the bug?

dvargas92495:
Right I agree completely. Just wanted to mention it in case anyone else run into something similar and have a hard time debuggine because of the infinite loop nature of the issue

csagar131:
If anyone found a any solution of this issue, because currently i am facing similar issue, I am navigating a router there a loader function which is making a api calls, but whenever api is down I am running into an infinite loop of client side errors.

Mange:
This happened to me when I was using a value imported from a `.server.ts` file inside of a component render, which meant that server-side everything worked but client-side it would crash because the imported value was `undefined`.

Took me a long time to figure out since the tab would log around 300k errors per second and filled up the RAM so I had to binary search the code to find the crashing line by commenting out things and then killing the chrome tab on every reload that the error was still active.

Example code:

```ts
import { MyEnum } from "some_api.server";

export const loader = () => ({value: MyEnum.Foo });

export default function MyPage() {
  const { value } = useLoaderFunction();
  if (value == MyEnum.Foo) { // <-- Crash here; MyEnum is undefined since the import was stripped away in client bundle.
[...]

PR #7133, opened 2023-08-10
Update Remix to use route.lazy from React Router

brophdawg11:
Leverage `route.lazy` from React Router

**TODO:**
* [x] Needs https://github.com/remix-run/react-router/pull/10778 to be added to a RR experimental release before all of CI will pass
* [x] Handle forced HMR revalidation


packages/remix-react/browser.tsx
```diff
@@ -169,6 +169,26 @@ if (import.meta && import.meta.hot) {
...
+    let hydratedPathname = window.location.pathname;
+    if (initialPathname !== hydratedPathname) {
+      let errorMsg =
+        `Initial URL (${initialPathname}) does not match URL at time of hydration ` +
+        `(${hydratedPathname}), reloading page...`;
+      console.error(errorMsg);
+      window.location.reload();
+      // Get out of here so the reload can happen - don't create the router
+      // since it'll then kick off unnecessary route.lazy() loads
+      return <></>;
```
brophdawg11, line 189: Without this we get into I think the same hydration error loop as https://github.com/remix-run/remix/issues/1678.  I hope to dig deeper into that later today/tomorrow so maybe it'll shed some light on why we need the short circuit, but once I move this up and return (and avoid any router initialization + lazy calls on the wrong route) the E2E test passes and all of my manual testing with rapid forward clicks into a Remix app to cause a mismatched URL at hydration time settle correctly on the final URL after the hard reload.

packages/remix-react/components.tsx
```diff
@@ -83,55 +82,6 @@ function useRemixContext(): RemixContextObject {
   return context;
 }
 
-////////////////////////////////////////////////////////////////////////////////
-// RemixRoute
-
-export function RemixRoute({ id }: { id: string }) {
```
brophdawg11, line 89: `RemixRoute` and `RemixRouteError` were sort of poor-mans implementations of `Component`/`ErrorBoundary` lazy loading.  We could assign them to a route definition up front (without a route module available) knowing that they'd pull from the lazily loaded `routeModulesCache` at render-time.  Now we can just assign `Component`/`ErrorBoundary` via `lazy()`.

The only downside is we lose these error messages - so if we think they are worth keeping we can stick with this approach.

packages/remix-react/routes.tsx
```diff
@@ -126,27 +125,63 @@ export function createClientRoutes(
...
+        if (!route.hasAction) {
+          let msg =
+            `Route "${route.id}" does not have an action, but you are trying ` +
+            `to submit to it. To fix this, please add an \`action\` function to the route`;
+          console.error(msg);
+          return Promise.reject(new Error(msg));
+        }
+
+        return fetchServerHandler(request, route);
+      },
```
brophdawg11, line 147: Lifted the forked logic from `createDataFunction` to inline here so we can share 100% of `fetchServerHandler`

packages/remix-react/routes.tsx
```diff
@@ -126,27 +125,63 @@ export function createClientRoutes(
...
+              if (needsRevalidation) {
+                mod.shouldRevalidate = wrapShouldRevalidateForHdr(
+                  route.id,
+                  mod.shouldRevalidate,
+                  needsRevalidation
+                );
+              }
+              return mod;
+            },
+          }),
```
brophdawg11, line 182: This is the meat of this PR - non-critical routes get everything populated via `lazy` now

PR #7137, opened 2023-08-11
fix(react): fix useLoaderData types in Yarn PnP

### LINKED ISSUE #6994: Type of `useLoaderData<typeof loader>()` doesn’t get inferred when using Yarn PnP

1.19.1

- [X] Yes

### Steps to Reproduce

1. `yarn create remix` with default settings (`my-remix-app`, Just the basics, Remix App Server, TypeScript, run `yarn install`)
2. `cd my-remix-app`
3. Make the following changes:
   ``` diff
   diff --git a/app/root.tsx b/app/root.tsx
   index 8cb74a167..ad3a65b41 100644
   --- a/app/root.tsx
   +++ b/app/root.tsx
   @@ -7,13 +7,17 @@
   import {
      Outlet,
      Scripts,
      ScrollRestoration,
   +  useLoaderData,
    } from "@remix-run/react";
    
    export const links: LinksFunction = () => [
      ...(cssBundleHref ? [{ rel: "stylesheet", href: cssBundleHref }] : []),
    ];
    
   +export const loader = () => ({ some: "data" });
   +
    export default function App() {
      return (
        <html lang="en">
          <head>
   ```
4. `yarn set version berry`
5. `yarn` (you should see the message “automatically enabling the compatibility node-modules linker 👍”, a line `nodeLinker: node-modules` will be added to `.yarnrc.yml`)
6. When using VSCode: open command panel (`⇧ ⌘ P` on macOS) → `TypeScript: Select TypeScript Version…` → `Use Workspace Version`
7. Observe that the inferred type of `data` is still (correctly) `SerializeObject<UndefinedToOptional<{ some: string; }>>`
8. Remove `nodeLinker: node-modules` from `.yarnrc.yml` (or change it to `nodeLinker: pnp`) and run `yarn` again. This will remove your `node_modules` folder and enable the [Plug‘n’Play module resolution mechanism](https://yarnpkg.com/features/pnp)
9. Run `yarn dlx @yarnpkg/sdks vscode`
10. You’ll have to set the VSCode TypeScript version again like in step 6
11. Observe that the inferred type of `data` is now `SerializeFrom<T>`. `Cmd`+clicking on `useLoaderData` will go to line 114 in `components.d.ts`:
    ```ts
    /**
     * Returns the JSON parsed data from the current route's `loader`.
     *
     * @see https://remix.run/hooks/use-loader-data
     */
    export declare function useLoaderData<T = AppData>(): SerializeFrom<T>;
    ```
    but it’s not possible to `Cmd`+click further on `SerializeFrom`.

### Expected Behavior

The inferred type of `useLoaderData<typeof loader>()` remains `SerializeObject<UndefinedToOptional<{ some: string; }>>` as in step 7

### Actual Behavior

The inferred type of `useLoaderData<typeof loader>()` is `SerializeFrom<T>` as described in step 11

markdalgleish:
Fixes #6994

The `useLoaderData` return type depends on access to the `SerializeFrom` type exported by `@remix-run/server-runtime`. Since this package is listed as a dev dependency of `@remix-run/react` rather than a regular dependency, this type isn't currently available to consumers when running Yarn PnP.

### DIFF
--- a/.changeset/useloaderdata-types-yarn-pnp.md
+++ b/.changeset/useloaderdata-types-yarn-pnp.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Fix types for `useLoaderData` when using Yarn PnP
--- a/packages/remix-react/package.json
+++ b/packages/remix-react/package.json
@@ -17,10 +17,10 @@
   "module": "dist/esm/index.js",
   "dependencies": {
     "@remix-run/router": "1.7.2",
+    "@remix-run/server-runtime": "1.19.3",
     "react-router-dom": "6.14.2"
   },
   "devDependencies": {
-    "@remix-run/server-runtime": "1.19.3",
     "@testing-library/jest-dom": "^5.17.0",
     "@testing-library/react": "^13.3.0",
     "@types/react": "^18.0.15",


MichaelDeBoey:
This feels very weird to do so and makes me think about the initiative @pcattori started at some point to look into extracting certain packages

I think back then the decision was made to copy/paste certain types and maintain 2 versions of it (for now)

Maybe we should either do that here as well _or_ since it's v2 it can be a perfect time to restart the initiative as well 🤔

markdalgleish:
I chatted offline with @MichaelDeBoey, but just updating here too.

My feeling is that this is okay as a short-term fix since it's not changing any of the relationships between the packages — it's simply fixing an issue where `@remix-run/react/package.json` is out of sync with reality. I agree that it's a good idea to revisit the current architecture at some point, but I don't think it needs to hold up this fix in the meantime, and there's no down side from a consumer's perspective.

pcattori:
> My feeling is that this is okay as a short-term fix since it's not changing any of the relationships between the packages — it's simply fixing an issue where `@remix-run/react/package.json` is out of sync with reality. I agree that it's a good idea to revisit the current architecture at some point, but I don't think it needs to hold up this fix in the meantime. We could duplicate the `SerializeFrom` type to avoid the `server-runtime` dependency, but I think that will make refactoring harder in the future, and including it as a dependency doesn't seem to have any down side from a consumer's perspective.

I agree. Originally, the package reorg work was going to happen early this year, but we deprioritized it since it would be a breaking change, plus we wanted to focus on HMR/HDR work. Let's add `@remix-run/server-runtime` back as a dependency and we can clean it up when we tackle package reorg in v2.

MichaelDeBoey (approved):

PR #7138, opened 2023-08-11
feat(remix-react): export `PrefetchBehavior` type

kinxori:
…om ./components and ./index

### DIFF
--- a/.changeset/sixty-insects-tell.md
+++ b/.changeset/sixty-insects-tell.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": minor
+---
+
+Export `PrefetchBehavior` type
--- a/contributors.yml
+++ b/contributors.yml
@@ -574,3 +574,4 @@
 - AlemTuzlak
 - ledniy
 - TrySound
+- kinxori
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -141,8 +141,9 @@ export function RemixRouteError({ id }: { id: string }) {
  * - "intent": Fetched when the user focuses or hovers the link
  * - "render": Fetched when the link is rendered
  * - "none": Never fetched
+ * - "viewport": Fetches while the link is in the viewport
  */
-type PrefetchBehavior = "intent" | "render" | "none" | "viewport";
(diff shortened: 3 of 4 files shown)


MichaelDeBoey:
@kinxori Can you provide a bit more info about the use-case why you want to use this type?

kinxori:
@MichaelDeBoey Sure!! 

I attached a video showcasing the actual usage. 

Bear with me if I am not as explicit. 😅

https://github.com/remix-run/remix/assets/116584982/d98ac89d-221e-4ed2-8530-b38a541e81c2

👉 The main purpose of this is to use the correct and explicit Type of "prefetch" attribute inside <Link>. 

I also noticed that the Types are supposed to come from "RemixLinkProps.PrefetchBehavior?:" but are not exported neither. 

(typescript lantern on type recommendation):
image

(RemixLinkProps.PrefetchBehavior?:" is not exported):
image

So, exporting the types for "prefetch" we can actually use the "PrefetchBehavior" instead of writing every property as string. 

I hope this helps! 😁

sergiodxa:
I think instead of `RemixLinkProps` you can use `LinkProps` directly, then `LinkProps["prefetch"]` should give you the type.

But I think that it's better to just export `PrefetchBehavior` directly instead of forcing app devs to extract it from the `LinkProps`.

sergiodxa:
Also, PRs that change code should go to the `dev` branch on the `remix-run` repo, could you change it on the PR?

sergiodxa (approved):

kinxori:
@sergiodxa PR changed from "main" to "dev". 

Is everything alright by now? 

If not, let me know! 😁

Thankyou! ✌️

kinxori:
By the way. I'm not sure If a clicked something wrong. I apologize in advanced! 😅

MichaelDeBoey (approved):

brophdawg11:
don't actually mind asking users to use `LinkProps["prefetch"]` as IMO it aligns more closely with `Link`/`LinkProps` being the "public" API.  The internal type used by `<Link prefetch>` feels a bit like an implementation detail to me. Also using `LinkProps["prefetch"]` makes it very explicit that this is a type specifically intended for use with `<Link prefetch>` and not something that should be leveraged elsewhere.

PR #7150, opened 2023-08-14
fix: use `headers.entries()` instead of `headers.raw()`

### LINKED ISSUE #4354: Remix should not rely on Response.headers.raw

1.7.2

### Steps to Reproduce

Any `Response` object that is created through other means than `@remix-run/node` (non node-fetch polyfills/undici/node builtins) fails here because Remix relies on the non standard `Reponse.headers.raw`:

https://github.com/remix-run/remix/blob/e260b2ba153271fed2e2c569abdfecdccf241625/packages/remix-express/server.ts#L125-L129

### Expected Behavior

Remix should use the standard `Response.headers` interface to maximize compatibility, eg:

````ts
  for (let [key, value] of nodeResponse.headers.entries()) {
    res.append(key, value);
  }
````

### Actual Behavior

Handing off a `Response` to Remix results in an error:

````ts
TypeError: nodeResponse.headers.raw is not a function
    at sendRemixResponse (C:\projects\alinea\examples\framework-remix\node_modules\@remix-run\express\dist\server.js:89:65)
    at C:\projects\alinea\examples\framework-remix\node_modules\@remix-run\express\dist\server.js:40:13
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
````

### PR DESCRIPTION
**Pull request by MichaelDeBoey**
Closes #4354

### DIFF
--- a/packages/remix-express/server.ts
+++ b/packages/remix-express/server.ts
@@ -118,10 +118,8 @@ export async function sendRemixResponse(
   res.statusMessage = nodeResponse.statusText;
   res.status(nodeResponse.status);
 
-  for (let [key, values] of Object.entries(nodeResponse.headers.raw())) {
-    for (let value of values) {
-      res.append(key, value);
-    }
+  for (let [key, value] of nodeResponse.headers.entries()) {
+    res.append(key, value);
   }
 
   if (nodeResponse.body) {
--- a/.changeset/orange-suits-nail.md
+++ b/.changeset/orange-suits-nail.md
@@ -1,0 +1,7 @@
+---
+"@remix-run/architect": patch
+"@remix-run/express": patch
+---
+
+- Switch to `headers.entries()` instead of non-spec-compliant `headers.raw()` in `sendRemixResponse`
+- Update to `@remix-run/web-fetch@4.3.7`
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by brophdawg11** [comment]
@mjackson the move to `raw` came from https://github.com/remix-run/remix/pull/241 - do you have context on whether it's safe to switch back?

File: packages/remix-express/server.ts [hunk 1]
```diff
@@ -118,7 +118,7 @@ export async function sendRemixResponse(
   res.statusMessage = nodeResponse.statusText;
   res.status(nodeResponse.status);
 
-  for (let [key, values] of Object.entries(nodeResponse.headers.raw())) {
+  for (let [key, values] of nodeResponse.headers.entries()) {
     for (let value of values) {
       res.append(key, value);
     }
```
brophdawg11, line 124: This breaks for me since `entries` returns strings, not arrays in `values` like `raw` did so this then iterates over the characters in the `values` string.  

I think the original bug actually boils down to a bug in the node-fetch `entries` implementation (which we inherited into the Remix fork) - I pushed a potential fix here: https://github.com/remix-run/web-std-io/pull/39

If we were to merge that fix, then I think this would become:

```suggestion
  for (let [key, value] of nodeResponse.headers.entries()) {
    res.append(key, value);
```

brophdawg11 (approved):

PR #7153, opened 2023-08-14
chore: update `esbuild`

MichaelDeBoey:
This should trigger the `deduplicate-yarn` workflow and the `yarn.lock` file should be deduplicated now

### DIFF
--- a/packages/remix-dev/package.json
+++ b/packages/remix-dev/package.json
@@ -36,7 +36,7 @@
     "chalk": "^4.1.2",
     "chokidar": "^3.5.1",
     "dotenv": "^16.0.0",
-    "esbuild": "0.17.6",
+    "esbuild": "^0.17.19",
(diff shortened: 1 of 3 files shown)


packages/remix-dev/package.json
```diff
@@ -36,7 +36,7 @@
     "chalk": "^4.1.2",
     "chokidar": "^3.5.1",
     "dotenv": "^16.0.0",
-    "esbuild": "0.17.6",
+    "esbuild": "^0.17.19",
```
brophdawg11, line 39: We should check with @pcattori/@markdalgleish/@mjackson here - I believe this is pinned intentionally because `esbuild` is `0.x` and will introduce breaking changes occasionally.  We've gotten bit a few times on upgrades and have to be really careful with them.  We probably need someone from the team to go through the release notes to ensure everything since `0.17.6` is ok to adopt.

PR #7154, opened 2023-08-14
chore: update `@types/cacache`

MichaelDeBoey:
This should trigger the `deduplicate-yarn` workflow and the `yarn.lock` file should be deduplicated now

### DIFF
--- a/packages/remix-dev/package.json
+++ b/packages/remix-dev/package.json
@@ -73,7 +73,7 @@
   },
   "devDependencies": {
     "@remix-run/serve": "1.19.3",
-    "@types/cacache": "^15.0.0",
+    "@types/cacache": "^17.0.0",
     "@types/gunzip-maybe": "^1.4.0",
     "@types/jsesc": "^3.0.1",
     "@types/lodash.debounce": "^4.0.6",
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
The Workflow is still failing because of the branch protection somehow 🤔
https://github.com/remix-run/remix/actions/runs/5857944230/job/15880905286

@mcansh Since you created the original PR, do you maybe have an idea on what should be done in order to make the workflow work?

mcansh:
i believe branch protection can be disabled for specific users, that's where i would check first

MichaelDeBoey:
@mcansh I don't think that's possible as the action itself isn't a user and you probably still need the PAT that way

mcansh:
the format workflow also runs on dev/main and appears to be able to commit without issue. it uses a custom token and commits as `remix-run-bot`, so we may need it to impersonate a user and bypass protection

MichaelDeBoey:
@mcanch yeah that one indeed seems to work as the PAT has `public_repo` permission, which is apparently able to bypass the "don't directly push to this branch" protection
Looked into the GitHub workflow docs and the only thing I could see is setting the `content` permission to `write` (which is just like you did and is currently the case)

Issue #6106, opened 2023-04-19
Don't treat a route as a resource route if it only has an ErrorBoundary


brophdawg11:
### What version of Remix are you using?

1.15.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

```jsx
import {
  Outlet,
  isRouteErrorResponse,
  useRouteError,
} from "@remix-run/react";

// this is required to make this a UI route, but we should be able to 
// default to this if no default is exported but a boundary is
// export default Outlet;

export function ErrorBoundary() {
  let error = useRouteError();
  // ...
}
```

### Expected Behavior

Any route exporting a default component, `ErrorBoundary`, or `CatchBoundary` is considered a UI route.

### Actual Behavior

Only routes exporting a default component are considered a UI route.

kiliman:
Interesting. I wonder what scenarios you would want to export an `ErrorBoundary` but not a `default` UI component.

To me, I think it should be an error.

brophdawg11:
The use-case I've seen is a singular top level error boundary in a pathless route below root - so it can still access valid loader data from the root loader (i18n info, etc.).  I think that might be the use-case Ryan was doing when he ran into it today.

PR #7155, opened 2023-08-14
Fix false-positive resource route identification on doc requests

brophdawg11:
This adds the corresponding logic from #6125 (closing #6106) for document requests.  A resource route should be a route that exports _no_ UI components, not just one that doesn't export a happy path UI component.

This way it's possible to do a self-contained splat route 404 page.  Previously, the below would be interpreted as a resource route and just return the raw text as a response and ignore the `ErrorBoundary`.

```jsx
// $.jsx
import { useRouteError } from "react-router-dom";

export async function loader({ request }) {
  throw new Response(`${new URL(request.url).pathname} not found`, {
    status: 404,
    statusText: "Not Found",
  });
}

export function ErrorBoundary() {
  let error = useRouteError();
  return (
    <>
      <h1>
        {error.status} {error.statusText}
      </h1>
      <p>{error.data}</p>
    </>
  );
}
```

### DIFF
--- a/.changeset/resource-route-boundary-only.md
+++ b/.changeset/resource-route-boundary-only.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/server-runtime": patch
+---
+
+Fix false-positive resource route classification on document requests for routes that only export an `ErrorBoundary`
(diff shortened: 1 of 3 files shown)


brophdawg11:
If an `ErrorBoundary` is specified and a default export is not, then it defaults the route's happy path UI component to `Outlet` - just like `<Route path="something">` would - https://github.com/remix-run/remix/blob/main/packages/remix-react/components.tsx#L132

PR #7157, opened 2023-08-14
Docs Tutorial: params.slug could be undefined

alexmuzenhardt:
Typescripts complains that params.slug could be string | undefined. A check would change this.
I also put the await getPost(params.slug) directly inside the object of the json-function. In the previous example we did this too and this would increase consistency.


alexmuzenhardt:
Sometimes I should continue first, before creating pull requests. Sorry. :)

PR #7162, opened 2023-08-14
docs(tutorials/jokes): update to be compatible with `v2_dev`

MichaelDeBoey:


docs/tutorials/jokes.md
```diff
@@ -1544,30 +1544,32 @@ So we've got a bit of extra work to do to avoid this development time problem.
...
+export const singleton = <Value>(
+  name: string,
+  valueFactory: () => Value
+): Value => {
+  const g = global as any;
+  g.__singletons ??= {};
+  g.__singletons[name] ??= valueFactory();
+  return g.__singletons[name];
+};
+```
```
pcattori, line 1563: We now have [docs for how to do this](https://remix.run/docs/en/main/guides/manual-mode#keeping-in-memory-server-state-across-rebuilds), so could you update this code to be consistent with that and maybe link to those docs instead?

pcattori (changes requested):

docs/tutorials/jokes.md
MichaelDeBoey: This is the same as we have in `indie-stack`, where I just took over whatever @jenseng had in his code + adapted it to accept a function (just like Kent did in his personal website and the epic stack)

I would argue to bring the docs more in line with what we have in there tbh

pcattori: Ok, let's update the docs to use `singleton` as defined here, but then link to those docs here since the docs will have a more detailed explanation.

docs/tutorials/jokes.md
```diff
@@ -1542,32 +1542,30 @@ This works just fine, but the problem is, during development, we don't want to c
 
 So we've got a bit of extra work to do to avoid this development time problem.
 
-Note that this isn't a remix-only problem. Any time you have "live reload" of server code, you're going to have to either disconnect and reconnect to databases (which can be slow) or do the workaround I'm about to show you.
+Note that this isn't a remix-only problem. Any time you have "live reload" of server code, you're going to have to either disconnect and reconnect to databases (which can be slow) or do the workaround [that's in our docs][singleton-workaround-docs].
```
pcattori, line 1545:
```suggestion
Note that this isn't a remix-only problem. Any time you have "live reload" of server code, you're going to have to either disconnect and reconnect to databases (which can be slow) or [use the `global` singleton workaround][singleton-workaround-docs].
```

pcattori (changes requested):

docs/tutorials/jokes.md
```diff
@@ -1542,32 +1542,30 @@ This works just fine, but the problem is, during development, we don't want to c
...
-} else {
-  if (!global.__db__) {
-    global.__db__ = new PrismaClient();
-  }
-  db = global.__db__;
-  db.$connect();
-}
+// Hard-code a unique key, so we can look up the client when this module gets re-imported
+const db = singleton("prisma", () => new PrismaClient());
+db.$connect();
```
kentcdodds, line 1568: Put the connect call within the callback, otherwise you'll call `$connect` on every file change which is probably not good.

Or just remove the connect call altogether. It's probably not necessary anyway.

kentcdodds (approved):
LGTM!

pcattori (approved):

PR #7166, opened 2023-08-15
tests: bring fixtures in line with templates

MichaelDeBoey:


brophdawg11:
To make this simpler in the future, I added a `scripts/copy-templates-to-fixtures.sh` script to do this and ran it - which brought along a few other small changes in [13aafdd](https://github.com/remix-run/remix/pull/7166/commits/13aafddb18fa3a76be6f08b78db21bdf3a5d7ec1).  Any concerns with that approach?

MichaelDeBoey:
@brophdawg11 I think we should keep this manual as we'll always have subtle small changes (like having the `0.0.0-local-version` or like we now have the problem where having `links` in the Node template fixture makes it crash, ...

brophdawg11:
Something in a fixture is causing a crash but it's not failing any tests?  That feels like it should cause a test failure...?  

My personal stance is that large-scale changes like this that are done manually worry me since it's _easy_ to mess it up manually and _hard_ to detect in a PR since the diff is so large (a bad combination).  I think some form of automated copy or some form of being able to read directly from templates in tests would make it much easier to handle these and get them merged in the future.

MichaelDeBoey:
> Something in a fixture is causing a crash but it's not failing any tests?

@brophdawg11 It's not because I commented it out (for now)
See `packages/remix-dev/_tests_fixtures/node/app/routes/_index.tsx`'s `links`

MichaelDeBoey:
@brophdawg11 It seems like we had a bug that I just discovered due to these updates
@markdalgleish fixed it in #7233

brophdawg11:
Am I understanding this correctly?

* `integration/helpers/node-template/app/root.tsx` `links` _used to_ have to be commented out due to the bug fixed by #7233.  If not commented out it would have crashed/caused a test failure?
* Now that #7233 is fixed - it no longer causes a crash, and that's why it no longer needs to be commented out and the tests are all passing?

> subtle small changes like having the 0.0.0-local-version or actual latest version

Is this actually a problem?  The unit tests don't care about the version since they're running off local repo code anyway.

If I run the copy script, I see 3 main things that get updated:

* `meta` functions are aligned - this is 👍 
* `@remix-run/*` deps go to `*` versions - I do not thinkg this has any impact on the unit tests, but please correct me if I'm wrong
* Deno's fixture handler creation moves from `mode: build.mode` to `mode: Deno.env.get("NODE_ENV")` - which also seems 👍 to me and doesn't cause failures

I'm just not seeing any reason we want to only bring these 95% in line instead of 100% in-line with templates if the actual templates being used don't cause any issues?

And beyond that - I'm half wondering why our uint etsts don't just run against the templates and why we can';t delete the fixtures entirely - but that's a larger question that's beyond the scope here I think.

MichaelDeBoey:
> * `integration/helpers/node-template/app/root.tsx` `links` _used to_ have to be commented out due to the bug fixed by #7233.  If not commented out it would have crashed/caused a test failure?
> * Now that #7233 is fixed - it no longer causes a crash, and that's why it no longer needs to be commented out and the tests are all passing?

@brophdawg11 Correct

> Is this actually a problem? The unit tests don't care about the version since they're running off local repo code anyway.

I presume it is, I wouldn't know why it's otherwise set to `0.0.0-local-version` tbh 🤷‍♂️

> * `@remix-run/*` deps go to `*` versions - I do not thinkg this has any impact on the unit tests, but please correct me if I'm wrong

Same reasoning as the `0.0.0-local-version` versions here

> * Deno's fixture handler creation moves from `mode: build.mode` to `mode: Deno.env.get("NODE_ENV")` - which also seems 👍 to me and doesn't cause failures

Seems like I missed updating that one in the Deno template in #7089 🙈

> I'm just not seeing any reason we want to only bring these 95% in line instead of 100% in-line with templates if the actual templates being used don't cause any issues?

I updated the script to also copy the templates into the `integration/helpers` folder
Things that pop up extra about the integration fixtures that aren't in templates:
- scripts that aren't linked to `node_modules` properly
- eslint is removed
- `...global.INJECTED_FIXTURE_REMIX_CONFIG,` is removed from `remix.config.js`
- node fixture has `@vanilla-extract/css` as extra dependency
  Should this be handled in the tests itself? cc/ @markdalgleish 

> And beyond that - I'm half wondering why our uint etsts don't just run against the templates and why we can';t delete the fixtures entirely - but that's a larger question that's beyond the scope here I think.

That's indeed beyond the scope of this ticket
But using the template folders for all fixture testing (in both `@remix-run/dev` & integration tests) is definitely something we can look at in a next PR I think

markdalgleish:
>node fixture has @vanilla-extract/css as extra dependency
>Should this be handled in the tests itself?

@MichaelDeBoey I guess we could manually define a custom package.json in the Vanilla Extract test?

MichaelDeBoey:
@brophdawg11 @markdalgleish I added `package.json` to the tests that needed `@vanilla-extract/css` and removed it from the fixture

The other problems I raised are still not fixed though 🤷‍♂️

brophdawg11:
@MichaelDeBoey yeah I actually wouldn't try to tackle integration test fixtures in this PR or with the script since those are actual running apps so commands/`package.json` stuff is important, whereas in unit tests we're not installing dependencies so the `package.json` stuff is effectively irrelevant.  I confirmed with @pcattori that the `local` version numbers are not important for unit test fixtures.

Can we put this back to just a script to update unit tests fixtures and get that merged, and tackle any updates to integration test fixtures in a separate PR?

MichaelDeBoey:
> Can we put this back to just a script to update unit tests fixtures and get that merged, and tackle any updates to integration test fixtures in a separate PR?

This PR was always about both updating integration + unit test fixtures

integration/package.json
```diff
@@ -1,3 +1,23 @@
...
+    "cross-env": "^7.0.3",
+    "express": "^4.17.1",
+    "isbot": "^3.6.8",
+    "npm-run-all": "^4.1.5",
+    "postcss": "^8.4.19",
+    "postcss-import": "^15.1.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "tailwindcss": "^3.3.0",
+    "typescript": "^5.1.0"
```
brophdawg11, line 21: If something is in the root, do we need to dup it here?  I think its just `react`/`react-dom`/`typescript`

brophdawg11 (approved):
Just one question otherwise LGTM

integration/package.json
MichaelDeBoey: I wanted to make them explicit here as I think we should at some point refactor to make them devDeps in the appropriate packages and root `package.json` only needs things like ESLint, rollup, Prettier, ...

PR #7178, opened 2023-08-15
chore: fix `deduplicate-yarn` workflow

MichaelDeBoey:
Let's use the PAT again, as we know that will work 🤷‍♂️


.github/workflows/deduplicate-yarn.yml
```diff
@@ -22,6 +22,8 @@ jobs:
     steps:
       - name: ⬇️ Checkout repo
         uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.FORMAT_PAT }}
```
brophdawg11, line 26: If we don't need `permissions` above anymore let's remove that too

PR #7180, opened 2023-08-16
fix(remix-dev): fix server builds where serverBuildPath ends in .cjs

lpsinger:
Fix a bug that caused the server build file to be emitted into the assets directory if the value of `serverBuildPath` ended in `.cjs`.

The bug was due to the server-side portion of the Remix compiler only iterating over files ending in `.js` and `.mjs`, but not `.cjs`.

Note that this diff is smaller than it appears: the largest number of lines affected are in integration/compiler-mjs-cjs-output-test.ts, in which I increased the indentation of the test fixture code in order to parametrize it over file extension.

- [x] Docs
- [x] Tests

Testing Strategy:

I found that the existing integration test `compiler-mjs-output-test.ts` was nearly what I needed. I renamed it to `compiler-mjs-cjs-output-test.ts` and parametrized it over server build file extension.

Note that this PR both introduces the new test and fixes the bug. To reproduce the test, just revert the part of the diff that affects `packages/remix-dev/compiler/server/write.ts` and run the integration test.


lpsinger:
Rebased.

mcansh (approved):

integration/compiler-mjs-cjs-output-test.ts
```diff
@@ -0,0 +1,76 @@
...
+            devDependencies: {
+              "@remix-run/dev": "0.0.0-local-version",
+              "@types/react": "0.0.0-local-version",
+              "@types/react-dom": "0.0.0-local-version",
+              typescript: "0.0.0-local-version",
+            },
+            engines: {
+              node: ">=18.0.0",
+            },
+          }),
```
MichaelDeBoey, line 44: `node-template`'s `package.json` already has `"type": "module",`, so is there a specific reason why we need a custom `package.json`?

```suggestion
```

integration/compiler-mjs-cjs-output-test.ts
```diff
@@ -0,0 +1,76 @@
...
+            export default function Index() {
+              let data = useLoaderData();
+              return (
+                <div>
+                  {data}
+                  <Link to="/burgers">Other Route</Link>
+                </div>
+              )
+            }
+          `,
```
MichaelDeBoey, line 62: Since we're not checking against anything that's in here, I guess we can also just use `node-template`'s default `app/routes/_index.tsx`?

```suggestion
```

MichaelDeBoey (changes requested):

integration/compiler-mjs-cjs-output-test.ts
lpsinger: I was just using what was already there in this integration test. I agree, but that increases the scope of this PR. Is that alright with you?

integration/compiler-mjs-cjs-output-test.ts
lpsinger: Same comment as above... I agree, but that increases the scope of this PR. Would you like me to apply these patches?

integration/compiler-mjs-cjs-output-test.ts
MichaelDeBoey: Ah yeah now I see you just copy/pasted the other test and made some changes

Imo we can just remove these files as they're not really necessary for these tests 🤷‍♂️

integration/compiler-mjs-cjs-output-test.ts
```diff
@@ -0,0 +1,35 @@
+import { test, expect } from "@playwright/test";
+import * as fs from "node:fs";
+import * as path from "node:path";
+
+import { createFixtureProject, js, json } from "./helpers/create-fixture.js";
```
MichaelDeBoey, line 5:
```suggestion
import { createFixtureProject, js } from "./helpers/create-fixture.js";
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

brophdawg11 (approved):

lpsinger:
Were you going to backport this to 1.x?

brophdawg11:
I wasn't planning on it - there are a _lot_ of bug fixes included in v2 and we can't backport them all to v1.  If there's a compelling reason to we could consider it but otherwise I think this just qualifies as a bug fixed in v2?  It's also worth noting that this didn't make the `2.0.0` release branch, so will land in `2.0.1`/`2.1.0` - whatever comes next.

mcansh:
> 🤖 Hello there,
> 
> We just published version `2.0.1` which includes this pull request. If you'd like to take it for a test run please try it out and let us know what you think!
> 
> Thanks!

really wish git log behaved the same in github actions as it does locally 🥲

PR #5529, opened 2023-02-21
test: dog food `@remix-run/testing`

mcansh:
Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

PR #7186, opened 2023-08-16
Dogfood @remix-run/testing on internal tests

### LINKED ISSUE #5528: dog food `@remix-run/testing`
TODO: look into act warnings

```log
console.error
      Warning: An update to RouterProvider inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at RouterProvider (/remix-run/remix/node_modules/react-router/lib/components.tsx:56:3)
          at RemixStub (/remix-run/remix/packages/remix-testing/dist/create-remix-stub.js:41:5)

```

brophdawg11:
Updated implementation of https://github.com/remix-run/remix/pull/5529 to dog food the `@remix-run/testing` package so we can stabilize it.  This PR moves `integration/meta-test.tsx` to `packages/remix-react/__tests__/integration/meta-test.tsx` and changes it from playwright tests to `createRemixStub` tests

```
yarn test:primary packages/remix-react/__tests__/integration/meta-test.tsx
```

This makes some changes to `createRemixStub` to permit `meta`/`links` implementations on routes.  It also combines the `manifest`/`routeModules` generation into the `patchRoutesWithContext` function (renamed `processRoutes`) so we just walk the route tree one time and generate everything.

I think this is a good starting point to get merged as an example, and then we can keep converting other modules over as we find time.

Notes:
* Internal testing via `@remix-run/testing` should be mostly (entirely?) for `packages/remix-react` so I'm just putting tests in `packages/remix-react/__tests__/integration` for now.  This allows our current `jest` setup to pick them up - simpler is better at this stage IMO.
* No renaming yet of `integration -> e2e` but once we're comfortable I think it's worth doing that and moving `__tests__/integration -> __integration__` in a follow up PR.

Replaces #5529 
Closes #5528


packages/remix-react/__tests__/integration/meta-test.tsx
```diff
@@ -0,0 +1,316 @@
...
+        Component: Meta,
+      },
+    ]);
+
+    let { container } = render(<RemixStub />);
+
+    // For some reason, prettyDOM strips the script tag (maybe because of
+    // dangerouslySetInnerHTML), so we just parse the HTML out into JSON and assert that way
+    let scriptTagContents =
+      container.querySelector("script")?.innerHTML || "{}";
```
markdalgleish, line 240: Should this test also ensure the correct `type` attribute is present?

```suggestion
      container.querySelector('script[type="application/ld+json"]')
        ?.innerHTML || "{}";
```

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -141,64 +131,73 @@ export function createRemixStub(
...
+      module: "", // any need for this?
+    };
+    manifest.routes[newRoute.id] = entryRoute;
 
-    modules[route.id!] = {
-      ErrorBoundary: undefined,
-      default: () => route.element,
+    // Add the route to routeModules
+    routeModules[route.id!] = {
+      ErrorBoundary: route.ErrorBoundary || undefined,
```
markdalgleish, line 183: Does this need to fall back to using `route.errorElement` too?

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -141,64 +131,73 @@ export function createRemixStub(
...
+
+    // Add the EntryRoute to the manifest
+    let entryRoute: EntryRoute = {
+      id: route.id,
+      path: route.path,
+      index: route.index,
+      parentId,
+      hasAction: route.action != null,
+      hasLoader: route.loader != null,
+      hasErrorBoundary: route.ErrorBoundary != null,
```
markdalgleish, line 176: Should this check `route.errorElement`?

packages/remix-testing/create-remix-stub.tsx
```diff
@@ -41,98 +77,52 @@ type RemixStubOptions = {
...
+          url: "",
+          version: "",
+        },
+        routeModules: {},
+      };
 
+      // Update the routes to include context in the loader/action and populate
+      // the manifest and routeModules during the walk
+      let patched = processRoutes(
+        // @ts-expect-error loader/action context types don't match :/
```
markdalgleish, line 113: Why don't they match?

brophdawg11: Remix forces a key/value context object in TS:

```ts
interface AppLoadContext {
  [key: string]: unknown;
}

interface DataFunctionArgs {
  request: Request;
  context: AppLoadContext;
  params: Params;
}
```

But React Router is more lenient since you can realistically pass anything through, it doesn't need to be key/value:

```ts
interface DataFunctionArgs {
  request: Request;
  params: Params;
  context?: any;
}
```

PR #7187, opened 2023-08-16
fix(dev): improve minified CSS compatibility

### LINKED ISSUE #6114: esbuild minifies CSS in a way that can break in old browsers

@remix-run/dev 1.15.0

- [X] Yes

### Steps to Reproduce

I found this compilation of styles.

image

when i use postcss-preset-env compile my style code, eg: `body { position: fixed; inset: 0; }`

First, it will be processed by postcss,get `body { position: fixed; top:0; bottom:0; left:0; right:0; } `. 

Then esbuild will process my style code,This will translate into `body { position: fixed; inset:0; }`

### Expected Behavior

Keep The postcss-preset-env dealing with me.

### Actual Behavior

Now, then esbuild will process my css code after then postcss.

markdalgleish:
Fixes #6114

As the changeset says, we now minify CSS targeting the Remix browser support baseline (i.e. anything that supports [ES module scripts](https://caniuse.com/es6-module)), whereas previously CSS was minified using esbuild's default "esnext" target. This meant that CSS in the production build could be using properties that are unsupported in some browsers.

This change also means you now have more control over CSS transformations using the built-in PostCSS support since they won't be reversed by esbuild's minification.

### DIFF
--- a/packages/remix-dev/compiler/css/bundle.ts
+++ b/packages/remix-dev/compiler/css/bundle.ts
@@ -6,9 +6,9 @@ import postcssDiscardDuplicates from "postcss-discard-duplicates";
 
 import type { Context } from "../context";
 
-export let write = async (ctx: Context, outputFiles: esbuild.OutputFile[]) => {
+export async function write(ctx: Context, outputFiles: esbuild.OutputFile[]) {
   let cssBundleFile = outputFiles.find((outputFile) =>
-    isBundle(ctx, outputFile, ".css")
+    isCssBundleFile(ctx, outputFile, ".css")
   );
   if (!cssBundleFile) return;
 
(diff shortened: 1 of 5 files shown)


packages/remix-dev/compiler/css/bundle.ts
```diff
@@ -45,16 +45,45 @@ export let write = async (ctx: Context, outputFiles: esbuild.OutputFile[]) => {
...
+): boolean {
   return (
     path.dirname(outputFile.path) === ctx.config.assetsBuildDirectory &&
     path.basename(outputFile.path).startsWith("css-bundle") &&
     outputFile.path.endsWith(extension)
   );
-};
+}
+
+interface GroupedCssBundleFiles {
```
MichaelDeBoey, line 62: This can just be a `type` I guess?

```suggestion
type GroupedCssBundleFiles = {
```

MichaelDeBoey (changes requested):

MichaelDeBoey (approved):

pcattori (approved):

PR #7198, opened 2023-08-20
fix(dev): support polyfill package imports in deps

### LINKED ISSUE #5521: Remix esbuild browser build external builtinModules of node.js causes import AWS-sdk error

v1.13.0

- [X] Yes

### Steps to Reproduce

import AWS from "aws-sdk";

aws-sdk dependency [buffer](https://github.com/feross/buffer) and other package for the browser from nodejs buildinModule, Remix esbuild browser make it external causes bundle for browser error

### Expected Behavior

no error in browser

### Actual Behavior

In Browser
Uncaught Error: Dynamic require of "buffer/" is not supported

markdalgleish:
Fixes #5521

This error was happening because all Node built-ins were marked as external. This meant that imports for `"buffer/"` (which is supposed to force the import of the package rather than the Node built-in) would be left as-is in the client bundle.

I added a feature upstream to esbuild-plugins-node-modules-polyfill (https://github.com/imranbarbhuiya/esbuild-plugins-node-modules-polyfill/pull/146) which allows us to inject empty fallbacks instead. This fixes the issue because it only matches on exact import specifiers (i.e. `"buffer"` matches as a Node built-in, but `"buffer/"` doesn't).


packages/remix-dev/compiler/js/compiler.ts
```diff
@@ -33,23 +33,10 @@ type Compiler = {
...
   let fakeBuiltins = nodeBuiltins.filter((mod) => dependencies.includes(mod));
-
-  if (fakeBuiltins.length > 0) {
-    throw new Error(
-      `It appears you're using a module that is built in to node, but you installed it as a dependency which could cause problems. Please remove ${fakeBuiltins.join(
-        ", "
-      )} before continuing.`
-    );
-  }
-  return nodeBuiltins.filter((mod) => !dependencies.includes(mod));
```
markdalgleish, line 52: This filter was redundant because an error would have already been thrown if this was the case. This is because this filter is the inverse of `fakeBuiltins`.

packages/remix-dev/package.json
```diff
@@ -36,7 +36,7 @@
     "chokidar": "^3.5.1",
     "dotenv": "^16.0.0",
     "esbuild": "0.17.6",
-    "esbuild-plugins-node-modules-polyfill": "^1.3.0",
+    "esbuild-plugins-node-modules-polyfill": "^1.4.0",
```
markdalgleish, line 39: This introduces the new `fallback: "empty"` option: https://github.com/imranbarbhuiya/esbuild-plugins-node-modules-polyfill/pull/146

MichaelDeBoey (approved):

PR #7199, opened 2023-08-21
fix(dev): avoid Cloudflare manifest import warning

markdalgleish:
Consumers can avoid bundling Cloudflare's built-in `__STATIC_CONTENT_MANIFEST` module with the following Remix config which is part of the Cloudflare Workers template:

```ts
  serverDependenciesToBundle: [
    /^(?!.*\b__STATIC_CONTENT_MANIFEST\b).*$/,
  ],
```

However, they still receive a warning in the terminal telling them that they didn't install it because `require.resolve` can't find it. In the same way that we ignore Node built-in modules, this PR extends our logic to ignore this Cloudflare built-in module.

### DIFF
--- a/.changeset/silence-cloudfare-manifest-warnings.md
+++ b/.changeset/silence-cloudfare-manifest-warnings.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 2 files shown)


markdalgleish:
Closing in favour of skipping this check entirely: https://github.com/remix-run/remix/pull/7214

PR #7203, opened 2023-08-21
feat(remix-node): use Web Crypto API in `createFileSessionStorage`

MichaelDeBoey:
https://nodejs.org/docs/latest-v20.x/api/crypto.html#cryptogetrandomvaluestypedarray
https://nodejs.org/docs/latest-v20.x/api/crypto.html#cryptowebcrypto
https://nodejs.org/docs/latest-v20.x/api/webcrypto.html#cryptogetrandomvaluestypedarray

### DIFF
--- a/packages/remix-node/sessions/fileStorage.ts
+++ b/packages/remix-node/sessions/fileStorage.ts
@@ -42,7 +42,7 @@ export function createFileSessionStorage<Data = SessionData, FlashData = Data>({
       while (true) {
         // TODO: Once Node v19 is supported we should use the globally provided
         // Web Crypto API's crypto.getRandomValues() function here instead.
-        let randomBytes = crypto.randomBytes(8);
+        let randomBytes = crypto.webcrypto.getRandomValues(new Uint8Array(8));
         // This storage manages an id space of 2^64 ids, which is far greater
         // than the maximum number of files allowed on an NTFS or ext4 volume
         // (2^32). However, the larger id space should help to avoid collisions
--- a/.changeset/node-web-crypto.md
+++ b/.changeset/node-web-crypto.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/node": patch
+---
+
+Switch from `crypto.randomBytes` to `crypto.webcrypto.getRandomValues` for file session storage ID generation


packages/remix-node/sessions/fileStorage.ts
```diff
@@ -42,7 +42,7 @@ export function createFileSessionStorage<Data = SessionData, FlashData = Data>({
       while (true) {
         // TODO: Once node v16 is available on AWS we should use the webcrypto
         // API's crypto.getRandomValues() function here instead.
-        let randomBytes = crypto.randomBytes(8);
+        let randomBytes = crypto.webcrypto.getRandomValues(new Uint8Array(8));
```
sergiodxa, line 45: I don't understand why Node put Web Crypto API under `crypto.webCrypto` instead of globally like in any other runtime

sergiodxa (approved):

packages/remix-node/sessions/fileStorage.ts
MichaelDeBoey: It will be under the global `crypto`, but only in Node v19
Since the new minimum version is v18, we need to do it like this unfortunately 😢

sergiodxa: That's nice to hear, so far I couldn't use WebCrypto in code that's not dependent of the runtime (some remix-auth strategies) because how to access the API depends if it's Node or basically anything else where it's global.

MichaelDeBoey: I updated the comments in our code to point to Node v19 in #7204, so we at least know that we can use the global `crypto` once we drop support for Node 18

packages/remix-node/sessions/fileStorage.ts
```diff
@@ -42,7 +42,7 @@ export function createFileSessionStorage<Data = SessionData, FlashData = Data>({
       while (true) {
         // TODO: Once Node v19 is supported we should use the globally provided
         // Web Crypto API's crypto.getRandomValues() function here instead.
```
brophdawg11, line 44: Can this go away now?

brophdawg11: oh nvm I see this is talking about node 19, not 18 🤦

PR #7204, opened 2023-08-21
chore: update Web Crypto API references to point to Node v19

MichaelDeBoey:
https://nodejs.org/docs/latest-v20.x/api/globals.html#crypto_1


packages/remix-server-runtime/crypto.ts
```diff
@@ -5,59 +5,57 @@ export type UnsignFunction = (
...
 
-// export async function sign(value: string, secret: string): Promise<string> {
-//   let key = await crypto.subtle.importKey(
-//     "raw",
-//     encoder.encode(secret),
-//     { name: "HMAC", hash: "SHA-256" },
-//     false,
-//     ["sign"]
-//   );
-
```
brophdawg11, line 22: Why move from `async function sign` to an arrow function?  I personally think the industry got too enamored with arrow functions.  If it's not an tiny/inline function, a predicate, or using `this` - I strongly prefer `function`.

MichaelDeBoey: I just copied over what we have in Cloudflare & Deno 🤷‍♂️

PR #7207, opened 2023-08-21
chore: update `@architect/functions`

MichaelDeBoey:


brophdawg11:
I don't know much about architect - but if this is a major version upgrade of the dep - does that mean breaking changes to the Remix surface area?

lpsinger:
> I don't know much about architect - but if this is a major version upgrade of the dep - does that mean breaking changes to the Remix surface area?

No.

Although the latest version of this dependency is now 8.0.0. @MichaelDeBoey, would you please update your PR to @architect/functions 8.0.0?

MichaelDeBoey:
@lpsinger I've updated this PR to use latest version of `@architect/functions` now

lpsinger:
Could you change it to 7.x? @architect/functions 8.x switched from using [the official AWS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/) to their own [aws-lite](https://aws-lite.org) client, which I am skeptical of.

MichaelDeBoey:
@lpsinger Can you explain why you're skeptical about the change?
The architect team seems to recommend it now, so I think we should update to their recommendations as well

Do you think those changes can break existing Remix apps?

lpsinger:
aws-lite is supposed to have quicker cold-start times than @aws-sdk/*, but all but the simplest projects are bound to have dependencies on @aws-sdk/* anyway through other projects that depend upon it. Since now we have to spin up both clients, aws-lite is going to make our cold-start times strictly longer.

Also, Architect has made some changes that may have dangerous consequences for my project (see https://github.com/architect/architect/issues/1483#issuecomment-2096692771), and we are stuck with using an old version of Architect until those issues are fixed.

If Remix requires @architect/functions 8.x, then my project may not be able to track the latest version of  Remix either.

MichaelDeBoey:
@lpsinger I've reverted to v7.x again 👍

lpsinger:
Thank you!

MichaelDeBoey:
Superseded by https://github.com/remix-run/react-router/pull/13556

PR #7202, opened 2023-08-21
chore(templates/remix): esm by default

pcattori:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

PR #7213, opened 2023-08-21
templates(remix): ESM by default

MichaelDeBoey:
Cherry-pick @pcattori's #7202 so this change is in latest templates already


pcattori:
We don't want ESM to be the default until v2, so this was left out of `main` intentionally.

MichaelDeBoey:
@pcattori We did enable it for other templates though

pcattori:
Yes, but this one is special because to flip to ESM, it requires `remix-serve` to have ESM support out-of-the-box, which is part of the v2 work.

PR #7215, opened 2023-08-21
refactor(deno): fix Deno lint errors

markdalgleish:
There seems to be a recently introduced bug in Deno's lint command where async method definitions aren't detected when validating usage of `await`. This works around the issue.


jacob-ebey (approved):

jacob-ebey:
https://github.com/denoland/deno/issues/20229

PR #7219, opened 2023-08-22
refactor(remix-node): use native Stream API

MichaelDeBoey:
- https://nodejs.org/docs/latest-v20.x/api/webstreams.html#class-readablestream
- https://nodejs.org/docs/latest-v20.x/api/webstreams.html#class-writablestream


brophdawg11:
We can't rely on these until they're stable 😕 

https://nodejs.org/docs/latest-v20.x/api/webstreams.html#web-streams-api

PR #7220, opened 2023-08-22
chore: update Jest

MichaelDeBoey:
Without this update #7217 & #7219 will keep failing

---

CC/ @SimenB to make sure we're doing what we're supposed to do

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,4 @@
-/** @type {import('@jest/types').Config.InitialOptions} */
+/** @type {import('jest').Config} */
 module.exports = {
   modulePathIgnorePatterns: [
     "<rootDir>/.tmp",
@@ -23,10 +23,13 @@ module.exports = {
     "packages/remix-server-runtime",
     "packages/remix-testing",
   ],
+  reporters:
+    process.env.GITHUB_ACTIONS === null
+      ? ["default"]
+      : [["github-actions", { silent: false }], "summary"],
   watchPlugins: [
     require.resolve("jest-watch-select-projects"),
     require.resolve("jest-watch-typeahead/filename"),
     require.resolve("jest-watch-typeahead/testname"),
   ],
-  reporters: ["default"],
 };
--- a/jest/jest.config.shared.js
+++ b/jest/jest.config.shared.js
@@ -5,16 +5,26 @@ const ignorePatterns = [
   "\\/\\.tmp\\/",
   "\\/\\.cache\\/",
 ];
-/** @type {import('@jest/types').Config.InitialOptions} */
+
+/** @type {import('jest').Config} */
 module.exports = {
-  testEnvironment: "node",
+  moduleNameMapper: {
+    "^@remix-run/web-blob$": require.resolve("@remix-run/web-blob"),
+    "^@remix-run/web-fetch$": require.resolve("@remix-run/web-fetch"),
+    "^@remix-run/web-file": require.resolve("@remix-run/web-file"),
+    "^@remix-run/web-form-data$": require.resolve("@remix-run/web-form-data"),
+    "^@remix-run/web-stream$": require.resolve("@remix-run/web-stream"),
+    "^@web3-storage/multipart-parser$": require.resolve(
+      "@web3-storage/multipart-parser"
+    ),
+  },
   modulePathIgnorePatterns: ignorePatterns,
-  watchPathIgnorePatterns: [...ignorePatterns, "\\/node_modules\\/"],
-  testMatch: ["<rootDir>/**/*-test.[jt]s?(x)"],
   setupFiles: ["<rootDir>/__tests__/setup.ts"],
+  testMatch: ["<rootDir>/**/*-test.[jt]s?(x)"],
   transform: {
     "\\.[jt]sx?$": require.resolve("./transform"),
   },
+  watchPathIgnorePatterns: [...ignorePatterns, "\\/node_modules\\/"],
   watchPlugins: [
     require.resolve("jest-watch-select-projects"),
     require.resolve("jest-watch-typeahead/filename"),
--- a/packages/create-remix/__tests__/create-remix-test.ts
+++ b/packages/create-remix/__tests__/create-remix-test.ts
@@ -242,7 +242,7 @@ describe("create-remix CLI", () => {
     });
 
     expect(stderr.trim()).toMatchInlineSnapshot(
-      `"▲  Oh no! The path \\"this/path/does/not/exist\\" was not found in this GitHub repo."`
+      `"▲  Oh no! The path "this/path/does/not/exist" was not found in this GitHub repo."`
     );
     expect(status).toBe(1);
     expect(fse.existsSync(path.join(projectDir, "package.json"))).toBeFalsy();
@@ -381,7 +381,7 @@ describe("create-remix CLI", () => {
     });
 
     expect(stderr.trim()).toMatchInlineSnapshot(
-      `"▲  Oh no! The path \\"this/path/does/not/exist\\" was not found in this GitHub repo."`
+      `"▲  Oh no! The path "this/path/does/not/exist" was not found in this GitHub repo."`
     );
     expect(status).toBe(1);
     expect(fse.existsSync(path.join(projectDir, "package.json"))).toBeFalsy();
@@ -1086,12 +1086,12 @@ describe("create-remix CLI", () => {
         });
 
         expect(stderr.trim()).toMatchInlineSnapshot(`
-        "▲  Oh no! Destination directory contains files that would be overwritten
-                 and no \`--overwrite\` flag was included in a non-interactive
-                 environment. The following files would be overwritten:
-                   package.json
-                   tsconfig.json"
-      `);
+                  "▲  Oh no! Destination directory contains files that would be overwritten
+                           and no \`--overwrite\` flag was included in a non-interactive
+                           environment. The following files would be overwritten:
+                             package.json
+                             tsconfig.json"
+              `);
         expect(status).toBe(1);
         expect(
           fse.existsSync(path.join(projectDir, "app/root.tsx"))
--- a/package.json
+++ b/package.json
@@ -49,6 +49,9 @@
     "watch": "rollup -c --watch --watch.onEnd=\"node scripts/copy-build-to-dist.mjs\"",
     "postinstall": "patch-package"
   },
+  "resolutions": {
+    "jsdom": "^22.0.0"
+  },
   "dependencies": {
     "@babel/core": "^7.21.8",
     "@babel/plugin-proposal-export-namespace-from": "^7.18.9",
@@ -74,7 +77,7 @@
     "@types/cross-spawn": "^6.0.2",
     "@types/dedent": "^0.7.0",
     "@types/glob": "7.2.0",
-    "@types/jest": "^27.4.1",
+    "@types/jest": "^29.5.4",
     "@types/jsonfile": "^6.1.0",
     "@types/lodash": "^4.14.182",
     "@types/react": "^18.2.20",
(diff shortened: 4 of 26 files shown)


SimenB:
Nothing here jumps out to me. A bit odd removing `require.resolve` fixes things, but shouldn't hurt.

---

Any particular reason why you don't go to v29? I'd be happy to help upgrading (I actually sent a PR at work 6 hours ago introducing `react-router` into the code base 😅)

jest/jest.config.shared.js
```diff
@@ -15,9 +15,12 @@ module.exports = {
   transform: {
     "\\.[jt]sx?$": require.resolve("./transform"),
   },
+  transformIgnorePatterns: [
+    "node_modules/(?!(@remix-run/web-(blob|fetch|file|form-data|stream)|@web3-storage/multipart-parser)/)",
```
MichaelDeBoey, line 19: We will be able to remove the `@remix-run/web-*` dependencies here once https://github.com/remix-run/web-std-io/pull/43 is merged

MichaelDeBoey:
@SimenB My plan was to first upgrade to v28 and make that work, but I can go straight to v29 as well if you think that would make it easier for you

SimenB:
I'm somehow surprised there's a difference. 😅 Breaking changes 28->29 (beyond snapshot format changes, which are configurable) were chosen as they were negligible.

MichaelDeBoey:
@SimenB I just didn't look into upgrading to v29 yet 🤷‍♂️
So if you say there's no _real_ difference, I'll just go ahead and upgrade to v29 👍

SimenB:
Ah! Fair enough enough 🙂 happy to take a look if it proves troublesome - feel free to tag me 👍

MichaelDeBoey:
@SimenB I guess the same issues with read-only globals will still be present in v29 so we still need to find a solution for that?

MichaelDeBoey:
@SimenB we're now getting the following warning

> ● Validation Warning:
> 
>   Unknown option "watchPlugins" with value ["jest-watch-select-projects", "jest-watch-typeahead/filename", "jest-watch-typeahead/testname"] was found.
>   This is probably a typing mistake. Fixing it will remove this message.
> 
>   Configuration Documentation:
>   https://jestjs.io/docs/configuration

Looking at the docs, I don't see what we're doing wrong tbh as `watchPlugins` is just there 🤔
https://jestjs.io/docs/configuration#watchplugins-arraystring--string-object

Next to that we're still getting the same read-only errors as we already had in v28

> TypeError: Cannot assign to read only property 'atob' of object '[object global]'

Together with the non-constructor error (which is probably because of the globals not being set)

> TypeError: Response is not a constructor

SimenB:
>Unknown option "watchPlugins" 

sounds like https://github.com/jestjs/jest/issues/13576 🙁 should probably dig deeper into it and figure out the root cause.

>Next to that we're still getting the same read-only errors as we already had in v28

_probably_ https://github.com/jestjs/jest/pull/12642 (currently https://github.com/jestjs/jest/blob/642267f6848869a64861c3176eb5a6d5980639c6/packages/jest-environment-node/src/index.ts#L79-L128)

both `atob` and `btoa` are already globals in Node since v16, so I don't think you need to polyfill them at all? it should at least be done conditionally

MichaelDeBoey:
> both `atob` and `btoa` are already globals in Node since v16, so I don't think you need to polyfill them at all? it should at least be done conditionally

@SimenB We'll remove the `atob` & `btoa` polyfills in #7206, but if people decide to override a global it should just be possible, no?

SimenB:
I'm unable to reproduce the override issue... Works fine for mw with both JSDOM and Node envs.

I'm also not seeing that issue running `yarn test:primary` on my own machine.

image

Is there any other command I must run to reproduce it?

MichaelDeBoey:
> Is there any other command I must run to reproduce it?

@SimenB No, that's the exact command I run locally as well

> I'm also not seeing that issue running `yarn test:primary` on my own machine.

Just tried it out myself as wel locally and indeed not seeing the read-only message anymore when using Node v18
I tried using Node v20 ([we're using `latest` in our workflow](https://github.com/remix-run/remix/blob/main/.github/workflows/test.yml#L25)) and got the errors showing up again

However, that makes the `TypeError: Response is not a constructor` error in Node v18 even stranger to me as the only changes in this PR are about Jest 🤔

> note that `fetch` is also available globally in newer versions of node - including `Response` etc

The team has decided to still polyfill them (for now) as Fetch API is still marked as `Experimental` in the Node docs

jenseng:
Another datapoint in favor of going straight to jest 29... [I upgraded react-router](https://github.com/remix-run/react-router/pull/10453) to use it a few months back and didn't run into any issues 🤞

MichaelDeBoey:
@jenseng This PR is already using v29, but still has issues for some reason 🤔

jenseng:
Ok I may have tracked it down... `Request` and friends are coming through as `undefined`, thus the error when you do `new ThingThatIsDeclaredButUndefined()`... here's what's happening:
1. `jest.setup.js` tries to install `Request` and friends via `installGlobals`
2. `installGlobals` gets them from `@remix-run/web-fetch`
3. `@remix-run/web-fetch` is getting imported its the `browser` conditional export
    ```json
    {
      "name": "@remix-run/web-fetch",
      // ...
      "exports": {
        ".": {
          "types": "./dist/src/lib.node.d.ts",
          "browser": "./src/lib.js", // <- it's picking this up 🤷 
          "require": "./dist/lib.node.cjs",
          "import": "./src/lib.node.js"
        },
        // ...
    ```
4. [./src/lib.js](https://github.com/remix-run/web-std-io/blob/main/packages/fetch/src/lib.js) re-exports the global `fetch`/`Request`/`Response`, since they should be available in a browser. But they are undefined in this case.

As for why jest is picking up the `browser` conditional export, perhaps @SimenB would have a better idea? Maybe the `require` one should come first?

MichaelDeBoey:
@jenseng Jest is only choosing the `browser` export when `testEnvironment` is set to `'jsdom'`, which makes sense imo

Although https://github.com/remix-run/web-std-io/pull/43 will have some changes to the `browser` field, I would still suspect that even if we're overriding the globals with the globals that they would be present as we're only testing with Node v20 in our Jest tests, which is a version where these globals are already present by default
So this would mean that we're just overriding these globals with themselves, which should be fine, no?

jenseng:
> So this would mean that we're just overriding these globals with themselves, which should be fine, no?

I've confirmed that `global.Request` and friends are undefined at this point in ./src/lib.js. So perhaps jsdom isn't actually loaded yet at this point?

MichaelDeBoey:
> I've confirmed that `global.Request` and friends are undefined at this point in ./src/lib.js. So perhaps jsdom isn't actually loaded yet at this point?

Then I have no idea what's happening tbh 🤷‍♂️
Hopefully @SimenB can give us some more insights on what's happening 🤔

SimenB:
JSDOM does not implement `fetch` (https://github.com/jsdom/jsdom/issues/1724), so it'll only be available globally in the Node environment.

---

I'm working on the `atob` override in v20 - that's a bug in Jest.

SimenB:
@MichaelDeBoey please give https://github.com/jestjs/jest/releases/tag/v29.6.4 a try

MichaelDeBoey:
> JSDOM does not implement `fetch` ([[jsdom/jsdom#1724]()](https://github.com/jsdom/jsdom/issues/1724)), so it'll only be available globally in the Node environment.

How would you suggest we go from here?
It's not unreasonable to use built-in functionality in the `browser` export, so nothing we can/should do on the `@remix-run/web-*` side imo

> please give https://github.com/jestjs/jest/releases/tag/v29.6.4 a try

Updated to latest version 👍

SimenB:
> It's not unreasonable to use built-in functionality in the `browser` export, so nothing we can/should do on the `@remix-run/web-*` side imo

Agreed - I'd manually polyfill fetch in the jest setup file

SimenB:
If you want the node polyfilling and not the browser ones, you can set up a module mapper using `require.resolve`

MichaelDeBoey:
@SimenB I'm personally not a fan of overriding the `exportConditions` I think, as that would mean all dependencies will use the `node` export instead of the `browser` one, right?
If so, this could hide some other bugs I suppose 🤔

So I guess `moduleNameMapper` is the only solution here?

SimenB:
least hacky is probably just

```
if (typeof window.fetch === 'undefined') {
  require('whatwg-fetch');
}
```

in your jest setup or something like that. I'd say that's the least likely to hide actual bugs

jest.config.js
```diff
@@ -23,10 +23,6 @@ module.exports = {
     "packages/remix-server-runtime",
     "packages/remix-testing",
   ],
-  watchPlugins: [
-    require.resolve("jest-watch-select-projects"),
-    require.resolve("jest-watch-typeahead/filename"),
-    require.resolve("jest-watch-typeahead/testname"),
-  ],
-  reporters: ["default"],
+  watchPlugins: ["select-projects", "typeahead/filename", "typeahead/testname"],
```
SimenB, line 26: any particular reason these are not `require.resolve` anymore?

jest/jest.config.shared.js
```diff
@@ -5,19 +5,26 @@ const ignorePatterns = [
...
+    "^@remix-run/web-stream$": require.resolve("@remix-run/web-stream"),
+    "^@web3-storage/multipart-parser$": require.resolve(
+      "@web3-storage/multipart-parser"
+    ),
+  },
   modulePathIgnorePatterns: ignorePatterns,
-  watchPathIgnorePatterns: [...ignorePatterns, "\\/node_modules\\/"],
-  testMatch: ["<rootDir>/**/*-test.[jt]s?(x)"],
   setupFiles: ["<rootDir>/__tests__/setup.ts"],
+  testEnvironment: "node",
```
SimenB, line 23: this is the default, fwiw

packages/create-remix/__tests__/create-remix-test.ts
```diff
@@ -242,7 +242,7 @@ describe("create-remix CLI", () => {
     });
 
     expect(stderr.trim()).toMatchInlineSnapshot(
-      `"▲  Oh no! The path \\"this/path/does/not/exist\\" was not found in this GitHub repo."`
```
SimenB, line 245: you can reduce the diff from the upgrade if you want by specifying `snapshotFormat`

jest.config.js
MichaelDeBoey: Just because the docs use string tbh
Is `require.resolve` preferred?

packages/create-remix/__tests__/create-remix-test.ts
MichaelDeBoey: The new defaults are fine imo

jest.config.js
SimenB: it's a bit faster as Jest doesn't have to do its own resolution

MichaelDeBoey: Reverted the change

jest.config.js
```diff
@@ -23,10 +23,10 @@ module.exports = {
     "packages/remix-server-runtime",
     "packages/remix-testing",
   ],
+  reporters: [["github-actions", { silent: false }], "summary"],
```
SimenB, line 26: note that you might want something like `reporters: process.env.GITHUB_ACTIONS == null ? undefined : [['github-actions', {silent: false}], 'summary'],`, otherwise you get no output when running tests locally. This is probably something we should fix in Jest...

SimenB (approved):

brophdawg11 (approved):

SimenB:
🎉

jest.config.js
```diff
@@ -23,10 +23,13 @@ module.exports = {
     "packages/remix-server-runtime",
     "packages/remix-testing",
   ],
+  reporters:
+    process.env.GITHUB_ACTIONS === null
```
SimenB, line 27: should be `==` - env variables that aren't set are `undefined`, not `null`

brophdawg11: Fixed in 517e40bae90ddec7a94a93f06c206cf93363210f

PR #7225, opened 2023-08-22
Improve Session API docs

aaronadamsCA:
Better clarify that it's only when using `cookieSessionStorage` that you need to commit a session after reading a flash value or unsetting a value.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,3 +1,4 @@
+- aaronadamsCA
 - aaronpowell96
 - aaronshaf
 - abereghici
--- a/docs/utils/sessions.md
+++ b/docs/utils/sessions.md
@@ -472,7 +472,7 @@ export async function action({
 
 Now we can read the message in a loader.
 
-<docs-info>You must commit the session whenever you read a `flash`. This is different than what you might be used to, where some type of middleware automatically sets the cookie header for you.</docs-info>
+<docs-info>When using `cookieSessionStorage`, you must commit the session after you `session.get()` a flash value. This is different than what you might be used to, where some type of middleware automatically sets the cookie header for you.</docs-info>
 
 ```tsx
 import { json } from "@remix-run/node"; // or cloudflare/deno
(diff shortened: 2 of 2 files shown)


brophdawg11 (approved):

aaronadamsCA:
Oops - now that the docs are already updated, @brophdawg11, I think I've noticed they are already wrong about sessions, and I've only made them worse with this change.

Could I run this by you before I actually fix this?

This part of the docs:

> With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).

Disagrees with the actual API designs:

- The `SessionIdStorageStrategy` interface defines async `createData()`, `updateData()`, and `deleteData()` methods.

- The `Session` interface is synchronous, so `get()`, `set()`, `flash()`, or `unset()` can't possibly be committing data, regardless of the session storage method. The source seems to confirm they only manipulate internal state data: https://github.com/remix-run/remix/blob/a873056de08be9b63def9567e6a7965d06a0f6c2/packages/remix-server-runtime/sessions.ts#L95-L142

- The `SessionStorage` interface, on the other hand, is asynchronous, and `getSession()`, `commitSession()`, and `destroySession()` appear to call the commit methods of the concrete session storage strategy: https://github.com/remix-run/remix/blob/a873056de08be9b63def9567e6a7965d06a0f6c2/packages/remix-server-runtime/sessions.ts#L263-L297

As such, I think the following doc changes are needed:

- Clarify that you always need to commit your changes after `get()` (for flash values), `set()`, `flash()`, or `unset()`. This is regardless of your session storage strategy.
- Clarify that when using `cookieSessionStorage`, every time you commit your changes you also need to update the cookie header. This is the actual difference when using `cookieSessionStorage`, as opposed to what the docs say now.

Happy to do this as long as I've finally got this right.

brophdawg11:
ok, I reverted this in `a2d91911e8236057fe4ef9a0dae03f90bf8d76c8` for now.  

To be totally honest, I'm not super familiar with the inner workings of our sessions APIs - Michael and Ryan mostly wrote them pre-v1 and they haven't needed to change much since.  I think what you're proposing generally sounds right but will probably look to get an official sign off from someone more familiar than me.  If you want to write up the proposed changes and tag me on the PR I'll get them in front of the right folks.

I'm also going to be digging into some open session-related bugs today so I oughta be able to provide a better answer after I've done that a bit :)

PR #7227, opened 2023-08-22
Update createMemorySessionStorage to use an internal hash for the id

brophdawg11:


MichaelDeBoey (approved):

ryanflorence:
In the future we plan to rename this to something more obvious it's only recommended for development and initial scaffolding. But we could also probably use the web crypto API for better unique IDs.

PR #7229, opened 2023-08-22
refactor!(serve): remove undocumented `createApp` node api

pcattori:
remix-serve should be a CLI, not a Node library

Testing Strategy:

Existing integration tests + tested locally with `LOCAL_BUILD_DIRECTORY`


packages/remix-serve/cli.ts
```diff
@@ -1,11 +1,19 @@
-import "./env";
```
pcattori, line 1: I inlined code from `./env` and `./index` into this file

jacob-ebey (approved):

PR #7231, opened 2023-08-22
feat(serve): manual mode integration

pcattori:
Testing Strategy:

Tested locally via `LOCAL_BUILD_DIRECTORY`

## TODO

- [x] Windows: URL file path for `import`
- [x] kill app server and esbuild after tests run


packages/remix-serve/cli.ts
```diff
@@ -32,9 +33,41 @@ async function run() {
     process.exit(1);
   }
 
-  let buildPath = url.pathToFileURL(
-    path.resolve(process.cwd(), buildPathArg)
-  ).href;
+  let buildPath = path.resolve(buildPathArg);
```
jacob-ebey, line 36: This will fail on windows, you need the url.PathToFileURL

templates/remix-javascript/package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "build": "remix build",
-    "dev": "remix dev",
+    "dev": "remix dev --manual",
```
jacob-ebey, line 7: Can we default to manual if no command is provided now that the serve CLI that will be used does cache busting?

jacob-ebey (changes requested):

templates/remix-javascript/package.json
pcattori: We could, but then we'd need a way to opt out of `--manual` which we don't currently have.

For v2, Michael and I agreed that we probably want a lever like `--mode=watch`, `--mode=live-reload`, `--mode=hot` or something like that. naming TBD

jacob-ebey: I just mean:
```js
if (!userProvidedCommand) {
  manual = true
}
```

PR #7241, opened 2023-08-23
Adjust flaky tests

brophdawg11:
Removed a few flakey tests that relied on global in-memory state which could conflict with one another during CI runs.  These are really testing router logic so I added tests for these flows in https://github.com/remix-run/react-router/pull/10810.

### DIFF
--- a/integration/redirects-test.ts
+++ b/integration/redirects-test.ts
@@ -15,43 +15,34 @@ test.describe("redirects", () => {
   test.beforeAll(async () => {
     fixture = await createFixture({
       files: {
-        "app/routes/action.tsx": js`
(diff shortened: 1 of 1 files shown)


integration/redirects-test.ts
```diff
@@ -15,155 +15,48 @@ test.describe("redirects", () => {
   test.beforeAll(async () => {
     fixture = await createFixture({
       files: {
-        "app/routes/action.tsx": js`
-          import { Outlet, useLoaderData } from "@remix-run/react";
-
-          if (typeof global.actionCount === "undefined") {
-            global.actionCount = 0;
-            global.actionRequests = new Set();
-          }
```
brophdawg11, line 24: Bad integration test, bad!

PR #7247, opened 2023-08-24
docs(utils/sessions): improve docs

aaronadamsCA:
See https://github.com/remix-run/remix/pull/7225#issuecomment-1688920918 for details. In the absence of a larger rewrite I think this should substantially improve clarity and reduce confusion.

Key bits include:

- Clarifying that regardless of your session storage strategy, if you don't `commitSession()`, you're losing your changes.
- Further clarifying that if you are using cookie session storage and you don't `Set-Cookie`, you're still losing your changes.
- Correcting the cookie size limit. It's actually 1024 bytes for a single cookie value in Chromium and Firefox (sources: [RFC 6265](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis), [Chrome Platform Status](https://chromestatus.com/feature/4946713618939904)). 4 KB is the limit for all cookie names + values combined.
- Consistently recommending `app/session.server.ts` over `app/session.ts`.

The one more opinionated thing I did as well was to remove the lengthy example under `session.flash()`, because there is already an example of the same thing earlier in the doc.

Please let me know if there's anything else I can do to help ship this, so that I can stop making the same mistakes over and over every time I read this doc page. Thanks!

@brophdawg11


docs/utils/sessions.md
```diff
@@ -21,8 +21,8 @@ Remix comes with several pre-built session storage options for common scenarios,
 
 This is an example of a cookie session storage:
 
-```ts filename=app/sessions.ts
-// app/sessions.ts
+```ts filename=app/sessions.server.ts
+// app/sessions.server.ts
```
MichaelDeBoey, line 25:
```suggestion
```

docs/utils/sessions.md
```diff
@@ -309,15 +318,15 @@ const { getSession, commitSession, destroySession } =
 export { getSession, commitSession, destroySession };
 ```
 
-## `createFileSessionStorage` (node)
+## `createFileSessionStorage` (Node.js, Deno)
```
MichaelDeBoey, line 321: Let's also update this in the list at the top of this page

MichaelDeBoey (changes requested):

docs/utils/sessions.md
aaronadamsCA: Done!

docs/utils/sessions.md
```diff
@@ -265,9 +271,11 @@ The `expires` argument to `createData` and `updateData` is the same `Date` at wh
...
 
-The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios. However, cookie-based sessions may not exceed the browser's max-allowed cookie length (typically 4kb).
+The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios.
 
-The downside is that you have to `commitSession` in almost every loader and action. If your loader or action changes the session at all, it must be committed. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away. With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
+However, cookie-based sessions may not exceed the browser's cookie size limit, typically 1024 bytes per cookie value (and 4 KB total for all cookies).
```
brophdawg11, line 276: I think the original 4k is accurate here.  It's 4k per _total_ cookie length (including name/attrs) and 1k per attribute value per [Chrome's docs](https://chromestatus.com/feature/4946713618939904) and the [spec](https://datatracker.ietf.org/doc/html/rfc6265#section-6.1) also states 4k but doesn't say anything about the attribute length limit.

I did a quick test and as expected this is the largest cookie I can set before it gets ignored:

```
"Set-Cookie": `test=${new Array(4092).fill("a").join("")}`,
```

Screenshot 2023-08-25 at 10 05 46 AM


Let's just leave it at 4k and not say anything about the 1k attribute limit.

The only "combined" cookie limits I've run into is ~15k in GCP before you start getting 429 errors.

docs/utils/sessions.md
```diff
@@ -309,15 +317,15 @@ const { getSession, commitSession, destroySession } =
 export { getSession, commitSession, destroySession };
 ```
 
-## `createFileSessionStorage` (node)
+## `createFileSessionStorage` (Node.js, Deno)
 
 For file-backed sessions, use `createFileSessionStorage()`. File session storage requires a file system, but this should be readily available on most cloud providers that run express, maybe with some extra configuration.
 
-The advantage of file-backed sessions is that only the session ID is stored in the cookie while the rest of the data is stored in a regular file on disk, ideal for sessions with more than 4kb of data.
+The advantage of file-backed sessions is that only the session ID is stored in the cookie while the rest of the data is stored in a regular file on disk, ideal for sessions with more than 1024 bytes of data.
```
brophdawg11, line 324:
```suggestion
The advantage of file-backed sessions is that only the session ID is stored in the cookie while the rest of the data is stored in a regular file on disk, ideal for sessions with more than 4k bytes of data.
```

docs/utils/sessions.md
brophdawg11:
```suggestion
However, cookie-based sessions may not exceed the browser's cookie size limit - typically 4k bytes per cookie.
```

MichaelDeBoey:
@aaronadamsCA Are you still interested in getting this PR merged?
If so, could you please rebase onto latest `main` & resolve conflicts?

docs/utils/sessions.md
```diff
@@ -265,9 +271,11 @@ The `expires` argument to `createData` and `updateData` is the same `Date` at wh
...
+The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios.
 
-The downside is that you have to `commitSession` in almost every loader and action. If your loader or action changes the session at all, it must be committed. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away. With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
+However, cookie-based sessions may not exceed the browser's cookie size limit, typically 1024 bytes per cookie value (and 4 KB total for all cookies).
+
+The other downside is that you need to update the `Set-Cookie` header in every loader and action that modifies the session. With other strategies you only need to set the session cookie once, because it doesn't actually store any session data, just the key to find it elsewhere.
```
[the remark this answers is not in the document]
aaronadamsCA: I've made this change.

docs/utils/sessions.md
aaronadamsCA: Thank you. I've made this change too, with an extra detail I did not know before:

> However, cookie-based sessions may not exceed browser cookie size limits of 4k bytes. If your cookie size exceeds this limit, `commitSession()` will throw an error.

This is good to call out, since it clarifies that exceeding the limit will result in an app crash. I was happy to see this safeguard present in the library—better to log an app crash than to have a page rejected by a browser.

https://github.com/remix-run/remix/blame/025ecac479def75dad3db4869ea1b4fb8917df2e/packages/remix-server-runtime/sessions/cookieStorage.ts#L53-L58

docs/utils/sessions.md
aaronadamsCA: Done.

aaronadamsCA:
Whoops, thanks for the bump, @MichaelDeBoey! I completely forgot about this.

I've updated the branch and made @brophdawg11's requested changes. The only thing that needs checking is the line numbers in the code block, I'm not sure how to preview those to ensure the correct lines are being highlighted.

aaronadamsCA:
Thanks for the merge @brophdawg11, though as I read these changes back a year and a half later, I desperately hope React Router middleware can at last solve all these race-condition pains. 😅

PR #7250, opened 2023-08-24
Speed up github actions for PRs

brophdawg11:
This PR splits up the GitHub actions a bit so we can be more granular (and therefore faster) on CI in PRs.  

* Most PRs do not need to run across Windows/OSX - so by default, PRs will only run Ubuntu/Chromium test suites (`test-pr-ubuntu.yml`)
* PRs touching `create-remix` or `remix-dev` are the most susceptible to Windows/OSX issues so when files in those packages are touched we also run the tests for Windows/OSX (`test-pr-windows-macos.yml`)
* `main`/`dev` branches should still have the full suite run against them (`test-full.yml`)

PR #7259, opened 2023-08-25
Switch release branches to use new workflows

brophdawg11:
**Update:**
The below suggestion has been implemented, so this PR now sets main/dev/release/nightly to run everything.

<hr/>

Follow up to #7250 to run the full suite of tests for node 18/20 on `release-*` branches and `nightly` tags

With this PR, here's the current CI approach (where CLI PR means touching code in `create-remix`/`remix-dev`):

| Job                   | PR             | CLI PR | main/dev       | release/nightly |
| --------------------- | -------------- | ------------------------- | -------------- | --------------- |
| Lint                  | ✅             | ✅                        | ✅             | ✅              |
| Build                 | ✅ node@latest | ✅ node@latest            | ✅ node@latest | ✅ node@18+20   |
| Unit (Ubuntu)         | ✅ node@latest | ✅ node@latest            | ✅ node@latest | ✅ node@18+20   |
| Unit (Windows)        |                | ✅ node@latest            | ✅ node@latest | ✅ node@18+20   |
| Integration (Chrome)  | ✅ node@latest | ✅ node@latest            | ✅ node@latest | ✅ node@18+20   |
| Integration (Firefox) |                | ✅ node@latest            | ✅ node@latest | ✅ node@18+20   |
| Integration (Edge)    |                | ✅ node@latest            | ✅ node@latest | ✅ node@18+20   |
| Integration (Safari)  |                | ✅ node@latest            | ✅ node@latest | ✅ node@18+20   |

I'm questioning the value of the main/dev runs though.  It effectively means 2x runs - on for the PR and one for the squashed commit to `dev`.  The intent here is to speed up PRs (which we've done) with the idea that ubuntu+chromium give us a pretty high level of confidence.  So there's less of a reason to re-run for everything else as soon as a commit lands.

What we definitely need to do is run them _regularly_ to catch weird PR-combination-type issues, but I think that can just be a nightly thing?  And if so, we can combine the workflow used by releases with that used by `main`/`dev` and just test everything on `main`/`dev` once a day and might as well do it on node 18 and 20 if we're only doing it that frequently.

This would be the new proposal:

| Job                   | PR             | CLI PR | main/dev/release/nightly |
| --------------------- | -------------- | ------------------------- | ------------------------ |
| Lint                  | ✅             | ✅                        | ✅                       |
| Build                 | ✅ node@latest | ✅ node@latest            | ✅ node@18+20            |
| Unit (Ubuntu)         | ✅ node@latest | ✅ node@latest            | ✅ node@18+20            |
| Unit (Windows)        |                | ✅ node@latest            | ✅ node@18+20            |
| Integration (Chrome)  | ✅ node@latest | ✅ node@latest            | ✅ node@18+20            |
| Integration (Firefox) |                | ✅ node@latest            | ✅ node@18+20            |
| Integration (Edge)    |                | ✅ node@latest            | ✅ node@18+20            |
| Integration (Safari)  |                | ✅ node@latest            | ✅ node@18+20            |

Thoughts?  If we're on board I'll update this PR to match the second table here.


.github/workflows/test-release.yml
```diff
@@ -0,0 +1,64 @@
...
+on:
+  push:
+    branches:
+      - release-*
+    tags:
+      - "v0.0.0-nightly-*"
+    paths-ignore:
+      - "docs/**"
+      - "scripts/**"
+      - "**/README.md"
```
brophdawg11, line 15: Same triggers as we had before in `.github/workflows/release-test.yml`

.github/workflows/test-release.yml
```diff
@@ -0,0 +1,64 @@
...
+
+jobs:
```
brophdawg11, line 17: Currently this is the same as `test-full.yml` but using node 18+20 instead of latest

MichaelDeBoey (approved):
I would say combining `main`/`dev`/`release`/`nightly` makes sense

.github/workflows/shared-build.yml
```diff
@@ -11,9 +11,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - name: 🛑 Cancel Previous Runs
-        uses: styfle/cancel-workflow-action@0.11.0
```
brophdawg11, line 15: We shouldn't need this and can use the built in [`concurrency`](https://docs.github.com/en/actions/using-jobs/using-concurrency) setting, which is recommended on the [README](https://github.com/styfle/cancel-workflow-action#readme) for this action these days as well.

This also allows us better control over cancellation behavior per workflow.  We do not want to cancel runs on `dev` for example since they may come in repeatedly and we don't want a bunch of cancellations queueing up which may obscure the commit which broke something:

Screenshot 2023-08-28 at 10 09 11 AM


So we only use the concurrency setting on the PR workflows, not the `test.yml` workflow

brophdawg11: We use this `cancel-workflow-action` in a bunch of other places, if we see this approach works without issue we can move those to use `concurrency` in a separate PR

.github/workflows/test-full.yml
```diff
@@ -1,12 +1,16 @@
 name: Branch
 
-# main/dev branches will get the full run across all OS/browsers
+# main/dev/release-* branches and nightly tags will get the full run across
+# all OS/browsers for multiple node versions
```
brophdawg11, line 4: This full workflow now runs on main/dev/release/nightly and runs for both node 18/20

MichaelDeBoey (approved):

PR #7191, opened 2023-08-17
docs: small updates

MichaelDeBoey:
Follow-up of @ryanflorence's docs changes

Closes #7441

PR #7260, opened 2023-08-25
docs(discussion/04-server-vs-client.md): add `headers`

gustavoguichard:
On lines 11-15 of this page, `headers` is listed along with `loader` and `action` so I thought it'd be cool to update the example to add a `headers` export. If that is not desirable, the `headers` item should probably removed from that list.

Also couple lil copy fixes


MichaelDeBoey:
Thanks for creating this PR @gustavoguichard 

However, I'm going to close this as the changes in here are already in #7191

gustavoguichard:
Cool @MichaelDeBoey ! Appreciate the progress on the docs ;)

PR #7264, opened 2023-08-26
chore: cleanup `createFixtureProject` in integration test helpers

MichaelDeBoey:
We don't include any `jsconfig.json` files in our fixtures anymore

### DIFF
--- a/integration/helpers/create-fixture.ts
+++ b/integration/helpers/create-fixture.ts
@@ -327,11 +327,6 @@ async function writeTestFiles(init: FixtureInit, dir: string) {
       let filePath = path.join(dir, filename);
       await fse.ensureDir(path.dirname(filePath));
       let file = init.files![filename];
-      // if we have a jsconfig we don't want the tsconfig to exist
-      if (filename.endsWith("jsconfig.json")) {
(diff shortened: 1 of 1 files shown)


brophdawg11 (dismissed):

integration/helpers/create-fixture.ts
```diff
@@ -327,11 +327,6 @@ async function writeTestFiles(init: FixtureInit, dir: string) {
       let filePath = path.join(dir, filename);
       await fse.ensureDir(path.dirname(filePath));
       let file = init.files![filename];
-      // if we have a jsconfig we don't want the tsconfig to exist
-      if (filename.endsWith("jsconfig.json")) {
-        let parsed = path.parse(filePath);
-        await fse.remove(path.join(parsed.dir, "tsconfig.json"));
-      }
```
brophdawg11, line 334: Couldn't someone still technically add one via the fixture setup?  I don't have a specific use-case offhand but it it worth removing this if it prevents that in the future?

```js
  files: {
    "jsconfig.json": ...
  }
}
```

brophdawg11 (approved):

PR #7265, opened 2023-08-27
chore: gitignore playwright-report dirs anywhere

lpsinger:
Running the integration tests results in the directory /integration/playwright-report, which should be ignored.


MichaelDeBoey:
@lpsinger We only accept docs changes on `main`, so this change should go to `dev` instead

Closing this as we already have this on `dev`
https://github.com/remix-run/remix/blob/c28c3cc6a577653a857f3f38a3d93eefa091f102/.gitignore#L16

PR #7269, opened 2023-08-27
feat(dev): make browser Node polyfills opt-in

markdalgleish:
This PR makes it so we no longer polyfill built-in Node modules by default. To allow consumers to opt back in to polyfills, this also adds a new `browserNodeBuiltinsPolyfill` option that mirrors the existing `serverNodeBuiltinsPolyfill`.

This also adds a `globals` property to both of these options so Node globals like `Buffer` can be enabled with a simple flag rather than needing to manually change production code to modify the global environment.

A couple of the tests have been updated since they were relying on Node builtins in the browser.

I originally planned on making the `browserNodeBuiltinsPolyfill` default to the value of `serverNodeBuiltinsPolyfill` if it was provided, but in practice I found it confusing that a Remix config with just `serverNodeBuiltinsPolyfill` defined would also have browser polyfills. It does mean there's a bit more boilerplate if you want to use the same polyfills across both, but I think it's better to be explicit, and it's easy to share a single config object across both since the Remix config is a JS file anyway.

✨ This PR also adds a bonus feature where the dev server restarts when the Remix config changes. (Thanks @pcattori!)


packages/remix-dev/compiler/js/compiler.ts
```diff
@@ -156,11 +149,12 @@ export const create = async (
 ): Promise<Compiler> => {
   let compiler = await esbuild.context({
     ...createEsbuildConfig(ctx, refs),
+    write: false,
```
markdalgleish, line 152: We need to manually write the result of the browser build to disk so that [esbuild-plugins-node-modules-polyfill](https://github.com/imranbarbhuiya/esbuild-plugins-node-modules-polyfill) can check the build output in memory for any missing/unconfigured polyfills after tree-shaking has occured. This is important in the context of Remix since our route files often depend on Node builtins. If `write` is enabled, the `outputFiles` are not kept in memory and are not passed to the `onEnd` callback of the esbuild plugin.

If you're interested, here's the PR where I implemented this: https://github.com/imranbarbhuiya/esbuild-plugins-node-modules-polyfill/pull/152

packages/remix-dev/compiler/compiler.ts
```diff
@@ -100,7 +101,8 @@ export let create = async (ctx: Context): Promise<Compiler> => {
     // js compilation (implicitly writes artifacts/js)
     let js = await tasks.js;
     if (!js.ok) throw error ?? js.error;
-    let { metafile, hmr } = js.value;
+    let { metafile, outputFiles, hmr } = js.value;
+    writes.js = JS.write(ctx.config, outputFiles);
```
markdalgleish, line 105: As explained in my [other comment](https://github.com/remix-run/remix/pull/7269/files#r1306751467), we now need to manage writing the output of the JS build to disk.

packages/remix-dev/compiler/js/compiler.ts
```diff
@@ -94,15 +95,7 @@ const createEsbuildConfig = (
...
     }),
-    nodeModulesPolyfillPlugin({
-      // For the browser build, we replace any Node built-ins that don't have a
-      // polyfill with an empty module. This ensures the build can pass without
-      // having to mark all Node built-ins as external which can result in other
-      // issues, e.g. https://github.com/remix-run/remix/issues/5521. We then
-      // rely on tree-shaking to remove all unused polyfills and fallbacks.
-      fallback: "empty",
-    }),
-    externalPlugin(/^node:.*/, { sideEffects: false }),
```
markdalgleish, line 105: Since [esbuild-plugins-node-modules-polyfill](https://github.com/imranbarbhuiya/esbuild-plugins-node-modules-polyfill) is now configured to manage fallbacks, this usage of `externalPlugin` was redundant.

packages/remix-dev/compiler/js/plugins/browserNodeBuiltinsPolyfill.ts
```diff
@@ -0,0 +1,37 @@
...
+  nodeModulesPolyfillPlugin({
+    // Rename plugin to improve error message attribution
+    name: "browser-node-builtins-polyfill-plugin",
+    // Only pass through the "modules" and "globals" options to ensure we
+    // don't leak the full plugin API to Remix consumers.
+    modules: ctx.config.browserNodeBuiltinsPolyfill?.modules ?? {},
+    globals: ctx.config.browserNodeBuiltinsPolyfill?.globals ?? {},
+    // Mark any unpolyfilled Node builtins in the build output as errors.
+    fallback: "error",
+    formatError({ moduleName, importer, polyfillExists }) {
```
markdalgleish, line 15: I added this functionality here: https://github.com/imranbarbhuiya/esbuild-plugins-node-modules-polyfill/pull/153.

packages/remix-dev/compiler/js/plugins/browserNodeBuiltinsPolyfill.ts
```diff
@@ -0,0 +1,37 @@
...
+export const browserNodeBuiltinsPolyfillPlugin = (ctx: Context) =>
+  nodeModulesPolyfillPlugin({
+    // Rename plugin to improve error message attribution
+    name: "browser-node-builtins-polyfill-plugin",
+    // Only pass through the "modules" and "globals" options to ensure we
+    // don't leak the full plugin API to Remix consumers.
+    modules: ctx.config.browserNodeBuiltinsPolyfill?.modules ?? {},
+    globals: ctx.config.browserNodeBuiltinsPolyfill?.globals ?? {},
+    // Mark any unpolyfilled Node builtins in the build output as errors.
+    fallback: "error",
```
markdalgleish, line 14: I added support for error fallbacks here: https://github.com/imranbarbhuiya/esbuild-plugins-node-modules-polyfill/pull/152

packages/remix-dev/compiler/server/plugins/serverNodeBuiltinsPolyfill.ts
```diff
@@ -0,0 +1,17 @@
...
+    // Rename plugin to improve error message attribution
+    name: "server-node-builtins-polyfill-plugin",
+    // Only pass through the "modules" and "globals" options to ensure we
+    // don't leak the full plugin API to Remix consumers.
+    modules: ctx.config.serverNodeBuiltinsPolyfill?.modules ?? {},
+    globals: ctx.config.serverNodeBuiltinsPolyfill?.globals ?? {},
+    // Since the server environment may provide its own Node polyfills,
+    // we don't define any fallback behavior here and allow all Node
+    // builtins to be marked as external
+    fallback: "none",
```
MichaelDeBoey, line 16: This is the default behavior, so technically not needed to explicitly add it here

packages/remix-dev/compiler/js/write.ts
```diff
@@ -0,0 +1,17 @@
+import * as path from "node:path";
+import type * as esbuild from "esbuild";
```
MichaelDeBoey, line 2: Any reason why we're not just importing `OutputFile` type here?

```suggestion
import type { OutputFile } from "esbuild";
```

packages/remix-dev/compiler/js/write.ts
```diff
@@ -0,0 +1,17 @@
...
+import fse from "fs-extra";
+
+import type { RemixConfig } from "../../config";
+
+export async function write(
+  config: RemixConfig,
+  outputFiles: esbuild.OutputFile[]
```
MichaelDeBoey, line 9:
```suggestion
  outputFiles: OutputFile[]
```

MichaelDeBoey (approved):

packages/remix-dev/compiler/server/plugins/serverNodeBuiltinsPolyfill.ts
markdalgleish: That's true, I was just using this as an excuse to document the behaviour since it impacts consumers.

packages/remix-dev/compiler/js/write.ts
markdalgleish: This was leftover after adapting the code from `packages/remix-dev/compiler/server/write.ts`.

MichaelDeBoey (approved):

pcattori (approved):

PR #7279, opened 2023-08-28
Update CI workflows to cancel based on built-in concurrency setting

brophdawg11:
Switch to built-in CI job cancellation as recommended in https://github.com/styfle/cancel-workflow-action

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,9 +24,6 @@ jobs:
       published_packages: ${{ steps.changesets.outputs.publishedPackages }}
       published: ${{ steps.changesets.outputs.published }}
     steps:
-      - name: 🛑 Cancel Previous Runs
-        uses: styfle/cancel-workflow-action@0.11.0
-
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -24,9 +24,6 @@ jobs:
       published_packages: ${{ steps.changesets.outputs.publishedPackages }}
       published: ${{ steps.changesets.outputs.published }}
     steps:
-      - name: 🛑 Cancel Previous Runs
-        uses: styfle/cancel-workflow-action@0.11.0
-
```
brophdawg11, line 29: I didn't add `concurrency` to the release workflow since we probably want any started release to finish?  For example if we start a `pre.0` release and then quickly fix a bug and kick off a `pre.1` - we would want them both to complete and publish to npm instead of leaving a gap in the prereleases?

1. It should never overlap since it runs once every 24 hours

MichaelDeBoey (approved):

PR #7280, opened 2023-08-28
fix(dev): kill app server when `remix dev` terminates

pcattori:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/devServer_unstable/index.ts
+++ b/packages/remix-dev/devServer_unstable/index.ts
@@ -7,6 +7,7 @@ import prettyMs from "pretty-ms";
 import execa from "execa";
 import express from "express";
 import pc from "picocolors";
+import exitHook from "exit-hook";
 
 import * as Channel from "../channel";
 import { type Manifest } from "../manifest";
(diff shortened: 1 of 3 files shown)


MichaelDeBoey (approved):

pcattori:
holding off on merging this. need to rebase to `release-next` after v2 prerelease is cut

packages/remix-dev/devServer_unstable/index.ts
```diff
@@ -267,6 +268,12 @@ export let serve = async (
 
   server.listen(options.port);
 
+  exitHook(() => {
+    state.appServer?.pid && kill(state.appServer.pid);
```
pcattori, line 272: probably want to call `state.appServer.kill()` instead to let the app server terminate gracefully

packages/remix-dev/devServer_unstable/index.ts
```diff
@@ -267,6 +268,12 @@ export let serve = async (
...
+    websocket.close();
+    server.close();
+  });
+
   return new Promise(() => {}).finally(async () => {
```
ngbrown, line 277: What does this `Promise` do? As far as I've tested it, it never resolves or rejects.

packages/remix-dev/devServer_unstable/index.ts
ngbrown, line 272: Isn't `kill` async?

PR #7291, opened 2023-08-29
feat(remix-dev)!: set `esbuild`'s `target` to `'node18'`

MichaelDeBoey:
Fixes #5203

PR #7292, opened 2023-08-29
compile server build to node 18

pcattori:
Closes #5203


MichaelDeBoey:
Seems like we created a PR at almost the same time: #7291 🙈

pcattori:
@MichaelDeBoey whoops, just saw that! merged this one since this was targeting `release-next`

PR #7309, opened 2023-08-30
add bug report for non working `shouldRevalidate`

buschco:
I added a test that demonstrates that `shouldRevalidate` is not working as expected.
see https://github.com/remix-run/remix/issues/7310

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -572,3 +572,4 @@
 - AlemTuzlak
 - ledniy
 - rogepi
+- buschco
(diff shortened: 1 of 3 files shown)


akamfoad:
Seems like it is a bug in React Router as well https://stackblitz.com/edit/should-revalidate-skipped-6-15-0

brophdawg11:
Commented in #7310 - this is working as expected because these are not revalidations.

PR #6708, opened 2023-06-27
Roll back minimatch version to fix Typescript errors

lpsinger:
Partially revert #6045. This is an alternative to #6620.

I think that this might fix the following error that you get when you run `npm run typecheck` on a freshly bootstrapped project generated using `npx create-remix@latest`:

```
node_modules/@types/glob/index.d.ts:29:42 - error TS2694: Namespace '"/private/tmp/my-remix-app/node_modules/minimatch/dist/cjs/index"' has no exported member 'IOptions'.

29     interface IOptions extends minimatch.IOptions {
                                             ~~~~~~~~

node_modules/@types/glob/index.d.ts:74:30 - error TS2724: '"/private/tmp/my-remix-app/node_modules/minimatch/dist/cjs/index"' has no exported member named 'IMinimatch'. Did you mean 'Minimatch'?

74         minimatch: minimatch.IMinimatch;
                                   ~~~~~~~~~~
```

But I don't know how to test `create-remix` on a branch.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

PR #6620, opened 2023-06-16
Update sort-package-json to 2.4.1

lpsinger:
I think that this might fix the following error that you get when you run `npm run typecheck` on a freshly bootstrapped project generated using `npx create-remix@latest`:

```
node_modules/@types/glob/index.d.ts:29:42 - error TS2694: Namespace '"/private/tmp/my-remix-app/node_modules/minimatch/dist/cjs/index"' has no exported member 'IOptions'.

29     interface IOptions extends minimatch.IOptions {
                                             ~~~~~~~~

node_modules/@types/glob/index.d.ts:74:30 - error TS2724: '"/private/tmp/my-remix-app/node_modules/minimatch/dist/cjs/index"' has no exported member named 'IMinimatch'. Did you mean 'Minimatch'?

74         minimatch: minimatch.IMinimatch;
                                   ~~~~~~~~~~
```

...the reason being that the latest versions of `glob` and `minimatch` both ship their own TypeScript definitions now, and this removes the transitiive dependency on `@types/glob`.

But I don't know how to test `create-remix` on a branch.

PR #7317, opened 2023-08-31
fix(remix-dev): remove `sort-package-json`

MichaelDeBoey:
When running `npx run typecheck` on a newly installed project, I get the following errors

```sh
node_modules/@types/glob/index.d.ts:29:42 - error TS2694: Namespace '"/Users/michaeldeboey/Downloads/my-remix-app/node_modules/minimatch/dist/mjs/index"' has no exported member 'IOptions'.

29     interface IOptions extends minimatch.IOptions {
                                            ~~~~~~~~

node_modules/@types/glob/index.d.ts:74:30 - error TS2724: '"/Users/michaeldeboey/Downloads/my-remix-app/node_modules/minimatch/dist/mjs/index"' has no exported member named 'IMinimatch'. Did you mean 'Minimatch'?

74         minimatch: minimatch.IMinimatch;
                                ~~~~~~~~~~
```

Running `npm ls @types/glob` to figure out where the dependency comes from, show us that it's a dependency of `sort-package-json`, which is a dependency of `@remix-run/dev`

```sh
my-remix-app@ [REDACTED]/my-remix-app
└─┬ @remix-run/dev@2.0.0-pre.4
  └─┬ sort-package-json@1.57.0
    └─┬ globby@10.0.0
      └── @types/glob@7.2.0
```

Since the `create-remix` change, we don't need `sort-package-json` in `@remix-run/dev` anymore so removed it from it's dependencies + also from the root

---

Closes #6620
Closes #6708

### DIFF
--- a/.changeset/metal-horses-camp.md
+++ b/.changeset/metal-horses-camp.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+REMOVE: removed dependency that caused TS error
--- a/package.json
+++ b/package.json
@@ -115,7 +115,6 @@
     "rollup-plugin-copy": "^3.3.0",
     "semver": "^7.3.7",
     "simple-git": "^3.16.0",
-    "sort-package-json": "^1.55.0",
     "to-vfile": "7.2.3",
     "typescript": "^5.1.0",
     "unified": "^10.1.2",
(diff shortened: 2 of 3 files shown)


brophdawg11:
Can you add a changeset here so this merging this PR will trigger a new prerelease?  You can prefix it with `REMOVE:` or something so we know it doesn't need to remain in the final release notes

PR #7319, opened 2023-08-31
feat!: align types with React Router

brophdawg11:
Aligning types with those used in React Router to remove the confusion around `LoaderArgs`/`LoaderFunctionArgs` etc.

See the changeset for the details and [this comment](https://github.com/remix-run/remix/pull/7319#issuecomment-1702974107) for the desired state


MichaelDeBoey:
> * Typed as `any`

@brophdawg11 I would argue that we should update RR to use `unknown` instead

Since that would be a breaking change on the RR side, I think it would be best to rename `ActionArgs` to `ActionFunctionArgs` and `LoaderArgs` to `LoaderFunctionArgs` and implement it here like

```ts
export type AppLoadContext = unknown;
```
```ts
export type DataFunctionArgs<Context = unknown> = {
  context: Context;
  params: Params;
  request: Request;
}
```

Once we also updated this in RR (this should be done in a major version), we can remove it here and just re-export RR types as you wanted to do in this PR

brophdawg11:
I'd prefer not to replace one set of duplicated types with another.  If `any` is truly not what we want, I'd rather just keep it as is in Remix today I think.  

I think naming the exports the same if they aren't identical is _more_ confusing since then `LoaderFunctionArgs` is different depending on where you import it form.  With VSCode auto-complete very often incorrectly recommending `react-router-dom` higher than `@remix-run/react` in the list - folks may end up with the wrong one unknowingly.  I think the current state of `LoaderArgs`/`LoaderFunctionArgs` is actually a benefit if they're different.

FWIW we have [chosen `any` specifically before](https://github.com/remix-run/react-router/pull/8702#discussion_r819907643).  Using `unknown` is certainly stricter and "better" from a TS purist point of view, but it also feels sometimes like when we type something as `unknown`, we _force_ all consumers to be purists as well.  I'll defer to @mjackson and @ryanflorence on any official preference though.

mjackson (commented):
LGTM. Only thing I think we could probably do a bit better is use `unknown` for data we get from the user in `AppData`, but it isn't a huge deal and would probably break some people's typechecking, so I'd say let's just leave it as `any` for now.

MichaelDeBoey:
> I'd rather just keep it as is in Remix today I think

The thing is: if we want to make it `unknown` at some point, that would be a breaking change, as `unknown` requires an explicit type check, `any` doesn't require anything at all

I get the idea of `any` being "easier to use", but imo that would people give a wrong feeling of (type) safety, so I think we should go with `unknown`, even if that means people will be forced to do extra checks
Isn't the whole thing about Remix "we learn you how to do it the proper way"?
Having the extra checks is "the proper way", so the way to go imo

brophdawg11:
I don't have a strong preference either way really, I just would like us to be consistent on types we don't know - stuff the user hands to us somewhere and we provide back to them elsewhere - `context`, `loaderData`, `handle`, `location.state` (I'm sure there are other's I'm not thinking of at the moment).  

The current state is:

|                | React Router | Remix                        |
| -------------- | ------------ | ---------------------------- |
| Context        | `any`        | `{ [key: string]: unknown }` |
| useLoaderData  | `unknown`    | `any` (via generic)          |
| handle         | `any`        | `any`                        |
| location.state | `any`        | `any` (inherited)            |

So do we:

1. Align on `any` today since it's the smallest change and also does not require a breaking RR change?  
   * We can add a generic to `DataFunctionArgs` so folks can type it stronger than `any`
   * Then in the near future we can export new types based on `unknown`, deprecate the `any` ones, and drop the `any` ones in the next major version of both RR and Remix
3. Or, move Remix to `unknown` in v2, and follow the deprecation path to move RR to `unknown` in v7
   * This means we'll be exporting duplicate types (`LoaderFunctionArgs` will be different in Remix than it is in React Router - but maybe only in the default value for the generic)

mjackson:
I'd personally go with option 2 since I think that's where we eventually want to be.

.changeset/align-rr-types.md
```diff
@@ -0,0 +1,20 @@
...
+* Change exposed `any` types to `unknown`
+  * `AppData`
+  * `useLocation.state`
+  * `useMatches()[i].data`
+  * `useFetcher().data`
+  * `MetaMatch.handle`
+* `useMatches()[i].handle` type changed from `{ [k: string]: any }` to `unknown`
+* `AppLoadContext` type changed from `{ [k: string]: unknown }` to `unknown`
+* Make `LoaderFunctionArgs`/`ActionFunctionArgs` generic to accept a `context` type
+* Rename `LoaderFunction`/`ActionFunction` to `LoaderFunctionArgs`/`ActionFunctionArgs`
```
MichaelDeBoey, line 19:
```suggestion
* Rename `LoaderArgs`/`ActionArgs` to `LoaderFunctionArgs`/`ActionFunctionArgs`
```

packages/remix-react/components.tsx
```diff
@@ -975,32 +976,16 @@ function dedupe(array: any[]) {
...
-   * @see https://remix.run/route/handle
-   */
-  handle: undefined | { [key: string]: any };
+export interface UIMatch<D = AppData> extends UIMatchRR<SerializeFrom<D>> {}
+
+/**
+ * Returns the active route matches, useful for accessing loaderData for
+ * parent/child routes or the route "handle" property
+ *
+ * @see https://remix.run/docs/hooks/use-matches
```
MichaelDeBoey, line 985:
```suggestion
 * @see https://remix.run/hooks/use-matches
```

packages/remix-server-runtime/reexport.ts
```diff
@@ -1,4 +1,10 @@
-export type { ErrorResponse } from "@remix-run/router";
+export type {
+  ActionFunction,
+  ActionFunctionArgs,
+  ErrorResponse,
+  LoaderFunction,
+  LoaderFunctionArgs,
+} from "@remix-run/router";
```
MichaelDeBoey, line 7: We should keep exporting these from `./routeModules`

packages/remix-react/data.ts
```diff
@@ -5,10 +5,8 @@ import {
 
 /**
  * Data for a route that was returned from a `loader()`.
- *
- * Note: This moves to unknown in ReactRouter and eventually likely in Remix
  */
-export type AppData = any;
+export type AppData = unknown;
```
MichaelDeBoey, line 9: Is it even useful to have this type exported?

We can't override it, so... 🤷‍♂️

packages/remix-react/routeModules.ts
```diff
@@ -121,7 +121,7 @@ export type RouteComponent = ComponentType<{}>;
  *
  * @see https://remix.run/route/handle
  */
-export type RouteHandle = any;
+export type RouteHandle = unknown;
```
MichaelDeBoey, line 124: Same here 🤔

.changeset/align-rr-types.md
```diff
@@ -0,0 +1,20 @@
+---
+"@remix-run/cloudflare": major
+"@remix-run/deno": major
+"@remix-run/node": major
+"@remix-run/server-runtime": major
```
MichaelDeBoey, line 5:
```suggestion
"@remix-run/react": major
"@remix-run/server-runtime": major
```

brophdawg11:
Unsure what's going on in CI - but worried it might be a weird `node_modules` cache issue.  Maybe closing/re-opening will clear something up...

Screenshot 2023-09-05 at 4 40 48 PM

MichaelDeBoey (changes requested):

packages/remix-react/components.tsx
```diff
@@ -975,32 +976,16 @@ function dedupe(array: any[]) {
...
-   * Any route data associated with the matched route
-   */
-  data: any;
-  /**
-   * The exported `handle` object of the matched route.
-   *
-   * @see https://remix.run/route/handle
-   */
-  handle: undefined | { [key: string]: any };
+export interface UIMatch<D = AppData> extends UIMatchRR<SerializeFrom<D>> {}
```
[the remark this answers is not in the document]
brophdawg11: I kept this as an interface to match what it was - is there a functional reason to prefer `type` here?

packages/remix-react/data.ts
brophdawg11: I'd be fine with that - we can run it by @mjackson and do that in a follow up PR if we want to drop all these aliases for `any`/`unknown` from the public API

packages/remix-server-runtime/reexport.ts
brophdawg11: yeah I realized I had them wrong locally when I tested on a playground app - pushing that up shortly

MichaelDeBoey: As long as it's merged _before_ v2 releases as a stable version, I'm fine with that

As removing them would be a breaking change as well

packages/remix-react/data.ts
MichaelDeBoey: As long as it's merged _before_ v2 releases as a stable version, I'm fine with that

As removing them would be a breaking change as well

packages/remix-react/components.tsx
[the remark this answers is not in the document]
MichaelDeBoey: Functionally they're _kinda_ equivalent, but:

- Interfaces can't express unions, mapped types, or conditional types. Type aliases can express any type.
- Interfaces with the same name in the same scope merge their declarations, leading to unexpected bugs
  Since this is a generic and it's not meant to be overridden, I think it's better to be safe than sorry and not make any (no pun intended 😅) unwanted side-effects possible
- Type aliases have an implicit index signature of `Record<PropertyKey, unknown>`

brophdawg11:
CI failure seems to be due to https://github.com/shelljs/shelljs/issues/1133

PR #7323, opened 2023-08-31
fix(remix-express): respect expressjs hostname property when behind reverse proxies

rossipedia:
`@remix-run/express` creates `request.url` directly from the `Host` HTTP header, but when behind a reverse proxy with `app.enable('trust proxy')` set, the actual hostname is provided via Express's `req.hostname` property, which is a "proxy aware" property that parses either `Host` or `X-Forwarded-Host` headers depending on the app configuration.

This can cause unexpected behavior when a user requests (for example) `https://remix-app.org.com/page`, but some condition causes a redirect to be thrown that uses `request.url` for a `returnTo` parameter (for instance: auth failure due to a protected resource).

If the app is behind a reverse proxy, and utilizing an internal DNS name (like an AWS internal host name), then _that_ host will be used to create `request.url`, and you'll get a response with a bad `Location` header, and will also be leaking internal hostnames externally.

For example: 
```ts
// server.js
const app = express();
app.enable('trust proxy');
// ... etc ...

// ~/routes/page.tsx
export function loader({request}: LoaderArgs) {
  await authenticator.isAuthenticated(request, {
    failureRedirect: `/login?returnTo=${encodeURIComponent(request.url)}`
  });
}
```

Request:
```
GET /page HTTP/1.1
Host: remix-app.internal.us-east.aws.io
X-Forwarded-Host: remix-app.org.com
```
Response:
```
HTTP/1.1 302 Found
Location /login?returnTo=http%3A%2F%2Fremix-app.internal.us-east.aws.io%2F <- INCORRECT
```

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

Made the change to `node_modules/@remix-run/express/dist/server.js` and verified that `request.url` reflected the `X-Forwarded-Host` header when `app.enable('trust proxy')` was set in my Express app.


packages/remix-express/server.ts
```diff
@@ -87,7 +87,11 @@ export function createRemixRequest(
   req: express.Request,
   res: express.Response
 ): Request {
-  let url = new URL(`${req.protocol}://${req.get("host")}${req.url}`);
+  let [, port] = req.get("host")?.split(":") ?? [];
```
brophdawg11, line 90: Should this try to get it directly from `X-Forwarded-Host`?  It looks like `req.hostname` strips it off, but wouldn't be run into issues if the proxy changed the port as well?

Locally I was testing this with the following `nginx` config (along with `127.0.0.1 example.com` in `/etc/hosts`):

```
http {
    server {
        listen 8080;
        server_name example.com;

        location / {
            proxy_pass http://localhost:3000;
            proxy_set_header X-Forwarded-Host example.com:8080;
        }
    }
}
```

When doing it that way it ends up with `http://example.com:3000` which works in this small setup since port 3000 is also exposed and `example.com` is an alias for `localhost` - but we can get the port if it exists using something like:

```js
  // req.hostname doesn't include port information so grab that from 
  // `X-Forwarded-Host` or `Host`
  let [,hostnamePort] = req.get('X-Forwarded-Host')?.split(':') ?? [];
  let [,hostPort] = req.get('host')?.split(':') ?? [];
  let port = hostnamePort || hostPort;
  // Use req.hostname here as it respects the "trust proxy" setting
  let resolvedHost = `${req.hostname}${port ? `:${port}` : ""}`;
  let url = new URL(`${req.protocol}://${resolvedHost}${req.url}`);
```

Thoughts?

rossipedia: Yeah that seems like a reasonable approach :+1:

packages/remix-express/server.ts
```diff
@@ -87,7 +87,11 @@ export function createRemixRequest(
...
+  // Use req.hostname here as it respects the "trust proxy" setting
+  // but unfortunately doesn't include any port information
+  let resolvedHost = `${req.hostname}${port ? `:${port}` : ""}`;
+  let url = new URL(`${req.protocol}://${resolvedHost}${req.url}`);
```
rossipedia, line 94:
```suggestion
  // req.hostname doesn't include port information so grab that from 
  // `X-Forwarded-Host` or `Host`
  let [,hostnamePort] = req.get('X-Forwarded-Host')?.split(':') ?? [];
  let [,hostPort] = req.get('host')?.split(':') ?? [];
  let port = hostnamePort || hostPort;
  // Use req.hostname here as it respects the "trust proxy" setting
  let resolvedHost = `${req.hostname}${port ? `:${port}` : ""}`;
  let url = new URL(`${req.protocol}://${resolvedHost}${req.url}`);
```

brophdawg11 (approved):

brophdawg11:
Oops, sorry this must have fallen off my radar.  Do you want to add a changeset and we can get this merged?

PR #7333, opened 2023-09-04
feat(remix-react): make RouteMatch a generic type

lpsinger:
Now the `RouteMatch` type is a generic type that supports type inference. So just as you can do `useLoaderData<typeof loader>()`, now you can also do `RouteMatch<typeof loader>`.

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.changeset/stale-moles-cheat.md
+++ b/.changeset/stale-moles-cheat.md
@@ -1,0 +1,7 @@
+---
+"@remix-run/react": minor
+---
+
+Make `RouteMatch` a generic type
+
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thanks for the PR!  This was actually also done as part of the larger type alignment with RR in https://github.com/remix-run/remix/pull/7319.  `RouteMatch` was also renamed to `UIMatch` to more clearly differentiate it from the other `*RouteMatch` types we use as return values from `matchRoutes` in React Router.

Feel free to give it a shot and see if it works for you!  `npx upgrade-remix pre`

PR #7343, opened 2023-09-05
Custom error instances are eaten by error catching

mikeybinns:
This adds a failing bug test which shows that custom error instances are being eaten by the error boundary or useRouteError.

Ideally, Remix should pass through an error instance as-is (as long as it's still an instance of Error), and it would fall to the user to check "Is this error an instance of custom error".

This change would not affect existing code because as the test shows, custom errors are also instances of Error.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -370,6 +370,7 @@
 - michaelgmcd
 - michaelhelvey
 - michaseel
+- mikeybinns
 - mikeybinnswebdesign
 - mirzafaizan
 - mitchelldirt
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -40,51 +40,55 @@ let appFixture: AppFixture;
 //    ```
 ////////////////////////////////////////////////////////////////////////////////
 
-test.beforeEach(async ({ context }) => {
-  await context.route(/_data/, async (route) => {
-    await new Promise((resolve) => setTimeout(resolve, 50));
-    route.continue();
-  });
-});
-
 test.beforeAll(async () => {
   fixture = await createFixture({
     config: {
-      future: { v2_routeConvention: true },
+      future: { v2_routeConvention: true, v2_errorBoundary: true },
     },
     ////////////////////////////////////////////////////////////////////////////
     // 💿 Next, add files to this object, just like files in a real app,
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
+      "app/customError.js": js`
+      export class CustomError extends Error {
+        // Implementation here doesn't matter for the test, however if looking
+        // for a use case, imagine a class that only shows the error message
+        // if in dev mode, or a class that links a tag to an error message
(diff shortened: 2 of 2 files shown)


mikeybinns:
Note: mikeybinnswebdesign is my old username in the CLA. If you think my name is a duplicate it's because it sort of is.

My correct name is mikeybinns and so if you want to remove one, please remove mikeybinnswebdesign.

I received an email recently saying I was removed and I expect it was due to this username change (I previously added my name to contributors but I don't think my work was merged)

brophdawg11:
This isn't something Remix can do for you automatically because it has to serialize that error across the network boundary as JSON.  And then we don't have access to the `class CustomError` instance when it needs to deserialize it back during hydration.  We can only deserialize built-in Error sub-types, such as `ReferenceError`, `TypeError`, etc.

What is your specific use case for this so we can potentially offer up an alternate solution?

mikeybinns:
Okay, I have a basically got two error classes, so I'll try and explain them both here:
First, I have UserError. This class is like a normal error, but it has some logic where it creates and adds a unique ID to assist with support queries, and instead of writing error messages in place, I have a glossary of error messages (some with interpolation for implementation details) so that if I need to look up an error, I can do it in one place instead of searching the code.

Second is DevError, which is the same as UserError, but it also has logic to display a generic message instead of the error if we're in production, so users don't get developer error messages.

I've currently been able to work around the issue by converting dev errors into user errors, and then converting them back in the boundary manually with some custom functions, however now I'm getting an issue where on production, the error isn't passed through, I only get it console logged, and then it is replaced with "Unexpected server error" on production builds. This means I lose the unique IDs and category tags so I can't display them to customers, rendering them less helpful than hoped.

I think this is probably a security feature rather than a bug, so my plan is to make a route wrapper function with a prop called "ExpectedErrorBoundary" so I can display expected errors by returning them from loaders instead of throwing them. If you have another solution or suggestion, feel free to let me know :)

brophdawg11:
Errors in Remix are expected to be things you _didn't_ handle - not control flows for things you _are_ handling.  And Errors are also always santized in production for security purposes so no server information is leaked to the client.

Instead, you're going to be better off throwing a `Response` with whatever information you need.  That way no data will be stripped and you can then put whatever info you want in there and convert it back into a specific error type:

```jsx
export function loader() {
  if (whatever) {
    throw json({ 
      type: 'CustomError', 
      message: 'something', 
      code: 123
    }, { 
      status: 500, 
      statusText: "Unexpected Server Error"
    });
  }
  ...
}

export function ErrorBoundary() {
  let error = useRouteError();
  if (isRouteErrorResponse(error)) {
    if (error.data.type === "CustomError") {
      let customError = new CustomError(error.data.message, error.data.code);
      return <CustomErrorUI error={customError} />
    }
    return <p>{error.status} {error.statusText}</p>;
  }
  return <p>{error.message || 'Unknown Error'}</p>
}
```

mikeybinns:
Sure, that makes sense, thank you 😄

PR #7348, opened 2023-09-06
fix(remix-dev): allow emitting `serverBuildPath` as cjs

mcansh:
in my opinion this should be back ported into remix v1

see https://github.com/remix-run/indie-stack/pull/255#issuecomment-1707457850 for rational

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


MichaelDeBoey (approved):
> in my opinion this should be back ported into remix v1

I agree we should release this as `1.19.4`

mcansh:
closing in favor of #7180

PR #7354, opened 2023-09-06
Remove AppData/RouteHandle types

brophdawg11:
Removing types which are just aliases for unknown.

I left `AppLoadContext` in for now since it's used in the adapters but I could be convinced to just hardcode `unknown` in those and remove that as well.

Follow up of https://github.com/remix-run/remix/pull/7319#discussion_r1316373471

### DIFF
--- a/.changeset/chilled-horses-look.md
+++ b/.changeset/chilled-horses-look.md
@@ -1,0 +1,8 @@
+---
+"@remix-run/cloudflare": major
+"@remix-run/deno": major
+"@remix-run/node": major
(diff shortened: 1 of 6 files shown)


rossipedia:
>  I could be convinced to just hardcode unknown in those and remove that as well.

Could you be convinced to leave it alone? Extending the `AppLoadData` interface feels pretty crucial.

MichaelDeBoey:
@rossipedia As of #7319, `AppLoadContext` is no interface anymore, but a type alias

PR #7355, opened 2023-09-06
fix(remix-react): make `useMatches`'s type generic

MichaelDeBoey:
Follow-up of #7319


MichaelDeBoey:
Closing this as @brophdawg11 pointed out that this doesn't make sense to do

Not all matches will have the same `data` & `handle`, so people will probably do

```ts
let matches = useMatches();
let rootMatch = matches[0] as UIMatch<typeof rootLoader>;
let thisRouteMatch = matches[matches.length - 1] as UIMatch<typeof loader>;
```

PR #7360, opened 2023-09-07
Revert AppLoadContext back to an interface for module augmentation su…

brophdawg11:
Reverting the `AppLoadContext` portion of https://github.com/remix-run/remix/pull/7319 as we didn't realize folks use module augmentation to define that in v1 and changing it to `unknown` breaks that ability (this feature was originally added in https://github.com/remix-run/remix/pull/1876).  

This PR brings back the limitation of needing to return an object from `getLoadContext` (and thus prevents using something like a `Map` or a custom `class`), but this is a less-invasive change for v2 - and we can revisit a change to allow those in the future if the need arises.

### DIFF
--- a/packages/remix-server-runtime/routes.ts
+++ b/packages/remix-server-runtime/routes.ts
@@ -1,12 +1,12 @@
-import type { AgnosticDataRouteObject } from "@remix-run/router";
+import type {
+  AgnosticDataRouteObject,
+  LoaderFunctionArgs as RRLoaderFunctionArgs,
+  ActionFunctionArgs as RRActionFunctionArgs,
+} from "@remix-run/router";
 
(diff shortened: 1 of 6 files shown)


packages/remix-server-runtime/routes.ts
```diff
@@ -87,7 +87,9 @@ export function createStaticHandlerDataRoutes(
       id: route.id,
       path: route.path,
       loader: route.module.loader
-        ? (args: LoaderFunctionArgs) =>
+        ? // Need to use RR's version here to permit the optional context even
+          // though we know it'll always be provided in remix
+          (args: RRLoaderFunctionArgs) =>
```
brophdawg11, line 92: At some point it could make sense to change RR from `context?:  unknown` to `context: unknown | undefined` which would, I think, eliminate this typing oddity.

Issue #7357, opened 2023-09-07
bug(templates): default entry.server.tsx fails type check with "Argument of type 'PassThrough' is not assignable to parameter of type 'BodyInit | null | undefined'."


lpsinger:
### What version of Remix are you using?

2.0.0-pre.8

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Due to #7271, the default entry.server.tsx file fails type checking with the following error message from TypeScript:

`Argument of type 'PassThrough' is not assignable to parameter of type 'BodyInit | null | undefined'.`

### Expected Behavior

The file entry.server.tsx should have no TypeScript errors.

### Actual Behavior

See screen shot below from VS Code.
![Screenshot 2023-09-07 at 09 01 20](https://github.com/remix-run/remix/assets/728407/76520b9e-9ac7-484d-943c-98678c329533)

brophdawg11:
Yeah I've noticed this on a few local tests too - just haven't gotten around to PR-ing it yet.  The fix, I'm pretty sure, is `Readable.toWeb(body) as ReadableStream`

brophdawg11:
Fixed by #7361 - will be out in a prerelease shortly 👍

PR #7361, opened 2023-09-07
Fix TS error in entry-server - make Passthrough a ReadableStream

brophdawg11:
Now that we're strictly using web fetch/streams and we no longer export the `fetch` API out through `@remix-run/node` (https://github.com/remix-run/remix/pull/7293), we are also using the globally available `Response` constructor in `entry.server.tsx` (https://github.com/remix-run/remix/pull/7271).  TS doesn't allow a node `PassThrough` stream to be passed directly, so we need to convert it to a web stream using node's `Readable.toWeb()` utility.

Closes: #7357

### DIFF
--- a/packages/remix-dev/config/defaults/entry.server.node.tsx
+++ b/packages/remix-dev/config/defaults/entry.server.node.tsx
@@ -48,11 +48,12 @@ function handleBotRequest(
         onAllReady() {
           shellRendered = true;
           const body = new PassThrough();
+          const stream = createReadableStreamFromReadable(body);
 
           responseHeaders.set("Content-Type", "text/html");
 
           resolve(
-            new Response(createReadableStreamFromReadable(body), {
+            new Response(stream, {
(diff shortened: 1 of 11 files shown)


packages/remix-dev/config/defaults/entry.server.node.tsx
```diff
@@ -47,11 +47,12 @@ function handleBotRequest(
         onAllReady() {
           shellRendered = true;
           const body = new PassThrough();
+          const webBody = Readable.toWeb(body) as ReadableStream;
```
brophdawg11, line 50: This cast seems to be required due to the underlying generic, but I couldn't find a way to get them to align without the `as`.

```
Type 'ReadableStream<T>' is missing the following properties from type 'ReadableStream<ArrayBufferView | undefined>': 
  values, [Symbol.asyncIterator]ts(2345)
```

PR #7362, opened 2023-09-07
fix(remix-server-runtime): make `DataFunctionArgs` a type alias

MichaelDeBoey:
Follow-up of @brophdawg11's #7360

I don't think we want people to override `DataFunctionArgs` as they can already override `AppLoadContext`

### DIFF
--- a/.changeset/small-impalas-invent.md
+++ b/.changeset/small-impalas-invent.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+[REMOVE] make `DataFunctionArgs` a type alias
--- a/packages/remix-server-runtime/routeModules.ts
+++ b/packages/remix-server-runtime/routeModules.ts
@@ -17,13 +17,10 @@ export interface RouteModules<RouteModule> {
 }
 
 // Context is always provided in Remix, and typed for module augmentation support.
-// RR also doesn't export DataFunctionArgs so we extend the two interfaces here
+// RR also doesn't export DataFunctionArgs, so we extend the two interfaces here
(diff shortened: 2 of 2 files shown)


brophdawg11:
🤣 I can't keep up with this type versus interface stuff - we export a ton of both 🤷 

Wanna add a changeset so it triggers a prerelease?  I'll lump it in with https://github.com/remix-run/remix/pull/7361 tomorrow

MichaelDeBoey:
@brophdawg11 Basically we should almost always use type aliases _unless_ we want to support interface augmentation (which is almost never the case)

brophdawg11:
Would you want to work with @pcattori to get together some sort of decision doc or guidelines around this stuff?  From a quick count, I counted ~20 or so interfaces in the 45 exports from `@remix-run/server-runtime` so we don't have any sort of rule at the moment.  Is there any way this stuff could be enforced via a linter?  Otherwise we'll keep drifting off track over time 😕

MichaelDeBoey:
@brophdawg11 I'll create a PR that makes all of them a type alias + enable linting rules that enforce this 👍

RR should only do that as part of v7 as it would technically be a breaking change to make it a type alias

brophdawg11:
We can merge this since it aligns better with the `LoaderFunctionArgs`/`ActionFunctionArgs` types, but for the rest of the associated PRs I'm going to close them out for now since we need to first define/document a stance as a team before we go mass-updating the current code.  And then we need to figure out if we want to churn the TS API with `V3_` types in v2 or just document it as a find/replace when v3 releases.

PR #7363, opened 2023-09-07
feat(remix-server-runtime)!: use type aliases instead of interfaces

MichaelDeBoey:
As requested by @brophdawg11 in https://github.com/remix-run/remix/pull/7362#issuecomment-1710730616, this PR converts all interfaces (both internal + exported) to type aliases

### DIFF
--- a/packages/remix-server-runtime/cookies.ts
+++ b/packages/remix-server-runtime/cookies.ts
@@ -1,12 +1,16 @@
-import type { CookieParseOptions, CookieSerializeOptions } from "cookie";
+import type {
+  CookieParseOptions as RRCookieParseOptions,
+  CookieSerializeOptions as RRCookieSerializeOptions,
+} from "cookie";
 import { parse, serialize } from "cookie";
 
 import type { SignFunction, UnsignFunction } from "./crypto";
 import { warnOnce } from "./warnings";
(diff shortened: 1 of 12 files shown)


packages/remix-server-runtime/cookies.ts
```diff
@@ -1,12 +1,16 @@
...
+  CookieSerializeOptions as RRCookieSerializeOptions,
+} from "cookie";
 import { parse, serialize } from "cookie";
 
 import type { SignFunction, UnsignFunction } from "./crypto";
 import { warnOnce } from "./warnings";
 
-export type { CookieParseOptions, CookieSerializeOptions };
+export type CookieParseOptions = RRCookieParseOptions;
+export type CookieSerializeOptions = RRCookieSerializeOptions;
```
MichaelDeBoey, line 11: These are still interfaces in RR, so re-exporting them as type aliases

Once RR exports these as types, we can go back to

```
export type { CookieParseOptions, CookieSerializeOptions };
```

MichaelDeBoey:

This would be something like
```ts
export interface Cookie {
  // ...
}
export type V3_Cookie = Cookie;
```

brophdawg11:

PR #7366, opened 2023-09-07
feat(remix-cloudflare-pages)!: use type aliases instead of interfaces

MichaelDeBoey:
Just like I did in #7363

This would be something like
```ts
export interface CreatePagesFunctionHandlerParams {
  // ...
}
export type V3_CreatePagesFunctionHandlerParams = CreatePagesFunctionHandlerParams;
```


brophdawg11:

PR #7368, opened 2023-09-07
feat(remix-react)!: use type aliases instead of interfaces

MichaelDeBoey:
Just like I did in #7363, #7366 & #7367

This would be something like
```ts
export interface AwaitProps {
  // ...
}
export type V3_AwaitProps = AwaitProps;
```


brophdawg11:

PR #7371, opened 2023-09-07
chore: use type aliases instead of interfaces

MichaelDeBoey:
Just like I did in #7363, #7366, #7367, #7368, #7369 & #7370

These are all internal types, so we can just merge it as-is

### DIFF
--- a/scripts/unist.d.ts
+++ b/scripts/unist.d.ts
@@ -2,10 +2,10 @@ import type { Parent, Literal } from "unist";
 
 export type Node = FlowNode | PhrasingNode | RootNode;
 
-export interface RootNode extends Parent {
+export type RootNode = Parent & {
   type: "root";
   children: FlowNode[];
-}
+};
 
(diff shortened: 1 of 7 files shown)


scripts/unist.d.ts
```diff
@@ -17,46 +17,46 @@ export type FlowNode =
...
   url?: string;
-}
+};
 
-export interface TextNode extends Literal {
+export type TextNode = Literal & {
   type: "text";
   value: string;
-}
+};
```
rossipedia, line 62: FWIW intersection types apparently perform far worse than `interface / extends`:

https://twitter.com/mattpocockuk/status/1698642590193619025

Might not matter at Remix's current size, but probably something to be aware of as the codebase grows

brophdawg11: Agreed - this is why I want us to define/document our team stance before we go making these types of changes.  I'm going to close this for now and we'll get a TS-cleanup on the books.

brophdawg11:

PR #7374, opened 2023-09-08
fix(dev): update esbuild, tree shake CSS bundle

markdalgleish:
Since esbuild v0.17.7+ no longer tree shakes CSS and this means we're stuck on esbuild v0.17.6, this PR is an experiment to see if we can manually tree shake CSS by performing two esbuild passes during the CSS bundle build. The first esbuild pass is effectively the same as before, except now we keep all generated CSS files in memory. We then check which files were referenced in the JS output and use this as input to a 2nd esbuild pass containing only referenced CSS files.

This PR is currently against the `release-next` branch so that it's up to date with the latest compiler code, but if it lands it's likely to go in `dev`.


pcattori (approved):
LGTM! Let's wait to merge this until after v2

MichaelDeBoey (approved):

PR #7391, opened 2023-09-11
fix(dev): surface PostCSS config errors

markdalgleish:
We were swallowing errors when loading PostCSS config so that it didn't break the build for those without a PostCSS config, but this suppresses real errors for PostCSS consumers when there's a problem with their config.

This PR updates the logic so that we only swallow the missing config error and ensure that all other errors are thrown.

I've also refactored all CSS plugins so that the PostCSS config is loaded lazily rather than up-front in the `setup` hook. Without this change, consumers see the error coming from the first plugin that happens to be instantiated which is confusing if they're not using the feature, e.g. the error comes from the CSS Modules plugin when they're not using CSS Modules.

### DIFF
--- a/packages/remix-dev/compiler/utils/postcss.ts
+++ b/packages/remix-dev/compiler/utils/postcss.ts
@@ -64,7 +64,14 @@ export async function loadPostcssPlugins({
 
       plugins.push(...postcssConfig.plugins);
     } catch (err) {
-      // If they don't have a PostCSS config, just ignore it.
+      // If they don't have a PostCSS config, just ignore it,
+      // otherwise rethrow the error.
+      if (
+        err instanceof Error &&
+        !/No PostCSS Config found/i.test(err.message)
+      ) {
+        throw err;
+      }
     }
   }
 
--- a/.changeset/postcss-surface-errors.md
+++ b/.changeset/postcss-surface-errors.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Surface errors when PostCSS config is invalid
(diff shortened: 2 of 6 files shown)


packages/remix-dev/compiler/utils/postcss.ts
```diff
@@ -64,7 +64,14 @@ export async function loadPostcssPlugins({
 
       plugins.push(...postcssConfig.plugins);
     } catch (err) {
-      // If they don't have a PostCSS config, just ignore it.
+      // If they don't have a PostCSS config, just ignore it,
+      // otherwise rethrow the error.
+      if (
+        err instanceof Error &&
+        !/No PostCSS Config found/i.test(err.message)
```
brophdawg11, line 71: Is this an error we throw or a dependency throws?  The raw string matching feels somewhat fragile but if it's not our error then we may not have a choice.  If it is our error - we could move to a more explicit check?

brophdawg11 (approved):

packages/remix-dev/compiler/utils/postcss.ts
markdalgleish: Yep, this isn't our error. I'm not a fan of this either, but the alternative would be to manually check for a PostCSS config ourselves before trying to load the config. This would mean checking for the following files:

- `package.json` (`postcss` field)
- `.postcssrc`
- `.postcssrc.json`
- `.postcssrc.yml`
- `.postcssrc.js`
- `.postcssrc.mjs`
- `.postcssrc.cjs`
- `.postcssrc.ts`
- `postcss.config.js`
- `postcss.config.mjs`
- `postcss.config.cjs`
- `postcss.config.ts`

While we could do this, this feels fragile too since PostCSS could change their config resolution logic in the future and we'd be hard-coding the current state.

We could improve this with a change upstream to allow ignoring missing config, but I didn't want that to block this fix.

brophdawg11: 👍 makes sense to me!

PR #7416, opened 2023-09-14
docs(quickstart): Minor fixes to commands

brookslybrand:
Some really minor changes to the [quickstart docs on dev](https://remix.run/docs/en/dev/start/quickstart)

Two issues not addressed in the code:
1. There is no `tsconfig.json` file, yet we're using `.tsx` files. Should we include a step to add a `tsconfig.json` file such as [this one](https://github.com/remix-run/remix/blob/30bcd9e9dfbeda1707329da017f9ceccfac21521/docs/guides/typescript.md?plain=1#L49-L70). If so, I can add that instruction
2. When running `remix-serve build/index.js` (or rather `npx remix-serve build/index.js`) I get a syntax error due to the `import` statements. A simple `"type": "module"` in the `package.json` fixes this, but I wasn't sure if there's a more preferred recommendation for the quick start

### DIFF
--- a/docs/start/quickstart.md
+++ b/docs/start/quickstart.md
@@ -19,7 +19,7 @@ cd my-remix-app
 npm i @remix-run/node @remix-run/react @remix-run/serve isbot react react-dom
 
 # install dev dependencies
-npm i -D @remix-run/dev typescript
+npm i -D @remix-run/dev typescript @types/react
 ```
 
 ## The Root Route
@@ -66,7 +66,7 @@ export default function App() {
 First build the app for production:
 
 ```shellscript nonumber
-remix build
+npx remix build
 ```
 
 You should now see a `build/` folder (the server version of your app) and `public/build` folder (the browser version) with some build artifacts in them. (This is all [configurable][remix-config].)
@@ -75,7 +75,7 @@ You should now see a `build/` folder (the server version of your app) and `publi
 
 ```shellscript nonumber
 # note the dash!
-remix-serve build/index.js
+npx remix-serve build/index.js
 ```
 
 You should be able to open up [http://localhost:3000][http-localhost-3000] and see the "hello world" page.
@@ -233,7 +233,7 @@ Now you can work on your app with immediate feedback. Give it a shot, change the
 There are default magic files Remix is using that most apps don't need to mess with, but if you want to customize Remix's entry points to the server and browser you can run `remix reveal` and they'll get dumped into your project.
 
 ```shellscript nonumber
-remix reveal
+npx remix reveal
 ```
 
 ```
@@ -248,7 +248,7 @@ Congrats, you can add Remix to your resume! Summing things up, we've learned:
 - `remix build` and `remix dev` compile your app into two things:
   - A request handler that you add to your own JavaScript server
   - A pile of static assets in your public directory for the browser
-- You can bring your own server with adapters to deploy anywhere.
+- You can bring your own server with adapters to deploy anywhere
 - You can set up a development workflow with HMR built-in
 
 In general, Remix is a bit "guts out". A few minutes of boilerplate but now you own your stack.
--- a/docs/start/quickstart.md
+++ b/docs/start/quickstart.md
@@ -73,6 +73,17 @@ You should now see a `build/` folder (the server version of your app) and `publi
 
 💿 **Run the app with `remix-serve`**
 
+First you will need to specify the type in `package.json` as module so that `remix-serve` can run your app.
+
+```jsonc filename=package.json lines=[2] nocopy
+{
+  "type": "module"
+  // ...
+}
+```
+
+Now you can run your app with `remix-serve`:
+
 ```shellscript nonumber
 # note the dash!
 npx remix-serve build/index.js
--- a/docs/start/quickstart.md
+++ b/docs/start/quickstart.md
@@ -19,19 +19,19 @@ cd my-remix-app
 npm i @remix-run/node @remix-run/react @remix-run/serve isbot react react-dom
 
 # install dev dependencies
-npm i -D @remix-run/dev typescript @types/react
+npm i -D @remix-run/dev
 ```
 
 ## The Root Route
 
 ```shellscript nonumber
 mkdir app
-touch app/root.tsx
+touch app/root.jsx
 ```
 
 `app/root.jsx` is what we call the "Root Route". It's the root layout of your entire app. Here's the basic set of elements you'll need for any project:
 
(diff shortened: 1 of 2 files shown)


docs/start/quickstart.md
```diff
@@ -66,7 +66,7 @@ export default function App() {
 First build the app for production:
 
 ```shellscript nonumber
-remix build
+npx remix build
```
brookslybrand, line 69: Alternatively we could recommend adding these commands to the `package.json`'s `scripts`, but that seems like a lot more work.

If I'm missing something to get regular 'ol `remix build` to work on it's own, let me know! If so, it's likely something we should reference in this doc

docs/start/quickstart.md
```diff
@@ -19,7 +19,7 @@ cd my-remix-app
 npm i @remix-run/node @remix-run/react @remix-run/serve isbot react react-dom
 
 # install dev dependencies
-npm i -D @remix-run/dev typescript
+npm i -D @remix-run/dev typescript @types/react
```
brophdawg11, line 22: Been a bit since I did the tutorial, but does this need/want `@types/react-dom` as well?

```suggestion
npm i -D @remix-run/dev typescript @types/react @types/react-dom
```

docs/start/quickstart.md
brophdawg11: I think it's safe to use `npx` since w'ere using `npm` above.  `yarn`/`pnpm` users should know what to switch it to 👍

brophdawg11 (commented):
Looks good!  Do you mind rebasing this and repointing to `release-next`?  That way this update will land in `main` when we do the v2 stable and will be immediately available on `remix.run/docs` (I'll give you the rundown on our release processes and such when we chat tomorrow!)

docs/start/quickstart.md
brookslybrand: Oh yeah, good point! I'll update it

brookslybrand:
> Looks good! Do you mind rebasing this and repointing to `release-next`? That way this update will land in `main` when we do the v2 stable and will be immediately available on `remix.run/docs` (I'll give you the rundown on our release processes and such when we chat tomorrow!)

Rebased and re-pointed to `release-next`, thanks for showing me `--onto`!

Any opinion on the 2 questions in the description? Also happy to wait for Ryan to weigh in

docs/start/quickstart.md
MichaelDeBoey: This should work without adding `npx` once dependencies are installed
At least when using npm

MichaelDeBoey:
> 1. There is no `tsconfig.json` file, yet we're using `.tsx` files. Should we include a step to add a `tsconfig.json` file such as [this one](https://github.com/remix-run/remix/blob/30bcd9e9dfbeda1707329da017f9ceccfac21521/docs/guides/typescript.md?plain=1#L49-L70). If so, I can add that instruction

Ryan wanted to keep it as small as possible and this just works with all the defaults if you don't specify a `tsconfig.json`, so with those 2 things in mind, I would just leave it as-is tbh

docs/start/quickstart.md
brophdawg11: Huh?  How?

```sh
~/me/shopify/repos/playgrounds/remix-playground> remix build
-bash: remix: command not found
```

brophdawg11:
yeah agreed if the tsconfig isn't needed for the app to run, let's leave it out

> When running `remix-serve build/index.js` ... I get a syntax error

Yeah I think having them add `"type": "module"` to `package.json` makes sense if it fixes this error since in v2 the default `serverModuleFormat` is ESM

docs/start/quickstart.md
MichaelDeBoey: `remix` & `remix-serve` bins are installed, so they should be available to call normally 🤔

kiliman: `bin` scripts still require `npx` when installed locally. If installed globally, those scripts are added to the path and can be called like any other command.

brookslybrand: ☝️ 

pretty much the reason `npx` was created from my understanding

brookslybrand:
> Ryan wanted to keep it as small as possible and this just works with all the defaults if you don't specify a `tsconfig.json`, so with those 2 things in mind, I would just leave it as-is tbh

Sounds good, I'll defer to you all. Just to say my peace, even though it "runs", you see all the type errors if you're looking at the code in your editor. I've seen plenty of new folks start a tutorial and get immediately stressed out when they see the squigglies. Wish it wasn't the case, but just trying to eliminate any way people can think they did something wrong.

> Yeah I think having them add "type": "module" to package.json makes sense if it fixes this error since in v2 the default serverModuleFormat is ESM

Cool, [added some instructions](https://github.com/remix-run/remix/pull/7416/files#diff-03e14cd14612482bf0dfc4f8cae9e7d9a38df77f64b0332c89f9e10d7cb9b2c9R76-R86)

brophdawg11:
@brookslybrand yeah I agree with you that the red squiggles are distracting and off-putting for a first-timer.  In the interest of a 5-minute tutorial I'm not sure TS is doing anything for us (and `root.tsx` doesn't actually contain any TS).  It also looks like we flip flop too (maybe just a typo) between JSX and TSX:

Screenshot 2023-09-14 at 3 28 17 PM

I'll defer to @ryanflorence but my vote would be to drop the `typescript` dependency and just roll with `root.jsx` for this 5-minute "single file Remix app".  Our more involved tutorials and templates will still use TS if you decide to dip in more than just a toe.

docs/start/quickstart.md
brophdawg11: oh phew I thought I missed an `npm` behavior change recently - I've given up on keeping up with package manager evolutions 🤣

MichaelDeBoey: I just looked for about 2h into why this was working for me and found the culprit 😅
https://github.com/MichaelDeBoey/dotfiles/blob/0e92b1d9b524f100ca8ec574f971c5f18154c7d0/path.zsh#L8

docs/start/quickstart.md
brookslybrand: Ended up removing these again and `typescript` as well, since the quickstart should not have TypeScript`

brookslybrand:
Alright, updated all `.jsx` to `.tsx` and ran through all the commands again. Everything worked as expected! There are no known open issues on this PR from my standpoint

brophdawg11 (approved):
Nice, looks good to merge!

PR #7420, opened 2023-09-14
docs(start/tutorial): update APIs and line numbers

brookslybrand:
Mostly just updating
- `LoaderArgs` -> `LoaderFunctionArgs`
- `ActionArgs` -> `ActionFunctionArgs`

Otherwise just fixing some line highlighting. Had a few comments/questions I'll call out in code comments

### DIFF
--- a/docs/start/tutorial.md
+++ b/docs/start/tutorial.md
@@ -36,7 +36,7 @@ npm install
 npm run dev
 ```
 
-You should now see an unstyled screen that looks like this:
+You should be able to open up [http://localhost:3000][http-localhost-3000] and see an unstyled screen that looks like this:
 
 <img class="tutorial" src="/docs-images/contacts/03.webp" />
 
@@ -496,14 +496,16 @@ export default function Contact() {
 
 TypeScript is very upset with us, let's make it happy and see what that forces us to consider:
 
-```tsx filename=app/routes/contacts.$contactId.tsx lines=[1,3,7-8]
-import type { LoaderArgs } from "@remix-run/node";
+```tsx filename=app/routes/contacts.$contactId.tsx lines=[1,3,7-10]
+import type { LoaderFunctionArgs } from "@remix-run/node";
 // existing imports
 import invariant from "tiny-invariant";
 
 // existing imports
 
-export const loader = async ({ params }: LoaderArgs) => {
+export const loader = async ({
+  params,
+}: LoaderFunctionArgs) => {
   invariant(params.contactId, "Missing contactId param");
   const contact = await getContact(params.contactId);
   return json({ contact });
@@ -518,10 +520,12 @@ Next, the `useLoaderData<typeof loader>()` now knows that we got a contact or `n
 
 We could account for the possibility of the contact being not found in component code, but the webby thing to do is send a proper 404. We can do that in the loader and solve all of our problems at once.
 
-```tsx filename=app/routes/contacts.$contactId.tsx lines=[6-8]
+```tsx filename=app/routes/contacts.$contactId.tsx lines=[8-10]
 // existing imports
 
-export const loader = async ({ params }: LoaderArgs) => {
+export const loader = async ({
+  params,
+}: LoaderFunctionArgs) => {
   invariant(params.contactId, "Missing contactId param");
   const contact = await getContact(params.contactId);
   if (!contact) {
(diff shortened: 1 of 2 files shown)


brookslybrand:
[Check the other tutorials for more in-depth examples.](https://github.com/remix-run/remix/blob/release-next/docs/start/tutorial.md?plain=1#L8)

Any link we should/could be providing here?

brookslybrand:
[Remix sends a 405 because there is no code on the server to handle this form navigation.](https://github.com/remix-run/remix/blob/release-next/docs/start/tutorial.md?plain=1#L552)

I was unable to get this to show. I just got an error

image

docs/start/tutorial.md
```diff
@@ -496,14 +496,16 @@ export default function Contact() {
 
 TypeScript is very upset with us, let's make it happy and see what that forces us to consider:
 
-```tsx filename=app/routes/contacts.$contactId.tsx lines=[1,3,7-8]
-import type { LoaderArgs } from "@remix-run/node";
+```tsx filename=app/routes/contacts.$contactId.tsx lines=[1,3,7-10]
+import type { LoaderFunctionArgs } from "@remix-run/node";
```
brophdawg11, line 500: Ah good catch, I should have done a search for these when we did the renames the other week 👍 .  I'm going to do a separate quick PR (probably tomorrow AM) to find/replace this elsewhere in the docs.

brophdawg11:
> Check the other tutorials

This _used_ to refer to https://remix.run/docs/en/main/tutorials/jokes but I'm not sure where that landed in the new docs setup though.

> Remix sends a 405

I wonder if that's just mis-wording.  When the client knows the server doesn't have an `action` for the route it doesn't even bother with the request and just rejects with an `Error` (not an `ErrorResponse` that would contain a 405 status code).  If it did work that way at one point it would have been pre `1.10.0` when we were still on the old `TransitionManager` so I can take a quick peek back through the old code tomorrow and see if we used to send a 405 status at all or if we always just raised a generic `Error`

brookslybrand:
> I wonder if that's just mis-wording. 

Gotcha, I should have been more specific too. The tutorial has a screenshot of a 405 page that doesn't exist

image

brookslybrand:
One thing I wanted to highlight (and I don't think there's really anything to do about it in the docs) is the fact that if you create a new contact and hit save in the editor, the hot data reloading (I assume) breaks Ryan's database he created. You can see it in the second half of this video. This is a pretty common flow users will take going through the tutorial, so might in some unexpected frustration.

https://github.com/remix-run/remix/assets/12396812/fe096022-5bdf-407f-873d-3c22507d17f0

@pcattori I know you asked me to share if anything with the compiler felt off.

pcattori:
> One thing I wanted to highlight (and I don't think there's really anything to do about it in the docs) is the fact that if you create a new contact and hit save in the editor, the hot data reloading (I assume) breaks Ryan's database he created. You can see it in the second half of this video. This is a pretty common flow users will take going through the tutorial, so might in some unexpected frustration.
> 
> @pcattori I know you asked me to share if anything with the compiler felt off.

Yea without `--manual` mode enabled, the server restarts and loses in-memory state. [Wrote some docs on how to get this all working with `--manual` mode](https://remix.run/docs/en/dev/guides/manual-mode#keeping-in-memory-server-state-across-rebuilds). Since `remix-serve` is compatible with `--manual` out-of-the-box (i.e. you can steps 1-4 to setup manual mode), this should work:

1. Change the [template `dev` script](https://github.com/ryanflorence/remix-tutorial-template/blob/main/package.json#L6) to: `remix dev --manual`
2. Use the [`remember` utility](https://remix.run/docs/en/dev/guides/manual-mode#keeping-in-memory-server-state-across-rebuilds) to wrap the in-memory database

brookslybrand:
Perfect, thanks for the thorough explanation @pcattori. Even more reason I need to read your manual mode doc!

Thanks for outlining some steps to avoid this issue in the tutorial. Nice to know it'd be work that can be done outside the actual content of the tutorial. That being said, I don't see this being a blocker for this PR.

brophdawg11:
The 405 looks like a small regression from pre React Router 6.4 integration.  We used to actually send the HTTP request, knowing it was going to fail, and got back a 405.  We made a conscious decision to short circuit and stop sending the useless request - but inadvertently lost the 405 status.  We just throw an `Error` but would be better off throwing an `ErrorResponse` to maintain the semantic nature of the 405 and would get us back to that UI in the screenshot.  Let's leave the tutorial as-is for now and I'll open a PR to dev to fix that (which can be released shortly after in a 2.0.1)

brookslybrand:
>  Let's leave the tutorial as-is for now and I'll open a PR to dev to fix that (which can be released shortly after in a 2.0.1)

Sounds good to me! Thanks for the thorough explanation

docs/start/tutorial.md
brophdawg11: https://github.com/remix-run/remix/pull/7426

brookslybrand:
> [Check the other tutorials for more in-depth examples.](https://github.com/remix-run/remix/blob/release-next/docs/start/tutorial.md?plain=1#L8)
> 
> Any link we should/could be providing here?

Removed this reference since there are no other tutorials we want to reference here.

pcattori:
> Perfect, thanks for the thorough explanation @pcattori. Even more reason I need to read your manual mode doc!
> 
> Thanks for outlining some steps to avoid this issue in the tutorial. Nice to know it'd be work that can be done outside the actual content of the tutorial. That being said, I don't see this being a blocker for this PR.

Agree, not a blocker for this PR. Here's the [PR for changes to the tutorial template](https://github.com/ryanflorence/remix-tutorial-template/pull/2)

brophdawg11 (approved):

Issue #7457, opened 2023-09-18
Node version in cf-pages template is outdated


kishanhitk:
### What version of Remix are you using?

2.0.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. Create a new remix app with the cf-pages template
2. Check the `.node-version` file, it has `16.13.0`
3. Try to deploy it on cf-pages, and you will get this error.
![image](https://github.com/remix-run/remix/assets/41117038/16fb62d4-9831-4fd6-b934-d9d9e14d6470)

### Expected Behavior

The node version should be the current LTS: 18.17.1

### Actual Behavior

The node version is outdated: 16.13.0

PR #7458, opened 2023-09-18
templates(cloudflare-pages): update `.node-version`

kishanhitk:
Closes: #7457

- [ ] Docs
- [ ] Tests

Testing Strategy:
1. Create a new template with cf-pages, and it will have the updated node js version


templates/cloudflare-pages/.node-version
```diff
@@ -1 +1 @@
-16.13.0
\ No newline at end of file
+18.17.1
```
MichaelDeBoey, line 1:
```suggestion
18.0.0
```

MichaelDeBoey (changes requested):

Issue #7438, opened 2023-09-15
`type { UIMatch } from '@remix-run/router'` is different from `type { UIMatch } from '@remix-run/react'`


lensbart:
### What version of Remix are you using?

2.0.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Compare both exports and observe that they are different:
* `UIMatch` from `@remix-run/router` takes two type arguments
* `UIMatch` from `@remix-run/react` takes one type argument

Probably the former should be used in both cases, since it’s more flexible.

### Expected Behavior

Both types are the same

### Actual Behavior

`!(`both types are the same`)`

lensbart:
Additionally, it would be nice if other exported types (e.g. `LoaderFunction`) would be exported from just one Remix package (i.e., not from both `@remix-run/node` and `@remix-run/router`), if that makes sense.

brophdawg11:
> It would be nice if other exported types (e.g. LoaderFunction) would be exported from just one Remix package

Only exporting from one package would make the layered approach of building Remix on top of React Router impossible.  `@remix-run/router` should be considered an implementation dependency so generally speaking you shouldn't be importing anything from there and shouldn't need it in your `package.json`.  If there are aspects you need in your application that are not exposed through something like `@remix-run/react` or `@remix-run/node` then please file a separate issue for those and we can see if something should be re-exported through a Remix-specific package.

lensbart:
@brophdawg11 nice — thanks for that suggestion. `UIMatch` is the only reason I currently installed the `@remix-run/router` dependency.

brophdawg11:
This is resolved by #7464 and will be available in the next release 👍

PR #7464, opened 2023-09-18
Add second generic to UIMatch for handle field

brophdawg11:
Add second `handle` generic to `UIMatch` to match the React Router version

Closes: https://github.com/remix-run/remix/issues/7438

### DIFF
--- a/.changeset/add-uimatch-handle.md
+++ b/.changeset/add-uimatch-handle.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Add second generic to `UIMatch` for `handle` field
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -52,6 +52,7 @@ import type {
   MetaDescriptor,
   MetaMatch,
   MetaMatches,
+  RouteHandle,
 } from "./routeModules";
 
 function useDataRouterContext() {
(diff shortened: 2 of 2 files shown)


lensbart:
I just tried it out, but it doesn’t work entirely satisfactory with things like `remix-typedjson`, since this is typed as `export type UIMatch<D = AppData, H = RouteHandle> = UIMatchRR<SerializeFrom<D>, H>`, which imposes `SerializeFrom` whereas using `remix-typedjson` the equivalent would be `RemixSerializedType`.

Would it be possible to just reexport `UIMatch` from `@remix-run/router` instead?

Thanks!

kiliman:
Since `remix-typedjson` changes the data that is returned, in addition to the types, it's possible some of the utility functions/types will need to be added to my package.

brophdawg11:
> Would it be possible to just reexport UIMatch from `@remix-run/router` instead?

If you need that to bypass the Remix typings - you can import that directly from `@remix-run/router`, but I believe the current Remix type is correct for the majority of Remix use cases.

lensbart:
@brophdawg11 that works for me. Still, I wonder if my use case (using `remix-typedjson` and typing the `export const handle` using the `loader` type) is too rare to care about the ergonomics. Either way, I’m fine with it, just some food for thought.

See also https://github.com/remix-run/remix/issues/7518#issuecomment-1734568909

kiliman:
@lensbart can you add an issue in `remix-typedjson` with an example that's failing?

Thanks!

PR #7482, opened 2023-09-19
Update arc template's deploy command

ngtrthinh169TPC:
This should help clarify the command for people that takes docs literally (or a bit dull, like me), since `arc` is not a command and you have to add `npx` to it. The full command is documented on [arc's quick start docs](https://arc.codes/docs/en/get-started/quickstart)

### DIFF
--- a/templates/arc/README.md
+++ b/templates/arc/README.md
@@ -35,7 +35,7 @@ If you make it through all of that, you're ready to deploy!
 2. Deploy with `arc`
 
    ```sh
-   arc deploy production
+   npx arc deploy production
    ```
 
 You're in business!


CONTRIBUTING.md
```diff
@@ -1 +1 @@
-Please see [our guide to contributing](docs/pages/contributing.md).
+Please see [our guide to contributing](docs/guides/contributing.md).
```
brophdawg11, line 1: I'm going to revert this change since it's already been fixed in `main` and is causing a conflict at the moment

```suggestion
Please see [our guide to contributing](docs/pages/contributing.md).
```

brophdawg11:
Thanks for the update!

PR #7491, opened 2023-09-19
fix(dev): fix MDX types when using pnpm

### LINKED ISSUE #7448: Remix v2 causes type errors for `mdx` components

2.0

- [X] Yes

### Steps to Reproduce

1. Using `create-remix` CLI, create an new project
2. Create a directory like `app/routes/example` within the project
3. Create the `mdx` file `app/routes/example/Foo.mdx` like:

```mdx
## Hello World
```

4. Import the new `mdx` file

```tsx
// app/routes/example/route.tsx
import Foo from './Foo.mdx'

export default function Route() {
  return <Foo /> // causes "JSX element type 'Content' does not have any construct or call signatures." error
}
```

### Expected Behavior

No Typescript error

### Actual Behavior

A type error

```
'Foo' cannot be used as a JSX component.
  Its type 'typeof import("*.mdx")' is not a valid JSX element type.ts(2786)
```

markdalgleish:
Fixes #7448

This issue doesn't surface when `@types/mdx` is hoisted to the top-level in `node_modules`, which is the behaviour of the official npm client, but pnpm is stricter about transitive dependencies and ensures that packages aren't hoisted to the top-level by default. This means that Remix consumers don't automatically get the global `*.md`/`*.mdx` types from `@types/mdx` in their app code and only see our augmented `attributes` and `filename` exports. To fix this, we now explicitly import the global types as part of the augmentation.

I've tested this change against a real app running both npm and pnpm to ensure this works across both, and I've confirmed that this change fixed the repro shared in the linked issue.

### DIFF
--- a/.changeset/fix-mdx-types-for-pnpm.md
+++ b/.changeset/fix-mdx-types-for-pnpm.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Fix types for MDX files when using pnpm
--- a/packages/remix-dev/modules.ts
+++ b/packages/remix-dev/modules.ts
@@ -69,12 +69,12 @@ declare module "*.jpg" {
   export default asset;
 }
 declare module "*.md" {
-  import "mdx/types";
+  import "mdx";
   export let attributes: any;
   export let filename: string;
 }
 declare module "*.mdx" {
-  import "mdx/types";
+  import "mdx";
   export let attributes: any;
   export let filename: string;
 }


MichaelDeBoey (approved):

bastiankistner:
I'm seeing the same issue with `RemixBrowser` and `Outlet` and probably others. Tried to switch from installing node_modules with pnpm to installing them with bun and out of a sudden, the issue is gone. Would be great if it could be package manager agnostic.

PR #7495, opened 2023-09-20
fix(remix-node): relax typings for createReadableStreamFromReadable

estyrke:
This enables NodeJS.ReadWriteStream to be used with it.

- [ ] Docs
- [ ] Tests

Testing Strategy:

This allows the Chakra UI example at https://github.com/remix-run/examples/blob/main/chakra-ui/ to be converted to v2 according to https://remix.run/docs/en/main/start/v2#removal-of-exported-polyfills

That said, I have only tested this by copy/pasting the code into my local entry.server.tsx.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -149,6 +149,7 @@
 - emzoumpo
 - eps1lon
 - esamattis
+- estyrke
 - evanwinter
 - everdimension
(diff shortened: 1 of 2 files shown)


nmackey:
would be great to get this merged

PR #6354, opened 2023-05-09
feat(remix-dev): add support for `mdx`' `providerImportSource` config

Jokcy:
- [ ] Docs
- [ ] Tests

Testing Strategy:

I directly modify my remix project node_modules

```js
let compiled = await xdm.compile(fileContents, {
      jsx: true,
      jsxRuntime: "automatic",
      rehypePlugins,
      remarkPlugins,
      providerImportSource: '@mdx-js/react'
    });
```

It work fine and MDXProvider worked.

I did not find tests file for this plugin, if unit tests is required I will try to add then.

PR #7502, opened 2023-09-20
feat(dev): support overriding components for Markdown elements

lpsinger:
You can use an [MDX provider](https://mdxjs.com/docs/using-mdx/#mdx-provider) to override which React components are used for standard Markdown elements. For example, suppose that for all routes under the `/products` path you want to replace all Markdown links with [`<Link>`](../components/link) components instead of HTML `<a>` elements. Create the following route module `app/routes/products.tsx`:

```tsx filename=app/routes/products.tsx
import { MDXProvider } from "@mdx-js/react";
import { Link, Outlet } from "@remix-run/react";

export default function Products() {
  return (
    <MDXProvider
      components={{
        a: ({ children, href, ref, ...props }) =>
          href ? (
            <Link to={href} {...props}>
              {children}
            </Link>
          ) : (
            <a {...props}>{children}</a>
          ),
      }}
    >
      <Outlet />
    </MDXProvider>
  );
}
```

Then `app/routes/products.hammers.mdx` might look like this:

```mdx filename=app/routes/products.hammers.mdx
# Hammers

If you like our hammers, then you should also check out our
[nails](/products/nails)!
```

- [x] Docs
- [x] Tests

Testing Strategy: Modified an existing integration test

### DIFF
--- a/.changeset/beige-experts-rest.md
+++ b/.changeset/beige-experts-rest.md
@@ -1,0 +1,7 @@
+---
+"integration-tests": minor
+"@remix-run/dev": minor
+"@remix-run/server-runtime": minor
+---
+
+Support overriding the React components for standard Markdown elements in MDX files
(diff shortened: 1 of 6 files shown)


lpsinger:
Note that this is to some extent a duplicate of #6354, although I prefer this implementation because it requires less configuration to override components.

lpsinger:
I rebased this. Could I get a review please?

lpsinger:
Rebased again. Sorry to keep pestering, I know maintainers are busy! Could I get a review please?

PR #7507, opened 2023-09-21
Don't attempt to fetch assets in the Cloudflare handler on POST/PUT r…

zwily:
…equests

While implementing file uploads in a Cloudflare app, I discovered that `await request.blob()` would often hang on larger files. By testing a lot of combinations of remix+wrangler, I discovered that this only started happening with wrangler3, which enables the workerd backend by default.

I discovered that I could avoid all the hangs by skipping this asset fetch for POST/PUT requests. Something about the request cloning must be triggering some sort of bug in wrangler/workerd. Plus, it also seems like we shouldn't bother trying to fetch a static asset on a POST/PUT.


packages/remix-cloudflare-pages/worker.ts
```diff
@@ -53,16 +53,19 @@ export function createPagesFunctionHandler<Env = any>({
...
-        context.request.url,
-        context.request.clone()
-      );
-      response =
-        response && response.status >= 200 && response.status < 400
-          ? new Response(response.body, response)
-          : undefined;
-    } catch {}
+    let method = context.request.method;
+    if (method !== "POST" && method !== "PUT") {
```
ngbrown, line 57: Probably would be better to positively match `"GET"` and `"HEAD"` rather than exclude.

zwily: Yeah, I wasn't possible that there wouldn't be some other verb that might be used legitimately.. like OPTIONS, maybe? Probably not. I can change it.

zwily: I changed it to check for GET/HEAD/OPTIONS.

PR #7531, opened 2023-09-25
fix(docs): entry server defaults broken links

fredericoo:
Broken links for default entry.server files ◡̈


brophdawg11 (approved):
Thanks!

PR #7536, opened 2023-09-25
Fix comment on default for serverModuleFormat

ngbrown:
Since the default for `serverModuleFormat` is now `esm`, the comments in the v2 guide for the different configurations of `serverBuildTarget` were incorrect.

- [x] Docs

### DIFF
--- a/docs/start/v2.md
+++ b/docs/start/v2.md
@@ -758,7 +758,7 @@ module.exports = {
   serverBuildPath: "server/index.js",
   serverMainFields: ["main", "module"], // default value, can be removed
   serverMinify: false, // default value, can be removed
-  serverModuleFormat: "cjs", // default value, can be removed
+  serverModuleFormat: "cjs", // default value in 1.x, add before upgrading
   serverPlatform: "node", // default value, can be removed
 };
 ```
@@ -774,7 +774,7 @@ module.exports = {
   serverDependenciesToBundle: "all",
   serverMainFields: ["browser", "module", "main"],
   serverMinify: true,
-  serverModuleFormat: "esm",
+  serverModuleFormat: "esm", // default value in 2.x, can be removed once upgraded
   serverPlatform: "neutral",
 };
 ```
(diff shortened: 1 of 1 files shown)


brophdawg11:
These comments are tricky since they are intended (originally) for v1 users preparing for v2.  So CJS is still the default on v1.  Now that v2 is released, ESM is the default but there are still 2 types of users reading these docs:

* Users upgrading to `1.19.x` and adopting all future flags to smooth the v2 upgrade
  * Current comments are accurate for these folks 
* Users going straight to v2 without the in-between step
  * Your proposed new comments are accurate for these folks

I think it's probably best to either remove these specific comments or change them to `default value in 1.x`/`default value in 2.x` for clarity?

ngbrown:
> These comments are tricky since they are intended (originally) for v1 users preparing for v2. So CJS is still the default on v1. Now that v2 is released, ESM is the default but there are still 2 types of users reading these docs:
> 
> * Users upgrading to `1.19.x` and adopting all future flags to smooth the v2 upgrade
>   
>   * Current comments are accurate for these folks

Yes, I considered that, but the document is, at some level, upgrading a user to v2, not to v1.9 with the v2 flags. For users that are upgrading to v1.9 with all flags enabled, would be best served to use the v1.9 documentation.

brophdawg11:
> but the document is, at some level, upgrading a user to v2

If the document were _only_ intended to be read by users upgrading straight to v2, then it wouldn't include the steps of telling them to set `future.v2_*` flags to true since those flags don't exist in v2.  Successfully or not, at the moment it is intentionally trying to be something consumable by both sets of users I mentioned above.

> would be best served to use the v1.19 documentation

In an ideal world, yes.  But in practice these users are going to land on the `main` docs and the vast majority of them won't know/remember to switch over to the v1.x docs, so we have to be pragmatic.

machour (changes requested):

Seconding Matt's suggestion here, using `default value in 1.x/default value in 2.x` would help avoid any confusion

ngbrown:
I updated the comments to the following:
```ts
serverModuleFormat: "cjs", // default value in 1.x, add before upgrading
...
serverModuleFormat: "esm", // default value in 2.x, can be removed once upgraded
```

brophdawg11 (approved):

PR #7539, opened 2023-09-26
Add information about exporting ErrorBoundary from Resource Routes

kasprownik:
This information would save my team a lot of debugging after upgrading remix, so it would be great to document this for other folks.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -585,3 +585,4 @@
 - robbtraister
 - TrySound
 - rogepi
+- kasprownik
--- a/docs/guides/resource-routes.md
+++ b/docs/guides/resource-routes.md
@@ -65,8 +65,8 @@ When the user clicks the link from the UI route, they will navigate to the PDF.
 Resource Route will change its behavior when you export `ErrorBoundary` along with the loader. 
 Remix will process it as a classic layout-less container route.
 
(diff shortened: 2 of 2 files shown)


kiliman:
You may want to clarify _why_ Remix behaves this way. Resource routes are not supposed to have a UI. Whereas an `ErrorBoundary` is explicitly UI for rendering errors. I think that's why Remix transitions to a UI route.

However, technically I think this should be considered an error. If the route doesn't export a default function (the UI component) but does export other UI-related exports, then it should return a build-time error.

kasprownik:
Thanks @kiliman, added it.

kasprownik:
> If the route doesn't export a default function (the UI component) but does export other UI-related exports, then it should return a build-time error.

@kiliman there is a specific use case when it's valid, I provided it in this change.

brophdawg11:
I don't think we need a whole section in the doc about `ErrorBoundary`, but instead just need to make it more explicit.  Right now the docs state:

> If a route doesn't export a default component, it can be used as a Resource Route

Could we extend that to be:

> If a route doesn't export a `default` component or an `ErrorBoundary`, it can be used as a Resource Route


If we want to include the specific error message for reference, then maybe that would be better suited as an entry in https://remix.run/docs/en/main/guides/gotchas?

Issue #7542, opened 2023-09-26
templates/express always loads initial build


nullday0:
### What version of Remix are you using?

2.0.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

`npx create-remix --template remix-run/remix/templates/express`

### Expected Behavior

When I change any file, HMR works correctly and page reloads with new build.
But if I refresh the page (F5) then it will show me the initial build.

### Actual Behavior

It should show the latest build every time I refresh the page.

nullday0:
I was going through the source code here - https://github.com/remix-run/remix/blob/main/templates/express/server.js#L80
And I saw this piece of code:
```
async function createDevRequestHandler(initialBuild) {
  let build = initialBuild;
...
```
The thing is that express will call createDevRequestHandler() on every new request, so it will reset build variable to initialBuild every time.

I changed that behaviour, so I made
```
const initialBuild = await reimportServer();
let currentBuild = initialBuild
```
and then 
```
async function handleServerUpdate() {
    // 1. re-import the server build
    currentBuild = await reimportServer();
    // 2. tell Remix that this app server is now up-to-date and ready
    broadcastDevReady(currentBuild);
  }
```
and also changed `build` to `currentBuild` in other places, and now it works as expected.

I'm not sure how it was supposed to work with current code in the repo. Please let me know what I am missing here.

ngbrown:
I can confirm this issue. It also prevents updated actions from using the new version. It was a mistake in the template to call `createDevRequestHandler` repeatedly, it should just be called once. To that end, I opened a pull request at #7544.

danielweinmann:
I can confirm that @ngbrown's solution on #7544 works. Thanks, @ngbrown 🚀

PR #7544, opened 2023-09-26
Fix express template to use re-loaded build in requests

ngbrown:
Closes: #7542

The template was re-importing the `initialBuild` on every request, so HMR would appear to work after changes, but any full request would not update. Actions would also not update.

The documentation for [Manual Mode](https://remix.run/docs/en/main/guides/manual-mode) is not affected because it wasn't trying to dynamically import `chokidar` (so it can be a dev dependency).

This is against the `main` branch because templates usage is pulled directly from the repo and not released as packages.

**Testing Strategy:**

I opened my windows machine and ran this script:

```
npx create-remix@latest --template remix-run/remix/templates/express
```

I updated the `server.js` with these changes and then ran:

```
cd my-test
npm run dev
```

### DIFF
--- a/templates/express/server.js
+++ b/templates/express/server.js
@@ -37,18 +37,15 @@ app.use(express.static("public", { maxAge: "1h" }));
 
 app.use(morgan("tiny"));
 
-app.all("*", async (...args) => {
-  if (process.env.NODE_ENV === "development") {
-    const handler = await createDevRequestHandler(initialBuild);
-    return handler(...args);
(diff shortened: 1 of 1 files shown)


danielweinmann:
This change fixed it for me 🚀 thanks, @ngbrown!

michaellopez:
For us too! Thank you @ngbrown ⭐

pcattori (approved):

PR #7550, opened 2023-09-27
docs(discussion/form-vs-fetcher): fix user avatar details popup example filename

craigayre:
Update `docs/discussions/form-vs-fetcher.md` user avatar details popup example route filename to match the one referenced in the code.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -98,6 +98,7 @@
 - confix
 - coryhouse
 - courtyenn
+- craigayre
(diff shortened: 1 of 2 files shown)


brophdawg11:
Looks like there's a conflict in `contributors.yml` - do you mind getting that resolved?

craigayre:
@brophdawg11 should be resolved now, thanks for reviewing!

PR #7552, opened 2023-09-27
Remove explicit `process.exit(0)` from cli.ts

rossipedia:
Calling `process.exit(0)` in the successful case is redundant, but also causes problems when wanting to use the CLI commands in any kind of automated capacity, if the output is larger than 8192 bytes. This is because the Promise resolves _before_ the stdout buffers have been flushed, and when `process.exit(0)` is called, the entire process terminates leaving buffered output unsent to the calling process.

This is usually not a problem when running the CLI directly, but can be problematic when invoked via a child process. For instance, @brophdawg11 supplied this sample code in [Discord](https://discord.com/channels/770287896669978684/940264701785423992/1156614372433608734):

```js
// tmp.js
const child_process = require('child_process')
const { matchRoutes } = require("react-router-dom")
const routes = JSON.parse(
  child_process.execSync("npx remix routes --json").toString()
)
const matches = matchRoutes(routes, "/")
```

However, attempting to run this on a project with a large number of routes, resulting in a JSON object larger than 8192 bytes, results in a deserialization error:

```sh
❯ node tmp.js
undefined:160
                "

SyntaxError: Unterminated string in JSON at position 8192
```
The default case for a node script that exits gracefully is an exit code of `0`. So, removing the explicit `process.exit(0)` leaves the process running until all I/O is completed and STDOUT buffers are flushed.

### Testing Strategy:

I tested this using the following patch on an installed version of `@remix-run/dev`, which resulted in letting the above JS code run successfully and parse the full JSON output (which is larger than 8192 bytes):

```diff
diff --git a/dist/cli.js b/dist/cli.js
index a7df897f9b2e3efee0f8d154dae72fc7450ca937..bd3d4462ebdbaa8ed89d1e9ec8ec0ce46cad7328 100644
--- a/dist/cli.js
+++ b/dist/cli.js
@@ -13,9 +13,7 @@
 
 var index = require('./index');
 
-index.cli.run().then(() => {
-  process.exit(0);
-}, error => {
+index.cli.run().catch(error => {
   if (error) console.error(error);
   process.exit(1);
 });
```

### DIFF
--- a/.changeset/soft-carrots-smile.md
+++ b/.changeset/soft-carrots-smile.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Stop calling `process.exit(0)` on successful CLI runs to avoid exiting prior to flushing the stdout buffers
--- a/packages/remix-dev/cli.ts
+++ b/packages/remix-dev/cli.ts
@@ -1,11 +1,6 @@
 import { cli } from "./index";
 
-cli.run().then(
-  () => {
-    process.exit(0);
-  },
-  (error: unknown) => {
-    if (error) console.error(error);
-    process.exit(1);
-  }
-);
+cli.run().catch((error: unknown) => {
+  if (error) console.error(error);
+  process.exit(1);
+});


brophdawg11:
Do you mind adding a patch changeset for `@remix-run/dev`?  Something like:

`Stop calling `process.exit(0)` on successful CLI runs to avoid exiting prior to flushing the stdout buffers`

rossipedia:
@brophdawg11 done!

brophdawg11:
Hm, I'm not sure what's going on with the CI tests here - it seems the playwright tests never kick off 😕 

Screenshot 2023-10-02 at 11 26 49 AM

I was able to reproduce locally via `yarn test:integration loader --project chromium` (which will run a small subset of the E2E tests) and it just hangs.  We'll need to dig a bit deeper into the test harness stuff and see why the removal of that is causing an issue.

brophdawg11:
@rossipedia Is this no longer an issue for you?  I was meaning to get back and dig into the CI issue here when we got through some of the recent work.

rossipedia:
🤦 No I accidentally deleted the repo when I was cleaning out my GH account. Lemme take a look. I'm at a point where `remix routes --json` on a relatively large app would be very useful.

Issue #3768, opened 2022-07-16
Source map URL replace does not work on cloudflare-pages build target.


fabiankaestner:
### What version of Remix are you using?

1.6.4

### Steps to Reproduce

Setup a cloudflare-workers remix app.
Start the application in development.
Provoke a stacktrace.

### Expected Behavior

I have sourcemaps available.

### Actual Behavior

I do not have sourcemaps, instead I get the stacktraces from the generated [[path]].js file.

fabiankaestner:
This happens because the server component generated for a cloudflare-pages deployment are named [[path]].js. The regex that is used for rewriting the source map URL doesn't account for unescaped [] in the filename, and thus fails to replace the URL correctly as implemented in PR #2065.
The problem is here:
https://github.com/remix-run/remix/blob/f752f8256ed5fba35b7878f7e14ef137968a115c/packages/remix-dev/compiler.ts#L514-L518
we should be escaping the [], for example like this
```ts
let escapedFilename = filename
    .replace(/\./g, "\\.")
    .replace(/\[/g, "\\[")
    .replace(/\]/g, "\\]");
let pattern = `(//# sourceMappingURL=)(.*)${escapedFilename}`;
let contents = Buffer.from(file.contents).toString("utf-8");
contents = contents.replace(new RegExp(pattern), `$1${filename}`);
await fse.writeFile(file.path, contents);
```

I'd be happy to do a PR with the change, however I'm unable to write a unit test for it. To be able to write a unit test for this, I'd have to be able to provoke the build artifacts naming to be like a cloudflare-pages build target.
Maybe someone that knows how the remix test work could help me out there?

kiliman:
Ah, thanks. At the time I wrote that, I hadn't seen any other patterns.

kiliman:
If your change works for both cloudflare and other adapters, I'd go ahead and submit the PR.

fabiankaestner:
Well, I don't see any reason it wouldn't work 🤔

kiliman:
You can also simplify the replace as
```ts
let escapedFilename = filename.replace(/([.\[\]])/g, '\\$1')
```
`[[abc.123]]` => `\\[\\[abc\\.123\\]\\]`

fabiankaestner:
Ooh, that's interesting, the . loses its unescaped meaning inside a character set?

brophdawg11:
This is resolved by #7574 and will; be available in the next release

xdivby0:
[...]

PR #5809, opened 2023-03-15
fix(remix-dev): fix sourcemap path replace in Cloudflare Pages build

fabiankaestner:
Closes: #3768

- [ ] Docs
- [ ] Tests

#### Testing Strategy:

I built the remix project, and linked remix-dev:

```
yarn build
cd packages/remix-dev/
yarn link
```
Then I linked the remix-dev build in a cloudflare-workers project:

```
cd some-cloudflare-project/
yarn link @remix-run/dev
yarn dev
```
and verified that the sourcemaps work.

PR #7574, opened 2023-10-02
fix: sourcemap takes into account special chars in output file

jacob-ebey:
Closes: https://github.com/remix-run/remix/issues/3768

- [ ] Docs
- [ ] Tests

Testing Strategy:

Initialized a new CF Pages template and built Remix into it. Ran the project, connected a debugger, threw an error in a loader and sourcemaps brought me back to the correct line in the source `.ts` route.

### DIFF
--- a/packages/remix-dev/compiler/server/write.ts
+++ b/packages/remix-dev/compiler/server/write.ts
@@ -14,7 +14,7 @@ export async function write(
     if ([".js", ".cjs", ".mjs"].some((ext) => file.path.endsWith(ext))) {
       // fix sourceMappingURL to be relative to current path instead of /build
       let filename = file.path.substring(file.path.lastIndexOf(path.sep) + 1);
-      let escapedFilename = filename.replace(/\./g, "\\.");
+      let escapedFilename = filename.replace(/([.[\]])/g, "\\$1");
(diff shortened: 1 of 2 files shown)


packages/remix-dev/compiler/server/write.ts
```diff
@@ -14,7 +14,7 @@ export async function write(
     if ([".js", ".cjs", ".mjs"].some((ext) => file.path.endsWith(ext))) {
       // fix sourceMappingURL to be relative to current path instead of /build
       let filename = file.path.substring(file.path.lastIndexOf(path.sep) + 1);
-      let escapedFilename = filename.replace(/\./g, "\\.");
+      let escapedFilename = filename.replace(/([.[\]])/g, "\\$1");
```
brophdawg11, line 17: I like this regex better than the 3-pass replace in https://github.com/remix-run/remix/pull/5809, but do we need to also escape the `[` in the regex, as suggested in https://github.com/remix-run/remix/issues/3768#issuecomment-1186553615?

```suggestion
      let escapedFilename = filename.replace(/([.\[\]])/g, "\\$1");
```

jacob-ebey: No, it's inside the existing group, so it's not needed.

Issue #7056, opened 2023-08-03
Deprecate Remix ESLint Config


brophdawg11:
ESLint is an opinion, that's more stacks territory than Remix territory.

Chatted with Michael and we think we should

* First, remove this dependency from all of our templates
* Also remove it from the stacks and duplicate the simplest set of plugins and recommended "extends" setups in stacks (react, typescript, jest, etc)
* Wait a few months 🕐 🕐 
* Deprecate the `@remix-run/eslint` config in v2 with a warning on install

sergiodxa:
I think a plugin for Remix specific stuff would be great, with a default config, but the config is not related to Remix so it makes sense to deprecate it

MichaelDeBoey:
> First, remove this dependency from all of our templates

@brophdawg11 Would this mean we have no ESLint setup at all anymore in the templates?
Or would we also just set the recommended configs on the plugins (ESLint, import, react, react-hooks, jsx-a11y) that are currently used in the config?
If so, I guess we can sometimes extend those recommended configs with some extra information (like providing the `Form`, `Link` & `NavLink` components to the configs) as that's what we currently also do and I think it can help people out

brophdawg11:
I believe that's where we're leaning.  Templates are not meant to be opinionated - they're meant to be the simplest way to spin up a working Remix app.  ESLint is, by design, quite opinionated so we'd like it to become a stacks choice.  Even then we want the stacks to make very few choices and rely on best practices in the form of "recommended" presets.  Then app teams are free to modify as needed once they create their app from a stack.

brophdawg11:
The stacks have been converted to simpler ESLint configs and no longer depend on `@remix-run/eslint-config`:
* https://github.com/remix-run/indie-stack/pull/261
* https://github.com/remix-run/blues-stack/pull/215
* https://github.com/remix-run/grunge-stack/pull/172

Aqzhyi:
> Deprecate Remix ESLint Config #7056


```
// @ts-check
import config from 'eslint-config-fsc-gov'
import remixConfig from '@remix-run/eslint-config'
                        ^^^^ ❌ Could not find a declaration file for module
import remixConfigNode from '@remix-run/eslint-config/node'
                            ^^^^ ❌ Could not find a declaration file for module

/** @type {typeof config} } */
export default [...config]
```

```
Oops! Something went wrong! :(

ESLint: 8.53.0

Error: Failed to patch ESLint because the calling module was not recognized.
[...]

PR #7597, opened 2023-10-05
Remove @remix-run/eslint-config dependency

brophdawg11:
For https://github.com/remix-run/remix/issues/7056, we want to get out of the opinionated game of managing an ESLint config so we want to remove dependencies on `@remix-run/eslint-config` so it can be deprecated.  We simplified the stacks to use a config based on industry recommended configs (https://github.com/remix-run/indie-stack/pull/261, https://github.com/remix-run/blues-stack/pull/215, https://github.com/remix-run/grunge-stack/pull/172).

This PR does the same for all the templates, leveraging the same minimal config as we used in the stacks - mostly based on "recommended" presets.

I also added a helper script for us to use in the future.  It's tough to fully validate the templates lint properly in the monorepo with the hoisted deps.  I added a new `scripts/lint-templates.sh` script that can be run outside the repo and will use `create-remix` to create an app from each of the local templates in a `temp/` dir, install deps, and run `npm run lint`:

```sh
# In this case, I'm running from a playgrounds/ folder that is a sibling to the remix repo
../remix/scripts/lint-templates.sh [path-to-remix-repo]
```


brophdawg11:
I think we might need to sync up the `eslint` deps in the template and the ones in the root repo `package.json` to get linting working on the template folders during a root `eslint` run...will look into that next week

PR #7617, opened 2023-10-09
docs(components/await): add notice about blocking promises

comountainclimber:
Updates the await docs to make it clear that that the component must be wrapped in a <Suspense> parent, otherwise the rendering of the component using the <Await> will be blocked until promise resolution


brophdawg11:
I think the code example is enough here

PR #7628, opened 2023-10-10
Add pdf as assets to loaders

MajorTom327:
I needed to use a pdf as an asset. The documentation said [Any files inside the app folder can be imported into your modules](https://remix.run/docs/en/main/file-conventions/asset-imports)

But that not totally true. The loaders are set for most cases but pdf wasn't included in the list.

This add the pdf as a file loader to make them able to be copied and used as an asset.

Here was the error for handling the pdf:
image

Here is the project and the PR that rise the problem: 
[MajorTom327/landing#200](https://github.com/MajorTom327/landing/pull/200/files#diff-afdd4d168b8c9f1999a8729c3e01b087a822b1860bc87dc93ca826520fdfb969R1)


sergiodxa (approved):

MajorTom327:
Can this be merged for next minor version ?

MichaelDeBoey:
As mentioned in the ["Remix Vite is Now Stable" blogpost](https://remix.run/blog/remix-vite-stable#next-steps), we won't add any new compiler features/improvements that aren't targeting Vite

> [...] from now on all new features and improvements that require compiler integrations will only be targeting Vite

PR #7649, opened 2023-10-12
Fix: using a local *.tgz stack

real34:
This PR allows to create a Remix application from a `*.tgz` Remix Stack tarball.

My use case was to directly consume tarballs created by [`pnpm pack`](https://pnpm.io/cli/pack), but only `tar.gz` extension was supported in Remix.\
:information_source: In `pnpm`, the `tgz` extension [is also hardcoded](https://github.com/pnpm/pnpm/blob/d15606ad1c33e2bbc3efde9468542276a215722c/releasing/plugin-commands-publishing/src/pack.ts#L97) and I considered it'd be better to support more tarballs formats in Remix.

- [x] Docs
- [x] Tests

Testing Strategy:
- I duplicated existing test case and fixture for `tar.gz` tarball
- before that, I ran a `create-remix` command by renaming my local archive to `tar.gz` and it worked as expected


brophdawg11 (approved):

brophdawg11:
Thanks!

PR #7656, opened 2023-10-12
makes it possible to run remix on subfolder by changing config files

MariMax:
I can setup server to work with subfolder, but cannot do the same for client part, this will allow to pass basename for router as a property for <RemixBrowser />

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -593,3 +593,4 @@
 - TrySound
 - rogepi
 - fredericoo
+- marimax
--- a/packages/remix-react/browser.tsx
+++ b/packages/remix-react/browser.tsx
@@ -35,7 +35,9 @@ declare global {
 }
 /* eslint-enable prefer-let/prefer-let */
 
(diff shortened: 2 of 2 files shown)


brophdawg11:
👋 Thank you for the PR!  Unfortunately, we don't accept new features directly through PRs - per our [Open Development Process](https://remix.run/blog/open-development) we ask that the community opens Proposals for new feature requests so they can be considered and prioritized based on popularity.  There's actually already a [proposal for basename support](https://github.com/remix-run/remix/discussions/2891) in Remix so you can give that an upvote to show your support for it.

Until then, you could use [patch-package](https://www.npmjs.com/package/patch-package) to add in this small change if it's important for your app.

Most of the time in a Remix app, it's suitable to just include the basename on your route definitions (i.e, `routes/_index.tsx -> routes/mybasename._index.tsx`).  If there's a use case where this is insufficient, pease feel free to include it on the Proposal for consideration!

PR #7682, opened 2023-10-16
Unstable Vite: support for custom servers

pcattori:
Expanded `createRequestHandler` to accept a function for `build` that acts as a dynamic way to get the latest build during dev. Then added APIs for custom server to delegate asset handling and SSR to Vite in dev:

```ts
import { createViteServer, loadViteServerBuild } from "@remix-run/dev"

// util to setup a Vite dev server in middleware mode
let vite = createViteServer()
//  ^? import("vite").ViteDevServer

// serve assets via built-in Vite middleware
app.use(vite.middlewares)

// serve SSR via Remix's request handler
app.use('*', createRequestHandler(() => loadViteServerBuild(vite)))
```

Names prefixed with `unstable_` for now

## Example usage

```ts
// server.mjs
import {
  unstable_createViteServer,
  unstable_loadViteServerBuild,
} from "@remix-run/dev";
import morgan from "morgan";
import sourceMapSupport from "source-map-support";

sourceMapSupport.install();
installGlobals();

let vite =
  process.env.NODE_ENV === "production"
    ? undefined
    : await unstable_createViteServer();

const app = express();
app.disable("x-powered-by");

if (vite) {
  app.use(vite.middlewares);
} else {
  app.use(compression());
  app.use(
    "/build",
    express.static("public/build", { immutable: true, maxAge: "1y" })
  );
}
app.use(morgan("tiny"));

app.all(
  "*",
  createRequestHandler({
    build: vite
      ? () => unstable_loadViteServerBuild(vite)
      : await import("./build/index.js"),
  })
);

const port = process.env.PORT || 3000;
app.listen(port, async () => console.log(`Express server listening on port ${port}`));
```

## Testing Strategy

Integration tests

## TODO

- [ ] switch to using `.ts` for vite config

### DIFF
--- a/packages/remix-dev/vite/index.ts
+++ b/packages/remix-dev/vite/index.ts
@@ -1,10 +1,26 @@
 // This file allows us to dynamically require the plugin so non-Vite consumers
 // don't need to have Vite installed as a peer dependency. Only types should
 // be imported at the top level.
+import type { ViteDevServer } from "vite";
+
 import type { RemixVitePlugin } from "./plugin";
+import { id } from "./vmod";
 
 export const unstable_vitePlugin: RemixVitePlugin = (...args) => {
   // eslint-disable-next-line @typescript-eslint/consistent-type-imports
   let { remixVitePlugin } = require("./plugin") as typeof import("./plugin");
   return remixVitePlugin(...args);
+};
+
+export const unstable_createViteServer = async () => {
+  let vite = await import("vite");
+  return vite.createServer({
+    server: {
+      middlewareMode: true,
+    },
+  });
+};
+
+export const unstable_loadViteServerBuild = async (vite: ViteDevServer) => {
+  return vite.ssrLoadModule(id("server-entry"));
 };
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -859,8 +859,10 @@ function getRoute(
   pluginConfig: ResolvedRemixVitePluginConfig,
   file: string
 ): Route | undefined {
-  if (!file.startsWith(pluginConfig.appDirectory)) return;
-  let routePath = path.relative(pluginConfig.appDirectory, file);
+  if (!file.startsWith(viteNormalizePath(pluginConfig.appDirectory))) return;
+  let routePath = viteNormalizePath(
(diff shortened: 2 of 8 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -446,6 +446,13 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
 
         viteChildCompiler = await createViteDevServer({
           ...viteUserConfig,
+          server: {
+            ...viteUserConfig.server,
+            // when parent compiler runs in middleware mode to support
+            // custom servers, we don't want the child compiler also
+            // run in middleware mode as that will cause websocket port conflicts
+            middlewareMode: false,
+          },
```
pcattori, line 455: @markdalgleish maybe we just want to unset all `server` options for child compiler since we don't want to use it as a server at all?

packages/remix-server-runtime/server.ts
```diff
@@ -58,12 +56,45 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
...
+  let serverMode: ServerMode;
+  let staticHandler: StaticHandler;
+  let errorHandler: HandleErrorFunction;
 
   return async function requestHandler(
     request,
     loadContext = {},
     { criticalCss } = {}
   ) {
+    _build = typeof build === "function" ? await build() : build;
```
pcattori, line 83: purposefully placing the `await` within `requestHandler` so that `createRequestHandler` doesn't need to change to `async` which would be breaking change.

packages/remix-dev/vite/index.ts
```diff
@@ -1,10 +1,26 @@
...
+  let vite = await import("vite");
+  return vite.createServer({
+    server: {
+      middlewareMode: true,
+    },
+  });
+};
+
+export const unstable_loadViteServerBuild = async (vite: ViteDevServer) => {
+  return vite.ssrLoadModule(id("server-entry"));
```
pcattori, line 25: note: duplicated `"server-entry"` literal from `plugin.ts` since creating an abstraction on our virtual modules seemed unnecessary so far. might revisit this.

pcattori:
Windows CI is failing. Going to investigate/resolve this with a Windows VM next week.

PR #7705, opened 2023-10-18
Update form-vs-fetcher.md

SeanGroff:
I know modals are frowned upon by the Remix team, but the act of performing a destructive action should always warn the user before confirming the destructive action. Usually this involves showing a :gasp: confirmation modal.
I'm not keen on my verbiage as I tried to provide a generic answer. Feel free to modify. I just wanted to bring this to light.

Unrelated, thanks for adding this article to the docs. My team has found it so helpful in clearing up confusion they've had over the past few months that i've pinned it in our internal slack.

### DIFF
--- a/docs/discussion/form-vs-fetcher.md
+++ b/docs/discussion/form-vs-fetcher.md
@@ -43,7 +43,7 @@ These actions are generally more subtle and don't require a context switch for t
 
 - **Updating a Single Field**: Maybe a user wants to change the name of an item in a list or update a specific property of a record. This action is minor and doesn't necessitate a new page or URL.
 
-- **Deleting a Record from a List**: In a list view, if a user deletes an item, they likely expect to remain on the list view, while warning the user of the destructive action.
+- **Deleting a Record from a List**: In a list view, if a user deletes an item, they likely expect to remain on the list view, with that item no longer in the list.
 
 - **Creating a Record in a List View**: When adding a new item to a list, it often makes sense for the user to remain in that context, seeing their new item added to the list without a full page transition.
 
--- a/docs/discussion/form-vs-fetcher.md
+++ b/docs/discussion/form-vs-fetcher.md
@@ -43,7 +43,7 @@ These actions are generally more subtle and don't require a context switch for t
 
 - **Updating a Single Field**: Maybe a user wants to change the name of an item in a list or update a specific property of a record. This action is minor and doesn't necessitate a new page or URL.
 
-- **Deleting a Record from a List**: In a list view, if a user deletes an item, they likely expect to remain on the list view, with that item simply disappearing.
+- **Deleting a Record from a List**: In a list view, if a user deletes an item, they likely expect to remain on the list view, while warning the user of the destructive action.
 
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thank you for the update, but I don't know that we need to add this here in the docs.  This is more of a general UX best practice, but this doc isn't a guide on the best UX to achieve these actions - but instead a guide on whether a form or a fetcher makes more sense for the action from a Remix standpoint.  We even have a "Deleting a Record" a few lines above in the `Form` section so then we'd have to add it there too if this were a UX-driven-guide.

Maybe the _"with that item simply disappearing"_ wording is a bit too strong and implies the lack of confirmation, and could be softened to _"with that item no longer in the list"_ to describe the desired end state?

> I know modals are frowned upon by the Remix team

Where are you getting this sentiment from?  Modals are often overused, but also have a valid place in web apps IMO.  Even 2 bullets below is _"Loading Data for a Popover or Combobox"_ 🙃 

> My team has found it so helpful in clearing up confusion they've had over the past few months that i've pinned it in our internal slack.

This is so great to hear!

SeanGroff:
> Thank you for the update, but I don't know that we need to add this here in the docs.  This is more of a general UX best practice, but this doc isn't a guide on the best UX to achieve these actions - but instead a guide on whether a form or a fetcher makes more sense for the action from a Remix standpoint.  We even have a "Deleting a Record" a few lines above in the `Form` section so then we'd have to add it there too if this were a UX-driven-guide.
> 
> 
> 
> Maybe the _"with that item simply disappearing"_ wording is a bit too strong and implies the lack of confirmation, and could be softened to _"with that item no longer in the list"_ to describe the desired end state?
> 
> 
> 
> > I know modals are frowned upon by the Remix team
> 
> 
> 
> Where are you getting this sentiment from?  Modals are often overused, but also have a valid place in web apps IMO.  Even 2 bullets below is _"Loading Data for a Popover or Combobox"_ 🙃 
> 
> 
> 
> > My team has found it so helpful in clearing up confusion they've had over the past few months that i've pinned it in our internal slack.
> 
> 
> 
> This is so great to hear!

I agree. Softening the response is the right move I think. An item simply disappearing is a scary UX

docs/discussion/form-vs-fetcher.md
```diff
@@ -43,7 +43,7 @@ These actions are generally more subtle and don't require a context switch for t
 
 - **Updating a Single Field**: Maybe a user wants to change the name of an item in a list or update a specific property of a record. This action is minor and doesn't necessitate a new page or URL.
 
-- **Deleting a Record from a List**: In a list view, if a user deletes an item, they likely expect to remain on the list view, with that item simply disappearing.
+- **Deleting a Record from a List**: In a list view, if a user deletes an item, they likely expect to remain on the list view, while warning the user of the destructive action.
```
brophdawg11, line 46:
```suggestion
- **Deleting a Record from a List**: In a list view, if a user deletes an item, they likely expect to remain on the list view, with that item no longer in the list.
```

brophdawg11 (approved):

SeanGroff:
Thanks @brophdawg11 
This reads much better and stays on point with the goal of this guide 👏

PR #7707, opened 2023-10-19
Correcting Form Validation Error Handling

jmarbutt:
The current example did not account for handling undefined returned from `useActionData` in addition to passing the correct type into the generic. This is a simple update to documentation

Closes: #

- [X] Docs
- [ ] Tests


brophdawg11 (approved):
Thanks!

PR #7711, opened 2023-10-19
Update remix file conventions demo to v2 link

shamsup:
Small docs change to link to the v2 version of the interactive demo of remix's file conventions for routing rather than the v1 routing version in the File Conventions > remix.config.js page. The link is already updated on the Route File Naming page


brophdawg11 (approved):
Thanks!

PR #7725, opened 2023-10-21
Unstable Vite: dedupe and optimize react deps

pcattori:
Currently, we require [userland config](https://github.com/pcattori/remix-template-vite/blob/main/vite.config.ts#L6-L8) to do this. We should just do it for the user automatically when the Vite plugin for Remix is used. Then the userland vite config can be simplified by omitting the `optimizeDeps` and `resolve` entries.

See [vite-plugin-react](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react/src/index.ts#L267-L275) for a reference implementation.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -404,6 +404,21 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
         return {
           appType: "custom",
           experimental: { hmrPartialAccept: true },
+          optimizeDeps: {
+            include: [
+              // pre-bundle React dependencies to avoid React duplicates,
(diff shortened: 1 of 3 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -404,6 +404,12 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
         return {
           appType: "custom",
           experimental: { hmrPartialAccept: true },
+          optimizeDeps: {
+            include: ["react", "react-dom/client"],
```
pcattori, line 408: should probably look into adding JSX import runtimes to `include` like in `vite-plugin-react`

PR #7734, opened 2023-10-22
s/ "Dependencies installing..." / "Installing dependencies..."

threepointone:
This reads a little better imo.

(Not a docs change, but I didn't test it because it's just a wording change)


brophdawg11:
Thanks!  Do you mind rebasing this onto dev and pointing to the `dev` branch since it touches source code?  I tried a simple repoint but it picks up other docs commits already on `main` and not in `dev` so it needs the rebase :/

PR #7736, opened 2023-10-22
fix(remix-server-runtime): fix `Jsonify` for objects with `toJSON` function

### LINKED ISSUE #7821: useLoaderData loses typings when accessing some data

2.1.0

- [X] Yes

### Steps to Reproduce

We use GraphQL's GraphQLError type in some responses. I just updated one of our smaller apps to 2.1.0 and it's now struggling with the return types on these errors. Noticed that Remix now has it's own Jsonify as of 2.1.0 instead of type-fest's.

```typescript
  return json({errors: [] as GraphQLError[]})
};

  const data = useLoaderData<typeof loader>(); // data is JsonifyObject<{errors: GraphQLError[]; }>

  // But when I try to access `errors`, it loses the type
  const errors = data.errors; // errors is now Unknown[] instead of the expected GraphQLError[]
}
```

I traced this back to GraphQLError having a `toJSON()` method. If I remove that, the types are fine. I tried the current version of type-fest for comparison, and it deals with the toJSON method without blowing away the types.

[Typescript Playground for some reference](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgKwM4QHZwL5wGZQQhwBEAAlAKYjAAeAtFAK4YD0GEAJpSQNwBQoSLERwmqSgBkIAQ25QAIjJgyc+QsXJUaDZmyoyAxjD6Dw0eEgDiUGWAAWARUkAxaCGUxKnAKJRCUGoERKQA5rYOAI4ANqZCFqIAUugYwHgAnkEapDDpYJT0eJSoJgL83IbRMlRwlTKoqHAAsul+AXCUtF4YnI1t0Ij8cHAwEIkAygDyAHIAFACUAFxwNnZOru6e3v2BrKxwAMJEIJQY8DD2wI0QTPAyPSP2lCN5xXCoTJzcGNGZAO7QADW-Gw-H4nWE8EMmBKcGisnkcAAvHB6ukMIY4AtkQA+QbDKgwJhQLBoTCzBCUfzQVDLADaAF1UY0WjtGdh5iCyhCEtw8DImNF4HgWMZgJg4AAVYowbEIIZwPZwACqqGAGFCL3yhRlAHJGslMGl-kDUAAaVEPQGUShgRoXZ65fJwP5PLAASVRhkMxXtTw61KgjTAhHysHSCuhGFhqBk4GizxRlMDtLgjOZzVagfZAmGUdhTu84zjYATyPeJbL9TghtSGQAPIWIHgK-HKDjc4r9sM4AA9AD8kZh8CpAUaKMLnGLbYAdKOaZ2lcMBwqFUqAEK3R7PDCUP41lLGrWO+zKOAAJWodDEEka8Nv2+PcGbAbHLrdcE9MCgmVGXp9DRwMAMBDtG8CcMoqgouIUgIlSSgqI2rwvvCchUjiCyLt2y6Dnmw5wNodA7OOcAQSoc4plhPYrsMa77OMECfnAHD7qeABujqMR4GBMDI0S-Pg0A+o+haLKBsKEbQxEAOrAfYkqvOWZEyBRb7Vqy2YMgIoJAA)

### Expected Behavior

Objects returned from actions/loaders that have a `toJSON()` method preserve their types.

### Actual Behavior

Types are lost on objects returned from actions/loaders that have a `toJSON()` method.

MichaelDeBoey:
I also added some extra tests for objects

---

Closes #7821

### DIFF
--- a/packages/remix-server-runtime/jsonify.ts
+++ b/packages/remix-server-runtime/jsonify.ts
@@ -13,7 +13,7 @@ export type Jsonify<T> =
   IsAny<T> extends true ? any :
 
   // toJSON
-  T extends { toJSON(): infer U } ? (U extends JsonValue ? U : unknown) :
+  T extends { toJSON(): infer U } ? (U extends JsonValue ? U : Jsonify<U>) :
 
   // primitives
   T extends JsonPrimitive ? T :
@@ -49,7 +49,7 @@ export type Jsonify<T> =
 
 // value is always not JSON => true
 // value is always JSON => false
-// value is somtimes JSON, sometimes not JSON => boolean
+// value is sometimes JSON, sometimes not JSON => boolean
 // note: cannot be inlined as logic requires union distribution
 type ValueIsNotJson<T> = T extends NotJson ? true : false;
 
@@ -150,8 +150,15 @@ type _tests = [
   // toJson
   Expect<Equal<Jsonify<{ toJSON(): "stuff" }>, "stuff">>,
   Expect<Equal<Jsonify<Date>, string>>,
-  Expect<Equal<Jsonify<{ toJSON(): undefined }>, unknown>>,
-  Expect<Equal<Jsonify<{ toJSON(): Date }>, unknown>>,
+  Expect<Equal<Jsonify<{ toJSON(): any }>, any>>,
+  Expect<Equal<Jsonify<{ toJSON(): undefined }>, never>>,
+  Expect<Equal<Jsonify<{ toJSON(): Date }>, string>>,
+  Expect<Equal<Jsonify<{ toJSON(): Map<unknown, unknown> }>, EmptyObject>>,
+  Expect<Equal<Jsonify<{ toJSON(): Int8Array}>, Record<string, number>>>,
+  Expect<Equal<Jsonify<{ toJSON(): [1, "two", Date, undefined, false]}>, [1, "two", string, null, false]>>,
+  Expect<Equal<Jsonify<{ toJSON(): [unknown, 1]}>, [unknown, 1]>>,
+  // @ts-expect-error
+  Expect<Equal<Jsonify<{ toJSON(): MyClass}>, {a: string}>>,
   Expect<Equal<Jsonify<BooleanWithToJson>, string>>,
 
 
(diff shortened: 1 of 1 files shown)


packages/remix-server-runtime/jsonify.ts
```diff
@@ -161,13 +167,27 @@ type _tests = [
   // object
   Expect<Equal<Pretty<Jsonify<{}>>, {}>>,
   Expect<Equal<Pretty<Jsonify<{a: string}>>, {a: string}>>,
+  Expect<Equal<Pretty<Jsonify<{a: Date}>>, {a: string}>>,
+  Expect<Equal<Pretty<Jsonify<{a: Map<unknown, unknown>}>>, {a: EmptyObject}>>,
+  Expect<Equal<Pretty<Jsonify<{a: Int8Array}>>, {a: Record<string, number>}>>,
+  Expect<Equal<Pretty<Jsonify<{a: [1, "two", Date, undefined, false]}>>, {a: [1, "two", string, null, false]}>>,
+  Expect<Equal<Pretty<Jsonify<{a: [unknown, 1]}>>, {a: [unknown, 1]}>>,
+  // @ts-expect-error
+  Expect<Equal<Pretty<Jsonify<{a: MyClass}>>, {a: {a: string}}>>,
```
MichaelDeBoey, line 176: I can't seem to get these nested classes to work

@pcattori probably has a good idea where to look at

packages/remix-server-runtime/jsonify.ts
```diff
@@ -160,14 +167,31 @@ type _tests = [
 
   // object
   Expect<Equal<Pretty<Jsonify<{}>>, {}>>,
+  // @ts-expect-error
+  Expect<Equal<Pretty<Jsonify<{a: any}>>, {a: any}>>,
```
MichaelDeBoey, line 171: I also don't really get why this one is failing tbh 🤔
If I change this to be equal to `{a?: any}`, this passes though

packages/remix-server-runtime/jsonify.ts
```diff
@@ -146,8 +146,15 @@ type _tests = [
   // toJson
   Expect<Equal<Jsonify<{ toJSON(): "stuff" }>, "stuff">>,
   Expect<Equal<Jsonify<Date>, string>>,
-  Expect<Equal<Jsonify<{ toJSON(): undefined }>, unknown>>,
-  Expect<Equal<Jsonify<{ toJSON(): Date }>, unknown>>,
+  Expect<Equal<Jsonify<{ toJSON(): any }>, any>>,
+  Expect<Equal<Jsonify<{ toJSON(): undefined }>, never>>,
+  Expect<Equal<Jsonify<{ toJSON(): Date }>, string>>,
```
pcattori, line 151: Haven't checked the other test cases, but this one is incorrect.

```js
let result = JSON.parse(JSON.stringify({ toJSON: () => new Date() }))
// => {}
```

So the result is an empty object, not a `string`. `toJSON` doesn't seem to always recursively serialize the return value.

pcattori (commented):
@MichaelDeBoey could you split this up into smaller, more targeted PRs?

packages/remix-server-runtime/jsonify.ts
MichaelDeBoey: I was checking, which did respond with a string

```
let result = JSON.parse(JSON.stringify({ toJSON: () => Date() }))
// => 'Mon Oct 23 2023 03:45:02 GMT+0200 (Midden-Europese zomertijd)'
```

MichaelDeBoey:
> could you split this up into smaller, more targeted PRs?

Can you explain how you would see the separate PRs?
For me this PR is fixing the `toJSON` and testing it for all possible use-cases

+ It's also adding these use-cases in the object case (which already works, except for `MyClass`)

Issue #7593, opened 2023-10-05
Remix V1 to V2: Error console logs a weird issue but everything works fine


AlemTuzlak:
### What version of Remix are you using?

v2.0.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

My entry.server.ts:
```ts

export default async function handleRequest(
  request: Request,
  responseStatusCode: number,
  responseHeaders: Headers,
  remixContext: EntryContext
) { 
  const instance = createInstance(); 
  const lng = await i18n.getLocale(request); 
  let ns = i18n.getRouteNamespaces(remixContext);
 
  await instance
    .use(initReactI18next) 
    .init({
      ...i18nextOptions,  
      lng,
      ns,  
      resources 
    });
 
  instance.services.formatter?.add("lowercase", value => value.toLowerCase());

  const ABORT_DELAY = 10000; 
  let markup = <RemixServer abortDelay={ABORT_DELAY} context={remixContext} url={request.url} />;

  return new Promise((resolve, reject) => {
    let shellRendered = false;

    const { pipe, abort } = renderToPipeableStream(markup, {
      onAllReady() {
        let body = new PassThrough();
        const stream = createReadableStreamFromReadable(body);
        responseHeaders.set("Content-Type", "text/html");

        resolve(
          new Response(stream, {
            status: responseStatusCode,
            headers: responseHeaders
          })
        );
        pipe(body);
      },
      onShellError(err: unknown) {
        reject(err);
      },
      onError(error: unknown) {
        responseStatusCode = 500;
        // Log streaming rendering errors from inside the shell.  Don't log
        // errors encountered during initial shell rendering since they'll
        // reject and get logged in handleDocumentRequest.
        if (shellRendered) {
          console.error(error);
        }
      }
    });
    setTimeout(abort, ABORT_DELAY);
  });
}
```

My remix.config.js:

```js
const { createRoutesFromFolders } = require("@remix-run/v1-route-convention");
const ignoredRoutePatterns = ["**/.*", "**/components/**", "**/integration/*.test.*"];

/** @type {import('@remix-run/dev').AppConfig} */
module.exports = { 
  ignoredRouteFiles: ignoredRoutePatterns,
  serverDependenciesToBundle: [/^marked.*/],
  tailwind: true,
  serverModuleFormat: "cjs", 
  watchPaths: ["./public/locales/en/common.json", "./public/locales/fr/common.json"],
[...]

PR #7745, opened 2023-10-23
Fix console warning for routes that do not export a default component

brophdawg11:
Closes https://github.com/remix-run/remix/issues/7593

### DIFF
--- a/packages/remix-react/server.tsx
+++ b/packages/remix-react/server.tsx
@@ -31,6 +31,7 @@ export function RemixServer({
   }
 
   let { manifest, routeModules, criticalCss, serverHandoffString } = context;
+  debugger;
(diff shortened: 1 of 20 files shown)


packages/remix-react/server.tsx
```diff
@@ -31,6 +31,7 @@ export function RemixServer({
   }
 
   let { manifest, routeModules, criticalCss, serverHandoffString } = context;
+  debugger;
```
theskillwithin, line 34: @brophdawg11 was this left in by accident?

brophdawg11: Yes it was!  Thanks!

theskillwithin (approved):

PR #7765, opened 2023-10-25
docs(hooks/use-fetcher): add notice about index routes in `load` section

rachaelmcq:
Adds additional doc to clarify how to use `load` on index routes. It was a bummer to find [this excellent explanation](https://github.com/remix-run/remix/discussions/3494#discussioncomment-2966946) of our problem buried in a discussion rather than in the docs themselves.

### DIFF
--- a/docs/hooks/use-fetcher.md
+++ b/docs/hooks/use-fetcher.md
@@ -112,7 +112,7 @@ fetcher.submit(
 
 Loads data from a route loader. While multiple nested routes can match a URL, only the leaf route will be called. 
 
-Note that when calling `load` on an index route's loader, you [must include](https://remix.run/docs/en/1.19.3/guides/routing#what-is-the-index-query-param) a `?index` query parameter in order to disambiguate between the `index.tsx` layout and the `root.tsx` route.
+Note that when calling `load` on an index route's loader, you must include an [`?index` query param](../guides/index-query-param) in order to disambiguate between the `index.tsx` layout and the `root.tsx` route.
 
 ```ts
 fetcher.load("/some/route");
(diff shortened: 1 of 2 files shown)


docs/hooks/use-fetcher.md
```diff
@@ -61,7 +61,9 @@ fetcher.submit(formData);
 
 ## `fetcher.load(href)`
 
-Loads data from a route loader. While multiple nested routes can match a URL, only the leaf route will be called.
+Loads data from a route loader. While multiple nested routes can match a URL, only the leaf route will be called. 
+
+Note that when calling `load` on an index route's loader, you [must include](https://remix.run/docs/en/1.19.3/guides/routing#what-is-the-index-query-param) a `?index` query parameter in order to disambiguate between the `index.tsx` layout and the `root.tsx` route.
```
akamfoad, line 66: > in order to disambiguate between the `index.tsx` layout and the `root.tsx` route.

I think the `root.tsx` is misleading here, `?index` is for every index route where there is need to disambiguate between a route and and _index route in its hierarchy, in RR and Remix `root` is usually used to refer to the root route, the first route that contains all other routes

```
- root.tsx <-- this is root
- routes/
  - posts.tsx <-- root layout for all routes beneath posts
  - posts._index.tsx <-- the index route beneath posts route
```

Also `_index.tsx` in this context isn't a layout per se, it is a leaf node among the route

MichaelDeBoey, line 66:
```suggestion
Note that when calling `load` on an index route's loader, you must include an [`?index` query param](../guides/index-query-param) in order to disambiguate between the `index.tsx` layout and the `root.tsx` route.
```

MichaelDeBoey (changes requested):

PR #7778, opened 2023-10-26
docs(utils/sessions): clarify `session.id`

gustavopch:
Just clarifying for other people who like me may be confused about how `session.id` works.


brophdawg11:
Thanks!

brophdawg11 (approved):

Issue #7467, opened 2023-09-18
Failed to create esm bundled application


KnisterPeter:
### What version of Remix are you using?

2.0.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

- Create a new app with `pnpx create-remix@latest test`.
- In `remix.config.js` add `serverDependenciesToBundle: [/.*/]`
- Run `pnpm run dev`.

### Expected Behavior

The application should start.

### Actual Behavior

The application actually fails while trying to `require` 'http' from node.
It looks like the package `@remix-run/node` does not contain any valid esm export. I assume it does fallback to commonjs instead and from there it takes `@remix-run/web-fetch` in commonjs as well.

Not sure if this is the correct assumption.

```
markusw@xxx:/tmp/test (main)$ pnpm run dev

> test@ dev /tmp/test
> remix dev --manual


 💿  remix dev

 info  building...
 info  built (797ms)
Error: Dynamic require of "http" is not supported
    at file:///tmp/test/build/index.js?t=1695049617985.849:11:9
    at node_modules/.pnpm/@remix-run+web-fetch@4.4.0/node_modules/@remix-run/web-fetch/dist/lib.node.cjs (file:///tmp/test/build/index.js?t=1695049617985.849:4452:16)
    at file:///tmp/test/build/index.js?t=1695049617985.849:17:50
    at node_modules/.pnpm/@remix-run+node@2.0.0_typescript@5.2.2/node_modules/@remix-run/node/dist/globals.js (file:///tmp/test/build/index.js?t=1695049617985.849:5476:20)
    at file:///tmp/test/build/index.js?t=1695049617985.849:17:50
    at node_modules/.pnpm/@remix-run+node@2.0.0_typescript@5.2.2/node_modules/@remix-run/node/dist/index.js (file:///tmp/test/build/index.js?t=1695049617985.849:9751:20)
    at file:///tmp/test/build/index.js?t=1695049617985.849:17:50
    at file:///tmp/test/build/index.js?t=1695049617985.849:20675:27
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)
^C ELIFECYCLE  Command failed.
```

KnisterPeter:
If the assumption above is correct (ESM exports missing) I will try to create a PR which addresses the issue.
Other than that I have no clue why we see commonjs for this in the server bundle if the output is set (and defaults) to ESM.

mao437303994:
[...]

Issue #7533, opened 2023-09-25
import_node3.Headers is not a constructor


dani-mp:
### What version of Remix are you using?

2.0.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

I just upgraded Remix to v2 and I'm getting this error at runtime:
```
ERROR    Invoke Error     {
    "errorType": "TypeError",
    "errorMessage": "import_node3.Headers is not a constructor",
    "stack": [
        "TypeError: import_node3.Headers is not a constructor",
        "    at convertApigRequestToNode (/var/task/build/server.cjs:1112109:19)",
        "    at /var/task/build/server.cjs:1112156:21",
        ...
    ]
}
```
I'm using `@remix-run/serve` and `serverModuleFormat: "cjs"`.

I deployed my Remix site to AWS using SST.

### Expected Behavior

The server doesn't crash.

### Actual Behavior

The server crashes.

brophdawg11:
Could you provide a minimal reproduction?

dani-mp:
To fix this, I had to add `nodejs: { format: 'cjs' }` to my `RemixSite` construct in sst. This wasn't needed before.

My `remix.config.js` also included `serverModuleFormat: "cjs"`, given that the default now is `esm`.

PR #7779, opened 2023-10-26
Add esbuild `banner` for ESM+NodeJS builds, to work around "dynamic require not supported" issue

brettdh:
## Bug report and failing test

Having `type: "module"` in `package.json` along with `serverDependenciesToBundle: "all"` in `remix.config.js` causes a build error:

    Error: Dynamic require of "http" is not supported

Oddly, in this integration test, it appears to happen at run time (unless I'm misreading the stack trace), whereas in the description of #7467, the error appears at build time. However, the error did appear at run time in my earlier testing as well, as indicated in the AWS Lambda logs in [my comment](https://github.com/remix-run/remix/issues/7467#issuecomment-1775411138).

## Bug fix

The general `Dynamic require of <pkg> not supported` issue is discussed at length in https://github.com/evanw/esbuild/issues/1921, and it appears that several other projects have had to apply a workaround for this problem. I first attempted to apply [this async IIFE](https://github.com/evanw/esbuild/issues/1921#issuecomment-1439609735) approach, but I realized that Remix already has a similar workaround applied for the `__file` and `__dirname` globals, so I removed those bits.

I'm not confident that this is the "right" fix (or a complete fix), but it makes the new integration test pass, and it doesn't appear to cause any other failures. It may also become irrelevant in the future if the esbuild issue is fixed, or if Remix fully [moves to Vite](https://github.com/remix-run/remix/discussions/7632), but I'm hoping it will at least unblock my current progress.

Closes: #7467 

## Tasks

- [x] Tests
  - [x] Add bugreport test that repros the issue
  - [ ] Move bugreport test to integration test
  - [x] Test this fix using patch-package with a real (proprietary) project where I'm also encountering the issue
- [ ] Docs
  - Not sure what would be appropriate to add here; happy to add something if someone can point me in the right direction
- [ ] Changeset
  - Waiting to see if folks think this looks like the right fix.

## Testing Strategy

Ran all the integration tests several times. Saw some flaky failures, and one not-as-flaky failure:

```
  1) [chromium] › defer-test.ts:823:3 › non-aborted › routes are interactive when deferred promises are suspended and after rejection in subsequent payload

    Test timeout of 30000ms exceeded.

    Error: page.waitForSelector: Target closed
    =========================== logs ===========================
    waiting for locator('#MANUAL_FALLBACK_ID') to be visible
fix: apply workaround for 'dynamic require not supported' error
    ============================================================

      830 |     await page.waitForSelector(`#${ROOT_ID}`);
      831 |     await page.waitForSelector(`#${DEFERRED_ID}`);
    > 832 |     await page.waitForSelector(`#${MANUAL_FALLBACK_ID}`);
          |                ^
      833 |     let idElement = await page.waitForSelector(`#${RESOLVED_DEFERRED_ID}`);
      834 |     let id = await idElement.innerText();
      835 |     expect(id).toBeTruthy();

        at /Users/brhiggins/src/remix/integration/defer-test.ts:832:16

    Pending operations:
      - page.waitForSelector at integration/defer-test.ts:832:16
```

Confirmed that this test also fails (pretty reliably) without my changes.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -65,6 +65,7 @@
 - bogas04
 - BogdanDevBst
 - bolchowka
+- brettdh
 - brookslybrand
 - brophdawg11
 - bruno-oliveira
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -6,6 +6,7 @@ import {
   createAppFixture,
   createFixture,
   js,
+  json,
 } from "./helpers/create-fixture.js";
 
 let fixture: Fixture;
@@ -58,6 +59,38 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
+      "package.json": json({
+        name: "bug-report-test",
+        private: true,
+        sideEffects: false,
+        type: "module",
+        dependencies: {
+          "@remix-run/css-bundle": "0.0.0-local-version",
+          "@remix-run/node": "0.0.0-local-version",
+          "@remix-run/react": "0.0.0-local-version",
+          "@remix-run/serve": "0.0.0-local-version",
+          isbot: "0.0.0-local-version",
+          react: "0.0.0-local-version",
+          "react-dom": "0.0.0-local-version",
+        },
+        devDependencies: {
+          "@remix-run/dev": "0.0.0-local-version",
+          "@types/react": "0.0.0-local-version",
+          "@types/react-dom": "0.0.0-local-version",
+          typescript: "0.0.0-local-version",
+
+          "@vanilla-extract/css": "0.0.0-local-version",
+          tailwindcss: "0.0.0-local-version",
+        },
+        engines: {
+          node: ">=18.0.0",
+        },
+      }),
+      "remix.config.js": js`
+        export default {
+          serverDependenciesToBundle: "all",
+        };
+      `,
       "app/routes/_index.tsx": js`
         import { json } from "@remix-run/node";
         import { useLoaderData, Link } from "@remix-run/react";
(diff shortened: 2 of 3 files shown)


brettdh:
Updated PR with a patch that works around the issue; also updated the description with testing details and feedback requests; PTAL

brettdh:
Update on the testing in my proprietary app running in AWS Lambda: the original issue is fixed, but now I'm getting [this error](https://github.com/remix-run/remix/issues/7533):

```
2023-10-27T16:39:57.889Z	1372b4b0-9cb0-4652-8f45-4443b3e20a55	ERROR	Invoke Error {
  "errorType": "TypeError",
  "errorMessage": "import_node3.Headers is not a constructor",
  "stack":[
    "TypeError: import_node3.Headers is not a constructor",
    "    at createRemixHeaders2 (file:///var/task/ui/build/index.mjs:34785:17)",
    "    at createRemixRequest2 (file:///var/task/ui/build/index.mjs:34780:14)",
    "    at Runtime.handler (file:///var/task/ui/build/index.mjs:34816:49)",
    "    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1173:29)"
  ]
}
```

brettdh:
After further digging, that last error appears to be an issue with remix-aws, so I'll open an issue over there.

brettdh:
The remix-aws issue is resolved and my project is working now 🥳

brettdh:
@brophdawg11 can this PR please be reopened before it completely falls off the radar? I'd be happy to hide the last few comments once it's reopened, in order to avoid future confusion and refocus the discussion on the originally-reported issue.

brettdh:
For whatever reason, I'm now getting this error when testing my app, despite not having made changes to this patch since I (thought I) had tested it successfully:

```
SyntaxError: Identifier 'require' has already been declared
```

Will investigate and report back.

brettdh:
I haven't been following closely, but I'm wondering about:
1. the expected pace of the Vite transition
2. whether it's likely to become stable "soon"
3. whether that means this PR will be quickly obsoleted
4. whether that's why it's not getting much attention 🙂

THOUSAND-SKY:
As far as I can tell, there isn't a way to pass this rule to `esbuild` from `remix.config.js` without this patch, is there?

PR #7790, opened 2023-10-27
Bug: cannot navigate to route using <Link> only with <a>

marcioluis:
- [ ] Docs
- [x] Tests

Testing Strategy:

I cannot navigate to route through <Link> if route has a dependency on a *.server.ts file.
If navigating using <a> element works fine.

`npx yarn bug-report-test`

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -58,28 +58,32 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
+      "app/utils/buggy.server.ts": js`
+        export function buggy() {
+          return "this is a bug";
+        }
+      `,
+      "app/routes/about.tsx": js`
+        import { buggy } from "../utils/buggy.server";
+        const forcebug = buggy();
+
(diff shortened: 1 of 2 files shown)


integration/bug-report-test.ts
```diff
@@ -58,30 +58,34 @@ test.beforeAll(async () => {
...
     ////////////////////////////////////////////////////////////////////////////
     files: {
+      "app/utils/buggy.server.ts": js`
+        export function buggy() {
+          return "this is a bug";
+        }
+      `,
+      "app/routes/about.tsx": js`
+        import { buggy } from "../utils/buggy.server";
+        const forcebug = buggy();
```
brophdawg11, line 68: This is your issue - components need to render both on the server (initial document request) and on the client (initial page hydration and subsequent client-side navigations).  So you cannot user server-only modules in components if you intend to hydrate your app into a SPA and do client-side navigations via `<Link>`.  Server-only modules can only be used in code that _only_ runs on the server - such as loaders and actions.

If you _want_ to use server only-modules in your components, you can do so if you treat your app as a MPA and do not include the `<Scripts>` components - this preventing Remix from hydrating your app into a SPA.

PR #7817, opened 2023-10-30
Update contributors.yml

tmsagarofficial:
Added my name to the contributors.yml


brophdawg11:
This is only for when you actually contribute to Remix though... We're always open to bug fix PRs though!

PR #7842, opened 2023-10-31
fix(vite): fix Fast Refresh load error when using defer

marbemac:
## To replicate the problem:

1. Fire up an remix + vite app (I tested on 2.2.0-pre.3)
2. Add a route with a loader that defers, e.g.:

```tsx
const getData = async () => {
  await new Promise(resolve => setTimeout(resolve, 1000));
  return {
    name: 'marc',
  };
};

  const user = getData();

  return defer({ user });
}

export default function Articles() {
  const { user } = useLoaderData<typeof loader>();

  return (
    <Suspense fallback={<div>loading user...</div>}>
      <Await resolve={user}>{user => <div>hi {user.name}</div>}</Await>
    </Suspense>
  );
}
```

3. Load the route in your browser, see error in console

Screenshot 2023-10-31 at 10 24 42 AM

## Why?

While the recommendation to put `<LiveReload />` before `<Scripts />` mitigates the problem in the non-deferred stream use-case, deferring the stream means that non-async modules on the page might be parsed after async modules (since async modules can be run before page parsing is even complete).

This PR updates the refresh runtime fragment to also be marked as async, which resolves the issue from what I can tell in my local testing.

Probably not the final solution here, as I know ya'll are trying to find a way to avoid having to re-order LiveReload/Scripts in general. But this should at least unblock anybody with an app that uses deferred streaming who is looking to try the vite plugin in the meantime.


markdalgleish:
Thanks for the PR! I just updated your branch, merging the latest from `dev` and adding a test that fails without your fix.

markdalgleish (approved):

markdalgleish (changes requested):
Test is failing in Firefox, seeing if I can fix it.

markdalgleish:
I could only reproduce it locally when running the test in Firefox via Playwright. I found a workaround that summed to do the trick though, so CI is passing now.

markdalgleish (approved):

vaishnavi1029:
Error: Remix Vite plugin can't detect preamble. Something is wrong.
 ❯ app/routes/stored._index.tsx:2:8
      1| export default function Comp(){
      2| return <h1>jcs</h1>
      }
      /// <reference types="vitest" />
/// <reference types="vite/client" />
import { vitePlugin as remix } from "@remix-run/dev";
import tsconfigPaths from "vite-tsconfig-paths";
import { JSDOMOptions } from "vitest";
installGlobals();

export default defineConfig({
  test:{
    environment:"jsdom",
    globals: true,
    setupFiles: ["./set-up-test.env.ts"],
  },
  plugins: [
    remix({
      ignoredRouteFiles: ["**/*.css"],
    }),
    tsconfigPaths(),
   
  ],
  
});
please help me solve this asap

Issue #7837, opened 2023-10-31
unstable_vitePlugin ignores multiple 'Set-Cookie' headers


nickprinsloo:
### What version of Remix are you using?

0.0.0-nightly-99a8d9c-20231031

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Please see https://github.com/nickprinsloo/remix-vite

1. Clone repository
2. Install dependencies
3. Run `pnpm dev-vite`
4. Click on "Try to set two cookies"

### Expected Behavior

I'd expect both cookies to be in the Cookies header (and displayed on the page)

<img width="850" alt="Screenshot 2023-10-31 at 11 56 41" src="https://github.com/remix-run/remix/assets/5994921/b6231cae-f736-4b03-9a6d-267484f133c9">

### Actual Behavior

Only the last cookie set is displayed.

Worth mentioning this is different behaviour compared to Remix without Vite and only affects the dev server.

PR #7843, opened 2023-10-31
fix: handle multiple "Set-Cookie" headers in unstable_vitePlugin dev

nickprinsloo:
Closes: #7837

- [ ] Docs
- [x] Tests

Testing Strategy:

Added additional test to the existing integration test for the Vite Node adapter.


markdalgleish:
Thanks for the PR! I noticed the same issue has already fixed in solid-start's implementation in the time since we used their adapter as our initial starting point. I've updated your PR to use theirs instead to keep the two implementations closer. I've also made the ordering more explicit in the test.

markdalgleish (approved):

nickprinsloo:
Great, makes sense. Thanks for taking a look.

PR #7879, opened 2023-11-02
docs(vite): add env variables configuration

wobsoriano:
This PR adds environment variable configuration docs when using the Remix vite plugin.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -601,3 +601,4 @@
 - SeanGroff
 - jmarbutt
 - KarelVerschraegen
+- wobsoriano
--- a/docs/future/vite.md
+++ b/docs/future/vite.md
@@ -293,6 +293,34 @@ export default defineConfig({
 });
 ```
 
+#### Configure environment variables
+
+Vite provides its [own mechanism](https://vitejs.dev/guide/env-and-mode.html#env-files) for exposing environment variables through [`import.meta.env`](https://vitejs.dev/guide/env-and-mode.html#env-files). You can either convert your existing environment variables to add a `VITE_` prefix and use the `import.meta.env` API, or you can use the [`vite-plugin-environment`](https://github.com/ElMassimo/vite-plugin-environment) plugin to keep using `process.env`:
+
+👉 **Install `vite-plugin-environment`**
+
+```shellscript nonumber
+npm install -D vite-plugin-environment
+```
+
+👉 **Add `vite-plugin-environment` to your Vite config**
+
+```ts filename=vite.config.ts
+import { unstable_vitePlugin as remix } from "@remix-run/dev";
+import { defineConfig } from "vite";
+import environmentPlugin from "vite-plugin-environment";
+
+export default defineConfig({
+  plugins: [
+    remix(),
+    environmentPlugin([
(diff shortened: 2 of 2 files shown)


markdalgleish:
Thanks for the PR!

The existing Remix compiler doesn't provide environment variables to the client, see https://remix.run/docs/en/main/guides/envvars. This means that anyone migrating an existing app won't assume that this is supported. Anyone who wants this behaviour in Vite is free to do so on their own, but our docs shouldn't need to mention it.

Since Remix apps have a server, rather than baking environment variables into the built assets, we instead recommend that environment variables are passed from the server to the client via the root loader at runtime. This way the environment variables on the client can change when the server environment changes.

KiwiKilian:
@markdalgleish thanks for your explanations!

> The existing Remix compiler doesn't provide environment variables to the client, see https://remix.run/docs/en/main/guides/envvars. This means that anyone migrating an existing app won't assume that this is supported. 

But what about the other way around, providing env vars to the server? As the Environment Variables docs state:

> If you're using the remix dev server to run your project locally, it has built-in support for [dotenv](https://www.npmjs.com/package/dotenv).

Would be great to know how to add server only env variables.

bbonamin:
Yep, I'm hitting this wall too when migrating a project to Vite, and I even need access to `process.env` vars in the initial `server.mjs`. What's the recommended approach to keep the same behavior as it currently exists today without Vite (that is, process.env vars from `.env` files are automagically available in the server).

stephen776:
> Yep, I'm hitting this wall too when migrating a project to Vite, and I even need access to `process.env` vars in the initial `server.mjs`. What's the recommended approach to keep the same behavior as it currently exists today without Vite (that is, process.env vars from `.env` files are automagically available in the server).

Same here.

bbonamin:
@stephen776 I found this workaround, but as the comment says, I'm not 100% sure of whether this exposes env vars to the client bundle... DON'T USE IN PROD just in case!
```
import { unstable_vitePlugin as remix } from "@remix-run/dev";
import { defineConfig, loadEnv } from "vite";
import tsconfigPaths from "vite-tsconfig-paths";

export default ({ mode }: { mode: string }) => {
  // Experimental. I'm not sure if this will expose env vars to the client!
  const env = loadEnv(mode, process.cwd(), "");
  process.env = { ...process.env, ...env };
  // Experimental - END.

  return defineConfig({
    plugins: [
      remix({
        ignoredRouteFiles: ["**/.*", "**/*.css", "**/*.test.{js,jsx,ts,tsx}"],
      }),
      tsconfigPaths(),
    ],
  });
};
```

stephen776:
> ```ts
> // Experimental. I'm not sure if this will expose env vars to the client!
>   const env = loadEnv(mode, process.cwd(), "");
>   process.env = { ...process.env, ...env };
>   // Experimental - END.
> ```

Does this still require prefixing all the var with `VITE_` ? If so, I think this will expose everything to the client

markdalgleish:
Sorry, I missed that `process.env` isn't being populated from `.env` files in dev on the _server_. There's an issue tracking this and I should have a fix up soon: https://github.com/remix-run/remix/issues/7934

Issue #7868, opened 2023-11-01
Vite with Large Apps: Very slow initial request and slow subsequent requests


dmarkow:
### What version of Remix are you using?

2.2.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

Adding per [a Discord discussion with @pcattori](https://discord.com/channels/770287896669978684/1169278500637773904/1169344627552288820). Given a large app with many routes and components, the first page request after starting `vite dev` can result in hundreds of network requests. In my case most pages make around 250-300 requests, but another person in that same discussion had close to 600. In addition, every following request to any route (via navigation or actions/loaders) has a significant delay (2+ seconds per page for me). Including both of these issues together since they seem closely related in the way the plugin works.

### Slow Initial Request

In my case, the first page request stalls for around 12 seconds while vite does its thing. After this, the app appears to load, but is not responsive for another 12 seconds while it runs 250 network requests for individual components and dependencies. This seems to be a common issue with Vite in dev mode due to the lack of bundling. But there may be some optimization opportunities on the remix plugin side.

I first ran `vite dev` and loaded a page, which let vite do it's initial `new dependencies optimized...` thing. I then restarted vite and ran `vite dev --profile`, made a browser request, then loaded the profile result into [speedscope.app](https://speedscope.app). Looking at it in Left Heavy mode, the time spent in `resolveConfig` and then `flatRoutes` is close to 9 seconds total. This is because `resolveConfig` is being called every time the plugin is loaded, which happens for each individual request. So even though it runs relatively quickly (around 30ms in my case), it's running hundreds of times and returning the same result every time. My initial thought was to cache the results of `resolveConfig`, but adding/renaming routes will break.

The next largest bottleneck in the remix plugin is at [the transform function for remix-remix-react-proxy](https://github.com/remix-run/remix/blob/main/packages/remix-dev/vite/plugin.ts#L649). For my initial page load, it's spending around 3-4 seconds here, most likely due to parsing the code of every single file in my project to see if it has "@remix-run/react" in it.

![CleanShot 2023-11-01 at 16 49 30@2x](https://github.com/remix-run/remix/assets/8009/91556add-c1a4-4de4-959d-f0a4db285117)

### Slow Subsequent Requests

[...]

PR #7883, opened 2023-11-02
perf(vite): fix LiveReload proxy bailout logic

markdalgleish:
This fixes an issue raised by @dmarkow in https://github.com/remix-run/remix/issues/7868#issuecomment-1791470740.

In the Vite dev server, we use a Babel transform to replace usage of `@remix-run/react` with a proxy module so that we can replace the `LiveReload` component with a Vite-specific implementation. Since this transformation doesn't come for free, we first check whether the file contains references to `@remix-run/react` and `LiveReload`. This should mean that, even if the transformation could be faster, it's only applied to a single file in your app.

However, there was an error in the boolean logic for this check that meant we were processing any file containing `@remix-run/react`, even if it was missing a reference to `LiveReload`. This PR fixes the logic so that both strings need to be present, as intended.

### DIFF
--- a/integration/vite-css-dev-test.ts
+++ b/integration/vite-css-dev-test.ts
@@ -36,6 +36,9 @@ test.describe("Vite CSS dev", () => {
               port: ${devPort},
               strictPort: true,
             },
+            optimizeDeps: {
+              include: ["react", "react-dom/client", "@remix-run/react"],
+            },
             plugins: [remix()],
           });
         `,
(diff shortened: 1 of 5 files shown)


pcattori (approved):

packages/remix-dev/vite/plugin.ts
```diff
@@ -666,11 +666,11 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
           return;
         }
 
+        let referencesLiveReload =
```
pcattori, line 669: Was trying to think of a better name, but names like `mightReferenceLiveReload` don't really read well when negated: `!mightReferenceLiveReload`, so I think its fine. Maybe `coarseFilterForLiveReload` or something might explain better what this is doing conceptually.

markdalgleish: Chatted offline, going with `hasLiveReloadHints` for now.

integration/vite-css-dev-test.ts
```diff
@@ -60,28 +63,28 @@ test.describe("Vite CSS dev", () => {
...
           .index {
             background: peachpuff;
             padding: ${TEST_PADDING_VALUE};
           }
         `,
-        "app/routes/_index/route.tsx": js`
-          import "./styles-bundled.css";
-          import linkedStyles from "./styles-linked.css?url";
-          import cssModulesStyles from "./styles.module.css";
+        "app/routes/_index.tsx": js`
```
markdalgleish, line 84: This is fixing an unrelated issue with the test while I'm here. The template already contains `app/routes/_index.tsx`, so we were adding a conflicting route with `app/routes/_index/route.tsx`.

integration/vite-css-dev-test.ts
```diff
@@ -36,6 +36,9 @@ test.describe("Vite CSS dev", () => {
               port: ${devPort},
               strictPort: true,
             },
+            optimizeDeps: {
+              include: ["react", "react-dom/client", "@remix-run/react"],
```
markdalgleish, line 40: We need this array defined so that we don't have multiple copies of these dependencies due to our project structure.

pcattori (approved):

PR #7889, opened 2023-11-03
Vite Plugin migration path, with some findings

Plopix:
Just a little more doc on the migration path.

The ENV vars issue was a struggle (especially if you don't know Vite)
You may also want to look here: https://github.com/withastro/astro/tree/main/packages/astro/src/vite-plugin-env

Also, but that is probably a bug, I had to `npx remix reveal` otherwise I had this error:

```....node_modules...dev/dist/config/defaults/entry.server.node.tsx" outside of root directory```

Most likely because I have PNPM workspaces and the `node_module` is indeed outside. => but that should work

I hope it helps

### DIFF
--- a/docs/future/vite.md
+++ b/docs/future/vite.md
@@ -90,23 +90,59 @@ Remix is now just a Vite plugin, so you'll need to hook it up to Vite.
 
 ```ts filename=vite.config.ts
 import { unstable_vitePlugin as remix } from "@remix-run/dev";
-import { defineConfig } from "vite";
+import tsconfigPaths from "vite-tsconfig-paths";
+import { defineConfig, loadEnv } from "vite";
 
-export default defineConfig({
-  plugins: [remix()],
-});
+export default ({ mode }: { mode: string }) => {
+    const env = loadEnv(mode, process.cwd(), "");
+    process.env = { ...process.env, ...env };
+    return defineConfig({
+        plugins: [
+            remix(),
+            tsconfigPaths(),
+        ],
+    });
+};
 ```
 
 The subset of [supported Remix config options][supported-remix-config-options] should be passed directly to the plugin:
 
 ```ts filename=vite.config.ts lines=[3-5]
-export default defineConfig({
-  plugins: [
-    remix({
-      ignoredRouteFiles: ["**/.*"],
-    }),
-  ],
-});
+export default ({ mode }: { mode: string }) => {
+    const env = loadEnv(mode, process.cwd(), "");
+    process.env = { ...process.env, ...env };
+    return defineConfig({
+        plugins: [
+            remix({
+              ignoredRouteFiles: ["**/.*"],
+            }),
+            tsconfigPaths(),
+        ],
+    });
+};
+```
+
+#### CSS Import
+
+If you were importing file like this
+
+```ts
+import stylesheet from "~/styles/app.css";
+
+export const links: LinksFunction = () => [
+    ...(cssBundleHref ? [{ rel: "stylesheet", href: cssBundleHref }] : []),
+    { rel: "stylesheet", href: stylesheet },
+];
+```
+
+now you can just do this:
+
+```ts
+import "~/styles/app.css";
+
+export const links: LinksFunction = () => [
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: docs/future/vite.md [hunk 1]
```diff
@@ -90,23 +90,59 @@ Remix is now just a Vite plugin, so you'll need to hook it up to Vite.
...
+```ts
+import stylesheet from "~/styles/app.css";
+
+export const links: LinksFunction = () => [
+    ...(cssBundleHref ? [{ rel: "stylesheet", href: cssBundleHref }] : []),
+    { rel: "stylesheet", href: stylesheet },
+];
+```
+
+now you can just do this:
```
kentcdodds, line 138: I strongly advise against recommending this. https://www.epicweb.dev/tips/do-not-bundle-all-your-css

Plopix: I can remove that I was recommended to add it.... so you think that's better to add the `?inline` ?

Plopix:
```// CSS
declare module '*.css' {
  /**
   * @deprecated Use `import style from './style.css?inline'` instead.
   */
  const css: string
  export default css
}```

Comment (kentcdodds) [reply docs/future/vite.md:138]: I just mean it's better to actually import the stylesheet and include it itself in most cases.

Comment (Plopix) [reply docs/future/vite.md:138]: I understand, it is just that without doing anything, developer will have to understand the change to be made.
Either do the thing you don't recommend or add the `?inline` which by itself seemed weird to me too.
So we have to make a decision, or mention both in the migration path I guess.
Let me know!

Comment (kentcdodds) [reply docs/future/vite.md:138]: I don't understand why any change needs to be made though. With vite can you not import the href anymore? That sounds wrong (and bad). It would probably be best to get an actual Remix maintainer involved to make a decision here...

Comment (Plopix) [reply docs/future/vite.md:138]: of course that's just a sharing feedback here. I am no one to take a decision but yet, indeed it displays a big warning if you don't do anything:

Screenshot 2023-11-08 at 12 26 27 PM

Screenshot 2023-11-08 at 12 27 13 PM

So all I am saying is (and the only goal of that PR): helping developer in the migration paths :)

Which ever is the decision, I will update the PR

Thank @kentcdodds to spot that, it's indeed very important to point people in the good direction!

Comment (markdalgleish) [reply docs/future/vite.md:138]: In theory Vite can support importing the href of a CSS file, but Vite currently has an issue where CSS files imported with the `?url` query string don't go through the CSS transformation pipeline. This means tools like PostCSS/Tailwind won't apply to your CSS. We have an issue tracking it from a Remix perspective here: https://github.com/remix-run/remix/issues/7786. 

In addition to this issue, it's also worth considering that we're trying to pick our battles a bit more when it comes to bundling so that we're better positioned to leverage tools like Vite. As a result, we're targeting CSS side-effect imports as the primary way of managing CSS since this lines up with Vite: https://vitejs.dev/guide/features.html#css.

File: docs/future/vite.md [hunk 2]
```diff
@@ -90,23 +90,59 @@ Remix is now just a Vite plugin, so you'll need to hook it up to Vite.
...
+            remix({
+              ignoredRouteFiles: ["**/.*"],
+            }),
+            tsconfigPaths(),
+        ],
+    });
+};
+```
+
+#### CSS Import
```
markdalgleish, line 125: This is already covered in the "Fix up CSS imports" section.

docs/future/vite.md
```diff
@@ -90,23 +90,59 @@ Remix is now just a Vite plugin, so you'll need to hook it up to Vite.
...
-import { defineConfig } from "vite";
-
-export default defineConfig({
-  plugins: [remix()],
-});
+import tsconfigPaths from "vite-tsconfig-paths";
+import { defineConfig, loadEnv } from "vite";
+
+export default ({ mode }: { mode: string }) => {
+    const env = loadEnv(mode, process.cwd(), "");
```
markdalgleish, line 97: This has been fixed in https://github.com/remix-run/remix/pull/7958 and will be out in the next release.

markdalgleish:
Thanks for the feedback and the PR!

The issues you've raised have all been addressed so this update shouldn't be necessary anymore.

PR #7908, opened 2023-11-04
Add caching around resolvePluginConfig() and only invalidate virtual modules when necessary

dmarkow:
Closes: #7868. This adds caching around `resolvePluginConfig()` to avoid recalculating routes hundreds of times on startup. There is one cache used for `remix-react-refresh-babel` and `remix-remove-server-exports`, and another used in the middleware to determine if the virtual modules need to be invalidated.

I was unable to get a single cache working, as the cache would get updated in one of the other plugin transforms so there was no way to tell if it was stale for the route request by the time the middleware ran.

- [ ] Docs
- [x] Tests: All existing Vite integration tests are passing with these changes

Testing Strategy:

Discussion with @pcattori, "we have pretty good tests for this sort of thing in integration/vite-dev-express-test so if those pass with your changes I'd be reasonably confident about it"

Also used this in developing our app the last couple of days, which included changing  #and adding new routes and lots of HMR/HDR. I didn't run into a single stale cache issue with HMR/HDR, and performance was a ton better than 2.2.0 as released.

### Benchmarking

On 2.2.0, my app startup took 24 seconds, HMR took 2+ seconds when there was a loader, and basic site navigation took around 2 seconds.

With the patch, app startup is just under 12 seconds and HMR with loaders and site nav are both around 400ms. (Note that some, but not all, of this improvement is due to the LiveReload logic fix added since 2.2.0).

Not profiled below, but my `vite build && vite build --ssr` times dropped from a combined 129 seconds to 24 seconds.

2.2.0:

CleanShot 2023-11-04 at 09 13 44@2x
CleanShot 2023-11-04 at 09 14 12@2x
CleanShot 2023-11-04 at 09 14 35@2x

Patched:

CleanShot 2023-11-04 at 09 14 53@2x
CleanShot 2023-11-04 at 09 15 09@2x
CleanShot 2023-11-04 at 09 15 21@2x

### DIFF
--- a/.changeset/gentle-ghosts-wait.md
+++ b/.changeset/gentle-ghosts-wait.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Cache resolution of Remix Vite plugin options
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -543,11 +543,12 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
         vite.httpServer?.on("listening", () => {
           setTimeout(showUnstableWarning, 50);
         });
+
         // We cache the pluginConfig here to make sure we're only invalidating virtual modules when necessary.
         // This requires a separate cache from `cachedPluginConfig`, which is updated by remix-hmr-updates. If
         // we shared the cache, it would already be refreshed by remix-hmr-updates at this point, and we'd
         // have no way of comparing against the cache to know if the virtual modules need to be invalidated.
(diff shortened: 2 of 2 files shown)


markdalgleish (approved):
Thanks for the PR!

markdalgleish:
Also, thanks for the help, and for being so thorough! I really appreciate it.

ZipBrandon:
@markdalgleish I'm using `0.0.0-nightly-903268f-20231106` and confirm that I'm back to 20s build times from 5m45 seconds. Thank you for getting this in and for @dmarkow authoring!

PR #7919, opened 2023-11-06
fix(vite): fix HMR preamble race condition

### LINKED ISSUE #7863: vite and defer seem to lead to `Uncaught Error: @vitejs/plugin-react can't detect preamble. Something is wrong.` and `TypeError: window.__remixClearCriticalCss is not a function`

markdalgleish:
Closes #7863.

This PR fixes a couple issues:

- Resolves a race condition between the React Refresh preamble script and the app code by ensuring that the preamble is imported first in the same script tag as the app code.
- Allows the `LiveReload` component to behave the same as the existing Remix compiler from a consumer perspective. You can now choose _not_ to render it, forcing you to perform a full refresh to see changes, and you can also render it after `Scripts`. I've updated the ordering of `LiveReload` and `Scripts` in the Vite tests to show this is the case.

### DIFF
--- a/.changeset/afraid-suns-yawn.md
+++ b/.changeset/afraid-suns-yawn.md
@@ -1,0 +1,6 @@
+---
+"@remix-run/dev": patch
+"@remix-run/react": patch
+---
+
+Support rendering of `LiveReload` component after `Scripts` in Vite dev
--- a/.changeset/early-trees-walk.md
+++ b/.changeset/early-trees-walk.md
@@ -1,0 +1,6 @@
+---
+"@remix-run/dev": patch
+"@remix-run/react": patch
+---
+
+Support optional rendering of `LiveReload` component in Vite dev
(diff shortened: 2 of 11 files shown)


MichaelDeBoey (approved):

pcattori (approved):
LGTM

Longterm, I'd like an API that decouples dev-only scripts from prod scripts. For example, I may want to use HMR in development but not ship any JS assets in production.

pcattori:
Also, I could have sworn that using `<LiveReload />` without `<Scripts />` worked for HMR in esbuild-based compiler, but I guess not. So this would get us parity (and then some) with the current Remix compiler.

PR #7932, opened 2023-11-07
Remove redundant `--watch` argument

sambostock:
This removes `--watch` from the `dev` script in the Express template.

The docs for `--watch-path` show it used without `--watch`, and explicitly say that it "starts Node.js in watch mode and specifies what paths to watch." Therefore, since we're already explicitly passing `server.js` as a `--watch-path`, we don't need to specify `--watch`, as Node.js will already be put into watch mode.

See https://nodejs.org/api/cli.html#--watch-path

### DIFF
--- a/templates/express/package.json
+++ b/templates/express/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "build": "remix build",
-    "dev": "remix dev --manual -c \"node --watch-path server.js --watch server.js\"",
+    "dev": "remix dev --manual -c \"node --watch-path server.js server.js\"",
     "start": "cross-env NODE_ENV=production node ./server.js",
     "typecheck": "tsc"
   },
--- a/contributors.yml
+++ b/contributors.yml
@@ -481,6 +481,7 @@
 - ryanjames1729
 - ryankshaw
 - sarahse
+- sambostock
 - sandulat
 - sandstone991
 - sbernheim4


templates/express/package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "build": "remix build",
-    "dev": "remix dev --manual -c \"node --watch-path server.js --watch server.js\"",
+    "dev": "remix dev --manual -c \"node --watch-path server.js server.js\"",
```
brophdawg11, line 7: Since we only need to watch a single file, should we use [`--watch`](https://nodejs.org/docs/latest-v18.x/api/cli.html#--watch)?   `--watch-path` seems intended to watch directories.

sambostock: From the docs:

> [By default, watch mode will **watch the entry point and any required or imported module**. Use `--watch-path` to specify what paths to watch.](https://nodejs.org/api/cli.html#--watch:~:text=By%20default%2C%20watch%20mode%20will%20watch%20the%20entry%20point%20and%20any%20required%20or%20imported%20module.%20Use%20%2D%2Dwatch%2Dpath%20to%20specify%20what%20paths%20to%20watch.)

My understanding/assumption is that we _don't_ want to watch the required/imported modules, because we use `chokidar` to watch the app files ourselves in `server.js`, and re-`createRequestHandler`, **without needlessly completely restarting the Express server**.

Therefore, we want to override the default and explicitly just watch `server.js`.

---

That said, in my team's app we've updated this to `--watch-path`:

- `server.js`
- `boot` (a directory from which we `import` a couple things right at the top of `server.js`, such as `boot/instrumentation.js` for Open Telemetry)
- `yarn.lock` (we should restart if we've updated deps)
- `.shadowenv.d` ([Shadowenv](https://github.com/Shopify/shadowenv) provides environment variables in development, so we should restart if they've changed)

brophdawg11: Ah, ok I see this in the docs now, makes sense

> This will turn off watching of required or imported modules

I'll defer to @pcattori or @markdalgleish for a final say

sambostock:
As a related aside, I noticed that Node versions prior to 18.17.0 seem to swallow log messages when in `watch` mode. Not sure if that can be captured somewhere... 🤷

brookslybrand:
Now that [we're using `vite`](https://github.com/remix-run/remix/blob/main/templates/express/package.json#L7) this change is no longer necessary

Thanks @sambostock for opening this PR, sorry it sat here for so long

PR #7933, opened 2023-11-07
chore: enforce sorted `contributors.yml`

sambostock:
This adds a test that enforces `contributors.yml` is sorted, and makes the test pass by sorting the file.


MichaelDeBoey (commented):
I don't think this is necessary as we have a bot that auto-sorts this file, but I'll let the team decide if we want to enforce people doing this themselves on every PR

sambostock:
Fair. One advantage of sorting the file itself is that it reduces the likelihood of merge conflicts from everyone appending to the bottom of the time. It also makes it obvious where you're supposed to insert your handle.

brophdawg11:
Yeah this should be happening automatically - we don't want to add any friction between someone and a potential contribution to Remix.  I'm unsure why it hasn't run recently though so we'll look into that...

PR #7937, opened 2023-11-08
fix(vite): fix dev FOUC with custom Express server

### LINKED ISSUE #7878: FOUC when using custom servers

2.2.0

- [X] Yes

### Steps to Reproduce

1. `npx create-remix@latest --template remix-run/remix/templates/unstable-vite-express`
2. Create and import a css file that does something noticeable like set background color to red or something
3. Load app, note flash on load, because CSS is not included in the server rendering and is only added on the client

### Expected Behavior

No flash of unstyled content when using custom servers.

### Actual Behavior

There is a flash of unstyled content when using custom servers.

---

This might be tricky to solve - the long standing approach to SSR w vite (the approach described in Vite's docs) is IMHO pretty cumbersome, would be amazing if remix could somehow smooth over the custom server + vite experience.

But that's a bit off topic, regarding this specific problem - one option is to somehow expose `getStylesForUrl` to the end user so that they can pass in the critical styles to their request handler similar to how the remix vite plugin does on request. The problem is that `getStylesForUrl` requires `cssModulesManifest`, which is scoped to the `remixVitePlugin` plugin.

The larger issue is that longer term, with the current custom server approach, any logic/features that remix's vite plugin covers in the `configureServer` middleware will need to be replicated by the end user in their custom server.

I'd be willing to try and tackle a solution here if ya'll have any preferred direction or other ideas!

markdalgleish:
Fixes #7878.

### DIFF
--- a/.changeset/slow-eyes-joke.md
+++ b/.changeset/slow-eyes-joke.md
@@ -1,0 +1,6 @@
+---
+"@remix-run/dev": patch
+"@remix-run/express": patch
+---
+
+Fix flash of unstyled content on initial page load in Vite dev when using a custom Express server
--- a/integration/vite-css-dev-express-test.ts
+++ b/integration/vite-css-dev-express-test.ts
@@ -1,0 +1,251 @@
+import { test, expect } from "@playwright/test";
+import fs from "node:fs/promises";
+import path from "node:path";
+import getPort from "get-port";
+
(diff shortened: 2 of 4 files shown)


gijsroge:
@markdalgleish just tested the pre-version, works flawlessly.

PR #7943, opened 2023-11-08
docs(tutorial): add note about multiple actions per page

lauhon:
As discussed [here](https://github.com/remix-run/remix/discussions/3138) the tutorial does not show the actual "recommended " way of handling multiple actions per page. I thought this would be a nice note

### DIFF
--- a/docs/start/tutorial.md
+++ b/docs/start/tutorial.md
@@ -994,6 +994,10 @@ When the user clicks the submit button:
 
 Add a `Form`, add an `action`, Remix does the rest.
 
+Note that we created the delete action in a separate file simply to explore that option. In reality we recommend you to stay
+in your page file and distinguish between the different actions, for example with a switch statement.
+You can learn more about that in [this awesome video](https://youtu.be/w2i-9cYxSdc?si=DDa78AYojCM6VCt7&t=117)
+
 ## Index Routes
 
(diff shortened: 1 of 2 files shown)


docs/start/tutorial.md
```diff
@@ -994,6 +994,10 @@ When the user clicks the submit button:
 
 Add a `Form`, add an `action`, Remix does the rest.
 
+Note that we created the delete action in a separate file simply to explore that option. In reality we recommend you to stay
+in your page file and distinguish between the different actions, for example with a switch statement.
+You can learn more about that in [this awesome video](https://youtu.be/w2i-9cYxSdc?si=DDa78AYojCM6VCt7&t=117)
+
```
brophdawg11, line 1000: > In reality we recommend you to stay in your page file and distinguish between the different actions

This feels pretty subjective and not necessarily an official recommendation.

Even in the [linked discussion](https://github.com/remix-run/remix/discussions/3138#discussioncomment-2715459), it's stated as such:

> But how you structure your code is a personal preference 

So I don't know that we want/need to make a statement like this in the tutorial?

lauhon: My POV is that I just started touching remix and that was a question I had after going through the tutorial. Wanted to save future learners from having to look for that info in discussions and yt videos. 

But if it's optional, as you say it might be best to just close this pr, fine with me ^^

PR #7948, opened 2023-11-08
Patch 1

FDiskas:
If I have nested ENV variables it is not populated
```env
FIRST_NAME="Joe"
SURNAME="Doe"
FULL_NAME="${FIRST_NAME} ${SURNAME}"
```
It should combine
```sh
source .env && echo $FULL_NAME   
Joe Doe
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -603,3 +603,4 @@
 - jmarbutt
 - KarelVerschraegen
 - mrkstwrt
+- fdiskas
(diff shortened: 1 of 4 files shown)


brophdawg11:
I'll defer to @pcattori here but I think our original intention with `dotenv` integration is to provide simple/minimal support out of the box, but if you needed more complex logic you would do that in your custom server/scripts (similar to our stance on remix-server versus a custom express server).  The additional `dotenv-expand` package could be argued as something that belongs in userland since it's nor part of the built-in/minimal dotenv support.

FDiskas:
I fully understand and partially agree. The idea is that: by default env variables works like that by default, but the framework parses env variables differently. As an end user I couldn't understand why it is not working 😔

PR #7958, opened 2023-11-10
fix(vite): load .env files into process.env in dev

### LINKED ISSUE #7934: Vite: `process.env` variables are not working inside loader

2.2.0

- [X] Yes

### Steps to Reproduce

Define custom environment variable on .env file:

```text
TEST="Custom value"
REMIX_TEST="Custom value"
VITE_TEST="Custom value"
```

On root.tsx add loader

```javascript
export const loader: LoaderFunction = async (args) => {
    console.log(process.env.TEST) // outputs undefined
    console.log(process.env.REMIX_TEST) // outputs undefined
    console.log(import.meta.env.VITE_TEST) // outputs Custom value

     return {
         test: process.env.TEST, // outputs undefined
         remix_test: process.env.REMIX_TEST, // outputs undefined
         vite_test: import.meta.env.VITE_TEST, // outputs Custom value
     }
}
```

### Expected Behavior

Server side code should be able to read data from `process.env`. I believe using `import.meta.env.VITE_TEST` on SSR code reveals that variable to client side code as expected by Vite documentation.

Maybe I'm missing something? Or it's a bug?

### Actual Behavior

Trying to get any `process.env` variables inside loaders return always `undefined` instead.

markdalgleish:
Fixes #7934

### DIFF
--- a/.changeset/tricky-news-thank.md
+++ b/.changeset/tricky-news-thank.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Populate `process.env` from `.env` files on the server in Vite dev
--- a/integration/vite-build-test.ts
+++ b/integration/vite-build-test.ts
@@ -23,6 +23,9 @@ test.describe("Vite build", () => {
           throw new Error("Remix should not access remix.config.js when using Vite");
           export default {};
         `,
+        ".env": `
+          ENV_VAR_FROM_DOTENV_FILE=true
+        `,
         "vite.config.ts": js`
           import { defineConfig } from "vite";
           import { unstable_vitePlugin as remix } from "@remix-run/dev";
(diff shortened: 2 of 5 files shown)


JacobMGEvans:
Hey! I love this, we came across this https://discord.com/channels/856971667393609759/1171477346046447616

seems like the backend code is not reading the Secret key properly in some situations.
https://github.com/JacobMGEvans/remix-and-vite-repro
Repro routes: 
http://localhost/:5173/working
http://localhost/:5173/not-working

markdalgleish:
@JacobMGEvans You should be using `process.env` on the server rather than `import.meta.env`. In Remix I'd recommend avoiding `import.meta.env` entirely since it bakes the environment variables into the code rather than allowing them to be dynamic at runtime. Instead you should use `process.env` on the server, and if the client needs access to it, pass it to the client via the loader.

john-griffin:
@markdalgleish what do you suggest we do for test environment? The env vars only seem to load in dev and prod.

PR #7959, opened 2023-11-10
docs(streaming): Update broken links in streaming.md

thomasverleye:
The `entry.server.ts` url is broken and there are a few more snippets that can be linked.


brophdawg11 (approved):
Thanks!

PR #7968, opened 2023-11-10
docs(v2): Fix devServerPort example

billc-dev:


docs/start/v2.md
```diff
@@ -719,10 +719,8 @@ module.exports = {
 ```js filename=remix.config.js good lines=[3-7]
 module.exports = {
-  future: {
-    v2_dev: {
-      port: 8002,
-    },
+  dev: {
+    port: 8002,
```
brophdawg11, line 723: This specific example is actually correct, but can be a bit confusing since this guide is ultimately targeting 2 audiences:
1. Folks on a 1.x release preparing for their eventual v2 upgrade.  This code is correct for those users (and that's the intention here)
2. Folks on a 2.x release.  This code is incorrect for them - they are targeted by a separate [section higher up](https://remix.run/docs/en/main/start/v2#after-upgrading-from-v1-to-v2)

Even (2) contains subgroups based on whether those folks did the future flags in v1 or if they're just going straight to v2 without the future flags and adopting the breaking changes.

We're definitely open to suggestions/PR on clearing up the confusion though if you have ideas!  Maybe for v3 we should consider 2 separate guides to separate "preparing for v3 while still on 2.x" and "upgrading to v3.x"

brophdawg11: Maybe instead of this change we could add a small link to this section to the hash heading up above?

billc-dev: I was confused because the [`devServerPort` section](https://remix.run/docs/en/main/start/v2#devserverport) started with

> In your `remix.config.js`, rename `devServerPort` to `dev.port`.

Which is different from the code example

Also `browserNodeBuiltinsPolyfill` is only typed in v2

One suggestion might be to add (v1) to the filename but adding a space between might not work. Not sure how you guys render your docs because I think that's not opensource right now. 
image

Another solution would be to add comments similar to what is done in `formMethod`'s [code section](https://remix.run/docs/en/main/start/v2#formmethod)

```js filename=remix.config.js good lines=[3-7]
module.exports = {
  // v1
  future: {
    v2_dev: {
      port: 8002,
    },
  },
  // v2
  dev: {
    port: 8002,
  },
};
```

brophdawg11:
I pushed a small clarification for this in 8ba9041df6b5e68298d566f2aa99f3dc2fcee97c

PR #7979, opened 2023-11-12
docs(guide): Adding a new guide for creating custom fileUploadHandler 

volvoDon:
Description
This PR introduces an example showcasing custom file upload functionality in Remix.js, focusing on AWS S3 integration. Recognizing the lack of detailed examples in this area, I've developed a guide that can be generalized for various upload scenarios. I'm working towards a guide that can fill in current gaps and I'm very open to feedback.

Key Aspects
Detailed AWS S3 file upload integration example.
Adaptable approach for different file upload needs.

I am open to suggestions on integrating this new content seamlessly. Feedback on clarity, usability, and documentation fit is highly welcomed.

Additional Notes
Aims to fill a gap in current file upload documentation.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -604,3 +604,4 @@
 - KarelVerschraegen
 - mrkstwrt
 - AltanS
+- volvoDon
--- a/docs/guides/file-upload-custom.md
+++ b/docs/guides/file-upload-custom.md
@@ -1,0 +1,178 @@
+---
+title: File Uploads Custom
+toc: false
+---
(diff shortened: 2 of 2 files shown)


brophdawg11:
This feels like it might be better suited for the [Examples](https://github.com/remix-run/examples) repository.  It actually looks like there is an S3 integration example in there: https://github.com/remix-run/examples/tree/main/file-and-s3-upload.  If they're showing the same things, then I think we don't need two - but if there's a material difference between the two such that both would make sense, feel free to open a PR over there with this example!

PR #7981, opened 2023-11-12
Fix route indentation in migration to v2 - v2.md

fayez-nazzal:
While following the migration guide, I found that switching the route filesystem structure was misleading:
- The `root.tsx` shouldn't change in path. The guide has it nested inside `app/routes` while it should remain in `app/`.
- It's not straightforward to compare and follow the changes. I changed the indentation and added `app/` on the top to make it easier to follow.


brophdawg11 (approved):
Thanks!

PR #8007, opened 2023-11-14
Update crossOrigin value used in links documentation

bmac:
Valid values are "anonymous" | "use-credentials". https://github.com/remix-run/remix/blob/43387828b3079d7f1f839b6bafa02e911f2be421/packages/remix-server-runtime/links.ts#L16


brophdawg11 (approved):
Thanks!

PR #8018, opened 2023-11-15
docs(guides/envvars): update Cloudflare description

jabranr:
This is true for both Cloudflare Pages and Workers-site with Remix v2.

### DIFF
--- a/docs/guides/envvars.md
+++ b/docs/guides/envvars.md
@@ -46,7 +46,7 @@ export async function loader() {
 }
 ```
 
-If you're using the `@remix-run/cloudflare-pages` adapter, env variables work a little differently. Since Cloudflare Pages are powered by Functions, you'll need to define your local environment variables in the [`.dev.vars`][dev-vars] file. It has the same syntax as `.env` example file mentioned above.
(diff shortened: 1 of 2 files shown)


docs/guides/envvars.md
```diff
@@ -46,7 +46,7 @@ export async function loader() {
 }
 ```
 
-If you're using the `@remix-run/cloudflare-pages` adapter, env variables work a little differently. Since Cloudflare Pages are powered by Functions, you'll need to define your local environment variables in the [`.dev.vars`][dev-vars] file. It has the same syntax as `.env` example file mentioned above.
+If you're using the `@remix-run/cloudflare` adapter, env variables work a little differently. Since Cloudflare Pages are powered by Functions, you'll need to define your local environment variables in the [`.dev.vars`][dev-vars] file. It has the same syntax as `.env` example file mentioned above.
```
brophdawg11, line 49: > Since Cloudflare Pages are powered by Functions

Do we need to alter the wording here too if it's applicable to both pages and workers?

jabranr: Agreed. It now applies to both Pages and Worker sites equally although there is a difference still behind the scenes i.e.
- CF Pages are still powered by Functions
- CF Workers with module syntax will only pass environment variables to the context
- CF Workers with service worker syntax (legacy) enable the environment variables globally

The whole paragraph may need a bit of redo.

brophdawg11:
Lets get the conflicts resolved

PR #8053, opened 2023-11-18
docs: update msw example to msw@2.x version + add ESM related details

mccuna:
I've updated the msw example based on the breaking changes they introduced in v2. Also, considering that Kent's epic-stack uses ESM by default I've added a few details regarding ESM modules to the msw example (might be useful for people using epic-stack as a reference).


brophdawg11 (approved):
Thanks!

PR #8100, opened 2023-11-22
Fixes #7951: prevent remix.init errors when init script is an ES6 module, update docs to clarify init script requirements

### LINKED ISSUE #7951: remix.init script fails on latest create-remix

2.2.0, 2.3.0

- [X] Yes

### Steps to Reproduce

- Create new Remix project (`npx create-remix@latest`)
- Add `remix.init` folder
- Add `remix.init/index.js` file:

```js
console.log("Greetings from remix.init :)");

process.exit();
```

- Use the new template to create a new Remix app: `npx create-remix@latest --template <template-url>`

### Expected Behavior

Remix init script runs successfully.

### Actual Behavior

```bash
  init   This template has a remix.init script. Do you want to run it?
         Yes

      ✔  Dependencies installed

      ✔  Git initialized

      ◼  Running template's remix.init script... 


      ▲  Oh no! Template's remix.init script failed
```

Running `npx remix init` indicates an ESM error:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/andrelandgraf/workspaces/tests/using-template/remix.init/index.js from /Users/andrelandgraf/workspaces/tests/using-template/node_modules/@remix-run/dev/dist/cli/commands.js not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/andrelandgraf/workspaces/tests/using-template/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).

    at Object.init (/Users/andrelandgraf/workspaces/tests/using-template/node_modules/@remix-run/dev/dist/cli/commands.js:97:16)
    at async Object.run (/Users/andrelandgraf/workspaces/tests/using-template/node_modules/@remix-run/dev/dist/cli/run.js:167:7) {
  code: 'ERR_REQUIRE_ESM'
}
```

leothorp:
Closes: #7951 

This addresses the two issues below, which are underlying https://github.com/remix-run/remix/issues/7951.

1. Due to the use of `require` to import the init script, remix.init/index.js currently cannot be defined as an ESM module without throwing `Error [ERR_REQUIRE_ESM]: require() of ES Module`. This file will be considered an ESM module by default if using `npx create-remix@latest`, creating a situation where defining a commonjs package.json in remix.init was required for it to not throw-  despite [the docs](https://remix.run/docs/en/main/guides/templates#customize-initialization) indicating that that the package.json should be optional.
 * fixed by switching to a dynamic import for ESM (which still falls back to a `require` for CJS modules) and adding a test covering ESM remix.init scripts.

2. [The docs](https://remix.run/docs/en/main/guides/templates#customize-initialization) don't indicate any required exports for the file, but it only works if exporting an init function. Updated docs and error messages to make the required contract more explicit, and added a test for the case of an init script without an export.

- [x] Docs
- [x] Tests

Testing Strategy:
- added new unit tests for both cases above; see comments below for the related code.
- Verified existing tests still pass.
- manual testing of all scenarios before/after the changes (es6 module, cjs module, default-exported function, no export).


packages/remix-dev/cli/commands.ts
```diff
@@ -44,12 +44,15 @@ export async function init(
...
 
-  let initFn = require(initScript);
+  let initFn = await import(initScript);
+
   if (typeof initFn !== "function" && initFn.default) {
     initFn = initFn.default;
   }
   try {
-    await initFn({ packageManager, rootDirectory: projectDir });
+    if (typeof initFn === "function") {
```
leothorp, line 53: Covered by the "succeeds for a remix.init script that doesn't default export a function" test.

packages/remix-dev/__tests__/init-test.ts
```diff
@@ -31,6 +32,31 @@ jest.mock("child_process", () => {
   };
 });
 
+// Copy over the current build of @remix-run/dev and execute remix init from the node_modules dir there.
+// Certain bugs only manifest when using this approach.
+const execRemixInitInProject = (projectDir: string) => {
```
leothorp, line 37: This stuff with copying over the newly-built version of @remix-run/dev and running it as a node module in the temp project is clunky, but was the only way I could reproduce the failing behavior (es6 module init script erroring) in a unit test. 

It seems like some aspect of the `run(["init"])` calls used by other tests here doesn't fully replicate the real-world behavior of running `remix init` as a project dependency, so this was the only workaround I found initially

markdalgleish:
Thanks for the PR!

Just a heads up that `create-remix` has its own code for running the init script rather than using the Remix CLI. You'll need to update it there too if you want to fix the `create-remix` flow — but given that the logic is a bit more more complicated now, maybe you could rework it to call out to `remix init` instead?

leothorp:
@markdalgleish I see, thanks for the info! 

Sounds good, I'll look at that second approach you suggested.

leothorp:
> Thanks for the PR!
> 
> Just a heads up that `create-remix` has its own code for running the init script rather than using the Remix CLI. You'll need to update it there too if you want to fix the `create-remix` flow — but given that the logic is a bit more more complicated now, maybe you could rework it to call out to `remix init` instead?

@markdalgleish Just updated the create-remix flow to call into the same function. Additionally, after seeing the more explicit error messages in the create-remix version, it seemed like it made more sense to just keep the exported function as a requirement (more explicit, less surface area for module side-effect related edge cases, etc.) Adapted `remix init` and the docs over to that version.  

Whenever it's convenient, mind taking another look to see if this looks like the right general idea? Thanks again.

packages/create-remix/index.ts
```diff
@@ -542,57 +541,20 @@ async function runInitScriptStep(ctx: Context) {
...
-  } catch (err) {
-    error("Oh no!", "Template's remix.init script failed");
-    throw err;
-  }
-
-  try {
-    await rm(initScriptDir);
-  } catch (err) {
-    error("Oh no!", "Failed to remove template's remix.init script");
+    await cli.run(
```
markdalgleish, line 546: Thanks for updating this! Could we try shelling out to the installed `remix` CLI instead of importing it so we can avoid `create-remix` depending on `@remix-run/dev` since it has a lot of large dependencies? We really want to keep the startup time of the create CLI to a minimum.

leothorp: Good point!

Offhand, calling out to the `remix` CLI did feel a bit heavy (though I didn't measure the impact.)

Makes sense that duplicating it wouldn't be the end of the world. However it seemed like the ideal might be to share only the relevant function's code, without incurring additional build size cost from unused dependencies. This diff shows a proof of concept of this via the use of a new remix-shared-internals package: https://github.com/leothorp/remix/pull/1/files

The builds in that branch are configured so that only the imported functions from this package are pulled into the builds for create-remix/ remix-dev, and anything unused is treeshaken away by Rollup (inspecting the build output confirms that `unusedFunctionToTreeshake` and `unusedFileToTreeshake` aren't present in the final result). The impact on the build sizes appears to be basically equivalent to just duplicating the relevant code. I'd imagine the ability to share code in this way could be useful for other things in the future as well.

Does this seem like a direction worth pursuing here? If not (or if it's too much of a scope expansion for this particular PR), happy to just call out to `remix` or duplicate the code for the time being. Thanks for taking a look.

leothorp: @markdalgleish just a re-ping here if it's helpful, no rush naturally. thanks.

markdalgleish:
@leothorp Sorry for the delay. I'm looking to get this in but wanted to first check whether there was a reason you closed it?

leothorp:
> @leothorp Sorry for the delay. I'm looking to get this in but wanted to first check whether there was a reason you closed it?

@markdalgleish ah, sorry about the confusion- was just cleaning up my open PRs that I wasn't sure were still relevant, but happy to reopen.

leothorp:
@markdalgleish anything else I can do to help get this over the line? no rush of course- and my apologies again for any confusion from closing it earlier.

leothorp:
@markdalgleish is this still relevant?

PR #8118, opened 2023-11-23
docs(route/should-revalidate): Update currentParams code example

SeanRoberts:
Change the logic in the example for currentParams to *prevent* revalidation if only a superfluous part of the param changed. Correct me if I'm wrong but in the existing example code if the IDs were the same but the human friendly-title changed the logic would flow through to `defaultShouldRevalidate` which would return `true` because the params changed.


brophdawg11 (approved):
Thanks!

PR #8120, opened 2023-11-24
fix(remix-dev/vite): remove Vite wrapper functions

markdalgleish:
Based on a suggestion by @pcattori, this PR removes the indirection of calling out to Vite via Remix APIs. Instead, we now keep the API surface area as small as possible and get consumers using custom servers to interact with Vite directly.


pcattori (commented):
Nice! Definitely better than what we had before IMO.

One thing I'm debating now is whether it makes sense to inline a small utility in `server.mjs` just to clean up the ternary / `.then(...)` logic. Something like:

```ts
// server.mjs
async function createDev() {
  if (process.env.NODE_ENV === "production") return undefined
  let vite = await import("vite")
  return vite.createServer({
    server: {
      middlewareMode: true
    }
  })
}

let dev = await createDev()

if (dev) {
  /* ... */
}
// etc.
```

To be clear, this is **not** API we would maintain and export, but rather just code we include as part of the examples/templates to make things a bit easier to follow.

pcattori (approved):

PR #8143, opened 2023-11-27
fix(remix-dev/vite): bundle CSS from client entry

markdalgleish:
Without this fix, CSS from the client entry appears to work in dev (after a flash of unstyled content), but the styles are then missing in the production build.

This PR ensures that:
- In dev, CSS from the client entry is included in the string of critical CSS to avoid FOUC.
- In production, any bundled CSS from the client entry is included with the root route.

To test this, CSS imported from the client entry has been added to Vite CSS tests for both dev and build.

### DIFF
--- a/integration/vite-css-dev-test.ts
+++ b/integration/vite-css-dev-test.ts
@@ -335,7 +335,7 @@ test.describe("Vite CSS dev", () => {
       // Ensure CSS updates were handled by HMR
       await expect(input).toHaveValue("stateful");
 
-      // The following change triggers a full page reload, so we check it after the HMR changes
+      // The following change triggers a full page reload, so we check it after all the checks for HMR state preservation
       let clientEntryCssContents = await fs.readFile(
         path.join(projectDir, "app/entry.client.css"),
         "utf8"
--- a/integration/vite-css-dev-test.ts
+++ b/integration/vite-css-dev-test.ts
@@ -222,8 +222,8 @@ test.describe("Vite CSS dev", () => {
 
   test.describe("with JS", () => {
     test.use({ javaScriptEnabled: true });
-    test("updates CSS", async ({ page }) => {
-      let pageErrors: unknown[] = [];
+    test("updates CSS", async ({ page, browserName }) => {
+      let pageErrors: Error[] = [];
       page.on("pageerror", (error) => pageErrors.push(error));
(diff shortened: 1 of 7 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -426,7 +429,15 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
...
         hasErrorBoundary: sourceExports.includes("ErrorBoundary"),
-        ...resolveBuildAssetPaths(pluginConfig, viteManifest, routeFilePath),
+        ...resolveBuildAssetPaths(
+          pluginConfig,
+          viteManifest,
+          routeFilePath,
+          // If this is the root route, we also need to include assets from the
+          // client entry file as this is a common way for consumers to import
+          // global reset styles, etc.
+          route.parentId === undefined ? [pluginConfig.entryClientFilePath] : []
```
pcattori, line 439: What do you think about creating a variable to make this explicit?

```ts
let isRoot = route.parentId === undefined
```

packages/remix-dev/vite/resolve-file-url.ts
```diff
@@ -0,0 +1,21 @@
+import * as path from "node:path";
+
+export const resolveFileUrl = (
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+  vite: typeof import("vite"),
```
markdalgleish, line 5: Note to self: If all callers of this util are async, can we make this util async and inline the dynamic import of Vite?

integration/vite-css-dev-test.ts
```diff
@@ -300,8 +332,28 @@ test.describe("Vite CSS dev", () => {
         UPDATED_TEST_PADDING_VALUE
       );
 
+      // Ensure CSS updates were handled by HMR
       await expect(input).toHaveValue("stateful");
 
+      // The following change triggers a full page reload, so we check it after the HMR changes
```
pcattori, line 338: Had a bit of trouble understanding this comment on my first read, so took a shot at clarifying:

```suggestion
      // The following change triggers a full page reload, so we check it after all the checks for HMR state preservation
```

Feel free to tweak it more!

pcattori (approved):

PR #8171, opened 2023-11-29
feat(vite): strict route exports

pcattori:
From the added docs:

With Vite, Remix gets stricter about which exports are allowed from your route modules.

Previously, Remix allowed user-defined exports from routes.
The Remix compiler would then rely on treeshaking to remove any code only intended for use on the server from the client bundle.

In contrast, Vite processes each module in isolation during development, so cross-module treeshaking is not possible.
You should already be separating server-only code into `.server` files or directories, so treeshaking isn't needed in those cases.
But routes are a special case since they intentionally blend client and server code.
Remix knows that exports like `loader`, `action`, `headers`, etc. are server-only, so it can safely remove them from the client bundle.
But there's no way to know when looking at a single route module in isolation whether user-defined exports are server-only.
That's why Remix's Vite plugin is stricter about which exports are allowed from your route modules.

In fact, we'd rather not rely on treeshaking for correctness at all.
Treeshaking is designed as a pure optimization, so relying on it for correctness is brittle.
Not only that, but relying on treeshaking also means that if tomorrow you or your coworker accidentally imports something you _thought_ was client safe, you might end up with server code in your client bundle.
In short, Vite made us eat our veggies, but turns out they were delicious!

So instead of treeshaking, its better to be explicit about what code is server only.
For route modules, that means only exporting Remix route exports.
For anything else, put it in a `.server` file or directory instead.

As a simple mental model, think of a Remix route module like a function and the exports like named arguments to the function.

```ts
// Not real API, just a mental model
let route = createRoute({ loader, mySuperCoolThing });
//                                ^^^^^^^^^^^^^^^^
// Object literal may only specify known properties, and 'mySuperCoolThing' does not exist in type
```

Just like how you shouldn't pass unexpected named arguments to a function, you shouldn't create unexpected exports from a route module.
The result is that Remix is simpler and more predictable.

---

Additionally, user-defined exports are a footgun for HMR as the Remix and React Fast Refresh may not have a way to handle updates for custom exports. See https://remix.run/docs/en/main/discussion/hot-module-replacement#supported-exports .

This also results in a nice invariant:

> Code in a Remix route only runs on for that route

👆 that may seem like just fancy words, but reliable invariants like that are invaluable to me when working on complex projects, esp. on larger teams and with Remix beginners.

## Fullstack components?

@kentcdodds wrote a fantastic [blog post about fullstack components](https://www.epicweb.dev/full-stack-components) in Remix. The idea is to colocate a component that depends on API with the route that provides that API. Specifically, Kent recommends colocating the component and loader in the same file.

Notably, the component uses `useFetcher` and manually fetches data for the API route it depends on rather than using `useLoaderData` as the whole point of a fullstack component is to use it anywhere, not just within routes nested under the route defining the loader. Indeed, "nesting" for resource routes isn't meaningful. As a result, the component depends on the API via that API's endpoint.

For example, with the default file routing convention in v2, you can still get colocation of this code in the same directory:

```tsx
// app/routes/api/blah/route.ts

export const loader = () => {}
export type Loader = typeof loader
```

```ts
// app/routes/api/blah/component.ts
import type { Loader } from "./route"

export const FullstackComponent1 = () => {
  let fetcher = useFetcher<Loader>()
  // ... more code here ...
  fetcher.submit(
      { },
      { method: 'get', action: '/api/blah' },
    )
}
```

The only import needed from `app/routes/api/blah/route.ts` is the type for the loader. Notably, user-defined `type` exports are fine since they are stripped before the code runs.

Reasonable people can disagree, but I think this actually models the relationship between the route and the component more accurately. Just like any other part of my Remix code, if I need to access a loader from another route, I can `fetch` that endpoint and grab types from that route's `loader` by importing them.

### DIFF
--- a/docs/future/vite.md
+++ b/docs/future/vite.md
@@ -634,6 +634,43 @@ const posts = import.meta.glob("./posts/*.mdx", {
 });
 ```
 
+#### Strict route exports
+
+With Vite, Remix gets stricter about which exports are allowed from your route modules.
+Previously, the Remix compiler would allow any export from routes.
+While this was convenient, it was also a common source of bugs that were hard to track down because they only surfaced at runtime.
+
+👉 **Move any non-route exports to a separate file**
+
+For example, with take this route:
+
+```ts filename=app/routes/super-cool.tsx
+// ❌ This is the problematic export
+export const mySuperCoolThing =
+  "Some value I wanted to colocate with my route!";
+
+// ✅ This is a valid Remix route export, so it's fine
+export const loader = () => {};
+
+// ✅ This is also a valid Remix route export
+export default function SuperCool() {}
+```
+
+Using the default route convention in v2, you can refactor to:
+
+```ts filename=app/routes/super-cool/route.tsx
+export const loader = () => {};
+
+export default function SuperCool() {}
+```
+
+```ts filename=app/routes/super-cool/utils.ts
+export const mySuperCoolThing =
+  "Some value I wanted to colocate with my route!";
+```
+
+That way your utilities are still colocated with your route, but in a separate module.
+
 ## Troubleshooting
 
 Check out the [known issues with the Remix Vite plugin on GitHub][issues-vite] before filing a new bug report!
--- a/docs/future/vite.md
+++ b/docs/future/vite.md
@@ -637,15 +637,46 @@ const posts = import.meta.glob("./posts/*.mdx", {
 #### Strict route exports
 
 With Vite, Remix gets stricter about which exports are allowed from your route modules.
-Previously, the Remix compiler would allow any export from routes.
-While this was convenient, it was also a common source of bugs that were hard to track down because they only surfaced at runtime.
 
-👉 **Move any non-route exports to a separate file**
+Previously, Remix allowed user-defined exports from routes.
+The Remix compiler would then rely on treeshaking to remove any code only intended for use on the server from the client bundle.
 
-For example, with take this route:
+In contrast, Vite processes each module in isolation during development, so cross-module treeshaking is not possible.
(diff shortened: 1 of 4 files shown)


mjackson (approved):
LGTM. Thanks for being so thorough here!

docs/future/vite.md
```diff
@@ -615,6 +615,43 @@ const posts = import.meta.glob("./posts/*.mdx", {
...
 
+#### Strict route exports
+
+With Vite, Remix gets stricter about which exports are allowed from your route modules.
+Previously, the Remix compiler would allow any export from routes.
+While this was convenient, it was also a common source of bugs that were hard to track down because they only surfaced at runtime.
+
+👉 **Move any non-route exports to a separate file**
+
+For example, with take this route:
```
markdalgleish, line 626:
```suggestion
For example, with this route:
```

docs/future/vite.md
```diff
@@ -615,6 +615,43 @@ const posts = import.meta.glob("./posts/*.mdx", {
 });
 ```
 
+#### Strict route exports
+
+With Vite, Remix gets stricter about which exports are allowed from your route modules.
+Previously, the Remix compiler would allow any export from routes.
+While this was convenient, it was also a common source of bugs that were hard to track down because they only surfaced at runtime.
```
markdalgleish, line 622: However, this approach doesn't work in Vite because it processes each module in isolation during development without any tree shaking. This means that any exports that were only intended for use on the server will also end up on the client, leading to errors or other unexpected behavior.

markdalgleish (approved):
Nice one! I've added a couple of docs suggestions, but this looks great.

PR #8174, opened 2023-11-29
fix (remix-serve): Fix source map loading

kiliman:
This PR fixes the issue where source maps are not getting used. This was traced to how the map file is loaded when HMR reloads the server file with the `?t=timestamp` suffix.

This causes the source map to not be found and therefore not used.

Adding the hook for `retrieveSourceMap`, this strips out the unexpected characters from the filename and properly returns the source map.

Closes: https://github.com/remix-run/remix/discussions/8168.


kiliman:
Ugh. I created 2 PRs back-to-back, one for Remix App Server and one for the Express template. I forgot to branch off of upstream/dev on the 2nd PR, so that template fix is added to this one.

I created the 2nd PR, which only has the template fix, but if you merge this one, that PR is obviated. Since these two are related, I don't think it would be an issue.

Anyway, let me know how you guys want to handle this. Sorry, it's been a while doing PRs.

Alex-Levacher:
@kiliman Thank you so much for the fix 💌

brophdawg11:
This looks good to me on both express and remix-serve.  Here's a screenshot of the before/after (I applied these changes to `remix-serve` directly in `node_modules` and restarted:

Screenshot 2023-12-01 at 9 11 25 AM

I'm going to kick this over to @pcattori and @jacob-ebey for a confirmation though since they're more in tune with the HMR + source maps flows.

brophdawg11 (approved):

pcattori (approved):

IgnisDa:
Hi @kiliman! This PR seems to have introduced a new bug, as mentioned in the above issue. I would be really grateful if you could take a look.

kiliman:
Thanks. I'll take a look in the morning.

kiliman:
@IgnisDa this is fixed in PR #8321

PR #8176, opened 2023-11-29
fix (templates): Add source map retrieval fix to Express template

kiliman:
This PR updates the the Express template and fixes the issue where source maps are not getting used. This was traced to how the map file is loaded when HMR reloads the server file with the `?t=timestamp` suffix.

This causes the source map to not be found and therefore not used.

Adding the hook for `retrieveSourceMap`, this strips out the unexpected characters from the filename and properly returns the source map.

### DIFF
--- a/templates/express/server.js
+++ b/templates/express/server.js
@@ -9,7 +9,19 @@ import express from "express";
 import morgan from "morgan";
 import sourceMapSupport from "source-map-support";
 
-sourceMapSupport.install();
+sourceMapSupport.install({
+  retrieveSourceMap: function (source) {
+    // get source file without the `file://` prefix or `?t=...` suffix
+    const match = source.match(/^file:\/\/(.*)\?t=[.\d]+$/);
+    if (match) {
+      return {
+        url: source,
+        map: fs.readFileSync(`${match[1]}.map`, "utf8"),
+      };
+    }
+    return null;
+  },
+});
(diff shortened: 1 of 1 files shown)


brophdawg11:
What's the best way to recreate the original issue just so I can give this a quick test locally?  I assume this impacts server-side sourcemaps?  I tried putting a `debugger` in a route component and loader and using the VSCode debugger but it took me to the correct source file before and after an HMR update.

kiliman:
Yeah, I wasn't sure how to do an automated test for this.

This is how I'm testing now:
```ts
// routes/_index.tsx
import { Form } from "@remix-run/react";

export async function action() {
  throw new Error("Oops!");
}

  return (
    <div>
      <h1>Welcome to Remix</h1>
      <Form method="post">
        <button>Submit</button>
      </Form>
    </div>
  );
}
```
Click on the submit button. It should throw a server-side exception.

Check the stack trace to make sure it shows the actual source file and not the compiled version.

The issue shows up when you make a change that causes Remix to rebuild (imports index.js?t=timestamp). 

Press the submit button again. It should still show the correct stack trace. Without the fix, it will show the compiled version since it didn't know how to find the map file for the file with the timestamp suffix.

kiliman:
Yeah, sorry I messed up the PRs. It's been awhile.

PR #8177, opened 2023-11-30
fix(vite): add remix-node to optimizeDeps.include

markdalgleish:
The integration test named `vite-dotenv-test` was consistently failing for me when testing in Safari locally and in CI. Debugging locally, I was able to see the following errors in the console:

error

The tests even failed when setting `optimizeDeps.force` to `true`. The only way I was able to get it to pass was to ensure that `@remix-run/node` was also included in the `optimizeDeps.include` array.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -557,6 +557,7 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
               // That means that before Vite pre-bundles dependencies (e.g. first time dev server is run)
               // mismatching Remix routers cause `Error: You must render this element inside a <Remix> element`.
               "@remix-run/react",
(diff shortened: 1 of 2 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -557,6 +557,7 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
               // That means that before Vite pre-bundles dependencies (e.g. first time dev server is run)
               // mismatching Remix routers cause `Error: You must render this element inside a <Remix> element`.
               "@remix-run/react",
+              "@remix-run/node",
```
pcattori, line 560: Can we add a comment for this?

pcattori: https://github.com/remix-run/remix/pull/8185

PR #8189, opened 2023-11-30
Support HMR/HDR for Client Data

brophdawg11:
~~Quick stab at duplicating the current loader HDR logic to apply to `clientLoader`~~

No need to duplicate - can just check the exports slightly differently in the current plugin.

Vite HDR is working.  HMR of `HydrateFallback` is not working currently - see `TODO` comment in the code


packages/remix-dev/devServer_unstable/hdr.ts
```diff
@@ -71,11 +71,23 @@ export let detectLoaderChanges = async (
...
+                `./${file}`
+              )};`;
+            } else if (theExports.includes("loader")) {
               contents = `export { loader } from ${JSON.stringify(
                 `./${file}`
               )};`;
+            } else if (theExports.includes("clientLoader")) {
+              contents = `export { clientLoader } from ${JSON.stringify(
+                `./${file}`
+              )};`;
```
brophdawg11, line 88: Enable HDR for clientLoader by adapting the current tree shaking approach for fine-grained `loader`-change detection to also support `clientLoader`.  Need these branches since `esbuild` will complain if we try to import a non-existent named export

PR #8211, opened 2023-12-04
fix(remix-dev/vite): add Vite commands to dev CLI

markdalgleish:
This PR adds `remix vite:dev` and `remix vite:build` commands to the Remix CLI. We may want to bikeshed these commands, but I personally like this format.

Note, this is an alternative, less-clever solution to https://github.com/remix-run/remix/pull/8206.

The advantages of this approach:
- Again, it's less clever, which is a good thing.
- It's closer to how people expect it to work.
- There's no "uncanny valley" effect where the Vite CLI exposes options that don't make sense (and probably don't work correctly) if a plugin kicks off multiple builds internally (e.g. `--watch`, `--profile`, etc.), which also means consumers can't trust the Vite CLI docs anymore when using Remix. This is a pretty big knock against https://github.com/remix-run/remix/pull/8206.

The disadvantage is that we're further away from being "just a Vite plugin", but if you look at how the build process works in this PR, you can see we're still using the Vite plugin as the source of truth for our build config even when running multiple builds. I think this is a pretty nice hybrid since you don't need a dedicated Remix config. In fact `vite build && vite build --ssr` still works, it just won't be guaranteed to work with different plugin configurations in the future.


packages/remix-dev/cli/run.ts
```diff
@@ -115,6 +130,12 @@ export async function run(argv: string[] = process.argv.slice(2)) {
       "-p": "--port",
       "--tls-key": String,
       "--tls-cert": String,
+
+      // vite
+      // --force and --port are already defined above
+      "--strictPort": Boolean,
+      "--config": String,
+      "--host": String,
```
pcattori, line 138: Let's pass thru all flags that are part of Vite's JS API

pcattori (approved):

PR #8221, opened 2023-12-06
refactor(vite): handle all future flags

markdalgleish:
This PR modifies the Vite plugin to pass the entire `future` config object through to the server entry rather than manually adding each value. This ensures that any new future flags don't also need to be added to the Vite plugin, and it also ensures that the types are always correct. If we ever need to add any Vite-specific future flags, we can handle those within the Vite plugin and then forward the rest.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -334,6 +334,7 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
         serverBuildPath,
         serverModuleFormat,
         relativeAssetsBuildDirectory,
+        future,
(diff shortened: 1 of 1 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -376,11 +374,7 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
       export const assetsBuildDirectory = ${JSON.stringify(
         pluginConfig.relativeAssetsBuildDirectory
       )};
-      ${
-        pluginConfig.future
```
markdalgleish, line 380: This check was redundant since the object always exists.

pcattori (approved):

PR #8236, opened 2023-12-06
Minor client data implementation cleanups

brophdawg11:


packages/remix-react/browser.tsx
```diff
@@ -266,9 +273,10 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
       mapRouteProperties,
     });
 
-    // As long as we didn't SSR a `HydrateFallback`, we can initialize immediately since
-    // there's no initial client-side data loading to perform
-    if (!didServerRenderFallback) {
+    // We can call initialize() immediately if the router doesn't have any
+    // loaders to run on hydration
+    if (router.state.initialized) {
+      routerInitialized = true;
```
brophdawg11, line 279: We don't need `didServerRenderFallback` anymore now that `router.state.initialized` is fixed up for partial hydration

PR #8267, opened 2023-12-12
feat(vite): error messages when .server files are referenced by client

pcattori:
https://github.com/remix-run/remix/assets/1477317/77eceb06-034a-4f6f-89a0-a2215686ada2

## TODO

- [x] changeset

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -962,8 +962,10 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
               "",
               `    '${id}' imported by route '${importerShort}'`,
               "",
-              `  The only route exports that can reference server-only modules are: ${serverOnlyExports}`,
-              `  but other route exports in '${importerShort}' depend on '${id}'.`,
+              `  The only route exports that can reference server-only modules are:`,
+              `    ${serverOnlyExports}`,
+              "",
+              `  But other route exports in '${importerShort}' depend on '${id}'.`,
               "",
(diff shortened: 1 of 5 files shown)


integration/helpers/vite-template/tsconfig.json
```diff
@@ -14,7 +14,9 @@
     "forceConsistentCasingInFileNames": true,
     "baseUrl": ".",
     "paths": {
-      "~/*": ["./app/*"]
+      "~/*": ["./app/*"],
+      "#dir/*": ["./app/.server/*"],
+      "#file": ["./app/utils.server"]
```
pcattori, line 19: this shouldn't be committed

packages/remix-dev/vite/plugin.ts
```diff
@@ -927,22 +929,88 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
...
+
+        throw Error(
+          [
+            colors.red(`Server-only module referenced by client`),
+            "",
+            `    '${id}' imported by '${importerShort}'`,
+            "",
+            `  * If all code in '${importerShort}' is server-only:`,
+            "",
+            `    Rename it to '${dotServerFile}'`,
```
markdalgleish, line 996: I'm wondering if this should also mention `.server` directories?

pcattori: This is meant to be a "quickfix" option to help unblock anyone migrating. We could maybe reword to make that clearer. That said, I can't think of a good way to mention `.server` directories that doesn't feel heavier. Any suggestions?

markdalgleish (approved):
This is so much nicer!

PR #8269, opened 2023-12-12
Enhance client data type inference

brophdawg11:
WIP for enhanced type inference for client data functions.  We should be able to infer when it's a `clientLoader`/`clientAction` based on the types parameters and avoid deserializing if it doesn't return a `Response`...

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -1007,12 +1007,10 @@ export function useMatches(): UIMatch[] {
 type InferDataType<T = AppData> = (
   T extends (...args: any[]) => unknown
     ? ReturnType<T> extends Promise<Response>
-      ? Parameters<T> extends [ClientLoaderFunctionArgs]
-        ? SerializeFrom<T>
-        : Parameters<T> extends [ClientActionFunctionArgs]
-          ? SerializeFrom<T>
-          : Awaited<ReturnType<T>>
-      : SerializeFrom<T>
+      ? SerializeFrom<T>
+      : Parameters<T> extends [ClientLoaderFunctionArgs | ClientActionFunctionArgs]
+        ? Awaited<ReturnType<T>>
+        : SerializeFrom<T>
     : SerializeFrom<T>
 );
 
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -54,6 +54,7 @@ import type {
   MetaMatches,
   RouteHandle,
   ClientLoaderFunctionArgs,
+  ClientActionFunctionArgs,
 } from "./routeModules";
 
 function useDataRouterContext() {
@@ -1001,15 +1002,19 @@ export function useMatches(): UIMatch[] {
   return useMatchesRR() as UIMatch[];
 }
 
+// Deserialize unless it's a clientLoader/clientAction returning a non-Response
 // prettier-ignore
-type InferDataType<T = AppData> =
+type InferDataType<T = AppData> = (
   T extends (...args: any[]) => unknown
-    ? Parameters<T> extends [ClientLoaderFunctionArgs]
-      ? ReturnType<T> extends Promise<Response>
+    ? ReturnType<T> extends Promise<Response>
+      ? Parameters<T> extends [ClientLoaderFunctionArgs]
         ? SerializeFrom<T>
-        : Awaited<ReturnType<T>>
+        : Parameters<T> extends [ClientActionFunctionArgs]
+          ? SerializeFrom<T>
+          : Awaited<ReturnType<T>>
       : SerializeFrom<T>
     : SerializeFrom<T>
+);
 
 /**
  * Returns the JSON parsed data from the current route's `loader`.
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -53,8 +53,6 @@ import type {
   MetaMatch,
   MetaMatches,
   RouteHandle,
-  ClientLoaderFunctionArgs,
-  ClientActionFunctionArgs,
 } from "./routeModules";
 
 function useDataRouterContext() {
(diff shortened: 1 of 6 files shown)


packages/remix-react/components.tsx
```diff
@@ -1000,13 +1002,25 @@ export function useMatches(): UIMatch[] {
...
+// prettier-ignore
+type InferDataType<T = AppData> = (
+  T extends (...args: any[]) => unknown
+    ? ReturnType<T> extends Promise<Response>
+      ? SerializeFrom<T>
+      : Parameters<T> extends [ClientLoaderFunctionArgs | ClientActionFunctionArgs]
+        ? Awaited<ReturnType<T>>
+        : SerializeFrom<T>
+    : SerializeFrom<T>
+);
```
brophdawg11, line 1015: We return `Awaited<ReturnType<T>>` only when the generic is a function that doesn't return a `Response`and has parameters of `ClientLoaderFunctionArgs` or `ClientActionFunctionArgs`

packages/remix-server-runtime/serialize.ts
```diff
@@ -1,16 +1,41 @@
...
+  T extends (...args: any[]) => infer Output
+    ? Output extends Promise<TypedResponse>
+      ? SerializeFromImpl<T>
+      : Parameters<T> extends [ClientLoaderFunctionArgs | ClientActionFunctionArgs]
+        ? Output extends Promise<TypedDeferredData<infer D>>
+          ? D // client defer() is just the raw data
+          : Awaited<Output> // client naked object is just the object
+        : SerializeFromImpl<T>
+    : SerializeFromImpl<T>
+);
```
brophdawg11, line 29: Exporting this as part of `SerializeFrom` ensures we get this across the board in `useLoaderData`, `useActionData`, `useFetcher`, `useRouteLoaderData`, meta `matches.data`, etc.

packages/remix-react/routeModules.ts
```diff
@@ -58,7 +62,7 @@ export type ClientLoaderFunction = ((
  * Arguments passed to a route `clientLoader` function
  */
 export type ClientLoaderFunctionArgs = RRLoaderFunctionArgs<undefined> & {
-  serverLoader: <T = AppData>() => Promise<SerializeFrom<T>>;
+  serverLoader: <T = AppData>() => Promise<SerializeFromImpl<T>>;
```
brophdawg11, line 65: The generics on these calls to the server should only ever be a server loader/action so we can just use `SerializeFromImpl` directly

packages/remix-server-runtime/serialize.ts
```diff
@@ -1,16 +1,41 @@
...
```
pcattori, line 29: Trying out my own suggestion here (https://twitter.com/pcattori/status/1734712411700023443):

```suggestion
export type SerializeFrom<T> = (
  T extends (...args: any[]) => infer Output ?
    Output extends Promise<TypedResponse> ? SerializeFromImpl<T> : 
    Parameters<T> extends [ClientLoaderFunctionArgs | ClientActionFunctionArgs] ?
      Output extends Promise<TypedDeferredData<infer D>> ? D : // client defer() is just the raw data
      Awaited<Output> // client naked object is just the object
    : 
    SerializeFromImpl<T>
  : 
  SerializeFromImpl<T>
);
```

pcattori (changes requested):
Could you add test cases to the bottom of the `serialize.ts` file?

packages/remix-server-runtime/jsonify.ts
```diff
@@ -21,6 +21,9 @@ export type Jsonify<T> =
   T extends Number ? number :
   T extends Boolean ? boolean :
 
+  // Promises JSON.stringify to an empty object
+  T extends Promise<unknown> ? EmptyObject :
```
brophdawg11, line 25: Screenshot 2023-12-13 at 3 26 19 PM

packages/remix-server-runtime/serialize.ts
```diff
@@ -1,21 +1,59 @@
...
  * For example:
  * `type LoaderData = SerializeFrom<typeof loader>`
  */
 export type SerializeFrom<T> =
-  T extends (...args: any[]) => infer Output ? Serialize<Awaited<Output>> :
+  T extends (...args: any[]) => infer Output ?
+    Parameters<T> extends [ClientLoaderFunctionArgs | ClientActionFunctionArgs] ?
+      // Client data functions may not serialize
+      SerializeClientDataFunction<Awaited<Output>>
+    :
```
brophdawg11, line 24: Cleaned it up so that this is all that's new and we don't touch the flow for existing loaders/actions

packages/remix-server-runtime/serialize.ts
```diff
@@ -1,21 +1,60 @@
...
  * For example:
  * `type LoaderData = SerializeFrom<typeof loader>`
  */
 export type SerializeFrom<T> =
-  T extends (...args: any[]) => infer Output ? Serialize<Awaited<Output>> :
+  T extends (...args: any[]) => infer Output ?
+    Parameters<T> extends [ClientLoaderFunctionArgs | ClientActionFunctionArgs] ?
+      // Client data functions may not serialize
+      SerializeClient<Awaited<Output>>
+    :
```
brophdawg11, line 24: Cleaned it up so that this is all that's new and we don't touch the flow for existing loaders/actions

packages/remix-server-runtime/serialize.ts
```diff
@@ -1,21 +1,60 @@
...
+  :
+  Output extends TypedResponse<infer U> ? Jsonify<U> :
+  Awaited<Output>
+
+// prettier-ignore
+type DeferValueClient<T> =
+  T extends undefined ? undefined :
+  T extends Promise<unknown> ? Promise<Awaited<T>> :
+  T;
+
```
brophdawg11, line 57: Copies of the `Serialize`/`DeferValue` types with `Jsonify` stuff removed (unless it's a `TypedResponse` and then we still call `Jsonify`)

packages/remix-server-runtime/serialize.ts
```diff
@@ -49,16 +88,45 @@ type DeferValue<T> =
 
 type Pretty<T> = { [K in keyof T]: T[K] };
 
-type Loader<T> = () => Promise<
-  | TypedResponse<T> // returned responses
-  | TypedResponse<never> // thrown responses
->;
+type Loader<T> = () => Promise<TypedResponse<T>>;
```
brophdawg11, line 91: @pcattori and I worked through some examples and we don't need to represent the never types since they don't show up in the signature if you throw conditionally.

PR #8288, opened 2023-12-14
Update wrangler version

supachaidev:
Because the HMR bug is solved in wrangler v3.19.0 (https://github.com/cloudflare/workers-sdk/issues/4318#issuecomment-1840232207), we can now pin the wrangler version to "^3.20.0".

### DIFF
--- a/templates/cloudflare-pages/package.json
+++ b/templates/cloudflare-pages/package.json
@@ -31,7 +31,7 @@
     "eslint-plugin-react": "^7.33.2",
     "eslint-plugin-react-hooks": "^4.6.0",
     "typescript": "^5.1.0",
-    "wrangler": "3.8.0"
+    "wrangler": "^3.20.0"
   },
   "engines": {
     "node": ">=18.0.0"


brookslybrand:
Thanks @supachaidev! I'm testing now. Will you add it to `templates/cloudflare-workers` as well?

brookslybrand:
Alright, tested and this does seem to fix it!

Unfortunately HMR seems to be broken on `templates/cloudflare-workers` though if you go to `^3.20.0`, I'm not totally sure why. I think that would need to be fixed before merging this in as well

supachaidev:
@brookslybrand With `templates/cloudflare-workers`, the HMR works if you refresh the page once after running the dev server with `npm run dev`. I'm not sure whether this behavior comes from Remix HMR or wrangler.

pmbanugo:
> @brookslybrand With `templates/cloudflare-workers`, the HMR works if you refresh the page once after running the dev server with `npm run dev`. I'm not sure whether this behavior comes from Remix HMR or wrangler.

I noticed this in earlier versions of wrangler perhaps even before v3, while using the workers template. I had to refresh the page once and then the websocket connection for refreshing the page worked after that. Perhaps the workers and pages template use a different semantic under the hood.

I'd advice though that Remix should recommend running on Pages if using Cloudflare and discontinue that template. Cloudflare also recommends people to use Pages for web apps as well. There are more benefits to using Pages than deploying on Workers

shaneuuid:
> > @brookslybrand With `templates/cloudflare-workers`, the HMR works if you refresh the page once after running the dev server with `npm run dev`. I'm not sure whether this behavior comes from Remix HMR or wrangler.
> 
> I noticed this in earlier versions of wrangler perhaps even before v3, while using the workers template. I had to refresh the page once and then the websocket connection for refreshing the page worked after that. Perhaps the workers and pages template use a different semantic under the hood.
> 
> I'd advice though that Remix should recommend running on Pages if using Cloudflare and discontinue that template. Cloudflare also recommends people to use Pages for web apps as well. There are more benefits to using Pages than deploying on Workers

I hope it doesn't go that way. Not everyone wants to and/or is able to use Pages. Deploying to Workers offers tremendous flexibility. It's also the easiest way to get started with Remix in any capacity. Users can go from nothing to global deploys in <5 mins. Pages, while great, is an additional deployment dependency users have to learn and pay for.

PR #8316, opened 2023-12-17
Replace broken placekitten avatar with picsum in tutorial

gregdingle:
This is nearly a simple docs change. 

I was just doing the remix-tutorial and I noticed an image was broken. It looks like `placekitten.com` is simply no longer. With a quick web search, I found `picsum.photos` and a photo on there which seems to serve the same purpose. But feel free to swap in something else. 

I'm not looking to take sides on dogs vs cats 🙂 

**Before**

CleanShot 2023-12-16 at 16 28 40

**After**

CleanShot 2023-12-16 at 16 28 56

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -186,6 +186,7 @@
 - gonzoscript
 - graham42
 - GregBrimble
+- gregdingle
 - GSt4r
 - guatedude2
 - guerra08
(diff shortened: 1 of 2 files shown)


brophdawg11:
https://placekitten.com/ looks OK to me?  Maybe it was just down for a moment when you were looking?

brophdawg11:
Going to close this for now since the photos should be ok

Issue #8309, opened 2023-12-16
Error related to sourcemaps in production build


IgnisDa:
### Reproduction

Using a unstable vite template and building and deploying it gives an error about a sourcemap file missing.

All remix packages are at `v2.4.0`.

[Here](https://github.com/IgnisDa/ryot/blob/2801f78d73cdf0bf81d77740aef28d05cc652e3f/Dockerfile#L56) is the part of the dockerfile that copies relevant files.

### System Info

```shell
System:
    OS: Linux 6.6 Arch Linux
    CPU: (16) x64 AMD Ryzen 7 5800H with Radeon Graphics
    Memory: 19.07 GB / 27.26 GB
    Container: Yes
    Shell: 5.2.15 - /bin/bash
  Binaries:
    Node: 20.5.1 - ~/.proto/bin/node
    Yarn: 4.0.0 - ~/.proto/bin/yarn
    npm: 10.2.1 - /usr/sbin/npm
    bun: 1.0.7 - ~/.proto/bin/bun
```

### Used Package Manager

yarn

### Expected Behavior

The production server to work fine.

### Actual Behavior

Here is the stderr for the server:
```
Error: ENOENT: no such file or directory, open '/home/ryot/build/server/index.js.map'
    at readFileSyncUtf8 (node:internal/fs/read/utf8:20:3)
    at Object.readFileSync (node:fs:467:12)
    at Array.retrieveSourceMap (/home/ryot/node_modules/@remix-run/serve/dist/cli.js:49:37)
    at /home/ryot/node_modules/source-map-support/source-map-support.js:85:24
    at mapSourcePosition (/home/ryot/node_modules/source-map-support/source-map-support.js:216:21)
    at wrapCallSite (/home/ryot/node_modules/source-map-support/source-map-support.js:397:20)
    at Function.prepareStackTrace (/home/ryot/node_modules/source-map-support/source-map-support.js:446:39)
    at maybeOverridePrepareStackTrace (node:internal/errors:140:29)
    at prepareStackTrace (node:internal/errors:114:5)
    at getStackString (node:internal/util/inspect:1239:16)
```

VHall1:
Can confirm, I'm having the same issue on Windows 11.

As far as I can tell, it seems like this is related to the [2.4.0 release](https://github.com/remix-run/remix/releases/tag/remix%402.4.0). Pegged all remix packages locally to version 2.3.1 and that seems to have "solved" the issue.

[...]

PR #8321, opened 2023-12-18
fix (remix-serve): convert file url to path when loading source maps

VHall1:
This PR fixes an issue where the leading slash on source map paths would cause remix-serve to crash when an error gets thrown on Windows.

Tested both on Windows and Mac.

Closes: [8309](https://github.com/remix-run/remix/issues/8309)

### DIFF
--- a/.changeset/angry-poets-accept.md
+++ b/.changeset/angry-poets-accept.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/serve": patch
+---
+
+Use node `fileURLToPath` to convert source map URL to path
--- a/contributors.yml
+++ b/contributors.yml
@@ -623,3 +623,4 @@
 - timkraut
 - alexanderson1993
 - signed
+- VHall1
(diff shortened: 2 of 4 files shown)


kiliman:
Ugh. I should have tested my "fix" on Windows.

Seems like something that `source-map-support` should automatically handle.

I verified that `url.fileURLToPath()` works correctly on macOS and Windows. It even ignores the search param.

Thanks for fixing this.

kiliman (approved):

brophdawg11 (approved):
Thanks!

brophdawg11:
Thank you fore the PR, but https://placekitten.com/ is working fine for me, it may have just had a temporary outage

brophdawg11:
Sorry for the accidental close - commented/closed this thinking I was on a different PR 🤦

kiliman:
Ha, yeah. @brophdawg11 this fixes the source-map issue from the previous PR but now works on Windows (yeah, I should have tested it on Windows, primarily since it's path-related 😳)

Issue #8296, opened 2023-12-14
🗺️  Server Bundles (formerly known as Server Builds (formerly known as Bundle Shredding))


ryanflorence:
- PR: https://github.com/remix-run/remix/pull/8332

- Cloudflare Pages
- Vercel build API

B3nten:
netlify to round out the trifecta?

pcattori:
https://github.com/remix-run/remix/pull/8596 removed `unstable_` prefix from server bundles, so this is just waiting for Vite to stabilize.

markdalgleish:
Since this is primarily to support Vercel at the moment, I'd consider this open until the `adapter` API stabilises, but there's still a little more work left on that.

PR #8332, opened 2023-12-19
feat(remix-dev/vite): Add unstable_serverBundles

markdalgleish:
See updated docs for details, but the high level summary is that this lets you split your server into multiple bundles with a request handler in each, with each one designed to handle a subset of routes.

We also generate a manifest in `build/server/bundles.json` so consumers have everything they need to direct requests to the correct server bundle. We may provide higher level utilities for consuming the manifest, but this is good enough to get the ball rolling.

Also see https://github.com/remix-run/remix/issues/8296.


docs/future/server-bundles.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+Each `route` object in the `branch` array contains the following properties:
+
+- `id` — The unique ID for this route, named like its `file` but relative to the app directory and without the extension, e.g. `app/routes/gists.$username.tsx` will have an `id` of `routes/gists.$username`.
+- `path` — The path this route uses to match on the URL pathname.
+- `file` — The absolute path to the entry point for this route.
+- `index` — Whether or not this route is an index route.
+- `caseSensitive` — Whether or not the `path` is case-sensitive.
+- `parentId` — The unique `id` for this route's parent route, if there is one.
```
markdalgleish, line 43: TODO: Prune this object.

markdalgleish, line 43: TODO: Prune this object.

integration/vite-server-bundles-test.ts
```diff
@@ -0,0 +1,364 @@
...
+      await withBundleServer(cwd, "bundle-a", async (port) => {
+        await page.goto(`http://localhost:${port}/bundle-a`);
+        await expect(
+          page.locator('[data-route-file="bundle-a._index.tsx"] [data-mounted]')
+        ).toBeVisible();
+
+        await page.goto(`http://localhost:${port}/bundle-a`);
+        await expect(
+          page.locator('[data-route-file="bundle-a._index.tsx"] [data-mounted]')
+        ).toBeVisible();
```
pcattori, line 166: duplicate test case

packages/remix-dev/vite/build.ts
```diff
@@ -29,7 +36,149 @@ async function extractRemixPluginConfig({
...
+    branch.push(route);
+    currentRouteId = route.parentId;
+  }
+
+  return branch.reverse();
+}
+
+function normalizePath(filePath: string) {
+  return filePath.replace(/\\/g, "/");
+}
```
pcattori, line 84: consider using vite's `normalizePath` for consistency

pcattori (dismissed):

Issue #8336, opened 2023-12-20
create-remix template issues installing deps and importing via default path alias


swalker326:
### Reproduction

[repo](https://github.com/swalker326/remix-import-error)

`npx create-remix@latest`

![Screenshot 2023-12-19 at 7 17 01 PM](https://github.com/remix-run/remix/assets/8460808/560487e8-5742-480f-a793-18718d8db2a7)

create any component and try to import it via path alias `import { Component } from "~/components/Component";`

error:
`'Component' is declared but its value is never read.ts(6133)`

I was getting some failing imports but unable to reproduce. 

### System Info

```shell
System:
    OS: macOS 14.1.2
    CPU: (10) arm64 Apple M1 Max
    Memory: 166.30 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.2.0 - ~/.nvm/versions/node/v20.2.0/bin/node
    npm: 9.6.6 - ~/.nvm/versions/node/v20.2.0/bin/npm
    pnpm: 8.12.1 - ~/.nvm/versions/node/v20.2.0/bin/pnpm
    bun: 1.0.2 - /opt/homebrew/bin/bun
    Watchman: 2023.11.27.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 119.0.6045.199
    Safari: 17.1.2
  npmPackages:
    @remix-run/css-bundle: ^2.4.0 => 2.4.0
    @remix-run/dev: ^2.4.0 => 2.4.0
    @remix-run/node: ^2.4.0 => 2.4.0
    @remix-run/react: ^2.4.0 => 2.4.0
    @remix-run/serve: ^2.4.0 => 2.4.0
```

### Used Package Manager

npm

### Expected Behavior

Importing with the path alias throws no error.

### Actual Behavior

Imports throw an error saying cannot resolve.

swalker326:
The issue is in the .eslintrc.cjs 

My initial approach was to install a new package, `eslint-import-resolver-typescript` and update `settings.["import/resolver"].typescript` to an empty object. This fix works but only if I move that "import/resolver" key outside of settings, which is not where the docs say it should go. 

Upon further tinkering I found that if I just move the key outside of settings, everything seems to work correctly, even without installing the package. 

![Screenshot 2023-12-19 at 9 50 01 PM](https://github.com/remix-run/remix/assets/8460808/6d5855b0-e138-4d6c-a30d-6dcffa73bcda)

[...]

PR #8337, opened 2023-12-20
Fix 8336 eslint path alias error

swalker326:
Closes: #8336 

Testing Strategy:

1. add a test component
1. triggered the error by importing it with the ~/
1. applied the fix to the .eslintrc
1. restarted eslint to make sure that:
  -the error was resolved on the import
  -eslint was running without error


templates/remix/app/components/Component.tsx
```diff
@@ -0,0 +1,5 @@
+export function Component() {
+  return (
+    <div>Am I real?</div>
+  )
+}
```
brophdawg11, line 5: I confirmed this seems to work locally - mind removing this new component addition/usage before we get this merged?

brophdawg11:
Thanks for the PR!  We'll also want to add this to all the other `templates/` (except `remix-tutorial` and `remix-javascript` I think) but I'm happy to do that part once this is ready for the primary template :)

templates/remix/package.json
```diff
@@ -25,7 +25,8 @@
     "@typescript-eslint/eslint-plugin": "^6.7.4",
     "eslint": "^8.38.0",
     "eslint-config-prettier": "^9.0.0",
-    "eslint-plugin-import": "^2.28.1",
+    "eslint-import-resolver-typescript": "^3.6.1",
```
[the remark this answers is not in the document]
glitchassassin: On my end, it looks like the moved key is just causing eslint to crash before reporting any errors (other eslint errors are also suppressed)

templates/remix/app/components/Component.tsx
swalker326: Yea I was in the process of updating them when I realized something was still a bit off. Per [this comment](https://github.com/remix-run/remix/pull/8337#discussion_r1434506330)

I'm leaving the key in the correct place, and adding the package required for paths to work to each template. Tested it locally and it seems to work correctly. Sound good?

swalker326:
So after much pain and troubleshooting I tracked the issue down to `"plugin:import/recommended"` Also chased down the issue in [their repo](https://github.com/import-js/eslint-plugin-import/issues/2765) and it looks like they maybe have a fix waiting deployment. Not sure if it will resolve the issue though. Removing the plugin allowed the project to lint, typecheck and run without issue. 

I also went into each template and did the following:

1. add a test component
2. triggered the error by importing it with the ~/ 
3. applied the fix to the .eslintrc
4. restarted eslint to make sure that: 
   - the error was resolved on the import 
   - eslint was running without error

brophdawg11:
Hm, are we sure we want to remove the preset entirely?  Doesn't that remove other potentially valuable rules just to get rid of the one problematic one?  Personally, I'd prefer a targeted fix to the problematic rule.  Full disclosure - I don't know what else is in the preset - I'll try to poke in and check soon though

brophdawg11:
I'm going to close this in favor of #8441 so we can keep the "recommended" preset and just fix the alias issue by adding the new `settings` config

Issue #7639, opened 2023-10-11
🗺️ Remix SPA Target


ryanflorence:
- RFC: #7638

brophdawg11:
To be released as `unstable_ssr` in 2.5.0 via #8457

PR #8338, opened 2023-12-20
SPA Mode

brophdawg11:
**Todo:**
* [ ] Tests
* [ ] Docs
* [ ] Template

RFC: https://github.com/remix-run/remix/discussions/7638
Closes: https://github.com/remix-run/remix/issues/7639

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -348,6 +348,9 @@ export interface RemixConfig {
 
   /**
    * Enable SPA Mode.  Default's to `false`.
+   *
+   * This is the inverse of the user-level `ssr` config and used throughout
+   * the codebase to avoid confusion with Vite's `ssr` config
    */
   isSpaMode: boolean;
 
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -487,7 +487,7 @@ export async function resolveConfig(
       let file = path.basename(userEntryServerFile);
       throw new Error(`SPA Mode: ${file} is not permitted in SPA Mode`);
     }
-    entryServerFile = `entry.server.spa.tsx`;
+    entryServerFile = "entry.server.spa.tsx";
 
     // TODO: Should we check the rest of the server related configs and throw
     // errors if present?  We need to be able to create and run a request handler
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -483,10 +483,7 @@ export async function resolveConfig(
   if (isSpaMode) {
     if (userEntryServerFile) {
       let file = path.basename(userEntryServerFile);
-      throw new Error(
-        `You cannot have an ${file} file when you have configured ` +
-          "`ssr:false` in your `remix.config.js`"
-      );
+      throw new Error(`SPA Mode: ${file} is not permitted in SPA Mode`);
     }
     // TODO: Probably should do the same for the rest of the server-related
     // configs?  We need to be able to create and run a request handler to
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -161,10 +161,10 @@ export interface AppConfig {
   serverPlatform?: ServerPlatform;
 
   /**
-   * Enable server-side rendering for your application. When disabled, Remix
-   * will request the `/` path at build-time and save it as an `index.html` file
-   * with your assets so your application can be deployed as a SPA without
-   * server-rendering. Default's to `true`.
+   * Enable server-side rendering for your application. Disable to use Remix in
+   * "SPA Mode", which will request the `/` path at build-time and save it as
+   * an `index.html` file with your assets so your application can be deployed
+   * as a SPA without server-rendering. Default's to `true`.
    */
   ssr?: boolean;
 
@@ -347,12 +347,9 @@ export interface RemixConfig {
   serverPlatform: ServerPlatform;
 
   /**
-   * Enable server-side rendering for your application. When disabled, Remix
-   * will request the `/` path at build-time and save it as an `index.html` file
-   * with your assets so your application can be deployed as a SPA without
-   * server-rendering. Default's to `true`.
+   * Enable SPA Mode.  Default's to `false`.
    */
-  ssr?: boolean;
+  isSpaMode: boolean;
 
   /**
    * Whether to support Tailwind functions and directives in CSS files if `tailwindcss` is installed.
@@ -451,7 +448,6 @@ export async function resolveConfig(
   serverMainFields ??=
     serverModuleFormat === "esm" ? ["module", "main"] : ["main", "module"];
   serverMinify ??= false;
-  let isSpaMode = appConfig.ssr === false;
 
   let serverNodeBuiltinsPolyfill = appConfig.serverNodeBuiltinsPolyfill;
   let browserNodeBuiltinsPolyfill = appConfig.browserNodeBuiltinsPolyfill;
@@ -480,18 +476,19 @@ export async function resolveConfig(
   let pkgJson = await PackageJson.load(rootDirectory);
   let deps = pkgJson.content.dependencies ?? {};
 
+  let isSpaMode = appConfig.ssr === false;
   if (isSpaMode) {
+    // Don't allow the user to provide their own entry.server.tsx - we'll use an
+    // internal renderToString implementation since no streaming is required
     if (userEntryServerFile) {
       let file = path.basename(userEntryServerFile);
       throw new Error(`SPA Mode: ${file} is not permitted in SPA Mode`);
     }
-    // TODO: Probably should do the same for the rest of the server-related
-    // configs?  We need to be able to create and run a request handler to
-    // generate index.html, so we need _some_ aspect of server available.
-
-    // No server loaders = no need for streaming, so we can just use a super-simple
-    // renderToString implementation
     entryServerFile = `entry.server.spa.tsx`;
+
+    // TODO: Should we check the rest of the server related configs and throw
+    // errors if present?  We need to be able to create and run a request handler
+    // to generate index.html, so we need _some_ aspect of "server" available.
   } else if (userEntryServerFile) {
     entryServerFile = userEntryServerFile;
   } else {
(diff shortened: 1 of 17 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -1336,3 +1368,41 @@ async function getRouteMetadata(
...
+) {
+  // Create a handler and call it for the `/` path - rendering down to the
+  // proper HydrateFallback ... or not!  Maybe they have a static landing page
+  // generated from routes/_index.tsx.
+  let build = await import(serverBuildPath);
+  let { createRequestHandler: createNodeRequestHandler } = await import(
+    "@remix-run/node"
+  );
+  let handler = createNodeRequestHandler(build, mode);
+  let response = await handler(new Request("http://localhost/"));
```
jrestall, line 1385: Please consider the suggestion documented at https://github.com/remix-run/remix/discussions/7638#discussioncomment-7918072 to allow either an index.html or index.js to be generated. It's the smallest change which allows the generation of a Remix App library file that I could think of but maybe the team has better ideas.

packages/remix-dev/config.ts
```diff
@@ -462,7 +479,22 @@ export async function resolveConfig(
...
+    // access to the `ssr:false` flag in the vite config (the streaming template
+    // works just fine so maybe instea dof having this we _only have this version
+    // in the template...).  We let users manage an entry.server file in SPA mode
+    // so they can de ide if they want to hydrate the full document or just an
+    // embedded `<div id="app">` or whatever.
+    entryServerFile = "entry.server.spa.tsx";
+
+    // TODO: Should we check the rest of the server related configs and throw
+    // errors if present?  We need to be able to create and run a request handler
+    // to generate index.html, so we need _some_ aspect of "server" available.
```
brophdawg11, line 496: @pcattori @markdalgleish Curious to get your thoughts here - worth throwing build-time errors for a subset of configs that don't make sense in a SPA world?

packages/remix-dev/vite/plugin.ts
```diff
@@ -1060,6 +1073,25 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
...
+            .filter((exp) => SERVER_ONLY_EXPORTS.includes(exp));
+          if (serverOnlyExports.length > 0) {
+            let message = [
+              `SPA Mode: ${serverOnlyExports.length} invalid route export${
+                serverOnlyExports.length > 1 ? "s" : ""
+              } in \`${route.file}\`:`,
+              ...serverOnlyExports.map((exp) => `  - \`${exp}\``),
+              "",
+              // TODO: Docs!
+              //"See https://remix.run/docs/en/main/future/vite#strict-route-exports",
```
brophdawg11, line 1088: Note to self

packages/remix-react/browser.tsx
```diff
@@ -198,7 +200,10 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
     // towards determining the route matches.
     let initialPathname = window.__remixContext.url;
     let hydratedPathname = window.location.pathname;
-    if (initialPathname !== hydratedPathname) {
+    if (
+      initialPathname !== hydratedPathname &&
+      !window.__remixContext.isSpaMode
+    ) {
```
brophdawg11, line 206: Don't perform this check in SPA mode since we want to be able to hydrate any URL via `route.lazy`

packages/remix-react/browser.tsx
```diff
@@ -213,48 +218,56 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
-          hydrationData.loaderData[routeId] = undefined;
-        } else if (manifestRoute && !manifestRoute.hasLoader) {
-          // Since every Remix route gets a `loader` on the client side to load
-          // the route JS module, we need to add a `null` value to `loaderData`
-          // for any routes that don't have server loaders so our partial
-          // hydration logic doesn't kick off the route module loaders during
-          // hydration
-          hydrationData.loaderData[routeId] = null;
+    let hydrationData = undefined;
+    if (!window.__remixContext.isSpaMode) {
```
brophdawg11, line 226: This is just a whitespace change - we don't have to prep any `hydrationData` in SPA mode

packages/remix-react/components.tsx
```diff
@@ -605,12 +605,22 @@ export type ScriptProps = Omit<
...
 
+  // Use these `matches` instead :)
+  // In SPA Mode we only want to import root on the critical path, since we
+  // want the generated HTML file to be able to be loaded at non-/ paths as
+  // well.  This lets the router handle initial match loads via lazy().
+  //
+  // TODO: If users only ever intend to serve the `/` path from index.html and
+  // thus they want to include all initial matches here...should we give them
+  // a config to do so?
+  let matches = isSpaMode ? [dontUseTheseMatches[0]] : dontUseTheseMatches;
```
brophdawg11, line 622: @mjackson @ryanflorence Thoughts?  I like the ability to load any URL (`/`, `/path`, `/path/child`) via index.html (like you can do in vite/RR) - but in order to do so we have to slightly de-optimize by not choosing which route to pre-load.  If a user wanted to only ever support landing on index.html via `/`, we could optimize and include the sub-matches here.

packages/remix-react/routes.tsx
```diff
@@ -497,10 +550,12 @@ function getRouteModuleComponent(routeModule: RouteModule) {
...
 export function shouldHydrateRouteLoader(
   route: EntryRoute,
-  routeModule: RouteModule
+  routeModule: RouteModule,
+  isSpaMode: boolean
 ) {
   return (
-    routeModule.clientLoader != null &&
-    (routeModule.clientLoader.hydrate === true || route.hasLoader !== true)
+    (isSpaMode && route.id !== "root") ||
```
brophdawg11, line 557: Never hydrate below the root in SPA mode

packages/remix-dev/config.ts
markdalgleish: I think it makes sense to throw when resolving the plugin config.

brophdawg11: 👍 I'll probably ping you or @pcattori to make sure I come up with the right list of configs to throw on

marbemac:
Just a heads up - tailwind users will likely run into the following error (importing css file from `root.tsx`):

Screenshot 2023-12-27 at 2 54 06 PM

I guess this is because the typical tailwind config file targets all ts/tsx files in the project (along lines of below), and spa mode shifts the pre/css transform stuff to be run in client context or something?

```ts
import type { Config } from 'tailwindcss';

export default {
  content: [
    'app/**/*.{ts,tsx}',
  ],
} satisfies Config;
```

pcattori:
@marbemac that issue should be fixed by #8353

packages/remix-react/components.tsx
brophdawg11: We agreed we do want to hydrate at any path

docs/guides/spa-mode.md
```diff
@@ -0,0 +1,96 @@
...
+- "Your boss couldn't care less about the UX ceiling of SPA architecture and won't give your dev teams time/capacity to re-architect things" [- Kent C. Dodds][kent-tweet]
+
+That's why we added support for **SPA Mode** in [2.5.0][2.5.0] (per this [RFC][rfc]), which builds heavily on top of the [Client Data][client-data] APIs.
+
+## What is SPA Mode?
+
+SPA Mode is basically what you'd get if you had your own [React Router + Vite][rr-setup] setup using `createBrowserRouter`/`RouterProvider`, but along with some extra Remix goodies:
+
+- File-based routing (or config-based via [`routes()`][routes-config])
+- Automatic route-based code-spitting via [`route.lazy`][route-lazy]
```
david-crespo, line 21: So every route is automatically turned into a separate file in the Vite output? Any way to opt out of this? My [main work app](https://github.com/oxidecomputer/console) (a Vite+RR app I'm excited to try SPA mode on) has one main bundle and two secondary ones for big libraries, and I like that!

```
dist/assets/Terminal-wrK8jj1j.js                            287.61 kB │ gzip:  71.75 kB │ map:   590.14 kB
dist/assets/TimeSeriesChart-wrSD895q.js                     370.29 kB │ gzip: 103.16 kB │ map: 1,646.72 kB
dist/assets/app-DtIhrgNj.js                               1,038.45 kB │ gzip: 323.32 kB │ map: 4,913.94 kB
```

marbemac: Curious - what's the benefit?

brophdawg11: Not at the moment - that's just how the Remix manifest works under the hood.  I don't know enough of the vite configs though to know if you have any more granular control over the chunking logic. 

You don't mind the cost of your larger-than-neccesary startup bundle?  If the goal is quicker client-side navs, then I'd reach for `<Link prefetch="render">`

david-crespo: That asset is downloaded once and cached in the browser, so no matter what page they hit after that (except for 3 or 4 pages that use the extra chunks) there is nothing to download. I figured in practice the extra transfer time on uncached pageload would not be the dominant factor in perceived performance, but now that I'm checking it, you might be right and it would be better to DL less up front. I'll give SPA mode a try either way as soon as it's out!

aarjithn: Another point is larger bundles invalidate more often - so you could potentially get better caching with smaller bundles.

david-crespo: Another very interesting point, thank you. It hadn’t occurred to me the hash filenames might _ever_ be the same across releases (my one big file always changes its hash, of course). I’ll have to test that out.

brophdawg11:
@Abbbdab Thanks for testing this out!  Would you mind filing Github issues with steps to reproduce and tagging me if you run into problems?  That way we can keep this PR restricted to code-related comments.  Thanks!

packages/remix-dev/config.ts
brophdawg11: It's only really `unstable_serverBundles` - added a warning in https://github.com/remix-run/remix/pull/8338/commits/b80ca86b5b70d1b9464d5f8233361bfec6e130c1.  Went with a warning since it feels weird to totally break a build if someone wants to just quickly try out SPA mode.  Instead we just ignore the `unstable_serverBundles` config

jacob-ebey (approved):
useId test please :)

packages/remix-dev/vite/plugin.ts
brophdawg11: Still pondering this, but we're going to get this out as an `unstable_ssr` flag without this functionality to start.  Hopefully that'll let some folks play around with it and we can gather some feedback and continue to evaluate this use case.

PR #8382, opened 2023-12-27
Update UIMatch type to cover cases where loaderData is undefined

lifeiscontent:
when using a UIMatch loaderData may be undefined in the scenario that the loader failed, so this makes the types a bit more safe making sure that consumers of the UIMatch type check for `undefined`

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -986,7 +986,7 @@ function dedupe(array: any[]) {
 }
 
 export type UIMatch<D = AppData, H = RouteHandle> = UIMatchRR<
-  SerializeFrom<D>,
+  SerializeFrom<D> | undefined,
   H
 >;
 


brophdawg11:
Is this specifically for `ErrorBoundary` use-cases?

lifeiscontent:
@brophdawg11 yep

brophdawg11:
I chatted with @mjackson on this and we agreed we don't want this to _always_ have `undefined` on the types since that is incorrect the vast majority of the time - as `data` is always populated in components.  

Additionally, `match.data` is `unknown` out of the box and only gets the `loaderData` when you effectively cast it via the generic:

```ts
let matches = useMatches();
let match = matches[matches.length - 1] as UIMatch<typeof loader>;
```

So I think in reality that is where the issue arises.  If you're using `useMatches` inside an `ErrorBoundary`, you shouldn't be saying "trust me, data is the typeof the loader" since that loader could have thrown.

I think the more accurate usage in boundaries is:

```ts
let matches = useMatches();
let match = matches[matches.length - 1] as UIMatch;
let data = match.data as SerializeFrom<typeof loader> | undefined
```

That way you're properly casting according to the error scenario.

If any TS ninjas want to look into making `as UIMatch<typeof loader | undefined>` achieve that - I think we'd potentially be open to that.  But for now I think the manual cast is the right solution.

PR #8385, opened 2023-12-27
chore: use isbot version 4

### LINKED ISSUE #8407: isbot v4 breaking change in templates and hidden files
### Reproduction

- Run `npx create-remix@latest` with all default options
- Delete `entry.server.ts`
- Run `npm remove isbot`
- Run `npm install` to make sure the dependency is removed from `node_modules`
- Run `npm run dev`

It will print the following and add `"isbot": "latest"` to `package.json`:

```
adding `isbot` to your package.json, you should commit this change
```

- The app should fail to build.

**Note: As of this issue being filed, it appears a new version has been marked as `latest`, so the breaking v4 version is no longer installed. As such, the reproduction doesn't break anything. However, the breaking v4 version will be marked as `latest` soon.**

If you want to break it now, run `npm install isbot@next` to install the new `isbot@4`.

https://www.npmjs.com/package/isbot#version-4-will-become-the-latest-version-on-npm-on-january-2024

### System Info

```shell
System:
    OS: macOS 13.6.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 31.58 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.18.2 - ~/.volta/tools/image/node/18.18.2/bin/node
    Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 9.8.1 - ~/.volta/tools/image/node/18.18.2/bin/npm
    pnpm: 8.6.10 - ~/.volta/bin/pnpm
    bun: 1.0.11 - ~/.bun/bin/bun
    Watchman: 2023.11.20.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 120.0.6099.129
    Safari: 17.0
  npmPackages:
    @remix-run/css-bundle: ^2.4.1 => 2.4.1 
    @remix-run/dev: ^2.4.1 => 2.4.1 
    @remix-run/node: ^2.4.1 => 2.4.1 
    @remix-run/react: ^2.4.1 => 2.4.1 
    @remix-run/serve: ^2.4.1 => 2.4.1
```

### Used Package Manager

npm

### Expected Behavior

The dev server should build and start up without issue.

### Actual Behavior

The app fails to build with an invalid import error.

fifi98:
The latest `isbot` version 3 has been marked as deprecated in favour of the version 4 which is going to become the "latest" version on npm on January 2024 (https://github.com/omrilotan/isbot). 

Version 4 includes some breaking changes listed here: https://github.com/omrilotan/isbot/blob/next/CHANGELOG.md. The main point of this release seems to be reducing the package size and improving performance.

This pull requests makes Remix use the latest `isbot` version, which is 4.1.0 at the moment. Templates and docs have been also updated to reflect that change.

Closes #8407


brophdawg11:
Thanks!!

brophdawg11:
Unfortunately, I realized we can't upgrade to v4 like this after all because it would be a breaking change to existing apps with v3 pinned in their deps who have not run `npx remix reveal` and are using the `entry.server.tsx` version built-into `@remix-run/dev`, since v3 does not export `isbot` as a named export.  I'm going to revert this for now and we'll figure out the best way to solve this.

Issue #8392, opened 2023-12-28
[VITE] isBot crashes the server in 2.4.1 if isbot is over v4


ZeldOcarina:
### Reproduction

https://stackblitz.com/edit/remix-run-remix-vwrwea

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (8) arm64 Apple M1
    Memory: 198.97 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    bun: 1.0.3 - ~/.bun/bin/bun
  Browsers:
    Chrome: 120.0.6099.129
    Safari: 17.2.1
  npmPackages:
    @remix-run/dev: ^2.4.1 => 2.4.1 
    @remix-run/eslint-config: ^2.4.1 => 2.4.1 
    @remix-run/express: ^2.4.1 => 2.4.1 
    @remix-run/node: ^2.4.1 => 2.4.1 
    @remix-run/react: ^2.4.1 => 2.4.1 
    vite: ^5.0.10 => 5.0.10
```

### Used Package Manager

npm

### Expected Behavior

Server should start with isbot since it's being used as part of entry.server.ts.

### Actual Behavior

`isbot` is being used in `entry.server.ts`, when upgrading to 2.4.1 and using:

```js
app.all(
  "*",
  createRequestHandler({
    // @ts-expect-error
    build: viteDevServer
      ? () => viteDevServer.ssrLoadModule("virtual:remix/server-build")
      : await import("./build/server/index.js"),
  }),
);
```

server won't start and crashes with àTypeError: __vite_ssr_import_4__.default is not a function`

ZeldOcarina:
Fix is simple:

![image](https://github.com/remix-run/remix/assets/42982095/d3b001cc-7ee0-4ca8-ba28-3bb6d6500bc7)

It just took half an hour to debug, maybe you want to update ìsbot`dependency to latest version and adjust the starter? If it has already been done, sorry for this and pls disregard.

michaeldebetaz:
Thanks for the fix ! Undofortunately, the expected type seems to have changed for the `isbot` function. In `entry.server.tsx`, `request.headers.get()` returns `string` or `null`, but `isbot()` expects `string`
[...]

PR #8415, opened 2024-01-02
Handle isbot v4 in a backwards-compatible manner

### LINKED ISSUE #8407: isbot v4 breaking change in templates and hidden files

brophdawg11:
**Background:**
* Remix apps need `isbot` installed for use by the default provided `entry.server.tsx` (and the one internally used by `remix dev` if one is not present in the app)
* `remix dev` adds `isbot` to `package.json` if it does not exist but it does so with the `latest` tag
* `isbot` released v4 on 1/1/2024 and moved from a default to a named export
* Any Remix apps that did a fresh install post-1/1/2024 that got the `latest` v4 version ran into build/runtime issues since their `entry.server` (or the internal one) assumed a default export which no longer existed
  * This should only be a problem for fresh installs that re-evaluate `latest` - it should not be an issue AFAIK for apps using lockfiles since those should have the v3 version pinned internally
* The fix for those apps is to just downgrade `isbot` from `latest` to `^3` in their `package.json`

**Solution**
* Moving forward, we'll use `isbot` v4 and `remix dev` will update package.json with `"isbot": "^4"` instead of `latest`
* However, the internal templates used by `remix dev` can't assume v4 since there are Remix apps in the wild that may have `"isbot": "3.x.y"` in their `package.json` (if they changed it from `latest`) that are using the internal entry.server.tsx file. 
* So, we need our internal `remix dev` defaults to handle both `isbot` v3 or v4
* Templates are updated to v4 so any net-new apps moving forward will install v4, and the `entry.server.tsx` in templates will assume v4 for net new apps

Reimplements #8385 which was reverted in #8414
Closes #8407, #8392


templates/arc/app/entry.server.tsx
```diff
@@ -39,6 +39,25 @@ export default function handleRequest(
...
+  }
+
+  // isbot@4
+  if ("isbot" in isbot && typeof isbot.isbot === "function") {
+    return isbot.isbot(userAgent);
+  }
+
+  return false;
+}
+
```
MichaelDeBoey, line 60: I guess templates can use v4 as they will only be used when creating new apps?

MichaelDeBoey (changes requested):

templates/arc/app/entry.server.tsx
brophdawg11: 🤣 yeah I realized this too late yesterday and left a post-it note on my desk as I logged off: _"remove isbot changes from templates?"_

docs/guides/migrating-react-router-app.md
```diff
@@ -66,7 +66,7 @@ export default function handleRequest(
   remixContext: EntryContext,
   loadContext: AppLoadContext
 ) {
-  return isbot(request.headers.get("user-agent"))
+  return isbot(request.headers.get("user-agent") || "")
```
brophdawg11, line 69: All docs are updated to reflect the v4 API

integration/defer-test.ts
```diff
@@ -997,7 +997,7 @@ test.describe("aborted", () => {
             remixContext: EntryContext,
             loadContext: AppLoadContext
           ) {
-            return isbot(request.headers.get("user-agent"))
+            return isbot(request.headers.get("user-agent") || "")
```
brophdawg11, line 1000: All internal unit/e2e tests and the repo itself are updated to use v4

packages/remix-dev/config.ts
```diff
@@ -493,7 +493,7 @@ export async function resolveConfig(
       pkgJson.update({
         dependencies: {
           ...pkgJson.content.dependencies,
-          isbot: "latest",
+          isbot: "^4",
```
brophdawg11, line 496: going forward, `remix dev` will add v4 to the `package.json` instead of `latest` - this is the primary "fix"

packages/remix-dev/config/defaults/entry.server.cloudflare.tsx
```diff
@@ -22,7 +22,7 @@ export default async function handleRequest(
     }
   );
 
-  if (isbot(request.headers.get("user-agent"))) {
+  if (isBotRequest(request.headers.get("user-agent"))) {
```
brophdawg11, line 25: Internal implementations of `entry.server.tsx` will support either v3 or v4 for backward compatibility with existing Remix apps who have pinned v3 in their package.json

templates/arc/app/entry.server.tsx
```diff
@@ -24,7 +24,7 @@ export default function handleRequest(
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   loadContext: AppLoadContext
 ) {
-  return isbot(request.headers.get("user-agent"))
+  return isbot(request.headers.get("user-agent") || "")
```
brophdawg11, line 27: Templates use v4 API moving forward

MichaelDeBoey (approved):

templates/arc/app/entry.server.tsx
omrilotan: The use case has been addressed with version 4.2.0 to accept "null" as input, with the intention to keep user code cleaner.

https://github.com/omrilotan/isbot/blob/main/CHANGELOG.md#420
> Accept null in place of user agent string to allow header value to be used "as is" (request.headers.get("user-agent"))

docs/guides/migrating-react-router-app.md
omrilotan: This use case is supported as is starting isbot@4.2.0

```suggestion
  return isbot(request.headers.get("user-agent"))
```

PR #8426, opened 2024-01-03
docs(testing): remove redundant usage of `waitFor`

kentcdodds:
This is a simple docs change. Wrapping a `find*` query in `waitFor` is unnecessary since that's what it's doing anyway.


markdalgleish (approved):

markdalgleish:
Thank you!

PR #8435, opened 2024-01-04
Add "eslint-import-resolver-typescript" to the remix template

rossipedia:
When creating a new project with `create-remix@latest`, if you attempt to import a `*.css` file (for instance, when using Tailwind), eslint complains about the resolver:

image

The fix is to run `npm install -D eslint-import-resolver-typescript`, after which eslint will no longer complain about the wrong interface or the `*.css` file.


brophdawg11:
I did this and grabbed the original fix from #8337 in #8441 so closing this in favor of that

PR #8437, opened 2024-01-05
docs(discussions/hot-module-replacement): add "Survive 'require' Cache Purge" section

gramotei:
Based on this https://stackoverflow.com/a/72662281

### DIFF
--- a/docs/discussion/hot-module-replacement.md
+++ b/docs/discussion/hot-module-replacement.md
@@ -184,3 +184,34 @@ In some cases, React cannot distinguish between existing components being change
 [meta]: ../route/meta
 [use-loader-data]: ../hooks/use-loader-data
 [react-keys]: https://react.dev/learn/rendering-lists#why-does-react-need-keys
+
+## Survive 'require' Cache Purge
+
+Remix purges the require cache on every request in development to support `<LiveReload/>`. To make sure your cache survives these purges, you need to assign it to the global object.
+
+```ts
+// utils/singleton.server.ts
+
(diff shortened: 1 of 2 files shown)


hilja:
It's mentioned here though https://remix.run/docs/en/main/guides/manual-mode#keeping-in-memory-server-state-across-rebuilds It also mentions Kent’s package, which is great https://www.npmjs.com/package/@epic-web/remember

gramotei:
I think it's worth moving to HMR section and add mention in manual. I would never look into manual mode because that's another topic

docs/discussion/hot-module-replacement.md
```diff
@@ -184,3 +184,34 @@ In some cases, React cannot distinguish between existing components being change
...
 [use-loader-data]: ../hooks/use-loader-data
 [react-keys]: https://react.dev/learn/rendering-lists#why-does-react-need-keys
+
+## Survive 'require' Cache Purge
+
+Remix purges the require cache on every request in development to support `<LiveReload/>`. To make sure your cache survives these purges, you need to assign it to the global object.
+
+```ts
+// utils/singleton.server.ts
+
```
MichaelDeBoey, line 194:
```suggestion
```ts filename=utils/singleton.server.ts
```

MichaelDeBoey (commented):
Not sure if we need to include this, since we already have https://remix.run/docs/en/main/guides/manual-mode#keeping-in-memory-server-state-across-rebuilds in the docs 🤔

Issue #8443, opened 2024-01-07
tsc fails on fresh unstable-vite install due to remix & vite typescript conflicts


setvik:
### Reproduction

Compiling a fresh install of unstable-vite remix as follows:
```
npx create-remix@latest --template remix-run/remix/templates/unstable-vite vitetest
cd vitetest
# Using Typescript (tsc) 5.3.2
npx tsc
```

results in 49 typescript errors due to module definition conflicts between remix and vite:
```
Found 49 errors in 4 files.

Errors  Files
    23  node_modules/@remix-run/dev/dist/modules.d.ts:3
     2  node_modules/@remix-run/react/dist/browser.d.ts:30
    23  node_modules/vite/client.d.ts:8
     1  node_modules/vite/types/importMeta.d.ts:17
```

### System Info

```shell
NOTE: Tested with Typescript 5.3.2

System:
    OS: macOS 14.2.1
    CPU: (8) arm64 Apple M2
    Memory: 4.35 GB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.3.1 - ~/.nvm/versions/node/v20.3.1/bin/node
    Yarn: 1.22.21 - ~/.nvm/versions/node/v20.3.1/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v20.3.1/bin/npm
  Browsers:
    Chrome: 120.0.6099.129
    Safari: 17.2.1
  npmPackages:
    @remix-run/dev: ^2.4.1 => 2.4.1
    @remix-run/node: ^2.4.1 => 2.4.1
    @remix-run/react: ^2.4.1 => 2.4.1
    @remix-run/serve: ^2.4.1 => 2.4.1
    vite: ^5.0.0 => 5.0.11
```

### Used Package Manager

npm

### Expected Behavior

Compiling a fresh install of unstable-vite with `npx tsc` should complete successfully.
```
npx create-remix@latest --template remix-run/remix/templates/unstable-vite vitetest
cd vitetest
# NOTE: Using Typescript (tsc) 5.3.2
npx tsc
```

### Actual Behavior

```
Found 49 errors in 4 files.

Errors  Files
    23  node_modules/@remix-run/dev/dist/modules.d.ts:3
     2  node_modules/@remix-run/react/dist/browser.d.ts:30
    23  node_modules/vite/client.d.ts:8
     1  node_modules/vite/types/importMeta.d.ts:17
```

[...]

PR #8444, opened 2024-01-07
fix typescript conflicts between remix and vite

setvik:
Closes: #8443

Testing Strategy:

```
npx create-remix@latest --template remix-run/remix/templates/unstable-vite vitetest
cd vitetest

# update remix typescript definitions included in this PR
vim packages/remix-dev/modules.ts
vim packages/remix-react/browser.tsx

npx tsc
```

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -530,6 +530,7 @@
 - selfish
 - sergiocarneiro
 - sergiodxa
+- setvik
 - shairez
 - shamsup
 - shashankboosi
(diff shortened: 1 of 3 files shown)


markdalgleish (commented):
Thanks for the PR!

I tested this locally, and the asset type fixes only appear to work because this actually breaks the types.

This is because the `"*.(module.css)"` style syntax you've used doesn't work as intended. In effect this is just defining types for file names that literally end in `.(module.css)`, e.g. `app/styles/test.(module.css)`. I've confirmed locally that this unintended behaviour is what's happing instead.

As a result I'm closing this in favour of #8447.

The type clash with `@remix-run/react` should be fixable by removing the global type augmentation since it's not part of the public API. I might do this as a follow-up PR.

PR #8446, opened 2024-01-07
fix (remix-serve): fallback to default path if source maps aren't available when building stack trace

VHall1:
Another update to source map loading. This should resolve issues where people reported getting server crashes when they had source maps disabled or when errors were being thrown by 3rd party packages (since source maps aren't available for those). All this does really is check that the file exists before loading it, and returning null if it doesn't (using the original path as a fallback). Mostly based on the patch suggested here: https://github.com/remix-run/remix/issues/8309#issuecomment-1868374521

Testing:

**3rd party error (tiny-invariant)**
image

**generic error**
image

**generic error after manually deleting source map**
image

Closes: #8309

### DIFF
--- a/.changeset/dry-books-hug.md
+++ b/.changeset/dry-books-hug.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/serve": minor
+---
+
+Allow remix-serve CLI to fallback to default stack trace if source maps are not available
--- a/.changeset/dry-books-hug.md
+++ b/.changeset/dry-books-hug.md
@@ -1,5 +1,5 @@
 ---
-"@remix-run/serve": minor
+"@remix-run/serve": patch
 ---
 
-Allow remix-serve CLI to fallback to default stack trace if source maps are not available
+Don't try to load sourcemaps if they don't exist on disk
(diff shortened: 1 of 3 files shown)


VHall1:
Not sure if this warrants a version bump, but happy to add a changeset if it does.

brophdawg11 (approved):
Thanks!

brophdawg11:
> Not sure if this warrants a version bump, but happy to add a changeset if it does.

yeah - since this changes source code, it should have a changeset.  Would you mind pointing the PR to to `dev` as well?

VHall1:
yep, sounds good to me! Will add the changeset and point to the correct branch later today.

VHall1:
sorry for the mess, should be pointing to dev now.

.changeset/dry-books-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/serve": minor
+---
+
+Allow remix-serve CLI to fallback to default stack trace if source maps are not available
```
brophdawg11, line 5:
```suggestion
---
"@remix-run/serve": patch
---

Don't try to load sourcemaps if they don't exist on disk
```

PR #8458, opened 2024-01-08
fix(create-remix): only update "*" versions for Remix dependencies

ahuth:
This PR updates `create-remix` so that it only updates "*" dependencies for `@remix-run/*` packages.

For example, when create-remix runs, we currently get this change

```diff
"dependencies": {
-  "@remix-run/react": "*",
+  "@remix-run/react": "^2.4.1",
```

Unfortunately, this also affects non-Remix dependencies, such as:

```diff
"dependencies": {
-  "@remix-run/react": "*",
+  "@remix-run/react": "^2.4.1",
-  "@myorg/some-random-dep": "*"
+  "@myorg/some-random-dep": "^2.4.1"
```

which breaks when version 2.4.1 (or whatever the current Remix release is) doesn't exist.

Closes: # n/a

- [ ] Docs
- [x] Tests

Testing Strategy:

Unit test

### DIFF
--- a/.changeset/wicked-suits-heal.md
+++ b/.changeset/wicked-suits-heal.md
@@ -1,0 +1,5 @@
+---
+"create-remix": patch
+---
+
+Only update `*` versions for `@remix-run/*` dependencies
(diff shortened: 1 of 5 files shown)


markdalgleish (approved):
Thanks for the PR!

PR #8459, opened 2024-01-08
fix(remix-run/react): fix import.meta.hot type clash

markdalgleish:
This PR is another fix related to #8443.

The global `import.meta.hot` types defined in `@remix-run/react` are clashing with the types provided by Vite. Since these types were not intended for consumers, I've opted to simplify and just use `@ts-expect-error` directives in the two spots that we use `import.meta.hot` (which was typed as `any` anyway).

### DIFF
--- a/.changeset/ninety-ducks-sin.md
+++ b/.changeset/ninety-ducks-sin.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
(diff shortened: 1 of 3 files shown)


pcattori (approved):

jacob-ebey (approved):
🚢 it. I always considered this API surface internal, and I seriously doubt any apps are actually relying on custom HMR logic.

PR #8468, opened 2024-01-10
fix(remix-dev/vite): tree-shake unused route exports

markdalgleish:
Similar to our existing compiler, this PR wraps all client route modules with a virtual module that re-exports all Remix client route exports. This means that any custom route exports that aren't consumed within the client module graph will be removed in the production build. I've added a test asserting that this is the case.

Note that while this is only of value in the production build, I've used the same technique in dev for consistency. This makes it easier to understand how Remix works from both a debugging and a teaching perspective.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -46,7 +46,20 @@ const supportedRemixConfigKeys = [
 type SupportedRemixConfigKey = typeof supportedRemixConfigKeys[number];
 type SupportedRemixConfig = Pick<RemixUserConfig, SupportedRemixConfigKey>;
 
-const SERVER_ONLY_EXPORTS = ["loader", "action", "headers"];
+const SERVER_ONLY_ROUTE_EXPORTS = ["loader", "action", "headers"];
+const CLIENT_ROUTE_EXPORTS = [
+  "clientAction",
+  "clientLoader",
+  "default",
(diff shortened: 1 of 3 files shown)


pcattori (approved):

packages/remix-dev/vite/plugin.ts
```diff
@@ -796,10 +809,27 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
         });
         await viteChildCompiler.pluginContainer.buildStart({});
       },
-      transform(code, id) {
+      async transform(code, id) {
```
pcattori, line 812: Reading through the [Vite conventions for virtual modules](https://vitejs.dev/guide/api-plugin#virtual-modules-convention) again and wondering if we should do something to ensure the client routes don't get processed/handled by other plugins. We could just prepend with `\0` or model them fully as vmods.

markdalgleish: I think it should be okay if it's handled by other plugins. In terms of resolving the module, since it's a real file on disk, it'll be treated the same as the original route file. In terms of transforming the file, it's going to behave as if you'd manually written a re-export file, which also would work fine.

The benefit of this approach from what I understand is that Vite will treat this pseudo-virtual module the same as the route file in terms of file changes, invalidation, HMR. This means we don't need to do anything special to invalidate virtual modules when the underlying route changes.

PR #8469, opened 2024-01-10
feat(vite): support adding plugin config via env var

TooTallNate:
Extra Vite plugin configuration may be supplied by importing the file specified by the `REMIX_VITE_EXTRA_CONFIG` env var. This allows for additional configuration values to be specified at build-time without modifying the `vite.config` file.


markdalgleish (commented):
Thanks for the PR!

Just closing this based on our offline discussion. We're opting to support this use case via an adapter API instead: https://github.com/remix-run/remix/pull/8514

PR #8476, opened 2024-01-10
POC of Library Mode for Remix SPAs

brophdawg11:
This is a POC of what it might look like to add support for https://github.com/remix-run/remix/discussions/7638#discussioncomment-7912638 which would allow Remix SPA's to be built as embeddable JS files instead of full HTML documents.

I don't think we can rely on Vite's `library` config directly here since once you set that it changes up the Vite output pretty significantly.

### DIFF
--- a/integration/spa-mode-test.ts
+++ b/integration/spa-mode-test.ts
@@ -16,11 +16,13 @@ test.describe("SPA Mode", () => {
   let fixture: Fixture;
   let appFixture: AppFixture;
 
-  test.beforeAll(async () => {
-    fixture = await createFixture({
-      compiler: "vite",
-      files: {
-        "vite.config.ts": js`
+  test.describe("shared build", () => {
+    test.beforeAll(async () => {
+      fixture = await createFixture({
+        compiler: "vite",
+        spaMode: true,
+        files: {
+          "vite.config.ts": js`
           import { defineConfig } from "vite";
           import { unstable_vitePlugin as remix } from "@remix-run/dev";
 
@@ -28,7 +30,7 @@ test.describe("SPA Mode", () => {
             plugins: [remix({ unstable_ssr: false })],
           });
         `,
-        "app/root.tsx": js`
+          "app/root.tsx": js`
           import * as React from "react";
           import { Form, Link, Links, Meta, Outlet, Scripts } from "@remix-run/react";
 
@@ -92,7 +94,7 @@ test.describe("SPA Mode", () => {
             );
           }
         `,
-        "app/routes/_index.tsx": js`
+          "app/routes/_index.tsx": js`
           import * as React  from "react";
           import { useLoaderData } from "@remix-run/react";
 
@@ -123,7 +125,7 @@ test.describe("SPA Mode", () => {
             );
           }
         `,
-        "app/routes/about.tsx": js`
+          "app/routes/about.tsx": js`
           import { useActionData, useLoaderData } from "@remix-run/react";
 
           export function meta({ data }) {
@@ -153,7 +155,7 @@ test.describe("SPA Mode", () => {
             );
           }
         `,
-        "app/routes/error.tsx": js`
+          "app/routes/error.tsx": js`
           import { useRouteError } from "@remix-run/react";
 
           export async function clientLoader({ serverLoader }) {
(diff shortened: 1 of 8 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -1492,15 +1492,34 @@ async function handleSpaMode(
...
-      colors.bold(path.relative(process.cwd(), assetsBuildDirectory)) +
-      " directory"
-  );
+  let html = await response.text();
+
+  if (response.headers.get("Content-Type") === "application/javascript") {
+    await fse.writeFile(
+      path.join(assetsBuildDirectory, "assets", "index.js"),
+      html
+    );
```
brophdawg11, line 1501: If the user returns a `application/javascript` response from the handler, we write it out as JS instead of HTML

packages/remix-react/components.tsx
```diff
@@ -594,6 +594,65 @@ export type ScriptProps = Omit<
...
+    // If we got a handoff, insert it after the imports (used for SPA mode library mode)
+    handoff ? remixContextJsString(handoff) : "",
+    `window.__remixRouteModules = {${matches
+      .map((match, index) => `${JSON.stringify(match.route.id)}:route${index}`)
+      .join(",")}}`,
+    `import(${JSON.stringify(manifest.entry.module)});`,
+  ]
+    .filter((l) => l)
+    .join("\n");
+}
```
brophdawg11, line 625: Extracted from `Scripts` for re-use

packages/remix-react/components.tsx
```diff
@@ -594,6 +594,65 @@ export type ScriptProps = Omit<
...
+//       status: 200,
+//     });
+//   }
+export function UNSAFE_remixSpaModeLibraryJs(ctx: EntryContext) {
+  return remixEntryJsString(
+    ctx.manifest,
+    ctx.staticHandlerContext.matches,
+    ctx.serverHandoffString
+  );
+}
```
brophdawg11, line 654: This is the key - right now, the suggestion in the RFC is to basically recreate this logic which is all implementation details.  If we expose a utility to do this outside of a React render tree, it can be used without rendering `<RemixServer>` in `entry-server.tsx`

jrestall: Thanks so much for looking into this Matt. Exporting this logic as a utility will make the approach much simpler.

With this, teams can incrementally migrate legacy PHP/ASP.NET sites to Remix by embedding a Remix SPA.

packages/remix-dev/vite/plugin.ts
```diff
@@ -1527,34 +1527,46 @@ async function handleSpaMode(
...
+  if (response.headers.get("Content-Type") === "application/javascript") {
+    let jsDir = path.join(assetsBuildDirectory, "assets");
+    let jsPath = path.join(jsDir, "index.js");
+    await fse.writeFile(jsPath, text);
+
+    viteConfig.logger.info(
+      "SPA Mode: index.js has been written to your " +
+        colors.bold(path.relative(process.cwd(), jsDir)) +
+        " directory"
+    );
```
brophdawg11, line 1548: If the user returns a `application/javascript` response from the handler, we write it out as JS instead of HTML

packages/remix-dev/vite/plugin.ts
```diff
@@ -120,6 +122,13 @@ export type RemixVitePluginOptions = RemixConfigJsdocOverrides &
      * as a SPA without server-rendering. Default's to `true`.
      */
     unstable_ssr?: boolean;
+    /**
+     * Specify a JS filename to enable "library mode" which will write out a
+     * JS file instead of an `index.html` file so you can embed your Remix SPA
+     * Mode app as a library in an HTML page generated elsewhere.  Requires
+     * `unstable_ssr: false`.
+     */
+    unstable_library?: string;
```
brophdawg11, line 131: When in spa mode, users can specify an `unstable_library 'app.js'` config to specify they want the app written out as a JS file they can embed into HTML generated by some other system.

packages/remix-dev/vite/plugin.ts
```diff
@@ -1483,46 +1501,72 @@ async function getRouteMetadata(
...
+      ["root"],
+      createServerHandoffString({
+        state: {},
+        url: "http://localhost/",
+        future,
+        isSpaMode: true,
+      })
     );
-  }
+    await fse.writeFile(jsPath, js);
```
brophdawg11, line 1526: Best viewed with whitespace changed hidden, but when `library` is specified we use the new `remixEntryJsString` utility to write out a JS file containing the same "hydration" JS that `<Scripts>` would normally generate.

packages/remix-react/components.tsx
```diff
@@ -778,26 +783,10 @@ export function Scripts(props: ScriptProps) {
...
-          .map(
-            (match, index) => `${JSON.stringify(match.route.id)}:route${index}`
-          )
-          .join(",")}};
-
-import(${JSON.stringify(manifest.entry.module)});`;
+      : remixEntryJsString(
+          manifest,
+          matches.map((m) => m.route.id)
+        );
```
brophdawg11, line 789: Internal `<Scripts>` logic extracted to internal `UNSAFE` helpers in `@remix-run/server-runtime/serverHandoff` so we can also leverage them from the vite plugin

packages/remix-dev/vite/plugin.ts
```diff
@@ -1001,12 +1015,16 @@ export const remixVitePlugin: RemixVitePlugin = (options = {}) => {
             );
           }
 
-          if (pluginConfig.isSpaMode) {
+          if (isSpaMode) {
             await handleSpaMode(
+              // TODO: Is this readily available anywhere?
+              await ssrBuildContext.getManifest(),
```
brophdawg11, line 1021: @pcattori @markdalgleish Is it an issue to generate this on demand like this?  Is there a cached version anywhere I can access?

integration/spa-mode-test.ts
```diff
@@ -308,152 +310,232 @@ test.describe("SPA Mode", () => {
...
-      await app.goto("/");
-      expect(await page.locator("[data-route]").textContent()).toBe("Index");
+  test.describe("One-off builds", () => {
+    test("writes out a JS file in library mode", async ({ page }) => {
+      fixture = await createFixture({
+        compiler: "vite",
+        spaMode: true,
+        library: "app.js",
+        files: {
+          "vite.config.ts": js`
```
brophdawg11, line 470: This test shows the basic public-facing usage for library mode:

* set `unstable_library: "app.js"` in your vite plugin config
* `root.tsx` should return a subset of document HTML, not an `<html>` document
* change `entry.client.tsx` to `createRoot(div).render()`, not `hydrateRoot(document)`
* then, serve your library within some other HTML doc (this is handled internally by our test fixture)
  * `<div id="app"/><script type="module" src="/assets/app.js"></script>`

packages/remix-server-runtime/entry.ts
```diff
@@ -28,7 +28,10 @@ export interface AssetsManifest {
   routes: RouteManifest<EntryRoute>;
   url: string;
   version: string;
-  hmrRuntime?: string;
+  hmr?: {
+    timestamp?: number;
+    runtime: string;
+  };
```
brophdawg11, line 34: @pcattori I'm pretty sure this is just a type bug fix - but could use confirmation on that and that this isn't by design for any reason

brophdawg11:
Closing this as we decided it did not make sense as a first class API, see [this comment](https://github.com/remix-run/remix/discussions/7638#discussioncomment-8149881).

PR #8495, opened 2024-01-12
Fix quirks mode warning for SPA template

mrm007:
1. Set up a new Remix app using the instructions for [SPA Mode](https://remix.run/docs/en/main/future/spa-mode#usage)
1. Start the dev server with `remix vite:dev`
1. Open Firefox Dev Tools and observe the warning:

```
This page is in Quirks Mode. Page layout may be impacted. For Standards Mode use “<!DOCTYPE html>”
```

### DIFF
--- a/.changeset/wild-timers-impress.md
+++ b/.changeset/wild-timers-impress.md
@@ -2,4 +2,4 @@
 "@remix-run/dev": patch
 ---
 
-Fix quirks mode warning for SPA template and default server entry
+Automatically prepend `<!DOCTYPE html>` if not present to fix quirks mode warnings for SPA template
--- a/.changeset/wild-timers-impress.md
+++ b/.changeset/wild-timers-impress.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Fix quirks mode warning for SPA template and default server entry
--- a/packages/remix-dev/config/defaults/entry.server.spa.tsx
+++ b/packages/remix-dev/config/defaults/entry.server.spa.tsx
@@ -9,9 +9,9 @@ export default function handleRequest(
   responseHeaders: Headers,
   remixContext: EntryContext
 ) {
-  let html = renderToString(
-    <RemixServer context={remixContext} url={request.url} />
-  );
+  let html =
+    "<!DOCTYPE html>\n" +
+    renderToString(<RemixServer context={remixContext} url={request.url} />);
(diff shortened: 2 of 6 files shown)


packages/remix-dev/config/defaults/entry.server.spa.tsx
```diff
@@ -9,9 +9,9 @@ export default function handleRequest(
   responseHeaders: Headers,
   remixContext: EntryContext
 ) {
-  let html = renderToString(
-    <RemixServer context={remixContext} url={request.url} />
-  );
+  let html =
+    "<!DOCTYPE html>\n" +
+    renderToString(<RemixServer context={remixContext} url={request.url} />);
```
brophdawg11, line 14: We probably shouldn't do this blindly.  There is no guarantee that root returns a complete `<html>` document.  It's possible (in SSR and SPA) to hydrate only a portion of the document by returning a `<div>` from `root.tsx` and updating `entry.client.tsx` to hydrate a specific element.  It's not the recommended usage - but we do like to leave as much control in the app developers hands and make minimal internal assumptions.

I think it would probably be safe to do this conditionally though?

```suggestion
  let html = renderToString(
    <RemixServer context={remixContext} url={request.url} />
  );
  if (html.startsWith('<html')) {
    html = "<!DOCTYPE html>\n" + html;
  }
```

brophdawg11:
Thank you for the PR!  Would you mind adding a changeset and pointing this to `dev` since it touches source code?

brophdawg11:
Also worth nothing that if this is an issue for anyone before this update gets released, you can `npx remix reveal` and make this change in your own `app/entry.server.tsx`

packages/remix-dev/config/defaults/entry.server.spa.tsx
mrm007: That's a very good point

.changeset/wild-timers-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Fix quirks mode warning for SPA template and default server entry
```
brophdawg11, line 5:
```suggestion
Automatically prepend `<!DOCTYPE html>` if not present to fix quirks mode warnings for SPA template
```

brophdawg11:
Thanks @mrm007!

mrm007:
Thanks for adding tests!

PR #8520, opened 2024-01-15
docs(vite): document adding additional Vite dev server options

ovi-ispas:
Document adding additional Vite dev server options.

### DIFF
--- a/docs/future/vite.md
+++ b/docs/future/vite.md
@@ -182,6 +182,21 @@ export default defineConfig({
 });
 ```
 
+Additional dev server options such as changing the default dev port (5173) or automatically opening the app in the browser on server start can be provided using Vite's [server options](https://vitejs.dev/config/server-options.html):
+```ts filename=vite.config.ts lines=[2-5]
+export default defineConfig({
+  server: {
+    port: 3000,
+    open: true,
+  },
+  plugins: [
+    remix({
+      ignoredRouteFiles: ["**/.*"],
+    }),
+  ],
+});
+```
+
 #### TypeScript integration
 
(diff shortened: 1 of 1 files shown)


brophdawg11:
@pcattori and @markdalgleish should make the call here, but IMO I don't know that we should single out the `server` aspect of `vite` here since there's plenty of stuff you can/should configure in vite outside of the remix plugin.  I think we want to avoid trying to duplicate the vite config docs in our docs and I like that we already link out to them higher up: 

> This means that, for any additional bundling features you'd like to use, you should reference [Vite documentation][vite] and the [Vite plugin community][vite-plugins] rather than the Remix documentation.

> Vite has many [features][vite-features] and [plugins][vite-plugins] that are not built into the existing Remix compiler.
The use of any such features will render the existing Remix compiler unable to compile your app, so only use them if you intend to use Vite exclusively from here on out.

Maybe we could expand on that slightly and include a bulleted list with links to links to a few common vite configs you might want to override in a Remix app (`server`, `build`, `mode`, etc.)

ovi-ispas:
Thanks for the comment.

Totally valid point @brophdawg11. My initial intention with this PR was not to redundantly add vite configs unrelated  to the remix plugin itself, more like to mention that in the context of the _migration_ you have to explicitly also set the port back to what you were used to (3000) if that is what you want, so it won't be a surprise when the app starts on 5173. 

Maybe that is not that relevant and just unnecessary information, in which case feel free to close and disregard this PR. 👍

markdalgleish (approved):
Thanks for the PR!

Your point about this being a migration issue makes sense so I've moved it into the Remix App Server migration docs and reworded it so it only focuses on the port changing.

PR #8529, opened 2024-01-16
docs(assets): link to supported asset file types

tmcw:
This documentation is currently misleading – only a subset of files are importable with asset imports. This links to the source code and summarizes its scope.


markdalgleish (approved):
Thanks for the PR!

PR #8531, opened 2024-01-16
Cloudflare support for Vite

pcattori:
This currently relies on `wrangler@beta`, so let's wait to merge this until the next stable release of `wrangler` that includes [`getBindingsProxy`](https://github.com/cloudflare/workers-sdk/pull/4523). CF core team says that should happen next week or so.

## Usage

```ts
// vite.config.ts
import {
  unstable_vitePlugin as remix,
  unstable_vitePluginAdapterCloudflare as cloudflare,
} from "@remix-run/dev";

export default defineConfig({
  plugins: [
    remix({
      adapter: await cloudflare(),
    }),
  ],
});
```

The `cloudflare` adapter reads bindings from `wrangler.toml` and sets them on the `env` field of load context:

```ts
// app/routes/_index.tsx

// ... some code ...

const key = "__my-key__";

export async function loader({ context }: LoaderFunctionArgs) {
  const { MY_KV } = context.env;
  const value = await MY_KV.get(key);
  return json({ value });
}

// ... some more code ...
```

## TODO

- [x] `build` script: `ssr` conditions as part of adapter API
- [x] docs
- [x] tests
- [x] changeset


cloudkite:
Exciting 🙌 does the build target workers or pages? Or can we choose the build target? Thanks!

pcattori:
> Exciting 🙌 does the build target workers or pages? Or can we choose the build target? Thanks!

Pages. CF core team flagged that [Workers is now deprecated](https://developers.cloudflare.com/workers/configuration/sites/) since Pages now is able to do everything Workers could do

sergiodxa:
@pcattori I would like to try this on my own website since I deploy to CF Pages, how could I do it? Would it be enough if I use patch-package to patch the Vite plugin to include the changes here?

jfsiii:
@pcattori thanks for highlighting "do not use Worker Sites for new projects".

We have a production app using Sites and I'd love to adopt Vite without having to also switch to Pages. It's clear we'll need to eventually but that prerequisite pushes Vite so much further away.

Would supporting Workers and Pages be a lot of extra work/complexity/etc? Perhaps Workers could be supported for some bridge period then removed later?

pcattori:
@sergiodxa I'll include instructions for trying things out once its in a good state for that. Right now there are a couple pieces missing.

@jfsiii I haven't looked into how much work supporting CF Workers would be. There's a decent chance that this work will already support CF Workers but not something we test for right now, so would love feedback from y'all if you try it out.

templates/unstable-vite-cloudflare/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "dev": "remix vite:dev",
+    "build": "remix vite:build",
+    "start": "wrangler pages dev ./build/client --kv MY_KV",
+    "deploy": "wrangler pages deploy ./build/client",
+    "lint": "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint .",
+    "typecheck": "tsc"
+  },
+  "dependencies": {
+    "@remix-run/cloudflare": "2.5.0",
+    "@remix-run/cloudflare-pages": "2.5.0",
```
markdalgleish, line 15: These should be `*`?

packages/remix-dev/vite/adapter.ts
```diff
@@ -0,0 +1,85 @@
+import type {
```
pcattori, line 1: should change this file back to be named `node/adapter.ts`

templates/unstable-vite-cloudflare/.cloudflare/adapter.ts
```diff
@@ -0,0 +1,18 @@
+// Eventually, `getCloudflareDevBindings` will be replaced by the official `getBindingsProxy` API from CloudFlare
+// https://github.com/cloudflare/workers-sdk/pull/4523
+// At that point, Remix will absorb this adapter so you can import it directly in your `vite.config.ts`:
+//
+//   import { adapter as cloudflare } from "@remix-run/cloudflare"
+//
+// Until then, this `.cloudflare/` directory provides the adapter implementation.
```
pcattori, line 7: Create a `.cloudflare/README.md` to put this explanation into, re-export this adapter from `.cloudflare/index.ts`, and add a comment in `vite.config.ts` that guides users to read more in the README if they want.

pcattori:
CF merged the [`getBindingsProxy` PR](https://github.com/cloudflare/workers-sdk/pull/4523) and that is now available in latest beta release of `wrangler`. They said stable release should be sometime next week, so let's wait for that and then we should be able to nuke the entire `.cloudflare/` dir in the template. It also means we can read CF bindings from `wrangler.toml` instead of requiring duplication of those bindings in `vite.config.ts`!

templates/unstable-vite-cloudflare/.cloudflare.ts
```diff
@@ -0,0 +1,7 @@
+import { getBindingsProxy } from "wrangler";
+
+export const adapter = () => async () => {
```
pcattori, line 3: TODO: this file should be removed and this adapter should come from `@remix-run/cloudflare`

templates/unstable-vite-cloudflare/functions/[[path]].ts
```diff
@@ -0,0 +1,14 @@
+import { type ServerBuild } from "@remix-run/cloudflare";
+import { createPagesFunctionHandler } from "@remix-run/cloudflare-pages";
+
+const build = async (): Promise<ServerBuild> => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  // eslint-disable-next-line import/no-unresolved
+  return import("../build/server");
+};
```
Comment (pcattori) [inline templates/unstable-vite-cloudflare/functions/[[path]].ts:9]: @markdalgleish I'm tempted to design a `getBuild` API that returns `Promise<ServerBuild | undefined>` to avoid this ugly boilerplate. Server bundles complicate things slightly, though we could just return `undefined` when server bundles are present.

templates/unstable-vite-cloudflare/vite.config.ts
```diff
@@ -0,0 +1,21 @@
...
+// Check out .cloudflare/README.md for why there's a `.cloudflare` directory in this template
+// In the future, you'll be able to import the `cloudflare` adapter from `@remix-run/cloudflare`
+import { adapter as cloudflare } from "./.cloudflare";
+
+export default defineConfig({
+  ssr: {
+    resolve: {
+      externalConditions: ["workerd", "worker"],
+    },
+  },
```
pcattori, line 14: TODO: this should be set automatically by the `cloudflare` adapter

templates/unstable-vite-cloudflare/functions/[[path]].ts
Comment (pcattori) [reply templates/unstable-vite-cloudflare/functions/[[path]].ts:9]: Writing the above snippet made me think that we should just include the `throw Error` part within `getBuild` and return `Promise<ServerBuild>` instead.

sergiodxa:
Since the bindings are set automatically, will context.env get correct types for the bindings? Like context.env.MY_KV bring a KVNamespace? Or we will need it still override AppLoadContext?

templates/unstable-vite-cloudflare/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "build": "remix vite:build",
+    "start": "wrangler pages dev ./build/client --kv MY_KV",
+    "deploy": "wrangler pages deploy ./build/client",
+    "lint": "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint .",
+    "typecheck": "tsc"
+  },
+  "dependencies": {
+    "@remix-run/cloudflare": "*",
+    "@remix-run/cloudflare-pages": "*",
+    "@remix-run/react": "*",
```
hi-ogawa, line 16: Currently Remix includes `@remix-run/node` in `optimizeDeps` but this template doesn't have direct dependency on it. This might cause Vite's warning for non-flat node_modules like pnpm where `@remix-run/node` is not visible from root project. 

I have a similar case where I use `@remix-run/server-runtime` directly without `@remix-run/node` and I've seen this warning on startup:

```
Failed to resolve dependency: @remix-run/node, present in 'optimizeDeps.include'
```

I'm not sure what's the best way to avoid this within Remix, but for now, what I did was to exclude manually by custom plugin `config` hook https://github.com/hi-ogawa/demo-remix-unstable-vite-cloudflare-workers/blob/2bdd6a4e15dc99584aa23897a5c1a40d7b185f6e/vite.config.ts#L15-L17

pcattori:
> Since the bindings are set automatically, will context.env get correct types for the bindings? Like context.env.MY_KV bring a KVNamespace? Or we will need it still override AppLoadContext?

The bindings are defined in a TOML file (`wrangler.toml`) so there's no way to get automatic types for it without type gen. Not something I'm inherently opposed to, but we don't have type gen as part of the scope for the "stabilizing Vite" work.

templates/unstable-vite-cloudflare/functions/[[path]].ts
Comment (pcattori) [reply templates/unstable-vite-cloudflare/functions/[[path]].ts:9]: That said, this is something we can land after Vite is stabilized.

cloudcreatr:
How do I access the D1 database? Am i supposed to make wrangler file or pass it through wrangler dev --d1 flag....

pcattori:
> How do I access the D1 database? Am i supposed to make wrangler file or pass it through wrangler dev --d1 flag....

The `wrangler dev --d1` flag only sets D1 when running `wrangler`, but with Vite, `wrangler` is never run. So if you want a consistent experience across Vite Dev Server, `wrangler dev`, and your deployment, its best to define it within `wrangler.toml`.

🚨 As of today, the latest stable version of `wrangler` (3.23.0) does not support `wrangler.toml` for CF pages. This feature is available on `wrangler@beta` or the next release of wrangler (presumably 3.24)

pcattori:
> @pcattori cf pages doesn't support toml file you said.. but the movie demo which Ryan launched uses toml file and it works fine and he applied migration that are visible in local db also.. but i think as you said.. toml doesn't work and the latest remix and cloudflare is kind off broken.. when i apply migration. The changes is not reflected in local db even after passing flags and creating a toml file as recommended by cf docs.. i wanted to know.. there's no server file in app dir.. how is streaming handled.. also.. i wanted to setup this locally with binding for my project just to see and work and play around.. even if it's beta.. how can I set up like . Thanks for your work

CF core team mentioned that CF pages reading from `wrangler.toml` before v3.24 was an accidental implementation detail and may not read _all_ values from `wrangler.toml` so something may have just worked, but not guaranteed. For how streaming is setup, checkout the [default `entry.server.tsx` for CF](https://github.com/remix-run/remix/blob/main/packages/remix-dev/config/defaults/entry.server.cloudflare.tsx).

@cloudcreatr good questions, but to keep this conversation focused on the changes introduced in this PR, please ask any follow-up questions about how CF + Remix generally works in our Discord. Thanks!

templates/unstable-vite-cloudflare/vite.config.ts
pcattori: from chatting w/ Mark: lets do `{ remixConfig, viteConfig }` as the return shape for adapter config

docs/future/vite.md
```diff
@@ -6,8 +6,6 @@ title: Vite (Unstable)
 
 [Vite][vite] is a powerful, performant and extensible development environment for JavaScript projects. In order to improve and extend Remix's bundling capabilities, we now support Vite as an alternative compiler. In the future, Vite will become the default compiler for Remix.
 
-<docs-warning>Note that Cloudflare is not yet supported when using Vite.</docs-warning>
-
```
markdalgleish, line 10: 😍

cloudcreatr:
Can I use it now 😀.. finally 🫡.. thanks @pcattori

templates/unstable-vite-cloudflare/.cloudflare.ts
pcattori: actually from `@remix-run/dev` since we don't want to make `@remix-run/cloudflare` a build-time dependency

docs/future/vite.md
```diff
@@ -80,6 +79,62 @@ A function for assigning addressable routes to [server bundles][server-bundles].
...
+
+There are two ways to run your Cloudflare app locally:
+
+```shellscript nonumber
+# Vite
+remix vite:dev
+
+# Wrangler
+remix vite:build # build app before running wrangler
+wranger pages dev ./build/client
```
NuroDev, line 114: Shouldn't this be 'wrangler'?

pcattori: It is using `wrangler` via `wrangler pages dev`.

reichhartd: Typo `wranger ` instead of `wrangler` 😄

pcattori: Oh now I see it 😅 yep you're right

tsteckenborn:
Am I doing something stupid, or is running it in dev via `"remix vite:dev",` currently always crashing completely if a route is not matching anything?

```
No routes matched location "..." 
/Users/.../Documents/div/.../node_modules/.pnpm/wrangler@3.24.0/node_modules/wrangler/wrangler-dist/cli.js:29572
throw a;
        ^
TypeError: Cannot read properties of undefined (reading 'get')
```

pcattori:
Can you provide a repro in stackblitz? My guess is that something is not playing nice with `pnpm`

pcattori:
> Am I doing something stupid, or is running it in dev via `"remix vite:dev",` currently always crashing / exiting completely if a route is not matching anything?
> 
> ```
> No routes matched location "..." 
> /Users/.../Documents/div/.../node_modules/.pnpm/wrangler@3.24.0/node_modules/wrangler/wrangler-dist/cli.js:29572
> throw a;
>         ^
> TypeError: Cannot read properties of undefined (reading 'get')
> ```

@tsteckenborn Can you provide a repro in stackblitz? My guess is that something is not playing nice with `pnpm`

patdx:
What's the recommended way to customize the `getLoadContext` in the vite Cloudflare mode?

For example I was attaching this cache adapter on each request: https://github.com/AdiRishi/cachified-adapter-cloudflare-kv

Should I create a custom server for dev mode and edit the getLoadContext in the catch-all route for prod mode? Or create an entry.server.ts file?

pcattori:
@patdx there's still a couple things to figure out for a general way `loadContext` since it affects _all_ possible servers, not just CF. What you're doing seems like an excellent workaround in the meantime.

tsteckenborn:
Unfortunately a bit complex environment to give a proper example (monorepo, pnpm), but did anybody else encounter such a behavior?

2024-02-02 at 19 23 21@2x

D1 and R2 Buckets are resolved properly, but KV isn't. It seems to exist an array named KV_Namespaces instead.

PR #8541, opened 2024-01-17
Add Panda CSS to styling guide

anubra266:
Add Panda CSS to styling guide

### DIFF
--- a/docs/styling/panda-css.md
+++ b/docs/styling/panda-css.md
@@ -1,0 +1,105 @@
+---
+title: Panda CSS
+---
+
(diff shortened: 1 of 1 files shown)


markdalgleish (commented):
Thanks for the PR!

We currently only have documentation for styling solutions that have Remix-specific integrations. Since Panda is a standalone tool that integrates with PostCSS, it doesn't need any additional integrations with the Remix compiler.

Also note that with our move towards Vite, most of our styling documentation will be removed in the future since consumers will be interacting with Vite directly.

PR #8542, opened 2024-01-18
refactor(remix-dev/vite): resolve absolute serverBuildDirectory

markdalgleish:
All paths on the resolved Remix config should be absolute but `serverBuildDirectory` was root-relative which made it awkward for any consuming code.

Since we're now resolving options from the Vite plugin, I've refactored the code to make it clearer that we're only calling `resolveConfig` for options that are compatible with the Remix esbuild compiler.

### DIFF
--- a/packages/remix-dev/vite/build.ts
+++ b/packages/remix-dev/vite/build.ts
@@ -139,18 +139,21 @@ async function getServerBuilds({
       }
       serverBundlesManifest.routeIdToServerBundleId[route.id] = bundleId;
 
-      let serverBundleDirectory = path.join(serverBuildDirectory, bundleId);
+      let relativeServerBundleDirectory = path.relative(
+        rootDirectory,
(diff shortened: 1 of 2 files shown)


packages/remix-dev/vite/build.ts
```diff
@@ -139,18 +139,21 @@ async function getServerBuilds({
...
+        rootDirectory,
+        path.join(serverBuildDirectory, bundleId)
+      );
       let serverBuildConfig = serverBuildConfigByBundleId.get(bundleId);
       if (!serverBuildConfig) {
         serverBundlesManifest.serverBundles[bundleId] = {
           id: bundleId,
           file: normalizePath(
-            path.join(serverBundleDirectory, serverBuildFile)
+            path.join(relativeServerBundleDirectory, serverBuildFile)
```
markdalgleish, line 151: This file is a case where we _want_ it to be relative since it's a root-relative manifest file.

PR #8553, opened 2024-01-18
docs: add how to fix cjs/esm bugs video link

brookslybrand:
Adding a link to the [video](https://www.youtube.com/watch?v=jmNuEEtwkD4) I made for fixing cjs/esm issues.

Feel free to suggest changes to the language or order in the docs

### DIFF
--- a/docs/future/vite.md
+++ b/docs/future/vite.md
@@ -723,11 +723,11 @@ check out our [discussion on Hot Module Replacement][hmr] to learn more about th
 
 #### ESM / CJS
 
-For a walkthrough of fixing an example bug checkout [🎥 How to Fix CJS/ESM Bugs in Remix][how-fix-cjs-esm].
-
 Vite supports both ESM and CJS dependencies, but sometimes you might still run into issues with ESM / CJS interop.
 Usually, this is because a dependency is not properly configured to support ESM.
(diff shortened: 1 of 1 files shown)


docs/future/vite.md
```diff
@@ -723,6 +723,8 @@ check out our [discussion on Hot Module Replacement][hmr] to learn more about th
 
 #### ESM / CJS
 
+For a walkthrough of fixing an example bug checkout [🎥 How to Fix CJS/ESM Bugs in Remix][how-fix-cjs-esm].
+
 Vite supports both ESM and CJS dependencies, but sometimes you might still run into issues with ESM / CJS interop.
 Usually, this is because a dependency is not properly configured to support ESM.
 And we don't blame them, its [really tricky to support both ESM and CJS properly][modernizing-packages-to-esm].
```
markdalgleish, line 730: I feel like we need the context of the issue before the link—what do you think?

```suggestion
Vite supports both ESM and CJS dependencies, but sometimes you might still run into issues with ESM / CJS interop.
Usually, this is because a dependency is not properly configured to support ESM.
And we don't blame them, its [really tricky to support both ESM and CJS properly][modernizing-packages-to-esm].

For a walkthrough of fixing an example bug, check out [🎥 How to Fix CJS/ESM Bugs in Remix][how-fix-cjs-esm].
```

PR #8557, opened 2024-01-19
refactor(vite): decouple server bundle config from serverBuildDirectory

markdalgleish:
When exploring replacing the separate `serverBuildDirectory` and `assetsBuildDirectory` options with a single `buildDirectory` option, I realised that a blocker for this would be the fact that the mechanism for building server bundles is coupled to the `serverBuildDirectory` option.

To address this, this PR decouples the two so that now, the Remix CLI only injects the `serverBundleId`, not its resolved output directory. Once this lands, it'll be trivial to change the shape of our config without having to touch the server bundle build process.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -194,9 +194,9 @@ export type ResolvedVitePluginConfig = Pick<
   ssr: boolean;
 };
 
-export type ServerBuildConfig = {
+export type ServerBundleBuildConfig = {
   routes: RouteManifest;
-  serverBuildDirectory: string;
+  serverBundleId: string;
 };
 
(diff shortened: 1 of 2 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -194,9 +194,9 @@ export type ResolvedVitePluginConfig = Pick<
   ssr: boolean;
 };
 
-export type ServerBuildConfig = {
+export type ServerBundleBuildConfig = {
   routes: RouteManifest;
-  serverBuildDirectory: string;
+  serverBundleId: string;
```
markdalgleish, line 199: This is the core change. The rest of the PR is in service of this.

PR #8564, opened 2024-01-19
template(vite): Improve vite-express template to match official express template server behavior

ngbrown:
- Set eslint node environment for server.mjs
- Use compression and disable x-powered-by in express
- Use morgan for logging requests
- Make port configurable by environment variable

These changes make the express server behavior more like the normal express from the official template: https://github.com/remix-run/remix/tree/main/templates/express 

Testing Strategy:

I made these changes after running `create-remix` with the `remix-run/remix/templates/unstable-vite-express` template. I tested both `dev` and `buld`/`start` modes.

After updating the source, I then ran the following:

```ps1
npx create-remix@latest --template /dev/OSS/remix/templates/unstable-vite-express
```

And again the install worked along with the `dev`, `build`, and `start` commands.

### DIFF
--- a/templates/unstable-vite-express/.eslintrc.cjs
+++ b/templates/unstable-vite-express/.eslintrc.cjs
@@ -74,7 +74,7 @@ module.exports = {
 
     // Node
     {
-      files: [".eslintrc.js"],
+      files: [".eslintrc.js", "server.mjs"],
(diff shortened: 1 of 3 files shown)


markdalgleish (approved):
Thank you!

PR #8565, opened 2024-01-20
Guide readers to fixing TS errors to avoid getting stuck and frustrated

radulescuandrew:
Guide readers to not get stuck in fixing typescript errors while copy/pasting code

Closes: #

- [X] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -484,6 +484,7 @@
 - ptitFicus
 - pwbriggs
 - pyr0gan
+- radulescuandrew
 - ramiroazar
 - RATIU5
 - raulrpearson


brookslybrand:
Hey @radulescuandrew, thanks for opening up this PR! I think it would be better if the callouts are closer to the code that introduces type errors, that way people will be less likely to miss them.

I've opened up #8579 with this and another little fix. Please let me know if you notice any spots I missed

radulescuandrew:
@brookslybrand It's even better what you've done. Cheers!

PR #8568, opened 2024-01-21
"node" instead of "deno"

gomflo:


sergiodxa:
The usage of Deno there is correct, why do you propose change it to Node?

PR #8582, opened 2024-01-23
refactor(remix-dev/vite): move server bundle ID to build context

markdalgleish:
This refactor is in preparation to expose the resolved Remix config as public API (e.g. via the adapter `buildEnd` hook). One thing holding this back was that the private `serverBundleId` property was part of the resolved config type even though it'll never be present from a consumer perspective.

As part of this refactor I realised `serverBundleId` should have been on the `ssrBuildContext` object all along. To raise its prominence, I've renamed this to just `buildContext`.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -197,7 +197,6 @@ export type ResolvedVitePluginConfig = Pick<
   buildDirectory: string;
   manifest: boolean;
   serverBuildFile: string;
-  serverBundleId?: string;
   serverBundles?: ServerBundlesFunction;
   ssr: boolean;
 };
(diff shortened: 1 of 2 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -197,7 +197,6 @@ export type ResolvedVitePluginConfig = Pick<
   buildDirectory: string;
   manifest: boolean;
   serverBuildFile: string;
-  serverBundleId?: string;
```
markdalgleish, line 200: This is the core change in this PR. Everything else is in service of this change.

packages/remix-dev/vite/plugin.ts
```diff
@@ -816,16 +844,6 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
...
-        ssrBuildContext =
-          viteConfig.build.ssr && viteCommand === "build"
-            ? {
-                isSsrBuild: true,
-                getBrowserManifest: createBrowserManifestForBuild,
-              }
-            : {
-                isSsrBuild: false,
-              };
-
```
markdalgleish, line 828: This logic moved earlier in the plugin lifecycle to the `config` hook, making use of `viteConfigEnv` rather than the using the resolved Vite config. This is safe because config resolution won't result in a different value for `ssr`.

PR #8586, opened 2024-01-24
refactor(remix-dev/vite): clean resolved Remix config

markdalgleish:
This is a follow-up to https://github.com/remix-run/remix/pull/8582, in preparation to expose the resolved Remix config as public API (e.g. via the adapter buildEnd hook).

The only remaining properties on the resolved `remixConfig` object that aren't present in the user-facing config type are `rootDirectory`, `entryClientFilePath` and `entryServerFilePath`. These have now been moved to the `buildContext` object. Since a lot of code assumes these are on `remixConfig`, it required a bit of a refactor to accomodate this.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -199,7 +199,7 @@ export type ServerBundleBuildConfig = {
   serverBundleId: string;
 };
 
-type ServerBuildContext =
+type RemixPluginServerContext =
   | {
       isSsrBuild: false;
       getBrowserManifest?: never;
@@ -211,7 +211,7 @@ type ServerBuildContext =
       serverBundleId: string | undefined;
     };
 
-export type BuildContext = ServerBuildContext & {
+export type RemixPluginContext = RemixPluginServerContext & {
   rootDirectory: string;
   entryClientFilePath: string;
   entryServerFilePath: string;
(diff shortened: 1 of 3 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -416,19 +419,17 @@ const getServerBundleBuildConfig = (
...
+    ...(typeof buildContext.serverBundleId === "string"
+      ? [buildContext.serverBundleId]
+      : [])
   );
 
-export let getServerBuildRootDirectory = (
-  remixConfig: ResolvedVitePluginConfig
-) => getServerBuildDirectory(remixConfig, { serverBundleId: undefined });
+export let getServerBuildRootDirectory = (buildContext: BuildContext) =>
+  getServerBuildDirectory(buildContext);
```
pcattori, line 432:
```suggestion
export let getServerBuildRootDirectory = getServerBuildDirectory;
```

Though at that point, might be worth just exporting `getServerBuildDirectory` directly

packages/remix-dev/vite/plugin.ts
```diff
@@ -218,6 +211,13 @@ type BuildContext =
       serverBundleId: string | undefined;
```
pcattori, line 211: Is there a reason this is `string | undefined` instead of `?: string`?

markdalgleish: So that you're forced to add the key to the object, even if the value is undefined.

packages/remix-dev/vite/plugin.ts
markdalgleish: Nice catch, I missed this!

Issue #8589, opened 2024-01-24
Vite build creates build/client/.vite/manifest.json, which gets exposed as static asset


nramovs-sr:
### Reproduction

https://remix.run/.vite/manifest.json

### System Info

```shell
System:
    OS: macOS 14.3
    CPU: (10) arm64 Apple M1 Max
    Memory: 156.59 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 3.7.0 - ~/.nvm/versions/node/v20.10.0/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    bun: 1.0.3 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 120.1.61.120
    Chrome: 121.0.6167.85
    Chrome Canary: 122.0.6261.3
    Safari: 17.3
  npmPackages:
    @remix-run/dev: ^2.5.1 => 2.5.1 
    @remix-run/eslint-config: ^2.5.1 => 2.5.1 
    @remix-run/express: ^2.5.1 => 2.5.1 
    @remix-run/node: ^2.5.1 => 2.5.1 
    @remix-run/react: ^2.5.1 => 2.5.1 
    @remix-run/testing: ^2.5.1 => 2.5.1 
    vite: ^5.0.12 => 5.0.12
```

### Used Package Manager

yarn

### Expected Behavior

Manifest should not be exposed in the client's public folder.

### Actual Behavior

build/client/.vite/manifest.json gets created, and is exposed to the internet along with other client assets through `app.use(express.static("build/client", { maxAge: "1h" }));` (from vite express template). This file contains information about source file structure.

pcattori:
Fixed by #8599

PR #8599, opened 2024-01-25
feat(remix-dev/vite): modify manifest paths

markdalgleish:
Fixes https://github.com/remix-run/remix/issues/8589

Vite manifest files are now moved from the `client` and `server` output directories to the new top-level `build/.vite` directory. Since we're managing these files on disk now, I've also updated the logic so that the Vite manifests are deleted if `build.manifest` isn't set to `true` in their Vite config.

For consistency, the Remix build manifest path has been changed to `build/.remix/manifest.json`.


pcattori (approved):

packages/remix-dev/vite/build.ts
```diff
@@ -270,23 +303,42 @@ export async function build(
     });
   }
 
-  // Since we're potentially running multiple Vite server builds with different
-  // output directories, we need to clean the root server build directory
-  // ourselves rather than relying on Vite to do it, otherwise you can end up
-  // with stale server bundle directories in your build output
-  await cleanServerBuildDirectory(viteConfig, ctx);
+  await cleanBuildDirectory(viteConfig, ctx);
```
markdalgleish, line 306: Now that manifest files are written to the root of the build directory, the comment that was written here previously now applies more generally to any Remix build since it may result in a different set of manifest files.

PR #8604, opened 2024-01-26
WIP: DevScripts

pcattori:
DO NOT MERGE

just a hacky prototype for a `<DevScripts />` component designed to be included in the `<head />` so that its guaranteed to execute before normal `<Scripts />`. Better name than `<LiveReload />` since we do HMR + HDR too.


pcattori:
Obsoleted by ##8609

PR #8609, opened 2024-01-26
Vite: replace `<LiveReload/>` with `<DevScripts/>`

pcattori:
**This is a breaking change for projects using the unstable Vite plugin.**

The `<LiveReload/>` component has a confusing name as it now also supports HMR and HDR.
Additionally, it provides an bespoke client-side runtime that is obsoleted by Vite.
To get our Vite plugin working, we were doing some compiler magic to swap out the
implementation of `<LiveReload/>`.
This was always meant as a temporary measure.

Now we have a better solution in the form of a new `<DevScripts/>` component specifically
designed with Vite's HMR capabilities in mind.

The `<LiveReload />` component will cease to provide HMR and HDR capabilities in Vite,
so you'll need to replace `<LiveReload/>` with `<DevScripts/>` in your app.

The `<DevScripts/>` component should be placed in the `<head/>` of your app so that it
can be loaded before any other scripts as required by React Fast Refresh.

```diff
  import {
-   LiveReload,
+   DevScripts,
    Outlet,
  }

  export default function App() {
    return (
      <html>
        <head>
+         <DevScripts />
        </head>
        <body>
-         <LiveReload />
          <Outlet />
        </body>
      </html>
    )
  }
```

---

Tried to directly inject the HMR runtime into the `<head/>` instead (https://github.com/remix-run/remix/pull/8604) but that still had issues with race conditions.


packages/remix-react/components.tsx
```diff
@@ -1060,6 +1060,18 @@ export function useFetcher<TData = AppData>(
   return useFetcherRR(opts);
 }
 
+export const DevScripts =
+  process.env.NODE_ENV === "production"
```
houmark, line 1064: @pcattori Wouldn't it be better to check for `process.env.NODE_ENV === "development"` and anything else than that should return null?

Maybe someone runs a NODE_ENV like `staging` or `preview` but in those cases, you wouldn't have HMR active.

templates/unstable-vite/app/root.tsx
```diff
@@ -15,12 +15,12 @@ export default function App() {
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <Meta />
         <Links />
+        <DevScripts />
       </head>
       <body>
         <Outlet />
         <ScrollRestoration />
         <Scripts />
-        <LiveReload />
```
brophdawg11, line 23: We probably need this change in `templates/spa/root.tsx` as well which is based on vite

pcattori:
Alternative would be to have something like:

```ts
```diff
  import {
-   LiveReload,
    Scripts,
    Outlet,
  }

  export default function App() {
    return (
      <html>
        <body>
-         <LiveReload />
           <Scripts />
          <Outlet />
        </body>
      </html>
    )
  }
```

Where `<Scripts />` automatically includes dev scripts. For users who want to ship without JS, they can do `<Scripts devOnly />` to still get the benefits of HMR/HDR in dev.

pcattori:
Superceded by #8636

PR #8619, opened 2024-01-29
templates(cloudflare-pages): add cf (IncomingRequestCfProperties) to context

smcstewart:
Testing Strategy:

Opened a terminal window and ran:
1. Ran `create-remix` from the updated template locally:

```
pnpm dlx create-remix@2.5.1 my-remix-app --template file:///src/remix/templates/cloudflare-pages
cd my-remix-app
pnpm run dev
```

2. Verified that the Remix dev server started up.
3. Opened another terminal window, verifying `GET / 200 OK` appeared in the logs upon the following:

```
http http://localhost:8788
```

4. Then added a loader to the `/app/routes/_index.tsx`:

```
export const loader = async ({ context }) => {
  console.log(JSON.stringify(context, null, 2));
  return null;
}
```

5. And run the dev server again in one window, and the `http` command in another:

```
pnpm run dev
```

```
http http://localhost:8788
```

6. Verified that the `context.cf` object appeared in the logs.
7. Finally, deployed to Cloudflare Pages, and verified the same.

### DIFF
--- a/templates/cloudflare-pages/server.ts
+++ b/templates/cloudflare-pages/server.ts
@@ -8,6 +8,6 @@ if (process.env.NODE_ENV === "development") {
 
 export const onRequest = createPagesFunctionHandler({
   build,
-  getLoadContext: (context) => ({ env: context.env }),
+  getLoadContext: (context) => ({ env: context.env, cf: context.request?.cf }),
   mode: build.mode,
 });
--- a/contributors.yml
+++ b/contributors.yml
@@ -555,6 +555,7 @@
 - sjparsons
 - skube
 - smathson
+- smcstewart
 - smeijer
 - sndrem
 - sobrinho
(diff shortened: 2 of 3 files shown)


arjunyel:
@smcstewart thank you friend! Can you update to make the changes to the new Cloudflare + Vite integration?

templates/cloudflare-pages/server.ts
```diff
@@ -8,6 +8,6 @@ if (process.env.NODE_ENV === "development") {
 
 export const onRequest = createPagesFunctionHandler({
   build,
-  getLoadContext: (context) => ({ env: context.env }),
+  getLoadContext: (context) => ({ env: context.env, cf: context.request?.cf }),
```
sergiodxa, line 11: Is waitUntil already provided to `context` here? It would be great to also pass that function by default

smcstewart: I don't think so, but will take a look. And I'll look at the Vite template as well.

PR #8620, opened 2024-01-29
Convert Vite plugin `adapter` API to `presets`

markdalgleish:
This PR makes the following changes: 
- Expands the `adapter` API into a more general `presets` API, supporting an array of preset objects that can define config to be merged with defaults before the user config is applied at the end.
- Removes the `loadContext` and `viteConfig` options from the old `adapter` API.
- Converts the Cloudflare adapter into a preset.
- Adds a `remixConfigResolved` hook to the preset which provides support for config validation after all presets have been applied.
- Adds a required `name` property to presets so that we can roll out improved error messaging support to the new `remixConfigResolved` hook in the future.
- Makes the `buildEnd` option available as a top-level option, not just for adapters.

Note that the documentation for preset authors is pretty light right now while this feature is still fresh. I'll add more docs as a follow-up, but the priority is to get this out for now.


packages/remix-dev/vite/plugin.ts
```diff
@@ -450,6 +537,25 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
...
+    // with an empty object. Conceptually this means that presets are operating
+    // on the user config object as it would be defined in your Vite config.
+    let accumulatedRemixConfigPresets: VitePluginConfig = {};
+
+    for (let preset of remixUserConfig.presets ?? []) {
+      let { remixConfig: getConfigPreset } = preset;
+      let remixConfigPreset =
+        getConfigPreset &&
+        omit(
+          await getConfigPreset({ remixConfig: accumulatedRemixConfigPresets }),
```
markdalgleish, line 551: I'm thinking I might pull this `remixConfig` arg from the preset since it's of limited value and arguably pretty confusing. Because it's the accumulator, it only has access to config values defined in earlier presets. The original intention of this arg was to support config validation, but this should now be handled in the `remixConfigResolved` hook since that contains the final config.

packages/remix-dev/vite/plugin.ts
```diff
@@ -537,7 +629,11 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
       serverBundles,
       serverModuleFormat,
       unstable_ssr,
-    };
+    });
+
+    for (let preset of remixUserConfig.presets ?? []) {
+      await preset.remixConfigResolved?.({ remixConfig });
```
markdalgleish, line 635: I'm thinking the args might need to be fleshed out a bit more here. If this hook is designed for config validation, you don't have enough knowledge to provide context for what caused the error, e.g. did the invalid value come from user config, or did it come from another preset, and if so, which one? To aid that last point, we probably also want to ensure that all presets have a `name` property defined.

markdalgleish: Note, it doesn't have to block this PR, except for adding a required `name` property to presets.

packages/remix-dev/vite/plugin.ts
markdalgleish: Updated.

packages/remix-dev/vite/plugin.ts
markdalgleish: The `name` property has been added as a required preset property since we want to ensure all presets have this from day one. We can always add more args to the `remixConfigResolved` hook without it being a breaking change so this is good enough for now.

packages/remix-dev/vite/plugin.ts
```diff
@@ -433,6 +431,98 @@ export let getServerBuildDirectory = (ctx: RemixPluginContext) =>
...
+                ...(configB.ignoredRouteFiles ?? []),
+              ])
+            ),
+          }
+        : {}),
+      ...(mergeRequired("presets")
+        ? {
+            presets: [...(configA.presets ?? []), ...(configB.presets ?? [])],
+          }
+        : {}),
```
pcattori, line 477: I don't think we want to support merging `presets`. Presets should all be declared upfront. That could also simplify `updateRemixPluginContext` where `omit` is called for `presets` key.

markdalgleish: This code is currently not used since we never pass two configs with presets through this function. I only included it here for completeness in case this function ever ends up being public API.

Issue #8623, opened 2024-01-29
vite/spa: `npm run start` using http-server does not support SPA routes


ngbrown:
### Reproduction

Run this script (PowerShell):

```ps1
npx create-remix@latest remix-spa-test -y --template remix-run/remix/templates/spa
cd .\remix-spa-test
echo "export default function Hello() { return <>Hello</>; }" | Out-File ./app/routes/hello.tsx -Encoding utf8
echo "import { Link } from `"@remix-run/react`";`n`nexport default function Index() { return <Link to=`"/hello`">Navigate</Link>; }" | Out-File ./app/routes/_index.tsx  -Encoding utf8
npm run build
npm run start
```

Open browser directly to http://127.0.0.1:8080/hello (change port if needed). Also try reloading the browser (`F5`) after navigating to the `/hello` route from the root (`/`) path.

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (24) x64 12th Gen Intel(R) Core(TM) i7-12850HX
    Memory: 31.75 GB / 63.67 GB
  Binaries:
    Node: 18.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.6.7 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (121.0.2277.83)
    Internet Explorer: 11.0.19041.3636
  npmPackages:
    @remix-run/dev: ^2.5.1 => 2.5.1
    @remix-run/node: ^2.5.1 => 2.5.1
    @remix-run/react: ^2.5.1 => 2.5.1
    vite: ^5.0.12 => 5.0.12
```

### Used Package Manager

npm

### Expected Behavior

When previewing a build, the server should support SPA fallback mode (serving multiple paths from a single `index.html` file) when using the SPA template. Use `vite preview` or `sirv-cli --single` as alternatives. (`vite preview` is powered by `sirv`).

No simple solutions that I found are especially production ready as they don't set the `cache-control` header for `./assets` files separately from other files.

### Actual Behavior

Navigating directly to a sub-route such as http://127.0.0.1:8080/hello fails with an HTTP ERROR 404.

Navigating to the page when starting at http://127.0.0.1:8080/ works correctly, but reloading once navigated to http://127.0.0.1:8080/hello also fails with an HTTP ERROR 404.

[...]

PR #8624, opened 2024-01-29
vite: Enable use of `vite preview` to preview built SPA sites

ngbrown:
This enables use of `vite preview` for serving built SPA sites. This works better than `http-server` because it correctly supports the SPA fallback (i.e. serving a statically deployed site with all routes being served from a root `/index.html`).

This is more than a template change because the Remix Vite plugin was setting `appType: "custom"`, preventing `vite preview` from working in `"spa"` mode. The patch also sets the `build.outDir` when previewing so additional command line options aren't needed.

The SPA template has been updated to use `npm run preview` (or `yarn preview`) to preview the built site. The `start` script (`npm run start`) has been removed from the SPA template because it isn't something that should be used in production.

Closes: #8623

- [x] Docs (template/spa/README.md)

Testing Strategy:

I opened up my windows machine and ran this script (PowerShell):

```ps1
Remove-Item -Recurse .\scripts\playground\template.local\
Copy-Item -Recurse .\templates\spa .\scripts\playground\template.local
yarn playground:new
cd .\playground\playground-1706546017969 # use directory name produced by previous command
echo "export default function Hello() { return <>Hello</>; }" | Out-File ./app/routes/hello.tsx -Encoding utf8
echo "import { Link } from `"@remix-run/react`";`n`nexport default function Index() { return <Link to=`"/hello`">Navigate</Link>; }" | Out-File ./app/routes/_index.tsx  -Encoding utf8
npm run dev
npm run build
npm run preview
```

Open browser to link provided by `vite preview` and click on the "Navigate" link to arrive at `http://localhost:4173/hello` (port may differ). Refresh the page (e.g. `F5`), and the page should load without a 404. Also try navigating directly to that URL in a new tab.

### DIFF
--- a/.changeset/tender-pens-tan.md
+++ b/.changeset/tender-pens-tan.md
@@ -1,0 +1,8 @@
+---
+"@remix-run/dev": patch
+---
+
+vite: Enable use of `vite preview` to preview built SPA sites
+
+In the SPA template, `start` script (`npm run start`) has been renamed to `preview` (`npm run preview`).
+In the SPA template, `npm run preview` uses `vite preview` for serving built SPA sites.
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -897,7 +897,12 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
 
         return {
           __remixPluginContext: ctx,
-          appType: "custom",
+          appType:
+            viteCommand === "serve" &&
+            viteConfigEnv.mode === "production" &&
+            ctx.remixConfig.unstable_ssr === false
+              ? "spa"
+              : "custom",
           optimizeDeps: {
             include: [
               // Pre-bundle React dependencies to avoid React duplicates,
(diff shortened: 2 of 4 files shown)


ngbrown:
I also implemented the required Remix changes as a patch against `@remix-run/dev@2.5.1`.

[@remix-run+dev+2.5.1.patch](https://github.com/remix-run/remix/files/14088013/%40remix-run%2Bdev%2B2.5.1.patch)

brophdawg11:
> This works better than http-server because it correctly supports the SPA fallback.

What is the "SPA Fallback" you're referring to here?  Remix SPA mode renders the root `HydrateFallback` on initial load while client JS loaders and `clientLoaders` execute - is "SPA Fallback" referring to another form of fallback UI?

> has been removed from the SPA template because it isn't something that should be used in production

I kind of liked `start` + `http-server` _because_ it was something you could deploy in production.  But I'm not too sure how many folks are serving their SPA's that way in prod versus a static CDN host, or a custom HTTP server of their own - so maybe it's fine for the template not to have an opinion there.

brophdawg11:
oh, I see the linked issue now - nvm my question about the fallback, we can discuss in #8623

ngbrown:
> What is the "SPA Fallback" you're referring to here?

I was using what seemed to be the most common parlance used in the web server discussions, but it's not super obvious. I meant the option to serve up the root-directory `index.html` when a path `404`'s, no matter the route. It's something that became common when react and react-router were serving multiple URL paths from a single `index.html` file.

ngbrown:
@brophdawg11 I see that you are about to stabilize SPA mode. I just rebased this on the most recent `dev` branch. I also tested it again.

Let me know if you have any suggestions for what you would like to see changed.

brophdawg11 (approved):
Thanks!

PR #8646, opened 2024-01-31
copy types when using LOCAL_BUILD_DIRECTORY

pcattori:
## TODO

- [ ] set `chmod` perms for `node_modules/.bin/remix`

### DIFF
--- a/rollup.utils.js
+++ b/rollup.utils.js
@@ -7,23 +7,6 @@ const path = require("node:path");
 
 const REPO_ROOT_DIR = __dirname;
 
-let activeOutputDir = "build";
-if (process.env.LOCAL_BUILD_DIRECTORY) {
-  let appDir = path.resolve(process.env.LOCAL_BUILD_DIRECTORY);
-  try {
-    fse.readdirSync(path.join(appDir, "node_modules"));
-  } catch {
-    console.error(
-      "Oops! You pointed LOCAL_BUILD_DIRECTORY to a directory that " +
-        "does not have a node_modules/ folder. Please `npm install` in that " +
-        "directory and try again."
-    );
-    process.exit(1);
-  }
-  console.log("Writing rollup output to", appDir);
-  activeOutputDir = appDir;
-}
-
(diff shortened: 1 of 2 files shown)


rollup.utils.js
```diff
@@ -7,23 +7,6 @@ const path = require("node:path");
...
-    console.error(
-      "Oops! You pointed LOCAL_BUILD_DIRECTORY to a directory that " +
-        "does not have a node_modules/ folder. Please `npm install` in that " +
-        "directory and try again."
-    );
-    process.exit(1);
-  }
-  console.log("Writing rollup output to", appDir);
-  activeOutputDir = appDir;
-}
```
brophdawg11, line 25: I think we lose the `yarn watch` functionality if we do this since it calls rollup directly.  There's a `watch.onEnd` script we could add the copy to though if we go this route

scripts/build.mjs
```diff
@@ -6,6 +8,24 @@ const tsc = process.env.CI || args.includes("--tsc") || publish;
...
+          "Oops! You pointed LOCAL_BUILD_DIRECTORY to a directory that " +
+            "does not have a node_modules/ folder. Please `npm install` in that " +
+            "directory and try again."
+        );
+        process.exit(1);
+      }
+      fse.copySync("build", appDir, { overwrite: true });
+      fse.chmodSync(path.join(appDir, "node_modules/.bin/remix"), 0o755);
+    }
+  })
```
brophdawg11, line 28: I think we lose the `triggerLiveReload` behavior if we just `cp -r` here?  I believe we need to update one of the app files to trigger it, I don't think copying new files into `node_modules` as enough IIRC?

Should we be thinking about the `vite` cache here?  Or should we just be using `--force` in our apps?

brophdawg11:
Maybe a better approach here to preserve `yarn watch` and live reload behavior would be to just do this full copy in a rollup `writeBundle` hook?  That way it would preserve both of those existing behaviors?

PR #8649, opened 2024-01-31
vite: cloudflare preset load context

pcattori:
Closes: #

- [x] Docs
- [x] Tests

Testing Strategy:

### DIFF
--- a/.changeset/spicy-buses-play.md
+++ b/.changeset/spicy-buses-play.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Vite: add dev load context option to Cloudflare preset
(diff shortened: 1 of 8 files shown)


markdalgleish (approved):

kiliman:
@pcattori Is `setRemixDevLoadContext` considered a _public_ API?

I'm using a custom Express server, so it would be nice if I could set the load context in the Vite dev server as well.

pcattori:
@kiliman the idea is that for Node-based custom servers, you can use the _same_ server for both dev and prod where you can choose to delegate asset handling and route handling to Vite in dev. In that world, you already have direct access to `createRequestHandler` where you can do whatever you'd like with `loadContext`. And the best part is that your load context is automatically consistent in dev and prod.

The issue with CF is that (as of today) Vite can't delegate server code execution to other runtimes (like `workerd`) so you _can't_ run the same server for dev and prod (yet). So `setRemixDevLoadContext` is a temporary stopgap for CF to approach parity with Node-based custom servers.

kiliman:
True. Thanks!

Issue #8472, opened 2024-01-10
Unnecessary package `eslint-config-prettier` (or template should also come with Prettier and configure it's usage)


fernandocanizo:
### Reproduction

Just do an `pnpx create-remix` and review `package.json`, it will have `eslint-config-prettier`. It also affects the official templates: express, haven't checked the rest.

### System Info

```shell
Irrelevant, this is a report about unused extra dev-dependencies. Also the package manager used is irrelevant for this bug.
```

### Used Package Manager

pnpm

### Expected Behavior

To not have spurious packages installed by default.

### Actual Behavior

Package [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) gets included in `package.json`, this package is for turning off Eslint style rules that may conflict with Prettier, but the template doesn't include Prettier, also to be really put into use, it has to be configured on Eslint configuration file in order to turn off style rules on Eslint.

So the current delivered way doesn't make sense.

I believe is a leftover artifact from PR #7597 , which on its first commit correctly adds configuration for this Eslint plugin by adding it to the `extends` field:
```js
// ...
extends: [
  // ...
  "prettier",
],
```

But then, on commit [ca6b95f](https://github.com/remix-run/remix/pull/7597/commits/ca6b95fb38c28c9ad78fb0431d7f963445f06aad), when adapting Eslint configuration for the remaining official templates, it completely forgets about it.

I was planning on starting a discussion to add a pre-configured development environment with basic industry-standard defaults setup from the start:

- `eslint` only for errors
- `prettier` for style
- `husky` to add git hooks
- `lint-staged` to only apply hooks to what's gonna be commited

But it seems, according to #7056 , that that's more pertaining to templates (aka stacks).

So in **conclusion**, package `eslint-config-prettier` shouldn't be part of the default installation.

brophdawg11:
Yeah this looks leftover - we'd be happy to accept a PR to remove this from the templates/stacks!

PR #8671, opened 2024-02-02
Drops eslint-config-prettier on templates (#8472)

fernandocanizo:
Removes unused package `eslint-config-prettier` from all templates that were installing it.

Closes: #8472 

- [ ] Docs
- [ ] Tests

Testing Strategy:

This PR removes an **unused dependency**. I didn't run any tests. Only checked all `.eslintrc.cjs` from templates to ensure none had "prettier" configured on the `extends` field, and also verified that no template had Prettier as dependency.


contributors.yml
```diff
@@ -21,8 +21,8 @@
 - Alarid
 - alcpereira
 - AlemTuzlak
-- alex-ketch
 - alexanderson1993
+- alex-ketch
```
brophdawg11, line 25: Would you mind removing these sorting changes unrelated to adding your name?  Our CI handles automatically sorting this file 👍

PR #8676, opened 2024-02-03
docs: fix tutorial contact action highlight line numbers

FAL-coffee:
From what I can read from the preceding and following text, I felt that the 'formData.get' part should be highlighted here.

スクリーンショット 2024-02-04 1 36 37


brophdawg11 (approved):
Thanks!

Issue #7786, opened 2023-10-27
`?url` CSS imports used in `links` are not bundled correctly in production


pcattori:
### What version of Remix are you using?

v0.0.0-nightly-0c99349-20231027

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

```tsx
// app/root.tsx
import tailwindHref from "~/tailwind.css?url";

export const links = () => [
  { rel: "stylesheet", href: tailwindHref },
];
```

### Expected Behavior

Processed styles show up in production

### Actual Behavior

Processed styles only show up in development

### Related Vite issues

- https://github.com/vitejs/vite/issues/2522
- https://github.com/vitejs/vite/issues/7842
- https://github.com/vitejs/vite/issues/13416

pcattori:
Workaround: use side-effect `import`s instead of deprecated `?url` imports:

```diff
// app/root.tsx
- import { cssBundleHref } from "@remix-run/css-bundle";

- import tailwindHref from "~/tailwind.css?url";
+ import "~/tailwind.css";

- export const links = () => [
-   ...(cssBundleHref ? [{ rel: "stylesheet", href: cssBundleHref }] : []),
-   { rel: "stylesheet", href: tailwindHref },
- ];
```

markdalgleish:
I think this is the same issue, logged against Vite: https://github.com/vitejs/vite/issues/13416

684efs3:
If I use the side-effect method (+ import "~/tailwind.css";), when I save in my editor, VS Code, all of the CSS will be removed from the localhost site. 

![Screenshot 2023-10-30 at 12 25 44 PM](https://github.com/remix-run/remix/assets/48864759/b75b28a1-645d-4526-9e1e-24238e7577ef)

If I hit [Command] [R] to refresh the browser, then the CSS styles will be applied. 

![Screenshot 2023-10-30 at 12 26 12 PM](https://github.com/remix-run/remix/assets/48864759/fae1b820-bece-4bdc-bccd-f1133475643d)

markdalgleish:
@684efs3 Could you please share a repro?

c43721:
I'm experiencing the same issue as @684efs3 but I don't have a reproduction. I'll edit this if I can get it repro'd on a new vite project with tailwind.

pcfreak30:
> Workaround: use side-effect `import`s instead of `?url` imports:
> 
> ```diff
> // app/root.tsx
[...]

PR #8684, opened 2024-02-05
Support Vite v5.1.0's `.css?url` imports

markdalgleish:
Fixes https://github.com/remix-run/remix/issues/7786

This adds built in support for `.css?url` imports in Vite v5.1.0 (currently in beta) to ease migration of [Remix's regular CSS imports.](https://remix.run/docs/en/main/styling/css)

```diff
-import styles from "~/styles/dashboard.css";
+import styles from "~/styles/dashboard.css?url";
```

The only thing we need to do to support this out of the box is to enable `build.cssMinify` in server builds by default. If we don't do this, content hashes differ between server and client since the server doesn't have minification enabled by default.

I've opted to call out the beta status of this feature rather than waiting for Vite 5.1.0 so that it doesn't need to be a blocker for a stable release of the Remix Vite plugin.

We had an existing test for Vite v5.0.x's `.css?url` imports but it didn't use any CSS processing. In order to exercise the new `.css?url` support in Vite v5.1.0, I've updated our dev dependency to `v5.1.0-beta.6` and modified the test case to use a minimal PostCSS plugin (copied from the Remix compiler's PostCSS tests) so that CSS processing is required for the test to pass. I've confirmed that the updated test fails when using Vite v5.0.x.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -909,6 +909,7 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
           ...(viteCommand === "build" && {
             base: ctx.remixConfig.publicPath,
             build: {
+              cssMinify: viteUserConfig.build?.cssMinify ?? true,
               ...(!viteConfigEnv.isSsrBuild
                 ? {
                     manifest: true,
--- a/.changeset/shaggy-sheep-cheer.md
+++ b/.changeset/shaggy-sheep-cheer.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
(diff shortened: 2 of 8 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -909,6 +909,7 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
           ...(viteCommand === "build" && {
             base: ctx.remixConfig.publicPath,
             build: {
+              cssMinify: viteUserConfig.build?.cssMinify ?? true,
```
markdalgleish, line 912: This is the meaningful change. The rest of this PR is in service of this.

PR #8688, opened 2024-02-05
vite: externalize getbindingsproxy from cloudflare preset

pcattori:
## TODO
- [x] docs
- [x] changeset


templates/unstable-vite-cloudflare/load-context.ts
```diff
@@ -0,0 +1,34 @@
...
+type Bindings = {
+  // Add types for bindings configured in `wrangler.toml`
+  MY_KV: KVNamespace;
+};
+
+declare module "@remix-run/cloudflare" {
+  interface AppLoadContext {
+    env: Bindings;
+
+    extra: string; // Example of augmenting load context
```
markdalgleish, line 15: I think this line and the implementation should be commented out so that the load context doesn't have extra stuff (heh) on it.

Alternatively, do you think this requirement is rare enough that it could just be documented rather than having it in the template? I'm happy either way, I just want to make sure we're not clogging up the template with an edge case.

templates/unstable-vite-cloudflare/functions/[[path]].ts
```diff
@@ -4,8 +4,9 @@ import { createPagesFunctionHandler } from "@remix-run/cloudflare-pages";
 // @ts-ignore - the server build file is generated by `remix vite:build`
 // eslint-disable-next-line import/no-unresolved
 import * as build from "../build/server";
+import { getLoadContext } from "../load-context";
 
 export const onRequest = createPagesFunctionHandler({
   build,
-  getLoadContext: (context) => ({ env: context.env }),
+  getLoadContext,
```
Comment (markdalgleish) [inline templates/unstable-vite-cloudflare/functions/[[path]].ts:11]: This pattern is really nice.

.changeset/nervous-rivers-chew.md
```diff
@@ -0,0 +1,50 @@
...
+
+```ts
+cloudflare(getBindingsProxy, {
+  getRemixDevLoadContext: ({ request, env}) => {
+    // return augmented context here
+  },
+})
+```
+
+It is also now recommended that you use `load-context.ts` file to define a shared `getLoadContext` function that you can use for both `getRemixDevLoadContext` in `vite.config.ts` as well as within `functions/[[path]].ts`.
```
markdalgleish, line 49: wdyt?

```suggestion
When using `getRemixDevLoadContext`, it's recommended that you create a `load-context.ts` file to define a shared `getLoadContext` function that you can use for both `getRemixDevLoadContext` in `vite.config.ts` as well as within `functions/[[path]].ts`.
```

docs/future/vite.md
```diff
@@ -147,41 +138,110 @@ export default defineConfig({
...
 
-<docs-info>Vite will not use your Cloudflare Pages Functions (`functions/*`) in development as those are purely for Wrangler routing.</docs-info>
+</docs-info>
 
-#### Augmenting Cloudflare load context in development
+#### Augmenting Cloudflare load context
 
-The Cloudflare preset accepts a `getRemixDevLoadContext` function that can be used to augment the load context in development:
+If you'd like to add additional properties to the load context,
+you can export a `getLoadContext` function from `load-context` that you can wire up to Vite and Cloudflare Pages:
```
markdalgleish, line 165:
```suggestion
you can export a `getLoadContext` function from `load-context.ts` that you can wire up to Vite and Cloudflare Pages:
```

markdalgleish (approved):

markdalgleish:
A couple of minor wording suggestions but nothing blocking 👍

PR #8696, opened 2024-02-06
docs(tutorial): fix typo escape char in shellscript

eeershov:
Closes: #

- [x] Docs

Added backslash to escape the dollar sign, so that it is not treated as the start of a variable or special expression

```shellscript nonumber
touch app/routes/posts.$slug.tsx
```
->
```shellscript nonumber
touch app/routes/posts.\$slug.tsx
```


brophdawg11 (approved):
Thanks!

PR #8639, opened 2024-01-31
doc(templates/spa): fix typo `yo` -> `to`

maksnester:
Closes: #

- [x] Docs

just a dummy typo fix

PR #8707, opened 2024-02-07
template(spa): Provide example of server that can handle SPA routes

ngbrown:
Because this also fixes the typo of "yo" -> "to" this pull request supersedes #8639.

Because `http-server` isn't able to serve the `index.html` file as a fallback for what would otherwise be 404 routes (see #8624) it is best to suggest a different server to preview or run a built SPA application. This change doesn't change how the `start` script in `package.json` uses `http-server` because that is not specifically addressed in the `README.md`.

This change includes part of the `README.md` updates from #8624, but without the changes to `packages/remix-dev/vite/plugin.ts` needed to support `vite preview`.

### DIFF
--- a/templates/spa/README.md
+++ b/templates/spa/README.md
@@ -22,16 +22,20 @@ npm run dev
 
 ## Production
 
-When you are ready yo build a production version of your app, `npm run build` will generate your assets and an `index.html` for the SPA.
+When you are ready to build a production version of your app, `npm run build` will generate your assets and an `index.html` for the SPA.
 
 ```shellscript
 npm run build
 ```
 
-You can serve this from any server of your choosing, for a simple example, you could use [http-server](https://www.npmjs.com/package/http-server):
+### Deployment
+
+You can then serve your app from any HTTP server of your choosing. The server should be configured to serve multiple paths from a single root `/index.html` file (commonly called "SPA fallback"). Other steps may be required if the server doesn't directly support this functionality.
+
+For a simple example, you could use [sirv-cli](https://www.npmjs.com/package/sirv-cli):
 
 ```shellscript
-npx http-server build/client/
+npx sirv-cli build/client/ --single
 ```
 
 [remix-vite-docs]: https://remix.run/docs/en/main/future/vite


templates/spa/README.md
```diff
@@ -22,16 +22,18 @@ npm run dev
...
 ```
 
-You can serve this from any server of your choosing, for a simple example, you could use [http-server](https://www.npmjs.com/package/http-server):
+You can serve this from any server of your choosing. The server should be configured to serve multiple paths from a single root `/index.html` file (commonly called "SPA fallback"). Other steps may be required if the host doesn't directly support this functionality.
+
+For a simple example, you could use [sirv-cli](https://www.npmjs.com/package/sirv-cli):
 
 ```shellscript
-npx http-server build/client/
+npx sirv-cli build/client/ --single --ignores "^/assets/"
```
brophdawg11, line 36: Do we need `--ignores "^/assets/"` here?  It seemed to work ok for me without that?

templates/spa/README.md
```diff
@@ -22,16 +22,18 @@ npm run dev
...
 
-When you are ready yo build a production version of your app, `npm run build` will generate your assets and an `index.html` for the SPA.
+When you are ready to build a production version of your app, `npm run build` will generate your assets and an `index.html` for the SPA.
 
 ```shellscript
 npm run build
 ```
 
-You can serve this from any server of your choosing, for a simple example, you could use [http-server](https://www.npmjs.com/package/http-server):
+You can serve this from any server of your choosing. The server should be configured to serve multiple paths from a single root `/index.html` file (commonly called "SPA fallback"). Other steps may be required if the host doesn't directly support this functionality.
```
brophdawg11, line 31:
```suggestion
You can then serve your app from any HTTP server of your choosing. The server should be configured to serve multiple paths from a single root `/index.html` file (commonly called "SPA fallback"). Other steps may be required if the server doesn't directly support this functionality.
```

templates/spa/README.md
ngbrown: It does work without it, but the http server would then return `200 OK` along with the `index.html` contents for any missing assets.

templates/spa/README.md
ngbrown: Updated the `readme.md` to match the changes made in the `dev` branch at 787648f7. Hopefully this minimizes any conflict on branch merges.

templates/spa/README.md
ngbrown: For example, I'm not sure how Remix detects that it needs to perform a reload because the files on the server got updated. Would that feature be impacted if the assets never `404`?

ngbrown: I should have tested this. Missing asset files will 404 because they have a file extension.

This could be considered a limitation of the `--single` mode where any path with an extension not fall-back to `index.html`.

brophdawg11 (approved):
Thanks!

PR #8709, opened 2024-02-07
Support an optional Layout export from the root route

brophdawg11:
RFC https://github.com/remix-run/remix/discussions/8702
Closes https://github.com/remix-run/remix/discussions/8703

### DIFF
--- a/docs/file-conventions/root.md
+++ b/docs/file-conventions/root.md
@@ -61,13 +61,13 @@ export default function App() {
         {/* Child routes render here */}
         <Outlet />
 
-        {/* Script tags go here */}
-        {/* If you use a nonce-based content security policy for scripts, you must provide the `nonce` prop. Otherwise, omit the nonce prop as shown here. */}
-        <Scripts />
-
         {/* Manages scroll position for client-side transitions */}
         {/* If you use a nonce-based content security policy for scripts, you must provide the `nonce` prop. Otherwise, omit the nonce prop as shown here. */}
         <ScrollRestoration />
+
+        {/* Script tags go here */}
+        {/* If you use a nonce-based content security policy for scripts, you must provide the `nonce` prop. Otherwise, omit the nonce prop as shown here. */}
+        <Scripts />
 
         {/* Sets up automatic reload when you change code */}
         {/* and only does anything during development */}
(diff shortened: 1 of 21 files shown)


docs/file-conventions/root.md
```diff
@@ -64,16 +79,98 @@ export default function App() {
 }
 ```
 
-You can pass extra props to `<Scripts />` like `<Scripts crossOrigin />` for hosting your static assets on a different server than your app.
```
brophdawg11, line 67: Moved this and the nonce section below over to the `<Scripts>` doc

aaronadamsCA:
Does this require the Vite dev server?

In my usage so far it works fine with `remix vite:dev`, but with `remix dev` the hydrated document is not wrapped with the `Layout` component.

I'm not sure if you want a bug report, or if you would rather just document this limitation. It probably depends on how aggressively you want to drop non-Vite support.

ngbrown:
I currently use the `<html className />` for setting the user preferred `dark` class. It should also match the rendered state, so it is synchronized between default browser preference and server with a cookie. Even setting this in `<body className />` still needs additional context to be rendered from the `Layout` component.

I can't set it at any lower of a level than `body` and still have [`headless-portal-root` elements](https://headlessui.com/react/dialog#rendering-to-a-portal) pick up the `dark` class.

Is there a possibility to pass more than just `children` to the `Layout` component?

brophdawg11:
> Is there a possibility to pass more than just children to the Layout component?

No, `Layout` is the top-most component so there's nowhere to determine prop values and pass them through.  `Layout` is really just syntactic sugar for the simple case where the app shell is shared between ahppy and error-path UIs.  If you need more complex differences than you may be better off not using `Layout` and using your own component that accepts props.  However, chances are you're going to be determining those prop values from `useLoaderData` and `useRouteError` and could just as easily grab that info from `useRouteLoaderData("root")` and `useRouteError` in `Layout`

ngbrown:
>  grab that info from `useRouteLoaderData("root")` and `useRouteError` in `Layout`

Thanks, `useRouteLoaderData("root")` was the missing link for me. It gave me the added benefit of having the root `ErrorBoundary` have more of the configuration from the root loader (if the root loader isn't what crashed).

PR #8710, opened 2024-02-07
Don't call serverLoader for client-only example

jkjustjoshing:
I was reading [this page](https://remix.run/docs/en/main/guides/client-data) and noticed an example for

> A route that only depends on a client loader looks like this.

still included a call to `serverLoader()`. This seems wrong, so this PR removes that call.

### DIFF
--- a/docs/guides/client-data.md
+++ b/docs/guides/client-data.md
@@ -135,7 +135,7 @@ export async function clientLoader({
 clientLoader.hydrate = true;
 
 export function HydrateFallback() {
-  return <p>Skeleton rendered during SSR</p>;
+  return <p>Skeleton rendered during SSR</p>; // (2)
 }
 
 export default function Component() {
--- a/docs/guides/client-data.md
+++ b/docs/guides/client-data.md
@@ -128,7 +128,7 @@ import type { ClientLoaderFunctionArgs } from "@remix-run/react";
 export async function clientLoader({
   request,
 }: ClientLoaderFunctionArgs) {
-  const clientData = await getClientData(request);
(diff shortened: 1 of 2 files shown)


docs/guides/client-data.md
```diff
@@ -127,22 +127,15 @@ import type { ClientLoaderFunctionArgs } from "@remix-run/react";
...
 }: ClientLoaderFunctionArgs) {
-  const [serverData, clientData] = await Promise.all([
-    serverLoader(),
-    getClientData(request),
-  ]);
-  return {
-    ...serverData, // (4)
-    ...clientData, // (4)
-  };
+  const clientData = await getClientData(request);
```
brophdawg11, line 131:
```suggestion
  const clientData = await getClientData(request); // (2)
```

docs/guides/client-data.md
```diff
@@ -127,22 +127,15 @@ import type { ClientLoaderFunctionArgs } from "@remix-run/react";
...
-  };
+  const clientData = await getClientData(request); // (2)
+  return clientData;
 }
 // Note: you do not have to set this explicitly - it is implied if there is no `loader`
 clientLoader.hydrate = true;
 
 export function HydrateFallback() {
-  return <p>Skeleton rendered during SSR</p>; // (2)
+  return <p>Skeleton rendered during SSR</p>;
```
brophdawg11, line 138:
```suggestion
  return <p>Skeleton rendered during SSR</p>; // (2)
```

brophdawg11 (approved):
Thanks!

Issue #7529, opened 2023-09-25
Trailing slash causes redirect loop


joelvh:
### What version of Remix are you using?

2.0.0

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. Create a page, ideally a nested path such as `app/routes/folder.file._index.tsx` (URL path `/folder/file`)
2. Set up a reverse proxy (e.g. Shopify App Proxy) pointing to that path (e.g. `https://www.example.com/folder/file`)
3. Update `remix.config.js` to set `publicPath = "/folder/file/build/` so assets link through the reverse proxy
4. Visit the proxy path (e.g. `https://example.myshopify.com/folder/file`) which reverse proxies to `www.example.com`
5. Observe: browser path does not have a trailing slash, but the request from the reverse proxy to the Remix app has a trailing slash (e.g `https://www.example.com/folder/file/`)
6. Observe: browser reloads infinitely with error message in browser console that says [`Initial URL (/folder/file/) does not match URL at time of hydration (/folder/file), reloading page...`](https://github.com/remix-run/remix/blob/94695c3d96ac72daeb18dddd77e6dc6590d16bf4/packages/remix-react/browser.tsx#L183C10-L184C50)

### Expected Behavior

Expect the refresh behavior to realize that a trailing slash should be ignored.

Temporary fix is to place the following in `entry.client.tsx`:

```ts
  const initialPathname = window.__remixContext.url
  const hydratedPathname = window.location.pathname

  /**
   * Fix initial pathname to match browser if we're dealing with trailing slashes.
   */
  if (
    initialPathname !== hydratedPathname &&
    initialPathname.replace(/\/+$/, '') === hydratedPathname.replace(/\/+$/, '')
  ) {
    window.__remixContext.url = hydratedPathname
  }
```

### Actual Behavior

Refresh doesn't realize the two paths are effectively the same.

**UPDATE:** It turns out this is also an issue with links and causes the inverse issue. The `Link` will have a path such as `/folder/file` and the path in the browser becomes `/folder/file/`, which I believe is updated by the router. I've updated the script that fixes this (above). I've also updated the steps to reproduce to call out that we are using a custom `publicPath`.

brophdawg11:
This should be resolved via #9695 and available in the next release 👍

brophdawg11:
We had to revert the fix in #9695 so re-opening this.  We should be able to remove this check entirely instead in #9890

[...]

PR #8730, opened 2024-02-11
Reproduce bug where Remix adds trailing slash to the route URL in case there is a public dir with the same name

silvenon:
Reproduces #7529, i.e. the root issue, I believe.

I wasn't aware this issue has been opened yet, I just realized that one of my routes, `/literature.tsx`, had an infinite redirect loop, it took me quite a while to realize that it's because I have a `public/literature` directory as well. The reason for the redirect loop was because I was removing trailing slashes, and this particular page wouldn't budge. (As a workaround I renamed my `public/literature` directory.)

I can add simulate the exact infinite redirect loop in the test if that makes it better.

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -58,30 +58,12 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/_index.tsx": js`
-        import { json } from "@remix-run/node";
-        import { useLoaderData, Link } from "@remix-run/react";
-
-        export function loader() {
-          return json("pizza");
-        }
-
+      "app/routes/about.tsx": js`
         export default function Index() {
-          let data = useLoaderData();
-          return (
-            <div>
-              {data}
-              <Link to="/burgers">Other Route</Link>
-            </div>
-          )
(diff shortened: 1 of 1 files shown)


silvenon:
I haven't read the issue carefully, but it sounds like it stems from the same problem as mine.

brophdawg11:
It looks like this is actually due to built-in default behavior of the [`express.static`](https://expressjs.com/en/4x/api.html#express.static) middleware used by [`remix-serve`](https://github.com/remix-run/remix/blob/main/packages/remix-serve/cli.ts#L134).

You can disable this by using your own express server and setting `redirect:false`:

```js
app.use(express.static("public", { maxAge: "1h", redirect: false }));
```

silvenon:
Thanks! Now I know why my Express server has been acting weird… I wonder what's the reason for this default behavior.

I assume nothing is going to be done about this in Remix, then?

brophdawg11:
Yeah we don't do any customizations in [remix-serve](https://remix.run/docs/en/main/other-api/serve) - it's just got the enough of the basics to be "production ready" if you don't need to customize anything.

> Remix is designed for you to own your server, but if you don't want to set one up you can use the Remix App Server instead. It's a production-ready, but basic Node.js server built with Express. If you find you want to customize it, use the @remix-run/express adapter instead.

PR #8737, opened 2024-02-12
fix(vite): css file paths when cwd is not the root directory

ksjogo:
Currently the getStylesForUrl functionality is broken when using vite if the current working directory (cwd) is not the rootDirectory.

The issue is that a relative path to the root directory is created and passed to getStylesForFiles.
getStylesForFiles is accessing files by generating a normalized path, this path is normalized on the CWD and not the root directory though.
So if there is a difference between CWD and root, it will miss that path difference.

Testing Strategy:
I did test it with our new remix vite setup.
First remix contribution though, so I am not aware of further conventions here, please advise which tests would help.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -339,6 +339,7 @@
 - knowler
 - konradkalemba
 - krolebord
+- ksjogo
 - kubaprzetakiewicz
 - kuldar
 - kumard3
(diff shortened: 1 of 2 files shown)


markdalgleish (commented):
Thanks for the PR!

This looks like the same issue that https://github.com/remix-run/remix/pull/8752 fixes in a more targeted way, so I'm closing this PR in favour of that one.

If you'd like to validate that the fix in that PR works for you too, that would be super helpful!

PR #8740, opened 2024-02-12
Fix Vite client route requests with pre-processed search params

cythrawll:
Issue if you have middleware that pass query string through any processing. anything WHATWG compatible will render a query param without a value ending with `=` this breaks matching in the vite plugin.

Included code so that vite plugin will match against the supported query string if it ends in '='.

another solution would just change the query string to something a bit more compatible like `?client-route=1` Would be willing to open a PR for that. I went for this one due to it being compatible with current solution.

Testing Strategy:

Tested on a custom express server  that had a searchParams filter that parses the query string that transforms `?client-route` into `?client-route=` because of how WhatWG URL serializes searchParams.

the middleware mock basically looks like this (with 'http-hook' package installed): 
```ts
process.on('httpHooks:request', (req) => {
  const url = new URL(req.url, `http://localhost:3000`);
  const searchParams = url.searchParams;
  req.url = `${url.pathname}?${url.searchParams}`;
});
```

see stackblitz: https://stackblitz.com/edit/remix-run-remix-xzvnld?file=server.ts

Came across this because I have some security hooks/middleware that process the query string.

### DIFF
--- a/.changeset/vite-whatwg-querystring.md
+++ b/.changeset/vite-whatwg-querystring.md
@@ -2,4 +2,4 @@
 "@remix-run/dev": minor
 ---
 
-Vite: Add support for client route query string that has been processed by WHATWG url parser/serializer which adds a `=` to query params without a value.
+Vite: Fix issue where client route file requests fail if search params have been processed before reaching the Remix Vite plugin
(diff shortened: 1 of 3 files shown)


markdalgleish (approved):
Thanks for the PR!

PR #8747, opened 2024-02-13
Fix a bug with SPA mode when the root route had no children

brophdawg11:


integration/spa-mode-test.ts
```diff
@@ -431,6 +431,90 @@ test.describe("SPA Mode", () => {
...
+          "vite.config.ts": js`
+            import { defineConfig } from "vite";
+            import { vitePlugin as remix } from "@remix-run/dev";
+
+            export default defineConfig({
+              plugins: [remix({
+                // We don't want to pick up the app/routes/_index.tsx file from
+                // the template and instead want to use only the src/root.tsx
+                // file below
+                appDirectory: "src",
```
markdalgleish, line 451: Hey, that's a neat trick.

brophdawg11: lol - it took me far too long to figure out why my test wasn't failing and I finally dug deep enough to find out the copy a template first, so even though I wasn't defining an index route, my test ended up with one.  And I started down the road of a flag to delete files/not copy the template before I realized I could do this instead

PR #8749, opened 2024-02-13
Vite: Cloudflare Proxy plugin

pcattori:
- [x] auto-set ssr conditions
- [x] caches / dispose
- [x] test wrangler
- [x] template
- [x] docs
- [x] wrangler typegen
- [x] deployment checklist (setup resources/bindings in CF dashboard)
- [x] remove `dispose` from `cloudflare` context
- [x] rename plugin to `devCloudflareProxyVitePlugin as remixDevCloudflareProxy`
- [x] way to set options for `getPlatformProxy`. passthru?
- [x] order validation
- [x] changeset

```ts
cloudflareProxy({ getLoadContext, persist: true })
```


packages/remix-dev/vite/cloudflare-proxy-plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+                nodeReq,
+                nodeRes
+              ) => {
+                let request = fromNodeRequest(nodeReq);
+                let loadContext = options.getLoadContext
+                  ? await options.getLoadContext({ request, context })
+                  : context;
+                let res = await handler(request, loadContext);
+                await toNodeRequest(res, nodeRes);
+              };
```
pcattori, line 69: inline this function

docs/future/presets.md
```diff
@@ -15,27 +15,20 @@ The config returned by each preset is merged in the order they were defined. Any
...
 
 ```ts filename=vite.config.ts lines=[3,11]
-import {
-  vitePlugin as remix,
-  cloudflarePreset as cloudflare,
-} from "@remix-run/dev";
+import { vitePlugin as remix } from "@remix-run/dev";
-import { getBindingsProxy } from "wrangler";
+// TODO: better example
```
pcattori, line 23: @markdalgleish

markdalgleish: Maybe we should flip the heading order around so we talk about using a preset _after_ showing you how to write one. That way we can just show how to consume the example preset.

aaronadamsCA:
Just a friendly heads-up that this is a wider breaking change than what was mentioned in the changelog. All `@remix-run/cloudflare-pages` projects with a custom `getLoadContext` function are affected. Our apps with `server.ts` files all need to be fixed up now.

It's a better API, no doubt—I'm glad for the change! Just don't want anyone caught out like we were.

pcattori:
@aaronadamsCA can you share what error or issue you encountered and what you change you made to your `getLoadContext` functions?

pcattori:
Ah I think I see the issue. We changed how the context was passed in from `(cloudflare) => ...` to `({context: { cloudflare } }) => ...` . Good call out, I'll see how we can patch any breaks for this

PR #8752, opened 2024-02-14
Fix Vite critical CSS resolution when cwd differs from project root

frandiox:
Hi 👋 

While integrating Remix<>Hydrogen in workerd, I found a weird issue with path resolution when applying critical CSS. Turns out it was resolving `app/entry.client.tsx` using `process.cwd()` (by omission) instead of `rootDirectory`, which can be different in some situations.


markdalgleish (approved):
Thanks for catching this!

PR #8761, opened 2024-02-15
Remove condescending error message

dylantf:
This passive-aggressive "Hey developer :wave_emoji:" message is hostile to your developers and not necessary. This PR removes this.

### DIFF
--- a/integration/client-data-test.ts
+++ b/integration/client-data-test.ts
@@ -534,9 +534,6 @@ test.describe("Client Data", () => {
 
       await app.goto("/parent/child");
       let html = await app.getHtml();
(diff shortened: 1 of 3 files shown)


kiliman:
Hmm... I think Remix is just trying to inject some personality. I don't consider it condescending, but more like self-deprecating. It's basically saying, _"Our error UI really sucks. We're sure you can come up with something way better!"_

ryanflorence:
It is condescending to our super ugly default error component that I personally wrote, not to the developer. You should absolutely provide a better error page with your app's branding and messaging.

Issue #7641, opened 2023-10-11
🗺️ Single Fetch


ryanflorence:
- RFC: #7640

mjackson:
I was talking to @jacob-ebey about this and he suggested we need some hooks for encoding/decoding loader data. I've asked him to write more about that here.

ryanflorence:
@jacob-ebey that sounds great to me. @brophdawg11 and I had discussed a long time ago that @remix-run/router will eventually need a pluggable "load the data" implementation, one use case we were thinking of was for SPAs that want to combine a bunch of graphql fragments into a single query and request. The `encode/decode` step will also make it much better for adding superjson, etc.

I'd get Matt's take on it, I know he's thought a lot about it already.

jacob-ebey:
Step 1: https://github.com/remix-run/react-router/pull/11086

brophdawg11:
We can talk about this on the call today but curious if we need 3/4 to be separate router APIs above?  In the initial POC for `fetchStrategy`(https://github.com/remix-run/react-router/pull/11001), it would have been responsible for both 3 and 4 (calling loaders and decoding their results).

I could probably be convinced pretty easily that `fetchStrategy` and `decodeResponse` as separate APIs is easier for the end user, but here's what I might envision for a single `fetchStrategy` API surface:

```js
createBrowserRouter(routes, {
  fetchStrategy({ request, matches, defaultStrategy }) {
    // 1. Default loader execution / Default response decoding
    let results = await defaultStrategy();

    // 2. Default loader execution / Custom response decoding
    let results = await defaultStrategy((response) => customDecode(response));

    // 3. Custom loader execution / Default response decoding
    let results = Promise.all(matches.map(async (match) => {
      let response = await callLoaderCustom(match);
      // This is where it gets a bit tricky - no super clean way to expose a `defaultDecode()` 
      // function but I would argue we do not need to since it's just standard web fetch stuff 
      // and a one liner if you know your app only deals in JSON/whatever
      let data = await response.json();
      return { type: ResultType.Success, data };
    }));

    // 4. Custom loader execution / Custom response decoding
    let results = Promise.all(matches.map(async (match) => {
      let response = await callLoaderCustom(match);
      let data = await customDecode(response);
      return { type: ResultType.Success, data };
    }));
  }
});
```

The comment in code snippet (3) above is what I was getting at in [this comment][(https://github.com/remix-run/react-router/pull/11086#discussion_r1414482649) as well.

[...]

PR #8773, opened 2024-02-16
Add future flag for Single Fetch

brophdawg11:
Closes: https://github.com/remix-run/remix/issues/7641
RFC: https://github.com/remix-run/remix/discussions/7640

**Todo:**
* [ ] Docs
* [x] Tests
  * Most of the integration tests pass as-is if we just enable to flag
  * Need to figure out a good way to enable tests for this while keeping the current tests enabled without this flag
  * Probably best to pick a handful of the most important "data sensitive" integration suites (loader-test, catch-boundary-test, error-boundary-test, form-test, etc.) and copy/paste all the tests so they run a second time with the flag
  * That gives us regression coverage of both approaches
  * Then when we prep to release v3 we just delete the original set of tests
* [x] Changeset


packages/remix-react/browser.tsx
```diff
@@ -275,9 +272,18 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
         v7_partialHydration: true,
         v7_prependBasename: true,
         v7_relativeSplatPath: window.__remixContext.future.v3_relativeSplatPath,
+        // Single fetch enables this underlying behavior
+        v7_skipActionErrorRevalidation:
+          window.__remixContext.future.unstable_singleFetch === true,
```
brophdawg11, line 277: Single fetch opts into this behavior automatically

packages/remix-react/routes.tsx
```diff
@@ -249,16 +249,34 @@ export function createClientRoutes(
...
     let routeModule = routeModulesCache[route.id];
 
-    async function fetchServerLoader(request: Request) {
+    async function fetchServerLoader(
+      request: Request,
+      unwrap: boolean,
+      singleFetch: unknown
+    ) {
       if (!route.hasLoader) return null;
-      return fetchServerHandler(request, route);
```
brophdawg11, line 254: Instead of always hitting the server here, if single fetch is enabled we will have passed the `singleFetch` method down from `dataStrategy` so use that to ensure we make the right HTTP call to the server

brophdawg11: By proxying this all the way though and branching here as deep as possible - we avoid having to re-implement any of the logic around critical/lazy routes, revalidation, client data etc. in data
strategy and get to leverage all of the existing battle-tested flows.

packages/remix-react/single-fetch.ts
```diff
@@ -0,0 +1,203 @@
...
+    // This function is the way for a loader/action to "talk" to the server
+    let singleFetch: (routeId: string) => Promise<unknown>;
+    if (request.method !== "GET") {
+      // Actions are simple since they're singular - just hit the server
+      singleFetch = async (routeId) => {
+        let url = singleFetchUrl(request.url);
+        let init = await createRequestInit(request);
+        let result = await fetchAndDecode(url, init);
+        return unwrapSingleFetchResult(result as SingleFetchResult, routeId);
+      };
```
brophdawg11, line 37: Action implementation.

Below is the more complex loader implementation which lets all loaders latch onto a single promise for the call.

packages/remix-react/single-fetch.ts
```diff
@@ -0,0 +1,203 @@
...
+// - If neither calls `serverLoader`, we won't make the single fetch call
+// - We delay the single fetch call until the **first** one calls `serverLoader`
+// - However, we cannot wait around to know if the other one calls
+//   `serverLoader`, so we include both of them in the `X-Remix-Routes`
+//   header
+// - This means it's technically possible that the second route never calls
+//   `serverLoader` and we never read the response of that route from the
+//   single fetch call, and thus executing that `loader` on the server was
+//   unnecessary.
+function addRevalidationParam(
```
brophdawg11, line 120: This method handles our fine-grained revalidation query param.  This should only show up on `.data` requests if the app is has implemented shouldRevalidate on one or more of the `matches`.  Otherwise, we don't include a query param and while we may run a few extra loaders, the cacheability of the request should improve greatly resulting in a net-positive impact on server load.

packages/remix-react/single-fetch.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+
+  return url.href;
+}
+
+function singleFetchUrl(reqUrl: string) {
+  let url = new URL(reqUrl);
+  url.pathname = `${url.pathname === "/" ? "_root" : url.pathname}.data`;
+  return url;
```
brophdawg11, line 156: single fetch URLs are `/path/to/whatever.data`, except for the root route.  To avoid a super odd-looking (and potentially invalid?) URL like `/.data`, we use `/_root.data`

packages/remix-react/single-fetch.ts
```diff
@@ -0,0 +1,203 @@
...
+            errorResponse.statusText,
+            errorResponse.data,
+            (errorResponse as any).internal === true
+          ),
+        };
+      }
+    },
+  ]);
+  return decoded.value;
+}
```
brophdawg11, line 181: Make the `fetch` HTTP call and decode via turbo-stream

packages/remix-server-runtime/server.ts
```diff
@@ -265,7 +302,214 @@ async function handleDataRequestRR(
...
+  return new Response(
+    encode(result, [
+      (value) => {
+        if (value instanceof ErrorResponseImpl) {
+          return ["ErrorResponse", { ...value }];
+        }
+      },
+    ]),
+    { headers: resultHeaders }
+  );
```
brophdawg11, line 360: Encode single fetch responses with turbo-stream

packages/remix-server-runtime/server.ts
```diff
@@ -265,7 +302,214 @@ async function handleDataRequestRR(
...
+): Promise<[SingleFetchResult, Headers]> {
+  try {
+    let handlerRequest = new Request(handlerUrl, {
+      method: request.method,
+      body: request.body,
+      headers: request.headers,
+      signal: request.signal,
+      ...(request.body ? { duplex: "half" } : undefined),
+    });
+    let response = await staticHandler.queryRoute(handlerRequest, {
```
brophdawg11, line 378: Single fetch action calls use `queryRoute` since they only hit one route

packages/remix-server-runtime/server.ts
```diff
@@ -265,7 +302,214 @@ async function handleDataRequestRR(
...
+): Promise<[SingleFetchResults, Headers]> {
+  let context: StaticHandlerContext;
+  try {
+    let handlerRequest = new Request(handlerUrl);
+    let loadRouteIds = routesToLoad ? routesToLoad.split(",") : undefined;
+
+    let result = await staticHandler.query(handlerRequest, {
+      requestContext: loadContext,
+      loadRouteIds,
+    });
```
brophdawg11, line 432: Single fetch loader calls use `query` with an optional `loadRouteIds` if fine-grained revalidation is used

packages/remix-server-runtime/server.ts
```diff
@@ -265,7 +302,214 @@ async function handleDataRequestRR(
...
+      if (
+        build.assets.routes[routeId]?.hasLoader &&
+        context.loaderData[routeId] === undefined &&
+        mostRecentError
+      ) {
+        context.errors[mostRecentError[0]] = undefined;
+        context.errors[routeId] = mostRecentError[1];
+        mostRecentError = null;
+      }
+    }
```
brophdawg11, line 488: Don't love this - probably should move this behind a flag in `query` to turn off error bubbling.  The other option is to run a bunch of `queryRoutes` in parallel but I kinda like this mimicing document requests at the moment 🤷

packages/remix-react/single-fetch.ts
sergiodxa: `.data`? I imagine it's not `.json` because with RSC the response can be more than just JSON?

packages/remix-server-runtime/server.ts
```diff
@@ -265,7 +308,233 @@ async function handleDataRequestRR(
...
+
+    let result = await staticHandler.query(handlerRequest, {
+      requestContext: loadContext,
+      loadRouteIds,
+    });
+    if (isResponse(result)) {
+      let redirect = getSingleFetchRedirect(result);
+
+      // We don't really know which loader this came from, include for all routes
+      // TODO: This feels like a hack - change internal query behavior here?
```
brophdawg11, line 459: Fix this by lifting to the root of the response and don't send data

Issue #6711, opened 2023-06-28
Dev server HMR and *.server.ts naming convention doesn't work with monorepos


jrestall:
### What version of Remix are you using?

v1.18

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. Create a monorepo using Nx or Turborepo such as Logan's example at https://github.com/mcansh/remix-shared-routes/ that has UI and/or shared route packages outside the app directory.
2. Setup the new dev server.
3. Modify files outside the main app directory and note HMR doesn't work.
4. Add files outside the main app directory following the *.server.ts pattern and note they don't get replaced with an empty module.

### Expected Behavior

HMR and *.server.ts naming convention works with code outside the app dir.

### Actual Behavior

Following are the areas of code that limit HMR and *.server conventions to only work with code in the app dir. I'm happy to submit a PR to fix this if a better approach can be agreed.

### HMR
https://github.com/remix-run/remix/blob/40a4d7d5e25eb5edc9a622278ab111d881c7c155/packages/remix-dev/compiler/js/plugins/hmr.ts#L118-L122

https://github.com/remix-run/remix/blob/40a4d7d5e25eb5edc9a622278ab111d881c7c155/packages/remix-dev/compiler/js/plugins/hmr.ts#L134-L136
If the goal here is to not add react refresh to node_modules code, could we perhaps explicitly check whether the file is in node_modules instead?

### Naming Convention
https://github.com/remix-run/remix/blob/40a4d7d5e25eb5edc9a622278ab111d881c7c155/packages/remix-dev/compiler/plugins/emptyModules.ts#L24-L28
Instead of limiting this convention to the app dir, if the goal is to exclude node_modules, could we change it to something like the following?
```ts
    args.path.includes("node_modules")
```

DanielSchiavini:
HMR is working fine for me with NX monorepo

jrestall:
> HMR is working fine for me with NX monorepo

Are you using typescript [path aliases](https://moonrepo.dev/docs/guides/javascript/typescript-project-refs#importing-source-files-from-local-packages) to share code in the monorepo or a separate build for the shared packages? I think path aliases are causing the issue since it means the linked source code is outside the app dir.

We use path aliases to share feature routes between multiple remix/react-router apps. Would be great to hear what your Nx setup with Remix looks like.

DanielSchiavini:
[...]

Issue #8790, opened 2024-02-19
In Vite Dev Mode, server code on the browser


lfantone:
### Reproduction

https://stackblitz.com/edit/remix-run-remix-1jhxxt

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.05 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.volta/tools/image/node/20.11.0/bin/node
    Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 10.4.0 - ~/.volta/tools/image/npm/10.4.0/bin/npm
    pnpm: 7.2.1 - ~/.volta/bin/pnpm
  Browsers:
    Chrome: 121.0.6167.184
    Edge: 121.0.2277.128
    Safari: 17.2.1
```

### Used Package Manager

npm

### Expected Behavior

When manually defining the **application routes** using the `routes` property in the _vite config_ all the application should work as expected. This means that _loader functions_ or _any server related code_ should be removed from the browser bundles in **development**. This behaviour is appears when running `remix vite:dev`.

It would be nice that server related code is not present in the browser bundles even when the _routes_ are defined manually in the config file.

### Actual Behavior

When navigating to a route that has been declared in the `vite.config.js` file doesn't work in the browser, you can see the below error in the console:

```
@remix-run_node.js?v=1752416c:1534 Uncaught ReferenceError: process is not defined
    at ../node_modules/util/util.js (@remix-run_node.js?v=1752416c:1534:5)
    at __require (chunk-WXXH56N5.js?v=1752416c:12:50)
    at ../node_modules/stream-slice/index.js (@remix-run_node.js?v=1752416c:7114:16)
    at __require (chunk-WXXH56N5.js?v=1752416c:12:50)
    at ../node_modules/@remix-run/node/dist/upload/fileUploadHandler.js (@remix-run_node.js?v=1752416c:7311:23)
    at __require (chunk-WXXH56N5.js?v=1752416c:12:50)
    at ../node_modules/@remix-run/node/dist/index.js (@remix-run_node.js?v=1752416c:7499:29)
    at __require (chunk-WXXH56N5.js?v=1752416c:12:50)
    at @remix-run_node.js?v=1752416c:7601:16
```

Looking into what has being delivered to the browser, we can still see the `loader` function and any server related code.
[...]

PR #8792, opened 2024-02-19
fix(vite-plugin): remove directoy check on route file

lfantone:
This PR tries to solve [this issue](https://github.com/remix-run/remix/issues/8790)

What I did was to remove the check against the `appDirectory`, the change look safe on the test I run which is based in the reproduction case submitted in the issue. But I don't have the background to know if that check is really needed.
Let me know and I can update the PR 😄

### DIFF
--- a/.changeset/cyan-fans-sleep.md
+++ b/.changeset/cyan-fans-sleep.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Remove directory check to allow the usage of defined routes outside of the app directory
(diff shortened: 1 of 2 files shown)


jrestall:
Nice this would also fix https://github.com/remix-run/remix/issues/6711#issuecomment-1880183280

pcattori:
Superceded by #8795 

Thanks @lfantone for filing the issue and putting up a PR! 🙏 We want this bug fix to land in time for out 2.7 release and rebasing PRs onto the release branch via GitHub UI is messy... So instead, I recreated this PR in #8795 and tagged you as a co-author in that commit.

PR #8797, opened 2024-02-20
Vite: Pass `remixUserConfig` to preset config hook

markdalgleish:
This change allows presets to configure Remix differently based on the user config.

The use case driving this change is to allow Vercel to avoid setting `serverBundles` when consumers have enabled SPA mode, otherwise Vercel preset consumers get the following warning and there's no way to avoid it:

> ⚠️  SPA Mode: the serverBundles config is invalid with unstable_ssr:false and will be ignored

### DIFF
--- a/integration/vite-presets-test.ts
+++ b/integration/vite-presets-test.ts
@@ -7,15 +7,46 @@ import dedent from "dedent";
 
 import { viteBuild, test, createProject } from "./helpers/vite.js";
 
+const js = String.raw;
+
 const files = {
-  "vite.config.ts": dedent`
+  "vite.config.ts": dedent(js`
     import { vitePlugin as remix } from "@remix-run/dev";
     import fs from "node:fs/promises";
     import serializeJs from "serialize-javascript";
 
+    let isDeepFrozen = (obj: any) =>
+      Object.isFrozen(obj) &&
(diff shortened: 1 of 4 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -527,6 +529,9 @@ let deepFreeze = (o: any) => {
 
 export type RemixVitePlugin = (config?: VitePluginConfig) => Vite.Plugin[];
 export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
+  // Prevent mutations to the user config
+  remixUserConfig = deepFreeze(remixUserConfig);
```
markdalgleish, line 533: I figured it's safest to freeze the user config up front rather than delaying it.

integration/vite-presets-test.ts
```diff
@@ -46,16 +77,20 @@ const files = {
             }),
           },
 
+          // Ensure remixConfig is called with a frozen Remix user config
```
markdalgleish, line 80: It's a good thing we had a test for a frozen resolved config object. It prompted me to add this behaviour here too.

pcattori (approved):

PR #8802, opened 2024-02-20
fix(remix-architect): upgrade packages

Xiphe:
moves @architect/functions to a peer dependency in order to allow host app to determine the installed version
upgrades @architect/functions and @types/aws-lambda to latest versions

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -634,6 +634,7 @@
 - xdivby0
 - xHomu
 - XiNiHa
+- Xiphe
 - xstevenyung
 - yasoob
 - yauri-io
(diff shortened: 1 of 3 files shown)


mjsol12:
I would appreciate understanding what factors are contributing to the decision to keep this update pending, as doing so could significantly enhance our ability to leverage architect functionalities.

bmccarthynn:
@MichaelDeBoey Let us know if there is anything we can do to push this forward. It has become a blocker with the latest version of Arc 11 which removes the usage of aws-sdk and causes a conflict with older dependencies, as noted in the PR description.

PR #8829, opened 2024-02-21
Fix build errors with Vite `.css?url` imports

### LINKED ISSUE #8813: Error: [remix] dest already exists.
### Reproduction

Using this template 👉 https://github.com/kiliman/remix-vite-template that doesn't differ very much from the official one 👉 https://github.com/remix-run/remix/tree/main/templates/vite-express

It gives the following error when running `remix vite:build`:

This only happens to me after upgrading to `2.7.0` of Remix

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 104.86 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - ~/.nvm/versions/node/v20.11.1/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.1/bin/npm
    bun: 1.0.28 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 121.0.6167.184
    Safari: 17.3.1
```

### Used Package Manager

npm

### Expected Behavior

Build normally like the previous version (2.6.0)

### Actual Behavior

Breaking the builds

markdalgleish:
Fixes #8813

The linked issue was introduced in #8796 but slipped through because the Vite CSS test didn't assert the build command succeeded without any output to `stderr`. The test still managed to pass because the build output still worked if you ignored the error message which meant all of our assertions still passed. I've updated the test to catch this issue along with the fix.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -758,6 +758,22 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
     return JSON.parse(manifestContents) as Vite.Manifest;
   };
 
+  let getViteManifestAssetPaths = (
+    viteManifest: Vite.Manifest
+  ): Set<string> => {
+    // Get .css?url imports and CSS entry points
+    let cssUrlPaths = Object.values(viteManifest)
+      .filter((chunk) => chunk.file.endsWith(".css"))
+      .map((chunk) => chunk.file);
+
+    // Get bundled CSS files and generic asset types
+    let chunkAssetPaths = Object.values(viteManifest).flatMap(
+      (chunk) => chunk.assets ?? []
+    );
+
+    return new Set([...cssUrlPaths, ...chunkAssetPaths]);
+  };
+
   let createBrowserManifestForBuild = async (): Promise<BrowserManifest> => {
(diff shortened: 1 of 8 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -1238,27 +1254,8 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
...
-          // Handle generic assets that only exist in SSR module graph
-          let ssrChunkAssetPaths = Object.values(ssrViteManifest).flatMap(
-            (chunk) => chunk.assets ?? []
-          );
-
-          let ssrAssetPaths = new Set<string>([
-            ...ssrCssUrlFilePaths,
-            ...ssrChunkAssetPaths,
-          ]);
+          let clientAssetPaths = getViteManifestAssetPaths(clientViteManifest);
```
markdalgleish, line 1257: This was the root of the issue. The logic for extracting assets from the client and server builds differed. These need to be the same because we compare their assets to detect server-only assets that need to be moved to the client build directory. The linked issue was because we were attempting to move a file to a path that already existed.

PR #8834, opened 2024-02-21
feat: cloudflare websocket proxy

izznatsir:
WebSocket proxy for Cloudflare Durable Object based WebSocket.

- [ ] Docs
- [ ] Tests

Testing Strategy:

Has been tested by running an echo room via Durable Object based WebSocket in separate service locally. Must be in separate service as Cloudflare Pages Functions (non advanced mode) [does not support inlined Durable Object](https://developers.cloudflare.com/pages/functions/bindings/#interact-with-your-durable-object-namespaces-locally). Will try to add an integration test later.

### DIFF
--- a/packages/remix-dev/vite/cloudflare-proxy-plugin.ts
+++ b/packages/remix-dev/vite/cloudflare-proxy-plugin.ts
@@ -1,10 +1,13 @@
+import { Server } from "node:http";
+
 import { createRequestHandler } from "@remix-run/server-runtime";
 import {
   type AppLoadContext,
   type ServerBuild,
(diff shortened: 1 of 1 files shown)


rlcurrall:
Looks like there are merge conflicts, is that the only thing blocking this from moving forward?

gregory:
@izznat any update on this?

izznat:
When React Router v7 is out I will try to revisit. The proxy might not be necessary if Vite Environment API is being integrated to the official adapter.

izznat:
With `@cloudflare/vite-plugin` this PR is no longer needed.

PR #8843, opened 2024-02-21
Support custom Vite `build.assetsDir` option

ericchernuka:
I was in need of having the assets directory in a different location due to a proxy. Leveraging Vite's `build#assetsDir`, I was able to achieve most of what I needed, but there was a straggler file at `build/client/assets/manifest-{hash}.js` even though the configuration of the asset was set to a different directory. To ensure this file adheres to the Vite configuration, I've added handling to build the path from the Vite configuration.

Testing Strategy:
I wasn't able to find any test regarding the Vite plugin functionality so I ended up utilizing manual testing using the playground. To recreate this, I leveraged the `template.local` feature of the `scripts/playground` directory.
1. Run the following from the project root `npx create-remix@latest ./scripts/playground/template.local --template remix-run/remix/templates/vite-cloudflare`
2. Update package.json references for `@remix/*` to all reference `*` to match existing base template
3. Run `yarn build` to build fresh version with the changes in this PR
4. Run `yarn playground:new` to create a fresh template
5. Install dependencies for this new playground
6. Run `yarn build`
7. The result should be a build directory with `client` and `server` directories. The `client` directory should be a subdirectory named `assets,` which includes a `manifest-{hash}.js.`
8. Update vite.config.ts to match the following:
```typescript
export default defineConfig({
  plugins: [remixCloudflareDevProxy(), remix(), tsconfigPaths()],
  build: {
    assetsDir: "assets-remix"
  }
});
```
9. From the new playground root, run `yarn build`. The result should be a build directory with two directories, `client` and `server`. Within the `client` directory, there should no longer be an `assets` directory containing `manifest-{hash}.js`, and it should now be part of the `assets-remix` directory.

### DIFF
--- a/.changeset/tender-keys-develop.md
+++ b/.changeset/tender-keys-develop.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Support custom Vite `build.assetsDir` option
--- a/contributors.yml
+++ b/contributors.yml
@@ -171,6 +171,7 @@
 - emzoumpo
 - enbonnet
 - eps1lon
+- ericchernuka
 - esamattis
 - evanwinter
 - everdimension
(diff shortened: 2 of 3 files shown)


markdalgleish (approved):
Thanks for the PR!

PR #8848, opened 2024-02-22
Migrate to pnpm

markdalgleish:


integration/package.json
```diff
@@ -1,14 +1,21 @@
 {
-  "name": "integration-tests",
+  "name": "integration",
   "version": "0.0.0",
   "private": true,
   "description": "deps needed for integration tests",
   "type": "module",
-  "scripts": {},
+  "scripts": {
+    "tsc": "tsc"
```
markdalgleish, line 8: Note to self: Can this go back to being called `typecheck` now that templates aren't in the workspace?

package.json
```diff
@@ -66,14 +46,24 @@
     "@babel/preset-react": "^7.18.6",
     "@babel/preset-typescript": "^7.21.5",
     "@changesets/cli": "^2.25.2",
+    "@cloudflare/kv-asset-handler": "^0.3.0",
```
markdalgleish, line 49: Note to self: Why is this here?

package.json
```diff
@@ -107,14 +100,16 @@
...
+    "fs-extra": "^7.0.1",
     "glob": "8.0.3",
+    "isbot": "^4.1.0",
     "jest": "^29.6.4",
     "jest-watch-select-projects": "^2.0.0",
     "jest-watch-typeahead": "^2.2.2",
     "jsonfile": "^6.0.1",
     "lodash": "^4.17.21",
     "mime": "^3.0.0",
-    "patch-package": "^6.5.0",
```
markdalgleish, line 117: pnpm has this functionality build in 🎉 https://pnpm.io/cli/patch

packages/remix-cloudflare/package.json
```diff
@@ -13,9 +13,12 @@
   "license": "MIT",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
+  "scripts": {
+    "tsc": "tsc"
+  },
   "dependencies": {
     "@cloudflare/kv-asset-handler": "^0.1.3",
-    "@remix-run/server-runtime": "2.7.2"
+    "@remix-run/server-runtime": "workspace:^"
```
markdalgleish, line 21: Note to self: Double check we're using the right version specifiers.

packages/remix-dev/dependencies.ts
```diff
@@ -60,12 +60,14 @@ function getPackageDependenciesRecursive(
...
   try {
-    pkgPath = require.resolve(pkg);
+    pkgPath = require.resolve(pkg, { paths: [__dirname, process.cwd()] });
   } catch (err) {
     if (isErrorWithCode(err) && err.code === "ERR_PACKAGE_PATH_NOT_EXPORTED") {
       // Handle packages without main exports.
       // They at least need to have package.json exported.
-      pkgPath = require.resolve(`${pkg}/package.json`);
+      pkgPath = require.resolve(`${pkg}/package.json`, {
+        paths: [__dirname, process.cwd()],
```
markdalgleish, line 69: Note to self: Call this out in changesets?

packages/remix-dev/vite/plugin.ts
```diff
@@ -959,6 +959,23 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
             ctx.remixConfig.ssr === false
               ? "spa"
               : "custom",
+          ssr: {
+            external: [
```
markdalgleish, line 963: Note to self: Add a comment for this.

packages/remix-react/package.json
```diff
@@ -15,13 +15,19 @@
...
   "dependencies": {
     "@remix-run/router": "1.15.1",
-    "@remix-run/server-runtime": "2.7.2",
+    "@remix-run/server-runtime": "workspace:^",
     "react-router": "6.22.1",
     "react-router-dom": "6.22.1"
   },
   "devDependencies": {
+    "@remix-run/node": "workspace:^",
+    "@remix-run/react": "workspace:^",
```
markdalgleish, line 29: Note to self: Clean up circular dependencies.

package.json
markdalgleish: It's so linting passes for the templates.

package.json
```diff
@@ -107,14 +101,16 @@
...
+    "npm-run-all": "^4.1.5",
```
lifeiscontent, line 113: @markdalgleish I believe pnpm also has `npm-run-all` functionality as well https://pnpm.io/cli/run

integration/helpers/deno-template/package.json
```diff
@@ -1,27 +1,28 @@
...
   "type": "module",
   "scripts": {
-    "build": "node ../../../build/node_modules/@remix-run/dev/dist/cli.js build",
-    "dev": "node ../../../node_modules/npm-run-all/bin/npm-run-all/index.js build --parallel \"dev:*\"",
-    "dev:deno": "node ../../../node_modules/cross-env/src/bin/cross-env.js NODE_ENV=development deno run --unstable --watch --allow-net --allow-read --allow-env ./build/index.js",
-    "dev:remix": "node ../../../build/node_modules/@remix-run/dev/dist/cli.js watch",
-    "start": "node ../../../node_modules/cross-env/src/bin/cross-env.js NODE_ENV=production deno run --unstable --allow-net --allow-read --allow-env ./build/index.js"
+    "build": "remix build",
+    "dev": "npm-run-all build --parallel \"dev:*\"",
+    "dev:deno": "cross-env NODE_ENV=development deno run --unstable --watch --allow-net --allow-read --allow-env ./build/index.js",
```
hilja, line 10: Also `cross-env` is not needed, you can use the shell emulator https://pnpm.io/cli/run#shell-emulator

integration/helpers/create-fixture.ts
```diff
@@ -287,11 +287,6 @@ export async function createFixtureProject(
 
   await fse.ensureDir(projectDir);
   await fse.copy(integrationTemplateDir, projectDir);
-  await fse.copy(
-    path.join(__dirname, "../../build/node_modules"),
-    path.join(projectDir, "node_modules"),
-    { overwrite: true }
-  );
```
markdalgleish, line 294: This throws an error when combined with pnpm's use of symlinks. As an alternative approach, we now always copy `build` to `dist` and then use `workspace:*` dependencies so that the built versions of packages are available within workspace packages.

.changeset/config.json
```diff
@@ -26,7 +26,7 @@
   "access": "public",
   "baseBranch": "dev",
   "updateInternalDependencies": "patch",
-  "ignore": [],
+  "ignore": ["integration", "integration-*"],
```
markdalgleish, line 29: This gets rid of all the changesets noise for packages that aren't published, which would otherwise be made even worse in this PR now that each integration helper template is a workspace package.

.npmrc
```diff
@@ -1 +1,3 @@
-engine-strict=true
\ No newline at end of file
+engine-strict=true
+shell-emulator=true
+enable-pre-post-scripts=true
```
markdalgleish, line 3: We may want to remove this flag in the future, but I've enabled this flag for now to reduce the number of unnecessary changes.

.npmrc
```diff
@@ -1 +1,3 @@
-engine-strict=true
\ No newline at end of file
+engine-strict=true
+shell-emulator=true
```
markdalgleish, line 2: This allows us to remove `cross-env`.

package.json
```diff
@@ -32,29 +13,28 @@
...
     "playwright:integration": "playwright test --config ./integration/playwright.config.ts",
-    "vite-ecosystem-ci:build": "yarn build",
-    "vite-ecosystem-ci:before-test": "yarn playwright install",
-    "vite-ecosystem-ci:test": "yarn playwright:integration vite && yarn clean:integration",
+    "vite-ecosystem-ci:build": "pnpm build",
+    "vite-ecosystem-ci:before-test": "pnpm playwright install",
+    "vite-ecosystem-ci:test": "pnpm playwright:integration vite && pnpm clean:integration",
     "changeset": "changeset",
     "changeset:version": "changeset version && node ./scripts/remove-prerelease-changelogs.mjs && node ./scripts/patchup-version.mjs",
-    "changeset:release": "yarn build --tsc --publish && changeset publish",
```
markdalgleish, line 51: The `--publish` flag has been removed in this PR. Only `--tsc` is required.

scripts/build.mjs
```diff
@@ -1,12 +1,13 @@
...
-const tsc = process.env.CI || args.includes("--tsc") || publish;
+const tsc = process.env.CI || args.includes("--tsc");
 
-exec("yarn", ["rollup", "-c"])
-  .then(() => tsc && exec("yarn", ["tsc", "-b"]))
-  .then(() => publish && exec("node", ["scripts/copy-build-to-dist.mjs"]))
+exec("pnpm", ["rollup", "-c"])
+  .then(() => tsc && exec("pnpm", ["--recursive", "tsc", "-b"]))
+  .then(() =>
+    exec("node", ["scripts/copy-build-to-dist.mjs", ...(tsc ? ["--tsc"] : [])])
```
markdalgleish, line 9: We now always run this script. See https://github.com/remix-run/remix/pull/8848/files#r1513657148. Note that this means it's not guaranteed a `--tsc` build is being run, so we now need to pass the flag down to this script too.

scripts/publish-private.js
```diff
@@ -15,7 +15,7 @@ function getTaggedVersion() {
  * @param {string} tag
  */
 function publish(dir, tag) {
-  execSync(`npm publish --tag ${tag} ${dir}`, { stdio: "inherit" });
+  execSync(`pnpm publish ${dir} --tag ${tag}`, { stdio: "inherit" });
```
markdalgleish, line 18: Since I haven't attempted a publish, I've rearranged args to match the pnpm docs just in case the order is important.

integration/helpers/cf-template/package.json
```diff
@@ -1,29 +1,30 @@
...
+  "name": "integration-cf-template",
+  "version": "0.0.0",
   "private": true,
   "sideEffects": false,
   "type": "module",
   "scripts": {
-    "build": "node ../../../build/node_modules/@remix-run/dev/dist/cli.js build",
-    "dev": "node ../../../build/node_modules/@remix-run/dev/dist/cli.js dev --manual -c \"npm start\"",
-    "start": "node ../../../node_modules/wrangler/bin/wrangler.js dev ./build/index.js"
+    "build": "node ./node_modules/@remix-run/dev/dist/cli.js build",
```
lifeiscontent, line 8: not 100% sure on this one, but I believe if you use `pnpm remix build` this will do the same thing e.g. point to linked cli package

lifeiscontent: if that doesn't work, the next thing I'd try is using `pnpm exec`

markdalgleish: Oh, nice one.

markdalgleish: I tried this and unfortunately this doesn't work with our current setup because the package bins don't exist at the time of `pnpm install`.

PR #8864, opened 2024-02-23
Fix `serverBundles` issue with multiple browser manifests

markdalgleish:
This issue meant that each server bundle generated a separate subset of the browser manifest rather than generating a single unified browser manifest. To fix this issue I've had to split the singular Remix manifest into a separate manifest for both client and server. This ensures that while the server bundle only contains a subset of the manifest, the client always receives the full manifest.

cc @TooTallNate.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -200,7 +200,7 @@ export type ServerBundleBuildConfig = {
 type RemixPluginSsrBuildContext =
   | {
       isSsrBuild: false;
-      generateRemixManifests?: never;
+      getRemixServerManifest?: never;
       serverBundleBuildConfig?: never;
     }
   | {
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -26,7 +26,7 @@ import {
   type RemixConfig as ResolvedRemixEsbuildConfig,
   resolveConfig as resolveRemixEsbuildConfig,
 } from "../config";
-import { type Manifest as BrowserManifest } from "../manifest";
+import { type Manifest as RemixManifest } from "../manifest";
 import invariant from "../invariant";
 import {
   type NodeRequestHandler,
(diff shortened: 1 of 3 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -26,7 +26,7 @@ import {
   type RemixConfig as ResolvedRemixEsbuildConfig,
   resolveConfig as resolveRemixEsbuildConfig,
 } from "../config";
-import { type Manifest as BrowserManifest } from "../manifest";
+import { type Manifest as RemixManifest } from "../manifest";
```
markdalgleish, line 29: This type has been renamed for clarity since the manifest type is shared between client and server, and especially now that its contents can differ too.

packages/remix-dev/vite/plugin.ts
```diff
@@ -638,15 +640,6 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
       serverBundles = undefined;
     }
 
-    // Get the server bundle build config injected by the Remix CLI, if present.
-    let serverBundleBuildConfig = getServerBundleBuildConfig(viteUserConfig);
-
-    // For server bundle builds, override the relevant config. This lets us run
-    // multiple server builds with each one targeting a subset of routes.
-    if (serverBundleBuildConfig) {
-      routes = serverBundleBuildConfig.routes;
```
markdalgleish, line 647: This was the crux of the issue. Both client and server received a subset of routes when in reality only the server should have a subset.

packages/remix-dev/vite/plugin.ts
```diff
@@ -701,13 +694,20 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
     //   mark the old compiler as deprecated
     // - Remove `ServerBuild.mode` in v3
 
+    let routes = ctx.serverBundleBuildConfig
+      ? // For server bundle builds, the server build should only import the
+        // routes for this bundle rather than importing all routes
+        ctx.serverBundleBuildConfig.routes
+      : // Otherwise, all routes are imported as usual
+        ctx.remixConfig.routes;
```
markdalgleish, line 702: This is one of two main parts of the fix. Rather than modifying the resolved routes for the Remix config — which incorrectly impacts the client, not just the server — we use a subset of routes in the server build file.

packages/remix-dev/vite/plugin.ts
```diff
@@ -818,29 +822,49 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
...
     );
 
-    return manifest;
+    // The server manifest is the same as the browser manifest, except for
+    // server bundle builds which only includes routes for the current bundle,
+    // otherwise the server and client have the same routes
+    let remixServerManifest: RemixManifest = {
+      ...remixBrowserManifest,
+      routes: serverRoutes,
+    };
```
markdalgleish, line 857: This is the other main part of the fix. Here we only provide a subset of the manifest on the server when using server bundles, while still writing the full manifest to disk for the client.

TooTallNate (approved):

PR #8871, opened 2024-02-23
docs: Update query string package

nickprinsloo:
The documentation currently suggests using [query-string](https://github.com/sindresorhus/query-string) to support more complex structured data with FormData. Specifically it suggests that you can nest data using `user[name]` syntax. 

This form of nesting is explicitly [not supported](https://github.com/sindresorhus/query-string?tab=readme-ov-file#nesting) by `query-string`. 

This PR updates the documentation to suggest using [qs](https://github.com/ljharb/qs) instead (which does support nesting as described).

### DIFF
--- a/docs/guides/faq.md
+++ b/docs/guides/faq.md
@@ -176,7 +176,7 @@ export async function action({
 
 Using the same input name and `formData.getAll()` covers most cases for wanting to submit structured data in your forms.
 
-If you still want to submit nested structures as well, you can use non-standard form-field naming conventions and the [`query-string`][query_string] package from npm:
+If you still want to submit nested structures as well, you can use non-standard form-field naming conventions and the [`qs`][qs] package from npm:
 
 ```tsx
 <>
@@ -191,7 +191,7 @@ If you still want to submit nested structures as well, you can use non-standard 
 And then in your action:
 
 ```tsx
-import queryString from "query-string";
+import queryString from "qs";
 
 // in your action:
 export async function action({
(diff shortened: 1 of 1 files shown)


brophdawg11:
I wonder if we should link to both libraries and let folks choose - the `qs` package is 5x the size of `query-string`.  I know it's used on the server so size is less of a concern.  I do think most of the time folks are looking for array support, which `query-string` has.

nickprinsloo:
`FormData` natively supports arrays and I'm not sure what using `query-string` adds? The docs already discuss using `FormData` in this way for checkboxes but it works for any input. My assumption was that this section was specifically for nested object data which might include but isn't limited to arrays.

brophdawg11:
> FormData natively supports arrays

Agreed - we call that out right above this section in the docs.

I think it's more of the `[]` notation folks are used to from other languages/frameworks.  That was so common in PHP/Rails/etc that we've had a large number of folks think it's built into the spec lol.  Most folks _don't_ know that `FormData.getAll` is a thing and reach for `[]` out of the box.

> My assumption was that this section was specifically for nested object data which might include but isn't limited to arrays

It's for both IMO.  The first example works with `query-string` and `qs`.  The second only with `qs`.  I don't know if we want to default to recommending a 5x larger lib if users only need the first `[]` example?

```tsx
<>
  // arrays with []
  <input name="category[]" value="comedy" />
  <input name="category[]" value="comedy" />

  // nested structures parentKey[childKey]
  <input name="user[name]" value="Ryan" />
</>
```

Maybe we should split that example up and point to the differences in what each lib handles?  

> If you want `[]` prefer `FormData.getAll`, or use `query-string` or `qs`.  If you want nested objects, use `qs`.

klaemo:
There are more modern, lightweight libraries for this nowadays, like picoquery (as per https://github.com/es-tooling/module-replacements/blob/main/docs/modules/qs.md). I would shy away from recommending something as bloated as `qs` in the docs.

MichaelDeBoey:
Due to latest developments, I would back away from using packages maintained by the maintainer of `qs` tbh and go indeed with one of the alternatives mentioned in the @es-tooling repo like `picoquery`

PR #8894, opened 2024-02-26
Test attempting to use server assets in production

kevlened:
If you want to reference a file buffer on the server, you're unable to, because the remix plugin will move the server assets to the client build. [This seems intentional](https://github.com/remix-run/remix/blob/main/packages/remix-dev/vite/plugin.ts#L1260-L1276), but I'm not sure how to include static resources otherwise.

- [ ] Tests - I'm unsure how you'd like to support binary blobs in the test fixture, but this PR has the bones

Testing Strategy:

I attempted to run the integration test, but the binary file couldn't be loaded.

I created a sample repo using the vanilla vite template (remix-run/remix/templates/vite) as a base: https://github.com/kevlened/remix-vanilla/pull/2/files


markdalgleish (commented):
Thanks for the detailed description and repro.

In your failing test case, the `pdfFile` value is meant to be a publicly addressable URL, not designed for use with `fs.readFile`, e.g. the URL could point to a CDN.

Instead, if you're wanting to get access to file contents while ensuring that the file is part of the module graph, you should use `?raw` to import the asset as a string: https://vitejs.dev/guide/assets#importing-asset-as-string

I'm closing this PR, but feel free to re-open if I've missed something.

kevlened:
I use private binary files as templates on the server, so these can't be hosted assets.

When `?raw` is used on a file that isn't utf-8 encoded (like a pdf), there doesn't seem to be a way to recover the original bytes. I'd also expect `?url` to provide base64, but it doesn't.

I've tried other plugins such as [vite-plugin-arraybuffer](https://github.com/tachibana-shin/vite-plugin-arraybuffer), but that causes other issues.

The most natural approach seems to be retaining the files from the `ssrEmitsAssets` flag. I'm happy to hack around this limitation at the moment by copying these assets from the client assets back to the server assets following the build, but [there would ideally be a flag to override this behavior](https://github.com/remix-run/remix/pull/8042#issuecomment-1817319851).

kevlened:
After moving the files back, I still get a missing file error, so I'll have to investigate a bit more.

kevlened:
While I _was_ able to hack around this, the cleanest solution was to resolve [my issues](https://github.com/tachibana-shin/vite-plugin-arraybuffer/issues/5) with the [vite-plugin-arraybuffer](https://github.com/tachibana-shin/vite-plugin-arraybuffer) by using an older version.

markdalgleish:
Looks like you managed to get [vite-plugin-arraybuffer](https://github.com/tachibana-shin/vite-plugin-arraybuffer) fixed in v0.0.5. To me this plugin is the right solution.

We could have fixed this by copying static assets from the server build rather than moving them, but what I'm trying to avoid is deploying a bunch of unused static assets as part of the server bundle. If your code references a file URL or a CSS file, conceptually that's part of the client bundle, even if it's only ever imported in the server graph (so that it can be passed down via a loader).

PR #8896, opened 2024-02-26
Sync vite templates with others re: entry files and polyfills

brophdawg11:
Align vite templates with others:

**1.** `entry.*` files are included by default and can be removed if no customization is needed

```
> cp templates/remix/app/entry.* templates/vite/app/
> cp templates/express/app/entry.* templates/vite-express/app/
> cp templates/cloudflare-workers/app/entry.* templates/vite-cloudflare/app/
```

**2.** `installGlobals` is called by default and can be removed if desired (this also ensures that `vite dev` and `remix-serve` behave the same).  This is only needed for `templates/vite` since it's in the express server in `vite-express` and not needed for `vite-cloudflare`

### DIFF
--- a/templates/vite/app/entry.server.tsx
+++ b/templates/vite/app/entry.server.tsx
@@ -1,0 +1,140 @@
+/**
+ * By default, Remix will handle generating the HTTP Response for you.
+ * You are free to delete this file if you'd like to, but if you ever want it revealed again, you can run `npx remix reveal` ✨
+ * For more information, see https://remix.run/file-conventions/entry.server
+ */
+
+import { PassThrough } from "node:stream";
+
+import type { AppLoadContext, EntryContext } from "@remix-run/node";
+import { createReadableStreamFromReadable } from "@remix-run/node";
+import { RemixServer } from "@remix-run/react";
+import { isbot } from "isbot";
+import { renderToPipeableStream } from "react-dom/server";
+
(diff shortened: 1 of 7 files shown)


markdalgleish (approved):

brookslybrand (approved):
🎉

nikodunk:
Thank you for syncing this! I was confused by this just yesterday!! :)

templates/vite/app/entry.server.tsx
```diff
@@ -0,0 +1,140 @@
...
+          if (shellRendered) {
+            console.error(error);
+          }
+        },
+      }
+    );
+
+    setTimeout(abort, ABORT_DELAY);
+  });
+}
```
cliffordfajardo, line 140: I saw that there is a newer version of this file for express based remix apps already somewhere else in the remix repo?

I think the version in the link below is a bit simpler too?

https://github.com/remix-run/remix/blob/6c5f84279dc2f3b6745c01d9f68050028107dae5/scripts/playground/template/app/entry.server.tsx#L17

brophdawg11: yeah I know it's been discussed we could simplify these and I think just haven't gotten around to it.  If you're interested I'm sure we'd accept a PR after this gets merged to simplify the node ones - otherwise I'll try to get that handled 👍

PR #8892, opened 2024-02-26
Fix root default HydrateFallback when Layout exists

brophdawg11:
Same as #8859 but for the default root `HydrateFallback`

Closes #8880

PR #8906, opened 2024-02-27
Fix BoundaryShell usage in default HydrateFallback

brophdawg11:
Follow up to https://github.com/remix-run/remix/pull/8892

### DIFF
--- a/packages/remix-react/errorBoundaries.tsx
+++ b/packages/remix-react/errorBoundaries.tsx
@@ -2,7 +2,7 @@ import * as React from "react";
 import type { Location } from "@remix-run/router";
 import { isRouteErrorResponse } from "react-router-dom";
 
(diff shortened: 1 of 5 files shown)


packages/remix-react/errorBoundaries.tsx
```diff
@@ -146,7 +147,8 @@ export function BoundaryShell({
       </head>
       <body>
         <main style={{ fontFamily: "system-ui, sans-serif", padding: "2rem" }}>
-          {contents}
+          {children}
+          {renderScripts ? <Scripts /> : null}
```
brophdawg11, line 151: We can't blindly render `<Scripts/>` in the default `HydrateFallback` since if the user has their own `Layout` we'll get two `<Scripts/>` instances.  So we only want to render it when they don't have a `Layout`.

PR #8932, opened 2024-02-28
Log route module errors prior to reloading the page

### LINKED ISSUE #8791: <Link> tag reloads page instead of navigating
### Reproduction

https://stackblitz.com/~/github.com/Edran0111/debug-falc-on
https://github.com/Edran0111/debug-falc-on

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
    Memory: 2.33 GB / 15.85 GB
  Binaries:
    Node: 20.11.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.4.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (121.0.2277.112)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @remix-run/css-bundle: ^2.6.0 => 2.6.0 
    @remix-run/dev: ^2.6.0 => 2.6.0 
    @remix-run/eslint-config: ^2.6.0 => 2.6.0 
    @remix-run/node: ^2.6.0 => 2.6.0 
    @remix-run/react: ^2.6.0 => 2.6.0 
    @remix-run/serve: ^2.6.0 => 2.6.0
```

### Used Package Manager

npm

### Expected Behavior

I'm encountering a navigation issue between pages in my Remix application. When clicking on the "Mon compte" button(in the navbar), which is supposed to redirect to /login, the current page simply reloads instead of navigating to the intended route. Interestingly, all other <Link> components in my application (such as /a-propos, /a-propos#equipe, and /contact) work as expected. The issue is specific to the "Mon compte" button that points to /login.

I can access the /login page directly by entering the URL in the browser, and from there, navigation to other links like /oubli-mdp, /register, and others in the navbar works without any issues. However, I cannot navigate to the /login page from any other page outside of it.

Additionally, when I'm on the /login, /register, or /oubli-mdp pages, the Remix DevTools do not appear, even when appending "?rdt=true" to the URL. Strangely, the only workaround I've found is to remove the ActionFunction from login.tsx, which allows the navigation <Link> to work, but this is obviously not a viable solution.

Although the remix dev tools says one or two times that there is a hydration error out of 10 refresh, the message that is send does not help me with the correction of the problem.
image
This is what happens when i'm on the _index page, i click once on the button of "Mon compte", the loader function is triggered but I still go back on the _index page

As I'm relatively new to Remix and might be missing something, I'm unable to identify the root cause of this issue. I would appreciate any guidance or suggestions on how to resolve this without having to delete the entire ActionFunction.

### Actual Behavior

When I click on any button that redirects to /login, the page loads correctly without any issues, and it does not reload the page from where I initiated the click.

PS: I encountered a similar issue previously on my _index page, where I had a simple script for counting characters and a button to submit text for translation. An ActionFunction was set up in root.tsx to handle a switch case, which, for the time being, only included a logout case. The presence of the ActionFunction interfered with the character counting and the form submission on the _index page. The issue has been resolved by relocating the ActionFunction to a different route.

### LINKED ISSUE #8915: Redirects silently fail to a route that has an import error
### Reproduction

Repository: https://github.com/cha0s/remix-eat-error

**Heads up**: This is using the **`vite-express`** template and happens in **dev mode**.

There is only one commit on top of the output from `npx create-remix@latest --template remix-run/remix/templates/vite-express` at https://github.com/cha0s/remix-eat-error/commit/60640efde5590dd395171337c927aa20ce9b26c6

Steps:

- Clone repository and `npm install`
- `npm run dev`
- Go to browser and try to click the submit button

---

P.S. I did try to create a test for you guys, but the test wouldn't even build because of the import/export issue! This may be specifically a dev mode bug.

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-6600K CPU @ 3.50GHz
    Memory: 12.58 GB / 31.31 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 20.11.1 - ~/.nvm/versions/node/v20.11.1/bin/node
    Yarn: 1.22.21 - ~/.nvm/versions/node/v20.11.1/bin/yarn
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.1/bin/npm
    pnpm: 8.15.1 - ~/.nvm/versions/node/v20.11.1/bin/pnpm
    bun: 1.0.17 - ~/.bun/bin/bun
    Watchman: 4.9.0 - /usr/bin/watchman
  Browsers:
    Chromium: 122.0.6261.69
  npmPackages:
    @remix-run/dev: ^2.7.2 => 2.7.2 
    @remix-run/express: ^2.7.2 => 2.7.2 
    @remix-run/node: ^2.7.2 => 2.7.2 
    @remix-run/react: ^2.7.2 => 2.7.2 
    vite: ^5.1.0 => 5.1.4
```

### Used Package Manager

npm

### Expected Behavior

There should be some sort of error output about the incorrect import/export.

### Actual Behavior

The redirect to `/test` silently fails and lands the user back on `/` with no output.

brophdawg11:
Closes #8791, #8915

### DIFF
--- a/packages/remix-react/routeModules.ts
+++ b/packages/remix-react/routeModules.ts
@@ -190,10 +190,21 @@ export async function loadRouteModule(
     routeModulesCache[route.id] = routeModule;
     return routeModule;
   } catch (error: unknown) {
-    // User got caught in the middle of a deploy and the CDN no longer has the
-    // asset we're trying to import! Reload from the server and the user
-    // (should) get the new manifest--unless the developer purged the static
-    // assets, the manifest path, but not the documents 😬
+    // If we can't load the route it's likely one of 2 things:
+    // - User got caught in the middle of a deploy and the CDN no longer has the
+    //   asset we're trying to import! Reload from the server and the user
+    //   (should) get the new manifest--unless the developer purged the static
+    //   assets, the manifest path, but not the documents 😬
+    // - Or, the asset trying to be imported has an error (usually in vite dev
+    //   mode), so the best we can do here is log the error for visibility
+    //   (via `Preserve log`) and reload
+
+    // Log the error so it can be accessed via the `Preserve Log` setting
+    console.error(
+      `Error loading route module \`${route.module}\`, reloading page...`
+    );
+    console.error(error);
+
     if (
       window.__remixContext.isSpaMode &&
       // @ts-expect-error
@@ -203,10 +214,11 @@ export async function loadRouteModule(
       // on dev-time errors since it's a vite compilation error and a reload is
       // just going to fail with the same issue.  Let the UI bubble to the error
       // boundary and let them see the error in the overlay or the dev server log
-      console.error(`Error loading route module \`${route.module}\`:`, error);
       throw error;
     }
+
     window.location.reload();
+
     return new Promise(() => {
       // check out of this hook cause the DJs never gonna re[s]olve this
     });
--- a/.changeset/log-module-error.md
+++ b/.changeset/log-module-error.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Log any errors encountered loading a route module prior to reloading the page


packages/remix-react/routeModules.ts
```diff
@@ -190,10 +190,21 @@ export async function loadRouteModule(
...
+    //   assets, the manifest path, but not the documents 😬
+    // - Or, the asset trying to be imported has an error (usually in vite dev
+    //   mode), so the best we can do here is log the error for visibility
+    //   (via `Preserve log`) and reload
+
+    // Log the error so it can be accessed via the `Preserve Log` setting
+    console.error(
+      `Error loading route module \`${route.module}\`, reloading page...`
+    );
+    console.error(error);
```
brophdawg11, line 206: We've never logged this error before since it's swallowed on the reload and (I think) it was much harder to run into this for import issues in dev when using `esbuild`.  But using vite during dev, you don't find out about these module errors until runtime, so we can log them and they're available via `Preserve log`

I'm curious through if they could be surfaced through Vite's overlay? @pcattori @markdalgleish

packages/remix-react/routeModules.ts
```diff
@@ -203,10 +214,11 @@ export async function loadRouteModule(
       // on dev-time errors since it's a vite compilation error and a reload is
       // just going to fail with the same issue.  Let the UI bubble to the error
       // boundary and let them see the error in the overlay or the dev server log
-      console.error(`Error loading route module \`${route.module}\`:`, error);
       throw error;
     }
+
     window.location.reload();
```
brophdawg11, line 220: We could also consider enhancing this behavior in the future too.  Reloading the current page isn't necessarily what we want all the time.  There's 2 usages of this method:

* prefetching - in that case there's no real reason to hard reload since the user may not click on the link, so we could just let the prefetch no-op
* navigating - ideally for the most seamless UX we'd hard-navigate to the _next_ url instead of reloading the current URL in response to a clicked link.  The issue here is that currently `route.lazy` (which calls this) isn't aware of the next location so we'd have to plumb that through in RR if we wanted to smooth out that flow.

raRaRa: It has bugged me many times that the current page gets reloaded instead of the target URL.

PR #8947, opened 2024-02-29
docs: add vite setup to tailwind

hinryd:
The original steps does not work as intended and will cause error.
Vite can natively resolve CSS file, so directly import it should suffice.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -231,6 +231,7 @@
 - hi-ogawa
 - hicksy
 - himorishige
+- hinryd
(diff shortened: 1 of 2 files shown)


markdalgleish (commented):
Thanks for the PR!

I'm actually working on reworking these docs at the moment. With the move to Vite, we shouldn't need to document integrations like this anymore since the setup process isn't Remix specific. I'm closing this for now, but I'd love to hear your thoughts once we update the docs.

PR #8957, opened 2024-03-01
Update docs regarding SPA Template ssr.noExternal setting

brophdawg11:
Follow up to https://github.com/remix-run/remix/pull/8952

### DIFF
--- a/docs/future/spa-mode.md
+++ b/docs/future/spa-mode.md
@@ -222,13 +222,41 @@ startTransition(() => {
 
 - You cannot call `serverLoader`/`serverAction` from your `clientLoader`/`clientAction` methods since there is no running server -- those will throw a runtime error if called
 
-- It's important to note that Remix SPA mode generates your `index.html` file by performing a "pre-render" of your root route on the server during the build
+### Server Build
 
-  - This means that while you're creating a SPA, you still have a "server build" and "server render" step, so you do need to be careful about using dependencies that reference client-only aspects such as `document`, `window`, `localStorage`, etc.
-  - Generally speaking, the way to resolve these issues is to import any browser-only libraries from `entry.client.tsx` so they don't end up in the server build
-  - Otherwise, you can generally solve these by using [`React.lazy`][react-lazy] or the [`<ClientOnly>`][client-only] component from `remix-utils`
+It's important to note that Remix SPA mode generates your `index.html` file by performing a "pre-render" of your root route on the server during the build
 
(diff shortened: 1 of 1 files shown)


docs/future/spa-mode.md
```diff
@@ -228,8 +228,7 @@ startTransition(() => {
   - Generally speaking, the way to resolve these issues is to import any browser-only libraries from `entry.client.tsx` so they don't end up in the server build
   - Otherwise, you can generally solve these by using [`React.lazy`][react-lazy] or the [`<ClientOnly>`][client-only] component from `remix-utils`
 
-- The [SPA Mode template][spa-mode-template] enables the Vite [ssr.noExternal][vite-ssr-noexternal] option by default to automatically bundle all of your dependencies during the server build to avoid most ESM/CJS issues
-  - This may slow down your build a bit — if so, you can try removing this option, or switching to a more targeted array containing the specific dependencies you wish to bundle
+- If you are running into ESM/CJS issues with your app dependencies you may need to set the Vite [ssr.external][vite-ssr-external] and/or [ssr.noExternal][vite-ssr-noexternal] options in your config to exclude or include certain dependencies in your server bundle.
```
brookslybrand, line 231: We may want to be more specific that `noExternal` will bundle the dependencies in your server bundle. I worry a bit that the language as it stands is going to encourage people to just flip a coin putting dependencies in `ssr.external` vs `ssr.noExternal`, whereas they really are always going to want to try `ssr.noExternal` (unless they set that to `true`, but that feels like the unhappy path at this point).

Also might be good to include a quick code snipped, in case they don't realize `ssr` is a top level key in the vite config.

brophdawg11: 👍 

It's tricky to draw the line between how much we should explain vite stuff that's already documented on the vite pages we're linking too.  Bring too much over and our docs get bloated and potentially stale.  Bring too little over and folks who don't click through may not grok the point we're making. I'll push an update to try to make this a bit clearer though - and I agree the code snippet would be helpful

brophdawg11: [bf62380](https://github.com/remix-run/remix/pull/8957/commits/bf62380bd27ad37ec30af17caa1fd1aa3c5730c9)

docs/future/spa-mode.md
```diff
@@ -222,14 +222,41 @@ startTransition(() => {
 
 - You cannot call `serverLoader`/`serverAction` from your `clientLoader`/`clientAction` methods since there is no running server -- those will throw a runtime error if called
 
-- It's important to note that Remix SPA mode generates your `index.html` file by performing a "pre-render" of your root route on the server during the build
+### Server Build
```
brophdawg11, line 225: Moved the bottom 2 more complex bullet points to sub headings to make them linkable and facilitate a code snippet

Issue #8951, opened 2024-03-01
You cannot `useLoaderData` in an errorElement (routeId: root)


cleytoncarvalho:
### Reproduction

- Clone https://github.com/cleytoncarvalho/remix-error/
- Install dependencies, `npm install`
- Run the app, `npm run dev`
- Open http://localhost:5173
- Click the link "Open page with thrown error"

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (8) arm64 Apple M1
    Memory: 65.91 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.asdf/installs/nodejs/20.11.0/bin/node
    Yarn: 1.22.19 - ~/.asdf/installs/nodejs/20.11.0/bin/yarn
    npm: 10.2.4 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 7.22.0 - ~/.asdf/installs/nodejs/20.11.0/bin/pnpm
    bun: 1.0.29 - ~/.asdf/shims/bun
    Watchman: 2023.06.12.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 122.0.6261.94
    Safari: 17.2.1
  npmPackages:
    @remix-run/dev: ^2.8.0 => 2.8.0 
    @remix-run/node: ^2.8.0 => 2.8.0 
    @remix-run/react: ^2.8.0 => 2.8.0 
    @remix-run/serve: ^2.8.0 => 2.8.0 
    vite: ^5.1.0 => 5.1.4
```

### Used Package Manager

npm

### Expected Behavior

Show error in the ErrorBoundary

### Actual Behavior

Page shows "Unexpected Server Error"
Console shows "You cannot `useLoaderData` in an errorElement (routeId: root)"

The error occurs when useLoaderData is used inside Layout export from root.

manzano78:
There is a [discussion](https://github.com/remix-run/remix/discussions/8846) which highlights an API lack to correctly use data or error in the new Layout component. It exposes a workaround which could probably help you before an expected cleaner way!

brophdawg11:
[...]

PR #8958, opened 2024-03-01
Add note to docs on layout useLoaderData restriction

brophdawg11:
Added some docs to clear up confusion around the inability to use `useLoaderData` in a `Layout` component when rendering an error boundary

See https://github.com/remix-run/remix/issues/8951 and https://github.com/remix-run/remix/discussions/8846.

### DIFF
--- a/docs/file-conventions/root.md
+++ b/docs/file-conventions/root.md
@@ -150,9 +150,9 @@ export function ErrorBoundary() {
 
 `useLoaderData` is not permitted to be used in `ErrorBoundary` components because it is intended for the happy-path route rendering, and its typings have a built-in assumption that the `loader` ran successfully and returned something. That assumption doesn't hold in an `ErrorBoundary` because it could have been the `loader` that threw and triggered the boundary! In order to access loader data in `ErrorBoundary`'s, you can use `useRouteLoaderData` which accounts for the loader data potentially being `undefined`.
 
-Because your `Layout` component is used in both success and error flows, this same restriction holds. If you need to fork logic in your `Layout` depending on if it was a successful request or not, you can use `useRouteLoaderData("root")`:
+Because your `Layout` component is used in both success and error flows, this same restriction holds. If you need to fork logic in your `Layout` depending on if it was a successful request or not, you can use `useRouteLoaderData("root")` and `useRouteError()`:
 
-```tsx
+```tsx filename="app/root.tsx" lines=[6-7,19-29,32-34]
 export function Layout({
   children,
 }: {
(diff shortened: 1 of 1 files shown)


docs/file-conventions/root.md
```diff
@@ -146,6 +146,24 @@ export function ErrorBoundary() {
...
+```tsx
+export function Layout({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  const data = useRouteLoaderData("root");
+  const error = useRouteError();
+  return <html>...</html>;
+}
```
brookslybrand, line 164: Only suggestion I have is including a little more example of what you would do with this information. Open to a better example, just figure this might help folks understand the purpose a bit better

```suggestion
export function Layout({
  children,
}: {
  children: React.ReactNode;
}) {
  const data = useRouteLoaderData("root");
  const error = useRouteError();
  return (
    <html>
      <head>{/* */}</head>
      <body>
        {data ? (
          <RouteLayout>{children}</RouteLayout>
        ) : (
          <ErrorLayout>{children}</ErrorLayout>
        )}
        <ScrollRestoration />
        <Scripts />
      </body>
    </html>
  );
}
```

brookslybrand (approved):

docs/file-conventions/root.md
sergiodxa: Adding analytics using an env variable from the root loader is a good use case

If `useRouteLoaderData("root")` returns a value, you render `<Analytics token={loaderData.env.analytics} />`, otherwise you don't

docs/file-conventions/root.md
```diff
@@ -146,6 +146,24 @@ export function ErrorBoundary() {
 }
 ```
 
+**A note on `useLoaderData`in the `Layout` Component**
+
+`useLoaderData` is not permitted to be used in `ErrorBoundary` components because it is intended for the happy-path route rendering, and it's typings have a built-in assumption that the `loader` ran successfully and returned something. That assumption doesn't hold in an `ErrorBoundary` because it could hve been the `loader` that threw and triggered the boundary! In order to access loader data in `ErrorBoundary`'s, you can use `useRouteLoaderData` which accounts for the loader data being potentially `undefined`.
```
houmark, line 151: it's → its
 hve → have
 super nit: being potentially → potentially being

```suggestion
`useLoaderData` is not permitted to be used in `ErrorBoundary` components because it is intended for the happy-path route rendering, and its typings have a built-in assumption that the `loader` ran successfully and returned something. That assumption doesn't hold in an `ErrorBoundary` because it could have been the `loader` that threw and triggered the boundary! In order to access loader data in `ErrorBoundary`'s, you can use `useRouteLoaderData` which accounts for the loader data potentially being `undefined`.
```

docs/file-conventions/root.md
brophdawg11: It's tricky because there are other ways to do both of these.  I prefer I prefer doing the dynamic stylesheet via `meta`, and the `<Analytics>` component could probably be done via the root component - but I do understand folks will want to put stuff that feels "global" like this in the `Layout`.  I updated with the Analytics example and the CSS custom property example from the linked discussion. 👍

manzano78: @brophdawg11 thank you very much! I note that you recommend `meta` for dynamic stylesheet `<link>`s but the doc precisely discourages the `meta` export for stylesheets in favor ot the `links` one instead (which can't depend on the data):

> For asset links like stylesheets and favicons, you should use the [links export](https://remix.run/docs/en/main/route/links) instead.

 I always want to follow the official recommendations but this use case seems to be in a grey area. Could you maybe add to the docs that dynamic links are an exception of what is tolerated in the `meta` export? It's not ideal though because it's not the intention of this export (made for SEO). Any suggestion? :)

brophdawg11: The `link` here depended on loader data so it could not be done in `links` and would have to be done in `meta` (or could also be done in `Layout`)

PR #8995, opened 2024-03-06
Simplify the URL search param example in the State Management docs.

RDGthree:
This changes the example to use `Link` components for routing instead of a `Form` and `button`s (?), and reduces the fairly complex "bad" example to just the first half of said example.

Feels like an obvious change but @brookslybrand and @brophdawg11 weren't so sure. Do with it what you will.

### DIFF
--- a/docs/discussion/state-management.md
+++ b/docs/discussion/state-management.md
@@ -72,51 +72,10 @@ export function List() {
 }
 ```
 
-Now consider you want the URL to update when the user changes the view. Note the state synchronization:
-
-```tsx bad lines=[10,19,27]
-import {
-  useNavigate,
-  useSearchParams,
-} from "@remix-run/react";
-
-export function List() {
-  const navigate = useNavigate();
-  const [searchParams] = useSearchParams();
-  const [view, setView] = React.useState(
-    searchParams.get("view") || "list"
-  );
-
-  return (
-    <div>
-      <div>
-        <button
-          onClick={() => {
-            setView("list");
-            navigate(`?view=list`);
-          }}
-        >
-          View as List
-        </button>
-        <button
-          onClick={() => {
-            setView("details");
-            navigate(`?view=details`);
-          }}
-        >
-          View with Details
-        </button>
-      </div>
-      {view === "list" ? <ListView /> : <DetailView />}
-    </div>
-  );
-}
-```
-
-Instead of synchronizing state, you can simply read and set the state in the URL directly with boring old HTML forms.
+Instead of using state, you could just use links and search parameters.
 
 ```tsx good lines=[5,9-16]
-import { Form, useSearchParams } from "@remix-run/react";
+import { Link, useSearchParams } from "@remix-run/react";
 
 export function List() {
   const [searchParams] = useSearchParams();
@@ -124,14 +83,14 @@ export function List() {
 
   return (
     <div>
-      <Form>
-        <button name="view" value="list">
+      <div>
+        <Link to="?view=list">
           View as List
-        </button>
-        <button name="view" value="details">
+        </Link>
+        <Link to="?view=details">
           View with Details
         </button>
-      </Form>
+      </div>
       {view === "list" ? <ListView /> : <DetailView />}
     </div>
   );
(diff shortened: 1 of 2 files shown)


brookslybrand:
Thanks for putting up your suggestion @RDGthree!

I disagree with you that using a form with a button is complex or bad, it's a pretty standard way of using buttons to drive the view and url on the page (HTML forms without a POST are navigations as well)

I'm with you though that if I were writing this I would probably reach for links first.

However, there's lots of cases where you want to use a button instead of a link for semantic/a11y/styling reasons. At a minimum, a switch from a button to a link here would likely change the look of this page, which is probably not what you want to do, since these buttons are essentially tabs.

If you still feel this example is too complex as it is, I see 3 possible routes:

1. Change the example for a search `Form`, which is a case where you would definitely want to submit a form with a query

```
<Form>
  <input name="q" type="text" />
  <button>Search</button>
</Form>
```

With this we would need to change all of the examples leading up to it to tell the same story
2. Explain that buttons are being used here for semantic, a11y, and/or styling reasons, which is why we want to keep them as html `button`s
3. Call out or add an example showing that you you could use a `Link` instead. Good opportunity to say "By default, a form submission in HTML is just a GET request that automatically serializes your inputs as search params and adds the entry to the history stack"


Any thoughts?

docs/discussion/state-management.md
```diff
@@ -72,66 +72,25 @@ export function List() {
...
-        </button>
-      </div>
-      {view === "list" ? <ListView /> : <DetailView />}
-    </div>
-  );
-}
-```
-
-Instead of synchronizing state, you can simply read and set the state in the URL directly with boring old HTML forms.
+Instead of using state, you could just use links and search parameters.
```
brookslybrand, line 75: I think you cut out too much here. The point is still that you can "simply read and set the state in the URL directly"

RDGthree:
> However, there's lots of cases where you want to use a button instead of a link for semantic/a11y/styling reasons.

If clicking the button does routing, it's a link. Semantically & for accessibility reasons and styling doesn't override those two. That's what links are. Using a button is an anti-pattern.
CleanShot 2024-03-06 at 17 45 55@2x

I appreciate the thorough review in any case. Cheers.

brophdawg11:
I'm unsure why you closed this?  I think we all agree that _in this specific example_ where the search params are static, a `Link` **is** better than a `Form`.  Brooks indicated as much above:

> I'm with you though that if I were writing this I would probably reach for links first

I also like your change at the end converting it to a `Link`.  But I also agree with Brooks that the [middle part you removed](https://github.com/remix-run/remix/pull/8995/files#r1515132393) is fundamental to understand the story being told here.  If you want to revert that section I think we'd be happy to get this merged.

I think the point you're missing, and the thing we are trying to ensure isn't lost here, is that blanket statements such as _"if clicking the button does routing, it's a link"_ are just fundamentally incorrect.  Clicking a button to submit a `<form method="post">` is routing.  As is clicking a button to submit a `<form>` (which defaults to a GET request).

The key difference IMO is that `<form>` can handle cases where the user provides the query params - which links cannot do (while supporting PE).  So I think the ideal update to the docs changes the current example to use `<Link>` as you proposed, but then also adds a subsequent example showing them how you could change the two buttons to instead be driven by a dropdown using `<Form>`:

```jsx
export function List() {
  const [searchParams] = useSearchParams();
  const view = searchParams.get("view") || "list";

  return (
    <div>
      <Form>
        <label>
          Select a view:
          <select name="view" defaultValue={view}>
            <option value="list">List view</option>
            <option value="details">Detail view</option>
          </select>
        </label>
        <button type="submit">Go</button>
      </Form>
      {view === "list" ? <ListView /> : <DetailView />}
    </div>
  );
```

That way we're including both valid approaches, and hinting at when one may be chosen over the other.

If you're interested in re-opening, we'd be happy to get this merged - otherwise I'll tackle this in a new PR tomorrow.  As I said on Twitter/X - we do appreciate constructive feedback on the docs and I think that this is useful update to this example to explain the nuance between links and forms.

PR #9009, opened 2024-03-07
Update docs vite.md

ksjogo:
Depending on the setup the user might need to re-enable the react plugin when disabling remix for testing.
As otherwise React can be undefined when running vitest.
Might be good to mention this in the docs.

### DIFF
--- a/docs/future/vite.md
+++ b/docs/future/vite.md
@@ -1082,6 +1082,23 @@ export default defineConfig({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


markdalgleish (approved):
Thanks for the PR!

ksjogo:
Thanks for improving on it. :)
Was an issue I ran into when converting to vite and wanted to stop other people from hitting it as well.

PR #9015, opened 2024-03-07
Fix experimental releases for pnpm

brophdawg11:


scripts/publish.js
```diff
@@ -15,7 +15,13 @@ function getTaggedVersion() {
  * @param {string} tag
  */
 function publish(dir, tag) {
-  execSync(`pnpm publish ${dir} --access public --tag ${tag}`, {
+  let args = ["--access public", `--tag ${tag}`];
+  if (tag === "experimental") {
+    args.push(`--no-git-checks`);
+  } else {
+    args.push("--publish-branch release-next");
+  }
```
brophdawg11, line 23: By default, `pnpm` will only publish from `main`.  Turn off this check when publishing an experimental from `release-experimental-*` and change the allowed publish branch to `release-next` for normal releases.

scripts/utils.js
```diff
@@ -191,14 +191,20 @@ const updateDenoImportMap = async (importMapPath, nextVersion) => {
...
   await updateRemixVersion("remix", nextVersion);
   await updateRemixVersion("create-remix", nextVersion);
   for (let name of remixPackages.all) {
     await updateRemixVersion(`remix-${name}`, nextVersion);
   }
 
+  // Sync up the pnpm-lock.yaml file with the new experimental version
+  if (syncLockFile) {
+    console.log(chalk.green("  Syncing pnpm lockfile..."));
+    execSync("pnpm install --no-frozen-lockfile");
```
brophdawg11, line 205: Need to do this with the version bump otherwise the CI run will fail on `pnpm install --frozen-lockfile`

markdalgleish (approved):

PR #9026, opened 2024-03-10
Allow `.` in create-remix `--template` repo name

TooTallNate:
Period is a valid character in GitHub repository names.

### Testing Strategy

Build this branch and then run:

```
node packages/create-remix/dist/cli.js --template TooTallNate/nx.js-remix
```

It should work, instead of the current behavior of printing an error.


markdalgleish (approved):

markdalgleish:
Thanks for the PR!

PR #9031, opened 2024-03-11
docs(future/vite): Fix links

Deanmv:
Link contains `.md` so fails, current page where broken link is visible - https://remix.run/docs/en/main/future/vite#configuration


brophdawg11 (approved):
Thanks!

PR #9061, opened 2024-03-15
set node env to production in start script in vite template

aarbi:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

* fresh remix installation using `vite` template: npx create-remix@latest --template remix-run/remix/templates/vite
* npm run build
* npm start
previously this was throwing an error:
```
  if (!origin) throw Error("Dev server origin not set");
                     ^
Error: Dev server origin not set
    at Object.broadcastDevReady (/Users/aarbi/Batcave/jsnodes/remix/vite/node_modules/.pnpm/@remix-run+server-runtime@2.8.1_typescript@5.4.2/node_modules/@remix-run/server-runtime/dist/dev.js:17:22)
    at Server.onListen (/Users/aarbi/Batcave/jsnodes/remix/vite/node_modules/.pnpm/@remix-run+serve@2.8.1_typescript@5.4.2/node_modules/@remix-run/serve/dist/cli.js:123:17)
    at Object.onceWrapper (node:events:632:28)
    at Server.emit (node:events:530:35)
    at emitListeningNT (node:net:1906:10)
    at processTicksAndRejections (node:internal/process/task_queues:81:21)
```
We should not see this error anymore

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,3 +1,4 @@
+- aarbi
 - aaronpowell96
 - aaronshaf
 - AbePlays
--- a/templates/vite/package.json
+++ b/templates/vite/package.json
@@ -6,7 +6,7 @@
     "build": "remix vite:build",
     "dev": "remix vite:dev",
     "lint": "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint .",
-    "start": "remix-serve ./build/server/index.js",
+    "start": "cross-env NODE_ENV=production remix-serve ./build/server/index.js",
     "typecheck": "tsc"
   },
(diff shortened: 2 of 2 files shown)


markdalgleish:
I wasn't able to reproduce this issue following your steps. Could you share a minimal repro?

aarbi:
so I have NODE_ENV=development set on my machine, unsetting that resolved the issue.

Now do we want to enforce the value of it to production? if so this MR would still make sense.
I also noticed that this template folder has been removed, and promoted to the main template, so if there are any changes needed here I will move them accordingly

PR #9063, opened 2024-03-15
Fix SPA mode when single fetch is enabled

brophdawg11:
We can't use `renderToString` when single fetch is enabled, so don't default to (or reveal) `entry.server.spa.tsx` in SPA mode if single fetch is enabled. 

I also removed the `entry.server.tsx` file from `templates/spa` since there's no need for the user to customize it like there may be in non-SPA templates.


packages/remix-dev/cli/commands.ts
```diff
@@ -303,7 +303,8 @@ export async function generateEntry(
   let defaultEntryClient = path.resolve(defaultsDirectory, "entry.client.tsx");
   let defaultEntryServer = path.resolve(
     defaultsDirectory,
-    ctx?.remixConfig.ssr === false
+    ctx?.remixConfig.ssr === false &&
+      ctx?.remixConfig.future.unstable_singleFetch !== true
```
brophdawg11, line 307: Reveal the streaming `entry.server` in SPA mode when single fetch is enabled.  

I wonder if there's any good way to detect if a user is currently using `renderToString` in their own `entry.server` and they flip on single fetch.  From what I recall the error isn't super obvious and React sticks it in a `<template>` at the end of the HTML but no console errors get logged.

packages/remix-dev/config.ts
```diff
@@ -469,7 +469,7 @@ export async function resolveConfig(
   let pkgJson = await PackageJson.load(rootDirectory);
   let deps = pkgJson.content.dependencies ?? {};
 
-  if (isSpaMode) {
+  if (isSpaMode && appConfig.future?.unstable_singleFetch != true) {
```
brophdawg11, line 472: Default to the streaming entry server in SPA mode w/singleFetch if the user doesn't have an `entry.server`

jacob-ebey (approved):

packages/remix-dev/cli/commands.ts
brophdawg11: @jacob-ebey is going to look into a check at built time for this using the `entry.server` AST (separate PR)

PR #9073, opened 2024-03-18
Update single fetch logic when clientLoaders are present

brophdawg11:
Change the logic for `serverLoader` calls from `clientLoader`'s in single fetch mode so we're not over-fetching on the server if a `clientLoader` never calls `serverLoader`.

Todo:
* [x] Update prefetching logic to align

### DIFF
--- a/packages/remix-react/single-fetch.tsx
+++ b/packages/remix-react/single-fetch.tsx
@@ -137,17 +137,6 @@ export function getSingleFetchDataStrategy(
       // 2. if multiple call `serverLoader` only one fetch call is made
       let singleFetchPromise: Promise<SingleFetchResults>;
 
-      let makeSingleFetchCall = async (url: URL) => {
-        let { data } = await fetchAndDecode(url);
-        return data as SingleFetchResults;
-      };
-
-      let matchesHaveClientLoaders = matches.some(
-        (m) => m.shouldLoad && manifest.routes[m.route.id].hasClientLoader
-      );
-
-      // TODO: Need to fix prefetching to take this logic into account
-
       singleFetch = async (routeId) => {
         let results: SingleFetchResults;
         if (manifest.routes[routeId].hasClientLoader) {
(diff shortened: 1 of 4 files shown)


packages/remix-react/single-fetch.tsx
```diff
@@ -134,26 +134,31 @@ export function getSingleFetchDataStrategy(
...
-        if (!singleFetchPromise) {
-          singleFetchPromise = makeSingleFetchCall();
+        let results: SingleFetchResults;
+        if (manifest.routes[routeId].hasClientLoader) {
+          // When a route has a client loader, we make it's own call for just
+          // it's server loader data
+          let url = stripIndexParam(singleFetchUrl(request.url));
+          url.searchParams.set("_routes", routeId);
+          let { data } = await fetchAndDecode(url);
+          results = data as SingleFetchResults;
```
brophdawg11, line 145: `clientLoader` calls through to `serverLoader` make a call for their individual route

packages/remix-react/single-fetch.tsx
```diff
@@ -134,26 +134,31 @@ export function getSingleFetchDataStrategy(
...
+              routeModules,
+              matches.map((m) => m.route),
+              matches.filter((m) => m.shouldLoad).map((m) => m.route),
+              stripIndexParam(singleFetchUrl(request.url))
+            );
+            singleFetchPromise = fetchAndDecode(url).then(
+              ({ data }) => data as SingleFetchResults
+            );
+          }
+          results = await singleFetchPromise;
```
brophdawg11, line 160: Everyone else (non-`clientLoader` routes) latches onto the same promise

packages/remix-react/single-fetch.tsx
```diff
@@ -226,22 +225,29 @@ export function addRevalidationParam(
...
-    if (matchedIds !== loadIds) {
-      url.searchParams.set("_routes", loadIds);
-    }
+  let needsParam = matchedRoutes.some(
+    (r) =>
+      routeModules[r.id]?.shouldRevalidate ||
+      manifest.routes[r.id]?.hasClientLoader
+  );
+  if (!needsParam) {
+    return url;
```
brophdawg11, line 240: We don't bother adding a param if there are no `shouldRevalidate`'s and no `clientLoader`'s

packages/remix-react/single-fetch.tsx
```diff
@@ -226,22 +225,29 @@ export function addRevalidationParam(
...
+  }
+
+  let matchedIds = genRouteIds(matchedRoutes.map((r) => r.id));
+  let loadIds = genRouteIds(
+    loadRoutes
+      .filter((r) => !manifest.routes[r.id]?.hasClientLoader)
+      .map((r) => r.id)
+  );
```
brophdawg11, line 248: Load the routes requested that don't have client loaders

brophdawg11:
Yeah only for that route branch.  Otherwise we nullify the main benefit of `clientLoader` which is to avoid hitting the server loader since the single fetch call would hit it anyway even if you pulled data from the client side cache.  `serverLoader` is essentially the same as the old `?_data` `fetch` via the single fetch call/encoding.

You still get a multi-route single fetch call for any routes that do not have a client loader, so if you have `a/b/c/d` and only `d has a `clientLoader`, you'd get:

```
GET /a/b/c/d.data?_routes=a,b,c (always)

GET /a/b/c/d.data?_routes=d     (only if d called serverLoader())
```

PR #9077, opened 2024-03-19
Promote Vite templates, nest classic templates

markdalgleish:
This moves all non-Vite templates to a `classic-remix-compiler` subdirectory. The `remix` template is now the same as the `vite` template, but the `vite` template has been kept temporarily to give the create-vite CLI time to migrate over.


templates/classic-remix-compiler/fly/tsconfig.json
```diff
@@ -1,19 +1,10 @@
 {
-  "include": [
-    "**/*.ts",
-    "**/*.tsx",
-    "**/.server/**/*.ts",
-    "**/.server/**/*.tsx",
-    "**/.client/**/*.ts",
-    "**/.client/**/*.tsx"
```
pcattori, line 8: I though there was some reason we needed to include `.server` and `.client` explicitly for tsconfig?

pcattori: Ah or maybe that was only for the new `.server` and `.client` _directories_ since `.server`/`.client` files don't start with  a `.`. So probably git got confused and these lines never existed for the classic compiler.

pcattori (approved):

PR #9082, opened 2024-03-19
Move experimental releases to a manually triggered workflow

brophdawg11:
POC to streamline our experimental release process to [run it manually from Github Actions](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow).

Today, we have to:
* checkout a local branch with a specific name
* run `yarn version:experimental`
* push branch and tags to github to trigger a CI workflow base don the specific tag name

With this change we would:
* Push our feature branch we want to release from
* Run this workflow in Github Actions with our feature branch name as an input

The branching and versioning will be subsumed by the action - so less manual stuff to run and we'd be able to delete some of the forks in our `version.js` and `publish.js` script if this pans out.  Plus it would align experimental and nightly releases so we don't have 3 separate ways of releasing anymore (stable, nightly, experimental).

I'd like to merge this and give it a test run, and if it works we can move forward with trimming down `scripts/` and such.  If it doesn't work, we'll revert and pretend it never happened 🙈


.github/workflows/nightly.yml
```diff
@@ -49,10 +49,7 @@ jobs:
       - name: 🕵️ Check for changes
         id: version
         run: |
-          # get latest commit sha
-          SHA=$(git rev-parse HEAD)
-          # get first 7 characters of sha
-          SHORT_SHA=${SHA::7}
+          SHORT_SHA=$(git rev-parse --short HEAD)
```
brophdawg11, line 52: Slightly easier way to do this stolen from our `version.js` script - results in a 9 char SHA instead of 7.

.github/workflows/release-experimental-dispatch.yml
```diff
@@ -0,0 +1,59 @@
+name: 🧪 Experimental Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        required: true
```
brophdawg11, line 7: Can't run this without specifying the branch to checkout and build from

PR #9085, opened 2024-03-19
Scope single fetch headers calls to only loaded routes

brophdawg11:


jacob-ebey (approved):

packages/remix-server-runtime/server.ts
```diff
@@ -447,7 +446,6 @@ async function singleFetchLoaders(
   request: Request,
   handlerUrl: URL,
   staticHandler: StaticHandler,
-  matches: RouteMatch<ServerRoute>[] | null,
```
brophdawg11, line 450: These were no longer used

PR #9092, opened 2024-03-20
WIP: vite-plugin-react-swc integration

pcattori:
See https://github.com/vitejs/vite-plugin-react-swc/pull/191


packages/remix-dev/vite/plugin.ts
```diff
@@ -1676,46 +1678,7 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
...
-
-        code = result.code!;
-        let refreshContentRE = /\$Refresh(?:Reg|Sig)\$\(/;
-        if (refreshContentRE.test(code)) {
-          code = addRefreshWrapper(ctx.remixConfig, code, id);
-        }
-        return { code, map: result.map };
-      },
-    },
+    reactPlugin(),
```
brophdawg11, line 1681: 😍

pcattori:
For React plugin to function correctly, we need to only create `?client-route` virtual modules in production, not in development. Currently, the React plugin will only add React Refresh header/footer to files that define React components (which is correct), but the virtual modules re-export React components defined in their corresponding route modules. We used to manually inject React Refresh header/footer for these vmods too when we vendored/modified the React plugin.

`?client-route` vmods exist solely to guarantee same production-mode treeshaking from classic (esbuild-based) Remix compiler still happens when migrating to Remix Vite. Originally, I was opposed to having `?client-route` vmods exist only in the prod module graph, but since they exist _solely_ to power optimizations like treeshaking that only get applied in prod, its ok for that part of the module graph to be different. Especially since these vmods are one-liners that just re-export client-safe exports, so there isn't a bunch of complexity that could trip us up.

markdalgleish:
@pcattori I've merged main into this PR so usage of `?client-route=1` has been removed in dev.

pcattori:
This cleanup will be in RRv7, so no need to force it into Remix at the moment

PR #9093, opened 2024-03-20
Docs for single fetch

brophdawg11:


docs/guides/single-fetch.md
```diff
@@ -0,0 +1,155 @@
...
+
+Remix introduced support for "Single Fetch" ([RFC][rfc]) behind the [`future.unstable_singleFetch`][future-flags] flag in [`v2.9.0`][2.9.0] which allows you to opt-into this behavior which will become the default in Remix v3.
+
+## Overview
+
+When you enable Single Fetch, Remix will make a single HTTP call to your server on client-side transitions, instead of multiple HTTP calls in parallel (one per loader). If you are currently returning `Response` instances from your loaders (i.e., `json`/`defer`) then you shouldn't _need_ to make any changes to your app code, but please read through the "breaking" changes below to be aware of some of the underlying behavior changes.
+
+### Breaking Changes
+
+- Single fetch uses a new streaming format under the hood via [`turbo-stream`][turbo-stream], which means that we can stream down more complex data than just JSON, including `Date`, `Map`, `Set`, `Promise`, `RegExp`, and `BigInt` instances
```
sergiodxa, line 15: If I remember correctly, `Error` is also included in turbo-stream, even custom errors as far as you define them globally on the browser.

docs/guides/single-fetch.md
```diff
@@ -0,0 +1,155 @@
...
+- Naked objects returned from `loader` and `action` functions are no longer automatically converted into a JSON `Response` and are serialized as-is over the wire
```
sergiodxa, line 16: Is `json` then going to still serialize them? So `return json({ date: new Date() })` and `return { date: new Date() }` are going to return different things?

sergiodxa: I just read below, yes using `json` will convert the Date to a string

docs/guides/single-fetch.md
```diff
@@ -0,0 +1,155 @@
...
+### Streaming Timeout
+
+Previously, Remix has a concept of an `ABORT_TIMEOUT` built-into the default [`entry.server.tsx`][entry-server] files which would terminate the React renderer, but it didn't do anything in particular to clean up any pending deferred promises.
+
+Now that Remix is streaming internally, we can cancel the `turbo-stream` processing and automatically reject any pending promises and stream up those errors to the client. By default, this happens after 4950ms - a value that was chosen to be just under the current 5000ms `ABORT_DELAY` in most entry.server.tsx files - since we need to cancel the promises and let the rejections stream up through the React renderer prior to aborting the React sid eof things.
+
+You can control this by exporting a `streamTimeout` numeric value from your `entry.server.tsx` and Remix will use that as the number of milliseconds after which to reject any outstanding Promises from `loader`/`action`'s. It's recommended to decouple this value from the timeout in which you abort the React renderer - and you should always set the React timeout to a higher value so it has time to stream down the underlying rejections from your `streamTimeout`.
```
sergiodxa, line 49: What if instead of `export const streamTimeout = 10_000` we could do:
```ts
export const config: RemixServerConfig = {
  streamTimeout: 10_000
}
```
Then future options could be added to `RemixServerConfig` without adding more exports

docs/guides/single-fetch.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Revalidations
+
+Previously, Remix would always revalidate all active loaders after _any_ action submission, regardless of the result of the action. You could opt-out of revalidation on a per-route basis via [`shouldRevalidate`][should-revalidate].
+
+With Single Fetch, if an `action` returns or throws a `Response` with a `4xx/5xx` status code, Remix will _not revalidate_ loaders by default. You can then opt-into revalidation on a per-route basis. If an `action` returns or throws anything that is not a 4xx/5xx Response, then the revalidation behavior is unchanged. The reasoning here is that in most cases, if you return a `4xx`/`5xx` Response, you didn't actually mutate any data so there is no need to reload data.
+
+Revalidation is handled via a `?_routes` query string parameter on the single fetch HTTP call which limits the loaders being called. This means that when you are doing fine-grained revalidation, you will have cache enumerations based on the routes being requested - but all of the information isin the URL so you should not need any special CDN configurations (as opposed to if this was done via a custom header that required your CDN to respect the `Vary` header).
```
sergiodxa, line 91:
```suggestion
Revalidation is handled via a `?_routes` query string parameter on the single fetch HTTP call which limits the loaders being called. This means that when you are doing fine-grained revalidation, you will have cache enumerations based on the routes being requested - but all of the information is in the URL so you should not need any special CDN configurations (as opposed to if this was done via a custom header that required your CDN to respect the `Vary` header).
```

docs/guides/single-fetch.md
jacob-ebey: Not all, but a small subset that has the generic new Error(message) constructor signature + being available globally in both environments.

docs/guides/single-fetch.md
manzano78: So if I understand, if we need to set headers and/or a particular status code to our loader response, we must use `json()` or `defer()` and consequently we can't benefit anymore from what turbo-stream brings (nested promises, Dates, etc). If my assumption is right, maybe this tradeoff users will have to make should be noted somewhere in the docs :)

sergiodxa: There's a TODO section about headers, from the RFC, the idea was to give a way to add headers inside the loader directly

docs/guides/single-fetch.md
fernandocanizo: Not sure if this is the right place to discuss this, but since you propose a code change, I'll jump in: it would be beneficial to let the code make the decision suggested in the docs. That way we remove a trip-point. Instead of just having docs specifying that you must ensure stream timeout to be lesser than the `ABORT_DELAY` configuration, set it up by calling a function that takes care of this. Something like:

```ts
// `lib.ts`, or maybe in `vite.config.ts`

// define this somewhere else and export it
import { ABORT_DELAY } from "config.ts"

export const getProperTimeout = () => {
  const timeDifferenceInMs = 10
  return ABORT_DELAY - timeDifferenceInMs
}
```

docs/guides/single-fetch.md
brophdawg11: Yeah that's a good call out to add to the docs - we serialize `error.name` if it exists, and then if `global[serializedName]` exists on the client we'll re-construct with that instead of `new Error`

docs/guides/single-fetch.md
brophdawg11: @sergiodxa - the exported object is definitely nice for typings, but I don't foresee us adding _too_ many more exports and I have a hunch that you'll have more first-hand control over this in the future when we move towards an RSC-enabled landscape. So I don't know if it makes sense to introduce an object export which may never grow beyond 1-2 fields.

@fernandocanizo - I don't personally like the idea of moving these configs anywhere farther from entry-server since they're specifically runtime (and server-only) decisions - whereas fields in the configs are mostly (if not exclusively) build-time decisions. I think no matter what, there will be some developer-education needed which we'll do via docs or even comments in the template source code.  In your proposed API if `getProperTimeout` is not exported by the developer, then the same potential issue remains as if they didn't set them correctly.  I think the vast majority of current Remix devs never touch `ABORT_DELAY` currently and the ones who do are experienced enough to know the implications.  I think the same will be true in the future with `streamTimeout`.

docs/guides/single-fetch.md
brophdawg11: Added a note in [6f55ec4](https://github.com/remix-run/remix/pull/9093/commits/6f55ec4dd6797493a425aab6225ff001b379db37)

docs/guides/single-fetch.md
brophdawg11: The docs have been updated with the new approach for `status`/`headers`.  The tl;dr; is that `headers()` is deprecated and you will now receive a mutable `response` instance (a "response stub") in your loaders/actions that you set `status`/`headers` on directly - and Remix will combine at the end.  You can still return `Response` instances with single fetch and they should _mostly_ work but you may need to make some small tweaks.  You can then incrementally migrate away from returning `Response` instances to returning naked objects and using the new `response` stub

kiliman: BTW: Is throwing a Response still a thing with single fetch?

PR #9094, opened 2024-03-20
Add "CSS Files" guide

markdalgleish:
Now that we're on Vite, this guide covers the pros and cons of bundled CSS vs Vite's `?url` CSS imports. This topic comes up quite a bit so I thought it'd be good to have a guide we can point people to.

### DIFF
--- a/docs/guides/css-files.md
+++ b/docs/guides/css-files.md
@@ -1,0 +1,170 @@
+---
+title: CSS Files
+---
+
+# CSS Files
+
+There are two main ways to manage CSS files in Remix:
+
+- [CSS bundling](#css-bundling)
+- [CSS URL imports](#css-url-imports)
+
+This guide covers the pros and cons of each approach, and provides some recommendations based on your project's specific needs.
+
+## CSS bundling
+
+CSS bundling is the most common approach for managing CSS files in the React community. In this model, styles are treated as module side effects and are bundled into one or more CSS files at the discretion of the bundler. It's simpler to use, requires less boilerplate, and gives the bundler more power to optimize the output.
+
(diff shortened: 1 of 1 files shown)


docs/guides/css-files.md
```diff
@@ -0,0 +1,170 @@
...
+import { Button } from "../components/Button";
+
+export default function HelloRoute() {
+  return <Button>Hello!</Button>;
+}
+```
+
+When consuming this component, you don't have to worry about managing individual CSS files. CSS is treated as private implementation detail of the component. This is a common pattern in many component libraries and design systems and scales quite nicely.
+
+#### CSS bundling is required for some CSS solutions
```
brookslybrand, line 47: Just wanted to check if you're using h4 to avoid having too many header links on the right side nav

If so [I've got a PR to indent them](https://github.com/remix-run/remix-website/pull/231)

brookslybrand (approved):
Looks great! Wasn't sure if there was any reason to include a link to [the Layout export](https://remix.run/docs/en/main/file-conventions/root#layout-export) when talking about links unmounting and remounting. While it's not the same as recovering from hydration issues, I wonder if some folks will be looking at this file trying to solve issues `Layout` solves.

Just a thought, again looks great and I say merge it

docs/guides/css-files.md
markdalgleish: Yep, but also to de-emphasise them a bit visually. I'll merge as is but feel free to tweak them if you like.

markdalgleish:
@brookslybrand Thanks for the review. You're right, the Layout export might be worth a mention, but I'll get this in as a starting point at least.

PR #9096, opened 2024-03-21
Update typescript-eslint

huijiewei:
// In TS 5.0, there is a option called verbatimModuleSyntax. add the following rules to correct this potential side effect import.
"@typescript-eslint/no-import-type-side-effects": "error",

// Method Shorthand Syntax Considered Harmful
// https://www.totaltypescript.com/method-shorthand-syntax-considered-harmful
"@typescript-eslint/method-signature-style": ["error", "property"],

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -242,6 +242,7 @@
 - Hopsken
 - houmark
 - hpnzm
+- huijiewei
 - humphd
 - huw
 - huyb1991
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thank you for the PR, but we're deprecating this plugin in favor of giving users full control over their eslint setups so we're not planning on making further upgrades.  I would recommend moving to the eslint setup in https://github.com/remix-run/remix/tree/main/templates/remix for the latest "recommended" starting point that uses industry standards presets and removes any Remix opinions on how your lint your code (it's your app, not ours!).  This also makes all dependencies a first class dependencies in your app so you have total control over the versions being used.

PR #9106, opened 2024-03-22
feat: use undici as fetch polyfill

jacob-ebey:
Use undici as our fetch polyfill going forward. This will allow us to firstly (and my my favorite part), stop maintaining our own, and secondly and arguably more important, will allow us to smoothly upgrade our users to the Node.js fetch when they are on a node 20+ as they will already be using it via the polyfill. Undici is the node fetch implementation, and now all our installGlobals() will be doing is giving you the latest version regardless of your node version, and I think that's pretty sweet.

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-react/__tests__/setup.ts
+++ b/packages/remix-react/__tests__/setup.ts
@@ -1,3 +1,8 @@
-import { installGlobals } from "@remix-run/node";
+const JSDOMFormData = global.FormData;
+global.TextDecoder = require("util").TextDecoder;
+global.TextEncoder = require("util").TextEncoder;
+global.ReadableStream = require("stream/web").ReadableStream;
+global.WritableStream = require("stream/web").WritableStream;
 
-installGlobals();
+require("@remix-run/node").installGlobals();
+global.FormData = JSDOMFormData;
--- a/jest/jest.config.shared.js
+++ b/jest/jest.config.shared.js
@@ -9,11 +9,6 @@ const ignorePatterns = [
 /** @type {import('jest').Config} */
 module.exports = {
   moduleNameMapper: {
-    "^@remix-run/web-blob$": require.resolve("@remix-run/web-blob"),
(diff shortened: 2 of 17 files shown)


packages/remix-react/__tests__/setup.ts
```diff
@@ -1,3 +1,8 @@
-import { installGlobals } from "@remix-run/node";
+const JSDOMFormData = global.FormData;
+global.TextDecoder = require("util").TextDecoder;
+global.TextEncoder = require("util").TextEncoder;
+global.ReadableStream = require("stream/web").ReadableStream;
+global.WritableStream = require("stream/web").WritableStream;
 
-installGlobals();
+require("@remix-run/node").installGlobals();
```
mjackson, line 7: It seems like `installGlobals` should also install `TextDecoder`, `TextEncoder`, `ReadableStream`, and `WritableStream`, no?

mjackson (commented):
this looks pretty sweet! 🙌  Nice work!

One small question: It seems like in addition to installing the `fetch` family in our polyfill, we should also install `ReadableStream` and `WritableStream`, as well as any other stream classes provided by undici. Do you agree @jacob-ebey ?

packages/remix-react/__tests__/setup.ts
jacob-ebey: This is a JSDom issue here.

garand:
Keep an eye out for this bug. I moved to Undici on a project and had to switch off because the API I was working with wasn't including a content-length.

https://github.com/nodejs/undici/pull/1540

jacob-ebey:
> Keep an eye out for this bug. I moved to Undici on a project and had to switch off because the API I was working with wasn't including a content-length.
> 
> [nodejs/undici#1540](https://github.com/nodejs/undici/pull/1540)

I added a test against this and it appears to have been fixed at some point: https://github.com/nodejs/undici/pull/2995

PR #9120, opened 2024-03-23
Update root.tsx with meta export

AndrewOt:
Changing the template to use the `meta` export function to do things the "remix" way.

I was using this template to upgrade my vite spa app to remix vite and ran into a hydration issue. After a bit of research and tinkering, I discovered that adding my own meta tags was causing the problem. So I figured I could just use the `meta` export. Then I thought I'd fix the template so no other poor soul would run into a hydration error because of this issue.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -38,6 +38,7 @@
 - andrelandgraf
 - andrewbrey
 - AndrewIngram
+- AndrewOt
 - anishpras
 - anmolm96
 - anmonteiro
--- a/templates/vite/app/root.tsx
+++ b/templates/vite/app/root.tsx
@@ -6,12 +6,17 @@ import {
   ScrollRestoration,
 } from "@remix-run/react";
 
+export const meta: MetaFunction = () => {
+  return [
+    { charSet: "utf-8" },
+    { viewport: "width=device-width,initial-scale=1" },
+  ];
+};
+
 export function Layout({ children }: { children: React.ReactNode }) {
   return (
     <html lang="en">
       <head>
-        <meta charSet="utf-8" />
-        <meta name="viewport" content="width=device-width, initial-scale=1" />
         <Meta />
         <Links />
       </head>


brophdawg11:
I think we'll want to keep this as-is, but could you provide a reproduction of the hydration issue for us to peek at?  

We used to include static tags like this in `meta()` but made a conscious decision a while back to move them to root component as `<meta>` tags to remove any unnecessary abstration/indirection.  This way we're doing things the HTML way as much as possible, without unnecessarily reaching for Remixism's.  `meta` is best used for tags that are dynamic based on your `loader` results.

Also, FWIW, eventually with React 19's automatic hoisting of `<meta>`/`<link>` tags to the `<head>`, `meta()` will very likely go away in a future version because you'll be able to just render the dynamic meta tags in your component and let React put them in the `<head>`.

kiliman:
In addition, [`<meta charset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#charset) is recommended to be in the first 1024 bytes of the HTML document as it signals the parser which encoding is being used. We can ensure this element is placed first by explicitly including it in the document vs the `meta()` function.

AndrewOt:
Oh interesting. Didn't know that about the `meta` tag. I would've thought you have some control over that in Remix if I put those tags in the `meta()` function, but I have very little knowledge of how Remix actually works.

Short story is that I never fully figured out what was causing the hydration error. Admittedly, I am a noob at figuring out hydration issues (because at work we still use CRA for all our react code), so I'm a bit new to SSR react. I was trying to upgrade a Vite SPA to Remix that use the MUI component library and got an error about the meta tags which I read was a hydration error. I ended up switching to NextJS (even though I'm a big Remix fan, it has horrible support for CSS in JS libraries :( and in Next, everything just...... worked) and was about to move on with my project. Thank you for looking into this. I will close this PR as it is N/a.

PR #9126, opened 2024-03-24
Remove old Vite template

markdalgleish:
The `vite` template is a duplicate of the newly updated `remix` template. It was only left here until `create-vite` updated to point to the default template. Now that Vite 5.2 has shipped including this PR — https://github.com/vitejs/vite/pull/16203 — the `vite` template is no longer needed.


kiliman:
Removing the Vite template breaks `https://remix.new` as it tries to fork that template. I'm not sure who's responsible for that URL.

image

hceresetti:
> Removing the Vite template breaks `https://remix.new` as it tries to fork that template. I'm not sure who's responsible for that URL.
> 
> image

I discovered this situation almost right now.

I thought to open an issue reporting such occasion.

But then, I found this PR and focused to read it.

Thank you for noticing it and leave me conscious of it.

PR #9138, opened 2024-03-26
Add error handling for not found static assets

heivo:
I noticed that without sending an 404 here the request will be handled by Remix and triggers the root loader. As an alternative setting fallthrough=false would also work, but that would show up in logs.

The issue occurs for example when the devtools are open and the browser tries to fetch sourcemaps.

In my case I was committing an CSRF token in the root loader so the CSRF token was not in sync with the client app anymore.

- [x] Docs
- [ ] Tests

### DIFF
--- a/docs/future/vite.md
+++ b/docs/future/vite.md
@@ -399,7 +399,12 @@ if (viteDevServer) {
     express.static("build/client/assets", {
       immutable: true,
       maxAge: "1y",
-    })
+      fallthrough: true
+    }),
+    (req, res) => {
+      res.status(404).send('Not found')
+      return
+    }
   );
 }
 app.use(express.static("build/client", { maxAge: "1h" }));


brophdawg11:
I'll check with the team, but I don't know if we'll want to adopt this since it's something users can configure, if needed, in their own express server - and it's veering into changing express default behavior.  We've found quite often that when we override a default behavior of a tool we're using (in which the Remix developer already has control) we "fix" the issue for some folks, and "break" it for other folks since there's usually never a one-size-fits-all solution.  And often times the tool has chosen the default with good reason - in this case `express.static` defaults to `fallthrough: true` so I imagine they had their reason for that being a sensible default.

But, FWIW, I think `fallthrough: false` be simpler here for your use case?  Per the [docs](https://expressjs.com/en/4x/api.html#express.static):

> Use false if you have mounted this middleware at a path designed to be strictly a single file system directory, which allows for short-circuiting 404s for less overhead

PR #9142, opened 2024-03-26
Update single fetch headers approach to use response stub API

brophdawg11:


packages/remix-server-runtime/headers.ts
```diff
@@ -16,10 +15,6 @@ export function getDocumentHeaders(
       ? context.matches.slice(0, boundaryIdx + 1)
       : context.matches;
 
-  if (loadRouteIds) {
-    matches = matches.filter((m) => loadRouteIds.includes(m.route.id));
-  }
```
brophdawg11, line 21: single fetch no longer uses `getDocumentHeaders`

packages/remix-server-runtime/routeModules.ts
```diff
@@ -27,6 +27,22 @@ export type DataFunctionArgs = RRActionFunctionArgs<AppLoadContext> &
     context: AppLoadContext;
   };
 
+export const ResponseStubOperationsSymbol = Symbol("ResponseStubOperations");
+export type ResponseStubOperation = [
+  "set" | "append" | "delete",
```
brophdawg11, line 32: Only track "write" operations

packages/remix-server-runtime/server.ts
```diff
@@ -406,60 +410,79 @@ async function singleFetchAction(
       signal: request.signal,
       ...(request.body ? { duplex: "half" } : undefined),
     });
-    let result = await staticHandler.queryRoute(handlerRequest, {
```
brophdawg11, line 409: `action`'s use `query` now as well and pass `skipLoaders` to just call the action.  I think this _should_ let us combine the `singleFetchAction`/`singleFetchLoaders` functions eventually too - but didn't look into that yet as part of this work

packages/remix-server-runtime/server.ts
```diff
@@ -406,60 +410,79 @@ async function singleFetchAction(
...
+
+    let responseStubs = getResponseStubs();
+    let result = await staticHandler.query(handlerRequest, {
       requestContext: loadContext,
+      skipLoaderErrorBubbling: true,
+      skipLoaders: true,
+      unstable_dataStrategy: getSingleFetchDataStrategy(responseStubs),
```
brophdawg11, line 419: We pass a factory-created `dataStrategy` to `query` now so we can have a set of `responseStub`'s scoped to this request

packages/remix-server-runtime/server.ts
```diff
@@ -788,10 +858,125 @@ function unwrapResponse(response: Response) {
...
+    get(target, prop, receiver) {
+      if (prop === "set" || prop === "append" || prop === "delete") {
+        return (name: string, value: string) => {
+          operations.push([prop, name, value]);
+          Reflect.apply(target[prop], target, [name, value]);
+        };
+      }
+      return Reflect.get(target, prop, receiver);
+    },
+  });
```
brophdawg11, line 874: Using a `Proxy` around `Headers` to track write operations

packages/remix-server-runtime/server.ts
```diff
@@ -788,10 +858,125 @@ function unwrapResponse(response: Response) {
...
+          }
+          for (let [k, v] of result.result.headers) {
+            if (k !== "Set-Cookie") {
+              responseStub.headers.set(k, v);
+            }
+          }
+
+          // Unsure why this is complaining?  It's fine in VSCode but fails with tsc...
+          // @ts-expect-error
+          for (let v of result.result.headers.getSetCookie()) {
```
brophdawg11, line 921: Confirming this is fine now that we use `undici`?  I know this was one of the areas we deviated from the spec and didn't yet implement this method

.changeset/response-stub.md
```diff
@@ -0,0 +1,25 @@
...
+  - `response.headers.delete`
+- Each `loader`/`action`receives it's own unique `response` instance so you cannot see what other `loader`/`action` functions have set (which would be subject to race conditions)
+- If all status codes are unset or have values <200, the deepest status code will be used for the HTTP response
+- If any status codes are set to a value >=300, the highest >=300 value will be used for the HTTP Response
+- Remix tracks header operations and will replay them in order (action if present, then loaders top-down) after all handlers have completed on a fresh `Headers` instance that will be applied to the HTTP Response
+  - `headers.set` on any child handler will overwrite values from parent handlers
+  - `headers.append` can be used to set the same header from both a parent and child handler
+  - `headers.delete` can be used to delete a value set by a parent handler, but not a value set from a child handler
+- Because single fetch supports naked object returns, and you no longer need to return a `Response` instance to set status/headers, the `json`/`redirect`/`redirectDocument`/`defer` utilities are considered deprecated when using Single Fetch
```
manzano78, line 23: @brophdawg11 , I totally understand that `json` and `defer` can be considered deprecated but I'm surprised for `redirect` and `redirectDocument`, especially `redirectDocument` which relies on a Remix custom header, ie an implementation that should be ignored in user land I think. So what will be the recommended way to redirect?

PR #9179, opened 2024-04-01
Migrate integration tests to Vite

markdalgleish:


integration/vite-abort-signal-test.ts
```diff
@@ -1,20 +1,19 @@
...
+let cwd: string;
+let stop: () => void;
 
 test.beforeAll(async () => {
-  fixture = await createFixture({
-    files: {
-      "app/routes/_index.tsx": js`
+  port = await getPort();
+  cwd = await createProject({
+    "vite.config.js": await viteConfig.basic({ port: port }),
```
jacob-ebey, line 15: <3

integration/vite-abort-signal-test.ts
```diff
@@ -42,24 +41,18 @@ test.beforeAll(async () => {
...
 
-test.afterAll(() => {
-  appFixture.close();
-});
+test.afterAll(() => stop());
 
 test("should not abort the request in a new event loop", async ({ page }) => {
-  let app = new PlaywrightFixture(appFixture, page);
-  await app.goto("/");
+  await page.goto(`http://localhost:${port}/`);
```
brophdawg11, line 51: Curious if `app -> page` is just a preference or related to the Vite change?  If a preference I think I could get on board - IIRC our `PlaywrightFixture` was mostly just an abstraction to make changing out the underlying testing setup easier (which came in handy when we went from puppeteer to playwright).  But I think we've migrated to using lower-level playwright primitives in so many cases we might not be getting the benefit we think we are there should we choose to switch away from Playwright in the future...

markdalgleish: It's a preference that was only related to the Vite change because our Vite test helpers were written independently from `PlaywrightFixture`. I prefer this lower level styling, but in the interest of time and to keep our changes more scoped, we've opted to migrate `PlaywrightFixture` to Vite by default instead: https://github.com/remix-run/remix/pull/9182.

markdalgleish:
Closing in favour of https://github.com/remix-run/remix/pull/9182.

PR #9182, opened 2024-04-02
Migrate Playwright fixture to Vite by default

markdalgleish:
This changes the default `compiler` option for `createFixture` to be `"vite"` and provides a basic Vite config by default. To support this, we've deprecated the `config` option on `createFixture` since it's coupled to the old `remix.config.js` format, which also means we've added a dedicated `singleFetch` flag.


markdalgleish:
Closing in favour of https://github.com/remix-run/react-router/pull/11395.

PR #9194, opened 2024-04-03
Fix Vite critical CSS resolution for route files with absolute paths

frandiox:
👋 Hey

I was trying to start using Remix presets to add some dev-only routes we have in Hydrogen, such as `/graphiql`. It mostly works but when it requests the critical CSS, it fails because it only supports files that are within the `app` directory. Our route files come from `node_modules` using absolute paths.

This small change does the trick because `path.resolve` does not modify absolute paths 👍

### DIFF
--- a/.changeset/hot-suits-jam.md
+++ b/.changeset/hot-suits-jam.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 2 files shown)


markdalgleish (approved):
Thanks for the PR!

PR #9198, opened 2024-04-04
chore(remix-node): put native fetch behind config option

jacob-ebey:
chore(remix-serve): default to native fetch for single-fetch

Since there are some quite major differences between our, and native fetch implementations we have opted to keep our polyfill around for existing apps until the next major to avoid introducing unexpected issues for users specifically around the garbage collection behavior of response bodies.

Due to reliance on a more spec compliant headers implementation for single fetch, single fetch opts you into native fetch.

### DIFF
--- a/.changeset/good-buses-divide.md
+++ b/.changeset/good-buses-divide.md
@@ -1,0 +1,6 @@
+---
+"@remix-run/node": patch
+"@remix-run/serve": patch
+---
+
+Put native fetch behind config option and default to native fetch for single-fetch.
--- a/.changeset/good-buses-divide.md
+++ b/.changeset/good-buses-divide.md
@@ -3,4 +3,7 @@
 "@remix-run/serve": patch
 ---
 
-Put native fetch behind config option and default to native fetch for single-fetch.
+- Put `undici` fetch polyfill behind a new `installGlobals({ nativeFetch: true })` parameter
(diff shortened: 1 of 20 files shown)


packages/remix-testing/jest.config.js
```diff
@@ -4,5 +4,5 @@ module.exports = {
   displayName: "testing",
   setupFiles: [],
   testEnvironment: "jsdom",
-  setupFilesAfterEnv: ["@testing-library/jest-dom", "./jest.setup.js"],
+  setupFilesAfterEnv: ["./jest.setup.js", "@testing-library/jest-dom"],
```
jacob-ebey, line 7: The order here matters due to testing-library setting up globals badly if they don't exist.

.changeset/good-buses-divide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@remix-run/node": patch
+"@remix-run/serve": patch
+---
+
+Put native fetch behind config option and default to native fetch for single-fetch.
```
brophdawg11, line 6:
```suggestion
- Put `undici` fetch polyfill behind a new `installGlobals({ nativeFetch: true })` parameter
- `remix-serve` will default to using `undici` for the fetch polyfill if `future._unstable_singleFetch` is enabled because the single fetch implementation relies on the `undici` polyfill
  - Any users opting into Single Fetch and managing their own polfill will need to pass the flag to `installGlobals` on their own to avoid runtime errors with Single Fetch

```

brophdawg11 (approved):
Made a minor update to the changeset so it reminds me to make the right updates to the release notes

PR #9210, opened 2024-04-08
update deno to use built in http server

wangbinyq:
deno std http server is marked as deprecated, and `Deno.serve` is the preferred api to use. 

https://deno.land/std@0.221.0/http/server.ts
https://docs.deno.com/runtime/manual/runtime/http_server_apis

### DIFF
--- a/templates/classic-remix-compiler/deno/server.ts
+++ b/templates/classic-remix-compiler/deno/server.ts
@@ -1,7 +1,6 @@
 import { createRequestHandlerWithStaticFiles } from "@remix-run/deno";
 // Import path interpreted by the Remix compiler
 import * as build from "@remix-run/dev/server-build";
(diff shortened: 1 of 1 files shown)


templates/classic-remix-compiler/deno/server.ts
```diff
@@ -1,13 +1,11 @@
...
 import * as build from "@remix-run/dev/server-build";
-import { serve } from "https://deno.land/std@0.128.0/http/server.ts";
 
 const remixHandler = createRequestHandlerWithStaticFiles({
   build,
   getLoadContext: () => ({}),
 });
 
 const port = Number(Deno.env.get("PORT")) || 8000;
-console.log(`Listening on http://localhost:${port}`);
```
wangbinyq, line 12: Remove this, as Deno.serve logs by default.

PR #9218, opened 2024-04-09
chore: remove unused `node-fetch` dependency from `package.json`

MichaelDeBoey:


wouterds:
Ah I was already wondering where this came from and if it was required for a peer dep 👍

PR #9254, opened 2024-04-18
Fix typings for response in LoaderFunctonArgs/ActionFunctionArgs

brophdawg11:
Screenshot 2024-04-18 at 3 00 37 PM

### DIFF
--- a/packages/remix-server-runtime/routeModules.ts
+++ b/packages/remix-server-runtime/routeModules.ts
@@ -40,6 +40,9 @@ export type ResponseStubOperation = [
 export type ResponseStub = {
   status: number | undefined;
   headers: Headers;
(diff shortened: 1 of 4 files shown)


packages/remix-server-runtime/routeModules.ts
```diff
@@ -40,6 +40,9 @@ export type ResponseStubOperation = [
 export type ResponseStub = {
   status: number | undefined;
   headers: Headers;
+};
+
+export type ResponseStubImpl = ResponseStub & {
```
brophdawg11, line 45: This is what we use internally now - since it contains the symbol.  We cast to `ResponseStub` when handing off to loaders/actions

PR #9261, opened 2024-04-19
Add remix.config.js relevance notice to routes discussion page

pavinduLakshan:
So I wanted to manually configure routes in my Remix app, and decided to add a `remix.config.js` to my vite project, but unfortunately it didn't work.

Later when I was randomly navigating through the docs site, I found out that the `remix.config.js` file is only applicable for the projects that use classic remix compiler[1]. For vite projects, the configs need to go inside remix vite plugin config.

I could have saved so much time, if that info is already mentioned in the doc I followed[2]. To prevent this happening for others, thought to send a PR adding the warning to the route discussion page.

[1] https://remix.run/docs/en/main/file-conventions/remix-config
[2] https://remix.run/docs/en/main/discussion/routes

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -494,6 +494,7 @@
 - pacexy
 - pandaiolo
 - panteliselef
+- pavinduLakshan
 - pcattori
 - penspinner
 - penx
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thank you for this PR!  I cherry-pick'd this commit into a new branch since we've made some changes to the docs for our upcoming 2.9.0 release that this will conflicts with, so I'll remake these changes against those docs top avoid conflicts.  

Superceded by #9363

PR #9272, opened 2024-04-21
opt-in type inference for single-fetch

pcattori:
Note: using `include` rather than `typeRoots`/`types` since those are designed for `@types` packages and `include` is lighter weight

### DIFF
--- a/.changeset/young-eagles-grab.md
+++ b/.changeset/young-eagles-grab.md
@@ -3,32 +3,4 @@
 ---
 
 Opt-in types for single-fetch
-
-To opt-in to type inference for single-fetch, add `future/single-fetch.d.ts` to `include` in your `tsconfig.json`:
-
-```json
-{
-  "include": [
-    "./node_modules/@remix-run/react/future/single-fetch.d.ts"
-  ]
-}
-```
-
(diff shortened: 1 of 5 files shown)


kiliman:
Awesome!

.changeset/young-eagles-grab.md
```diff
@@ -0,0 +1,34 @@
+---
+"@remix-run/react": patch
+---
+
+Opt-in types for single-fetch
+
+To opt-in to type inference for single-fetch, add `future/single-fetch.d.ts` to `include` in your `tsconfig.json`:
```
markdalgleish, line 7: I really like this API.

packages/remix-react/future/single-fetch.d.ts
```diff
@@ -0,0 +1,51 @@
...
+) => Serializable;
+
+type Loader = DataFunction & { hydrate?: boolean };
+type Action = DataFunction;
+
+declare module "@remix-run/react" {
+  export function useLoaderData<T>(): T extends Loader
+    ? Awaited<ReturnType<T>>
+    : never;
+
```
brophdawg11, line 47: We will need to fork in here and use the old `SerializeFrom` types if a user returns a `Response` from the `loader`/`action`since that will still go through `res.json()`

PR #9274, opened 2024-04-21
Update tutorial.md

EARodriguezM:
Closes: #

- [x] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -164,6 +164,7 @@
 - dwt47
 - dylanplayer
 - eastlondoner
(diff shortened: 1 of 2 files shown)


brophdawg11:
Neither of these seem to work for me unfortunately.  http redirects to https but the placekitten site seems to be down entirely.  Hopefully it will come back up soon - but this image is controllable by the end user so they can always change it out during the tutorial

Screenshot 2024-04-22 at 8 57 14 AM

PR #9301, opened 2024-04-24
Only add Remix packages to `ssr.external` in Remix repo

markdalgleish:
As noted by the changeset, this fixes an issue where consumers who had added Remix packages to Vite's `ssr.noExternal` option were being overridden by the Remix Vite plugin adding Remix packages to Vite's `ssr.external` option.

Our use of `ssr.noExternal` (fixing a Vite issue within the Remix repo) was incorrectly thought to be a no-op for consumers, so instead we now detect when we're in the Remix repo before applying the fix.

### DIFF
--- a/.changeset/thirty-donuts-clean.md
+++ b/.changeset/thirty-donuts-clean.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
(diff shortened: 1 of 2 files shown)


jacob-ebey (approved):

brophdawg11 (approved):
👍

d7omipni (commented):
5826213613

PR #9320, opened 2024-04-26
Added ?url to CSS import to fix "import not found: default" errors

andrioid:
Closes: #

- [x] Docs
- [ ] Tests

Testing Strategy:

This should hopefully make it so that fresh Remix projects attempting to use Tailwind don't get greeted with "import not found: default" (client) or "malformed url" (server) when following the instructions.

I suspect this is caused by new Vite behavior of automatically injecting imported CSS. The docs clearly expect an URL, but that's not what Vite is returning here.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -1,3 +1,4 @@
+- andrioid
 - aaronpowell96
 - aaronshaf
 - AbePlays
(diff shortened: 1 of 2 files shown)


andrioid:
I just noticed the note on the top of the document. Closing this PR. Might re-open with a complete overhaul of this page to better represent the current defaults, but don't have the time at the moment.

PR #9329, opened 2024-04-27
docs: update tailwind import with additional query string based on ta…

joonmo-you:
The official documentation provided by Tailwind suggests adding a query string at the end of the path to the Tailwind file.
A Remix application doesn't properly import the Tailwind file without the query string.


markdalgleish (approved):
Thanks for the PR!

Issue #9330, opened 2024-04-28
Single fetch: loader type us unknown


ziimakc:
### Reproduction

https://github.com/ziimakc/types_error_remix

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz
    Memory: 3.13 GB / 15.84 GB
  Binaries:
    Node: 18.16.1 - C:\Program Files\nodejs\node.EXE
    npm: 8.14.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.0.6 - ~\AppData\Roaming\npm\pnpm.CMD
  Browsers:
    Edge: Chromium (123.0.2420.97)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @remix-run/dev: 2.9.1 => 2.9.1
    @remix-run/node: 2.9.1 => 2.9.1
    @remix-run/react: 2.9.1 => 2.9.1
    @remix-run/serve: 2.9.1 => 2.9.1
    vite: 5.2.10 => 5.2.10
```

### Used Package Manager

pnpm

### Expected Behavior

For `const xxx = useLoaderData<typeof loader>()` xxx type should be `{ message: string }`

### Actual Behavior

For `const xxx = useLoaderData<typeof loader>()` xxx type is `unknown`

rphlmr:
Try to add `"node_modules/@remix-run/react/future/single-fetch.d.ts"` in `compilerOptions.types`. It works for me.

ziimakc:
@rphlmr thanks, it works using types with `./node_modules/@remix-run/react/future/single-fetch.d.ts`.

Probably we should update docs: https://remix.run/docs/en/main/guides/single-fetch#type-inference

Not sure why includes doesn't work the same way.

trevor-hackett:
From the [docs](https://www.typescriptlang.org/tsconfig/#include):
> Specifies a list of glob patterns that match files to be included in compilation

`include` isn't meant to add types to your app. It's meant to tell typescript about the files you want compiled/type checked.

`compilerOptions.types` is used to add global types to your app without needing to reference them in your code. Since `./node_modules/@remix-run/react/future/single-fetch.d.ts` comes last, it'll override the previous types from `@remix-run/node`

MarkoH17:
[...]

PR #9336, opened 2024-04-29
Change single fetch type override recommendation

brophdawg11:
Closes https://github.com/remix-run/remix/issues/9330


pcattori:
The [official docs for tsconfig `types`](https://www.typescriptlang.org/tsconfig/#types) only specify `@types` packages as being supported. Seems like it also works for other packages/paths without `@types` prefix, but unclear if this is intentional or something that isn't part of the contract and could change without warning in subsequent TS releases. I think we should still go with this since its meant to be a temporary bridge for migrations/upgrades and seems extremely unlikely that TS would change this overnight without warning _during_ this temporary period.

pcattori (approved):

PR #9337, opened 2024-04-29
fix: fix incorrect useActionData type

michaelgmcd:
When a page loads, `actionData` is undefined. The current single-fetch types do not support this.

Could be wrong, but we ran into this issue implementing Single Fetch in our project. TypeScript is happy with everything, but our page crashes due to an undefined `actionData`.


brophdawg11:
Thanks for the PR!  We actually already handled this in https://github.com/remix-run/remix/pull/9322 and should have that released in a patch release early this week

PR #9345, opened 2024-04-30
Add Vite-based `cloudflare-workers` template

markdalgleish:
This adapts the old classic compiler template to use Vite and `cloudflareDevProxyVitePlugin`.


templates/cloudflare-workers/vite.config.ts
```diff
@@ -0,0 +1,19 @@
...
+} from "@remix-run/dev";
+import tsconfigPaths from "vite-tsconfig-paths";
+
+export default defineConfig((env) => ({
+  plugins: [cloudflareDevProxyVitePlugin(), remix(), tsconfigPaths()],
+  ...(env.isSsrBuild && {
+    resolve: {
+      conditions: ["workerd", "worker", "browser"],
+      mainFields: ["browser", "module", "main"],
+    },
```
pcattori, line 14: These could be specified within the `ssr` field so that the `env.isSsrBuild` check is not needed

```suggestion
  ssr: {
    resolve: {
      conditions: ["workerd", "worker", "browser"],
      mainFields: ["browser", "module", "main"],
    },
  },
```

templates/cloudflare-workers/vite.config.ts
```diff
@@ -0,0 +1,19 @@
...
+    build: {
+      minify: true,
```
pcattori, line 16: Why is `minify: true` needed for CF workers?

pcattori (approved):
LGTM left a couple minor comments/questions, but nothing blocking

templates/cloudflare-workers/vite.config.ts
markdalgleish: This was ported from the old template. I guess the idea is to keep the file size to a minimum.

templates/cloudflare-workers/vite.config.ts
markdalgleish: `mainFields` isn't available on `ssr.resolve` 🤔

pcattori: TIL!

markdalgleish: I managed to clean this up to avoid the `env.isSsrBuild` check.

wouterds:
YES 💯

acusti:
i just migrated my cloudflare workers app to the vite setup using this template as my guide, thanks so much for doing this! a note for others making the migration: the big change when migrating from the remix classic compiler to this setup is that all cloudflare workers bindings (KV stores, DOs, env variables, cloudflare Ai, etc.) are exposed on `context.cloudflare.env` (they were previously available directly on context, i.e. `context.env`). likewise, the cloudflare workers [Context API](https://developers.cloudflare.com/workers/runtime-apis/context/) is now available on `context.cloudflare.ctx`, e.g. `context.cloudflare.ctx.waitUntil(…)`.

acusti:
another gotcha i ran into that almost caused me to abandon the whole thing before i worked it out is that after migrating over, when using [workers AI bindings](https://developers.cloudflare.com/workers-ai/configuration/bindings/) locally, wrangler felt the need to make me login again. however, doing so while the server was running (which is when i would be prompted to do so) would consistently crash the server process and fail to complete the login, causing much gnashing of teeth. i resolved this by running `wrangler login` (`yarn wrangler login` in my case, to be specific) and authenticating successfully, then running `yarn dev`.

note that with the classic compiler cloudflare integration, logging in to wrangler worked when prompted while the server was running, hence my confusion.

Issue #9314, opened 2024-04-25
Single fetch `response` parameter for loader is undefined if no default export exists in the route


emilioschepis:
### Reproduction

https://stackblitz.com/edit/remix-run-remix-x55her?file=app%2Froutes%2Fno-default.tsx

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @remix-run/dev: * => 2.9.1 
    @remix-run/node: * => 2.9.1 
    @remix-run/react: * => 2.9.1 
    @remix-run/serve: * => 2.9.1 
    vite: ^5.1.0 => 5.2.10
```

### Used Package Manager

npm

### Expected Behavior

The `response` parameter should always be a `ResponseStub`, even in resource routes without a default export.

### Actual Behavior

Visiting a resource route (in the stackblitz: `app/routes/no-default.tsx`) results in the response starting as `undefined` and an error being printed to the console:

```
The following error is a bug in Remix; please open an issue! https://github.com/remix-run/remix/issues/new
Error: Expected a Response to be returned from queryRoute
```

brophdawg11:
Resource routes don't need a `response` stub because they return actual `Response` instances that contain a `status` and `headers` - so you just set them directly.  No need for a stub when you have the real thing!

The response stub concept for single fetch is needed because those loaders don't return a `Response` - they return data that is merged with other loader data into a Response, so we need to give you some abstraction to handle status/headers.

kiliman:
I think this is a breaking change with the new `Single Fetch` feature.

Before, your resource routes could return a _naked_ object and Remix would automatically convert that to a JSON response. 

Now, if you return a _naked_ object from a resource route, you will get that error message.

It's not technically a Remix error. You should check to see if `unstable_singleFetch` is true and then let the developer know that that resource routes can no longer return a _naked_ object. They can still use the `json` helper, since it creates a new Response for you.

brophdawg11:
[...]

PR #9349, opened 2024-04-30
Resource Route Single Fetch Updates

brophdawg11:
When single fetch is enabled:

* Raw javascript objects returned from resource routes are automatically wrapped in `json()` and a deprecation warning is logged
* Resource routes will receive the `response` stub for consistency and it will be respected, but the preferred way to handle status/headers is via your returned `Response`

Closes https://github.com/remix-run/remix/issues/9314

### DIFF
--- a/.changeset/slow-peaches-matter.md
+++ b/.changeset/slow-peaches-matter.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/server-runtime": patch
+---
+
+Pass `response` stub to resource route handlerså when single fetch is enabled
--- a/docs/guides/single-fetch.md
+++ b/docs/guides/single-fetch.md
@@ -272,14 +272,22 @@ GET /a/b/c.data?_routes=routes/c
 
 ### Resource Routes
 
-Because of the new streaming format used by Single Fetch, raw JavaScript objects returned from `loader` and `action` functions are no longer automatically converted to `Response` instances via the `json()` utility.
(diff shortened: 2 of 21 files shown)


manzano78:
@brophdawg11, does it mean that we can’t benefit from turbo-stream Date/Set/Map/Promise types in a useFetcher which load()s a resource route?

brophdawg11:
@manzano78 No - when a fetcher hits a resource route it will still make a `/resource.data` request and it'll go through the `turbo-stream`-enabled code path.  If you were to CURL a resource route, you'd get this `json()` behavior

manzano78:
@brophdawg11 yes I can reproduce the turbo-stream format but since json() applies a simple `JSON.stringify(data)`, I still receive a string instead of a Date in my fetcher's data :-/

docs/guides/single-fetch.md
```diff
@@ -270,6 +270,74 @@ And then when `c` calls `serverLoader`, it'll make it's own call for just the `c
...
+  return json(data, {
+    status: 200,
+    headers: {
+      "X-Custom": "whatever",
+    },
+  });
+}
+```
+
+To keep things consistent, resource route `loader`/`action` functions will still receive a `response` stub and you can use it if you need to (maybe to share code amongst non-resource-route handlers):
```
haines, line 322: This is a great improvement, thank you! I ran into this after I refactored our `authenticate` helper to take the response stub to append the `Set-Cookie` header, so it's really nice to be able to use that from resource routes as well.

.changeset/slow-peaches-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/server-runtime": patch
+---
+
+Pass `response` stub to resource route handlerså when single fetch is enabled
```
ngbrown, line 5: There is an errant `å` in the changset text.

Issue #9359, opened 2024-05-02
Single fetch: allow setting a `nonce` on `StreamTransfer`'s `script` tags


haines:
### Expected Behavior

We have a CSP in place which requires a nonce on inline scripts, so we need to add that to the `script` tags created [in the `StreamTransfer` component](https://github.com/remix-run/remix/blob/bd6a30601107798d6771cf166417e2034691a9c4/packages/remix-react/single-fetch.tsx#L75-L95).

### Actual Behavior

There is no way afaict to set the nonce for these scripts.

0927777696:
> ### Expected Behavior
> 
> We have a CSP in place which requires a nonce on inline scripts, so we need to add that to the `script` tags created [in the `StreamTransfer` component](https://github.com/remix-run/remix/blob/bd6a30601107798d6771cf166417e2034691a9c4/packages/remix-react/single-fetch.tsx#L75-L95).
> 
> ### Actual Behavior
> 
> There is no way afaict to set the nonce for these scripts.

hi
![IMG-20240503-WA0006.jpg](https://github.com/remix-run/remix/assets/166762251/3f3bc712-b655-477a-91d9-f95038ca338e)

brophdawg11:
This fix is available in `2.9.2-pre.1` if you'd like to give it a shot and see if it resolves your issue!  Here's the draft release notes: https://github.com/remix-run/remix/blob/release-next/CHANGELOG.md#v292

brophdawg11:
`2.9.2` is released - https://github.com/remix-run/remix/releases/tag/remix%402.9.2

PR #9364, opened 2024-05-03
Allow a nonce to be set on single fetch stream transfer inline scripts

haines:
Closes: #9359

- [x] Docs
- [x] Tests

Testing Strategy:

I added a test case to `integration/single-fetch-test.ts`, and I patched `@remix-run/react` in our application with these changes and saw that it fixed our CSP violations.

### DIFF
--- a/.changeset/nasty-vans-brake.md
+++ b/.changeset/nasty-vans-brake.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": minor
+---
+
+Allow a nonce to be set on single fetch stream transfer inline scripts
--- a/.changeset/nasty-vans-brake.md
+++ b/.changeset/nasty-vans-brake.md
@@ -1,5 +1,5 @@
 ---
-"@remix-run/react": minor
+"@remix-run/react": patch
 ---
 
 Allow a nonce to be set on single fetch stream transfer inline scripts
--- a/integration/single-fetch-test.ts
+++ b/integration/single-fetch-test.ts
@@ -2478,7 +2478,7 @@ test.describe("single-fetch", () => {
       files: {
         ...files,
         "app/root.tsx": js`
-          import { Form, Link, Links, Meta, Outlet, Scripts } from "@remix-run/react";
+          import { Links, Meta, Outlet, Scripts } from "@remix-run/react";
 
           export function loader() {
             return {
@@ -2494,14 +2494,6 @@ test.describe("single-fetch", () => {
                   <Links />
                 </head>
                 <body>
-                  <Link to="/">Home</Link><br/>
-                  <Link to="/data">Data</Link><br/>
-                  <Link to="/a/b/c">/a/b/c</Link><br/>
-                  <Form method="post" action="/data">
-                    <button type="submit" name="key" value="value">
-                      Submit
-                    </button>
-                  </Form>
                   <Outlet />
                   <Scripts nonce="the-nonce" />
                 </body>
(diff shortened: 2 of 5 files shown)


brophdawg11:
Thanks!

Yeah this stuff is best tested through our E2E tests - want to try something like this in `integration/single-fetch-test.ts`?

You can run it with `pnpm test:integration single-fetch --project chromium`

```ts
  test("supports nonce on streaming script tags", async ({ page }) => {
    let fixture = await createFixture({
      config: {
        future: {
          unstable_singleFetch: true,
        },
      },
      files: {
        ...files,
        "app/entry.server.tsx": js`
          import { PassThrough } from "node:stream";

          import { createReadableStreamFromReadable } from "@remix-run/node";
          import { renderToPipeableStream } from "react-dom/server";

            request: Request,
            responseStatusCode: number,
            responseHeaders: Headers,
            remixContext: EntryContext
          ) {
            return new Promise((resolve, reject) => {
              const { pipe } = renderToPipeableStream(
                <RemixServer context={remixContext} url={request.url} nonce="the-nonce" />,
                {
                  onShellReady() {
                    const stream = createReadableStreamFromReadable(body);
                    resolve(
                      new Response(stream, {
                        headers: responseHeaders,
                        status: responseStatusCode,
                      })
                    );
                    pipe(body);
                  },
                  onShellError(error: unknown) {
                    reject(error);
                  },
                  onError(error: unknown) {
                    responseStatusCode = 500;
                  },
                }
              );
            });
          }
        `,
      },
    });
    let appFixture = await createAppFixture(fixture);
    await app.goto("/data", true);
    let scripts = await page.$$("script");
    expect(scripts.length).toBe(6);
    for (let s of scripts) {
      // Do nonce expect() calls here for the scripts that should have it - which I think might just be
      // any with `window.__remixContext.streamController` in the innterHTML?
      console.log(await s.getAttribute("nonce"), await s.innerHTML());
    }
  });
```

brophdawg11 (approved):
Thanks!

.changeset/nasty-vans-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/react": minor
```
brophdawg11, line 2: We can do patch here since it's part of an unstable feature and it'll be a minor bump once we stabilize the feature

```suggestion
"@remix-run/react": patch
```

integration/single-fetch-test.ts
```diff
@@ -2467,4 +2467,104 @@ test.describe("single-fetch", () => {
...
+                </head>
+                <body>
+                  <Link to="/">Home</Link><br/>
+                  <Link to="/data">Data</Link><br/>
+                  <Link to="/a/b/c">/a/b/c</Link><br/>
+                  <Form method="post" action="/data">
+                    <button type="submit" name="key" value="value">
+                      Submit
+                    </button>
+                  </Form>
```
brophdawg11, line 2504: No need to carry these over from the base root file

```suggestion
```

integration/single-fetch-test.ts
```diff
@@ -2467,4 +2467,104 @@ test.describe("single-fetch", () => {
...
+    let fixture = await createFixture({
+      config: {
+        future: {
+          unstable_singleFetch: true,
+        },
+      },
+      files: {
+        ...files,
+        "app/root.tsx": js`
+          import { Form, Link, Links, Meta, Outlet, Scripts } from "@remix-run/react";
```
brophdawg11, line 2481:
```suggestion
          import { Links, Meta, Outlet, Scripts } from "@remix-run/react";
```

PR #9372, opened 2024-05-03
Single-fetch typesafety: `defineLoader` and `defineAction`

pcattori:
```ts
// turbo-stream for bare objects
let loader1 = () => ({a: 1})
useLoaderData<typeof loader1>(); // { a: number }

// Jsonify for `json`/`defer`
let loader2 = () => json({a: 1})
useLoaderData<typeof loader2> // JsonifyObject<{a: number}>

// typesafety for args and return type via `defineLoader`
let loader3 = defineLoader(({ request }) => {
  //                          ^? Request
  return {a: 1, b: () => 2}
  //            ^ `b` being a function causes a type error for `defineLoader`
})
```

See changeset for more detail


packages/remix-react/future/single-fetch.d.ts
```diff
@@ -11,90 +11,52 @@ import type {
...
+type Serialize<T extends Loader | Action> =
   Awaited<ReturnType<T>> extends TypedDeferredData<infer D> ? D :
   Awaited<ReturnType<T>> extends TypedResponse<Record<string, unknown>> ? SerializeFrom<T> :
   Awaited<ReturnType<T>>;
 
 declare module "@remix-run/react" {
-  export function useLoaderData<T>(): T extends LoaderFunction_SingleFetch
-    ? SingleFetchSerialize_V2<T>
-    : never;
+  export function useLoaderData<T>(): T extends Loader ? Serialize<T> : T;
```
pcattori, line 25: @brophdawg11 we currently skip serialization logic if the user is passing in a custom type for `useLoaderData`:

```ts
type MyCustomType = { a: number }
let data = useLoaderData<MyCustomType>() // basically a type cast 😕
```

I'm thinking it might be a good opportunity to drop that with single-fetch since we already don't want to support it in v3. That would mean

```suggestion
   export function useLoaderData<T extends Loader>(): Serialize<T>;
```

packages/remix-server-runtime/single-fetch.ts
```diff
@@ -504,3 +507,51 @@ export function encodeViaTurboStream(
...
+  context: AppLoadContext;
+  response: ResponseStub;
+};
+
+export type Action = (
+  args: ActionArgs
+) => MaybePromise<DataFunctionReturnValue>;
+
+export let defineLoader = <T extends Loader>(loader: T): T => loader;
+export let defineAction = <T extends Action>(action: T): T => action;
```
pcattori, line 557: Should any user-facing exports be prefixed with `unstable_`? I think probably yes

packages/remix-react/future/single-fetch.d.ts
brophdawg11: yeah I agree - I think that was mostly just for back-compat with some early version of `useLoaderData`?

packages/remix-server-runtime/single-fetch.ts
brophdawg11: yeah that's what I've been trying to do and we can stabilize them all together

packages/remix-server-runtime/single-fetch.ts
```diff
@@ -504,3 +508,76 @@ export function encodeViaTurboStream(
...
+export type Loader = (
+  args: LoaderArgs
+) => MaybePromise<DataFunctionReturnValue>;
+export let defineLoader = <T extends Loader>(loader: T): T => loader;
+
+// clientLoader
+type ClientLoaderArgs = RRLoaderArgs<undefined> & {
+  serverLoader: <T extends Loader>() => Promise<Serialize<T>>;
+};
+type ClientLoader = (args: ClientLoaderArgs) => MaybePromise<Serializable>;
```
brophdawg11, line 562: We also need to support the `clientLoader.hydrate = true` property - here's what we have in the base types:

```ts
type ClientLoaderFunction = ((
  args: ClientLoaderFunctionArgs
) => ReturnType<RRLoaderFunction>) & {
  hydrate?: boolean;
};
```

brophdawg11: This seemed to be what got it working for me - I needed it in both of those places it seems

```ts
type ClientLoader = ((args: ClientLoaderArgs) => MaybePromise<Serializable>) & {
  hydrate?: boolean;
};
export let defineClientLoader = <T extends ClientLoader>(
  clientLoader: T
): T & { hydrate?: boolean } => clientLoader;
```

brophdawg11: yeah that crossed my mind too - let's discuss with Ryan if he wants to add that in if they're already making this change now.  We should land on our final preferred API in v2 so that when we stabilize the `v3_singleFetch` flag it's no code changes to v7

docs/guides/single-fetch.md
```diff
@@ -98,13 +98,66 @@ In order to ensure you get the proper types when using Single Fetch, we've inclu
...
+export const action = defineAction(() => {
+  return { hello: "world", badData: new CustomType() };
+  //                       ^^^^^^^ Type error: `badData` is not serializable
+});
+```
+
+Single-fetch supports the following return types:
+
+```ts
+type Serializable =
```
sergiodxa, line 142: Is this type exported somewhere?

packages/remix-server-runtime/single-fetch.ts
sergiodxa: I like @pcattori idea to pass the hydrate option directly to defineClientLoader, since it's already calling a function it makes sense to just pass the option.

docs/guides/single-fetch.md
```diff
@@ -98,13 +98,66 @@ In order to ensure you get the proper types when using Single Fetch, we've inclu
...
+  | undefined
+  | null
+  | boolean
+  | string
+  | symbol
+  | number
+  | bigint
+  | Date
+  | URL
```
sergiodxa, line 151: Could URLSearchParams be added to this too? I remember someone reporting in Discord that they tried to return this and it failed.

docs/guides/single-fetch.md
```diff
@@ -98,13 +98,66 @@ In order to ensure you get the proper types when using Single Fetch, we've inclu
...
+Not only does this give you types for any arguments, but it also ensures you are returning single-fetch compatible types:
+
+```ts
+export const loader = defineLoader(() => {
+  return { hello: "world", badData: () => 1 };
+  //                       ^^^^^^^ Type error: `badData` is not serializable
+});
+
+export const action = defineAction(() => {
+  return { hello: "world", badData: new CustomType() };
```
sergiodxa, line 134: What if CustomType class has a toJSON?

pcattori: The contract for single-fetch (and `turbo-stream` which is what it is using under-the-hood) is that a type `T` in becomes a type `T` out. So `CustomType` -> `string` is not supported. If you want you can manually call `toJSON()`:

```ts
export let loader = () => {
  return { a: myCustomTypeInstance.toJSON() }
}
```

The point is to have less coercion and magic. Having the serialization be basically a glorified ID function is the thing that unlocks a bunch of the simplified types features. Plus it means we can support recursive types much better than the current approach.

docs/guides/single-fetch.md
pcattori: We're using Jacob's [turbo-stream](https://github.com/jacob-ebey/turbo-stream) library under-the-hood so support for `URLSearchParams` would need to be added there.

docs/guides/single-fetch.md
gustavopch: @sergiodxa FIY, I raised the same question on Discord: https://discord.com/channels/770287896669978684/1233404747885576293/1233417203693523104

pcattori:
@brophdawg11 looks like there are still a couple places in the codebase that rely on `response` in `LoaderFunctionsArgs` and `ActionFunctionArgs`, so left that in for now.

pcattori:
> Also - if you don't use `defineLoader` should we still be trying to infer single fetch types? Or are we saying the two are mutually exclusive?

Bare objects are always single-fetch serialized after you enabled single-fetch future flag + types, as intended:

docs/guides/single-fetch.md
brophdawg11: Cross posting my response from discord as well for visibility

> We're hesitant to add to much "custom" stuff like this to turbo-stream because it's a stopgap solution that will be replaced with the RSC wire streaming format once we land the RSC work.  So the limitations of what can and can't be streamed are going to eventually be defined by React, not us

I.e., if we add `toJson` support now, folks will start relying on it.  Then if we incorporate RSC and it doens't support `toJSON`, it would be a breaking change for your apps.  So we're keeping this simpler now until we know in greater detail how the RSC story shakes out.

ziimakc:
@pcattori

Why do I need to install `react-router` as dependency to use `defineLoader`?
```
The inferred type of 'loader' cannot be named without a reference to '.pnpm/react-router@6.23.1_react@18.3.1/node_modules/react-router'. This is likely not portable. A type annotation is necessary.
```

PR #9378, opened 2024-05-06
docs: made example in `Route File Nmaing` stricter

soartec-lab:
First of all, thank you for maintaining this project. I'm a fan of this.

There was a difference between the file path and the actual routing in the document `Route File Naming`.
This can be confusing for readers, so we've updated all routes to match file paths.


brophdawg11:
Thanks!

PR #9395, opened 2024-05-09
Clean redundant `?client-route=1` usage in dev

markdalgleish:
This PR is aiming to address two things:
1) Support the move to vite-plugin-react-swc: https://github.com/remix-run/remix/pull/9092
2) Potentially unlock this fix in Hydrogen: https://github.com/Shopify/hydrogen/pull/2078

The `?client-route=1` virtual modules that wrap all client routes are only necessary for a build optimisation that removes unused exports from route files on the client. The only reason these virtual modules were used in development was to keep the dev and build code paths as similar as possible. However, since it's causing other issues, and it also necessitated _extra_ code in dev to special-case HMR for these files, we're opting to remove `?client-route=1` usage in dev entirely.

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -130,9 +130,11 @@ const CLIENT_ROUTE_EXPORTS = [
   "shouldRevalidate",
 ];
 
-// The "=1" suffix ensures client route requests can be processed before hitting
-// the Vite plugin since "?client-route" can be serialized as "?client-route="
-const CLIENT_ROUTE_QUERY_STRING = "?client-route=1";
+/** This is used to manage a build optimization to remove unused route exports
+from the client build output. This is important in cases where custom route
+exports are only ever used on the server. Without this optimization we can't
+tree-shake any unused custom exports because routes are entry points. */
+const BUILD_CLIENT_ROUTE_QUERY_STRING = "?__remix-build-client-route";
 
 // Only expose a subset of route properties to the "serverBundles" function
 const branchRouteProperties = [
@@ -318,10 +320,6 @@ const getHash = (source: BinaryLike, maxLength?: number): string => {
   return typeof maxLength === "number" ? hash.slice(0, maxLength) : hash;
 };
 
-const isClientRoute = (id: string): boolean => {
-  return id.endsWith(CLIENT_ROUTE_QUERY_STRING);
-};
-
 const resolveChunk = (
   ctx: RemixPluginContext,
   viteManifest: Vite.Manifest,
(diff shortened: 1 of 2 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -1271,8 +1269,8 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
           cssModulesManifest[id] = code;
         }
 
-        if (isClientRoute(id)) {
-          let routeModuleId = id.replace(CLIENT_ROUTE_QUERY_STRING, "");
+        if (id.endsWith(BUILD_CLIENT_ROUTE_QUERY_STRING)) {
```
markdalgleish, line 1272: Inlined `isClientRoute` here since it's now the only usage.

packages/remix-dev/vite/plugin.ts
```diff
@@ -1695,10 +1693,6 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
         let useFastRefresh = !ssr && (isJSX || code.includes(devRuntime));
         if (!useFastRefresh) return;
 
-        if (isClientRoute(id)) {
-          return { code: addRefreshWrapper(ctx.remixConfig, code, id) };
-        }
-
```
markdalgleish, line 1701: This code path is only used in dev, so no longer needed.

packages/remix-dev/vite/plugin.ts
```diff
@@ -1789,17 +1783,16 @@ function addRefreshWrapper(
   id: string
 ): string {
   let route = getRoute(remixConfig, id);
-  let acceptExports =
-    route || isClientRoute(id)
```
markdalgleish, line 1793: HMR is no longer needed for these virtual modules.

packages/remix-dev/vite/plugin.ts
```diff
@@ -130,9 +130,11 @@ const CLIENT_ROUTE_EXPORTS = [
...
 ];
 
-// The "=1" suffix ensures client route requests can be processed before hitting
-// the Vite plugin since "?client-route" can be serialized as "?client-route="
-const CLIENT_ROUTE_QUERY_STRING = "?client-route=1";
+/** This is used to manage a build optimization to remove unused route exports
+from the client build output. This is important in cases where custom route
+exports are only ever used on the server. Without this optimization we can't
+tree-shake any unused custom exports because routes are entry points. */
+const BUILD_CLIENT_ROUTE_QUERY_STRING = "?__remix-build-client-route";
```
markdalgleish, line 137: Since this query string is no longer visible in the browser's network tab, I'm not trying to make this look pretty anymore and instead I'm aiming for it to be more explicit.

markdalgleish: Also the `=1` suffix is no longer needed since that was to solve an issue when proxying during development. In the production builds this query string never goes over the network.

pcattori (approved):

PR #9398, opened 2024-05-09
Remove response stub from LoaderFunctionArgs/ActionFunctionArgs

brophdawg11:
Minor follow ups to https://github.com/remix-run/remix/pull/9372

### DIFF
--- a/packages/remix-server-runtime/data.ts
+++ b/packages/remix-server-runtime/data.ts
@@ -10,8 +10,8 @@ import type {
   ActionFunctionArgs,
   LoaderFunction,
   LoaderFunctionArgs,
-  ResponseStub,
 } from "./routeModules";
+import type { ResponseStub } from "./single-fetch";
 
 /**
  * An object of unknown type for route loaders and actions provided by the
@@ -49,7 +49,9 @@ export async function callRouteAction({
     request: stripDataParam(stripIndexParam(request)),
     context: loadContext,
     params,
-    response,
+    // Only provided when single fetch is enabled, and made available via
+    // `defineAction` types, not `ActionFunctionArgs`
+    ...(singleFetch ? { response } : null),
   });
 
   if (result === undefined) {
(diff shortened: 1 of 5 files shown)


packages/remix-server-runtime/data.ts
```diff
@@ -49,7 +49,9 @@ export async function callRouteAction({
     request: stripDataParam(stripIndexParam(request)),
     context: loadContext,
     params,
-    response,
+    // Only provided when single fetch is enabled, and made available via
+    // `defineAction` types, not `ActionFunctionArgs`
+    ...(singleFetch ? { response } : null),
```
brophdawg11, line 54: This is functionally equivalent, as `response` is only defined when `singleFetch` is enabled, but this quiets the type error when response is not in `LoaderFunctionArgs`

packages/remix-server-runtime/routeModules.ts
```diff
@@ -59,8 +40,6 @@ export type ActionFunction = (
 export type ActionFunctionArgs = RRActionFunctionArgs<AppLoadContext> & {
   // Context is always provided in Remix, and typed for module augmentation support.
   context: AppLoadContext;
-  // TODO: (v7) Make this non-optional once single-fetch is the default
-  response?: ResponseStub;
```
brophdawg11, line 63: Remove the previous ambiguity here.  Now users will get no `response` stub when using `LoaderFunctionArgs` and a non-optional `response` stub when using `defineLoader`

packages/remix-server-runtime/routeModules.ts
```diff
@@ -27,25 +27,6 @@ export type DataFunctionArgs = RRActionFunctionArgs<AppLoadContext> &
     context: AppLoadContext;
   };
 
-export const ResponseStubOperationsSymbol = Symbol("ResponseStubOperations");
```
brophdawg11, line 30: Moved alongside the rest of the single fetch types

Issue #9444, opened 2024-05-16
docs - createCookieSessionStorage says commitSession not needed


penx:
The docs for `createCookieSessionStorage` state:

> With other session storage strategies you only have to commit it when it's created

https://remix.run/docs/en/main/utils/sessions#createcookiesessionstorage

When we moved from `createCookieSessionStorage` to another storage solution (vercel kv), we followed this advice and removed calls to `commitSession` (assuming it was not needed, and that `session.set` would be all that is required) but it resulted in sessions not being saved.

Elsewhere in the docs for `createSessionStorage`:

> updateData will be called from commitSession

https://remix.run/docs/en/main/utils/sessions#createsessionstorage

Which seems to imply that anything based on `createSessionStorage` will require a call to `commitSession` in order to update the session, and I'm fairly sure explains why we would have bugs when removing calls to `commitSession`.

I think the docs for `createCookieSessionStorage` should instead state

> With other session storage strategies you only have to _send an updated cookie in the request header_ when it's created

I'll raise a PR to the docs for this.

kiliman:
Yes, the documentation needs to be corrected. For the session API, all session storage providers require you to call `commitSession` after mutation.

I think what the documentation was alluding to was you don't have to return a `set-cookie` header in the response after every commit, just the initial one.

In cookie session storage, the session data is stored entirely in the cookie, so any changes require you to send an updated cookie via the `set-cookie` header.

However, in other storage providers like Cloudflare KV or file-based sessions, the only thing stored in the cookie is the session ID. You would send that id in the `set-cookie` header on creation. The session ID is used to look up the session in the storage. Since the session id never changes, you don't need to keep sending a `set-cookie` header.

akomm:
Might be related with this issue?
https://github.com/remix-run/remix/discussions/9338

brophdawg11:
Thank you for opening this issue, and our apologies we haven't gotten around to it yet!

With the release of [React Router v7](https://remix.run/blog/react-router-v7) we are sun-setting continued development/maintenance on Remix v2. If you have not already [upgraded to React Router v7](https://reactrouter.com/upgrading/remix), we recommend you do so. We've tried to make the upgrade process as smooth as possible with our [Future Flags](https://remix.run/docs/en/main/start/future-flags). We are now in the process of cleaning up outdated issues and pull requests to improve the overall hygiene of our repositories.

We plan to continue to address 2 types of issues in Remix v2:

- Bugs that pose security concerns
[...]

PR #9445, opened 2024-05-16
docs(utils/sessions): Correction about `commitSession` in non-cookie sessions

penx:
Closes: #9444

- [x] Docs
- n/a Tests 

The docs for `createCookieSessionStorage` state:

> With other session storage strategies you only have to commit it when it's created

https://remix.run/docs/en/main/utils/sessions#createcookiesessionstorage

When we moved from `createCookieSessionStorage` to another storage solution (`createKvSessionStorage` from `@vercel/remix`), we followed this advice and removed calls to `commitSession` (assuming it was not needed, and that `session.set` would be all that is required) but it resulted in sessions not being saved.

Elsewhere in the docs for `createSessionStorage`:

> updateData will be called from commitSession

https://remix.run/docs/en/main/utils/sessions#createsessionstorage

Which seems to imply that anything based on `createSessionStorage` will require a call to `commitSession` in order to update the session, and I'm fairly sure explains why we would have bugs when removing calls to `commitSession`.

### DIFF
--- a/docs/utils/sessions.md
+++ b/docs/utils/sessions.md
@@ -277,7 +277,7 @@ For purely cookie-based sessions (where the session data itself is stored in the
 
 The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios. However, cookie-based sessions may not exceed the browser's max-allowed cookie length (typically 4kb).
 
-The downside is that you have to `commitSession` in almost every loader and action. If your loader or action changes the session at all, it must be committed. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away. With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
+The downside is that you have to send a "Set-Cookie" header from almost every loader and action, this can cause complications when writing to simultaneous requests for the same session. With other session storage strategies you only have to send a "Set-Cookie" header it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). Note that you still need to call `commitSession()` whenever you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
 
 ```ts
 import { createCookieSessionStorage } from "@remix-run/node"; // or cloudflare/deno
--- a/docs/utils/sessions.md
+++ b/docs/utils/sessions.md
@@ -277,7 +277,7 @@ For purely cookie-based sessions (where the session data itself is stored in the
 
 The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios. However, cookie-based sessions may not exceed the browser's max-allowed cookie length (typically 4kb).
 
-The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That means, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
+The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. This means, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
 
 This can cause complications if loaders or actions are writing to the same session at the same time.
 
--- a/docs/utils/sessions.md
+++ b/docs/utils/sessions.md
@@ -277,7 +277,13 @@ For purely cookie-based sessions (where the session data itself is stored in the
 
 The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios. However, cookie-based sessions may not exceed the browser's max-allowed cookie length (typically 4kb).
 
-The downside is that you have to send a "Set-Cookie" header from almost every loader and action, this can cause complications when writing to simultaneous requests for the same session. With other session storage strategies you only have to send a "Set-Cookie" header when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). Note that you still need to call `commitSession()` whenever you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
+The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That mean, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
+
+This can cause complications when handling simultaneous requests for the same session. 
+
+With other session storage strategies you only have to send a "Set-Cookie" header when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). 
+
+Note that you still need to call `commitSession()` whenever you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
 
 ```ts
 import { createCookieSessionStorage } from "@remix-run/node"; // or cloudflare/deno
--- a/docs/utils/sessions.md
+++ b/docs/utils/sessions.md
@@ -277,13 +277,13 @@ For purely cookie-based sessions (where the session data itself is stored in the
 
 The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios. However, cookie-based sessions may not exceed the browser's max-allowed cookie length (typically 4kb).
 
-The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That mean, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
+The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That means, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
 
-This can cause complications when handling simultaneous requests for the same session. 
+This can cause complications if loaders or actions are writing to the same session at the same time.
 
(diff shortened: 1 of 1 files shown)


docs/utils/sessions.md
```diff
@@ -277,7 +277,7 @@ For purely cookie-based sessions (where the session data itself is stored in the
...
 
-The downside is that you have to `commitSession` in almost every loader and action. If your loader or action changes the session at all, it must be committed. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away. With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
+The downside is that you have to send a "Set-Cookie" header from almost every loader and action, this can cause complications when writing to simultaneous requests for the same session. With other session storage strategies you only have to send a "Set-Cookie" header it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). Note that you still need to call `commitSession()` whenever you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
```
penx, line 280:
```suggestion
The downside is that you have to send a "Set-Cookie" header from almost every loader and action, this can cause complications when writing to simultaneous requests for the same session. With other session storage strategies you only have to send a "Set-Cookie" header when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). Note that you still need to call `commitSession()` whenever you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
```

docs/utils/sessions.md
```diff
@@ -277,7 +277,7 @@ For purely cookie-based sessions (where the session data itself is stored in the
...
 
-The downside is that you have to `commitSession` in almost every loader and action. If your loader or action changes the session at all, it must be committed. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away. With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
+The downside is that you have to send a "Set-Cookie" header from almost every loader and action, this can cause complications when writing to simultaneous requests for the same session. With other session storage strategies you only have to send a "Set-Cookie" header when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). Note that you still need to call `commitSession()` whenever you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
```
machour, line 280: Made a mix of old & new version, and added some line feeds to make it easier to read. What do you think?

```suggestion
The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.

This can cause complications when handling simultaneous requests for the same session. 

With other session storage strategies you only have to send a "Set-Cookie" header when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). 

Note that you still need to call `commitSession()` whenever you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
```

penx: Great thanks

“that mean” should probably be “that means”

docs/utils/sessions.md
```diff
@@ -277,7 +277,13 @@ For purely cookie-based sessions (where the session data itself is stored in the
 
 The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios. However, cookie-based sessions may not exceed the browser's max-allowed cookie length (typically 4kb).
 
-The downside is that you have to `commitSession` in almost every loader and action. If your loader or action changes the session at all, it must be committed. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away. With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
+The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That mean, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
```
penx, line 280:
```suggestion
The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That means, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
```

docs/utils/sessions.md
```diff
@@ -277,7 +277,13 @@ For purely cookie-based sessions (where the session data itself is stored in the
...
+
+This can cause complications when handling simultaneous requests for the same session.
```
penx, line 282:
```suggestion
This can cause complications if loaders or actions are writing to the same session at the same time.
```

docs/utils/sessions.md
```diff
@@ -277,7 +277,13 @@ For purely cookie-based sessions (where the session data itself is stored in the
...
 
-The downside is that you have to `commitSession` in almost every loader and action. If your loader or action changes the session at all, it must be committed. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away. With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
+The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That mean, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
+
+This can cause complications when handling simultaneous requests for the same session. 
+
+With other session storage strategies you only have to send a "Set-Cookie" header when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
```
penx, line 284:
```suggestion
With other session storage strategies you only have to send a "Set-Cookie" header when the session is created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). 
```

docs/utils/sessions.md
```diff
@@ -277,7 +277,13 @@ For purely cookie-based sessions (where the session data itself is stored in the
...
+The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That mean, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
+
+This can cause complications when handling simultaneous requests for the same session. 
+
+With other session storage strategies you only have to send a "Set-Cookie" header when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). 
+
+Note that you still need to call `commitSession()` whenever you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
```
penx, line 286:
```suggestion
Note that you still need to call `commitSession()` when you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
```

docs/utils/sessions.md
```diff
@@ -277,7 +277,13 @@ For purely cookie-based sessions (where the session data itself is stored in the
 
 The main advantage of cookie session storage is that you don't need any additional backend services or databases to use it. It can also be beneficial in some load-balanced scenarios. However, cookie-based sessions may not exceed the browser's max-allowed cookie length (typically 4kb).
 
-The downside is that you have to `commitSession` in almost every loader and action. If your loader or action changes the session at all, it must be committed. That means if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away. With other session storage strategies you only have to commit it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
+The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. That means, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
```
penx, line 280:
```suggestion
The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. This means, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
```

machour (approved):

penx:
@brophdawg11 anything needed to get this merged? 🙏

docs/utils/sessions.md
```diff
@@ -277,7 +277,13 @@ For purely cookie-based sessions (where the session data itself is stored in the
...
+The downside is that you have to `commitSession` and send a "Set-Cookie" header from every loader and action that changes the session. This means, for example, that if you `session.flash` in an action, and then `session.get` in another, you must commit it for that flashed message to go away.
+
+This can cause complications if loaders or actions are writing to the same session at the same time.
+
+With other session storage strategies you only have to send a "Set-Cookie" header when the session is created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere). 
+
+Note that you still need to call `commitSession()` when you change the session for anything based on `createSessionStorage`, you just don't need to send an updated header.
```
brophdawg11, line 286: I think the confusion in the original docs is that the phrase "commit" is being used when it really means "set header".  So we mean to say "with others you don't need to set the header every time" but we instead incorrectly imply that with others scenarios "you don't need to commitSession every time".

What about this slight update to the original wording?  Instead of re-iterating that you need to `commitSession` after changes (which is only needed because of the incorrect implication that you don't) - we can just remove the misleading implication otherwise and indicate the true difference which relates to the `Set-Cookie` header:

```suggestion
The downside is that you have to update the cookie via a `Set-Cookie` header in almost every loader and action. If your loader or action changes the session at all, it must be sent back as an updated cookie. That means if you `session.flash` in an action, and then `session.get` in another, you must update it for that flashed message to go away. With other session storage strategies you only have to send the `Set-Cookie` header it when it's created (the browser cookie doesn't need to change because it doesn't store the session data, just the key to find it elsewhere).
```

brophdawg11:
this has conflicts likely from the other large update to sessions docs - is this still relevant given those updates?

PR #9478, opened 2024-05-22
Fix potential github action smells

ceddy4395:
Hey! 🙂
I've made the following changes to your workflow: 

- Avoid jobs without timeouts
  - Jobs without a timeout can cause runners to be occupied when some process falls into an infinite loop or has multiple retries to perform a certain task, which will inevitably fail. Furthermore, it is also useful to be notified when a test-suite all of a sudden takes a lot longer than it used to, considering keeping tests fast is a good practice. 
- Use commit hash instead of tags for action versions
  - When using a tag as version, the code related to the tag can be changed after the tag is created, whereas when using the commit hash this cannot. Therefore, for consistency and security a commit hash should be used.
- Define permissions for workflows with external actions
  - Permissions should be used when running actions written by other developers because there may be security leaks exposed through these actions.   

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,11 @@ on:
       - "!release-manual"
       - "!release-manual-*"
 
+permissions: 
+  contents: write
+  pull-requests: write
+  
+
 jobs:
   release:
     name: 🦋 Changesets Release
@@ -20,20 +25,21 @@ jobs:
       github.repository == 'remix-run/remix' &&
       !contains(github.ref, 'nightly')
     runs-on: ubuntu-latest
+    timeout-minutes: 2
     outputs:
       published_packages: ${{ steps.changesets.outputs.publishedPackages }}
       published: ${{ steps.changesets.outputs.published }}
     steps:
       - name: ⬇️ Checkout repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4
         with:
           fetch-depth: 0
 
       - name: 📦 Setup pnpm
-        uses: pnpm/action-setup@v3.0.0
+        uses: pnpm/action-setup@a3252b78c470c02df07e9d59298aecedc3ccdd6d # v3.0.0
 
       - name: ⎔ Setup node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4
         with:
           node-version-file: ".nvmrc"
           cache: "pnpm"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -20,20 +25,21 @@ jobs:
...
       !contains(github.ref, 'nightly')
     runs-on: ubuntu-latest
+    timeout-minutes: 2
     outputs:
       published_packages: ${{ steps.changesets.outputs.publishedPackages }}
       published: ${{ steps.changesets.outputs.published }}
     steps:
       - name: ⬇️ Checkout repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4
```
alcpereira, line 34: This action is maintained by Github, there should not be any security/breaking changes issues. By pinning to a specific commit hash, you're limiting yourself from potential improvements to the current major version.

This is the same for `pnpm` and `changesets`, both are maintained by their respective organization.

.github/workflows/release.yml
```diff
@@ -20,20 +25,21 @@ jobs:
       github.repository == 'remix-run/remix' &&
       !contains(github.ref, 'nightly')
     runs-on: ubuntu-latest
+    timeout-minutes: 2
```
alcpereira, line 28: For this job, I don't think 2 minutes is enough, but core maintainers can confirm.
Usually releases are carefully monitored, so I don't think there is a real use case to this limit.

.github/workflows/release.yml
brophdawg11: I agree, I think we're fine keeping the numerical versions for these jobs

.github/workflows/release.yml
```diff
@@ -9,6 +9,11 @@ on:
       - "!release-manual"
       - "!release-manual-*"
 
+permissions: 
+  contents: write
+  pull-requests: write
```
brophdawg11, line 14: Do you have links to what these are doing versus what the default permissions are?  Is this more restrictive?  Or just more explicit?

.github/workflows/release.yml
brophdawg11: I agree - a quick peek at past runs shows we've had a few jobs in the 3-5 minute range so I wouldn't go lower than 10-15m here if it's just intended to catch true runaway/hung jobs.  Anything under that could result in false positives if github's infra is slowed down.

Issue #9471, opened 2024-05-20
Single fetch: Throwing a response stub with a 4xx status causes a turbo-stream error 


brophdawg11:
### Reproduction

Thrown error response stubs are properly handled on data requests but not on document requests - see https://github.com/remix-run/remix/pull/9454 and https://github.com/remix-run/remix/pull/9479

### System Info

```shell
n/a
```

### Used Package Manager

npm

### Expected Behavior

Thrown response stubs hydrate up as `ErrorResponse` instances for use with `useRouteError`/`isRouteErrorResponse`

### Actual Behavior

`turbo-stream` tries to serialize the `ReponseStub` instance and errors

haines:
We ran into a similar case with a redirect response stub in a resource route: https://github.com/remix-run/remix/pull/9479

PR #9479, opened 2024-05-22
Bug report: Single fetch: Throwing a response stub with a 3xx status from a resource route results in an unexpected server error

haines:
We had a couple of routes that were only there to redirect to another page in the application, and I forgot to put a default export in the modules.

We discovered that for these routes, the redirect works for data requests, but not for document requests.

I'm guessing this may just be the same problem as in https://github.com/remix-run/remix/issues/9471, but it manifests differently (presumably because without a default export it's not treated as a UI route). Instead of a turbo-stream error, the response stub propagates all the way to `returnLastResortErrorResponse` and we get the plaintext `500 Unexpected Server Error` page.

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -1,6 +1,5 @@
 import { test, expect } from "@playwright/test";
 
-import { PlaywrightFixture } from "./helpers/playwright-fixture.js";
 import type { Fixture, AppFixture } from "./helpers/create-fixture.js";
 import {
   createAppFixture,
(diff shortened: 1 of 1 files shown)


brophdawg11:
Thank you for the bug report!  This is handled by https://github.com/remix-run/remix/pull/9488

brophdawg11:
And also has some additional test cases added in https://github.com/remix-run/remix/pull/9501

jam-fran:
I'm using 2.10.2 and am still experiencing this bug. Here I had my index route that is intended to be a pure redirect: 

```
export const loader = defineLoader(async ({ response }) => {
  response.status = 302
  response.headers.set('Location', '/account')
  return null
})
```

When this was the only export in my route file, I got an Internal Server Error on document requests.

When I added an empty default export:

`export default function IndexRoute() {}`

It worked as expected.

brophdawg11:
I cannot reproduce that in a simple reproduction: https://stackblitz.com/edit/remix-run-remix-bydt2f

Could you open a new issue with a minimal reproduction of the issue?

PR #9480, opened 2024-05-22
chore: pin wrangler, update cf packages, add .dev.vars to gitignore

brookslybrand:
We've had a number of issues in the past with wrangler updating and breaking things. Going to resume pinning the version and update some docs while I'm here

Closes: #8288

### DIFF
--- a/templates/cloudflare/package.json
+++ b/templates/cloudflare/package.json
@@ -21,7 +21,7 @@
     "react-dom": "^18.2.0"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^4.20240405.0",
+    "@cloudflare/workers-types": "^4.20240512.0",
     "@remix-run/dev": "*",
     "@types/react": "^18.2.20",
     "@types/react-dom": "^18.2.7",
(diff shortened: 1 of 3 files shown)


templates/cloudflare/package.json
```diff
@@ -16,12 +16,11 @@
     "@remix-run/cloudflare-pages": "*",
     "@remix-run/react": "*",
     "isbot": "^4.1.0",
-    "miniflare": "^3.20240404.0",
```
brookslybrand, line 19: @pcattori was there a reason miniflare was included? It doesn't seem like it was doing anything, and removing it didn't break `dev`, `start`, or `prod`

pcattori: 🤷 don't remember, there was a lot of churn on the CF side when we were iterating towards the CF Vite stuff with them

brookslybrand: Cool, I'll just remove it since it worked fine without it locally for me

If I did something wrong people will yell at me and I can ask them to make our Cloudflare templates better. I mostly just want them to consistently work and be pretty minimal at the moment

PR #9481, opened 2024-05-22
chore: turn on all the future flags for vite templates

brookslybrand:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


templates/remix-javascript/vite.config.js
```diff
@@ -5,5 +5,13 @@ import { defineConfig } from "vite";
 installGlobals();
```
brophdawg11, line 5: Do we want `nativeFetch:true` here too?

templates/remix/vite.config.ts
```diff
@@ -6,5 +6,14 @@ import tsconfigPaths from "vite-tsconfig-paths";
 installGlobals();
```
brophdawg11, line 6: here too?

brophdawg11 (approved):

templates/remix-javascript/vite.config.js
brookslybrand: 🤦 

Yep! Thanks for pointing out what I already told you I'd add 😅

brookslybrand: Actually, I'm just going to bump all the templates to node 20 and remove `installGlobals` since we are going to deprecate it once node 18 is no longer LTS

Thanks!

PR #9488, opened 2024-05-23
Single Fetch: fix issues returning/throwing response stubs from resource routes

brophdawg11:


packages/remix-server-runtime/server.ts
```diff
@@ -591,7 +591,7 @@ async function handleResourceRequest(
         : null),
     });
 
-    if (typeof response === "object") {
+    if (typeof response === "object" && response !== null) {
       invariant(
         !(DEFERRED_SYMBOL in response),
```
brophdawg11, line 596: Avoids `cannot search for property in null` error

jacob-ebey (approved):

PR #9512, opened 2024-05-29
docs: update flow and language of single-fetch.md

brookslybrand:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/docs/guides/single-fetch.md
+++ b/docs/guides/single-fetch.md
@@ -4,43 +4,40 @@ title: Single Fetch
 
 # Single Fetch
 
-Remix introduced support for "Single Fetch" ([RFC][rfc]) behind the [`future.unstable_singleFetch`][future-flags] flag in [`v2.9.0`][2.9.0] which allows you to opt-into this behavior which will become the default in Remix v3.
+Single fetch is a new data data loading strategy and streaming format. When you enable Single Fetch, Remix will make a single HTTP call to your server on client-side transitions, instead of multiple HTTP calls in parallel (one per loader). Additionally, Single Fetch also allows you to send down naked objects from your `loader` and `action`, such as `Date`, `Error`, `Promise`, `RegExp`, and more.
 
-## Overview
+## Enabling Single Fetch
 
-When you enable Single Fetch, Remix will make a single HTTP call to your server on client-side transitions, instead of multiple HTTP calls in parallel (one per loader). If you are currently returning `Response` instances from your loaders (i.e., `json`/`defer`) then you shouldn't _need_ to make many changes to your app code, but please read through the "breaking" changes below to be aware of some of the underlying behavior changes - specifically around serialization and status/header behavior.
+Remix introduced support for "Single Fetch" ([RFC][rfc]) behind the [`future.unstable_singleFetch`][future-flags] flag in [`v2.9.0`][2.9.0] which allows you to opt-into this behavior. Single Fetch will be the default in [React Router v7][merging-remix-and-rr].
 
-### Breaking Changes
+Enabling Single Fetch is incredibly simple. If you are currently returning `Response` instances from your loaders (i.e., `json`/`defer`) then you shouldn't _need_ to make many changes to your app code, but please read through the "breaking" changes below to be aware of some of the underlying behavior changes - specifically around serialization and status/header behavior.
 
-- Single fetch uses a new streaming format under the hood via [`turbo-stream`][turbo-stream], which means that we can stream down more complex data than just JSON
-- Naked objects returned from `loader` and `action` functions are no longer automatically converted into a JSON `Response` and are serialized as-is over the wire
-- To get the most accurate type inference, you should do two things:
-  - Add `@remix-run/react/future/single-fetch.d.ts` to the end of your `tsconfig.json`'s `compilerOptions.types` array
-  - Begin using `unstable_defineLoader`/`unstable_defineAction` in your routes
-    - This can be done incrementally - you should have _mostly_ accurate type inference in your current state
-- Revalidation after an `action` `4xx`/`5xx` `Response` is now opt-in, versus opt-out
-- The [`headers`][headers] function is no longer used when Single Fetch is enabled, in favor of the new `response` stub passed to your `loader`/`action` functions
-- The old `installGlobals()` polyfill doesn't work for Single Fetch, you must either use the native Node 20 `fetch` API or call `installGlobals({ nativeFetch: true })` in your custom server to get the [undici-based polyfill][undici-polyfill]
+**1. Enable the future flag**
 
-## Details
+```diff
+  remix({
+    future: {
+      // ...
++     unstable_singleFetch: true,
+    },
+  }),
+```
 
-### Streaming Data Format
+**2. Update or remove `installGlobals`**
 
-Previously, Remix used `JSON.stringify` to serialize your loader/action data over the wire, and needed to implement a custom streaming format to support `defer` responses.
+Single Fetch requires using [`undici`][undici] as your `fetch` polyfill, or using the built-in `fetch` on Node 20+, because it relies on APIs available there that are not in the `@remix-run/web-fetch` polyfill. Please refer to the [Undici][undici-polyfill] section in the 2.9.0 release notes below for more details.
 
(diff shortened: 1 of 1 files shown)


docs/guides/single-fetch.md
```diff
@@ -367,7 +351,7 @@ This poses an interesting conundrum for [resource routes][resource-routes] which
...
 
-To ease backwards-compatibility and ease the adoption of the Single Fetch future flag, Remix v2 will handle this based on whether it's accessed from a Remix API or externally. In the future Remix will require you to return your own JSON response if you do not want raw objects to be streamed down for external consumption.
+To ease backwards-compatibility and ease the adoption of the Single Fetch future flag, Remix v2 will handle this based on whether it's accessed from a Remix API or externally. In the future Remix will require you to return your own [JSON response][returning-response] if you do not want raw objects to be streamed down for external consumption.
 
 The Remix v2 behavior with Single Fetch enabled is as follows:
```
brookslybrand, line 356: I can't comment where I want to, but down below I find these examples a little confusing. Since we're going to be deprecating `json` we should probably update these examples to use `new Response` instead, right?

I get that some instances we're trying to show a migration path, but I think it's a bit confusing to recommend using `json` when we also talk about how we're going to remove it in rr7

docs/guides/single-fetch.md
```diff
@@ -304,9 +290,7 @@ Each `loader`/`action` receives its own unique `response` instance so you cannot
   - `headers.append` can be used to set the same header from both a parent and child handler
   - `headers.delete` can be used to delete a value set by a parent handler, but not a value set from a child handler
 
-Because Single Fetch supports naked object returns, and you no longer need to return a `Response` instance to set status/headers, the `json`/`redirect`/`redirectDocument`/`defer` utilities can be considered deprecated when using Single Fetch.
-
-**\*However**, this does not mean you need to change them all right away in order to adopt Single Fetch!\*
+Because Single Fetch supports naked object returns, and you no longer need to return a `Response` instance to set status/headers, the `json`/`redirect`/`redirectDocument`/`defer` utilities should be considered deprecated when using Single Fetch.
```
brophdawg11, line 293: Maybe append this with something super clear like "these will remain for the duration of v2 so you don't need to remove the immediately.  They will likely be removed in the next major version so we recommend remove them incrementally between now and then." or something

brophdawg11 (commented):
👏

brophdawg11 (approved):

docs/guides/single-fetch.md
```diff
@@ -4,99 +4,95 @@ title: Single Fetch
...
+  ```diff
+  - installGlobals()
+  + installGlobals({ nativeFetch: true })
+  ```
 
-Once adopting Single Fetch, it is recommended that you incrementally remove the usage of `json`/`defer` throughout your application in favor of returning raw objects.
+- If you are using `remix-serve`, it will use `undici` automatically if Single Fetch is enabled.
 
-### React Rendering APIs
+**3. Replace `renderToString`**
```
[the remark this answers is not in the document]
brookslybrand: Good idea, I added this to the top

> For most Remix apps it's unlikely you're using `renderToString`, but if you have opted into using it in your `entry.server.tsx`, then continue reading, otherwise you can skip this step.

docs/guides/single-fetch.md
brookslybrand: Added! Thanks for the suggestion

docs/guides/single-fetch.md
```diff
@@ -416,7 +401,7 @@ export async function loader() {
 }
 ```
 
-To keep things consistent, resource route `loader`/`action` functions will still receive a `response` stub and you can use it if you need to (maybe to share code amongst non-resource-route handlers):
+To keep things consistent, resource route `loader`/`action` functions will still receive a `response` stub and you can use it if you need to (maybe to share code amongst non-resource route handlers):
 
 ```tsx filename=app/routes/resource.tsx
 // But you can still set values on the response stubstraightforward approach
```
brophdawg11, line 407:
```suggestion
// But you can still set values on the response stub
```

docs/guides/single-fetch.md
```diff
@@ -401,13 +385,14 @@ Note: It is _not_ recommended to use `defineLoader`/`defineAction` for externall
...
 
-Ad discussed above, the `headers` export is deprecated in favor of a new [`response` stub][responsestub] passed to your `loader` and `action` functions. This is somewhat confusing in resource routes, though, because you get to return the _actual_ `Response` - there's no real need for a "stub" concept because there's no merging results from multiple loaders into a single Response:
+As discussed above, the `headers` export is deprecated in favor of a new [`response` stub][responsestub] passed to your `loader` and `action` functions. This is somewhat confusing in resource routes, though, because you get to return the _actual_ `Response` - there's no real need for a "stub" concept because there's no merging results from multiple loaders into a single Response:
 
 ```tsx filename=app/routes/resource.tsx
 // Using your own Response is the most straightforward approach
   const data = await getData();
-  return json(data, {
+  const body = JSON.stringify(data);
```
brophdawg11, line 394:
```suggestion
```

docs/guides/single-fetch.md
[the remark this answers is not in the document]
michenly: ps. when I reading this, I am reminded of the `Inline Scripts` section below.

Both may not apply to everyone, but your app would break if you use these things and did not update.

Maybe both of those things should go into this step together.

docs/guides/single-fetch.md
```diff
@@ -4,99 +4,97 @@ title: Single Fetch
...
+  //    ^? { message: string, date: Date }
 }
 ```
 
+## Migrating a Route with Single Fetch
+
+The following changes are not required to take advantage of Single Fetch (refer to [Enabling Single Fetch][start] to get started). We recommend making the following changes on a route-by-route basis, as it's easier to validate changes to things like headers and data types.
+
+Making these changes will ensure a smooth, non-breaking upgrade to React Router v7.
+
```
michenly, line 94: ❤️ 💟 😍

docs/guides/single-fetch.md
```diff
@@ -4,99 +4,97 @@ title: Single Fetch
...
 ### Type Inference
 
-Without Single Fetch, any plain Javascript object returned from a `loader` or `action` is automatically serialized into a JSON response (as if you returned it via `json`). The type inference assumes this is the case and infer naked object returns as if they were JSON serialized.
+Without Single Fetch, any plain Javascript object returned from a `loader` or `action` is automatically serialized into a JSON response (as if you returned it via `json`). The type inference assumes this is the case and infers naked object returns as if they were JSON serialized.
```
michenly, line 97: The [Type Inference](https://remix.run/docs/en/main/guides/single-fetch#type-inference) in mentioned here under `Migrating a Route with Single Fetch` section as  optional change. 

But it is also mentioned in `Breaking Changes`
Screenshot 2024-05-30 at 5 15 24 PM

A little confuse as to if this is a Breaking or Optional change.

docs/guides/single-fetch.md
```diff
@@ -4,99 +4,97 @@ title: Single Fetch
...
 
-### Streaming Timeout
+## Breaking Changes
 
-Previously, Remix has a concept of an `ABORT_TIMEOUT` built-into the default [`entry.server.tsx`][entry-server] files which would terminate the React renderer, but it didn't do anything in particular to clean up any pending deferred promises.
+As mentioned above, if you are returning `Response` instances from your loaders (i.e., `json`/`defer`) you shouldn't _need_ to make many changes to your app code. To get better type inference and prepare for React Router v7 you can [migrate your routes one by one][migration-guide].
 
-Now that Remix is streaming internally, we can cancel the `turbo-stream` processing and automatically reject any pending promises and stream up those errors to the client. By default, this happens after 4950ms - a value that was chosen to be just under the current 5000ms `ABORT_DELAY` in most entry.server.tsx files - since we need to cancel the promises and let the rejections stream up through the React renderer prior to aborting the React side of things.
+There are a few important breaking changes Single Fetch introduces which are important to cover:
```
michenly, line 54: In a way the entire `Breaking Changes` section felt like it can be deleted and instead be listed as steps in `Enabling Single Fetch` section as must do.

Otherwise there is a inconsistency with `installGlobals` being the only thing that live in both `Breaking Change` and `Enabling Single Fetch`

docs/guides/single-fetch.md
brookslybrand: Yeah, I get what you're saying. It's more that the current type inference is buggy if you return naked objects (since `useLoader<typeof loader>()` assumes that the return value is serialized

There's probably a better way to say that, but basically if you want to get the best type safety you need to adopt `defineLoader`/`defineAction`

PR #9515, opened 2024-05-30
Fix code samples in data loading sections

ugultopu:
I'm following the docs step by step and these sections worked for me only after removing the `json` function.

### DIFF
--- a/docs/start/tutorial.md
+++ b/docs/start/tutorial.md
@@ -468,8 +468,7 @@ These params are most often used to find a record by ID. Let's try it out.
 
 <docs-info>The following code has type errors in it, we'll fix them in the next section</docs-info>
 
-```tsx filename=app/routes/contacts.$contactId.tsx lines=[1-2,5,7-10,13]
-import { json } from "@remix-run/node";
+```tsx filename=app/routes/contacts.$contactId.tsx lines=[1,4,6-9,12]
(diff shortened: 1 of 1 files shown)


brookslybrand:
@ugultopu what wasn't working with the `json` function exactly? I just ran through the tutorial up to your first example and it all works correctly for me

ugultopu:
The contact page didn't work for me. Attached is how it was rendered. I'm also attaching my working version (which lacks the `json` call). Maybe I missed something while following along the tutorial.

[remix-tutorial.zip](https://github.com/remix-run/remix/files/15500938/remix-tutorial.zip)
Screenshot 2024-05-30 at 4 58 35 PM

brookslybrand:
> The contact page didn't work for me. Attached is how it was rendered. I'm also attaching my working version (which lacks the `json` call). Maybe I missed something while following along the tutorial.

Here's a StackBlitz of it working with `json`

ugultopu:
Ok, I got it now. The initial contact component defines a contact as `const contact = {/* hardcoded dummy data */};`. Whereas the version that uses loader data needs to destructure it. I was omitting the destructuring. That is, instead of:

    const { contact } = useLoaderData<typeof loader>();

I was using:

    const contact = useLoaderData<typeof loader>();

PR #9535, opened 2024-05-31
Update future flags guide

brookslybrand:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/docs/guides/api-development-strategy.md
+++ b/docs/guides/api-development-strategy.md
@@ -1,0 +1,68 @@
+---
+title: Development Strategy
+---
+
+# Gradual Feature Adoption with Future Flags
+
+In our approach to software development, we aim to achieve the following goals for major releases:
+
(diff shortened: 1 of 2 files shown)


docs/guides/api-development-strategy.md
alcpereira, line 1: What do you think about suggesting `import { unstable_someFeature as someFeature } from @mypackage` in this section?

I've seen this pattern quite often, I think it helps a lot with migrations.

brookslybrand: Yep, I think that's a good approach (like what we're doing with `unstable_defineLoader`) right now.

Since this document is now just generally describing our Future Flag process and we don't have any examples of importing an unstable API, I'm not too concerned about it. I think I'd rather us just advise on recommended import patterns in the documentation for the specific feature (like with Single Fetch)

PR #9566, opened 2024-06-05
Avoid hydration loops when layout error renders also throw

### LINKED ISSUE #9553: Infinite loop on unhandle error using Layout Export with ErrorBoundary
#### This is the positive case with no Layout, but AppShell wrap around App & ErrorBoundary
1. Go to https://stackblitz.com/edit/remix-run-remix-jfuwjb
2. See the error loaded in browser
----------

#### This is the negative case with Layout Export ###
1. Go to https://stackblitz.com/edit/remix-run-remix-jfuwjb
2. Quit dev server
3. Rename root-with-layout.tsx to root.tsx
4. Open dev tool with console open
5. Start dev server
6. Click the link to see the error showing up in page

Following the example from https://remix.run/docs/en/main/start/tutorial to set up a base example use example here https://remix.run/docs/en/main/file-conventions/root to setup Layout Export and ErrorBoundary

This might be related to this close issue https://github.com/remix-run/remix/issues/1678#issuecomment-2127085554 but no longer related to React version.

### System Info

```shell
I can reproduce this with 
- Node 18 and Node 20
- React 18
- Remix 2.8 & 2.9
```

### Used Package Manager

npm

### Expected Behavior

That unhandle error would cause Application error in browser using Layout export

### Actual Behavior

Unhandle error cause infinite loading using Layout export

brophdawg11:
If you have a root `Layout` we will try to render that on the success _and_ error flows.  If the error flow _also_ throws during hydration (i.e., from a delayed resolved promise render) and we bubble to our default error boundary, then we can get into an infinite loop because we choose _not_ to render the surrounding `<html>` in the error boundary when a `Layout` exists.

Instead, if we are inside the default error boundary and the root route has both a `Layout` and an `ErrorBoundary` then we must have thrown trying to render the error via the `Layout` and we _should_ render the default `<html>` shell.  Otherwise we get into a react hydration loop because the DOM rejects when we try to add our default error boundary children (`h1`, `pre`) as children of `#document`.

Closes #9553

### DIFF
--- a/.changeset/mighty-experts-agree.md
+++ b/.changeset/mighty-experts-agree.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Avoid hydration loops when `Layout` `ErrorBoundary` renders also throw
--- a/integration/root-route-test.ts
+++ b/integration/root-route-test.ts
@@ -153,4 +153,74 @@ test.describe("root route", () => {
 
     console.error = oldConsoleError;
   });
+
+  test("Skip the Layout on subsequent client renders if the ErrorBoundary throws", async ({
+    page,
+  }) => {
+    let oldConsoleError;
+    oldConsoleError = console.error;
+    console.error = () => {};
+
+    fixture = await createFixture(
+      {
+        files: {
+          "app/root.tsx": js`
+            import * as React from "react";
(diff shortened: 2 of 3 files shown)


brophdawg11:
Note to self - add changeset

PR #9580, opened 2024-06-06
Dont prefetch server loader when client loader exists

### LINKED ISSUE #9575: Link prefetch calls loader despite the presence of clientLoader
### Reproduction

https://stackblitz.com/edit/remix-run-remix-s2qdww

Hover on link to 'Other', in network devtools you can see a request for loader, and server log shows that loader was called.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @remix-run/dev: * => 2.9.2 
    @remix-run/node: * => 2.9.2 
    @remix-run/react: * => 2.9.2 
    @remix-run/serve: * => 2.9.2 
    vite: ^5.1.0 => 5.2.12
```

### Used Package Manager

npm

### Expected Behavior

When using `<Link to="" prefetch="intent">`, prefetch will fetch the `loader` from network, while navigating to that page will correctly only call `clientLoader`. 

In case of using clientLoader as https://remix.run/docs/en/main/guides/client-data#skip-the-hop, that caused double network requests and result from `loader` is discarded.

### Actual Behavior

Either call clientLoader, or no loader at all.

When adding a clientLoader to existing route I would expect that loader _data request will never trigger again unless I explicitly call serverLoader, even if I forget to remove prefetch from Link elements

brophdawg11:
Closes #9575

### DIFF
--- a/packages/remix-react/links.ts
+++ b/packages/remix-react/links.ts
@@ -425,7 +425,11 @@ export function getDataLinkHrefs(
   let path = parsePathPatch(page);
   return dedupeHrefs(
     matches
-      .filter((match) => manifest.routes[match.route.id].hasLoader)
+      .filter(
+        (match) =>
+          manifest.routes[match.route.id].hasLoader &&
+          !manifest.routes[match.route.id].hasClientLoader
+      )
       .map((match) => {
         let { pathname, search } = path;
         let searchParams = new URLSearchParams(search);
--- a/.changeset/heavy-ways-rule.md
+++ b/.changeset/heavy-ways-rule.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Don't prefetch server loader data when clientLoader exists
--- a/integration/client-data-test.ts
+++ b/integration/client-data-test.ts
@@ -977,6 +977,48 @@ test.describe("Client Data", () => {
           'not have a server loader (routeId: "routes/parent.child")'
       );
     });
+
+    test("does not prefetch server loader if a client loader is present", async ({
+      page,
+    }) => {
+      appFixture = await createAppFixture(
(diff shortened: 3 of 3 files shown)


packages/remix-react/links.ts
```diff
@@ -425,7 +425,11 @@ export function getDataLinkHrefs(
   let path = parsePathPatch(page);
   return dedupeHrefs(
     matches
-      .filter((match) => manifest.routes[match.route.id].hasLoader)
+      .filter(
+        (match) =>
+          manifest.routes[match.route.id].hasLoader &&
+          !manifest.routes[match.route.id].hasClientLoader
+      )
```
brophdawg11, line 432: Check with @ryanflorence on this - current behavior is desired for folks who are calling `await serverLoader()` unconditionally because we can prefetch that so it resolves immediately.  This new solution hurts those scenarios in favor of apps that conditionally call `serverLoader()`.  Maybe we need to provide some way to enable "prefetching" (pre-executing) of client loaders - either by stashing their results into an internal cache or adding the `type` field to loaders we've previously discussed.

brophdawg11: 👍 Chatted and agreed we should only prefetch module links when a client loader exists

dantrain: Would be great to have the option to prefetch/pre-execute client loaders. In my case I am using a library that has its own cache, so calling the client loader is all that's needed.

PR #9581, opened 2024-06-06
Add note on layout error boundary usage

brophdawg11:


docs/file-conventions/root.md
```diff
@@ -150,7 +150,9 @@ export function ErrorBoundary() {
...
 
-Because your `Layout` component is used in both success and error flows, this same restriction holds. If you need to fork logic in your `Layout` depending on if it was a successful request or not, you can use `useRouteLoaderData("root")` and `useRouteError()`:
+Because your `Layout` component is used in both success and error flows, this same restriction holds. If you need to fork logic in your `Layout` depending on if it was a successful request or not, you can use `useRouteLoaderData("root")` and `useRouteError()`.
+
+<docs-warn>Because your `<Layout>` component is used for rendering the `ErrorBoundary`, you should be _very defensive_ to ensure that you can render your `ErrorBoundary` without encountering any render errors. If your `Layout` throws another error trying to render the boundary, then it can't can't be used and your UI will fall back to the very minimal built-in default `ErrorBoundary`.</docs-warn>
```
michenly, line 155: >If your `Layout` throws another error trying to render the boundary, then it can't can't be used and your UI will fall back to the very minimal built-in default `ErrorBoundary`

A little confuse about this sentence. 

In my example when your Layout throw,  it will get an infinite looping error that crash your browser. I feel like falling back to the very minimal built-in default would actually be preferable but doesnt seems to be the case?

brophdawg11: it does currently - https://github.com/remix-run/remix/pull/9566 fixe that loop and introduces falling back to the minimal root layout

michenly: awww amazing! That was really the fix I am looking for :)
Will follow that issue and watch for the release.

PR #9585, opened 2024-06-07
Add Tailwind CSS to the templates

brookslybrand:
- [x] Add tailwind
- [ ] Test all the templates


templates/cloudflare-workers/app/root.tsx
```diff
@@ -1,10 +1,16 @@
...
   Meta,
   Outlet,
   Scripts,
   ScrollRestoration,
 } from "@remix-run/react";
+import stylesheet from "~/tailwind.css?url";
+
+export const links: LinksFunction = () => [
+  { rel: "stylesheet", href: stylesheet },
+];
```
brookslybrand, line 13: Is there any reason to not just use a side-effect css import with tailwind?

pcattori:
Worth noting that Tailwind v4 is planning to release with a Vite plugin that should obsolete all/most of this setup. Might be worth waiting for that, but not sure when they plan to ship it

silvenon:
Also, by doing this we lose the benefit that `npx tailwindcss init` provides.

Setting up Tailwind will always take up to a few minutes, compared to months or years of building the project. How often are we starting new projects that those minutes became so unbearable?

Even though I love Tailwind, I feel like this would be a step backwards, and a maintenance burden.

brookslybrand:
> Worth noting that Tailwind v4 is planning to release with a Vite plugin that should obsolete all/most of this setup. Might be worth waiting for that, but not sure when they plan to ship it

Haha yeah, I thought about that. If it's a matter of weeks I'm happy to wait, but I have no idea how long it'll be (or if anyone knows)

brookslybrand:
> Also, by doing this we lose the benefit that `npx tailwindcss init` provides.
> 
> Setting up Tailwind will always take up to a few minutes, compared to months or years of building the project. How often are we starting new projects that those minutes became so unbearable?
> 
> Even though I love Tailwind, I feel like this would be a step backwards, and a maintenance burden for you.

I appreciate the thoughts. it's hard balance. The majority of times people run something like `npx create-remix@latest` is because they want to
- try Remix for the first time
- teach Remix
- Experiment with something they'll throw away

It's a pretty small amount of the time people run a create-* CLI to actually start a project they'll be working on for months or years. So I kind of few it the opposite, how many minutes are added for projects that are going to last months or years to either delete tailwind or modify the way they want to (even using `npx tailwindcss init`).

I definitely get why it can be seen as a step backwards, since it somewhat degrades the "purity" of the Remix template. I'm not sure I understand why you think it'll be a maintenance burden? Especially once tailwind 4 with the Vite plugin comes out

brookslybrand:
The other piece I should mention is that we want to make the templates a little nicer and more styled, so either we'll have to use something like tailwind, or ship a bunch of custom css or css modules that the user will delete. So either way in an attempt to make the templates more inviting, we're going to be including some degree of "you have to delete this stuff". I'm not sure the best way to strike that balance

templates/cloudflare-workers/app/root.tsx
kiliman: Since Vite is now the default, I prefer the side-effect import for tailwind. It's more idiomatic.

kiliman:
Yes, when I create an example on StackBlitz or Codesandbox, it's always a hassle if you want to use something like `flex` without resorting to inline styles or external CSS. Having Tailwind as part of the default template would definitely help.

brookslybrand:
> Ok, I'm convinced. I would rather provide that than the way Vite starts. Forget what I said about maintainenance burden, I was mostly refering to the purity, as you recognized.

Thanks for your thoughts! We're gonna be updating these templates a bit over the next week or two I imagine, so please reach out if you feel like anything is funky

Thanks @kiliman for the thoughts as well

templates/cloudflare-workers/app/root.tsx
brookslybrand: Also makes it easier to remove if you don't want to use tailwind.

Updated!

templates/cloudflare-workers/app/root.tsx
```diff
@@ -1,16 +1,11 @@
...
   Outlet,
   Scripts,
   ScrollRestoration,
 } from "@remix-run/react";
-import stylesheet from "./tailwind.css?url";
-
-export const links: LinksFunction = () => [
-  { rel: "stylesheet", href: stylesheet },
-];
+import "./tailwind.css?url";
```
kiliman, line 8: For side-effect imports, I don't think you want the `?url` suffix.

```ts
import "./tailwind.css";
```

brookslybrand: Yep, you're absolutely right, totally missed that when copying all these over

The good thing is testing showed this to pretty obviously be a mistake 😅

saifulapm:
Why not 
```ts
import stylesheet from '~/tailwind.css?url'

export const links: LinksFunction = () => {
  return [
    { rel: 'stylesheet', href: stylesheet },
  ]
}
```
instead of `import "./tailwind.css";` since `Vite v5.1` has shipped.

kiliman:
You can use either method. Since Remix/RR is going all in with Vite, it's probably better to go with the Vite way (side-effect import).

PR #9600, opened 2024-06-10
Fog of War

### LINKED ISSUE #8423: 🗺️ Fog of War
RFC: https://github.com/remix-run/react-router/discussions/11113

brophdawg11:
Implements Fog of War (https://github.com/remix-run/react-router/discussions/11113) for Remix via the manifest

* The minimal `__remixManifest` is sent up on page load in an inline `<script>` just like `__remixContext`
* Rendered `<Link>` components are batched up and a request is sent to the Remix server handler via a built-in `/__manifest` endpoint to match paths on the server and send back relevant route manifest entries
* Those entries are patched into `__remixManifest` and the client side router 
* If a "discover" pass isn't able to patch the manifest in-time and a user clicks a `Link`, we fall into the `patchRoutesOnMiss` handler which will load patches for that path during the `loading` phase

RR Sibling PR: https://github.com/remix-run/react-router/pull/11626
Closes #8423

### DIFF
--- a/packages/remix-react/browser.tsx
+++ b/packages/remix-react/browser.tsx
@@ -79,12 +79,19 @@ let hmrRouterReadyPromise = new Promise<Router>((resolve) => {
   return undefined;
 });
 
-let fogOfWar: {
-  enabled: boolean;
+type FogOfWarInfo = {
+  controller: AbortController | null;
+  // Currently rendered links that may need prefetching
   nextPaths: Set<string>;
+  // Paths we know the client can already match, so no need to perform client-side
+  // matching or prefetching for them.  Just an optimization to avoid re-matching
+  // on a larger and larger route tree over time
   knownGoodPaths: Set<string>;
+  // Routes the server was unable to match - don't ask for them again
   known404Paths: Set<string>;
 };
+
+let fogOfWar: FogOfWarInfo | null = null;
 
 // @ts-expect-error
 if (import.meta && import.meta.hot) {
@@ -202,7 +209,7 @@ if (import.meta && import.meta.hot) {
  * `app/entry.client.js`). This component is used by React to hydrate the HTML
  * that was received from the server.
  */
-export function RemixBrowser(props: RemixBrowserProps): ReactElement {
+export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
   if (!router) {
     // Hard reload if the path we tried to load is not the current path.
     // This is usually the result of 2 rapid back/forward clicks from an
@@ -311,20 +318,17 @@ export function RemixBrowser(props: RemixBrowserProps): ReactElement {
       }
     }
 
-    fogOfWar = {
-      enabled:
-        window.__remixContext.future.unstable_fogOfWar === true &&
-        !window.__remixContext.isSpaMode,
-      // Track rendered links for fog of war prefetching
-      nextPaths: new Set<string>(),
-      // Once a path has been matched in the client, track it so we never have to
-      // re-match - just an optimization to avoid re-matching on a larger and larger
-      // route tree over time
-      knownGoodPaths: new Set<string>(),
-      // Track routes that the server was unable to match so we don't ask for
-      // them again
-      known404Paths: new Set<string>(),
-    };
+    if (
+      window.__remixContext.future.unstable_fogOfWar === true &&
+      !window.__remixContext.isSpaMode
+    ) {
+      fogOfWar = {
+        controller: null,
+        nextPaths: new Set<string>(),
+        knownGoodPaths: new Set<string>(),
+        known404Paths: new Set<string>(),
+      };
+    }
 
     // We don't use createBrowserRouter here because we need fine-grained control
     // over initialization to support synchronous `clientLoader` flows.
@@ -350,17 +354,17 @@ export function RemixBrowser(props: RemixBrowserProps): ReactElement {
             window.__remixRouteModules
           )
         : undefined,
-      ...(fogOfWar.enabled
+      ...(fogOfWar
         ? {
             async unstable_patchRoutesOnMiss({ path, patch }) {
-              if (!fogOfWar.known404Paths.has(path)) {
-                await fetchAndApplyManifestPatches(
-                  [path],
-                  window.__remixContext.basename,
-                  window.__remixManifest.version,
-                  patch
-                );
-              }
+              if (fogOfWar!.known404Paths.has(path)) return;
+              await fetchAndApplyManifestPatches(
+                [path],
+                fogOfWar!.known404Paths,
+                window.__remixContext.basename,
+                window.__remixManifest.version,
+                patch
+              );
             },
           }
         : {}),
@@ -426,27 +430,31 @@ export function RemixBrowser(props: RemixBrowserProps): ReactElement {
 
   // eslint-disable-next-line react-hooks/rules-of-hooks
   React.useEffect(() => {
-    // Don't perform any prefetching if the app asked not to or the user has saveData enabled
-    if (!fogOfWar.enabled || navigator.connection?.saveData === true) {
+    // Don't perform prefetching without flag or if the user has `saveData` enabled
+    if (!fogOfWar || navigator.connection?.saveData === true) {
       return;
     }
 
     async function fetchPatches() {
       fogOfWarAbortControllerRef.current?.abort();
 
-      let lazyPaths = getFogOfWarPaths(window.__remixContext.basename);
+      let lazyPaths = getFogOfWarPaths(
+        fogOfWar!,
+        window.__remixContext.basename
+      );
       if (lazyPaths.length === 0) {
         return;
       }
 
       try {
-        fogOfWarAbortControllerRef.current = new AbortController();
+        fogOfWar!.controller = new AbortController();
         await fetchAndApplyManifestPatches(
           lazyPaths,
+          fogOfWar!.known404Paths,
           window.__remixContext.basename,
           window.__remixManifest.version,
           router.patchRoutes,
-          fogOfWarAbortControllerRef.current.signal
+          fogOfWar!.controller.signal
         );
       } catch (e) {
         console.error("Failed to fetch manifest patches", e);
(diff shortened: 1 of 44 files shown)


packages/remix-react/browser.tsx
```diff
@@ -324,6 +345,15 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
         : undefined,
+      async unstable_patchRoutesOnMiss(path, _, patch) {
+        if (!known404Paths.has(path)) {
+          await fetchAndApplyManifestPatches(
+            [path],
+            window.__remixManifest.version,
+            patch
+          );
+        }
+      },
```
brophdawg11, line 356: If we click a link that doesn't match, try to fetch that path from the server during the `loading` phase.  In an ideal scenario, this never runs because the pre-discovery logic all lands prior to link clicks.

packages/remix-react/browser.tsx
```diff
@@ -381,6 +411,80 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+
+    async function fetchPatches() {
+      if (fogOfWarAbortControllerRef.current) {
+        fogOfWarAbortControllerRef.current.abort();
+      }
+
+      let lazyPaths = getFogOfWarPaths();
+      if (lazyPaths.length === 0) {
+        return;
+      }
```
brophdawg11, line 432: Short circuit there are no actively rendered paths to fetch - which means that they are all either matchable via the currently known routes, or they are a known 404 from a prior fetch.

packages/remix-react/browser.tsx
```diff
@@ -381,6 +411,80 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+          let el = node as Element;
+          let links = Array.from(el.querySelectorAll("a[data-discover]"));
+          if (el.tagName === "A" && el.getAttribute("data-discover")) {
+            links.push(el);
+          }
+          links.forEach((el) => registerPath(el.getAttribute("href")));
+          debouncedFetchPatches();
+        });
+      });
+    });
```
brophdawg11, line 469: This logic detects newly rendered links on navigations and internal route state changes (i.e., click this button to show a link) and batches them up via the debounced fetch

packages/remix-react/browser.tsx
```diff
@@ -381,6 +411,80 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+
+    document.body
+      .querySelectorAll("a[data-discover]")
+      .forEach((a) => registerPath(a.getAttribute("href")));
+
+    fetchPatches();
```
brophdawg11, line 475: This does the initial fetch for all routes rendered in the initial HTML payload

packages/remix-react/components.tsx
```diff
@@ -210,6 +220,9 @@ let NavLink = React.forwardRef<HTMLAnchorElement, RemixNavLinkProps>(
           {...prefetchHandlers}
           ref={mergeRefs(forwardedRef, ref)}
           to={to}
+          data-discover={
+            !isAbsolute && discover === "render" ? "true" : undefined
+          }
```
brophdawg11, line 225: `data-discover` opts into the manifest patching.  It's enabled by default (`discover="render"`), users can opt out via `<Link discover="click">`/`<NavLink discover="click">`

brophdawg11: The use of a `data-` attribute here instead of just react state is for a few reasons:

1. Makes it easier to detect via `MutationObserver` and not a mess of React context/effects where Links "register" themselves with some parent context
2. It scales very well into an RSC world where you may just want to render `<a href="">` and not _have_ to use a client component just to participate in fog of war
3. If we ever remove `Link` in favor of event delegation, it works there too

packages/remix-react/components.tsx
```diff
@@ -846,13 +862,14 @@ export function Scripts(props: ScriptProps) {
           ? `__remixContext.a=${deferredScripts.length};`
           : "");
 
+    // TODO: Put fog of war behind a future flag
```
brophdawg11, line 865: Note to self

packages/remix-react/components.tsx
```diff
@@ -914,11 +935,15 @@ import(${JSON.stringify(manifest.entry.module)});`;
...
-      return matches;
+      // TODO: With fog of war we won't always be able to match the next
+      // routes synchronously.
+      // TODO: Is this even used since we don't even render these when
+      // isHydrated is false?  Maybe dead code?
+      // invariant(
+      //   matches,
+      //   `No routes match path "${navigation.location.pathname}"`
+      // );
+      return matches || [];
```
brophdawg11, line 946: Note to self

packages/remix-react/browser.tsx
```diff
@@ -381,6 +421,81 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+      return;
+    }
+
+    async function fetchPatches() {
+      fogOfWarAbortControllerRef.current?.abort();
+
+      let lazyPaths = getFogOfWarPaths(window.__remixContext.basename);
+      if (lazyPaths.length === 0) {
+        return;
+      }
```
brophdawg11, line 440: Short circuit there are no actively rendered paths to fetch - which means that they are all either matchable via the currently known routes, or they are a known 404 from a prior fetch.

packages/remix-react/browser.tsx
```diff
@@ -324,6 +350,20 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+            async unstable_patchRoutesOnMiss({ path, patch }) {
+              if (!fogOfWar.known404Paths.has(path)) {
+                await fetchAndApplyManifestPatches(
+                  [path],
+                  window.__remixContext.basename,
+                  window.__remixManifest.version,
+                  patch
+                );
+              }
+            },
```
brophdawg11, line 364: If we click a link that doesn't match, try to fetch that path from the server during the loading phase. In an ideal scenario, this never runs because the pre-discovery logic all lands prior to link clicks.

packages/remix-react/browser.tsx
```diff
@@ -381,6 +425,88 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+    async function fetchPatches() {
+      fogOfWarAbortControllerRef.current?.abort();
+
+      let lazyPaths = getFogOfWarPaths(
+        fogOfWar!,
+        window.__remixContext.basename
+      );
+      if (lazyPaths.length === 0) {
+        return;
+      }
```
brophdawg11, line 447: Short circuit there are no actively rendered paths to fetch - which means that they are all either matchable via the currently known routes, or they are a known 404 from a prior fetch.

packages/remix-react/components.tsx
```diff
@@ -905,27 +938,7 @@ import(${JSON.stringify(manifest.entry.module)});`;
...
-      invariant(
-        matches,
-        `No routes match path "${navigation.location.pathname}"`
-      );
-      return matches;
-    }
-
-    return [];
-  }, [navigation.location, router.routes, router.basename]);
-
```
brophdawg11, line 926: @jacob-ebey This seems like dead code to me?  It only returns a non-empty array when we're in the middle of a client-side `navigation` (which implies we've already hydrated), but the `routePreloads` it gets concatenated into below is only ever rendered when we're not hydrated.  I think it's likely leftover from the `TransitionManager` implementation and maybe before `<PrefetchPageLinks>`?

packages/remix-server-runtime/server.ts
```diff
@@ -133,6 +132,27 @@ export const createRequestHandler: CreateRequestHandlerFunction = (
...
+      } catch (e) {
+        handleError(e);
+        return new Response("Unknown Server Error", { status: 500 });
+      }
+    }
+
+    let matches = matchServerRoutes(routes, url.pathname, _build.basename);
+    if (matches && matches.length > 0) {
+      Object.assign(params, matches[0].params);
+    }
```
brophdawg11, line 154: Moved this down so we don't bother matching on `/__manifest` requests

packages/remix-react/fog-of-war.ts
```diff
@@ -0,0 +1,259 @@
...
+            registerPath(node.getAttribute("href"));
+          } else if (node.tagName !== "A") {
+            node
+              .querySelectorAll("a[data-discover]")
+              .forEach((el) => registerPath(el.getAttribute("href")));
+          }
+          debouncedFetchPatches();
+        });
+      });
+    });
```
brophdawg11, line 152: This logic detects newly rendered links on navigations and internal route state changes (i.e., click this button to show a link) and batches them up via the debounced fetch

packages/remix-react/fog-of-war.ts
```diff
@@ -0,0 +1,259 @@
...
+        console.error("Failed to fetch manifest patches", e);
+      }
+    }
+
+    // Register and fetch patches for all initially-rendered links
+    document.body
+      .querySelectorAll("a[data-discover]")
+      .forEach((a) => registerPath(a.getAttribute("href")));
+
+    fetchPatches();
```
brophdawg11, line 129: This does the initial fetch for all routes rendered in the initial HTML payload

packages/remix-react/fog-of-war.ts
```diff
@@ -0,0 +1,259 @@
...
+        fogOfWar!.known404Paths,
+        manifest,
+        routeModules,
+        future,
+        isSpaMode,
+        basename,
+        patch
+      );
+    },
+  };
```
brophdawg11, line 70: If we click a link that doesn't match, try to fetch that path from the server during the loading phase. In an ideal scenario, this never runs because the pre-discovery logic all lands prior to link clicks.

packages/remix-react/fog-of-war.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    // Fetch patches for all currently rendered links
+    async function fetchPatches() {
+      fogOfWar?.controller?.abort();
+
+      let lazyPaths = getFogOfWarPaths(fogOfWar!, router);
+      if (lazyPaths.length === 0) {
+        return;
+      }
```
brophdawg11, line 104: Short circuit there are no actively rendered paths to fetch - which means that they are all either matchable via the currently known routes, or they are a known 404 from a prior fetch.

PR #9626, opened 2024-06-17
Remove AbortController and respect URL length limits

brophdawg11:
* Using the abort controller causes log noise in `StrictMode`, and there's no real benefit to aborting since the response is still valid so might as well patch those routes to avoid subsequent requests if we navigate back to those routes.
* Switch to a `POST` request if the URL goes beyond 4196 characters (a fairly conservative limit) to avoid running into URL or Request size limits
  * Only GET requests get the built-in `Cache-Control` response header

### DIFF
--- a/packages/remix-react/fog-of-war.ts
+++ b/packages/remix-react/fog-of-war.ts
@@ -24,6 +24,10 @@ type FogOfWarInfo = {
   known404Paths: Set<string>;
 };
 
+// 7.5k to come in under the ~8k limit for most browsers
+// https://stackoverflow.com/a/417184
+const URL_LIMIT = 7680;
+
 let fogOfWar: FogOfWarInfo | null = null;
 
 export function isFogOfWarEnabled(future: FutureConfig, isSpaMode: boolean) {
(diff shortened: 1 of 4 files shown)


packages/remix-react/fog-of-war.ts
```diff
@@ -193,10 +187,7 @@ export function useFogOFWarDiscovery(
       attributeFilter: ["data-discover", "href"],
     });
 
-    return () => {
-      fogOfWar?.controller?.abort();
```
brophdawg11, line 197: This is technically against the "rules of react" because we're not cleaning up our side effect, but functionally it shouldn't cause any issues since the requests are additive and shouldn't have race conditions

packages/remix-react/fog-of-war.ts
```diff
@@ -213,33 +221,43 @@ function getFogOfWarPaths(fogOfWar: FogOfWarInfo, router: Router) {
       return false;
     }
 
-    let matches = matchRoutes(router.routes, path, router.basename);
-    if (matches) {
-      knownGoodPaths.add(path);
-      nextPaths.delete(path);
-      return false;
-    }
```
brophdawg11, line 221: This could produce false positives, so we only add to `knownGoodPaths` in response to a manifest fetch where we matched on the server

Issue #9657, opened 2024-06-25
forwarding of request cookie headers comma separated instead of semi-colon separated


Courey:
### Reproduction

if you use version 2.8.1, and in an auth loader you set 
`return redirect(someAuthURL, { headers: { 'Set-Cookie': cookie } })`
it will work as expected.

if you then upgrade to version 2.9.2, the cookies you passed into the redirect are comma separated and break auth since there are no headers named for what you are looking for since they end up as one giant single value header that is _session with all the other headers as its value:
```
{"_session":"[REDACTED], _session.legacy=[REDACTED], csrfToken=[REDACTED], X-CSRFToken=[REDACTED], csrf=[REDACTED], session=[REDACTED]"}
```

### System Info

```shell
## This is the version that breaks:
  System:
    OS: macOS 13.4
    CPU: (12) arm64 Apple M2 Max
    Memory: 31.86 GB / 96.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    Yarn: 1.22.21 - ~/dev/gcn.nasa.gov/node_modules/.bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
  Browsers:
    Chrome: 126.0.6478.115
    Safari: 16.5
  npmPackages:
    @remix-run/architect: ^2.9.2 => 2.9.2
    @remix-run/css-bundle: ^2.9.2 => 2.9.2
    @remix-run/dev: ^2.9.2 => 2.9.2
    @remix-run/eslint-config: ^2.9.2 => 2.9.2
    @remix-run/node: ^2.9.2 => 2.9.2
    @remix-run/react: ^2.9.2 => 2.9.2

## this is the version where it works
  System:
    OS: macOS 13.4
    CPU: (12) arm64 Apple M2 Max
    Memory: 30.14 GB / 96.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    Yarn: 1.22.21 - ~/dev/temporary/node_modules/.bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
  Browsers:
    Chrome: 126.0.6478.115
    Safari: 16.5
  npmPackages:
    @remix-run/architect: ^2.8.1 => 2.8.1
[...]

PR #9664, opened 2024-06-27
reverting createRemixHeaders to previous fix for handling multiple cookies

Courey:
Closes: #9657

Testing Strategy:
I ran this locally and compared the output. This was also a re-introduced bug so this change is just going back to how it was a couple of months ago. It was introduced [here](https://github.com/remix-run/remix/commit/ad83e53c42b9da24c89408a14d5154563de3b477#diff-e37d9e295b178aa51a4880acae4b8bc7f82759e6868583513a5fa52b16846408L91-R93) and that is the change I am reverting.


brophdawg11:
Do you want to take a stab at a unit test to prevent a regression like this in the future?

Courey:
I can do that, my issue was that in order to write a unit test, I have to change the argument typing for `createRemixHeaders` because it's not a string array in this scenario, so it would need to accept both an object and a string array. It isn't a problem at runtime since the package is bundled and ends up as just plain javascript so the typing doesn't matter and it still works just fine, but for test writing the types need to match. Is it okay to change the typing on the argument to do this?

Courey:
Working on it now!

Courey:
The challenge I am having is that [this test](https://github.com/remix-run/remix/blob/%40remix-run/architect%402.10.0/packages/remix-architect/__tests__/server-test.ts#L224-L234) should have covered it. But I think that it's because the issue arises when using `@remix-run/web-fetch` and the tests are run `native fetch` (I haven't yet dug deeply into that yet). So if that is the case, I am unsure of how to write a test that runs using `@remix-run/web-fetch` when the others are using `native fetch`. Is there an already existing way to do this? Am I wrong in thinking that's the issue I'm having with writing a test that captures the scenario?

brophdawg11:
ah ok I don't think I realized this was only an issue when using web-fetch.  Let me play around with a way to try to test this with both fetch implementations.  

In the meantime - would you mind rebasing this and pointing it to the `dev` branch since it modifies source code?

brophdawg11:
Since `createRemixHeaders` is internal, how about if we expose an optional DI parameter we can use to force a `Headers` implementation for unit testing purposes?

```ts
export function createRemixHeaders(
  requestHeaders: APIGatewayProxyEventHeaders,
  requestCookies?: string[],
  _Headers?: typeof Headers
): Headers {
  // `_Headers` should only be used for unit testing purposes so we can unit test
  // the different behaviors of the @remix-run/web-fetch `Headers` implementation
  // and the node/undici implementation.  See:
  // https://github.com/remix-run/remix/issues/9657
  let HeadersImpl = _Headers || Headers;
  let headers = new HeadersImpl();

  for (let [header, value] of Object.entries(requestHeaders)) {
    if (value) {
      headers.append(header, value);
    }
  }

  if (requestCookies) {
    // Do the semi-colon joining manually to avoid differences between Remix
    // and node/undici `Headers` implementations
    headers.append("Cookie", requestCookies.join("; "));
  }

  return headers;
}
```

Then to test we can add a test for the `web-fetch` implementation:

```ts
import { Headers as RemixHeaders } from "@remix-run/web-fetch";
...
    it("handles multiple request cookies when using @remix-run/web-fetch", () => {
      let headers = createRemixHeaders(
        {},
        ["__session=some_value", "__other=some_other_value"],
        RemixHeaders
      );
      expect(headers.get("cookie")).toEqual(
        "__session=some_value; __other=some_other_value"
      );
    });
```

We'll just need to install `@remix-run/web-fetch` as a dependency in the architect package

Courey:
That test does pass! However, I am encountering an issue with typing:
Screenshot 2024-07-01 at 12 36 52 PM
I'm still fiddling with it, but I figured I'd let you know why it was taking a while since you gave me the test.

brophdawg11:
oh - yeah for the test I would just `@ts-expect-error` that - our old implementation is known to be missing `getSetCookie` 😕

brophdawg11:
Looking good!  Could you add a [changeset](https://github.com/remix-run/remix/pull/9664#issuecomment-2194761380) as well so we can get this merged?  Just run `pnpm changeset` and then choose the architect package and a `patch` semver bump

Courey:
I just realized that I didn't add it to the package.json in architect! Hold on merging till I get that added please!

Courey:
Screenshot 2024-07-03 at 2 14 18 PM
I'm running the full test suite locally and I see these failures, would this be new to this branch? Just trying to be thorough.

packages/remix-architect/__tests__/server-test.ts
```diff
@@ -230,6 +231,17 @@ describe("architect createRemixHeaders", () => {
         "__session=some_value; __other=some_other_value"
       );
     });
+    it("handles multiple request cookies when using @remix-run/web-fetch", () => {
+      let headers = createRemixHeaders(
+        {},
+        ["__session=some_value", "__other=some_other_value"],
+        //@ts-expect-error
```
brophdawg11, line 238:
```suggestion
        // @ts-expect-error types don't align since it's not fully spec compliant
```

packages/remix-architect/__tests__/server-test.ts
```diff
@@ -230,6 +231,17 @@ describe("architect createRemixHeaders", () => {
         "__session=some_value; __other=some_other_value"
       );
     });
+    it("handles multiple request cookies when using @remix-run/web-fetch", () => {
```
brophdawg11, line 234:
```suggestion

    it("handles multiple request cookies when using @remix-run/web-fetch", () => {
```

packages/remix-architect/__tests__/server-test.ts
```diff
@@ -3,6 +3,7 @@ import path from "node:path";
 import { createRequestHandler as createRemixRequestHandler } from "@remix-run/node";
 import type { APIGatewayProxyEventV2 } from "aws-lambda";
 import lambdaTester from "lambda-tester";
+import { Headers as RemixHeaders } from "@remix-run/web-fetch";
```
brophdawg11, line 6:
```suggestion
import { createRequestHandler as createRemixRequestHandler } from "@remix-run/node";
import { Headers as RemixHeaders } from "@remix-run/web-fetch";
import type { APIGatewayProxyEventV2 } from "aws-lambda";
import lambdaTester from "lambda-tester";
```

brophdawg11 (approved):
Thanks!

PR #9665, opened 2024-06-27
Support Fog of War form discovery and fix link prefetching

brophdawg11:
* We forgot to enable Fog of War discovery for `<Form>` components in the first pass
* Add an E2E test to ensure that we can prefetch after FOW discovery - requires an RR prerelease before CI will pass

### DIFF
--- a/packages/remix-react/components.tsx
+++ b/packages/remix-react/components.tsx
@@ -13,11 +13,13 @@ import type {
 } from "@remix-run/router";
 import type {
   FetcherWithComponents,
+  FormProps,
   LinkProps,
   NavLinkProps,
 } from "react-router-dom";
 import {
   Await as AwaitRR,
+  Form as RouterForm,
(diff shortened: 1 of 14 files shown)


packages/remix-react/components.tsx
```diff
@@ -271,6 +273,33 @@ let Link = React.forwardRef<HTMLAnchorElement, RemixLinkProps>(
...
+ * submit information via your website.
+ *
+ * @see https://remix.run/components/form
+ */
+let Form = React.forwardRef<HTMLFormElement, RemixFormProps>(
+  ({ discover = "render", ...props }, forwardedRef) => {
+    let isAbsolute =
+      typeof props.action === "string" && ABSOLUTE_URL_REGEX.test(props.action);
+    return (
+      <RouterForm
```
ryanflorence, line 291: we missed the `<form ref={forwardedRef}>`

PR #9675, opened 2024-06-29
Update tutorial.md

MasanoriOnishi:
fix path name
otherwise, I cannot execute delete when clicking delete button.

"app/routes/contacts.\$contactId.destroy.tsx"

-->

"app/routes/contacts.$contactId_.destroy.tsx"

Closes: #

- [x] Docs

Testing Strategy:

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -417,6 +417,7 @@
 - marshallwalker
 - martensonbj
 - marvinwu
+- MasanoriOnishi
 - matchai
 - mathieusteele
 - matt-l-w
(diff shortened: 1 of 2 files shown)


brophdawg11:
This should not be necessary since `destroy` is a resource route so it has no UI - and the feature you're adding is to alter the [UI layout nesting](https://remix.run/docs/en/main/file-conventions/routes#nested-urls-without-layout-nesting).  If you are having issues submitting to that route, please open an issue with a minimal reproduciton and we can take a look.  Thanks!

Issue #9212, opened 2024-04-08
`TypeError: Can't modify immutable headers.` in handleDataRequestRR (Cloudflare Pages)


ilyasotkov:
### Reproduction

Go to https://github.com/ilyasotkov/polished-truth-b001, open in Codespaces or local devcontainer, follow the steps in README.

This pull request https://github.com/remix-run/remix/pull/6783 is the culprit, present since 1.19.0, worked fine before that. The error seems quite baffling to me because the code below is so basic, I can't imagine any useful Remix app not doing something like this. Am I missing something here? Any workarounds?

```ts
import { Form } from '@remix-run/react';

export const loader = async () => {
  const r = new Request(`https://httpbin.org/get`);
  return await fetch(r);
};

export const action = async () => {
  return null;
};

export default function ButtonForm() {
  return (
    <div>
      <Form method="post">
        <button type="submit">Send Request</button>
      </Form>
    </div>
  );
}
```

### System Info

```shell
System:
    OS: Linux 6.2 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (2) x64 AMD EPYC 7763 64-Core Processor
    Memory: 5.77 GB / 7.74 GB
    Container: Yes
    Shell: 5.2.15 - /bin/bash
  Binaries:
    Node: 18.20.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.5.0 - /usr/local/bin/npm
  npmPackages:
    @remix-run/cloudflare: ^2.8.1 => 2.8.1 
    @remix-run/cloudflare-pages: ^2.8.1 => 2.8.1 
    @remix-run/dev: ^2.8.1 => 2.8.1 
    @remix-run/react: ^2.8.1 => 2.8.1 
    vite: ^5.1.0 => 5.2.8 
```
```


### Used Package Manager

npm

### Expected Behavior

No error on button press.

### Actual Behavior

✘ [ERROR] TypeError: Can't modify immutable headers.

      at handleDataRequestRR
  (file:///projects/apps/polished-truth-b001/node_modules/@remix-run/server-runtime/dist/server.js:158:22)
      at async requestHandler
  (file:///projects/apps/polished-truth-b001/node_modules/@remix-run/server-runtime/dist/server.js:94:18)
      at async handleFetch
  (file:///projects/apps/polished-truth-b001/node_modules/@remix-run/cloudflare-pages/dist/esm/worker.js:75:18)
[...]

PR #9693, opened 2024-07-02
Fix native fetch immutable headers issue

### LINKED ISSUE #9691: Environment variables not working in dev mode when navigating to a different route
### Reproduction

Reproduction: https://github.com/nareshbhatia/remix-nested-layouts

```shell
npm ci
npm run dev
```

1. Point your browser to http://localhost:3000/
2. Click on the `Movies` tab. App navigates to http://localhost:3000/movies. However it crashes with error: `TypeError: immutable`.
3. Refresh the browser without changing the URL. Page refreshes correctly showing a list of movies.

This indicates that the loader is reading the `API_URL` environment variable correctly for full page render, however when navigating from page to page, it can't read the environment variable.

This problem does not exist in production mode. See prod deployment at https://remix-nested-layouts-coral.vercel.app/

### System Info

```shell
System:
    OS: macOS 13.6.7
    CPU: (4) x64 Intel(R) Core(TM) i5-7600K CPU @ 3.80GHz
    Memory: 55.42 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    Watchman: 2022.11.14.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 126.0.6478.127
    Safari: 17.5
  npmPackages:
    @remix-run/dev: ^2.10.0 => 2.10.0
    @remix-run/node: ^2.10.0 => 2.10.0
    @remix-run/react: ^2.10.0 => 2.10.0
    @remix-run/serve: ^2.10.0 => 2.10.0
    vite: ^5.3.2 => 5.3.2
```

### Used Package Manager

npm

### Expected Behavior

I should be able to navigate to any page, or click on various movies in the movie list and the app should not crash (in dev mode). The basic issue is that environment variables are not being read when navigating from page to page.

### Actual Behavior

App crashes when navigating from page to page.

brophdawg11:
Fixes #9691

### DIFF
--- a/packages/remix-server-runtime/__tests__/server-test.ts
+++ b/packages/remix-server-runtime/__tests__/server-test.ts
@@ -153,6 +153,48 @@ describe("shared server runtime", () => {
       expect(resourceLoader.mock.calls.length).toBe(1);
     });
 
+    test("calls resource route loader throwing response with immutable headers", async () => {
+      let build = mockServerBuild({
+        root: {
+          default: {},
+        },
+        "routes/resource": {
+          parentId: "root",
+          path: "resource",
+          loader() {
+            let headers = new Headers({ "x-test": "yes" });
+            let headersProxy = new Proxy(headers, {
+              get(target, prop, receiver) {
+                if (prop === "set") {
+                  throw new TypeError("immutable");
+                }
+                return Reflect.get(target, prop, receiver);
+              },
+            });
+            // Mock a "response" that will pass the `isResponse` check
+            throw {
+              status: 400,
+              statusText: "Bad Request",
+              headers: headersProxy,
+              body: "text",
+              text: () => Promise.resolve("text"),
+            };
+          },
+        },
+      });
+      let handler = createRequestHandler(build, ServerMode.Development);
+
+      let request = new Request(`${baseUrl}/resource`, {
+        method: "get",
+      });
+
+      let result = await handler(request);
+      expect(result.status).toBe(400);
+      expect(result.headers.get("x-test")).toBe("yes");
+      expect(result.headers.get("X-Remix-Catch")).toBe("yes");
+      expect(await result.text()).toBe("text");
+    });
+
     test("calls sub resource route loader", async () => {
       let rootLoader = jest.fn(() => {
         return "root";
@@ -612,7 +654,7 @@ describe("shared server runtime", () => {
       expect(indexLoader.mock.calls.length).toBe(1);
     });
 
-    test("data request calls loader returning raw fetch response", async () => {
+    test("data request calls loader returning response with immutable headers", async () => {
       let build = mockServerBuild({
         root: {
           default: {},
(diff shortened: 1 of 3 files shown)


packages/remix-server-runtime/__tests__/server-test.ts
```diff
@@ -612,6 +612,35 @@ describe("shared server runtime", () => {
...
+    test("data request calls loader returning raw fetch response", async () => {
+      let build = mockServerBuild({
+        root: {
+          default: {},
+        },
+        "routes/_index": {
+          parentId: "root",
+          index: true,
+          loader() {
+            return fetch(`https://remix.run/docs/en/main?_data=root`);
```
brophdawg11, line 624: I don't really want to keep this here but the only way to replicate this issue was to actually get a legit `Response` from a `fetch` call which is why we didn't run into this before.  Thoughts on how to best prevent a regression in this area without spamming our prod server?

packages/remix-server-runtime/server.ts
```diff
@@ -360,12 +360,12 @@ async function handleDataRequest(
 
     // Mark all successful responses with a header so we can identify in-flight
     // network errors that are missing this header
-    response.headers.set("X-Remix-Response", "yes");
```
brophdawg11, line 363: If this response is a direct `undici` `fetch` `Response`, then the headers are immutable and we can't change them.  I didn't want to clone the response in 100% of scenarios since this isn't an issue with `return json()` and alternatives, so I am only cloning if this operation throws the `immutable` error

myfrontpocket:
So, I think this is the same issue here: https://github.com/remix-run/remix/issues/9212

I've updated to 2.10.2, and this does restore the previous behavior, with the exception that returning the error response results in a framework exception (that I suppose would be caught by an ErrorBoundary).  Shouldn't this only happen if we `throw` the response rather than `return` it?

```
const response = await fetch('some-url.com')

if (!response.ok) {
  return response; // this acts a throw
}

// continue processing...
```

brophdawg11:
I do not see that behavior.  Please open a ne wissue with a reporduction and we can take a look.  But my quick attempt works as expected:

```jsx
// routes/page.tsx
import { useLoaderData } from "@remix-run/react";

  const response = await fetch("http://localhost:5173/resource");

  if (!response.ok) {
    return response;
  }

  return { message: "from loader" };
}

export default function Comp() {
  const data = useLoaderData();
  return <p>data: {data?.message}</p>;
}

// routes/resource.tyx

export function loader() {
  return json({ message: "from resource route" }, { status: 500 });
}
```

myfrontpocket:
> I do not see that behavior. Please open a ne wissue with a reporduction and we can take a look. But my quick attempt works as expected:
> 
> ```js
> // routes/page.tsx
> import { useLoaderData } from "@remix-run/react";
> 
> export async function loader() {
>   const response = await fetch("http://localhost:5173/resource");
> 
>   if (!response.ok) {
>     return response;
>   }
> 
>   return { message: "from loader" };
> }
> 
> export default function Comp() {
>   const data = useLoaderData();
>   return <p>data: {data?.message}</p>;
> }
> 
> // routes/resource.tyx
> import { json } from "@remix-run/node";
> 
> export function loader() {
>   return json({ message: "from resource route" }, { status: 500 });
> }
> ```

Hmmm... is it the same for an action function?  Not sure if it'll make a difference. This is where I'm currently seeing it happen

Issue #9692, opened 2024-07-02
double slash followed by any char after domain causes infinite page reloading


dimanche-matin:
### Reproduction

Add double slash followed by any char after domain (e.g. [https://example.com//foo]()), then Enter.

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1
    Memory: 59.33 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.7.1 - /opt/homebrew/bin/node
    Yarn: 1.22.21 - /opt/homebrew/bin/yarn
    npm: 10.5.0 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 125.1.66.115
    Chrome: 126.0.6478.127
    Edge: 126.0.2592.81
    Safari: 17.5
  npmPackages:
    @remix-run/dev: ^2.9.2 => 2.9.2 
    @remix-run/node: ^2.9.2 => 2.9.2 
    @remix-run/react: ^2.9.2 => 2.9.2 
    @remix-run/serve: ^2.9.2 => 2.9.2 
    vite: ^5.1.0 => 5.2.13
```

### Used Package Manager

npm

### Expected Behavior

404 Not Found

### Actual Behavior

Infinite reloading with this console error log:
"Initial URL (/) does not match URL at time of hydration (//foo), reloading page..."

Some web sites listed on https://remix.run/showcase are affected.

brophdawg11:
This should be resolved via #9695 and available in the next release 👍

dimanche-matin:
Bug persists on `2.11.0-pre.0`.

When navigating on `//foo` path :

On `2.9.2` :

On client :
<img width="485" alt="remix_2 9 2_p1_client_crop-min" src="https://github.com/user-attachments/assets/27a7cdcf-8a8a-4574-89d1-70df77ccb2ec">

On server :
Empty

On `2.11.0-pre.0` :

On client :
<img width="487" alt="remix_2 11 0-pre 0_p1_client_crop-min" src="https://github.com/user-attachments/assets/a7a72438-7ece-4b0d-8895-147563919291">

On server :
Empty

When navigating on `/foo//bar` path :

On `2.9.2` :

404 Not Found

On `2.11.0-pre.0` :

On client :
[...]

Issue #8872, opened 2024-02-23
encoded character in url causes "Initial URL does not match URL at time of hydration"


WillSmithTE:
### Reproduction

https://github.com/WillSmithTE/grunge-test-2

https://19przkzqw9.execute-api.us-west-2.amazonaws.com

1. open the app above
2. open the whistler blog
3. infinite refresh because "Initial URL does not match URL at time of hydration"

This is deployed with the grunge stack on lambda, and the whistler blog title (and url) has an encoded character, "," (or "%2C")

in the html, this is the remix context, note the "decoded" comma

```
<script>
            window.__remixContext = {
                "url": "/blog/Good%20day%20at%20Whistler,%20but%20a%20bit%20icy",
                "state": {
                    "loaderData": {
                        "root": {
                            "user": null
                        },
                        "routes/blog.$title": {
                            "blog": {
                                "id": "a",
                                "title": "Good day at Whistler, but a bit icy"
                            }
                        }
                    },
                    "actionData": null,
                    "errors": null
                },
                "future": {
                    "v3_fetcherPersist": false,
                    "v3_relativeSplatPath": false,
                    "v3_throwAbortReason": false
                }
            };
        </script>
        ```

https://github.com/remix-run/remix/assets/26863411/091a8e6d-cd1e-4a07-9727-bcddb5e449be

### System Info

```shell
System:
    OS: macOS 14.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 130.39 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.0/bin/npm
    Watchman: 2023.12.04.00 - /usr/local/bin/watchman
  Browsers:
    Brave Browser: 121.1.62.153
    Chrome: 121.0.6167.184
    Safari: 17.1
  npmPackages:
    @remix-run/architect: ^2.7.2 => 2.7.2 
    @remix-run/css-bundle: ^2.7.2 => 2.7.2 
    @remix-run/dev: ^2.7.2 => 2.7.2 
    @remix-run/node: ^2.7.2 => 2.7.2 
    @remix-run/react: ^2.7.2 => 2.7.2 
    vite: ^5.0.0 => 5.1.4
```

### Used Package Manager

npm

[...]

PR #9695, opened 2024-07-02
Change hydration check from URL to matches

brophdawg11:
The original issue this check was fixing was https://github.com/remix-run/remix/issues/1757 - see about reproducing that in today's landscape - maybe the original infinite loop is no longer an issue? 

**Update:** I'm having trouble reproducing the original issue so I'm inclined to remove this check from RR v7 entirely.  But for now, switching to matches should make it less error-prone.

Closes https://github.com/remix-run/remix/issues/9692, https://github.com/remix-run/remix/issues/8872, https://github.com/remix-run/remix/issues/7529

### DIFF
--- a/packages/remix-react/browser.tsx
+++ b/packages/remix-react/browser.tsx
@@ -61,7 +61,6 @@ let stateDecodingPromise:
   | undefined;
 let router: Router;
 let routerInitialized = false;
-let alreadyReloadedIgnoreStrictModeRerender = false;
 let hmrAbortController: AbortController | undefined;
 let hmrRouterReadyResolve: ((router: Router) => void) | undefined;
 // There's a race condition with HMR where the remix:manifest is signaled before
@@ -195,10 +194,6 @@ if (import.meta && import.meta.hot) {
  */
 export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
   if (!router) {
-    if (alreadyReloadedIgnoreStrictModeRerender) {
-      return <></>;
-    }
-
     // When single fetch is enabled, we need to suspend until the initial state
     // snapshot is decoded into window.__remixContext.state
     if (window.__remixContext.future.unstable_singleFetch) {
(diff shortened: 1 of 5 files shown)


calumpeak:
@brophdawg11 I just bumped to the latest version of Remix Run (2.11.0) and my application fails to load with a hydration error as `ssrMatches.length` is undefined. Just flagging this here as I believe this is the PR that introduces that check - cleared the cache, and everything. It may be that other users have this issue. I'll be rolling back to stable

```
client.ts:19 [vite] connecting...
chunk-D74W65WU.js?v=d1ff1c99:6884 Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at RemixBrowser (chunk-D74W65WU.js?v=d1ff1c99:6884:81)
    at renderWithHooks (chunk-7NZKRGB6.js?v=d1ff1c99:12171:26)
    at mountIndeterminateComponent (chunk-7NZKRGB6.js?v=d1ff1c99:14921:21)
    at beginWork (chunk-7NZKRGB6.js?v=d1ff1c99:15902:22)
    at HTMLUnknownElement.callCallback2 (chunk-7NZKRGB6.js?v=d1ff1c99:3674:22)
    at Object.invokeGuardedCallbackDev (chunk-7NZKRGB6.js?v=d1ff1c99:3699:24)
    at invokeGuardedCallback (chunk-7NZKRGB6.js?v=d1ff1c99:3733:39)
    at beginWork$1 (chunk-7NZKRGB6.js?v=d1ff1c99:19761:15)
    at performUnitOfWork (chunk-7NZKRGB6.js?v=d1ff1c99:19194:20)
    at workLoopConcurrent (chunk-7NZKRGB6.js?v=d1ff1c99:19185:13)
```
image

packages/remix-react/browser.tsx
```diff
@@ -270,6 +247,35 @@ export function RemixBrowser(_props: RemixBrowserProps): ReactElement {
...
+
+      if (hasDifferentSSRMatches && !window.__remixContext.isSpaMode) {
+        let ssr = ssrMatches.join(",");
+        let client = (initialMatches || []).map((m) => m.route.id).join(",");
+        let errorMsg =
+          `SSR Matches (${ssr}) do not match client matches (${client}) at ` +
+          `time of hydration , reloading page...`;
+        console.error(errorMsg);
+
+        window.location.reload();
```
ngbrown, line 272: This window reload seems to be creating an infinite reload loop on Remix 2.11.0 on a 404 page. With or without a custom `ErrorBoundary`. See #9821.

It's hard to break the browser and get it into debug mode when it's doing this, but it seems that it's because `hasDifferentSSRMatches` is set to `true` because `(initialMatches || []).length !== ssrMatches.length` is true. `initialMatches` is `0`, because `matchRoutes` doesn't match a 404 to the `root` route? while `ssrMatches` is length `1` with the `root` route.

PR #9696, opened 2024-07-02
Properly bind cloudflare workers’ ctx.waitUntil and ctx.passThroughOnException

acusti:
follow-up to #9345

this fixes usage of the [cloudflare context API’s](https://developers.cloudflare.com/workers/runtime-apis/context/) `waitUntil` and `passThroughOnException` utils. without this, invoking:

```js
context.cloudflare.ctx.waitUntil(doAsyncWork());
```

results in `TypeError: Illegal invocation` when run in a remix app deployed to cloudflare workers. the version of `waitUntil` passed to `getAssetFromKV` also has to be bound properly, so i just hoisted the binding so that the correct version of the function can you used both with `getAssetFromKV` and to build the `loadContext` that gets passed to `handleRemixRequest`.

Testing Strategy:

i tested my remix app both locally (using `remix vite:dev`) and in the deployed workers version of the app, and using the bound version of `waitUntil` ensures that the app works in both environments.


brophdawg11:
Would you mind signing the [CLA](https://github.com/remix-run/remix/pull/9696#issuecomment-2204390762)?

brookslybrand (approved):
Thanks @acusti. I've confirmed that this is a problem both when running wrangler and when deployed. This change fixes the issue

Once you sign the CLA we'll merge this in 👍

acusti:
@brookslybrand @brophdawg11 thanks for looking into it, and sorry for the CLA delay. i signed it in a different PR so i thought i’d already done it, but i forgot that it hadn’t been merged, so i just cherry-picked the commit over.

PR #9700, opened 2024-07-03
fix publicPath joining with module paths to support absolute publicPath

asmadsen:
Currently if you set the `base` property in `vite.config.ts` to an full url (ie `http://localhost:5173/`), when running `vite:dev` the _modulepreload_ link tags remove the second slash(`/`) in `http://local...`.

Current behaviour:
```
<link rel="modulepreload" href="http:/localhost:5173/app/entry.client.tsx" crossorigin="anonymous">
```

Expected behaviour:
```
<link rel="modulepreload" href="http://localhost:5173/app/entry.client.tsx" crossorigin="anonymous">
```

This is not the case when first running `vite:build` then serving the output with `remix-serve ./build/server/index.js`, then it behaves as expected.

I also checked the classic remix compiler, it also behaves as expected.

# Testing Strategy:

> In a fresh remix project, I pointed `@remix/run` to the local path of this MR using `pnpm link`.
> Making sure the `base` property inside `vite.config.ts` was set to the origin of the local dev server (ie `http://localhost:5173/`)
>
> ```
> pnpm create remix my-test
> cd my-test
> pnpm run dev
> ```
>
> Then I verified that the _modulepreload_ link href works as the expected behaviour above.

I'm a bit unsure if all of the places should be changed, however since they all are based on the same path I think this should be correct. And in the limited testing I did there where no issues.


packages/remix-dev/vite/plugin.ts
```diff
@@ -977,21 +977,21 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
...
     return {
       version: String(Math.random()),
-      url: path.posix.join(
-        ctx.remixConfig.publicPath,
-        VirtualModule.url(browserManifestId)
-      ),
+      url: new URL(
+        VirtualModule.url(browserManifestId),
+        ctx.remixConfig.publicPath
+      ).href,
```
brophdawg11, line 983: I don't think `ctx.remixConfig.publicPath` is guaranteed to be an absolute URL here, is it?  If it's not, this will throw an exception:

```sh
> node
Welcome to Node.js v20.12.2.
Type ".help" for more information.
> new URL("file.txt", "/public")
Uncaught TypeError: Invalid URL
    at new URL (node:internal/url:796:36) {
  code: 'ERR_INVALID_URL',
  input: 'file.txt',
  base: '/public'
}
```

packages/remix-dev/vite/plugin.ts
```diff
@@ -40,6 +40,12 @@ import { resolveFileUrl } from "./resolve-file-url";
 import { removeExports } from "./remove-exports";
 import { importViteEsmSync, preloadViteEsm } from "./import-vite-esm-sync";
 
+function combineURLs(baseURL: string, relativeURL: string) {
+  return relativeURL
+      ? baseURL.replace(/\/+$/, '') + '/' + relativeURL.replace(/^\/+/, '')
+      : baseURL;
```
brophdawg11, line 46: Could you add a comment with some examples of why this approach is needed for future context if folks are in this file?

markdalgleish: I've added some unit and integration tests to capture this instead.

markdalgleish (approved):
I've added some tests, but otherwise this looks great to me. Thanks for the PR!

Issue #9717, opened 2024-07-05
`useLoaderData` returns `undefined` when used in `root.tsx`


npapagna:
### Reproduction

1. Go to https://stackblitz.com/edit/remix-run-remix-cuwoyi?file=app%2Froutes%2F_index.tsx
2. Notice that both `root.tsx` and `_index.tsx` render `undefined` as the output of the `useActionData` hook (see rendered messages on the page.
3. Click the `Click to submit` button to call the `clientAction` form the `_index.tsx` route.
4. Notice that `useActionData` returns `{toast: "foo"}` in the `_index.tsx` route, but returns `undefined` in `root.tsx` (see rendered messages on the page.

### System Info

```shell
System:
    OS: macOS 13.6.7
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 25.92 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - ~/.nvm/versions/node/v20.11.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.16.0/bin/yarn
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.1/bin/npm
    pnpm: 8.14.0 - /usr/local/bin/pnpm
  Browsers:
    Brave Browser: 124.1.65.126
    Chrome: 126.0.6478.127
    Safari: 16.6
  npmPackages:
    @remix-run/dev: ^2.9.2 => 2.9.2 
    @remix-run/node: ^2.9.2 => 2.9.2 
    @remix-run/react: ^2.9.2 => 2.9.2 
    vite: ^5.1.0 => 5.2.12
```

### Used Package Manager

npm

### Expected Behavior

I was expecting to access the (the serialized data from the most recent route action)[https://remix.run/docs/en/main/hooks/use-action-data] in `root.tsx` to implement an `Toasts` component.

The `clientAction` on `_index.tsx` would do stuff and return `{toast: "message"}` as part of the response (other routes do this as well) and in `root.tsx` I wanted to render the toasts (when present).

This is a Remix SPA project (that's why we use a `clientAction` instead of `action`).

### Actual Behavior

`useActionData` returns the `clientAction` response only in the route where the client action is defined.

kiliman:
In Remix, the _actual behavior_ is the _correct behavior_. `useActionData` only returns the action response from the route it's defined in.

[...]

PR #9744, opened 2024-07-12
Docs: Clarify `useActionData` Hook Scope

npapagna:
The current documentation for the `useActionData` hook might unintentionally imply that it can be used to access data from any route action. However, the hook is limited to accessing data associated with the current route action where it's called.

This PR clarifies this limitation in the documentation to prevent confusion.

See #9717 for more information.


brophdawg11 (approved):
Thanks!

PR #9753, opened 2024-07-15
feat(remix-serve): Allow opting out of using `source-map-support`

lforst:
This PR adds a way to opt out of using `source-map-support` when using `remix-serve`, which can cause issues with certain observability providers like Sentry.

Closes: https://github.com/remix-run/remix/discussions/9752

- [x] Docs
- [ ] Tests

Testing Strategy:

I had a hard time coming up with a way to test this in the existing testing framework. I am not too familiar with the code base. If someone can point me into the right direction I am happy to add tests! :)

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -383,6 +383,7 @@
 - levippaul
 - LewisArdern
 - lfantone
+- lforst
 - lifeiscontent
 - lili21
 - lionotm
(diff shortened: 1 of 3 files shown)


brophdawg11:
Thank you for the PR but [`remix-serve`](https://remix.run/docs/en/main/other-api/serve) is intentionally opinionated/unconfigurable so it can be the simplest possible solution for folks who don't want to do any customizations.  

There is already a way to opt-out of source map support by using the express adapter with your own `server.js` file which you can customize as needed.

> It's a production-ready, but basic Node.js server built with Express. If you find you want to customize it, use the @remix-run/express adapter instead.

PR #9756, opened 2024-07-15
Add a failing test for a GET request with a body

blittle:
Add a failing test for when there is a GET request with a body. The request throws, which produces a 500 error in prod. Instead it should respond with a 400 bad request.

The issue is Remix Server Runtime has [a function that proxies](https://github.com/remix-run/remix/blob/f51ec665c3adbc24548310c5f6bb43f67057458a/packages/remix-server-runtime/data.ts#L153) the original request object and changes it into a new request object. This throws an error, which results in a 500 level error from Remix. While it is proper to error, I don't think this should be a 500 level error. It is a bad request from the client, and the client should get a 400 level error. 500 means something bad went wrong on the server, or something we did wrong. 400 means something the client did wrong (like send a body in a GET).  Often time 500 vs 400 server errors also effect SLAs. We have a guaranteed "uptime", and 500s showing up imply that the server went down, which it didn't.

Closes: #

- [ ] Docs
- [x] Tests

Testing Strategy:

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -76,6 +76,7 @@
 - benwis
 - bgschiller
 - binajmen
+- blittle
 - bmac
 - bmarvinb
 - bmontalvo
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -99,21 +99,13 @@ test.afterAll(() => {
 ////////////////////////////////////////////////////////////////////////////////
 
 test("[description of what you expect it to do]", async ({ page }) => {
-  let app = new PlaywrightFixture(appFixture, page);
-  // You can test any request your app might get using `fixture`.
-  let response = await fixture.requestDocument("/");
-  expect(await response.text()).toMatch("pizza");
-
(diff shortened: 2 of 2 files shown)


brophdawg11:
FWIW - this error may surface in different spots based on adapters.  In cloudflare deployments I think it happens in there since cloudflare gives you a `fetch` `Request` out of the box.  But I think if this was in an express app, it would throw at the adapter layer when converting the express `req` to a `fetch` `Request`.  So we may need checks in more than one spot.

brophdawg11:
Chatted with Bret and we think this needs to be fixed at the server layer.  

When using Express, it's actually impossible to get into this state because we never even include any `body` on GET request creation at the adpter layer: https://github.com/remix-run/remix/blob/main/packages/remix-express/server.ts#L110

So it's unclear just how a fetch `Request` is being created in Bret's case since it should be invalid to create one in that state in the first place. 

Bret is planning to fix this at the server layer prior to handing off to Remix.

PR #9758, opened 2024-07-15
Include .client/.server directories in Tailwind globs

brettscott:
## Problem
Tailwind does not work when a component with Tailwind classes is within a `/.client/` folder.

## Reproduction
https://stackblitz.com/edit/remix-run-remix-jetmkv?file=tailwind.config.ts

- Remove `'./app/**/.client/**/*.{js,jsx,ts,tsx}',` from `tailwind.config.ts` and the styles within `TextOnClientOnly.tsx` will no longer be processed by Tailwind.

## Same for .server ?
My assumption is that it's not necessary / needed for `/.server/` server-only components 🤷


markdalgleish (commented):
Thanks for the PR!

We consider `.client` modules to be an edge-case feature for keeping modules with browser-only side effects out of the server. We didn't intend for them to be used for components and we expect that the vast majority of projects will never need this feature.

What's the real-world scenario you need this for? Is there any downside to moving the UI code out of a `.client` module?

brettscott:
Hey @markdalgleish, 

I came across this when attempting to migrate the [Lexical Playground](https://playground.lexical.dev/) ([repo](https://github.com/facebook/lexical/tree/main/packages/lexical-playground)) to Remix.  

The playground  has an Excalidraw plugin using  [Excalidraw](https://github.com/excalidraw/excalidraw) virtual whiteboard, which fails to server-render.

It took longer than I expected to work out why component styles were not being rendered by Tailwind, hence this PR to save other's time.

markdalgleish (approved):
I found a way to include `.client` and `.server` directories without needing to provide multiple globs, which was my main objection. I've updated the PR to use this format. Now that it's more streamlined, it seems like a no-brainer to include it even if it's a bit of an edge case, especially when it comes to `.server` directories.

Thanks again for the PR!

brettscott:
@markdalgleish your approach looks slick - thanks!

PR #9763, opened 2024-07-16
Rename future flag from fogOfWar to lazyRouteDiscovery

brophdawg11:
The "fog of war" phrasing is a nice metaphor for this and works well in blog posts but it's pretty confusing to see as a future flag without the proper context or video game familiarity.  We're going to rename the flag for clarity and we'll still use the "fog of war" metaphor in docs and blog posts etc.

### DIFF
--- a/.changeset/happy-dots-sleep.md
+++ b/.changeset/happy-dots-sleep.md
@@ -1,0 +1,8 @@
+---
+"@remix-run/dev": minor
+"@remix-run/react": minor
+"@remix-run/server-runtime": minor
+"@remix-run/testing": minor
+---
+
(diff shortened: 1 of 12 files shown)


.changeset/happy-dots-sleep.md
```diff
@@ -0,0 +1,8 @@
+---
+"@remix-run/dev": minor
+"@remix-run/react": minor
+"@remix-run/server-runtime": minor
+"@remix-run/testing": minor
```
brophdawg11, line 5: Did `minor` here instead of `patch` since I'm not considering this a change to a current flag but moreso the `deletion` of a current flag and the introduction of a new one.

PR #9772, opened 2024-07-18
docs: Add a note about Single Fetch to the headers documentation

tmcw:
This is a documentation-only PR that adds a note about the upcoming Single Fetch feature to the `headers` reference page.

It might be confusing to consult this documentation and not understand why it isn't working - I think it's helpful to add a note about how the `headers` export is on its way to deprecation.

- [x] Docs


brophdawg11:
Based on the latest roadmap planning this won't e the case anymore :). Stay tuned for updates in an upcoming release.

PR #9774, opened 2024-07-18
Convert cloudflare workers template’s main entry file from JS → TS

acusti:
follow-up to https://github.com/remix-run/remix/pull/9345

This enables type checking of the main workers entry file in the cloudflare workers template by converting it from `.js` to `.ts` and by fixing an incompatibility with the `AppLoadContext` type that is surfaced by the new type checking. Note that I also renamed `server.js` → `worker.ts`, but I’m happy to rename it back to `server.ts` if that is preferred. Mainly, I just saw it as an opportunity to slightly simplify the migration from the classic compiler cloudflare worker setup to the vite-based one. Lastly, I removed the redundant `SCRIPT` argument from the `wrangler dev` command (it’s already defined as the `main` file via `wrangler.toml`; [reference here](https://developers.cloudflare.com/workers/wrangler/commands/#dev)). Lastly, I added `// eslint-disable-line import/no-unresolved` to `import * as build from "./build/server";` because it was causing a lint error from the [CI lint check](https://github.com/remix-run/remix/actions/runs/9995968691/job/27629268955), and I added `// @ts-ignore This file won’t exist if it hasn’t yet been built` above it because running `tsc` before building the app in my CI was causing a “Cannot Find Module” error.

**Testing Strategy:**

I converted the `main` file of my remix cloudflare workers app from `.js` to `.ts` and resolved the resulting typescript errors that were produced. All the changes are types-only, so there is no effect on runtime behavior.


markdalgleish (approved):
Thanks for the PR!

Just a heads up that I've rolled the `worker.ts` rename back to `server.ts` mainly just to keep it in line with the old template.

Issue #6507, opened 2023-05-29
Submitting empty `Form` on chrome throws error


akamfoad:
### What version of Remix are you using?

1.16.1

### Are all your remix dependencies & dev-dependencies using the same version?

- [X] Yes

### Steps to Reproduce

1. Create a Remix App
2. Create a form with no input but a submit button
3. click the submit button

Or you can see this [reproduction app](https://stackblitz.com/edit/remix-run-remix-phwauf?file=app%2Froutes%2F_index.tsx).

Notice when the checkbox is checked submitting works fine, but when its unchecked, it fails to submit and crashes.

This only happens on Chrome, it works fine on Firefox.

Also using `form.submit` still works, `form.requestSubmit` or normal submit causes the crash to happen.

### Expected Behavior

To submit the form and send the request to the action regardless of what browser I use.

### Actual Behavior

The `Form` crashes with this error stack

```
TypeError: Failed to fetch
    at fetchData (https://remixrunremixphwauf-t1s0--3000--272d3407.local-credentialless.webcontainer.io/build/_shared/chunk-Q6KF7EY5.js:10638:139)
    at https://remixrunremixphwauf-t1s0--3000--272d3407.local-credentialless.webcontainer.io/build/_shared/chunk-Q6KF7EY5.js:10857:26
    at runHandler (https://remixrunremixphwauf-t1s0--3000--272d3407.local-credentialless.webcontainer.io/build/_shared/chunk-Q6KF7EY5.js:5992:26)
    at callLoaderOrAction (https://remixrunremixphwauf-t1s0--3000--272d3407.local-credentialless.webcontainer.io/build/_shared/chunk-Q6KF7EY5.js:6031:22)
    at handleAction (https://remixrunremixphwauf-t1s0--3000--272d3407.local-credentialless.webcontainer.io/build/_shared/chunk-Q6KF7EY5.js:5016:22)
    at startNavigation (https://remixrunremixphwauf-t1s0--3000--272d3407.local-credentialless.webcontainer.io/build/_shared/chunk-Q6KF7EY5.js:4960:32)
    at Object.navigate (https://remixrunremixphwauf-t1s0--3000--272d3407.local-credentialless.webcontainer.io/build/_shared/chunk-Q6KF7EY5.js:4890:18)
    at https://remixrunremixphwauf-t1s0--3000--272d3407.local-credentialless.webcontainer.io/build/_shared/chunk-Q6KF7EY5.js:8058:15
[...]

PR #9778, opened 2024-07-19
Bug: `Form` crash on submit with empty multipart form data

aaronadamsCA:
The combination of `<Form>`, `encType="multipart/form-data"`, and an empty form body is crashing the Remix client on submit with error `TypeError: Failed to fetch`.

This PR provides eight tests, seven passing and one failing.

Multipart form encoding is a sensible default for a larger app where many, but not all, forms contain file inputs. A few of our smaller forms are all checkboxes, and eventually we figured out they were crashing if everything was unchecked. We eventually identified `<Form>` as the culprit.

For now we can work around this by setting `encType="application/x-www-form-urlencoded"` on our "oops all checkboxes" forms.

---

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/integration/bug-report-test.ts
+++ b/integration/bug-report-test.ts
@@ -58,28 +58,104 @@ test.beforeAll(async () => {
     // `createFixture` will make an app and run your tests against it.
     ////////////////////////////////////////////////////////////////////////////
     files: {
-      "app/routes/_index.tsx": js`
-        import { json } from "@remix-run/node";
-        import { useLoaderData, Link } from "@remix-run/react";
+      "app/routes/html-form.tsx": js`
+        import { json, useActionData } from "@remix-run/react";
 
-        export function loader() {
+        export async function action({ request }) {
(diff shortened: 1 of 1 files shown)


brophdawg11:
I think this is a dup of https://github.com/remix-run/remix/issues/6507 which looks to be a bug in Chrome.  For now the recommended workaround is to include a hidden input in your forms to avoid this specific issue

aaronadamsCA:
Thanks @brophdawg11, obviously I did not find that previously closed issue. I truly cannot believe this is an error in Chromium:

```js
new Request("/", { method: "post", body: new FormData() }).formData();
```

I guess there's nothing Remix can do at the framework level here, since the `TypeError` is opaque and the form data can't be examined. I guess we'll just need to work around it!

brophdawg11:
yeah it took some digging on our end to track that down as well!

PR #9791, opened 2024-07-24
fix: update pending-ui.md starred icons

Pouet--:
I believe the starred icons has been swapped in the example


brophdawg11 (approved):
Thanks!

Issue #9498, opened 2024-05-28
Build fails when using vite-imagetools plugin due to assets already existing in build/client


ivosabev:
I get the following error when building Remix. 

```
Error: [remix] dest already exists.
    at /Users/xxx/site/node_modules/fs-extra/lib/move/move.js:51:31
    at /Users/xxx/site/node_modules/universalify/index.js:22:54 {
  code: 'PLUGIN_ERROR',
  plugin: 'remix',
  hook: 'writeBundle'
}
ERROR: "build-remix" exited with 1.
```

It seems that there is a build step where `build/server/assets` are moved over to `build/client/assets`, but the images already exist there and the build fails due to the way `fs-extra/lib/move/move.js` behaves.

```
// fs-extra/lib/move/move.js
function doRename (src, dest, overwrite, isChangingCase, cb) {
  if (isChangingCase) return rename(src, dest, overwrite, cb)
  if (overwrite) {
    return remove(dest, err => {
      if (err) return cb(err)
      return rename(src, dest, overwrite, cb)
    })
  }
  pathExists(dest, (err, destExists) => {
    if (err) return cb(err)
    if (destExists) return cb(new Error('dest already exists.')) // the cause of the error
    return rename(src, dest, overwrite, cb)
  })
}
```

I am using `vite-imagetools` which causes the assets to already exist in `build/client/assets`.

When calling directly `vite build` everything works. So to me it seems that Remix should do a better job at moving the files from `build/server/assets` to `build/client/assets` and not fail.

justb3a:
I can verify the bug 😇  same on my side. As soon as I include the same image twice, remix build is failing.

```sh
x Build failed in 270ms
Error: [remix] dest already exists.
    at /Users/…/node_modules/fs-extra/lib/move/move.js:51:31
    at /Users/…/node_modules/universalify/index.js:21:38 {
  code: 'PLUGIN_ERROR',
  plugin: 'remix',
  hook: 'writeBundle'
}
```

remix packages @2.9.2
vite @5.2.11
vite-imagetools @7.0.2

alexandercarls:
I can confirm that this bug was introduced in Remix version [2.9.2](https://github.com/remix-run/remix/blob/main/CHANGELOG.md#patch-changes) likely with https://github.com/remix-run/remix/pull/9305

[...]

PR #9800, opened 2024-07-29
Fix  error when using vite-imagetool

tobigumo:
This PR resolves issue #9498 and is also related to PR #9305.

In PR #9305, the client asset list was changed to a file list. However, vite-imagetool generates assets during the build process, and these assets are not included in the file list, which seems to be causing the error.

I have modified the code to check both the asset list and the file list. This change appears to allow the build to complete without errors, and vite-imagetool seems to function correctly.

However, this is not a fundamental solution, and since I am not very familiar with Remix, I would appreciate any feedback on this PR.

Closes: #9498

- [ ] Docs
- [ ] Tests

Testing Strategy:

I didn't see any tests for the `plugin` file, so I didn't include them here.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -632,6 +632,7 @@
 - TimonVS
 - tjefferson08
 - tmcw
+- tobigumo
 - tombiju
 - tombohub
 - tombyrer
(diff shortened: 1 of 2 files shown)


markdalgleish (commented):
Thanks for the PR!

Since we've hit up against this issue a couple of times now, I've opted for a safer solution instead via #9901 where we use the file system as the source of truth rather than inferring it from the Vite manifest.

PR #9806, opened 2024-07-30
Add DISABLE_ROUTES_LOGS env to disable route logs

dchueri:
The logs that appear when a user accesses an Remix route in production make it difficult to locate other logs, especially when there are many simultaneous users.

image

Testing Strategy:
Simply adding DISABLE_ROUTES_LOGS to the envs will be enough to suppress just these logs.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -138,6 +138,7 @@
 - davecranwell-vocovo
 - DavidHollins6
 - davongit
+- dchueri
 - dcramer
 - Deanmv
 - defjosiah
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thank you for the PR but [remix-serve](https://remix.run/docs/en/main/other-api/serve) is intentionally opinionated/unconfigurable so it can be the simplest possible solution for folks who don't want to do any customizations.

If you wish to control the logging implementation (and anything else about the server), you can using the express adapter with your own `server.js` file.

> It's a production-ready, but basic Node.js server built with Express. If you find you want to customize it, use the @remix-run/express adapter instead.

mmacdonaldOCP:
While I understand the position of the team, I do think something like this would be helpful. For the project I've been working on I've been happy with the simplicity of the existing tools, but I really wish there was a way to disable the `morgan` logs while retaining the rest of the setup.

PR #9824, opened 2024-08-04
Replace placekitten service with placecats

iamajvillalobos:
## Description

It's been a few months since Placekitten went down. I've been looking for alternatives that still feature cats (there's a reason we use them! :)).

## Screenshot
Screenshot 2024-08-04 at 11 01 37 AM

### DIFF
--- a/docs/start/tutorial.md
+++ b/docs/start/tutorial.md
@@ -168,7 +168,7 @@ export default function Contact() {
   const contact = {
     first: "Your",
     last: "Name",
-    avatar: "https://placekitten.com/200/200",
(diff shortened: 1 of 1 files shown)


brophdawg11:
Thank you for the PR but placekitten is working fine for me - and this is just sample code so folks are free to change it to whatever they need while going through the tutorial.

Screenshot 2024-08-08 at 1 49 06 PM

jbwl:
I confirm that placekitten.com is down (Cloudflare App down warning) since weeks. Not reopening this though.

PR #9830, opened 2024-08-05
docs(data-loading): fix SearchCheckbox snippet

jackhkmatthews:
The suggested controlled SearchCheckbox component didn't work for me - when checking a checkbox it would flick to checked and then revert to unchecked.

I suspect this was the useEffect running unexpectedly, maybe because of the arrays and objects in the dependency array.

React docs: [Removing unnecessary object dependencies](https://github.com/remix-run/remix/blob/main/CONTRIBUTING.md).

This updates the required dependencies to a single string variable and seems to work better.

My first PR here, let me know what I've missed! 👍 🙏

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -276,6 +276,7 @@
 - izznatsir
 - jacargentina
 - jack-r-warren
+- jackhkmatthews
 - jacob-ebey
 - JacobParis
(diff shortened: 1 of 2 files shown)


brophdawg11 (approved):
Thanks!

PR #9834, opened 2024-08-05
fix #9826: support conditional return types from loader when serializing

### LINKED ISSUE #9826: useLoaderData types broken if unstable_data is conditionally used
### Reproduction

With the following loader:

```typescript
export const loader = unstable_defineLoader(async ({ request }) => {
  const isBot = isbot(request.headers.get('user-agent'));

  // Prevents making a session for every bot request
  if (isBot) {
    return { notification: null };
  }

  const session = await sessionStorage.getSession(request.headers.get('Cookie'));
  const notification = session.get('notification') ?? null;

  return unstable_data({ notification }, { 'Set-Cookie': await sessionStorage.commitSession(session) });
});
```

`const { notification } = useLoaderData<typeof loader>();` returns a type issue. The issue exists because of the conditional use of `unstable_data`. If I remove the "isBot" block, the type issue is gone.

Screenshot 2024-08-04 at 9 01 46 PM

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M2 Pro
    Memory: 2.85 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.16.0 - /opt/n/bin/node
    npm: 10.8.1 - /opt/n/bin/npm
    pnpm: 9.4.0 - /opt/n/bin/pnpm
    Watchman: 2024.06.17.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 127.0.6533.89
    Safari: 17.5
```

### Used Package Manager

pnpm

### Expected Behavior

No type issue. In the use case above, `notification` would be `NotificationType | null`

### Actual Behavior

Type issue attached above.

michaelgmcd:
Tested this in my own Remix project. See #9826 for more details.

Before:
Screenshot 2024-08-05 at 3 55 10 PM

After:
Screenshot 2024-08-05 at 3 56 05 PM

### DIFF
--- a/packages/remix-server-runtime/single-fetch.ts
+++ b/packages/remix-server-runtime/single-fetch.ts
@@ -387,16 +387,25 @@ type DataFunctionReturnValue =
   | TypedDeferredData<Record<string, unknown>>
   | TypedResponse<Record<string, unknown>>;
 
+type Unwrap<T extends DataFunctionReturnValue> = T extends TypedDeferredData<
+  infer D
+>
+  ? D
+  : T extends TypedResponse<Record<string, unknown>>
+  ? SerializeFrom<T>
+  : T extends DataWithResponseInit<infer D>
+  ? D
+  : T;
+
 // Backwards-compatible type for Remix v2 where json/defer still use the old types,
 // and only non-json/defer returns use the new types.  This allows for incremental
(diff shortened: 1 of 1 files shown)


pcattori:
Superceded by #9999

Issue #9884, opened 2024-08-18
Turbo responses are not compressed by Cloudflare


lilouartz:
### Reproduction

e.g. This page is hosted on Cloudflare https://pillser.com/vitamins/vitamin-b1.data

### System Info

```shell
Cloudflare
```

### Used Package Manager

npm

### Expected Behavior

I would expect response to be compressed.

### Actual Behavior

Response is not compressed.

The reason for this is because `content-type` is set to `text/x-turbo`, and this is not recognized as a cacheable value by Cloudflare.

https://developers.cloudflare.com/speed/optimization/content/brotli/content-compression/

## Possible solution

Change content type to `application/json` since it is valid JSON.

PR #9889, opened 2024-08-19
Change content type on single fetch requests

brophdawg11:
Closes https://github.com/remix-run/remix/issues/9884

### DIFF
--- a/.changeset/popular-meals-hide.md
+++ b/.changeset/popular-meals-hide.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/server-runtime": patch
+---
+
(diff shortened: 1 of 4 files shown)


jacob-ebey (approved):

ivogt:
Would you consider doing the same for defer? It has the same issue as single fetch.

brophdawg11:
I don't think defer needs updating because it's being deprecated and replaced with Single Fetch very soon.  But beyond that, it's a stable API so one could argue changing the content type is a breaking change so we could not make that change in v2 anyway.  You may be able to change it in userland via [`handleDataRequest`](https://remix.run/docs/en/main/file-conventions/entry.server#handledatarequest) though.

Issue #9776, opened 2024-07-19
Type instantiation is excessively deep and possibly infinite.ts(2589) when using readonly


punkpeye:
### Reproduction

https://stackblitz.com/edit/remix-run-remix-msgs25?file=app%2Froutes%2F_index.tsx

### System Info

```shell
N/A
```

### Used Package Manager

pnpm

### Expected Behavior

No type error.

### Actual Behavior

Getting error:

> Type instantiation is excessively deep and possibly infinite.ts(2589)

punkpeye:
The only code necessary to replicate is this:

```ts
export const loader = defineLoader(async () => {
  type RecentChat = {};

  const recentChats: readonly RecentChat[] = [];

  return {
    recentChats,
  };
});

const Route = () => {
  const data = useLoaderData<typeof loader>();
};
```

lilouartz:
@brophdawg11 I am facing the same issue. This is a blocker for adopting `turbo-stream`.

lilouartz:
I dug into types. Looks like fixing this is super simple:

```diff
- type Serializable = undefined | null | boolean | string | symbol | number | Array<Serializable> | 
+ type Serializable = undefined | null | boolean | string | symbol | number | Array<Serializable> | ReadonlyArray<Serializable> | 
{
    [key: PropertyKey]: Serializable;
} | bigint | Date | URL | RegExp | Error | Map<Serializable, Serializable> | Set<Serializable> | Promise<Serializable>;
```

punkpeye:
This fixed the issue!

lilouartz:
@brophdawg11 @jacob-ebey Can this me merged upstream?

At the moment, I have to patch local node_modules every time I update Remix.

brophdawg11:
`defineLoader` is going away in favor of a better type story soon so this will be resolved once that lands

punkpeye:
> `defineLoader` is going away in favor of a better type story soon so this will be resolved once that lands

That's great news, but apps that currently rely on `defineLoader` cannot lint properly.

brophdawg11:
It's an unstable API [not intended for production use](https://remix.run/docs/en/main/guides/single-fetch).  `// @ts-expect-error` is your friend here 😄

punkpeye:
What `defineLoader` and `defineAction` have been replaced with?

How do I type them?

0t4u:
> What `defineLoader` and `defineAction` have been replaced with?
> 
> How do I type them?

[...]

Issue #9714, opened 2024-07-05
The inferred type of 'loader' cannot be named without a reference to


vixalien:
### Reproduction

Install packages with `pnpm`

Try using the new `defineLoader` from unstable fetch.

make sure to add

```json
    "types": [
      "@remix-run/react/future/single-fetch.d.ts"
    ]
 ```
 
 to `tsconfig.json`

### System Info

```shell
System:
    OS: Linux 6.9 Alpine Linux
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 23.21 GB / 31.06 GB
    Container: Yes
    Shell: 5.2.26 - /bin/bash
  Binaries:
    Node: 22.1.0 - ~/.nvm/versions/node/v22.1.0/bin/node
    Yarn: 1.22.22 - ~/.local/share/pnpm/yarn
    npm: 10.7.0 - ~/.nvm/versions/node/v22.1.0/bin/npm
    pnpm: 9.4.0 - ~/.local/share/pnpm/pnpm
  npmPackages:
    @remix-run/dev: ^2.10.2 => 2.10.2 
    @remix-run/eslint-config: ^2.10.2 => 2.10.2 
    @remix-run/node: ^2.10.2 => 2.10.2 
    @remix-run/react: ^2.10.2 => 2.10.2 
    @remix-run/serve: ^2.10.2 => 2.10.2 
    vite: ^5.2.11 => 5.3.3
```

### Used Package Manager

pnpm

### Expected Behavior

The code to work correctly

### Actual Behavior

I get the following typescript errors:

```
The inferred type of 'loader' cannot be named without a reference to '.pnpm/@remix-run+server-runtime@2.10.2_typescript@5.5.3/node_modules/@remix-run/server-runtime/dist/single-fetch'. This is likely not portable. A type annotation is necessary.ts(2742)

The inferred type of 'loader' cannot be named without a reference to '.pnpm/react-router@6.24.1_react@19.0.0-rc-f994737d14-20240522/node_modules/react-router'. This is likely not portable. A type annotation is necessary.ts(2742)
```

This may be due to pnpm's strict dependencies isolation.

jansedlon:
Yes this happens to me as well

vixalien:
You're right. That's a mistake from my side

pcattori:
[...]

Issue #9649, opened 2024-06-24
'Possibly Infinite' type when attempting to use new Single Fetch types


lsthornt:
### Reproduction

I have been running across various flavors of this issue, but I think I've distilled it to the simplest reproduction here: [Typescript Playground reproduction](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgVwHYGcYEMBGAbAUwH0ATAgM2FQIEEBjGYCVOAXznKghDgHIABKARDAAHgFooaAPSoIZXgG4AUMpgBPMATj1GzAGJoGTFgF44AJQIxkUVABVNBADwatEcigzZ8xMpWpdEwA+FTdtIOYaKABzdDhzAAUsKCwQawIodGdI1ENUY2ZQtScdQtQrGzsEy2tbBycc8vzy4vCdAHcsYBgCElzK+pqaLp6+pr0KurtQoA)

Essentially, when attempting to use the new single fetch DataFunction types, you can trigger `Type instantiation is excessively deep and possibly infinite. ts(2589)`. I've found it helpful/necessary to use these types to develop boilerplate-reducing wrappers around `unstable_defineAction`/`unstable_defineLoader`.

Falling back to the currently-exposed `ActionFunction` type resolves the issue, but risks developing against the incorrect types.

I recognize that this is a issue on functions explicitly labeled as `unstable`, but I figured it may be helpful to expose as these future APIs get hardened.

### System Info

```shell
Typescript Playground
```

### Used Package Manager

npm

### Expected Behavior

One can develop against the new ActionFunction types added in 2.9.2 without typescript errors

### Actual Behavior

Attempting to use the internal `DataFunctionReturnValue` triggers a Typescript error.

lsthornt:
Related, if useful: I've found that trying to create generic functions at all utilizing the `unsafe_define...` types will often crush the Typescript server.

A common case for me is wanting to use the new equivalent of `ActionFunction`, which is only accessible as `ReturnType<typeof unsafe_defineAction>`. Using that type in conjunction with something like `useFetcher<T>` will generally freeze the TS server.

ksmithut:
I'll add my own example to this. Slight variation because I'm using the jsdoc flavor of TypeScript, but the error is the same.

<details>
<summary>vite.config.js</summary>

```js
import { vitePlugin as remix } from '@remix-run/dev'
import { defineConfig } from 'vite'
import { installGlobals } from '@remix-run/node'

[...]

Issue #9514, opened 2024-05-30
Single Fetch: TypeScript implementation of defineLoader doesn't support records with arrays


KubaJastrz:
### Reproduction

I copied the implementation of `Serializable` into TypeScript playground. Let me know if you want me to submit an actual reproduction with loader code.

https://www.typescriptlang.org/play/?#code/C4TwDgpgBAyhBOBLAhgG0QL2QI1dAvFAK4B2AJhAGaIkRlQA+UJRqqjU2A9l3siRwDOwJCQDmQkAFtu7JixkIOAQXjxkIADxwkaTDjwA+DgG8AUFCgBtANYQQALigAFeF0jxQAaXsBdJzoo6Fi4EADcZgC+HNiIYjTAHAAiyMDQTACqAEoAMhxZEGIAogAeYBxFalzwHACyyGDaCEH6oQA0sM16IUYccMBNusEGEMZMrlxSiIIQgy09oxFmZgkIlMgAxtApwMhQ5pbowk4508AAkmlSVr4RkSskafDrW1CnwpcQUvsWUIhkTmEojEd2WFA2qGQ8GgGy4JGEUDIqWQTh2yGWAHoMVBlIJBHESDQJGQuBBBCQAOSJADu1RsUAcGLMsPhiWwyABnSGrTwUEISN2SyxUHqYDARKgtPgNkEUAA-HLmXCEWIeJzAt0Rnyfoczk4BcgAHRHYCGqQNAAUFsQVwAlHzjBaDpY-pybV9Df82r9Irbbd7IkA

```ts
type Serializable = undefined | null | boolean | string | symbol | number | Array<Serializable> | {
  [key: PropertyKey]: Serializable;
} | bigint | Date | URL | RegExp | Error | Map<Serializable, Serializable> | Set<Serializable> | Promise<Serializable>;


interface Data {
  list: ListItem[];
}
interface ListItem {
  id: string;
}

declare const data: Data

// Assigning doesn't work :/
const bad: Serializable = data;

// Mapping works ??
const good: Serializable = {
  list: data.list.map((item) => ({
    id: item.id,
  })),
}
```

### System Info

```shell
npmPackages:
    @remix-run/dev: ^2.9.2 => 2.9.2 
    @remix-run/node: ^2.9.2 => 2.9.2 
    @remix-run/react: ^2.9.2 => 2.9.2 
    @remix-run/serve: ^2.9.2 => 2.9.2 
[...]

PR #9893, opened 2024-08-20
Single-fetch typesafety

### LINKED ISSUE #9428: Single Fetch: useFetcher without generics arg triggers TS error `instantiation is excessively deep...`
### Reproduction

https://stackblitz.com/edit/remix-run-remix-k6ghpy?file=app%2Froutes%2F_index.tsx

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @remix-run/dev: ^2.9.2 => 2.9.2 
    @remix-run/node: ^2.9.2 => 2.9.2 
    @remix-run/react: ^2.9.2 => 2.9.2 
    @remix-run/serve: ^2.9.2 => 2.9.2 
    vite: ^5.1.0 => 5.2.11
```

### Used Package Manager

npm

### Expected Behavior

Hi! For better typesafety we should give a loader/action function type parameter to the generic of `useFetcher`. But we sometimes don't do it, because of laziness or the action is too simple to bother, or nothing is done with the data. In these cases `useFetcher` (without generics arg) was just behaving normally, silently casting the `"data": any`.

### Actual Behavior

With SingleFetch the definition of `useFetcher` changed and it seems that we don't have default type parameter anymore:

```ts
  export function useFetcher<T extends Loader | Action>(
    opts?: Parameters<typeof useFetcherRR>[0]
  ): FetcherWithComponents<Serialize<T>>;
```
Would it be possible to have the `=  any` default parameter back?

pcattori:
Fixes #9428 #9514 #9649 #9714 #9776

## TODO

- [x] fix tests
- [x] `unstable_` prefix for `Future.singleFetch`
- [x] remove `Future` re-exports
- [x] Docs (`vite.config.ts` -> `declare module "@remix-run/server-runtime"`)
- [x] changeset


packages/remix-server-runtime/future.ts
```diff
@@ -0,0 +1 @@
+export interface Future {}
```
brophdawg11, line 1: Where does this get values added to it to trigger the ternary below?

```ts
// prettier-ignore
type SingleFetchEnabled =
  Future extends { singleFetch: infer T extends boolean } ? T : false
```

punkpeye:
I think that the code in this PR has the same issue as described here https://github.com/remix-run/remix/issues/9776

pcattori:
> I think that the code in this PR has the same issue as described here #9776

Turns out the new code with `Serialize<T>` helper doesn't have this issue, but just added some test cases and explicit handling for `ReadonlyArray` to be safe.

pcattori:
@romansndlr don't get too excited, this is just a quick patch for improve types for single-fetch in Remix. The full typesafety story is coming soon in RRv7 😉 but yea this is the first step towards that

brophdawg11 (approved):
👌 great work!

romansndlr:
> @romansndlr don't get too excited, this is just a quick patch for improved types for single-fetch in Remix, though yea that is pretty exciting on its own! The full typesafety story is coming soon in RRv7 😉 this is the first step towards that

For sure! I assumed as much... This is defiantly existing on it's own but also for what's to come!

docs/guides/single-fetch.md
```diff
@@ -145,145 +145,120 @@ Without Single Fetch, any plain Javascript object returned from a `loader` or `a
...
-  | Set<Serializable>
-  | Promise<Serializable>;
+export default function Component() {
+  const data = useLoaderData<typeof loader>();
+  //    ^? { planet: string, date: Date, notSoRandom: undefined }
+}
 ```
 
-There are also client-side equivalents un `defineClientLoader`/`defineClientAction` that don't have the same return value restrictions because data returned from `clientLoader`/`clientAction` does not need to be serialized over the wire:
+Methods are also not serializable, so class instances get slimmed down to just their serializable properties:
```
sergiodxa, line 199: @pcattori Couldn't Remix call `classInstance.toJSON()`? If it's not defined then grab the serializable properties automatically

pcattori: It _could_ but that's what `json` is for

PR #9895, opened 2024-08-21
Debugging of jest tests does not work

bbonch:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

In the root directory run to debug any test
```
node --inspect-brk ./node_modules/jest/bin/jest.js --runInBand packages/create-remix/__tests__/create-remix-test.ts
``` 
It fails with
```
Error: Cannot find module '@web3-storage/multipart-parser'
``` 
This module is used only in remix-server-runtime. 
I removed moduleNameMapper from jest.config.shared.js
I ran tests inside packages/remix-server-runtime successfully.
I seems @web3-storage/multipart-parser mapping is not required.
After that debugging works

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -701,3 +701,4 @@
 - zayenz
 - zhe
 - zwhitchcox
+- bbonch
--- a/jest/jest.config.shared.js
+++ b/jest/jest.config.shared.js
@@ -8,11 +8,6 @@ const ignorePatterns = [
 
 /** @type {import('jest').Config} */
 module.exports = {
-  moduleNameMapper: {
-    "^@web3-storage/multipart-parser$": require.resolve(
-      "@web3-storage/multipart-parser"
-    ),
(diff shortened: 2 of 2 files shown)


brophdawg11:
Huh, odd.  I've never had problems debugging but I have an alias I use that seems to work fine:

```sh
alias npx-debug="npx --node-options=\"--inspect-brk\""
```

brophdawg11:
> I ran tests inside packages/remix-server-runtime successfully.

It looks like this breaks the test in CI for other packages, so I don't think we added this specifically for `server-runtime`, but for other packages that rely on `server-runtime` and then need to be able to import this package during tests.  I confirmed that ` pnpm test:primary remix-react` fails locally with this change so I'm going to close this out.  

I would switch to debugging tests using the `npx --node-options` approach above which seems to work fine 👍

PR #9901, opened 2024-08-22
Fix `dest already exists` build errors

### LINKED ISSUE #9498: Build fails when using vite-imagetools plugin due to assets already existing in build/client

markdalgleish:
Fixes #9498.

This fix was inspired by #9800, but I decided to make this more error-proof by inspecting files on disk rather than inspecting the Vite manifest files/assets and potentially getting it wrong (as we have already).

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -829,11 +829,6 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
     return JSON.parse(manifestContents) as Vite.Manifest;
   };
 
-  let getViteManifestFilePaths = (viteManifest: Vite.Manifest): Set<string> => {
-    let filePaths = Object.values(viteManifest).map((chunk) => chunk.file);
-    return new Set(filePaths);
-  };
-
   let getViteManifestAssetPaths = (
     viteManifest: Vite.Manifest
   ): Set<string> => {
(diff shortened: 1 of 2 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -1419,8 +1411,9 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
           let movedAssetPaths: string[] = [];
           for (let ssrAssetPath of ssrAssetPaths) {
             let src = path.join(serverBuildDirectory, ssrAssetPath);
-            if (!clientFilePaths.has(ssrAssetPath)) {
-              let dest = path.join(clientBuildDirectory, ssrAssetPath);
+            let dest = path.join(clientBuildDirectory, ssrAssetPath);
+
+            if (!fse.existsSync(dest)) {
```
markdalgleish, line 1416: Since (assuming no errors) we were guaranteed to either move the SSR asset or delete it depending on whether it's present in the client assets, it's much safer to use the file system as the source of truth rather than trying to infer what's on disk based on the Vite manifest.

Issue #9844, opened 2024-08-08
dev server unexpectedly loads CSS when using `?inline` imports


mayank99:
### Reproduction

https://stackblitz.com/edit/vitejs-vite-sfsfcx?file=styles.css&terminal=dev

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    @remix-run/dev: ^2.11.1 => 2.11.1 
    @remix-run/node: ^2.11.0 => 2.11.1 
    @remix-run/react: ^2.11.0 => 2.11.1 
    @remix-run/serve: ^2.11.0 => 2.11.1 
    vite: ^5.4.0 => 5.4.0
```

### Used Package Manager

npm

### Expected Behavior

Vite supports importing CSS strings without automatically injecting CSS into the page using an `?inline` query param (see [docs](https://vitejs.dev/guide/features.html#disabling-css-injection-into-the-page)).

```js
import rawCssText from './styles.css?inline'
```

This works as expected in vanilla Vite (see [comparison stackblitz](https://stackblitz.com/edit/vitejs-vite-cy4kgj?terminal=dev)) so it should also work fine in Remix.

### Actual Behavior

The CSS gets added to the page only when using the dev server (prod build works fine) and only on initial load (it gets removed after HMR).

Specifically, it gets added as a `<style>` tag inside the `<head>`.

dj-rabel:
Experienced the same behavior in our project today and I think I know whats causing that.

https://github.com/remix-run/remix/blob/cc65962b1a96d1e134336aa9620ef1dad7c5efb1/packages/remix-dev/vite/styles.ts#L71-L75

There is a check for `?url`, but none for `?inline` and `?raw`. Thats why the initial serverside render includes the css.

mayank99:
Just confirmed this is fixed in `2.12.1-pre.0` 👍

PR #9910, opened 2024-08-23
Exclude `?inline` and `?raw` from critical styles in dev

dj-rabel:
- css files with query ?url, ?inline or ?raw shouldn't be injected on SSR bei vite spec
- a combination of these query-params is possible. This is checked as well

Closes: #9844

- [ ] Docs (not needed?)
- [x] Tests

Testing Strategy:

- I added a test for the new method, statically checking some possible css urls
- I did not know how to create a test for the whole thing, so I ran a build an copied the file "styles.js" into my regular project and ran the server. To show that, I adjusted the stackblitz from the original Issue (#9844)
-> Here it is: https://stackblitz.com/edit/vitejs-vite-qjxcev
-> And here executing dev without the override as a negative test https://stackblitz.com/edit/vitejs-vite-qjxcev?terminal=dev


markdalgleish:
Thanks for the PR! This is a really great catch.

I've pushed some refactors to make the code a little easier to follow by avoiding a regular expression, and expanded the tests a bit.

markdalgleish (approved):

PR #9914, opened 2024-08-26
Add failing test for syncronous blocker.proceed calls

brophdawg11:
Add a failing integration test for https://github.com/remix-run/react-router/issues/11613 since it relies on specific behavior of browser `history.go` calls that doesn't seem to be accurately implemented in JSDOM.

This should fail initially and be fixed once we point to an experimental RR release with the fix from https://github.com/remix-run/react-router/pull/11930


brophdawg11:
Huh - weird - this integration test fails for me locally but is passing on GH actions 🤔 

```
pnpm test:integration block --project chromium
```

Issue #9798, opened 2024-07-28
Vite dependency optimization is causing errors in development


tavoyne:
### Reproduction

See #8876 and #8917.

```ts
import {
  vitePlugin as remix,
  cloudflareDevProxyVitePlugin as remixCloudflareDevProxy,
} from "@remix-run/dev";
import { defineConfig } from "vite";
import tsconfigPaths from "vite-tsconfig-paths";

import { getLoadContext } from "./loadContext";

export default defineConfig({
  plugins: [
    remixCloudflareDevProxy({ getLoadContext }),
    remix(),
    tsconfigPaths(),
  ],
});
```

### System Info

```shell
System:
  OS: macOS 14.1
  CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Memory: 1.79 GB / 32.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
  npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
Browsers:
  Chrome: 126.0.6478.183
  Chrome Canary: 129.0.6622.0
  Safari: 17.1
npmPackages:
  @remix-run/cloudflare: ^2.10.3 => 2.10.3 
  @remix-run/cloudflare-pages: ^2.10.3 => 2.10.3 
  @remix-run/dev: ^2.10.3 => 2.10.3 
  @remix-run/eslint-config: ^2.10.3 => 2.10.3 
  @remix-run/react: ^2.10.3 => 2.10.3 
  vite: ^5.3.5 => 5.3.5
```

### Used Package Manager

npm

### Expected Behavior

Pages load smoothly.

### Actual Behavior

Errors like the following on page loads:

![image](https://github.com/user-attachments/assets/cbaf44e8-f7f8-4985-a8b1-bfefe62f11a1)

andrecasal:
Yep, all pages visited for the first time fail to load because Vite optimizes the dependencies, but sends a `504 Outdated Dependency` to the browser.

Only happens in during dev and doesn't affect production, but having all tests (or page loads) fail on first try is a bad experience.

PR #9921, opened 2024-08-29
Optimize dependencies found through route module graphs

### LINKED ISSUE #8803: Vite plugin: Outdated Optimize Dep errors
### Reproduction

I don't think a StackBlitz is appropriate in this case, here's a repo — https://github.com/silvenon/repro/tree/remix-vite, make sure to switch to the `remix-vite` branch.

It's built on top of your latest Vite template, with added login & logout logic via cookie session storage because I was initially trying to reproduce a logout bug that I was experiencing in Safari (but not in Chrome!) in my app after upgrading to the Vite plugin.

I wasn't able to reproduce that particular one, but I decided to leave in the complexity because I think it's useful for a bug I'm reporting.

### System Info

```shell
  System:
    OS: macOS 14.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 64.98 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.6.2 - ~/.n/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.2.4 - ~/.n/bin/npm
    pnpm: 8.6.10 - /opt/homebrew/bin/pnpm
    bun: 1.0.23 - ~/.bun/bin/bun
  Browsers:
    Chrome: 121.0.6167.184
    Edge: 121.0.2277.128
    Safari: 17.3.1
  npmPackages:
    @remix-run/dev: ^2.7.0 => 2.7.0 
    @remix-run/node: ^2.7.0 => 2.7.0 
    @remix-run/react: ^2.7.0 => 2.7.0 
    @remix-run/serve: ^2.7.0 => 2.7.0 
    vite: ^5.1.3 => 5.1.3
```

### Used Package Manager

npm

### Expected Behavior

Safari has been my default browser for several months now, so after upgrading my Remix app I expected my app's development experience to work similar to how developing regular Remix and Vite apps worked.

### Actual Behavior

A quick checks tells me that on Chrome this seems to work fine, but on Safari eventually I continuously start getting "Outdated Optimize Dep" 504 errors in my console with the accompanying "Importing a module script failed" error:

```
[Error] Failed to load resource: the server responded with a status of 504 (Outdated Optimize Dep) (react_jsx-dev-runtime.js, line 0)
[Error] Unhandled Promise Rejection: TypeError: Importing a module script failed.
[Error] Failed to load resource: the server responded with a status of 504 (Outdated Optimize Dep) (@remix-run_react.js, line   0)
[Error] Failed to load resource: the server responded with a status of 504 (Outdated Optimize Dep) (react.js, line 0)
[Error] Failed to load resource: the server responded with a status of 504 (Outdated Optimize Dep) (react-dom_client.js, line   0)
```

These errors break client-side JS, most notably HMR stops working.

One way to get them to show up is to start one Remix Vite app, stop it, and start another one. Both have to use the same port. This quirk does not happen with regular Vite apps, I checked, that's why I'm reporting it here.

The only way I found to make these errors go away immediately is to go to Safari Settings → Privacy → Manage Website Data and clear data from "localhost". Otherwise they stay even if you restart the dev server, if you delete the `.cache` directory or run `npm ci`.

That's as far as I got, hopefully that helps!

pcattori:
Fixes #8917 #9798
Might fix #8803 #8876

See changeset for description

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "vite": "5.1.3",
     "vite-tsconfig-paths": "^4.2.2",
     "wait-on": "^7.0.1",
-    "wrangler": "^3.24.0"
+    "wrangler": "^3.72.3"
   },
   "engines": {
     "node": ">=18.0.0"
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -1075,6 +1075,9 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
               : undefined,
           },
           optimizeDeps: {
+            entries: Object.values(ctx.remixConfig.routes).map((route) =>
+              path.join(ctx.remixConfig.appDirectory, route.file)
+            ),
             include: [
               // Pre-bundle React dependencies to avoid React duplicates,
               // even if React dependencies are not direct dependencies.
@@ -1497,7 +1500,12 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
       name: "remix-dot-server",
       enforce: "pre",
       async resolveId(id, importer, options) {
-        if (options?.ssr) return;
+        // https://vitejs.dev/config/dep-optimization-options
+        let isOptimizeDeps =
+          viteCommand === "serve" &&
+          (options as { scan?: boolean })?.scan === true;
+
+        if (options?.ssr || isOptimizeDeps) return;
 
         let isResolving = options?.custom?.["remix-dot-server"] ?? false;
         if (isResolving) return;
--- a/.changeset/hot-dogs-applaud.md
+++ b/.changeset/hot-dogs-applaud.md
@@ -1,0 +1,10 @@
+---
+"@remix-run/dev": patch
+---
+
+Automatically optimize dependencies during development
+
+Now Remix will tell Vite to find dependencies by crawling imports starting with your route modules.
+This should resolve most (if not all) `504 Outdated Dependency` errors that could previously break HMR.
+
+For users who were previously working around this limiation, you no longer need to explicitly add routes to Vite's `optimizeDeps.entries` nor do you need to disable the `remix-dot-server` plugin.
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -132,7 +132,7 @@
     "vite": "5.1.3",
     "vite-tsconfig-paths": "^4.2.2",
     "wait-on": "^7.0.1",
-    "wrangler": "^3.24.0"
+    "wrangler": "^3.72.3"
```
pcattori, line 135: Unrelated to this PR, our CI integration suite was failing due to outdated `wrangler` version, so batched that fix in here.

packages/remix-dev/vite/plugin.ts
```diff
@@ -1497,7 +1500,12 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
       name: "remix-dot-server",
       enforce: "pre",
       async resolveId(id, importer, options) {
-        if (options?.ssr) return;
+        // https://vitejs.dev/config/dep-optimization-options
+        let isOptimizeDeps =
+          viteCommand === "serve" &&
+          (options as { scan?: boolean })?.scan === true;
```
markdalgleish, line 1506: Hmmm, why is this not on the type? 🤔

packages/remix-dev/vite/plugin.ts
```diff
@@ -1075,6 +1075,9 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
               : undefined,
           },
           optimizeDeps: {
+            entries: Object.values(ctx.remixConfig.routes).map((route) =>
```
markdalgleish, line 1078: Looking at the docs for this option:

> If build.rollupOptions.input is specified, Vite will crawl those entry points instead

I wonder if the root cause is that we're only applying the Rollup build options when the `build` command is running:

https://github.com/remix-run/remix/blob/8f38118e44298d609224c6074ae6519d385196f1/packages/remix-dev/vite/plugin.ts#L1126

PR #9922, opened 2024-08-30
Automatically optimize dependencies of the client entry

pcattori:
Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/packages/remix-dev/vite/plugin.ts
+++ b/packages/remix-dev/vite/plugin.ts
@@ -1075,9 +1075,12 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
               : undefined,
           },
           optimizeDeps: {
-            entries: Object.values(ctx.remixConfig.routes).map((route) =>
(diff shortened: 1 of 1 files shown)


packages/remix-dev/vite/plugin.ts
```diff
@@ -1075,9 +1075,12 @@ export const remixVitePlugin: RemixVitePlugin = (remixUserConfig = {}) => {
               : undefined,
           },
           optimizeDeps: {
-            entries: Object.values(ctx.remixConfig.routes).map((route) =>
-              path.join(ctx.remixConfig.appDirectory, route.file)
-            ),
+            entries: [
+              ctx.entryClientFilePath,
```
markdalgleish, line 1079: Following on from this comment: https://github.com/remix-run/remix/pull/9921/files#r1740220074

The client entry is already captured in the Rollup build options, so if that turns out to fix it, it'll cover this case too.

PR #9923, opened 2024-08-30
chore: replace function to const

molezov:
replaced function to const in remix default template

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:


brophdawg11:
We're ok with `function` here

PR #9938, opened 2024-09-04
Update/fix Single Fetch revalidation behavior

brophdawg11:


integration/error-boundary-v2-test.ts
```diff
@@ -416,7 +416,7 @@ test.describe("single fetch", () => {
         await waitForAndAssert(
           page,
           app,
-          "#child-error",
+          "#parent-error",
```
brophdawg11, line 419: When the entire single fetch call fails we'll report this error for all the routes so it'll naturally "bubble" to the parent

PR #9954, opened 2024-09-06
Fix HDR for single fetch

brophdawg11:


integration/vite-hmr-hdr-test.ts
```diff
@@ -112,6 +113,61 @@ test("Vite / HMR & HDR / mdx", async ({ page, viteDev }) => {
   expect(page.errors).toEqual([]);
 });
 
+test.describe("single fetch", () => {
```
brophdawg11, line 116: Didn't have copies of these tests with the flag enabled so didn't catch this until we migrated the changes to RR v7 where single fetch is always on

PR #9958, opened 2024-09-06
Align single fetch prefetching with new revalidation logic

brophdawg11:


integration/single-fetch-test.ts
```diff
@@ -3253,9 +3253,9 @@ test.describe("single-fetch", () => {
       let app = new PlaywrightFixture(appFixture, page);
       await app.goto("/", true);
 
-      // A/B can be prefetched, C doesn't get prefetched due to its `clientLoader`
+      // root/A/B can be prefetched, C doesn't get prefetched due to its `clientLoader`
       await page.waitForSelector(
-        "nav link[rel='prefetch'][as='fetch'][href='/a/b/c.data?_routes=routes%2Fa%2Croutes%2Fa.b']",
+        "nav link[rel='prefetch'][as='fetch'][href='/a/b/c.data?_routes=root%2Croutes%2Fa%2Croutes%2Fa.b']",
```
brophdawg11, line 3258: This was wrong before because the root route has a loader and should be re-fetched by default

PR #9968, opened 2024-09-10
perf(remix-dev/vite): use `structuredClone` instead of `JSON.parse(JSON.stringify())`

xSyki:
`JSON.parse(JSON.stringify(` is not a recommended method for doing deepCopy. I suggest changing it to native structuredClone

https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

https://web.dev/articles/structured-clone

## Benchmarks

https://jsperf.app/watawu

## Support

https://caniuse.com/?search=structuredClone

Testing Strategy:

Called function directly

### DIFF
--- a/.changeset/tall-jars-destroy.md
+++ b/.changeset/tall-jars-destroy.md
@@ -1,5 +1,5 @@
 ---
-"@remix-run/dev": major
+"@remix-run/dev": patch
 ---
 
 Change JSON.parse(JSON.stringify to structuredClone native function
--- a/.changeset/tall-jars-destroy.md
+++ b/.changeset/tall-jars-destroy.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": major
+---
+
+Change JSON.parse(JSON.stringify to structuredClone native function
--- a/.changeset/tall-jars-destroy.md
+++ b/.changeset/tall-jars-destroy.md
@@ -2,4 +2,4 @@
 "@remix-run/dev": patch
 ---
 
-Change JSON.parse(JSON.stringify to structuredClone native function
+Use `structuredClone` instead of `JSON.parse(JSON.stringify())`
--- a/packages/remix-dev/vite/static/refresh-utils.cjs
+++ b/packages/remix-dev/vite/static/refresh-utils.cjs
@@ -13,7 +13,7 @@ function debounce(fn, delay) {
 const enqueueUpdate = debounce(async () => {
   let manifest;
   if (routeUpdates.size > 0) {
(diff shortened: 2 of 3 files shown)


.changeset/tall-jars-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/dev": major
```
MichaelDeBoey, line 2: Since Remix supports Node@>=18.0.0 and this is available as of Node v17, this is not a breaking change

```suggestion
"@remix-run/dev": patch
```

.changeset/tall-jars-destroy.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Change JSON.parse(JSON.stringify to structuredClone native function
```
MichaelDeBoey, line 5: Use `structuredClone` instead of `JSON.parse(JSON.stringify())`

MichaelDeBoey (changes requested):

xSyki:
@MichaelDeBoey Fixed

MichaelDeBoey (approved):

xSyki:
@MichaelDeBoey Thanks for approve

xSyki:
@MichaelDeBoey Can we merge that one?

packages/remix-dev/vite/static/refresh-utils.cjs
```diff
@@ -13,7 +13,7 @@ function debounce(fn, delay) {
 const enqueueUpdate = debounce(async () => {
   let manifest;
   if (routeUpdates.size > 0) {
-    manifest = JSON.parse(JSON.stringify(__remixManifest));
+    manifest = structuredClone(__remixManifest);
```
brophdawg11, line 16: I think without feature detection, this would technically be a breaking change for some end users because of the browser support.  Remix uses [`<script type=module>`](https://caniuse.com/?search=type%3D%22module%22) so we have a loose baseline of browser support back to Chrome 61.  But then due to PE, we can also support versions prior to 61 because browsers won't understand `<script type="module">` and ignore it.

`structuredClone` is not available until Chrome 98 so this PR would break existing Remix users on Chrome 61 through 97 (anyone on Chrome <61 wouldn't have HMR since they wouldn't have scripts loaded).  

```suggestion
    manifest = typeof structuredClone === 'function' ? 
      structuredClone(__remixManifest) : 
      JSON.parse(JSON.stringify(__remixManifest));
```

xSyki:
@brophdawg11 I changed it to what you suggested

MichaelDeBoey (approved):

Issue #9438, opened 2024-05-15
request.signal is never fired in dev server


DanielBaulig:
### Reproduction

https://stackblitz.com/edit/remix-run-remix-jkyczz?file=app%2Froutes%2Fstream.tsx

Based on discussion: https://github.com/remix-run/remix/discussions/8461

The effects of the bug are twofold: 
- Actual abort signal behavior cannot be tested and verified using the dev server.
- The dev server might leak resources that should have been freed when the request was cancelled.

Not that the bug only applies to the dev server. Using the express server the request abort signal is fired correctly.

### System Info

```shell
See Stackblitz
```

### Used Package Manager

npm

### Expected Behavior

Dev server should print 'Received Abort!' when the response gets cancelled a second in.

### Actual Behavior

Silence.

remorses:
Faced the same issue, this is very important please fix

PR #9975, opened 2024-09-11
Add support for fetcher.abort()

brophdawg11:
Remix sibling for https://github.com/remix-run/react-router/pull/11985

PR #9976, opened 2024-09-11
Proxy request.signal through in vite dev

brophdawg11:
Found this wasn't working right while testing https://github.com/remix-run/remix/pull/9975.  Aligned with how we do it in the express adapter.

Closes https://github.com/remix-run/remix/issues/9438

### DIFF
--- a/.changeset/hungry-experts-suffer.md
+++ b/.changeset/hungry-experts-suffer.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Abort request.signal in vite dev when the node response is closed
--- a/.changeset/hungry-experts-suffer.md
+++ b/.changeset/hungry-experts-suffer.md
@@ -2,4 +2,4 @@
 "@remix-run/dev": patch
 ---
 
-Abort request.signal in vite dev when the node response is closed
+Properly abort `request.signal` during `vite dev` when the node response is closed
--- a/packages/remix-dev/vite/cloudflare-proxy-plugin.ts
+++ b/packages/remix-dev/vite/cloudflare-proxy-plugin.ts
@@ -70,7 +70,7 @@ export const cloudflareDevProxyVitePlugin = <Env, Cf extends CfProperties>({
               )) as ServerBuild;
 
               let handler = createRequestHandler(build, "development");
-              let req = fromNodeRequest(nodeReq);
+              let req = fromNodeRequest(nodeReq, nodeRes);
               let loadContext = getLoadContext
                 ? await getLoadContext({ request: req, context })
                 : context;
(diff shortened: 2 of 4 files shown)


jacob-ebey (approved):

.changeset/hungry-experts-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/dev": patch
+---
+
+Abort request.signal in vite dev when the node response is closed
```
brophdawg11, line 5:
```suggestion
Properly abort `request.signal` during `vite dev` when the node response is closed
```

jvaill:
It's working better now! But I'm still having issues with the signal not being aborted sometimes.

Referencing this: https://github.com/sergiodxa/remix-utils/issues/317#issuecomment-2369224485

Eiley2:
Is there a workaround for this? Maybe check every once is a while if a connection is closed to abort that connection? I'm having the same issue with node and the updated version of remix in production too.


sse.time.ts
```typescript
import { LoaderFunctionArgs } from "@remix-run/node"
import { eventStream } from "remix-utils/sse/server"

export async function loader({ request }: LoaderFunctionArgs) {
  return eventStream(request.signal, function setup(send) {
    const timer = setInterval(() => {
      console.log(new Date().toLocaleTimeString())
      send({ event: "time", data: new Date().toLocaleTimeString() })
    }, 1000)

    return function clear() {
      console.log("SSE connection closed")
      clearInterval(timer)
    }
  })
}
``` 

and this is my route:


```typescript

export default Route(){
  let time = useEventSource("/sse/time", { event: "time" })
  if (!time) {
    time = new Date().toLocaleTimeString()
  }

return <p>{time}</p>
```

jvaill:
I tracked this down to some weirdness in undici.

It seems like the signal is working correctly until a new undici `Request()` is constructed. At that point, undici creates an underlying signal. That ends up getting passed to the remix handler.

[Here's where `Request()` is created in remix.](https://github.com/remix-run/remix/blob/main/packages/remix-server-runtime/data.ts#L86)

[Here's an issue I opened on the undici repo with a minimal test case.](https://github.com/nodejs/undici/issues/3644)

jvaill:
This should fix the remaining issues: https://github.com/remix-run/remix/pull/10030

cc @Eiley2

jvaill:
Possible fix: https://github.com/remix-run/remix/pull/10030

mkromann:
I am running into the same issue, and cannot get any EventSource to close with request.signal. Is there a workaround for how to handle this?

Right now it is a huge block for developing with SSE. We have to restart the whole app constantly.

PR #9992, opened 2024-09-16
docs(hooks/use-revalidator): update code example

jacobparis:
The original example looks like a GPT solution, the interval was never accounted for and the useEffect would just keep running immediately as soon as the revalidation completes

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

### DIFF
--- a/docs/hooks/use-revalidator.md
+++ b/docs/hooks/use-revalidator.md
@@ -38,16 +38,34 @@ The state of the revalidation. Either `"idle"` or `"loading"`.
 Initiates a revalidation.
 
 ```tsx
-function useLivePageData() {
-  const revalidator = useRevalidator();
-  const interval = useInterval(5000);
 
-  useEffect(() => {
-    if (revalidator.state === "idle") {
-      revalidator.revalidate();
+function useLivePageData() {
+  const revalidator = useRevalidator()
+
+  useInterval(() => {
+    if (revalidator.state === 'idle') {
+      revalidator.revalidate()
     }
-  }, [interval, revalidator]);
+  }, 5000)
 }
+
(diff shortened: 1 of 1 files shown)


docs/hooks/use-revalidator.md
```diff
@@ -38,16 +38,34 @@ The state of the revalidation. Either `"idle"` or `"loading"`.
...
+      id = setTimeout(tick, delay)
     }
-  }, [interval, revalidator]);
+
+    id = setTimeout(tick, delay)
+
+    return () => clearTimeout(id)
+  }, [callback, delay])
 }
+
```
brophdawg11, line 68: I think the idea behind `useInterval` was to hide the details around the `setTimeout`/`useEffect` so the docs would be focused on `useRevalidator` - but looks like we simplified it down too much such that it had a bug.  Do you think we need to include the implementation of `useInterval` in your corrected version?  Or do you think folks can grok the usage of `useRevalidator` without it?

alexanderson1993: Just link to Dan Abramov's blog post and call it a day https://overreacted.io/making-setinterval-declarative-with-react-hooks/

jacobparis: There's very little content on this page so I don't see a pressing need to minimize it, but if you want to drop the implementation then that's fine with me 

I just like being able to go to docs and copy/paste things that work, and having the useInterval there to also copy makes that easier

brophdawg11: yeah I like linking out to an implementation of it better than bloating the code block with unrelated lines personally

PR #9993, opened 2024-09-16
docs(guides/single-fetch): update references

haines:
There are still a couple of references to the old way of configuring type inference for single fetch in the docs; this PR updates them with the new way of doing things.


brophdawg11 (approved):
Thanks!

PR #9994, opened 2024-09-16
docs(guides/spa-mode): fix default `sirv-cli` configuration

ivan-kobzar:
Small change in [SPA Mode](https://remix.run/docs/en/main/guides/spa-mode) page.
To make `sirv-cli` works with default deployment setup we need to provide `--host` and `--port` flags.
Should be useful for beginners and save some time.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -706,3 +706,4 @@
 - zayenz
 - zhe
 - zwhitchcox
+- ivan-kobzar
--- a/docs/guides/spa-mode.md
+++ b/docs/guides/spa-mode.md
@@ -86,7 +86,7 @@ To deploy, you can serve your app from any HTTP server of your choosing. The ser
 For a simple example, you could use [sirv-cli][sirv-cli]:
 
 ```shellscript
-npx sirv-cli build/client/ --single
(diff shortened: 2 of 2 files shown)


brophdawg11:
What are you mean by "default deployment setup"?

ivan-kobzar:
> What are you mean by "default deployment setup"?

I mean fly.io for example

brophdawg11:
Where do we specify that Fly.io is the default deployment setup for Remix SPA Mode (even if we did, isn't [8080 the default port](https://fly.io/docs/reference/configuration/#the-http_service-section)?)  

If you're deploying to Fly, then you have the ability to run an SSR server and we'd probably recommend doing that to unlock the added Remix functionality.  SPA Mode is mostly intended for deployment to static file servers such as Github Pages, a CDN, etc.

Because SPA Mode's output is nothing more than an `index.html` file and some JS assets that can be deployed just about anywhere (a running server on Fly, a CDN, github pages, etc), I don't think we should be making any assumptions in the docs about flags to be passed.  They _may_ be accurate for you and your specific Fly deployment setup, but they may be incorrect and cause issues for others deploying elsewhere.  We link to the `sirv-cli` docs right above that example so that folks can easily discover these types of flags.

PR #10002, opened 2024-09-18
docs(remix-server-runtime): mention Netlify Edge runtime

serhalp:
👋🏼 This README lists supported server runtimes, but the Netlify Edge Functions runtime wasn't listed.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -593,6 +593,7 @@
 - selfish
 - sergiocarneiro
 - sergiodxa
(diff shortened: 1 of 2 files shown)


brophdawg11:
Thank you for the PR - but we list out all the third party/community adapters in the [docs](https://remix.run/docs/en/main/other-api/adapter#community-adapters) and we've got Netlify Edge included there so I think we're ok there.

PR #10017, opened 2024-09-23
Fix Cloudflare template imports

jsparkdev:
Fixed incorrect module imports for Cloudflare templates.


markdalgleish (approved):
Thanks for the PR!

PR #10019, opened 2024-09-23
Fix docs for environment variables in Cloudflare

MonsterDeveloper:
[Current docs](https://remix.run/docs/en/main/guides/envvars#local-development) say that env variables in Cloudflare environment are accessible through `context.env`, which was true for the classic compiler, but doesn't work with Vite.

Now as Vite is the default for Remix, env variables in Cloudflare are accessible through `context.cloudflare.env`. This PR reflects this change.

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -471,6 +471,7 @@
 - mochi-sann
 - mohammadhosseinbagheri
 - monitaure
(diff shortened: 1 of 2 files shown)


markdalgleish (approved):
Thanks for the PR!

PR #10030, opened 2024-09-26
Fix request signal not aborting after request is garbage collected

jvaill:
Closes: https://github.com/remix-run/remix/pull/9976#issuecomment-2373035840

Testing Strategy:

Add `remix-utils` and the following routes. Start the dev server and open `/time` in a bunch of different tabs. Refresh the tabs randomly, close them, create new ones, etc, and eventually you'll trigger the GC in the right way where `SSE connection closed ${number of open conns}` will be logged to the console with the number of open connections being greater than the number of open tabs. A good way to see if you've triggered this is just to close all of them and see if the logged connection count is greater than 0.

**routes/sse.time.ts**

```typescript
import { LoaderFunctionArgs } from "@remix-run/node"
import { eventStream } from "remix-utils/sse/server"

let conns = 0;

export async function loader({ request }: LoaderFunctionArgs) {
  conns++;
  return eventStream(request.signal, function setup(send) {
    const timer = setInterval(() => {
      send({ event: "time", data: new Date().toLocaleTimeString() })
    }, 1000)

    return function clear() {
      conns--;
      console.log("SSE connection closed", conns);
      clearInterval(timer)
    }
  })
```

**routes/time.tsx**

```typescript
import { useEventSource } from "remix-utils/sse/react"

export default Route(){
  let time = useEventSource("/sse/time", { event: "time" })
  if (!time) {
    time = new Date().toLocaleTimeString()
  }

return <p>{time}</p>
```


jvaill:
Fyi - the underlying problem appears to be a bug in undici:

https://github.com/nodejs/undici/issues/3644#issuecomment-2377644058

brophdawg11:
> It seems like globals should be installed in development as well? Manually invoking installGlobals() changes the Request object and solves this issue.

FWIW we are deprecating the `@remix-run/web-fetch` package in favor of using the built-in `fetch` on Node 20 and above - something you can opt-into via `installGlobals({ nativeFetch: true })` and is usually done when you opt into Single Fetch (see https://remix.run/docs/en/main/guides/single-fetch#breaking-changes).  

For now, it looks like there is some traction on the `node`/`undici` end so do we think they'll be getting a fix out anytime soon?  It would be nice not to have to put this type of workaround/hack into the Remix codebase to fix a bug in `undici`/`node`...

jvaill:
@brophdawg11 Another update on this - (IIRC, been a few weeks) at first I thought this was going to fix signals in production, but it seems like that's an entirely different bug altogether. When signals *are* working (i.e., in development), this fixes signals sporadically not being aborted due to GC.

I gave up on getting signals to work and ended up switching from SSE to websockets 😅

Eiley2:
> @brophdawg11 Another update on this - (IIRC, been a few weeks) at first I thought this was going to fix signals in production, but it seems like that's an entirely different bug altogether. When signals _are_ working (i.e., in development), this fixes signals sporadically not being aborted due to GC.
> 
> I gave up on getting signals to work and ended up switching from SSE to websockets 😅

Did you implement your own server with Express and Remix? Or what did you use? Considering to do the same tbh

mkromann:
Is it correct to say that SSEs are not working right now in Remix? I have been trying the last few days to find a workaround to this but been unable. We are starting up a new project and this is a curveball. It would be sad to go another route than Remix!

These issues are also reported here:
https://github.com/sergiodxa/remix-utils/issues/317
https://github.com/dan-cooke/remix-sse/issues/27

jvaill:
> > @brophdawg11 Another update on this - (IIRC, been a few weeks) at first I thought this was going to fix signals in production, but it seems like that's an entirely different bug altogether. When signals _are_ working (i.e., in development), this fixes signals sporadically not being aborted due to GC.
> > I gave up on getting signals to work and ended up switching from SSE to websockets 😅
> 
> Did you implement your own server with Express and Remix? Or what did you use? Considering to do the same tbh

I ended up adding a separate service to handle websockets. Basically, it uses `createServer` (`node:http`) and upgrades the connection to a websocket via `WebSocketServer` (`ws`). I use redis pub/sub to trigger messages from the remix app that get picked up by the service and proxied to the client. Might be a bit overly complex depending on your needs.

PR #10033, opened 2024-09-26
docs(route/meta): add favicon example

omerycll:
Added favicon example code.


MichaelDeBoey:
The docs explicitly say that `links` export should be used for favicons

> For asset links like stylesheets and favicons, you should use the `links` export instead.

PR #10034, opened 2024-09-26
templates(cloudflare-workers): use static assets

edmundhung:
This PR updates the cloudflare-workers template to use the Static Assets feature [announced today](https://blog.cloudflare.com/builder-day-2024-announcements/#static-asset-hosting).

> Note: Static assets is currently in open beta

You can learn more about Static Assets in the [documentation](https://developers.cloudflare.com/workers/static-assets/?cf_history_state=%7B%22guid%22%3A%22C255D9FF78CD46CDA4F76812EA68C350%22%2C%22historyId%22%3A14%2C%22targetId%22%3A%22167AF5232FF9C5279F7F203D0E95B587%22%7D).

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

To test, please run `create-remix` with the template on my fork 

```sh
npx create-remix@latest --template edmundhung/remix/templates/cloudflare-workers
```

Then deploy it with wrangler by running:

```sh
npm run deploy
```

Woohoo! Your Remix app is now running on Cloudflare Workers: https://template.remix-run.workers.dev

### DIFF
--- a/templates/cloudflare-workers/package.json
+++ b/templates/cloudflare-workers/package.json
@@ -12,7 +12,6 @@
     "typecheck": "tsc"
   },
   "dependencies": {
-    "@cloudflare/kv-asset-handler": "^0.3.2",
     "@remix-run/cloudflare": "*",
     "@remix-run/react": "*",
     "@remix-run/server-runtime": "*",
@@ -21,7 +20,7 @@
     "react-dom": "^18.2.0"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^4.20240512.0",
+    "@cloudflare/workers-types": "^4.20240925.0",
     "@remix-run/dev": "*",
     "@types/react": "^18.2.20",
     "@types/react-dom": "^18.2.7",
@@ -39,7 +38,7 @@
     "typescript": "^5.1.6",
     "vite": "^5.1.0",
     "vite-tsconfig-paths": "^4.2.1",
-    "wrangler": "3.57.1"
+    "wrangler": "^3.78.10"
   },
   "engines": {
     "node": ">=20.0.0"
--- a/templates/cloudflare-workers/package.json
+++ b/templates/cloudflare-workers/package.json
@@ -38,7 +38,7 @@
     "typescript": "^5.1.6",
     "vite": "^5.1.0",
     "vite-tsconfig-paths": "^4.2.1",
-    "wrangler": "^3.78.10"
+    "wrangler": "3.78.10"
   },
   "engines": {
     "node": ">=20.0.0"
--- a/templates/cloudflare-workers/package.json
+++ b/templates/cloudflare-workers/package.json
@@ -38,7 +38,7 @@
     "typescript": "^5.1.6",
     "vite": "^5.1.0",
     "vite-tsconfig-paths": "^4.2.1",
-    "wrangler": "3.78.10"
+    "wrangler": "3.84.0"
   },
   "engines": {
     "node": ">=20.0.0"
--- a/templates/cloudflare-workers/package.json
+++ b/templates/cloudflare-workers/package.json
@@ -20,7 +20,7 @@
     "react-dom": "^18.2.0"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^4.20240925.0",
+    "@cloudflare/workers-types": "^4.20241022.0",
     "@remix-run/dev": "*",
     "@types/react": "^18.2.20",
     "@types/react-dom": "^18.2.7",
--- a/templates/cloudflare-workers/server.ts
+++ b/templates/cloudflare-workers/server.ts
@@ -1,56 +1,37 @@
-import { getAssetFromKV } from "@cloudflare/kv-asset-handler";
 import { createRequestHandler, type ServerBuild } from "@remix-run/cloudflare";
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore This file won’t exist if it hasn’t yet been built
 import * as build from "./build/server"; // eslint-disable-line import/no-unresolved
-// eslint-disable-next-line import/no-unresolved
-import __STATIC_CONTENT_MANIFEST from "__STATIC_CONTENT_MANIFEST";
+import { getLoadContext } from "./load-context";
 
-const MANIFEST = JSON.parse(__STATIC_CONTENT_MANIFEST);
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const handleRemixRequest = createRequestHandler(build as any as ServerBuild);
 
 export default {
   async fetch(request, env, ctx) {
-    const waitUntil = ctx.waitUntil.bind(ctx);
-    const passThroughOnException = ctx.passThroughOnException.bind(ctx);
     try {
-      const url = new URL(request.url);
-      const ttl = url.pathname.startsWith("/assets/")
-        ? 60 * 60 * 24 * 365 // 1 year
-        : 60 * 5; // 5 minutes
-      return await getAssetFromKV(
-        { request, waitUntil },
-        {
-          ASSET_NAMESPACE: env.__STATIC_CONTENT,
-          ASSET_MANIFEST: MANIFEST,
-          cacheControl: {
-            browserTTL: ttl,
-            edgeTTL: ttl,
+      const loadContext = getLoadContext({
+        request,
+        context: {
+          cloudflare: {
+            // This object matches the return value from Wrangler's
+            // `getPlatformProxy` used during development via Remix's
+            // `cloudflareDevProxyVitePlugin`:
+            // https://developers.cloudflare.com/workers/wrangler/api/#getplatformproxy
+            cf: request.cf,
+            ctx: {
+              waitUntil: ctx.waitUntil.bind(ctx),
+              passThroughOnException: ctx.passThroughOnException.bind(ctx),
+            },
+            caches,
+            env,
           },
         }
-      );
-    } catch (error) {
-      // No-op
-    }
-
-    try {
-      const loadContext = {
-        cloudflare: {
-          // This object matches the return value from Wrangler's
-          // `getPlatformProxy` used during development via Remix's
-          // `cloudflareDevProxyVitePlugin`:
-          // https://developers.cloudflare.com/workers/wrangler/api/#getplatformproxy
-          cf: request.cf,
-          ctx: { waitUntil, passThroughOnException },
-          caches,
-          env,
-        },
-      };
+      });
       return await handleRemixRequest(request, loadContext);
     } catch (error) {
       console.log(error);
       return new Response("An unexpected error occurred", { status: 500 });
     }
   },
-} satisfies ExportedHandler<Env & { __STATIC_CONTENT: KVNamespace<string> }>;
+} satisfies ExportedHandler<Env>;
--- a/templates/cloudflare-workers/vite.config.ts
+++ b/templates/cloudflare-workers/vite.config.ts
@@ -4,6 +4,7 @@ import {
   cloudflareDevProxyVitePlugin,
 } from "@remix-run/dev";
 import tsconfigPaths from "vite-tsconfig-paths";
+import { getLoadContext } from "./load-context";
 
 declare module "@remix-run/cloudflare" {
   interface Future {
@@ -13,7 +14,9 @@ declare module "@remix-run/cloudflare" {
 
 export default defineConfig({
   plugins: [
-    cloudflareDevProxyVitePlugin(),
+    cloudflareDevProxyVitePlugin({
+      getLoadContext,
+    }),
     remix({
       future: {
         v3_fetcherPersist: true,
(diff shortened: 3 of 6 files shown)


templates/cloudflare-workers/server.ts
```diff
@@ -1,39 +1,13 @@
...
-            edgeTTL: ttl,
-          },
-        }
-      );
-    } catch (error) {
-      // No-op
-    }
-
     try {
       const loadContext = {
```
saaskitten, line 12: Shouldn't this be getting value from `getLoadContext()` to make it consistent between local Vite dev and production cloudflare workers?

templates/cloudflare-workers/wrangler.toml
```diff
@@ -1,12 +1,14 @@
...
 workers_dev = true
 # https://developers.cloudflare.com/workers/platform/compatibility-dates
-compatibility_date = "2023-04-20"
+compatibility_date = "2024-09-26"
 
-[site]
-bucket = "./build/client"
+[assets]
+# https://developers.cloudflare.com/workers/static-assets/binding/
+directory = "./build/client"
```
saaskitten, line 11: Can we also add observability?

```toml
[observability]
enabled = true
```

templates/cloudflare-workers/server.ts
edmundhung: Ah, yes! Totally missed it. Thanks

templates/cloudflare-workers/wrangler.toml
```diff
@@ -1,12 +1,14 @@
+#:schema node_modules/wrangler/config-schema.json
```
moishinetzer, line 1: Where's this coming from?

edmundhung: If you use an IDE extension that support json schema over TOML, this will give you auto completion and validation on your wrangler.toml, like [this one](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml#completion-and-validation-with-json-schema).

moishinetzer: fabulous

templates/cloudflare-workers/package.json
```diff
@@ -39,7 +38,7 @@
     "typescript": "^5.1.6",
     "vite": "^5.1.0",
     "vite-tsconfig-paths": "^4.2.1",
-    "wrangler": "3.57.1"
+    "wrangler": "^3.78.10"
```
brookslybrand, line 41: In the past we've pinned this version because we've had unexpected breaking changes with `wrangler`

```suggestion
    "wrangler": "3.78.10"
```

edmundhung: Sorry to hear about the past issues with breaking changes! I'll make sure the team works to avoid this going forward. Happy to pin the version for now to get this moving. 👍🏼

templates/cloudflare-workers/wrangler.toml
brookslybrand: We try to keep our templates very minimal and let users add the platform features they'd like.

Plus this feature is in beta, so I'm not super comfortable automatically turning it on for everyone using this template

templates/cloudflare-workers/package.json
brookslybrand: Thanks Edmund! I appreciate that, and really appreciate you updating this template.

templates/cloudflare-workers/wrangler.toml
saaskitten: Is the concern more on runtime crashing or something? If yes, this doesn't affect runtime. 

This config is very useful for Remix users as it provides immediate error logging without having to setup complicated o11y tools like Sentry.

saaskitten: Is the concern more on runtime crashing or something? If yes, this doesn't affect runtime. 

This config is very useful for Remix users as it provides immediate error logging without having to setup complicated o11y tools like Sentry.

I bumped into errors when trying to update the template myself locally and this flag saved the day during the debugging.

edmundhung: I agree that the observability logs are really helpful for debugging. However, it's not entirely free—there’s a generous free tier, but costs might apply once you exceed that. To avoid unexpected charges, it’s better to let developers opt into this themselves.

saaskitten: Oh ok.

templates/cloudflare-workers/wrangler.toml
laishere: It's generated by the wrangler.

> Where's this coming from?

templates/cloudflare-workers/wrangler.toml
OnurGvnc: https://x.com/CherryJimbo/status/1839753265535430757

tsteckenborn:
@edmundhung while you're on it, shouldn't https://github.com/edmundhung/remix/blob/1b52fe4b118735692b94d5338fca39c1b8bc55ae/templates/cloudflare-workers/app/root.tsx#L8 be from `@remix-run/cloudflare`?

ayuhito:
I've been testing this template and it's been working flawlessly. However, has anyone else managed to make the asset binding work? I keep getting `TypeError: Cannot read properties of undefined (reading 'fetch')` when using the `Fetcher` interface of the asset binding.

This only occurs when running the `dev` script with `remix:vite dev`. `wrangler dev` seems to work, although that requires the whole build step.

tsteckenborn:
@ayuhito are you when running `pnpm dev`also seeing the state error or is that something on my machine? 😕

So it's showing the page, but I'm seeing the error in the terminal.

```
> remix vite:dev

Re-optimizing dependencies because lockfile has changed
  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
TypeError [ERR_INVALID_STATE]: Invalid state: Controller is already closed
    at ReadableByteStreamController.close (node:internal/webstreams/readablestream:1157:13)
```

laishere:
> @edmundhung while you're on it, shouldn't https://github.com/edmundhung/remix/blob/1b52fe4b118735692b94d5338fca39c1b8bc55ae/templates/cloudflare-workers/app/root.tsx#L8 be from `@remix-run/cloudflare`?

And this too, they are from the old commits:
https://github.com/edmundhung/remix/blob/1b52fe4b118735692b94d5338fca39c1b8bc55ae/templates/cloudflare-workers/app/routes/_index.tsx#L1

ayuhito:
> also seeing the state error or is that something on my machine?

Nope. I've got a different error to that. I was alluding to the fact that `cloudflareDevProxyVitePlugin` might not perfectly handle the static assets interface in my comment.

FraBle:
Is there any rough timeline for when this PR might land? (excited about using it)

petebacondarwin (approved):
This looks good to me. If we can land this we can fix up the create-cloudflare tool to point to it.

focux:
Does this same setup work for SPA mode?

templates/cloudflare-workers/wrangler.toml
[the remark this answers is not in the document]
acusti: @petebacondarwin for the majority of use cases, there’s no need for the binding. the static assets feature intercepts requests and checks if the URL matches anything in the assets `directory` ([source](https://developers.cloudflare.com/workers/static-assets/routing/)). if it finds a file, it serves that. if not it, invokes your worker. as a result, the ASSETS binding’s only purpose is to provide programmatic access from your worker to the available assets that are already being automatically served when appropriate. that’s why the `create-cloudflare` remix template doesn’t default to including the binding: https://github.com/cloudflare/workers-sdk/blob/main/packages/create-cloudflare/templates-experimental/remix/templates/wrangler.toml

acusti:
Static Assets (aka Workers Assets) is currently in beta, but according to the [Wrangler configuration docs](https://developers.cloudflare.com/workers/wrangler/configuration/#top-level-only-keys), Workers Sites has been deprecated in favor of Workers Assets:

> `site` `object` optional deprecated
> * See the [Workers Sites](https://developers.cloudflare.com/workers/wrangler/configuration/#workers-sites) section below for more information. Cloudflare Pages and Workers Assets is preferred over this approach.

also, i’ve been using it in production (same as the implementation introduced here) without issue for a few weeks now.

my only comment on this PR is that i’m not sure whether or not it’s worth including the custom `getLoadContext` pattern for [augmenting load context](https://remix.run/docs/en/main/guides/vite#augmenting-load-context), because it seems like more of an advanced pattern. but it also seems pretty harmless, and maybe it will help template consumers discover the pattern if they find the need for it, so i don’t mind either way.

--------------

@focux i haven’t tried it, but i can’t imagine why it wouldn’t work for SPA mode. the static assets feature matches any request to any file in the specified `directory` (`./build/client/` in Remix), so it serves any built JS and HTML (and any other static files placed in the `./public/` directory) automatically.

--------------

@FraBle you can use this right away for new projects by using the fork:
```
npx create-remix@latest --template edmundhung/remix/templates/cloudflare-workers
```

or just manually apply the diff of this PR to an existing Remix / Cloudflare Workers project if you have one.

MichaelDeBoey (approved):

templates/cloudflare-workers/vite.config.ts
```diff
@@ -4,10 +4,13 @@ import {
...
+import { getLoadContext } from "./load-context";
 
 export default defineConfig({
   plugins: [
-    cloudflareDevProxyVitePlugin(),
+    cloudflareDevProxyVitePlugin({
+      getLoadContext,
+    }),
     remix({
       future: {
```
brookslybrand, line 15: @edmundhung let's go ahead and add 

```
v3_lazyRouteDiscovery: true,
v3_singleFetch: true,
```

to the config

templates/cloudflare-workers/server.ts
```diff
@@ -1,56 +1,37 @@
...
-            edgeTTL: ttl,
+      const loadContext = getLoadContext({
+        request,
+        context: {
+          cloudflare: {
+            // This object matches the return value from Wrangler's
+            // `getPlatformProxy` used during development via Remix's
+            // `cloudflareDevProxyVitePlugin`:
+            // https://developers.cloudflare.com/workers/wrangler/api/#getplatformproxy
+            cf: request.cf,
```
brookslybrand, line 21: This causes a type error, should it just be cast to `any`?

laishere: I think we can just set it to
```ts
cf: request.cf!
```
Or, we can change `load-context.ts` to:
```ts
type GetLoadContextArgs = {
  request: Request;
  context: {
    cloudflare: Omit<PlatformProxy<Env, IncomingRequestCfProperties>, "dispose" | "caches" | "cf"> & {
      caches: PlatformProxy<Env, IncomingRequestCfProperties>['caches'] | CacheStorage;
      cf: Request['cf'];
    };
  };
}
```
The first is simpler, but the second is better.

brookslybrand: I'd prefer correctness, I'll defer to @edmundhung

acusti: when I ran into this issue in my own codebase, I looked into it and wound up going with the non-null assertion operator (`cf: request.cf!`) based on the Cloudflare docs, which explain that “the `request.cf` object is not available in the Cloudflare Workers dashboard or Playground preview editor”: https://developers.cloudflare.com/workers/runtime-apis/request/#incomingrequestcfproperties

in every other case, the `request.cf` object is defined and available, so I figured it seemed safe to assert that in this context.

AdiRishi:
Really keen for this to land!

I do have a question, I haven't been able to find a good answer to this yet. How can we set `cache-control` headers for static asssets server via the new method? Previously the `ttl` was being configured in code, and in Cloudflare Pages it's configured via the `_headers` file. What's the strategy here? I can see under the new template all assets use `public, max-age=0, must-revalidate` as the cache control header.

edmundhung:
> Really keen for this to land!
> 
> I do have a question, I haven't been able to find a good answer to this yet. How can we set `cache-control` headers for static asssets server via the new method? Previously the `ttl` was being configured in code, and in Cloudflare Pages it's configured via the `_headers` file. What's the strategy here? I can see under the new template all assets use `public, max-age=0, must-revalidate` as the cache control header.

They are not supported yet. But you can find some workarounds under the `Static Assets` section: https://developers.cloudflare.com/workers/static-assets/compatibility-matrix/

templates/cloudflare-workers/wrangler.toml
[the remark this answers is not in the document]
edmundhung: Let's keep it minimal as Brooks suggest [here](https://github.com/remix-run/remix/pull/10034#discussion_r1778684491).

templates/cloudflare-workers/vite.config.ts
edmundhung: I see that you added the config already. I have rebased my branch to reflect those.

templates/cloudflare-workers/server.ts
edmundhung: The `request.cf` object is not always available as @acusti shared so I adopted the 2nd solution @laishere suggested.

brookslybrand (approved):
Looks great to me, thanks so much @edmundhung!

tsteckenborn:
Is there already a port of this to RR7? @brookslybrand it seems like the type in `@react-router/cloudflare` over there for the cloudflare `createRequestHandler` is `createPagesFunctionHandlerParams`, which might be misleading as it could also be a worker?

PR #10041, opened 2024-09-30
templates(classic-compiler/cloudflare-workers): use static assets

zwily:
This PR updates the classic-remix-compiler/cloudflare-workers template to use the new [Static Assets feature](https://blog.cloudflare.com/builder-day-2024-announcements/#static-asset-hosting). It also modernizes the template to match the current vite ones as much as possible. (Includes tailwind by default, uses `load-context.ts` for augmenting context, etc)

A separate PR (#10034) is open to update the vite template, but this is for the classic compiler. There are a couple reasons to not use vite with cloudflare yet:

 * Your code doesn't run in wrangler, so it doesn't match the execution environment very well.
 * Some bindings don't work through the platform proxy (like Durable Objects).

Once Vite 6 comes out with Vite Environments and your code can run in wrangler, this won't be necessary. But for now it is for at least some of us.

Testing Strategy:

To test, please run create-remix with the template on my fork

```
npx create-remix@latest --template https://github.com/zwily/remix/tree/workers-assets-classic-compiler/templates/classic-remix-compiler/cloudflare-workers
```

Then deploy it with wrangler by running:

```
npm run deploy
```


zwily:
I have submitted a PR to wrangler to fix the issue:

https://github.com/cloudflare/workers-sdk/pull/6866

zwily:
OK, wrangler is fixed and this template is pulling the right version.

Wrangler is also logging every file change on every build which is obnoxious, but not breaking. I have another PR for that here: https://github.com/cloudflare/workers-sdk/pull/6873 but that shouldn't block this.

zwily:
Okay, this template is in working shape now with current wrangler. 

(The only thing is we have `--x-dev-env` disabled for now because of this issue in wrangler which is closed but not fixed: https://github.com/cloudflare/workers-sdk/issues/6876)

MichaelDeBoey:
@zwily It seems the mentioned bug is now fixed?

PR #10047, opened 2024-10-01
fix abort behavior for renderToReadableStream

brophdawg11:


packages/remix-dev/config/defaults/entry.server.cloudflare.tsx
```diff
@@ -13,7 +13,9 @@ export default async function handleRequest(
   const body = await renderToReadableStream(
     <RemixServer context={remixContext} url={request.url} />,
     {
-      signal: request.signal,
+      // If you wish to abort the rendering process, you can pass a signal here.
+      // Please refer to the templates for example son how to configure this.
+      // signal: controller.signal,
```
brophdawg11, line 18: Stop passing the `request.signal` to fix the issue - but do not add any default timeout behavior to avoid a breaking change

templates/classic-remix-compiler/cloudflare-pages/app/entry.server.tsx
```diff
@@ -9,6 +9,8 @@ import { RemixServer } from "@remix-run/react";
 import { isbot } from "isbot";
 import { renderToReadableStream } from "react-dom/server";
 
+const ABORT_DELAY = 5000;
```
brophdawg11, line 12: Templates can have the timeout logic included so that net new apps have it going forward

jacob-ebey (approved):

PR #10055, opened 2024-10-03
Override cookie encode

Evanion:
## Current behaviour
Right now, the Base64 encoding is applied before the value is sent to the cookie serializer. but the `options` object still takes parameters for overriding the `encode`/ `decode` functions.

## Expected behaviour 
I would expect that when I add my own `encode`/`decode` functions, that the value isn't Base64 encoded anymore.

## Side effects
One side effect of this change, is that you can't sign cookies using the `secrets` property any more. But the user can still handle that themselves in the `encode`/`decode` functions.

This will give more flexibility to the cookie utils in remix, and allow developers to keep a good DX, by not having to use different APIs for different cookies minimizing confusion.

## Testing Strategy
[These new tests covers this code](https://github.com/remix-run/remix/compare/main...Evanion:override-cookie-encode?expand=1#diff-09eb98fd8e0c75af0f2010a7fa9344eeca487b5f71d03787d765ea871b2741dd)

### DIFF
--- a/packages/remix-server-runtime/__tests__/cookies-test.ts
+++ b/packages/remix-server-runtime/__tests__/cookies-test.ts
@@ -193,6 +193,38 @@ describe("cookies", () => {
       );
     });
   });
+
+  describe("encode/decode options", () => {
+    it("Will not Base64 encode/decode the cookie if `encode` and `decode` are set to custom functions", async () => {
+      let testValue = { hello: "world" };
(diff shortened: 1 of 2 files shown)


Evanion:
I'm unsure what `changeset` I should add. Strictly speaking, there is a breaking change, so following semver, it's a major bump.

PR #10072, opened 2024-10-07
Stabilize future flags

brophdawg11:
Stacked on https://github.com/remix-run/remix/pull/9980

### DIFF
--- a/packages/remix-dev/config.ts
+++ b/packages/remix-dev/config.ts
@@ -38,7 +38,7 @@ interface FutureConfig {
   v3_relativeSplatPath: boolean;
   v3_throwAbortReason: boolean;
   v3_singleFetch: boolean;
-  unstable_lazyRouteDiscovery: boolean;
+  v3_lazyRouteDiscovery: boolean;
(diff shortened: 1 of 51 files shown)


packages/remix-dev/config.ts
```diff
@@ -529,12 +529,6 @@ export async function resolveConfig(
     entryServerFile = `entry.server.${serverRuntime}.tsx`;
   }
 
-  if (isSpaMode && appConfig.future?.unstable_lazyRouteDiscovery === true) {
-    throw new Error(
-      "You can not use `future.unstable_lazyRouteDiscovery` in SPA Mode (`ssr: false`)"
-    );
-  }
```
brophdawg11, line 536: I think this was unintentionally left in here from a first pass - we decided up front that SPA mode would just not have fog of war so its never enabled at run time, so no need to fail the build (which would be a problem in RR v7 anyway)

```js
export function isFogOfWarEnabled(future: FutureConfig, isSpaMode: boolean) {
  return future.v3_lazyRouteDiscovery === true && !isSpaMode;
}
```

PR #10107, opened 2024-10-14
Add `routes.ts` support behind future flag

markdalgleish:
See changeset and docs for more details.


integration/vite-route-config-test.ts
```diff
@@ -0,0 +1,348 @@
...
+      "vite.config.js": `
+        import { vitePlugin as remix } from "@remix-run/dev";
+
+        export default {
+          plugins: [remix({
+            routes: () => {},
+          })]
+        }
+      `,
+      "app/routes.ts": `export default INVALID(`,
```
brophdawg11, line 50: I don't think this is intended to be an invalid syntax file - maybe invalid copy/paste from line 86?  I think this test probably wants to assert that even with a valid `routes.ts` we error if you're using the `routes` config?

integration/vite-route-config-test.ts
```diff
@@ -0,0 +1,348 @@
...
+        import { vitePlugin as remix } from "@remix-run/dev";
+
+        export default {
+          ${await viteConfig.server({ port })}
+          plugins: [remix({
+            routes: () => {},
+          })]
+        }
+      `,
+      "app/routes.ts": `export default INVALID(`,
```
brophdawg11, line 73: Same question as above

packages/remix-dev/config/routes.ts
```diff
@@ -40,7 +55,335 @@ export interface ConfigRoute {
...
+        ...route,
+        children: prefix(prefixPath, route.children),
+      };
+    }
+    return route;
+  });
+}
+
+const helpers = { route, index, layout, prefix };
+export { route, index, layout, prefix };
```
brophdawg11, line 302: Do we need to do this instead of inline exports?

brophdawg11 (approved):

integration/vite-route-config-test.ts
markdalgleish: Ah, nice catch 👍

packages/remix-dev/config/routes.ts
markdalgleish: I formatted the code like this to make it easier to spot if the `relative` API gets out of sync with the top-level exports.

brophdawg11: ah ok - that makes sense

packages/remix-dev/index.ts
```diff
@@ -5,6 +5,19 @@ export type { AppConfig, RemixConfig as ResolvedRemixConfig } from "./config";
...
+  RouteManifestEntry as UNSAFE_RouteManifestEntry,
+  RouteConfig as UNSAFE_RouteConfig,
+  RouteConfigEntry as UNSAFE_RouteConfigEntry,
+} from "./config/routes";
+export {
+  defineRoutes as UNSAFE_defineRoutes,
+  routeManifestToRouteConfig as UNSAFE_routeManifestToRouteConfig,
+  getRouteConfigAppDirectory as UNSAFE_getRouteConfigAppDirectory,
+} from "./config/routes";
+export { flatRoutes as UNSAFE_flatRoutes } from "./config/flat-routes";
```
markdalgleish, line 20: This isn't needed in React Router because `@remix-run/dev/routes` has access to the internals of the dev package, whereas in this case we need to expose these internals across package boundaries.

docs/start/future-flags.md
```diff
@@ -476,6 +476,143 @@ You may find some usage for the new [`<Link discover>`][discover-prop] API if yo
 
 Opt into automatic [dependency optimization][dependency-optimization] during development. This flag will remain in an "unstable" state until React Router v7 so you do not need to adopt this in your Remix v2 app prior to upgrading to React Router v7.
 
+## routes.ts
+
+Config-based routing is the new default in React Router v7. Support for `routes.ts` and its related APIs in Remix are designed as a migration path to help minimize the number changes required when moving your Remix project over to React Router v7. Since React Router v7 is not yet stable, these APIs are also considered unstable.
```
brookslybrand, line 481:
```suggestion
Config-based routing is the new default in React Router v7. Support for `routes.ts` and its related APIs in Remix are designed as a migration path to help minimize the number of changes required when moving your Remix project over to React Router v7. Since React Router v7 is not yet stable, these APIs are also considered unstable.
```

brophdawg11 (approved):

PR #10115, opened 2024-10-15
fix(create-remix): add fallback git author to initial commit

klibardi:
When run in an environment where git global config has not been set via
```
git config --global user.email "you@example.com"
git config --global user.name "Your Name"
```
and cannot be inferred from local username and hostname, the initialization process fails (git exits with exit code 128).

Closes: #

- [ ] Docs
- [ ] Tests

Testing Strategy:

Manual, just tried running it with various git setups.

### DIFF
--- a/packages/create-remix/index.ts
+++ b/packages/create-remix/index.ts
@@ -507,10 +507,24 @@ async function gitInitStep(ctx: Context) {
     while: async () => {
       let options = { cwd: ctx.cwd, stdio: "ignore" } as const;
       let commitMsg = "Initial commit from create-remix";
+      const commitUser = "create-remix";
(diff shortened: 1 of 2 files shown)


packages/create-remix/index.ts
```diff
@@ -507,10 +507,24 @@ async function gitInitStep(ctx: Context) {
     while: async () => {
       let options = { cwd: ctx.cwd, stdio: "ignore" } as const;
       let commitMsg = "Initial commit from create-remix";
+      const commitUser = "create-remix";
+      const commitEmail = "<>"
```
MichaelDeBoey, line 511: I think this should be the fallback and we should use the available global defaults first

PR #10126, opened 2024-10-16
Add deprecations for future flags

brophdawg11:


.changeset/rude-clocks-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/dev": minor
+---
+
+Log deprecation warnings for v3 future flags
```
brophdawg11, line 5:
```suggestion
- Log deprecation warnings for v3 future flags
- Add `@deprecated` annotations to `json`/`defer` utilities
```

wouterds:
How does this work in combination with
>⚠️ REMIX FUTURE CHANGE: Externally-accessed resource routes will no longer be able to return raw JavaScript objects or `null` in React Router v7 when Single Fetch becomes the default. You can prepare for this change at your convenience by wrapping the data returned from your `action` function in the `routes/users` route with `json()`.  For instructions on making this change, see https://remix.run/docs/en/v2.13.1/guides/single-fetch#resource-routes

tmcw:
There's quite a lot of documentation that still includes the `json()` method - the [documentation for loader itself](https://remix.run/docs/en/main/route/loader) uses `json()`, as well as [the tutorials](https://remix.run/docs/en/main/start/tutorial#updating-data). Folks who start from those directions will get warnings from Remix, which is not a very good onboarding experience.

danielvanc:
> > There's quite a lot of documentation that still includes the `json()` method  ... Folks who start from those directions will get warnings from Remix, which is not a very good onboarding experience.
> 
> Not to mention there is nothing I can find pointing me to what I should do instead? It would be nice to also have the deprecation message point us to an alternative.

Yeah i agree. it should at least show examples of links to pages like these:
https://reactrouter.com/dev/start/framework/data-loading

bhouston:
I recently adopted Remix and I am noticing that my code has deprecation warnings.  It appears to be related to this issue.

The issue is that I am using json() in order to return data and a status code.  What is the replacement code for this?

peter-skillmasters:
The return type of `Response.json()` is just `Response`, which means I lose type information when I use it. So if you need to apply a custom status code or use `commitSession` with a JSON body, it's a pretty major DX downgrade from `json()`. What's the recommendation here? 

```ts
function oldLoader() {
  return json({ foo: "bar"}, {status: 201})
}

type OldLoaderReturn = ReturnType<typeof useLoaderData<typeof oldLoader>>
// => { foo: string }

function newLoader() {
  return Response.json({ foo: "bar"}, {status: 201})
}

type NewLoaderReturn = ReturnType<typeof useLoaderData<typeof newLoader>>
// => any
```

peter-skillmasters:
> If you're using single fetch, I believe this is what you're looking for: https://remix.run/docs/en/main/utils/data

Thank you! That's very helpful. The deprecation message should probably point to this as an alternative

PR #10140, opened 2024-10-19
Missing version makes Yarn workspace fail

Arthurmtro:
Actually, it's not possible to create a new remix project in a Yarn workspace and directly install packages, you have to manually add a version
image

By doing this change, we just ensure Remix can be seamlessly integrated into a yarn workspace :)


Arthurmtro:
Hello !
Any news on that possible quick fix ? :)

PR #10145, opened 2024-10-21
Update single fetch deprecation notice to also include info about null

JakeGinnivan:
Resource routes can no longer just return null, call this out in the deprecation notice.

It just took me a minute to realise it was the null which was triggering the message.

### DIFF
--- a/.changeset/grumpy-llamas-leave.md
+++ b/.changeset/grumpy-llamas-leave.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/server-runtime": patch
+---
+
+Update externally-accessed resource routes warning to cover null usage as well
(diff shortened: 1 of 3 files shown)


JakeGinnivan:
I'll update the tests if this is actually correct. Let me know.

JakeGinnivan:
Done @brophdawg11. Thanks

Let me know if I need to squash/apply conventional commits. This started as a bit of a drive by, haha

brophdawg11:
Looks good!  No need to change any commits - I'll squash-merge the PR.

Would you mind adding a [changeset](https://github.com/remix-run/remix/pull/10145#issuecomment-2425682413) as well?  Should be able to just type `pnpm changeset` and follow the prompts for a patch change for the server-runtime package

JakeGinnivan:
Done @brophdawg11. Thanks for your help

brophdawg11 (approved):

PR #10173, opened 2024-10-29
Deprecate SerializeFrom

brophdawg11:


kentcdodds:
I'm not sure how to handle these cases without `SerializeFrom`:

- https://github.com/epicweb-dev/epic-stack/blob/6c7eb1d529a15e0d32c49b310e8b0a7711aa01f9/app/utils/user.ts#L2
- https://github.com/epicweb-dev/epic-stack/blob/6c7eb1d529a15e0d32c49b310e8b0a7711aa01f9/app/routes/settings%2B/profile.connections.tsx#L7
- https://github.com/epicweb-dev/epic-stack/blob/6c7eb1d529a15e0d32c49b310e8b0a7711aa01f9/app/routes/users%2B/%24username_%2B/__note-editor.tsx#L12

And I have some here as well: https://github.com/search?q=repo%3Aepicweb-dev%2Fepicshop%20serializefrom&type=code

Am I missing something?

bpetetot:
@kentcdodds 

I was able to reproduce the behavior of `SerializeFrom` by doing:

```
type SerializeFrom<T> = ReturnType<typeof useLoaderData<T>>;
```

It can be used in the same way as the current `SerializeFrom` for data from `loader` functions. You can make an equivalent type with `useActionLoader` if you need it.

artemis-prime:
Please, please, please document this better!  What a PITA to have to poke around to use such core functionality!

felixmokross:
To me it seems the current implementation in RR7 doesn't cover all that `SerializeFrom` in Remix used to cover. I used it in two ways so far:

```ts
// Given these type definitions:
type SerializeFrom<T> = ReturnType<typeof useLoaderData<T>>;
type Car = {
  …
  price: Prisma.Decimal;
};

// 1)
type ResultDto = SerializeFrom<typeof loader>;

// 2)
type CarDto = SerializeFrom<Car>;
```

Use case (1) still works since I upgraded from Remix to React Router (incl. migration to Single Fetch). But if I use it as shown in (2), `CarDto` just is the same as `Car`, it doesn't serialize the props. I use Prisma Decimals, where this is an issue.

If I look into the RR7 package, I see the following type definition for the return type of `useLoader`:
```ts
type SerializeFrom<T> = T extends (...args: infer Args) => unknown ? Args extends [ClientLoaderFunctionArgs | ClientActionFunctionArgs] ? ClientDataFrom<T> : ServerDataFrom<T> : T
```

So if `T` is not a function, it just returns `T`. Shouldn't it return `Serialize<T>`?

To work around the issue, I see that I could define `CarDto` like this (assuming we have a `car` prop on the object returned by the loader):

```ts
type CarDto = SerializeFrom<typeof loader>["car"];
```
But I would like to avoid having to always use a specific `loader` to define `CarDto`. I have centralized type files being used in many places which should not depend on specific routes.

PR #10174, opened 2024-10-29
Deprecate eslint config package

brophdawg11:


kentcdodds:
Anyone coming here disappointed about the deprecation of the lint config should take a look at what I did about it: https://github.com/epicweb-dev/config

PR #10225, opened 2024-11-13
gotchas.md: remove typeof window gotcha

avindra:
Deno since version 2 has erased the `window` global.

https://deno.com/blog/v2.0-release-candidate#changes-to-global-variables

### DIFF
--- a/docs/guides/gotchas.md
+++ b/docs/guides/gotchas.md
@@ -8,30 +8,6 @@ Rendering your app on the server and in the browser with React has some inherent
 
 This document should help you get over these bumps.
 
-## `typeof window` checks
-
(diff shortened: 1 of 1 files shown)


docs/guides/gotchas.md
```diff
@@ -8,30 +8,6 @@ Rendering your app on the server and in the browser with React has some inherent
...
-
-Because the same JavaScript code can run in the browser as well as the server, sometimes you need to have a part of your code that only runs in one context or the other:
-
-```ts bad
-if (typeof window === "undefined") {
-  // running in a server environment
-} else {
-  // running in a browser environment
-}
-```
```
brookslybrand, line 21: This part still seems relevant, it seems like only the deno specific part should be removed

PR #10254, opened 2024-11-21
feat: use `?inline` to load critical styles

sapphi-red:
The default export of css files was removed in Vite v5, but it was accidently left for SSR. Vite v6 removes that (https://github.com/vitejs/vite/pull/17922). This PR changes the critical style loader to use `?inline` imports to make remix compatible with Vite v6.

Testing Strategy: I ran `pnpm i && pnpm vite-ecosystem-ci:build && pnpm vite-ecosystem-ci:test` with both Vite 5.x and 6.x.


markdalgleish (commented):
Thanks for the PR! Could you point this at the `dev` branch instead, and add a changeset?

PR #10255, opened 2024-11-21
feat: support Vite 6 HMR

sapphi-red:
In Vite v5 and below, Vite mixed the client modules and the SSR modules in a single module graph. This caused a unnecessary HMR when a file that only affects a single environment (e.g. editing a SSR only module triggered a HMR on the client side unconditionally).

In Vite v6, the module graphs are separated for each environment to give a way to remove that unnecessary HMR. But that means the frameworks now have to trigger HMR on their own when a module in a different environment is changed.

This PR implements that HMR triggering logic. It triggers the client HMR when the SSR module is changed like it was working in Vite 5. This code triggers unnecessary HMR as same as Vite v5, but at least makes remix work with Vite 6. I think that can be improved in a separate PR.

Testing Strategy: I ran `pnpm i && pnpm vite-ecosystem-ci:build && pnpm vite-ecosystem-ci:test` with both Vite 5.x and 6.x.

### DIFF
--- a/.changeset/cyan-bats-work.md
+++ b/.changeset/cyan-bats-work.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/dev": minor
+---
+
+Add support for Vite 6 (hot update related parts)
--- a/contributors.yml
+++ b/contributors.yml
@@ -591,6 +591,7 @@
 - sambostock
 - sandstone991
 - sandulat
+- sapphi-red
 - sarahse
 - sathvik-k
 - sbernheim4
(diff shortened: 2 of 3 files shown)


markdalgleish (commented):
Thanks for the PR! Same as https://github.com/remix-run/remix/pull/10254, could you point this at the `dev` branch and add a changeset?

sapphi-red:
I did the same with #10254 🙂

markdalgleish (commented):
Now that React Router v7 is in vite-ecosystem-ci (https://github.com/vitejs/vite-ecosystem-ci/pull/348) and the Remix test has been removed (https://github.com/vitejs/vite-ecosystem-ci/pull/353), we're free to leave Remix without Vite 6 support. Anyone wanting to use Vite 6 should migrate to React Router v7.

PR #10261, opened 2024-11-22
docs(file-conventions/remix-config): Remove hidden metadata from front matter for remix-config

Jr-14:
I noticed that remix-config was missing in the sidebar on File conventions. The link is also hidden and tucked away in https://remix.run/docs/en/main/guides/migrating-react-router-app#configuration. I thought it'd be good to have remix config displayed in File conventions for easier access.

`remix.config.js` is missing in File Conventions
image

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -330,6 +330,7 @@
 - joshball
 - JoshuaKGoldberg
 - jplhomer
+- Jr-14
 - jrf0110
 - jrubins
 - jsbmg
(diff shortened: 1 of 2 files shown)


Jr-14:
Had a read through https://remix.run/docs/en/main/guides/vite#classic-remix-compiler-vs-remix-vite - will close this PR as remix vite is what the project wants to use moving forward

PR #10306, opened 2024-12-05
refactor: remove redundant '@remix-run/node/install' import

kt3k:
`remix-serve` cli seems calling `installGlobals` function of `@remix-run/node` package twice. In my usage, it's called with 2 different `nativeFetch` flags and that seems causing unnecessary dependency (`@remix-run/web-fetch`) loaded on the server runtime.

This PR removes the first call of `installGlobals` (by removing import of `@remix-run/node/install`), and tries to avoid the above situation.

### DIFF
--- a/packages/remix-serve/cli.ts
+++ b/packages/remix-serve/cli.ts
@@ -1,4 +1,3 @@
-import "@remix-run/node/install";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";


kt3k:
Note:
- The only thing `@remix-run/node/install` does is calling `installGlobals` with default args. https://github.com/remix-run/remix/blob/main/packages/remix-node/install.js
- `run` function in `remix-serve` calls `installGlobals` with `nativeFetch` flag. This call overwrites the globals installed by the above call.

MichaelDeBoey (approved):

brophdawg11:
Thanks for the PR!  Would you mind repointing this to the `dev` branch?

.changeset/eighty-beans-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@remix-run/serve": patch
+---
+
+Remove redundant '@remix-run/node/install' import
```
brophdawg11, line 5:
```suggestion
Remove redundant '@remix-run/node/install' import from `remix-serve` because it manually calls `installGlobals` separately
```

PR #10317, opened 2024-12-09
Throw single fetch redirects fetched from the server

### LINKED ISSUE #10208: handling of redirects encountered in `serverLoader` changes with Single Fetch
### Reproduction

https://github.com/rossipedia/await-single-fetch-shenanigans

https://stackblitz.com/~/rossipedia/await-single-fetch-shenanigans

### System Info

```shell
System:
    OS: macOS 15.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 26.83 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.11.0 - ~/.local/state/fnm_multishells/43462_1731267897329/bin/node
    Yarn: 1.22.22 - ~/.local/state/fnm_multishells/43462_1731267897329/bin/yarn
    npm: 10.9.0 - ~/.local/state/fnm_multishells/43462_1731267897329/bin/npm
    pnpm: 9.12.0 - ~/.local/state/fnm_multishells/43462_1731267897329/bin/pnpm
    bun: 1.0.2 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 111.1.49.132
    Chrome: 130.0.6723.117
    Edge: 130.0.2849.80
    Safari: 18.1
```

### Used Package Manager

`npm`

### Expected Behavior

> [!NOTE]
> I originally thought this was an issue with the `<Await>` component, but I've since realized that the underlying behavioral change is with `serverLoader()`, and I assume by extension `turbo-stream`

With `v3_singleFetch` enabled, a redirect `Response` received from `serverLoader()` should result in a rejected `Promise`.

### Actual Behavior

With `v3_singleFetch` enabled, a redirect `Response` received from `serverLoader()` results in a resolved `Promise`.

When used with `<Await>`, note that the `Response` instance is passed to the `children` function if provided, which breaks the type contract for `<Await>`:

```tsx
export function loader() {
  if (someCondition) {
    throw redirect('/');
  }

  return {
    foo: 'bar',
  };
}

export function clientLoader({ serverLoader }: ClientLoaderFunctionArgs) {
  return {
    promise: serverLoader<typeof loader>(),
  }
}

clientLoader.hydrate = true;

export default function RouteComponent() {
  const { promise } = useLoaderData<typeof clientLoader>();

  return (
    <Suspense fallback={<>Loading...</>}>
      <Await resolve={promise}>
        {/*
        Here, the typings say that data is of type `{ foo: string }`, but at runtime
        if `someCondition` is true it will be a `Response` instance with a 302 status
        and `Location` header.
        */}
        {(data) => <Render data={data} />}
      </Await>
    </Suspense>
  );
}
```

brophdawg11:
Align the logic with [non-Single Fetch](https://github.com/remix-run/remix/blob/main/packages/remix-react/routes.tsx#L588) to always throw redirects from the client loader so short circuit any further processing and avoid type inference mismatches

Closes #10208

### DIFF
--- a/packages/remix-react/single-fetch.tsx
+++ b/packages/remix-react/single-fetch.tsx
@@ -491,7 +491,7 @@ function unwrapSingleFetchResult(result: SingleFetchResult, routeId: string) {
     if (result.replace) {
       headers["X-Remix-Replace"] = "yes";
     }
-    return redirect(result.redirect, { status: result.status, headers });
+    throw redirect(result.redirect, { status: result.status, headers });
   } else if ("data" in result) {
     return result.data;
   } else {
--- a/.changeset/unlucky-melons-promise.md
+++ b/.changeset/unlucky-melons-promise.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/react": patch
+---
+
+Throw unwrapped single fetch redirect to align with pre-single fetch behavior


packages/remix-react/single-fetch.tsx
```diff
@@ -491,7 +491,7 @@ function unwrapSingleFetchResult(result: SingleFetchResult, routeId: string) {
     if (result.replace) {
       headers["X-Remix-Replace"] = "yes";
     }
-    return redirect(result.redirect, { status: result.status, headers });
+    throw redirect(result.redirect, { status: result.status, headers });
```
brophdawg11, line 494: See https://github.com/remix-run/remix/blob/main/packages/remix-react/routes.tsx#L588

rossipedia:
❤️

jacob-ebey (approved):

PR #10330, opened 2024-12-13
docs: add warnings to use React Router

alexanderson1993:
Happy to change the copy if necessary. I'm pretty sure I've got all of the major starting point pages covered.


brookslybrand (approved):
Thanks so much @alexanderson1993, this is great and very helpful. I very much look forward to more people telling us we made a mistake in "killing" Remix 😉

Thanks for all you've done to support the Remix community. Really excited for what comes next under the Remix umbrella of OSS

PR #10351, opened 2024-12-19
Add Vite v6 support

markdalgleish:


integration/vite-route-added-test.ts
```diff
@@ -38,7 +38,12 @@ test.describe(async () => {
   });
   test.afterAll(() => stop());
 
-  test("Vite / dev / route added", async ({ page }) => {
+  test("Vite / dev / route added", async ({ page, browserName }) => {
+    test.skip(
+      browserName === "webkit",
+      "Safari caches too aggressively, browser manifest is cached with old routes"
+    );
+
```
markdalgleish, line 46: We also needed to do this in React Router, seems to be triggered by the Playwright update: https://github.com/remix-run/react-router/pull/11511

package.json
```diff
@@ -53,7 +53,7 @@
     "@octokit/graphql": "^4.8.0",
     "@octokit/plugin-paginate-rest": "^2.17.0",
     "@octokit/rest": "^18.12.0",
-    "@playwright/test": "^1.33.0",
+    "@playwright/test": "^1.49.1",
```
markdalgleish, line 56: It's not clear to me why it was related, but attempting to run tests against Vite 6 raised some errors that are fixed in later versions of Playwright.

.gitignore
```diff
@@ -14,7 +14,7 @@ tsconfig.tsbuildinfo
 /fixtures/my-remix-app
 /fixtures/deno-app
 /integration/playwright-report
-/test-results
+test-results
```
markdalgleish, line 17: I noticed the test-results directory within integration was showing up in my git diffs.

packages/remix-dev/__tests__/cli-test.ts
```diff
@@ -8,7 +7,7 @@ import semver from "semver";
 let execFile = util.promisify(childProcess.execFile);
 
 const TEMP_DIR = path.join(
-  fse.realpathSync(os.tmpdir()),
+  path.join(__dirname, ".tmp"),
```
markdalgleish, line 10: This was changed so that Vite can be resolved from cwd.

markdalgleish (commented):
Now that React Router v7 is in vite-ecosystem-ci (https://github.com/vitejs/vite-ecosystem-ci/pull/348) and the Remix test has been removed (https://github.com/vitejs/vite-ecosystem-ci/pull/353), we're free to leave Remix without Vite 6 support. Anyone wanting to use Vite 6 should migrate to React Router v7.

amitdahan:
> Anyone wanting to use Vite 6 should migrate to React Router v7.

Can this be reconsidered? 

We are migrating an app that currently still uses React Router v5, I've had success migrating to Remix (essentially rendering our entire previous app, within a single Remix route).

But when I tried the same with React Router v7, I encountered issues due to having 2 different majors of `react-router`.
Not sure whether React Router v7 intentionally doesn't support gradual migration from v5, or even v6, or if this is a bug with v7?)

markdalgleish:
We've seen some HMR issues with Remix in Vite 5 that are fixed by the upgrade to Vite 6, so I'm re-opening this PR since the work has already been done.

markdalgleish:
@amitdahan Yep, if you're running into issues with React Router, it'd be great if you could raise them over there.

Issue #10232, opened 2024-11-15
404 error using redirect with v3_lazyRouteDiscovery


srod:
### Reproduction

I'm doing a redirect in a AuthLayout (AuthLayout.tsx) according if session is true/false

And when navigating to /login I have a 404 error

Small reproduction : https://github.com/srod/remix-v3_lazyRouteDiscovery

If you go with your browser to /protected you will have that 404

Error:
No routes matched location "/login"
ErrorResponseImpl {status: 404, statusText: 'Not Found', internal: true, data: 'Error: No route matches URL "/login"',

Discord link if needed: https://discord.com/channels/770287896669978684/1306915010701824063

### System Info

```shell
System:
    OS: macOS 15.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 95.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.11.0 - ~/.config/nvm/versions/node/v22.11.0/bin/node
    Yarn: 1.22.22 - ~/.config/nvm/versions/node/v22.11.0/bin/yarn
    npm: 10.9.0 - ~/.config/nvm/versions/node/v22.11.0/bin/npm
    pnpm: 9.12.3 - ~/.config/nvm/versions/node/v22.11.0/bin/pnpm
    bun: 1.1.34 - ~/.bun/bin/bun
  Browsers:
    Chrome: 131.0.6778.69
    Edge: 131.0.2903.48
    Safari: 18.1
  npmPackages:
    @remix-run/dev: ^2.14.0 => 2.14.0
    @remix-run/node: ^2.14.0 => 2.14.0
    @remix-run/react: ^2.14.0 => 2.14.0
    @remix-run/serve: ^2.14.0 => 2.14.0
    vite: ^5.1.0 => 5.4.11
```

### Used Package Manager

npm

### Expected Behavior

Navigate to /login without 404 error.

### Actual Behavior

When navigate to /login we have an error.

maximeclaisse:
This issue also happens in this very simple case :

1. Create a remix app with ` npx create-remix@latest`
2. Create a valid page such as `export default function TestPage() { return <div>Test</div>; }`
3. In _index.tsx, add `useEffect(() => { navigate("/test");  }, []);`
4. Get a 404
5. Refresh the page and get actual page

DerJacques:
[...]

PR #10409, opened 2025-01-16
Bug report: 404 error when using navigate() with v3_lazyRouteDiscovery

jabrks:
Demonstrates the issue described in #10232 with an integration test. I'm not sure if this is still an issue in React Router 7, but I'd be happy to check if that would be useful. I encountered this after enabling the future flag whilst attempting to migrate an application over from Remix.

This is more evident in dev mode, so I borrowed the test setup from [`vite-dev-test.ts`](https://github.com/remix-run/remix/blob/c2a2c34801f1bcb21a6c79b56f9e10d9d8e61c9d/integration/vite-dev-test.ts)


jabrks:
I saw a similar report in https://github.com/remix-run/react-router/pull/12853 which made me realise that this isn't actually specific to Vite development mode, but simply more evident there due to effects running twice in development in Strict Mode

jabrks:
This was fixed in https://github.com/remix-run/remix/pull/10447

PR #10410, opened 2025-01-16
Properly handle empty responses in single fetch requests

### LINKED ISSUE #10328: Using `data` function with `single_fetch` enabled breaks when returning a status of 204
### Reproduction

https://stackblitz.com/edit/remix-run-remix-mem1r9ss?file=app%2Froutes%2F_index.tsx

There are two pages (`index` and `okay` with links on the page to each other)

Click the submit.  the `index page` has the action that returns a 204 and breaks, the `okay page` returns 200 and works.

### System Info

```shell
System:
    OS: macOS 15.1.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 175.98 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    Yarn: 1.22.22 - ~/.yarn/bin/yarn
    npm: 10.8.2 - ~/.nvm/versions/node/v20.17.0/bin/npm
    pnpm: 7.26.3 - ~/Library/pnpm/pnpm
  Browsers:
    Brave Browser: 129.1.70.126
    Chrome: 131.0.6778.140
    Edge: 131.0.2903.86
    Safari: 18.1.1
  npmPackages:
    @remix-run/dev: ^2.15.1 => 2.15.1
    @remix-run/node: ^2.15.1 => 2.15.1
    @remix-run/react: ^2.15.1 => 2.15.1
    @remix-run/serve: ^2.15.1 => 2.15.1
    vite: ^5.1.0 => 5.4.11
```

### Used Package Manager

yarn

### Expected Behavior

An action that uses the `data` function should be able to return a 204 (no content).  In my application, I have routes that are only actions that do not return data used as a quasi API (for setting user prefs, for example) that don't return anything (hence the 204). 

### Actual Behavior

An action that uses the `data` function and returns a 204 status fails with `Unable to decode turbo-stream response from URL...` when single_fetch is enabled.

brophdawg11:
Remix v2 version of https://github.com/remix-run/react-router/pull/12760 to ease the upgrade path to `v3_singleFetch -> RR v7`

Closes #10328

### DIFF
--- a/integration/package.json
+++ b/integration/package.json
@@ -10,7 +10,7 @@
   "dependencies": {
     "@cloudflare/kv-asset-handler": "^0.3.0",
     "@cloudflare/workers-types": "^4.20230518.0",
-    "@playwright/test": "^1.33.0",
+    "@playwright/test": "^1.49.1",
     "@remix-run/dev": "workspace:*",
     "@remix-run/express": "workspace:*",
     "@remix-run/node": "workspace:*",
--- a/.changeset/giant-dolls-wonder.md
+++ b/.changeset/giant-dolls-wonder.md
@@ -1,0 +1,6 @@
+---
+"@remix-run/react": patch
+"@remix-run/server-runtime": patch
+---
+
+Properly handle status codes that cannot have a body in single fetch responses (204, etc.)
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ tsconfig.tsbuildinfo
 /fixtures/my-remix-app
 /fixtures/deno-app
 /integration/playwright-report
-/test-results
+/integration/test-results
 /uploads
 
 .eslintcache
--- a/integration/single-fetch-test.ts
+++ b/integration/single-fetch-test.ts
@@ -1977,6 +1977,14 @@ test.describe("single-fetch", () => {
       }
     });
 
+    // Document requests
+    let documentRes = await fixture.requestDocument("/?index", {
+      method: "post",
+    });
+    expect(documentRes.status).toBe(204);
+    expect(await documentRes.text()).toBe("");
+
+    // Data requests
     await app.goto("/", true);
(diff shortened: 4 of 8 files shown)


integration/package.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": {
     "@cloudflare/kv-asset-handler": "^0.3.0",
     "@cloudflare/workers-types": "^4.20230518.0",
-    "@playwright/test": "^1.33.0",
+    "@playwright/test": "^1.49.1",
```
brophdawg11, line 13: Needed specifically to pick up changes in [1.37](https://github.com/microsoft/playwright/releases/tag/v1.37.0) that fix the running of tests on ubuntu

FELLIX000: Trying to make a Chain Based crypto company that uses Felix Tokens as a new  crypto currency called TX  Use Felix Coins to either trade for other currency that are part of Chainlink or God of Chains or any other currency that are based on chaining your lost wallet to your new wallet if you loose it Patrica Tree aka Genies was full of lost wallets. I'm hoping chaining will prevent lost and stolen wallets I would be exclusive to Coinbase And Metamask and any other company willing to accept my crypto also buying Felix coins give you rewards by keeping it and stacking in return the host which would me Felix and anyone else that would like to buy Felix coins would get a big investment bc it would be like BNB but in chainlink way so everything but Felix tokens would be worth same trading value as fellow tokens Felix tokens would be for gas and collecting to make back value to be same worth as all the other currency around. Also every five min someone stealing my data or stealing my dev containers and reusing them I had locked and people where complaining that it was locked but there was point I'm trying to make I'm original person that was given Genies but I've had no luck keeping it. Everyone wants to be me or link themselves to my blockchain since of its worth sad people got steal to get ahead in life.

Issue #9353, opened 2024-05-01
Re: `@remix-run/serve` sets high `max-age` and `immutable` cache header on everything


kenteramin:
### Reproduction

1. Create a remix app
2. Run `build` then `start`
3. Open the page
4. Check the headers on the `/favicon.ico` request

The favicon from `/public` comes with `Cache-Control: public, max-age=31536000, immutable`

### System Info

```shell
remix 2.9.1
```

### Used Package Manager

npm

### Expected Behavior

The server config from
https://github.com/remix-run/remix/blob/main/packages/remix-serve/cli.ts
says `app.use(express.static("public", { maxAge: "1h" }));`.
So it is expected to have `maxAge=1h`

### Actual Behavior

We actually get an `immutable` `maxAge=1y`

kenteramin:
There was already a bug of the same nature
https://github.com/remix-run/remix/issues/903

kenteramin:
So, actually the problem is that here
```
app.use(
  build.publicPath,
  express.static(build.assetsBuildDirectory, {
    immutable: true,
    maxAge: "1y",
  })
);
app.use(express.static("public", { maxAge: "1h" }));
```

`build.assetsBuildDirectory` is just `build/client`, so it handles all the requests to the public folder.

The config should actually be something like this
```
app.use(
  build.publicPath,
  express.static(build.assetsBuildDirectory + "/assets", {
    immutable: true,
    maxAge: "1y",
  })
);
app.use(express.static(build.assetsBuildDirectory, { maxAge: "1h" }));
```

istarkov:
The main problem is that these headers are also sent on a 404 error.
A 404 on static files is not uncommon during deployment, for example on Vercel. And a cached 404 makes it impossible to fully use the site.

remorses:
This issue seems to be fixed in the react-router codebase but not Remix
https://github.com/remix-run/react-router/blob/bc1c1c8ef87b7bf97f1f168604b89a5da61939da/packages/react-router-serve/cli.ts#L81

bhouston:
I've moved away from remix:server and adopted the Fastify integration and now it works properly:

https://github.com/mcansh/remix-fastify

/assets is immutable, cached forever because it has unique URLs.

/public is no-cache, but it will serve up 304 unmodified to save bandwidth.  

Those are good defaults.

brophdawg11:
Thank you for opening this issue, and our apologies we haven't gotten around to it yet!

[...]

PR #10441, opened 2025-01-28
Do not cache user public files as immutable

remorses:
The Remix `build.assetsBuildDirectory` is `build/client`, this folder contains user public files. The serve code is caching these files as immutable which means that even if the user changes the content of a file but not the name the file it will not be revalidated. 

This PR makes immutable only files generated by Remix, which contain an hash in the name. Files in `public` will be cached only for 1 hour instead.

Fix https://github.com/remix-run/remix/issues/9353

### DIFF
--- a/.changeset/fast-bottles-repeat.md
+++ b/.changeset/fast-bottles-repeat.md
@@ -1,0 +1,5 @@
+---
+"@remix-run/serve": patch
+---
+
+Do not make user files in public folder cache as immutable, recude cache max age to 1 hour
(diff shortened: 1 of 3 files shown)


sergiodxa:
Why do you have user files in `build/client` folder? That should only include the files generated at build time and hashed. If you're uploading user files to your file system you should add them in the `public` folder.

remorses:
Files in /public folder are moved to build/client by Vite

sergiodxa:
But that's build time thing, after you build the app and deploy it any file in public folder is not touched anymore.

PR #10605, opened 2025-05-03
Update quickstart to compliant with express 5

shge:
Changing
```js
app.all("*", createRequestHandler({ build }));
```
to
```js
app.all(/(.*)/, createRequestHandler({ build }));
```
to compliant with express 5.

cf. https://github.com/expressjs/express/issues/5936

### DIFF
--- a/docs/start/quickstart.md
+++ b/docs/start/quickstart.md
@@ -162,7 +162,7 @@ const app = express();
 app.use(express.static("build/client"));
 
 // and your app is "just a request handler"
-app.all("*", createRequestHandler({ build }));
+app.all(/(.*)/, createRequestHandler({ build }));
 
 app.listen(3000, () => {
   console.log("App listening on http://localhost:3000");


MichaelDeBoey (approved):

docs/start/quickstart.md
```diff
@@ -162,7 +162,7 @@ const app = express();
 app.use(express.static("build/client"));
 
 // and your app is "just a request handler"
-app.all("*", createRequestHandler({ build }));
+app.all(/(.*)/, createRequestHandler({ build }));
```
brophdawg11, line 165: Are these funcitonally equivalent?  If so, could we use this verison instead and avoid the regex?
```suggestion
app.all("/{*splat}", createRequestHandler({ build }));
```

shge: @brophdawg11 
Yes, also confirmed `"/{*splat}"` works fine.

PR #10611, opened 2025-05-08
fix(dev): update `vite-node`

MichaelDeBoey:
Fix for https://github.com/vitest-dev/vitest/security/advisories/GHSA-9crc-q9x8-hgqq & https://github.com/vitest-dev/vitest/security/advisories/GHSA-8gvc-j273-4wm5

---

CC/ @platocrat


brophdawg11:
Can we point this to dev and add a changeset?

PR #10645, opened 2025-05-30
Import history from remix-the-web

brophdawg11:


CoderDeltaLAN:
**Context**
- Remix app
- Response contains `utf8` characters but they appear broken (mojibake)
- Happens when returning text from loader/route

**Why**
The default `Response` constructor in Web APIs does not set charset unless you include it. Browsers assume `text/plain` or `text/html` **without charset** as ISO-8859-1, not UTF-8. That’s why characters break.

**Fix (minimum viable)**

Return response with explicit charset:

```ts
// app/routes/hello.tsx (loader or action)
import { json } from "@remix-run/node"

  // Option 1: use json() helper (auto sets utf-8)
  return json({ message: "¡Hola, mundo! Привет мир 你好" })

  // Option 2: raw text with headers
  // return new Response("¡Hola!", {
  //   headers: { "Content-Type": "text/plain; charset=utf-8" },
  // })
}
```

**Checklist**
- Always set `charset=utf-8` in `Content-Type` when returning raw text/HTML:
  - `text/html; charset=utf-8`
  - `text/plain; charset=utf-8`
  - `application/json; charset=utf-8`
- If serving files (CSV, etc.), set charset there too.
- Using `json()` or `redirect()` from Remix helpers automatically adds the correct headers.

**Docs**
- MDN Response API: https://developer.mozilla.org/docs/Web/API/Response/Response
- Remix response helpers: https://remix.run/docs/en/main/utils/json

If this helps, please **Mark as answer** so others can find it.

PR #10646, opened 2025-06-01
chore: remove unused tar fixtures

43081j:
`tar-fs` was a production dependency of `remix-dev` but only used in an unreferenced test fixture. This removes the fixture and the dependency.

If I just missed a usage somewhere, we can reintroduce the fixture but move the dependency to dev dependencies it seems.


43081j:
I actually meant to push this to the react-router repo!

let me know if its still worth doing here

MichaelDeBoey:
Done in https://github.com/remix-run/react-router/pull/13740

PR #10652, opened 2025-06-02
docs(guides): fix grammar + typos

MichaelDeBoey:

PR #10660, opened 2025-06-02
docs(decisions): fix grammar + typos

MichaelDeBoey:


brookslybrand:
See https://github.com/remix-run/remix/pull/10652#issuecomment-2950198421

decisions/0005-remixing-react-router.md
```diff
@@ -30,31 +30,31 @@ In [Remixing React Router][remixing router], Ryan gives an overview of the work
...
 
-There's one caveat here when it comes to back/forward button usage. When the user navigates back/forward in the history stack we get a `popstate` event _but the URL has already been updated_. So the best we can do there is react to the new URL. This is _not_ what the browser would do in a non-JS world, but we really have no choice.
+There's one issue here when it comes to back/forward button usage. When the user navigates back/forward in the history stack, we get a `popstate` event _but the URL has already been updated_. So the best we can do there is react to the new URL. This is _not_ what the browser would do in a non-JS world, but we really have no choice.
```
MichaelDeBoey:
```suggestion
There's one caveat here when it comes to back/forward button usage. When the user navigates back/forward in the history stack, we get a `popstate` event _but the URL has already been updated_. So the best we can do there is react to the new URL. This is _not_ what the browser would do in a non-JS world, but we really have no choice.
```

decisions/0005-remixing-react-router.md
```diff
@@ -132,7 +132,7 @@ Normal POST submissions that do not redirect will use a `REPLACE`:
 - Navigates to `/login` (history stack is `[/, /login]`)
 - Fills out and submits the `<Form method="post">`
 - Action does not redirect
-  - At this point, if the action returns a non-redirect and we were to PUSH the navigation we'd end up with a history stack of `[/, /login, /login]` and the user would be in a scenario where it would take them 2 back buttons to get "through" the login page from a subsequent route.
+  - At this point, if the action returns a non-redirect and we were to PUSH the navigation, we'd end up with a history stack of `[/, /login, /login]` and the user would be in a scenario where it would take them two back buttons to get "through" the login page from a later route.
```
MichaelDeBoey:
```suggestion
  - At this point, if the action returns a non-redirect and we were to PUSH the navigation, we'd end up with a history stack of `[/, /login, /login]` and the user would be in a scenario where it would take them two back buttons to get "through" the login page from a subsequent route.
```

decisions/0010-splitting-up-client-and-server-code-in-vite.md
```diff
@@ -61,18 +61,18 @@ The Plugin API makes this apparent:[^1]
...
+This approach lets Vite load and transform each module _once_ and cache the result[^2], which is a keystone for its speed.
 
 #### Handling mixed modules
 
 While `.server` modules are a great way to separate client and server code in most cases,
 there will always be a need to stitch together modules that mix client and server code.
 For example, you may want to migrate from the previous compiler to Vite without needing to manually split up mixed modules.
 
-But supporting mixed modules directly in Remix would require compile-time magic which would add substantial complexity.
+But supporting mixed modules directly in Remix would require compile-time magic, which would add significant complexity.
```
MichaelDeBoey:
```suggestion
But supporting mixed modules directly in Remix would require compile-time magic, which would add substantial complexity.
```

PR #10693, opened 2025-07-24
fix(tar-parser): fix for large repos

remorses:
Currently tar-parses fails to parse large tar files, this is because there is an edge case that is not handled: the problem was in the #parseBody() method where the parser wasn't handling the case when a file's content ended exactly at a buffer boundary.


remorses:
@mjackson

PR #10695, opened 2025-07-24
refactor(node-fetch-server): add type-safety to `socket` check in `createRequest`

MichaelDeBoey:
The actual code is doing exactly the same, but is introducing type guards for some extra type-safety inside the code itself.

I took inspiration from how React Router handles this check
https://github.com/remix-run/react-router/blob/87912477ce25a501f8cac71c1105dfe3b0bff5ca/packages/react-router-dev/vite/node-adapter.ts#L53-L55

This check was originally introduced in @TrySound's https://github.com/remix-run/remix/pull/10199 and resubmitted in https://github.com/remix-run/react-router/pull/13746


mjackson:
Thanks, @MichaelDeBoey, but I don't think we need this check.

PR #10708, opened 2025-08-01
docs:(guides/data-loading): mention Prisma Postgres

aidankmcalister:
Added a mention for [Prisma Postgres](https://prisma.io/postgres) so users have quick access to a quick and easy database

### DIFF
--- a/contributors.yml
+++ b/contributors.yml
@@ -15,6 +15,7 @@
 - AhmadMayo
 - ahmedeldessouki
 - ahuth
+- aidankmcalister
 - aiji42
 - ainoya
 - airjp73


brophdawg11:
Thank you for the PR but I don't think we want to go down the road of linking off to specific solutions in this space because there are too many and we don't want to appear too preferential.  We already have a link to Prisma so if they want prisma + postgres, they should be able to find that on their own.

brophdawg11:
Also I just realized this is the Remix repo - Remix v2 is [deprecated](https://github.com/remix-run/remix?tab=readme-ov-file#welcome-to-remix) in favor of React Router v7, so the new version of this doc is [here](https://reactrouter.com/start/framework/data-loading) and it no longer even has a specific call out to Prisma, so we're not making any specific recommendations anymore.

PR #10723, opened 2025-08-27
fix(tar-parser): fix for large repos

remorses:
## Summary
Fix tar-parser failing to parse large tar files due to edge case in buffer boundary handling.

## Problem
The tar-parser was failing when parsing large tar files. The issue occurred in the `#parseBody()` method where the parser wasn't properly handling the case when a file's content ended exactly at a buffer boundary.

## Solution
This PR addresses the edge case to ensure proper parsing of tar files regardless of size or how file contents align with buffer boundaries.


remorses:
@mjackson

remorses:
bump. is the tar package still broken?

mjackson:
Superseded by #11264.

PR #10726, opened 2025-08-27
fix(node-fetch-server): fix logic for aborting requests

MichaelDeBoey:
Taken from @brophdawg11's https://github.com/remix-run/remix/pull/10046 & https://github.com/remix-run/react-router/commit/d386c0d7ee055a1cbea12448899121b72facfde3


packages/node-fetch-server/src/lib/request-listener.ts
```diff
@@ -128,10 +128,14 @@ export function createRequest(
...
-    controller.abort()
-  })
+  let controller: AbortController | null = new AbortController()
+
+  // Abort once we can no longer write a response if we have
+  // not yet sent a response (i.e., `close` without `finish`)
+  // `finish` -> done rendering the response
+  // `close` -> response can no longer be written to
+  res.on('close', () => controller?.abort())
+  res.on('finish', () => (controller = null))
```
brophdawg11, line 138: @mjackson Here's a [small repro](https://stackblitz.com/edit/stackblitz-starters-x9nmv79p) of the issue we discovered in React Router that we fixed with this change.  If we only use the `close` event to trigger aborting request signals, then we inadvertently _also_ abort on successful/non-interrupted requests:

```js
import * as http from 'node:http';
import { createRequestListener } from '@remix-run/node-fetch-server';

function handler(request) {
  // This should only fire if the request was interrupted
  request.signal.addEventListener('abort', () => console.log('request aborted'));
  return new Response('Hello, world!');
}

let server = http.createServer(createRequestListener(handler));

server.listen(3000);
```

Instead, we want to listen for a `close` event that was not preceded by a `finish` event.  `finish` means we successfully finished sending response, so there is no longer any need to abort once we receive that.

mjackson:
Thanks, @MichaelDeBoey 🙏

mjackson:
Released in node-fetch-server v0.8.1

PR #10729, opened 2025-09-01
route-pattern: Add some (failing?) edge cases

machour:
Not sure if this is by design, or if it's an overlooked edge case, so I'm opening this PR to highlight some mismatches when using accents / arabic / emojis in URL.

PHP, for example, automatically decodes these for me.


pcattori:
Superseded by #11201

pcattori:
Thanks for the patience @machour ! Finally got `route-pattern` stable enough to tackle this

PR #10745, opened 2025-09-12
feat(node-fetch-server): support `statusText` in HTTP/1 responses

brophdawg11:
When looking into https://github.com/remix-run/react-router/pull/13927 (specifically https://github.com/remix-run/react-router/pull/13460), for compatibility so we can use `@remix-run/node-fetch-server` directly in the RR Vite plugin, I noticed that we were using `writeHead` but ignoring the `Response` `statusText` field. 

`statusText` isn't supported in HTTP/2, but can be used in HTTP/1 if you want custom status messages.  

https://github.com/remix-run/react-router/pull/13460

```js
import * as http from 'node:http';
import { createRequestListener } from '@remix-run/node-fetch-server';

function handler(request) {
  return new Response('Hello, world!', {
    status: 418,
    // ❌ This is not included on the response
    statusText: "I'm not a teapot, you are!",
  });
}

let server = http.createServer(createRequestListener(handler));

server.listen(3000);
```

### DIFF
--- a/packages/node-fetch-server/CHANGELOG.md
+++ b/packages/node-fetch-server/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This is the changelog for [`node-fetch-server`](https://github.com/remix-run/remix/tree/main/packages/node-fetch-server). It follows [semantic versioning](https://semver.org/).
 
+## HEAD
+
+- Support `statusText` in HTTP/1 responses (#10745)
+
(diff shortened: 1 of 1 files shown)


brophdawg11:
Failing typechecks in CI are unrelated to this PR: 

Screenshot 2025-09-12 at 11 34 12 AM

mjackson:
Looks good, thanks @brophdawg11! 🙏

PR #10756, opened 2025-09-21
feat(patternOfUrl): added urlpattern bridge

akshatsinha0:
First I would like to ask if exporting allowed? ...or if it need more expansions?
added urlpat to @remix-run/route-pattern. This return a native URLPattern for pathname only patterns (vars,etc). Added a createRouter helper inside route-pattern getter(pat,handler), all(pattern, handler), handle(request) will uses the concerned pattern when available, Fallback to RoutePattern.match otherwise
Exposes both urlpat and createRouter from the route-pattern package index.


MichaelDeBoey:
@akshatsinha0 We now have [`@remix-run/fetch-router`](https://github.com/remix-run/remix/tree/main/packages/fetch-router)

PR #10762, opened 2025-09-26
feat(fetch-proxy): use `X-Forwarded-Port` header when passing `xForwardedHeaders` option

jonsmorrow:
This change updates the logic for xForwardedHeaders=true so that hostname instead of host is set in X-Forwarded-Host so that the port is never included. Additionally, port is set in X-Forwarded-Port.


jonsmorrow:
I'm not 100% why the tests are failing for multipart-parser. They pass for me locally. They appear to pass in the workflow before the workflow fails.

mjackson:
Superseded by #11265.

mjackson:
Thanks for the PR, @jonsmorrow. I chose to leave the port in `X-Forwarded-Host` to mimic the behavior of other well-known proxies like node-http-proxy and add a separate `X-Forwarded-Port` header that should help satisfy your use case.

PR #10768, opened 2025-10-10
feat(fetch-router): add alias functions for standard HTTP verb paths

aaronjordan:
Drafted an idea I had watching the reveal livestream. This PR adds functional aliases for simple HTTP routes.

The route objects returned are super simple here, but providing them as functions to call would make the framework appear more consistent when defining a route using helpers. For instance, this would be an existing way to express some routes:

```ts
let routes = createRoutes({
  home:  { method: 'GET', pattern: '/' },
  books: resources('books'),
  profile: resource('profile'),
})
```

After this change, we would be able to use a consistent syntax for resources, formActions, and single-method routes.

```ts
let routes = createRoutes({
  home: get('/'),
  books: resources('books'),
  profile: resource('profile'),
})
```

I followed the `resource.ts` file to try to get the style right, please do advise on anything I could update!

### DIFF
--- a/packages/fetch-router/src/lib/route-helpers.ts
+++ b/packages/fetch-router/src/lib/route-helpers.ts
@@ -6,53 +6,53 @@ import type { RequestMethod } from './request-methods'
  * Shorthand for a DELETE route.
  */
 export function createDestroy<T extends string | RoutePattern>(pattern: T) {
-  return { method: 'DELETE', pattern: pattern } as BuildRouteObject<'DELETE', T>
+  return { method: 'DELETE', pattern: pattern } as RouteDefObject<'DELETE', T>
 }
 
 /**
  * Shorthand for a GET route.
  */
 export function createGet<T extends string | RoutePattern>(pattern: T) {
-  return { method: 'GET', pattern: pattern } as BuildRouteObject<'GET', T>
+  return { method: 'GET', pattern: pattern } as RouteDefObject<'GET', T>
 }
 
 /**
  * Shorthand for a HEAD route.
  */
 export function createHead<T extends string | RoutePattern>(pattern: T) {
-  return { method: 'HEAD', pattern: pattern } as BuildRouteObject<'HEAD', T>
+  return { method: 'HEAD', pattern: pattern } as RouteDefObject<'HEAD', T>
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/fetch-router/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ This is the changelog for [`fetch-router`](https://github.com/remix-run/remix/tr
 - Export `InferRouteHandler` and `InferRequestHandler` types
 - Re-export `FormDataParseError`, `FileUpload`, and `FileUploadHandler` from `@remix-run/form-data-parser`
 - Fix an issue where per-route middleware was not being applied to a route handler nested inside a route map with its own middleware
+- Adds functional aliases for routes that respond to a single HTTP verb
```
aaronjordan: Would this actually go under a new 0.7.0 heading?

sergiodxa: The CHANGELOG is updated when a release is created, this if merged would not be part of an already released version

mjackson:
Thanks for the PR, @aaronjordan. I had the same thought. Could you please also add helpers for `head`, `patch`, and `options`? That would complete the set of supported request methods.

Also, I'd like to see a test along with this work that asserts the types are still what we expect them to be when these helpers are used in a larger route config.

As for the CHANGELOG, we include new changes under a `## Unreleased` heading at the top of the CHANGELOG that gets swapped out with the release version when we cut a new release.

aaronjordan:
Thank you for the feedback! I've updated this PR to include these additional methods and a test which revealed a couple of issues that were addressed.

1. The `delete` keyword in JavaScript wins over an imported `delete` function, so I changed the related names to `createDestroy` and `destroy`. Happy to go any other direction there if preferred.
2. I added a `RouteDefObject` type that resolves to a more specific version of [this branch of RouteDef](https://github.com/aaronjordan/remix/blob/05c7650f0420b5a087687bc383a726596f06551c/packages/fetch-router/src/lib/route-map.ts#L100).

mjackson:
Thanks again for the PR, @aaronjordan! And sorry it took so long to get it merged. The router has changed a bunch over the past month, but I always had this PR in the back of my mind. I'm going to get this published in the next release.

PR #10774, opened 2025-10-13
fix(fetch-router): make `Router`'s `fetch` spec-compliant

MichaelDeBoey:
As @rossipedia [mentioned on Discord](https://discord.com/channels/770287896669978684/940264701785423992/1426664672878268591), `Router`'s `fetch` currently isn't spec-compliant if you pass both `Request` & `init`

> A [`RequestInit`](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit) object containing any custom settings that you want to apply to the request.
> 
> If you construct a new `Request` from an existing `Request`, any options you set in an _options_ argument for the new request replace any corresponding options set in the original `Request`.

https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#options

### DIFF
--- a/packages/fetch-router/CHANGELOG.md
+++ b/packages/fetch-router/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This is the changelog for [`fetch-router`](https://github.com/remix-run/remix/tree/main/packages/fetch-router). It follows [semantic versioning](https://semver.org/).
 
+## HEAD
+
+- Fix an issue where `Router`'s `fetch` wasn't spec-compliant
+
 ## v0.6.0 (2025-10-10)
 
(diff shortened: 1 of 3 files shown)


packages/fetch-router/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 This is the changelog for [`fetch-router`](https://github.com/remix-run/remix/tree/main/packages/fetch-router). It follows [semantic versioning](https://semver.org/).
 
+## HEAD
```
mjackson, line 5: Just FYI, I switched to using the `## Unreleased` heading.

mjackson:
This is a great PR, thanks!

PR #10777, opened 2025-10-13
chore(bookstore): update `dev` scripts in `package.json`

mcansh:
renamed `dev` to `dev:server` and added a new `dev` which runs all "dev:" prefixed scripts in parallel


mjackson:
I'd like to keep the commands separate for now, so you run them in 2 separate tabs. That way the output in the console doesn't get mixed up.

PR #10778, opened 2025-10-14
docs(fetch-router): update readme blog example

jimniels:
Was trying to get the basic blog example working in val.town

See inline comments for changes that would've helped me.

### DIFF
--- a/packages/fetch-router/README.md
+++ b/packages/fetch-router/README.md
@@ -456,9 +456,7 @@ let routes = createRoutes({
   },
 })
 
-let router = createRouter({
-  parseFormData: false
-})
+let router = createRouter()
 
 // Global logging middleware
 router.use((context, next) => {
@@ -513,8 +511,7 @@ router.map(routes, {
         </html>
       `)
     },
-    async create({ request }) {
-      let formData = await request.formData()
+    async create({ formData, request }) {
       let post = await db.createPost({
         title: formData.get('title'),
         content: formData.get('content'),
@@ -555,8 +552,7 @@ router.map(routes, {
         </html>
       `)
     },
-    async update({ params, request }) {
-      let formData = await request.formData()
+    async update({ formData, params, request }) {
       await db.updatePost(params.slug, {
         title: formData.get('title'),
         content: formData.get('content'),


packages/fetch-router/README.md
```diff
@@ -174,7 +174,7 @@ router.map(routes, {
 
 ### Resource-based Routes
 
-Create resource-based route maps with the `resource` and `resources` functions. This can help DRY up your route definitions when creating RESTful APIs, Rails-style routes, etc.
+Create resource-based route maps with the `resource` and `resources` functions. This can help DRY up your route definitions when creating RESTful APIs, Rails-style routes, etc. This example illustrates [Rails-style routes](https://guides.rubyonrails.org/routing.html#resource-routing-the-rails-default):
```
jimniels, line 177: For those of us not familiar with Rails examples, just provide a link for clarity :)

packages/fetch-router/README.md
```diff
@@ -456,7 +456,9 @@ let routes = createRoutes({
   },
 })
 
-let router = createRouter()
+let router = createRouter({
+  parseFormData: false
+})
```
jimniels: I was getting an error trying to run this example

```
TypeError: Body already consumed
    at InnerBody.consume (ext:deno_fetch/22_body.js:157:32)
    at consumeBody (ext:deno_fetch/22_body.js:268:34)
    at Request.formData (ext:deno_fetch/22_body.js:349:16)
    at create (https://esm.town/v/jimniels/fetch-router@18-main/main.ts:84:36)
    at dispatch (file:///opt/render/.cache/deno/npm/registry.npmjs.org/@remix-run/fetch-router/0.6.0/dist/index.js:2814:21)
    at next (file:///opt/render/.cache/deno/npm/registry.npmjs.org/@remix-run/fetch-router/0.6.0/dist/index.js:2820:21)
    at https://esm.town/v/jimniels/fetch-router@18-main/main.ts:26:10
    at dispatch (file:///opt/render/.cache/deno/npm/registry.npmjs.org/@remix-run/fetch-router/0.6.0/dist/index.js:2823:26)
    at runMiddleware (file:///opt/render/.cache/deno/npm/registry.npmjs.org/@remix-run/fetch-router/0.6.0/dist/index.js:2832:10)
    at Router.dispatch (file:///opt/render/.cache/deno/npm/registry.npmjs.org/@remix-run/fetch-router/0.6.0/dist/index.js:2976:41)
```

Digging into the source, it looks like this is a config option I need to pass in because the router is gonna read this data by default. So your options are:

### Option 1

```js
// configure router
let router = createRouter({
  parseFormData: false
})
// read formData in your handler
async create({ request }) {
  let formData = await request.formData();
  console.log(formData);
  // ...
},
```

### Option 2

```js
// use router defaults (which reads body)
let router = createRouter()
// access formData directly
async create({ request, formData }) {
  console.log(formData);
  // ...
},
```

I went with option 1 for the docs because it was the smallest change, but this wasn't obvious to me when I copy/pasted the example on why it wasn't working

packages/fetch-router/README.md
```diff
@@ -569,8 +571,16 @@ router.map(routes, {
 })
 
 export { router }
+
+// Use the router in server environments that deal in request/response objects, e.g.
+//
+// ```
+// import { router } from "./router"
+// export default async function handler(req: Request) { return router.fetch(req) }
```
jimniels, line 575: A note on how to integrate with hosting envs like deno.dev, val.town, etc. — might be useful for n00bs like myself

packages/fetch-router/README.md
jimniels: Oh, but if you turn it off in the router, then doing like `methodOverride` won't work.

So I changed the examples in the docs to take formData as args.

packages/fetch-router/README.md
```diff
@@ -553,8 +552,7 @@ router.map(routes, {
         </html>
       `)
     },
-    async update({ params, request }) {
-      let formData = await request.formData()
```
jimniels, line 557: [This doesn't work because the router is reading it](https://github.com/remix-run/remix/pull/10778#discussion_r2430161993) (which I assume it needs to do in order to handle method overrides)

So need to get `formData` from the function rather than reading `request.formData`

mjackson:
Thanks for the PR, @jimniels. I landed the built-in formdata parsing the week before Remix Jam, so the docs were a little behind. Apologies for anything that was unclear. I'll give the docs a more comprehensive overhaul this week.

PR #10782, opened 2025-10-17
feat(multipart-parser): add ReadableStream support for multipart parts

Nyoxis:
This PR implements native `ReadableStream` support for multipart parts, aiming to align streaming with standard web streams and the Fetch API.

This is a **breaking change** intended to improve streaming efficiency and API consistency.

Please refer to the related discussion for design rationale, trade-offs, and ongoing feedback: #10781

Also aware of another multipart streaming PR implementing a callback-based chunking approach: #10764 

Feedback is welcome, especially regarding API design and backward compatibility.

### DIFF
--- a/packages/multipart-parser/src/lib/multipart.node.ts
+++ b/packages/multipart-parser/src/lib/multipart.node.ts
@@ -30,15 +30,21 @@ export class StreamedMultipartPart extends MultipartPart {
     this.contentReadable = Readable.fromWeb(webMultipartPart.content as NodeReadableStream<Uint8Array>)
     this.#webMultipartPart = webMultipartPart
   }
-
+  /**
+   * Consumes stream of content into buffered content,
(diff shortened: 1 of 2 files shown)


mjackson:
The current design of the parser is to buffer each part as it is read from the stream because buffers are easier to work with in many instances. Servers that rely on multipart-parser for parsing multipart Request bodies should have enough memory to buffer a single part of the request payload at a time. For the majority of servers and workloads, this should not be a problem. Even highly constrained worker environments like [Cloudflare Workers offer 128MB](https://developers.cloudflare.com/workers/platform/limits/#memory).

You can use [the `maxFileSize` option](https://github.com/remix-run/remix/tree/main/packages/multipart-parser#limiting-file-upload-size) to limit the maximum size of file uploads so you don't exhaust the memory on your server.

PR #10796, opened 2025-10-22
Bring over RR cookies code into @remix-run/cookie package

brophdawg11:
Adds a new `@remix-run/cookie` package - brought over from https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/server-runtime/cookies.ts

### DIFF
--- a/packages/interaction/README.md
+++ b/packages/interaction/README.md
@@ -73,7 +73,7 @@ on(listItem, {
 
 Import builtins from their modules (for example, `@remix-run/interaction/press`, `@remix-run/interaction/keys`). Some interactions may coordinate with others (for example, calling `event.preventDefault()` in one listener can prevent a related interaction from firing).
 
-You can also [create your own interactions](#interactions).
+You can also [create your own interactions](#custom-interactions).
 
 ### Async listeners and reentry protection
 
@@ -193,7 +193,7 @@ on(button, {
 })
 ```
 
-## Interactions
+## Custom Interactions
 
 Define semantic interactions that can dispatch custom events and be reused declaratively.
 


packages/cookie/src/lib/cookie.test.ts
```diff
@@ -0,0 +1,201 @@
+import * as assert from 'node:assert/strict'
```
brophdawg11, line 3: Tests converted from jest to node

packages/cookie/src/index.ts
```diff
@@ -0,0 +1 @@
+export { createCookie, isCookie } from './lib/cookie.ts'
```
brophdawg11: React Router also currently exports `type IsCookieFunction` as public API but I wasn't sure what exactly that was necessary for?  Happy to add that back in here if needed - I personally just tend to do `typeof isCookie` when I run into those cases.

packages/cookie/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@remix-run/cookie",
+  "version": "0.1.0",
+  "description": "A toolkit for working with cookies in JavaScript",
+  "author": "Michael Jackson <mjijackson@gmail.com>",
```
brophdawg11, line 5: I copied this from `headers` and altered accordingly.

We should probably swap these all over to `hello@remix.run` now that they're migrated into the Remix repo?

packages/cookie/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Michael Jackson
```
brophdawg11, line 3: Should this be updated across the board now that it's moved into the Remix repo?

packages/cookie/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "dependencies": {
+    "cookie": "^1.0.2"
+  },
+  "devDependencies": {
+    "@types/node": "^24.6.0",
+    "esbuild": "^0.25.10"
+  },
+  "scripts": {
+    "build": "pnpm run clean && pnpm run build:types && pnpm run build:esm",
+    "build:esm": "esbuild src/index.ts --bundle --external:cookie --outfile=dist/index.js --format=esm --platform=neutral --sourcemap",
```
brophdawg11: Single runtime dep on `cookie` which is marked external for the build

packages/cookie/src/index.ts
mjackson: Agree, we don't need `IsCookieFunction`.

packages/cookie/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# `@remix-run/cookie` CHANGELOG
+
+This is the changelog for [`@remix-run/cookie`](https://github.com/remix-run/remix/tree/main/packages/cookie). It follows [semantic versioning](https://semver.org/).
+
+## HEAD
```
mjackson:
```suggestion
## Unreleased

- This is the initial release of this package.
```

packages/cookie/LICENSE
mjackson: Yep, let's update these to Shopify

packages/cookie/package.json
mjackson: Yep, that works 👍

packages/cookie/src/lib/cookie.ts
```diff
@@ -0,0 +1,229 @@
...
+    },
+  }
+}
+
+type IsCookieFunction = (object: any) => object is Cookie
+
+/**
+ * Returns true if an object is a Remix cookie container.
+ */
+export const isCookie: IsCookieFunction = (object): object is Cookie => {
```
mjackson: Make this a regular `function` and ditch the `IsCookieFunction` type entirely.

packages/cookie/src/lib/cookie.ts
```diff
@@ -0,0 +1,229 @@
...
+   * Serializes the given value to a string and returns the `Set-Cookie`
+   * header.
+   */
+  serialize(value: any, options?: SerializeOptions): Promise<string>
+}
+
+/**
+ * Creates a logical container for managing a browser cookie from the server.
+ */
+export const createCookie = (name: string, cookieOptions: CookieOptions = {}): Cookie => {
```
mjackson: Make this a regular `function`.

packages/cookie/src/lib/crypto.ts
```diff
@@ -0,0 +1,50 @@
+const encoder = /* @__PURE__ */ new TextEncoder()
+
+export const sign = async (value: string, secret: string): Promise<string> => {
```
mjackson: Make this a regular `function`.

packages/cookie/src/lib/crypto.ts
```diff
@@ -0,0 +1,50 @@
+const encoder = /* @__PURE__ */ new TextEncoder()
```
mjackson: This isn't a proper use of `/* @__PURE__ */`. Let's remove it.

packages/cookie/src/lib/crypto.ts
```diff
@@ -0,0 +1,50 @@
...
+export const sign = async (value: string, secret: string): Promise<string> => {
+  let data = encoder.encode(value)
+  let key = await createKey(secret, ['sign'])
+  let signature = await crypto.subtle.sign('HMAC', key, data)
+  let hash = btoa(String.fromCharCode(...new Uint8Array(signature))).replace(/=+$/, '')
+
+  return value + '.' + hash
+}
+
+export const unsign = async (cookie: string, secret: string): Promise<string | false> => {
```
mjackson: Make this a regular `function`.

packages/cookie/src/lib/crypto.ts
```diff
@@ -0,0 +1,50 @@
...
+    return valid ? value : false
+  } catch (error: unknown) {
+    // atob will throw a DOMException with name === 'InvalidCharacterError'
+    // if the signature contains a non-base64 character, which should just
+    // be treated as an invalid signature.
+    return false
+  }
+}
+
+const createKey = async (secret: string, usages: CryptoKey['usages']): Promise<CryptoKey> =>
```
mjackson: Make this a regular `function`.

packages/cookie/package.json
mjackson: Let's go ahead and bundle the `cookie` dep. I believe this means we can move `cookie` from `dependencies` to `devDependencies` since we only ever need it at build/test time, and not in production.

mjackson (changes requested):
Thanks for the PR!

packages/cookie/src/lib/cookie.ts
brophdawg11: oh good - I'm so glad you prefer this syntax too 😂

packages/cookie/src/lib/cookie.ts
```diff
@@ -0,0 +1,227 @@
...
+   * Parses a raw `Cookie` header and returns the value of this cookie or
+   * `null` if it's not present.
+   */
+  parse(cookieHeader: string | null, options?: ParseOptions): Promise<any>
+
+  /**
+   * Serializes the given value to a string and returns the `Set-Cookie`
+   * header.
+   */
+  serialize(value: any, options?: SerializeOptions): Promise<string>
```
MichaelDeBoey: I think we should use `unknown` instead of `any` for the reasons explained in https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any

```suggestion
  serialize(value: unknown, options?: SerializeOptions): Promise<string>
```

packages/cookie/src/lib/cookie.ts
```diff
@@ -0,0 +1,227 @@
...
+        ...serializeOptions,
+      })
+    },
+  }
+}
+
+/**
+ * Returns true if an object is a Remix cookie container.
+ */
+export function isCookie(object: any): object is Cookie {
```
MichaelDeBoey:
```suggestion
export function isCookie(object: unknown): object is Cookie {
```

packages/cookie/src/lib/cookie.ts
```diff
@@ -0,0 +1,227 @@
...
+  return (
+    object != null &&
+    typeof object.name === 'string' &&
+    typeof object.isSigned === 'boolean' &&
+    typeof object.parse === 'function' &&
+    typeof object.serialize === 'function'
+  )
+}
+
+async function encodeCookieValue(value: any, secrets: string[]): Promise<string> {
```
MichaelDeBoey:
```suggestion
async function encodeCookieValue(value: unknown, secrets: string[]): Promise<string> {
```

packages/cookie/src/lib/cookie.ts
```diff
@@ -0,0 +1,227 @@
...
+      }
+    }
+
+    return null
+  }
+
+  return decodeData(value)
+}
+
+function encodeData(value: any): string {
```
MichaelDeBoey:
```suggestion
function encodeData(value: unknown): string {
```

packages/cookie/src/lib/cookie.ts
```diff
@@ -0,0 +1,227 @@
...
+  return btoa(myUnescape(encodeURIComponent(JSON.stringify(value))))
+}
+
+function decodeData(value: string): any {
```
MichaelDeBoey:
```suggestion
function decodeData(value: string): unknown {
```

packages/cookie/README.md
```diff
@@ -0,0 +1,307 @@
...
+**Properties:**
+
+- `name: string` - The cookie name (readonly)
+- `isSigned: boolean` - Whether the cookie uses signing (readonly)
+- `expires?: Date` - Calculated expiration date (readonly)
+
+**Methods:**
+
+- `parse(cookieHeader: string | null, options?: ParseOptions): Promise<any>` - Parse cookie value from header
+- `serialize(value: any, options?: SerializeOptions): Promise<string>` - Serialize value to Set-Cookie header
```
MichaelDeBoey:
```suggestion
- `serialize(value: unknown, options?: SerializeOptions): Promise<string>` - Serialize value to Set-Cookie header
```

packages/cookie/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+### `isCookie(object)`
+
+Type guard to check if an object is a cookie container.
+
+**Parameters:**
+
+- `object: any` - Object to test
```
MichaelDeBoey:
```suggestion
- `object: unknown` - Object to test
```

MichaelDeBoey (changes requested):

brophdawg11:
@mjackson I updated this with the same thing we did in the session PR:

* `interface Cookie` + `function createCookie` are now collapsed into `class Cookie`
* `isCookie` removed in favor of `instanceof Cookie`

packages/cookie/src/lib/cookie.ts
```diff
@@ -0,0 +1,206 @@
...
+    } else {
+      return null
+    }
+  }
+
+  /**
+   * Serializes the given value to a string and returns the `Set-Cookie`
+   * header.
+   */
+  async serialize(value: unknown, serializeOptions?: SerializeOptions): Promise<string> {
```
mjackson: I have a feeling we should use `value: string` here instead of letting people use any JS value.

One piece of feedback we got in Remix v1 that stuck with me over the years is that anyone who wants to take a look at the value of a cookie needs a JSON parser, because we run everything through `JSON.stringify`. This makes it tricky (and slow) for anyone who is using a reverse proxy and wants to do some kind of routing based on its value. Even if they give us a string, e.g. `cookie.serialize('string value')`, we still encode it as JSON in `encodeData`, so they still need to parse the JSON at some point.

Forcing a `string` value here will also simplify our implementation and cut down on the number of tests we need for different JS value types.

Also, if someone wants to store arbitrary JS values in cookies, they can always do it themselves with `cookie.serialize(JSON.stringify(value))` and `JSON.parse(await cookie.parse(header))`. This will be safe since the return type of `cookie.parse()` will be `Promise<string>`.

packages/cookie/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+// Malformed cookie values return empty object or null
+await cookie.parse('test=invalid-base64@#$') // {}
+
+// Signed cookies with bad signatures return null
+let signedCookie = new Cookie('signed', { secrets: ['secret'] })
+await signedCookie.parse('signed=value.badsignature') // null
+```
+
+## API Reference
```
mjackson: We can leave this section out of the README. We will work on auto-generating API docs for everything here at some point soon.

packages/cookie/src/lib/cookie.ts
brophdawg11: awesome yeah that was something I meant to ask next time we synced up because folks have asked a few times to be able to customize the encoding, so it feels right to just get out of that game entirely.

brophdawg11: Just to confirm - you want to remove with `JSON.stringify`/`JSON.parse` _and_ `btoa`/`atob` right?  It's the latter base-64 encoding folks have mostly complained about IIRC because it makes it harder for them to debug cookie values

brophdawg11: nvm I think we still need the base64 for utf8 chars :/

mjackson: Yeah the purpose isn't to make this human readable. It's to simplify the process of reading cookies in a non-JS environment, like a reverse proxy. base64 decoding is trivial to do. JSON is much more difficult.

brophdawg11: Removed in https://github.com/remix-run/remix/pull/10796/commits/b8db25082a2b357d4149ae467b926dd6bf16666d

packages/cookie/src/index.ts
```diff
@@ -1 +1,6 @@
-export { Cookie, type CookieOptions } from './lib/cookie.ts'
+export {
+  Cookie,
+  type CookieOptions,
+  type CookieParseOptions,
+  type CookieSerializeOptions,
```
brophdawg11: `CookieParseOptions`/`CookieSerializeOptions` are re-exported out from here so the `@remix-run/session` package can get them without having a dep on `cookie-es`

packages/cookie/package.json
```diff
@@ -33,12 +33,12 @@
   },
   "devDependencies": {
     "@types/node": "^24.6.0",
-    "cookie": "^1.0.2",
+    "cookie-es": "^2.0.0",
```
brophdawg11: ESM friendly port of `cookie`: https://www.npmjs.com/package/cookie-es, avoids needing the `--main-fields=main` change I had to add below to get `cookie` working

packages/cookie/README.md
brophdawg11: https://github.com/remix-run/remix/pull/10796/commits/6f1d7fb41953e79a18d13f2eed1bcde76d3e8c10

PR #10797, opened 2025-10-22
Bring over RR session code into @remix-run/session package

brophdawg11:
Stacked on https://github.com/remix-run/remix/pull/10796

Adds a new `@remix-run/session` package - brought over from https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/server-runtime/sessions.ts

### DIFF
--- a/demos/bookstore/app/auth.test.ts
+++ b/demos/bookstore/app/auth.test.ts
@@ -2,11 +2,7 @@ import * as assert from 'node:assert/strict'
 import { describe, it } from 'node:test'
 
 import { router } from './router.ts'
-import {
-  getSessionCookie as sessionCookie,
-  assertContains,
-  getSessionCookie,
-} from '../test/helpers.ts'
+import { assertContains, getSessionCookie } from '../test/helpers.ts'
 
 describe('auth handlers', () => {
   it('POST /login with valid credentials sets session cookie and redirects', async () => {
--- a/demos/bookstore/app/cart.test.ts
+++ b/demos/bookstore/app/cart.test.ts
@@ -7,6 +7,7 @@ import {
   assertContains,
   loginAsCustomer,
   assertNotContains,
+  getSessionCookie,
 } from '../test/helpers.ts'
 
 describe('cart handlers', () => {
@@ -36,7 +37,7 @@ describe('cart handlers', () => {
     assertNotContains(html, 'Heavy Metal Guitar Riffs')
 
     // First, add item to cart to get a session
-    await router.fetch('http://localhost:3000/cart/api/add', {
+    response = await router.fetch('http://localhost:3000/cart/api/add', {
       method: 'POST',
       body: new URLSearchParams({
         bookId: '002',
@@ -47,6 +48,7 @@ describe('cart handlers', () => {
       },
       redirect: 'manual',
     })
+    sessionCookie = getSessionCookie(response)!
 
     // Now view cart with session
     request = requestWithSession('http://localhost:3000/cart', sessionCookie)
@@ -62,7 +64,7 @@ describe('cart handlers', () => {
     let sessionCookie = await loginAsCustomer(router)
 
     // Add first item
-    await router.fetch('http://localhost:3000/cart/api/add', {
+    let response = await router.fetch('http://localhost:3000/cart/api/add', {
       method: 'POST',
       body: new URLSearchParams({
         bookId: '001',
@@ -73,6 +75,7 @@ describe('cart handlers', () => {
       },
       redirect: 'manual',
     })
+    sessionCookie = getSessionCookie(response)!
 
     // Add second item with same session
     let addRequest2 = requestWithSession('http://localhost:3000/cart/api/add', sessionCookie, {
--- a/demos/bookstore/app/checkout.test.ts
+++ b/demos/bookstore/app/checkout.test.ts
@@ -2,7 +2,7 @@ import * as assert from 'node:assert/strict'
 import { describe, it } from 'node:test'
 
 import { router } from './router.ts'
-import { loginAsCustomer, requestWithSession } from '../test/helpers.ts'
+import { getSessionCookie, loginAsCustomer, requestWithSession } from '../test/helpers.ts'
 
 describe('checkout handlers', () => {
   it('GET /checkout redirects when not authenticated', async () => {
@@ -23,7 +23,8 @@ describe('checkout handlers', () => {
         slug: 'bbq',
       }),
     })
-    await router.fetch(addRequest)
+    let response = await router.fetch(addRequest)
+    sessionCookie = getSessionCookie(response)!
 
     // Submit checkout
     let checkoutRequest = requestWithSession('http://localhost:3000/checkout', sessionCookie, {


packages/session/src/lib/session.ts
```diff
@@ -0,0 +1,309 @@
...
+  return `__flash_${name}__`
+}
+
+/**
+ * Creates a new Session object.
+ *
+ * Note: This function is typically not invoked directly by application code.
+ * Instead, use a `SessionStorage` object's `getSession` method.
+ */
+export function createSession<Data = SessionData, FlashData = Data>(
```
mjackson: Seems like we could organize this code a little better by using a `Session` class. Then we wouldn't need a separate `interface` declaration up above. It would also make our `isSession` implementation simpler, or probably even redundant because people could just use `obj instanceof Session` instead.

We could still provide a `createSession` function for those who prefer doing things in a functional style, but internally it would just `return new Session(...)`. I actually have a similar situation in the router where I recommend using `createRouter` in all the docs, but the router itself is implemented as a class.

packages/fetch-router/src/lib/router.ts
```diff
@@ -92,9 +101,10 @@ export class Router {
       throw requestAbortError()
     }
 
-    let response = await this.dispatch(request)
+    let context = await this.#parseRequest(request)
```
mjackson: Good call here. The default handler should also have access to the fully parsed request context.

packages/fetch-router/src/lib/router.ts
```diff
@@ -160,7 +176,16 @@ export class Router {
...
-    let context = new RequestContext(request)
+    // We have to create the session here because it's an async operation to
+    // parse the cookie internally using `cookie.parse()`.
+    // - We can't use a `get session()` getter to lazily create the session because
+    //   `getSession` is async
+    // - We can't create the session in the `RequestContext` constructor because
+    //   constructors can't be async
+    // - If we assign the session in the middleware, then `context.session` has
+    //   to have a type of `Session | undefined` which is inconvenient for users
+    let session = await this.#sessionStorage.getSession(request.headers.get('Cookie'))
```
mjackson: I see two options here:

a) Always use a root `session` middleware (as you're already doing in `dispatch`). In that case, we can get this code out of here and set `context.session` in the middleware. Then call `next()` to pass the request downstream and set the `Set-Cookie` header on the way back up.

b) Inline the logic from the session middleware in `dispatch`. This is essentially the approach we are already taking with the form data parsing and method override, both of which could be built as separate middlewares that always run.

The main difference is form data and method override need to affect matching, so they need to run before we know the full request context. The session doesn't affect route matching, but it is part of the request context so it feels similar to the other 2.

I think what I'd like to try is doing the session decoding here in this method, just below the creation of the `context` object. Keep the `RequestContext` constructor clean with a single `request` arg, and manually set `context._session` similar to how we're currently setting `context.formData` and `context.method` below. Then use a getter in `context.session` that returns `this._session ??= new Session()` to automatically create a new session when it is accessed if we didn't find one in the cookie.

Then, just before `dispatch` returns its `response`, it could automatically set the `Set-Cookie` header.

Ofc, none of this runs if they don't have a `sessionStorage` set. In that case, they'll just get a blank, new `contex.session` that won't do anything. This is similar to PHP where the `$_SESSION` superglobal is blank and doesn't do anything useful unless you call `session_start()` somewhere before you access it.

packages/fetch-router/src/lib/router.ts
```diff
@@ -71,13 +78,15 @@ export class Router {
...
   #parseFormData: (ParseFormDataOptions & { suppressErrors?: boolean }) | boolean
+  #sessionStorage: SessionStorage
   #uploadHandler: FileUploadHandler | undefined
   #methodOverride: string | boolean
 
   constructor(options?: RouterOptions) {
     this.#defaultHandler = options?.defaultHandler ?? noMatchHandler
     this.#matcher = options?.matcher ?? new RegExpMatcher()
     this.#parseFormData = options?.parseFormData ?? true
+    this.#sessionStorage = options?.sessionStorage ?? createCookieSessionStorage()
```
mjackson: This should default to `undefined` meaning "I don't want to use sessions". We don't want to default to sending session cookies on every site, especially not in a GDPR world.

packages/fetch-router/package.json
```diff
@@ -41,14 +41,15 @@
     "@types/node": "^24.6.0",
     "esbuild": "^0.25.10"
   },
-  "dependencies": {
+  "peerDependencies": {
```
mjackson: I will take care of this on main so you don't need to do it in this PR. But thanks for bringing this to my attention.

mjackson (changes requested):
Overall we are headed in the right direction, thanks for the effort here this week! I hope the feedback helps.

packages/session/src/lib/session.ts
```diff
@@ -0,0 +1,309 @@
...
+import type { Cookie, CookieOptions } from '@remix-run/cookie'
+import { createCookie, isCookie } from '@remix-run/cookie'
+
+import { warnOnce } from './warnings.ts'
+
+/**
+ * An object of name/value pairs to be used in the session.
+ */
+export interface SessionData {
+  [name: string]: any
```
MichaelDeBoey: I think we should use `unknown` instead of `any` for the reasons explained in https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any


```suggestion
  [name: string]: unknown
```

packages/session/src/lib/session.ts
```diff
@@ -0,0 +1,309 @@
...
+      map.clear()
+      status = 'destroyed'
+    },
+  }
+}
+
+/**
+ * Returns true if an object is a Remix session.
+ */
+export function isSession(object: any): object is Session {
```
MichaelDeBoey:
```suggestion
export function isSession(object: unknown): object is Session {
```

packages/fetch-router/src/lib/request-context.test.ts
```diff
@@ -9,11 +10,14 @@ describe('new RequestContext()', () => {
         'Content-Type': 'application/json',
       },
     })
-    let context = new RequestContext(req)
+    let context = new RequestContext(req, createSession())
```
mjackson: We don't need this line.

packages/fetch-router/src/lib/router.ts
brophdawg11: Discussed offline we should be able to make it work such that a default session is available but never comitted to a cookie until they interact with it

packages/session/src/lib/session.ts
brophdawg11: 👍 yeah I didn't touch much of what we had in Remix v1 but I agree classes make sense to reduce a bit of the duplication of a separate interface

packages/session/src/file-storage.ts
```diff
@@ -0,0 +1 @@
+export { createFileSessionStorage } from './lib/storage/file-storage.ts'
```
brophdawg11, line 1: I needed to move this to a sub-export so that it didn't get picked up when a client-side API imported `@remix-run/fetch-router` and `@remix-run/fetch-router` importing the agnostic session stuff from `@remix-run/sesson`.

This happened from the bookstore `routes.ts`:

```ts
import { route, formAction, resources } from '@remix-run/fetch-router'

export let routes = route({ ... })
```

When we tried to run `pnpm dev:browser`, it would error on the `file-storage` imports of `node:fs`

demos/bookstore/app/models/cart.ts
```diff
@@ -10,27 +12,44 @@ export interface Cart {
   items: CartItem[]
 }
 
-// Store carts by session ID
+// Store carts by cartId, cartId will be stored in the session
+let nextCartId = 1
```
brophdawg11, line 16: No longer storing by session id because that's not something app code can always _directly_ access (`sesson.id` isn't populated on initial session creation - only when we eventually commit the session).  We don't want to commit a session until they interact with it, so there's no good way to key a cart by sessionId the first time we interact with the session.  Instead, we use a cartId that we store in the session (which "activates" the session and triggers the cookie), and we key the cart lookup by that.

demos/bookstore/app/middleware/cart.ts
```diff
@@ -0,0 +1,10 @@
+import type { Middleware } from '@remix-run/fetch-router'
+import { createCartIfNotExists } from '../models/cart.ts'
+
+/**
+ * Middleware that ensures the user session has an associated cart
+ * To be used on any route that mutates the cart
+ */
+export const ensureCart: Middleware = async ({ session }) => {
+  createCartIfNotExists(session)
```
brophdawg11, line 9: Middleware used by the cart API routes to ensure we always have a cart available if we're trying to mutate the cart

demos/bookstore/app/utils/session.ts
```diff
@@ -1,77 +1,21 @@
...
-  let sessionId = getSessionId(request)
-  let session = sessions.get(sessionId)
-
-  if (!session) {
-    session = { sessionId }
-    sessions.set(sessionId, session)
+declare module '@remix-run/session' {
+  interface SessionData {
+    cartId?: string
+    userId?: string
```
brophdawg11, line 7: Module augmentation to strongly type our session

demos/bookstore/app/cart.tsx
```diff
@@ -137,64 +137,55 @@ export default {
     },
 
     api: {
-      async add({ storage, formData }) {
-        // Simulate network latency
-        await new Promise((resolve) => setTimeout(resolve, 1000))
+      use: [ensureCart],
```
brophdawg11, line 140: Cart middleware that runs before all cart API handlers

demos/bookstore/package.json
```diff
@@ -12,6 +12,7 @@
     "@remix-run/node-fetch-server": "workspace:*"
   },
   "devDependencies": {
+    "@remix-run/session": "workspace:*",
```
mjackson, line 15: Should be a regular dep, not a dev dep.

packages/fetch-router/package.json
```diff
@@ -47,15 +47,16 @@
     "esbuild": "^0.25.10",
     "tsx": "^4.20.6"
   },
-  "peerDependencies": {
+  "dependencies": {
```
mjackson, line 50: These should all be peer deps, not regular deps. The strategy with individual packages is that they should all use peer deps with other `@remix-run/*` packages and use `--external` when building.

packages/fetch-router/src/lib/router.test.ts
```diff
@@ -1670,3 +1672,609 @@ describe('abort signal support', () => {
     assert.equal(handlerCalled, false)
   })
 })
+
```
mjackson, line 1675: This file is getting huge. Should we put these in a new `sessions.test.ts` file? I could probably separate out a few others too, like `file-uploads.test.ts` and `middleware.test.ts`

packages/session/package.json
```diff
@@ -0,0 +1,65 @@
...
+      },
+      "./file-storage": {
+        "types": "./dist/file-storage.d.ts",
+        "default": "./dist/file-storage.js"
+      },
+      "./package.json": "./package.json"
+    }
+  },
+  "dependencies": {
+    "@remix-run/cookie": "workspace:^",
```
mjackson, line 40: I've been using `workspace:*`. This could also work. I just want to be consistent about how we do it.

packages/session/src/index.ts
```diff
@@ -0,0 +1,11 @@
+export {
+  type SessionData,
+  type SessionIdStorageStrategy,
+  type SessionStorage,
+  type FlashSessionData,
+  createSessionStorage,
+  Session,
+} from './lib/session.ts'
+
+export { createCookieSessionStorage } from './lib/storage/cookie-storage.ts'
```
mjackson, line 10: The `file-storage` package uses a pattern where we have separate exports for all implementations, e.g. `@remix-run/file-storage/local` and `@remix-run/file-storage/memory`.

I think the same pattern makes sense here. Let's put both cookie and memory storage into their own exports. So we'll have:

- `@remix-run/session/cookie-storage`
- `@remix-run/session/file-storage`
- `@remix-run/session/memory-storage`

packages/session/src/lib/session.ts
```diff
@@ -0,0 +1,283 @@
...
+/**
+ * Session persists data across HTTP requests.
+ *
+ * Note: This class is typically not invoked directly by application code.
+ * Instead, use a `SessionStorage` object's `getSession` method.
+ */
+export class Session<Data = SessionData, FlashData = Data> {
+  #id: string
+  #map: Map<keyof Data | FlashDataKey<keyof FlashData & string>, unknown>
+  #status: 'new' | 'clean' | 'dirty' | 'destroyed'
```
mjackson, line 23: The `new` status feels redundant to me. `clean`, `dirty`, and `destroyed` make sense. But `new` feels the same as `clean` to me in the sense that in neither case does it need to be saved.

packages/session/src/lib/session.ts
```diff
@@ -0,0 +1,283 @@
+import type { ParseOptions, SerializeOptions } from 'cookie'
+import type { CookieOptions } from '@remix-run/cookie'
+import { Cookie } from '@remix-run/cookie'
+
+import { warnOnce } from './warnings.ts'
+
+/**
+ * An object of name/value pairs to be used in the session.
+ */
+export interface SessionData {
```
mjackson, line 10: This feels like an awkward way to achieve type safety for `session.data` because there's no way to define different data in different session cookies. We should either come up with a solution that works for N session cookies on a site or just return `unknown` values out of `session.data` for now.

demos/bookstore/app/middleware/cart.ts
mjackson: That's a great pattern.

packages/fetch-router/src/lib/router.ts
```diff
@@ -218,6 +262,38 @@ export class Router {
...
+      //   read in from a cookie, and when that happens we don't need to send up
+      //   a new cookie because we already have the ID in there
+      if (session.id === '') {
+        response.headers.append('Set-Cookie', cookie)
+      }
+    } else if (session.status === 'destroyed') {
+      let cookie = await this.#sessionStorage.destroySession(session)
+      response.headers.append('Set-Cookie', cookie)
+    }
+
```
gyx1000, line 292: What do you think about implementing an optional touchSession feature on storage?
In cases where sessions are automatically cleaned up by storage upon expiration (cron, ttl), this would allow the idle timer to be reset without updating the entire content.

The touchSession would only be called if the session status is clean and the storage implements touchSession.
The custom storage backend could also manage a touchTTL to avoid updating the backend each time a session is read.

Sorry if you have already ruled out this strategy and I missed the information. If so, what approach would you take?

brophdawg11:
Ended up starting a fresh PR to clean up some of this commit history after a few rebases onto the breaking changes in `main`.  Closing this in favor of #10811

packages/fetch-router/package.json
brophdawg11: huh yeah I thought I had done that but maybe a bad rebase flipped it back 👍

packages/fetch-router/src/lib/router.test.ts
brophdawg11: yeah I'll pull session tests out for now and we can pull the others out in a separate PR

packages/session/package.json
brophdawg11: Looks like we have some of each throughout the repo - I'll use `workspace:*` for this new stuff and we can do a larger pass separately

packages/session/src/index.ts
brophdawg11: good call - I updated to follow the same pattern here.  I think in RR we had a loose rule of only using a separate package when the build/runtime required it (i.e., `react-router/dom`) but I like the feel of these self contained sub-exports for semi-mutually-exclusive behaviors.

packages/session/src/lib/session.ts
brophdawg11: yeah this is carried forward from Remix v1 - happy to ditch it since it's just sort of a TS lie anyway

packages/session/src/lib/session.ts
brophdawg11: yeah the new only commit session on modifications alleviated the original use for `new` - removed

packages/fetch-router/src/lib/router.ts
brophdawg11: Lets land the session stuff in it's current state (largely following the implementation in RR) and then we can consider potential enhancements.  Feel free to open a discussion outlining the use case and we can further discuss if something like this would be beneficial there 👍

gyx1000: Yes! Thank you for the work already done! 👌

PR #10798, opened 2025-10-23
Add fetch-router package to README

brookslybrand:
@mjackson I would commit this directly but I just wanted to make sure there wasn't a reason `fetch-router` isn't listed in the main `README` first


mjackson:
LGTM. Thanks, @brookslybrand !

PR #10802, opened 2025-10-26
feat(fetch-router): make `@remix-run/html-template` an optional peerDep

MichaelDeBoey:
This way people who aren't using the `html` response helper, don't need to install this package


mjackson:
I'd like to leave it as non-optional while response helpers still live in the router package. They may move out at some point in the future, but I'm not sure yet.

PR #10805, opened 2025-10-29
proposal: named type parameters

frolic:
I've spent a lot of time over the last couple years deep in strongly typed libraries with the legendary [David Blass](https://x.com/ssalbdivad). One pattern of his that I've adopted is naming type parameters of generics like you would variable names (because that's effectively what they are). It felt weird/ugly to me at first, but after using it for a couple years, I find it much more readable, especially as types get more complex (e.g. >2 parameters in play) and while doing type transformations like mapped types.

I've mocked this up in a couple of files to demonstrate both the pure types side as well as how the type parameters are used within methods and how they nicely align with parameter names.

If ya'll are open to this, I'm happy to finish converting the rest of this package here, and will open up separate PRs for other packages that need this.

### DIFF
--- a/packages/fetch-router/src/lib/route-handlers.ts
+++ b/packages/fetch-router/src/lib/route-handlers.ts
@@ -6,24 +6,24 @@ import type { RequestMethod } from './request-methods.ts'
 import type { Route, RouteMap } from './route-map.ts'
 
 // prettier-ignore
-export type RouteHandlers<T extends RouteMap> =
-  | RouteHandlersWithMiddleware<T>
+export type RouteHandlers<routeMap extends RouteMap> =
+  | RouteHandlersWithMiddleware<routeMap>
   | {
-      [K in keyof T]: (
-        T[K] extends Route<infer M, infer P> ? RouteHandler<M, P> :
-        T[K] extends RouteMap ? RouteHandlers<T[K]> :
+      [name in keyof routeMap]: (
+        routeMap[name] extends Route<infer method, infer pattern> ? RouteHandler<method, pattern> :
+        routeMap[name] extends RouteMap ? RouteHandlers<routeMap[name]> :
         never
       )
     }
 
-type RouteHandlersWithMiddleware<T extends RouteMap> = {
+type RouteHandlersWithMiddleware<routeMap extends RouteMap> = {
   use: Middleware[]
-  handlers: RouteHandlers<T>
+  handlers: RouteHandlers<routeMap>
 }
 
-export function isRouteHandlersWithMiddleware<T extends RouteMap>(
-  handlers: any,
-): handlers is RouteHandlersWithMiddleware<T> {
+export function isRouteHandlersWithMiddleware<routeMap extends RouteMap>(
+  handlers: unknown,
+): handlers is RouteHandlersWithMiddleware<routeMap> {
   return (
     typeof handlers === 'object' && handlers != null && 'use' in handlers && 'handlers' in handlers
   )
@@ -32,18 +32,19 @@ export function isRouteHandlersWithMiddleware<T extends RouteMap>(
 /**
  * An individual route handler.
  */
-export type RouteHandler<M extends RequestMethod | 'ANY', P extends string> =
-  | RequestHandlerWithMiddleware<M, P>
-  | RequestHandler<M, Params<P>>
+export type RouteHandler<method extends RequestMethod | 'ANY', pattern extends string> =
+  | RequestHandlerWithMiddleware<method, pattern>
+  | RequestHandler<method, Params<pattern>>
 
-type RequestHandlerWithMiddleware<M extends RequestMethod | 'ANY', P extends string> = {
-  use: Middleware<M, Params<P>>[]
-  handler: RequestHandler<M, Params<P>>
+type RequestHandlerWithMiddleware<method extends RequestMethod | 'ANY', pattern extends string> = {
+  use: Middleware<method, Params<pattern>>[]
+  handler: RequestHandler<method, Params<pattern>>
 }
 
-export function isRequestHandlerWithMiddleware<M extends RequestMethod | 'ANY', P extends string>(
-  handler: any,
-): handler is RequestHandlerWithMiddleware<M, P> {
+export function isRequestHandlerWithMiddleware<
+  method extends RequestMethod | 'ANY',
+  pattern extends string,
+>(handler: unknown): handler is RequestHandlerWithMiddleware<method, pattern> {
   return typeof handler === 'object' && handler != null && 'use' in handler && 'handler' in handler
 }
 
@@ -51,8 +52,8 @@ export function isRequestHandlerWithMiddleware<M extends RequestMethod | 'ANY', 
  * Build a `RouteHandler` type from a string, route pattern, or route.
  */
 // prettier-ignore
-export type BuildRouteHandler<M extends RequestMethod | 'ANY', T extends string | RoutePattern | Route> =
-  T extends string ? RouteHandler<M, T> :
-  T extends RoutePattern<infer P> ? RouteHandler<M, P> :
-  T extends Route<infer _, infer P> ? RouteHandler<M, P> :
+export type BuildRouteHandler<method extends RequestMethod | 'ANY', route extends string | RoutePattern | Route> =
+  route extends string ? RouteHandler<method, route> :
+  route extends RoutePattern<infer pattern> ? RouteHandler<method, pattern> :
+  route extends Route<infer _, infer pattern> ? RouteHandler<method, pattern> :
   never
--- a/packages/fetch-router/src/lib/route-map.ts
+++ b/packages/fetch-router/src/lib/route-map.ts
@@ -4,27 +4,27 @@ import type { HrefBuilderArgs, Join, RouteMatch } from '@remix-run/route-pattern
 import type { RequestMethod } from './request-methods.ts'
 import type { Simplify } from './type-utils.ts'
 
-export interface RouteMap<T extends string = string> {
-  [K: string]: Route<RequestMethod | 'ANY', T> | RouteMap<T>
+export interface RouteMap<name extends string = string> {
+  [name: string]: Route<RequestMethod | 'ANY', name> | RouteMap<name>
 }
 
 export class Route<
-  M extends RequestMethod | 'ANY' = RequestMethod | 'ANY',
-  P extends string = string,
+  method extends RequestMethod | 'ANY' = RequestMethod | 'ANY',
+  pattern extends string = string,
 > {
-  readonly method: M | 'ANY'
-  readonly pattern: RoutePattern<P>
+  readonly method: method | 'ANY'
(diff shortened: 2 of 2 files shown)


packages/fetch-router/src/lib/route-map.ts
```diff
@@ -4,39 +4,39 @@ import type { HrefBuilderArgs, Join, RouteMatch } from '@remix-run/route-pattern
 import type { RequestMethod } from './request-methods.ts'
 import type { Simplify } from './type-utils.ts'
 
-export interface RouteMap<T extends string = string> {
-  [K: string]: Route<RequestMethod | 'ANY', T> | RouteMap<T>
+export interface RouteMap<name extends string = string> {
+  [name: string]: Route<RequestMethod | 'ANY', name> | RouteMap<name>
```
mjackson, line 8: Shouldn't this line be something like:

```ts
  [key: string]: Route<RequestMethod | 'ANY', name> | RouteMap<name>
```

using `key` instead of `name` for the index variable?

Naming the index variable the same as the generic is a little confusing.

packages/fetch-router/src/lib/route-handlers.ts
```diff
@@ -6,24 +6,24 @@ import type { RequestMethod } from './request-methods.ts'
...
   | {
-      [K in keyof T]: (
-        T[K] extends Route<infer M, infer P> ? RouteHandler<M, P> :
-        T[K] extends RouteMap ? RouteHandlers<T[K]> :
+      [name in keyof routeMap]: (
+        routeMap[name] extends Route<infer method, infer pattern> ? RouteHandler<method, pattern> :
+        routeMap[name] extends RouteMap ? RouteHandlers<routeMap[name]> :
         never
       )
     }
```
MichaelDeBoey, line 17:
```suggestion
export type RouteHandlers<TRouteMap extends RouteMap> =
  | RouteHandlersWithMiddleware<TRouteMap>
  | {
    [Key in keyof TRouteMap]: (
      TRouteMap[Key] extends Route<infer Method, infer Pattern> ? RouteHandler<Method, Pattern> :
      TRouteMap[Key] extends RouteMap ? RouteHandlers<TRouteMap[Key]> :
      never
  )
}
```

packages/fetch-router/src/lib/route-handlers.ts
```diff
@@ -6,24 +6,24 @@ import type { RequestMethod } from './request-methods.ts'
...
 
-type RouteHandlersWithMiddleware<T extends RouteMap> = {
+type RouteHandlersWithMiddleware<routeMap extends RouteMap> = {
   use: Middleware[]
-  handlers: RouteHandlers<T>
+  handlers: RouteHandlers<routeMap>
 }
```
MichaelDeBoey, line 22:
```suggestion
type RouteHandlersWithMiddleware<TRouteMap extends RouteMap> = {
  use: Middleware[]
  handlers: RouteHandlers<TRouteMap>
}
```

packages/fetch-router/src/lib/route-handlers.ts
```diff
@@ -6,24 +6,24 @@ import type { RequestMethod } from './request-methods.ts'
...
 
-export function isRouteHandlersWithMiddleware<T extends RouteMap>(
-  handlers: any,
-): handlers is RouteHandlersWithMiddleware<T> {
+export function isRouteHandlersWithMiddleware<routeMap extends RouteMap>(
+  handlers: unknown,
+): handlers is RouteHandlersWithMiddleware<routeMap> {
```
MichaelDeBoey, line 26:
```suggestion
export function isRouteHandlersWithMiddleware<TRouteMap extends RouteMap>(
  handlers: unknown,
): handlers is RouteHandlersWithMiddleware<TRouteMap> {
```

packages/fetch-router/src/lib/route-handlers.ts
```diff
@@ -32,27 +32,28 @@ export function isRouteHandlersWithMiddleware<T extends RouteMap>(
 /**
  * An individual route handler.
  */
-export type RouteHandler<M extends RequestMethod | 'ANY', P extends string> =
-  | RequestHandlerWithMiddleware<M, P>
-  | RequestHandler<M, Params<P>>
+export type RouteHandler<method extends RequestMethod | 'ANY', pattern extends string> =
+  | RequestHandlerWithMiddleware<method, pattern>
+  | RequestHandler<method, Params<pattern>>
```
MichaelDeBoey, line 37:
```suggestion
export type RouteHandler<Method extends RequestMethod | 'ANY', Pattern extends string> =
  | RequestHandlerWithMiddleware<Method, Pattern>
  | RequestHandler<Method, Params<Pattern>>
```

packages/fetch-router/src/lib/route-handlers.ts
```diff
@@ -32,27 +32,28 @@ export function isRouteHandlersWithMiddleware<T extends RouteMap>(
...
 
-type RequestHandlerWithMiddleware<M extends RequestMethod | 'ANY', P extends string> = {
-  use: Middleware<M, Params<P>>[]
-  handler: RequestHandler<M, Params<P>>
+type RequestHandlerWithMiddleware<method extends RequestMethod | 'ANY', pattern extends string> = {
+  use: Middleware<method, Params<pattern>>[]
+  handler: RequestHandler<method, Params<pattern>>
 }
```
MichaelDeBoey, line 42:
```suggestion
type RequestHandlerWithMiddleware<Method extends RequestMethod | 'ANY', Pattern extends string> = {
  use: Middleware<Method, Params<Pattern>>[]
  handler: RequestHandler<Method, Params<Pattern>>
}
```

packages/fetch-router/src/lib/route-handlers.ts
```diff
@@ -32,27 +32,28 @@ export function isRouteHandlersWithMiddleware<T extends RouteMap>(
...
 
-export function isRequestHandlerWithMiddleware<M extends RequestMethod | 'ANY', P extends string>(
-  handler: any,
-): handler is RequestHandlerWithMiddleware<M, P> {
+export function isRequestHandlerWithMiddleware<
+  method extends RequestMethod | 'ANY',
+  pattern extends string,
+>(handler: unknown): handler is RequestHandlerWithMiddleware<method, pattern> {
```
MichaelDeBoey, line 47:
```suggestion
export function isRequestHandlerWithMiddleware<
  Method extends RequestMethod | 'ANY',
  Pattern extends string,
>(handler: unknown): handler is RequestHandlerWithMiddleware<Method, Pattern> {
```

packages/fetch-router/src/lib/route-handlers.ts
```diff
@@ -32,27 +32,28 @@ export function isRouteHandlersWithMiddleware<T extends RouteMap>(
...
 // prettier-ignore
-export type BuildRouteHandler<M extends RequestMethod | 'ANY', T extends string | RoutePattern | Route> =
-  T extends string ? RouteHandler<M, T> :
-  T extends RoutePattern<infer P> ? RouteHandler<M, P> :
-  T extends Route<infer _, infer P> ? RouteHandler<M, P> :
+export type BuildRouteHandler<method extends RequestMethod | 'ANY', route extends string | RoutePattern | Route> =
+  route extends string ? RouteHandler<method, route> :
+  route extends RoutePattern<infer pattern> ? RouteHandler<method, pattern> :
+  route extends Route<infer _, infer pattern> ? RouteHandler<method, pattern> :
   never
```
MichaelDeBoey, line 59:
```suggestion
export type BuildRouteHandler<Method extends RequestMethod | 'ANY', TRoute extends string | RoutePattern | Route> =
  TRoute extends string ? RouteHandler<Method, TRoute> :
  TRoute extends RoutePattern<infer Pattern> ? RouteHandler<Method, Pattern> :
  TRoute extends Route<infer _, infer Pattern> ? RouteHandler<Method, Pattern> :
  never
```

packages/fetch-router/src/lib/route-map.ts
frolic: Oops, you're totally right, this was an overshoot on my part. `key` is totally fine here.

I think when I was going through this code, I was reading this as a mapped type rather than a map definition. In a mapped typed, I prefer to use something more descriptive than `key` because you end up using it as a variable.

packages/fetch-router/src/lib/route-handlers.ts
mjackson: I don't like this as much as `routeMap`. `TRouteMap` is an awkward name.

The only reason people put a `T` in front of it is so that the name of the generic doesn't match the name of the interface it `extends` in the constraint. But if we follow the practice that all our interfaces/types are declared with uppercase names, and all our generics are declared with lowercase names, we will never have this problem.

frolic:
Oops, just a point of clarity: I meant _type parameter_ names (the things that go into generics), not the names of the _generics_ themselves.

I generally follow the [ArkType's guidelines](https://github.com/arktypeio/arktype/blob/main/.github/CONTRIBUTING.md):

> We also have some unique casing rules for our TypeScript types to facilitate type-level code that can parallel its runtime implementation and be easily understood:
> 
> - Use PascalCase for...
>   - Entities/non-generic types (e.g. User, SomeData)
>   - Generic types with noun names, like Array<t>. As a rule of thumb, if your generic is named this way, all parameters have defaults.
> - Use camelCase for...
>   - Generic types with verb names like inferDomain<t>. Types named this way have at least one required parameter.
>   - Parameter names, e.g. t in Array<t>

The important bit for me is the parameter names being lowercase. And fortunately this has no downstream impacts in terms of breaking APIs, it's all internal, so things can be progressively migrated.

packages/fetch-router/src/lib/route-handlers.ts
frolic: This is an example where it would be quite easy to accidentally use `RequestMethod` instead of `Method` in the return type (i.e. `handler is RequestHandlerWithMiddleware<RequestMethod, Pattern>`), whereas the lowercase parameter names makes it much clearer what should be passed into the generic.

mjackson:
Hi @frolic, I went ahead and got a start on this in https://github.com/remix-run/remix/commit/bd43a6e5fd6f379fe26f3961c4f738c6e892c5b8, just to see how it feels. I like how it reads and will likely be doing more of this in this repo from here on out, thanks for the suggestion!

I'm going to leave this open until we have done this across all packages.

frolic:
Looks great! Already much more readable 🎉 

I'm still happy to contribute some effort here. Are there any packages you think would good for me to tackle? Maybe some of the smaller ones?

mjackson:
> Are there any packages you think would good for me to tackle? Maybe some of the smaller ones?

The next package I'd take a look at is route-pattern. It includes a parser for route patterns written in TypeScript that I think could benefit from some more readable generic names.

mjackson:
We are doing this in most places now, so I'm going to close. Thanks again for the suggestion, @frolic. I really like how easy it makes reading some of our more complex types 👍

PR #10808, opened 2025-10-31
feat(fetch-router): add static-middleware

markdalgleish:
See docs in PR for details.

Note, this also adds the following lower level primitives:
- `file` response helper for manually creating `Response` objects containing files.
- `Range`, `Content-Range`, `If-Match`, `If-Range` and `Allow` support to the `headers` package.

### DIFF
--- a/packages/lazy-file/src/fs.test.ts
+++ b/packages/lazy-file/src/fs.test.ts
@@ -189,6 +189,19 @@ describe('findFile', () => {
     teardown()
   })
 
+  it('allows overriding file.name with options.name', async () => {
+    setup()
+    createTestFile('assets/favicon.ico', 'icon content')
+
+    let result = await findFile(tmpDir, 'assets/favicon.ico', { name: 'custom.ico' })
+
+    assert.ok(result)
+    assert.equal(result.name, 'custom.ico')
+    assert.equal(await result.text(), 'icon content')
+
+    teardown()
+  })
+
   it('throws non-ENOENT errors', async () => {
     setup()
 
--- a/packages/lazy-file/src/fs.ts
+++ b/packages/lazy-file/src/fs.ts
@@ -71,6 +71,13 @@ function streamFile(
   })
 }
 
+export interface FindFileOptions {
+  /**
+   * Overrides the name of the file. Default is the relativePath argument.
+   */
+  name?: string
+}
+
 /**
  * Finds a file on the filesystem within the given root directory.
  *
@@ -78,19 +85,31 @@ function streamFile(
  * specified root directory.
  *
  * The returned file's `name` property will be set to the `relativePath` argument,
- * preserving the path structure relative to the root.
+ * unless overridden via `options.name`.
  *
  * @param root - The root directory to serve files from (absolute or relative to cwd)
  * @param relativePath - The relative path from the root to the file
+ * @param options - Options to override the file's metadata
  * @returns A `File` object, or null if not found
  *
  * @example
- * let file = await findFile('./public', 'assets/favicon.ico')
+ * let file = await findFile('./public', 'assets/logo.png')
  * if (file) {
- *   console.log(file.name) // "assets/favicon.ico"
+ *   console.log(file.name) // "assets/logo.png"
+ * }
+ *
+ * @example
+ * // Override the file name
+ * let file = await findFile('./public', 'assets/logo.png', { name: 'custom.png' })
+ * if (file) {
+ *   console.log(file.name) // "custom.png"
  * }
  */
-export async function findFile(root: string, relativePath: string): Promise<File | null> {
+export async function findFile(
+  root: string,
+  relativePath: string,
+  options?: FindFileOptions,
+): Promise<File | null> {
   // Ensure root is an absolute path
   root = path.resolve(root)
 
@@ -102,7 +121,7 @@ export async function findFile(root: string, relativePath: string): Promise<File
   }
 
   try {
-    return openFile(filePath, { name: relativePath })
+    return openFile(filePath, { name: options?.name ?? relativePath })
   } catch (error) {
     if (isNoEntityError(error) || isNotAFileError(error)) {
       return null


packages/fetch-router/src/lib/middleware/static.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Resolver that extracts the pathname from the request context, without a
+ * leading slash so it's suitable for use as a relative file path.
+ *
+ * @example
+ * // Request to http://example.com/assets/style.css
+ * // Returns: "assets/style.css"
+ */
+function requestPathnameResolver(context: RequestContext): string {
+  return new URL(context.request.url).pathname.replace(/^\/+/, '')
```
mjackson: The request context already contains a parsed URL, so you can just do it like this instead:


```suggestion
  return context.url.pathname.replace(/^\/+/, '')
```

packages/headers/src/index.ts
```diff
@@ -5,7 +5,9 @@ export { type CacheControlInit, CacheControl } from './lib/cache-control.ts'
 export { type ContentDispositionInit, ContentDisposition } from './lib/content-disposition.ts'
 export { type ContentTypeInit, ContentType } from './lib/content-type.ts'
```
mjackson, line 7: Missing `ContentRange` export above this line.

packages/fetch-router/src/lib/file-handler.test.ts
```diff
@@ -0,0 +1,1206 @@
...
+    } = {},
+  ): RequestContext<'GET', {}> {
+    let headers = new SuperHeaders(options.headers ?? {})
+    return {
+      formData: undefined,
+      storage: new AppStorage(),
+      url: new URL(url),
+      files: null,
+      method: 'GET',
+      request: new Request(url, {
```
mjackson: I think you can just `return new RequestContext(request)` here instead of manually mocking out all properties of a request context object.

packages/fetch-router/src/lib/middleware/static.ts
```diff
@@ -0,0 +1,73 @@
...
+>(root: string, options: StaticFilesOptions<Method, Params> = {}): Middleware<Method, Params> {
+  let { path: pathResolver = requestPathnameResolver, ...fileHandlerOptions } = options
+
+  let handler = createFileHandler<Method, Params>(
+    createFsFileResolver(root, pathResolver),
+    fileHandlerOptions,
+  )
+
+  return async (context, next) => {
+    let response = await handler(context)
```
mjackson: What if we unwound this a bit so we didn't have so much wrapped up inside the `handler`?

For example, we could do something like:

```tsx
let resolveFile = createFsFileResolver(root, pathResolver)
let file = resolveFile(context)

if (file == null) {
  return next()
}

return sendFile(file, context)
```

This way we wouldn't have to generate a 404 `Response` in the handler that we just throw away at this level.

demos/bookstore/app/router.ts
```diff
@@ -28,10 +28,47 @@ middleware.push(formData({ uploadHandler }))
 middleware.push(methodOverride())
 middleware.push(storeContext())
 
+middleware.push(
+  staticFiles('./public/root', {
```
mjackson: This is where `staticFiles` really shines, IMO. In the "global" middleware case.

Trying to re-use `staticFiles` as an inline middleware on a specific route feels a little awkward because we have to pass the param from the route to the middleware somehow. I think I'd prefer to have a more direct approach in a route handler, something like:

```tsx
router.get('/assets/*path', (context) => {
  let file = findFile('./assets', context.params.path)
  if (file != null) return sendFile(file, context) // sendFile could handle ETag, Ranges, etc.
  return new Response('Not Found', { status: 404 })
})
```

mjackson (commented):
This is looking great. Very clean! I love how organized you're keeping the code.

There are just a few small ways I think we could re-arrange things that would make it a little more ergonomic IMO. Hopefully the feedback I left helps.

packages/fetch-router/README.md
```diff
@@ -664,6 +665,177 @@ let button = html`<button>${icon} Click me</button>` // icon is not escaped
...
+  return res.file(file, context)
+})
+```
+
+##### Options
+
+The `file()` helper accepts an optional third argument with configuration options:
+
+```ts
+return res.file(file, context, {
```
markdalgleish: Context could be Headers / SuperHeaders.

packages/fetch-router/README.md
```diff
@@ -664,6 +665,177 @@ let button = html`<button>${icon} Click me</button>` // icon is not escaped
...
+
+When using strong ETags, you can provide a cache to avoid re-hashing files on every request:
+
+```ts
+return res.file(file, context, {
+  etag: 'strong',
+
+  // Provide a cache to avoid re-hashing files on every request.
+  // Defaults to `undefined`, meaning that there is no cache.
+  digestCache: new Map(),
```
markdalgleish: Let's drop this for now, come back to it if we need it.

demos/bookstore/app/router.ts
```diff
@@ -28,10 +30,43 @@ middleware.push(formData({ uploadHandler }))
...
+    lastModified: false,
+    acceptRanges: false,
+  }),
+)
+
 export let router = createRouter({ middleware })
 
-router.get(routes.assets, publicHandlers.assets)
-router.get(routes.images, publicHandlers.images)
+router.get(routes.assets, async (context) => {
```
markdalgleish: This could all be folded into a single top-level middleware. We can keep the routes for href generation, but we don't need separate route handlers for them.

packages/fetch-router/README.md
```diff
@@ -664,6 +665,177 @@ let button = html`<button>${icon} Click me</button>` // icon is not escaped
...
+})
+```
+
+By default, the middleware uses the full request pathname to resolve files. You can customize this with a `path` resolver function:
+
+```ts
+router.get('/assets/*path', {
+  middleware: [
+    staticFiles('./public', {
+      path: (context) => context.params.path,
```
markdalgleish: Let's drop this for now.

markdalgleish: Maybe introduce a filter function and dotfiles option?

packages/fetch-router/README.md
```diff
@@ -664,6 +665,177 @@ let button = html`<button>${icon} Click me</button>` // icon is not escaped
...
+When you need to map a route pattern to a directory of files on disk, you can use the `findFile()` function from `@remix-run/lazy-file/fs` to resolve files before sending them with the `file()` response helper:
+
+```ts
+import * as res from '@remix-run/fetch-router/response-helpers'
+import { findFile } from '@remix-run/lazy-file/fs'
+
+router.get('/assets/*path', async (context) => {
+  let file = await findFile('./public/assets', context.params.path)
+
+  if (!file) {
```
markdalgleish, line 770: Add a dotfile check here.

packages/headers/src/lib/if-range.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    // Try parsing as HTTP date first
+    let dateTimestamp = parseHttpDate(this.value)
+    if (dateTimestamp !== null && resource.lastModified != null) {
+      let resourceTimestamp =
+        resource.lastModified instanceof Date
+          ? resource.lastModified.getTime()
+          : resource.lastModified
+      return roundToSecond(dateTimestamp) === roundToSecond(resourceTimestamp)
```
markdalgleish: Double check how rounding works when initially sending the file.

packages/headers/src/lib/super-headers.ts
```diff
@@ -179,7 +208,7 @@ export class SuperHeaders extends Headers {
           let descriptor = Object.getOwnPropertyDescriptor(SuperHeaders.prototype, name)
           if (descriptor?.set) {
             descriptor.set.call(this, value)
-          } else {
+          } else if (value !== undefined) {
             this.set(name, value.toString())
```
markdalgleish, line 212:
```suggestion
            this.set(name, String(value))
```

markdalgleish: I've opted to roll back my changes to the constructor and its types for now since this opened up a bit of a can of worms. We can address this separately. For now I'm just going to avoid passing undefined values to the constructor rather than updating the constructor to handle them.

packages/headers/src/lib/if-range.ts
markdalgleish: Confirmed that milliseconds are simply removed, so `Math.floor` is the correct logic. I've renamed the util to better communicate this.

packages/fetch-router/README.md
markdalgleish: I decided to go with just a `filter` function since it means we're not providing any built-in opinions, and it makes it easier to migrate from middleware to a custom handler since you always own the filtering logic.

packages/fetch-router/README.md
markdalgleish: We need the request method too, so I've opted to accept a `request` arg here rather than the full context.

packages/fetch-router/src/lib/response-helpers/file.ts
```diff
@@ -0,0 +1,337 @@
...
+): Promise<Response> {
+  let {
+    cacheControl,
+    etag: etagStrategy = 'weak',
+    digest: digestOption = 'SHA-256',
+    lastModified: lastModifiedEnabled = true,
+    acceptRanges: acceptRangesEnabled = true,
+  } = options
+
+  if (request.method !== 'GET' && request.method !== 'HEAD') {
```
mjackson: We shouldn't need this check in the response helper. By the time someone calls `res.file()`, they already know they want to generate a file response, regardless of the request method.

packages/fetch-router/src/lib/response-helpers/file.ts
```diff
@@ -0,0 +1,337 @@
...
+ * let result = await findFile('./public', 'image.jpg')
+ * if (result) {
+ *   return file(result, request, {
+ *     cacheControl: 'public, max-age=3600'
+ *   })
+ * }
+ */
+export async function file(
+  fileToSend: File,
+  request: Request,
```
mjackson, line 92: Not a huge deal, but we could do just fine with `requestHeaders: Headers` here instead of a full-blown `Request`. It could simplify testing, for example.

packages/fetch-router/src/lib/response-helpers/file.ts
```diff
@@ -0,0 +1,337 @@
...
+
+function generateWeakETag(file: File): string {
+  return `W/"${file.size}-${file.lastModified}"`
+}
+
+type OmitNullableValues<T> = {
+  [K in keyof T as T[K] extends null | undefined ? never : K]: NonNullable<T[K]>
+}
+
+function omitNullableValues<T extends Record<string, any>>(headers: T): OmitNullableValues<T> {
```
mjackson, line 274: I believe we modified the `SuperHeaders` constructor to not error when it gets `null` and/or `undefined` values, right? Do we still need this?

packages/fetch-router/src/lib/response-helpers/file.ts
```diff
@@ -0,0 +1,337 @@
...
+
+/**
+ * Hashes a file using SubtleCrypto.
+ *
+ * @param file - The file to hash
+ * @param algorithm - Hash algorithm name (e.g., 'SHA-256')
+ * @returns The hash as a hex string
+ */
+async function hashFile(file: File, algorithm: string): Promise<string> {
+  let buffer = await file.arrayBuffer()
```
mjackson: This is totally just an optimization, but it's a shame we have to buffer the file here. Did you explore any way to create the digest incrementally, in a stream? I don't think `crypto.subtle` supports it, but I know `node:crypto` does. `node:crypto` is widely supported by all the various vendors we support. Maybe consider using it here?

packages/fetch-router/README.md
```diff
@@ -664,6 +665,153 @@ let button = html`<button>${icon} Click me</button>` // icon is not escaped
...
+```
+
+Or provide a custom digest function:
+
+```ts
+return res.file(file, request, {
+  etag: 'strong',
+
+  // Custom digest function
+  digest: async (file) => {
```
mjackson: Definitely a nit-pick, but I always prefer using shorthand methods over arrow functions in object literals like this. It's shorter and involves less syntax noise.


```suggestion
  async digest(file) {
```

mjackson (changes requested):
Looking good. So much great stuff in this PR! 💪

packages/fetch-router/src/lib/response-helpers/file.ts
markdalgleish: It's not just the headers that we use. We also check `request.method` so that we can ensure the body is empty on `HEAD` requests.

packages/fetch-router/src/lib/response-helpers/file.ts
markdalgleish: I explored this but it opened a can of worms. Properties that have a setter defined currently ignore `null`/`undefined`, but then the other properties were coercing undefined values to `"undefined"`. The inconsistency felt like a big foot gun. I'm keen to address this but didn't want it to get in the way of this PR, so this change was made to avoid needing to touch this logic for now.

PR #10811, opened 2025-11-03
Bring over RR session code into @remix-run/session package

brophdawg11:
Stacked on #10796 

This is a replacement for https://github.com/remix-run/remix/pull/10797 which got a bit messy after a few rebases on top of breaking router changes in `main`.  I squashed into a few logical commits to add the session package, integrate with fetch-router, and update the bookstore demo.

Issue #10794, opened 2025-10-21
Add `session-middleware` to `fetch-router`


mjackson:
The session middleware is responsible for populating the request context's `session` property.

Session middleware should properly retrieve the current session when a session cookie is sent in the request. If no session cookie is sent, a new session should be created.

When this work is done, the bookstore demo should be updated to use session middleware for automatic session management.

PR #10822, opened 2025-11-06
Add session package

### LINKED ISSUE #10790: Add `session` package
This package is a standalone version of Remix v1's session handling.

Current code can be found [here](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/server-runtime/sessions.ts).

It should also include some concrete implementations of session storage interfaces, including one for the filesystem and one for Redis.

mjackson:
This PR adds a new `@remix-run/session` package and a `session` middleware to `fetch-router`.

It's a complete rewrite of the session handling code we currently have in Remix v2 and RR v7. The main motivations for this work were:

- Less ceremony around session storage creation
- More powerful session management API that supports regenerating IDs and smarter/automatic handling of `session.flash()` values
- Add `FileSessionStorage` for storing sessions on the filesystem

This PR closes #10790 and #10794, and supersedes #10811

### DIFF
--- a/packages/session/src/lib/session.test.ts
+++ b/packages/session/src/lib/session.test.ts
@@ -50,10 +50,10 @@ describe('Session', () => {
   it('sets and gets values with complex types', () => {
     let session = new Session(undefined, [{ user: { id: 123, name: 'alice' } }, {}])
 
-    assert.equal(session.get('user'), { id: 123, name: 'alice' })
+    assert.deepEqual(session.get('user'), { id: 123, name: 'alice' })
 
     session.set('user', { id: 456, name: 'bob' })
-    assert.equal(session.get('user'), { id: 456, name: 'bob' })
+    assert.deepEqual(session.get('user'), { id: 456, name: 'bob' })
 
     // @ts-expect-error - should not allow different type for user
     session.set('user', { id: 456, name: 'bob', roles: ['admin'] })


packages/cookie/src/lib/cookie.ts
```diff
@@ -6,6 +6,16 @@ import {
...
 
+/**
+ * Creates a new `Cookie` object.
+ * @param name The name of the cookie
+ * @param options The options for the cookie
+ * @returns A new cookie instance
+ */
+export function createCookie(name: string, options?: CookieOptions): Cookie {
+  return new Cookie(name, options)
+}
```
brophdawg11, line 17: I'm curious what made you decide to bring this API back?  It looks like we use both in this PR - `createCookie()` in a few tests versus `new Cookie()` in the session middleware.

packages/fetch-router/src/lib/middleware/session.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Middleware that manages `context.session` based on the session cookie.
+ * @param cookie The cookie to use for the session (should be signed)
+ * @param options The options for the session middleware
+ * @returns The session middleware
+ */
+export function session(
+  cookie: Cookie | (CookieOptions & { name: string }),
```
brophdawg11: I like the decoupling of the the session cookie - it feels cleaner, but it does make for a bit of an awkward API when you want to customize the cookie options (statically) since we eliminated the ability to do that at cookie creation time:

```tsx
// My session cookie is signed.  Is it HttpOnly?  I'm not sure...
let sessionCookie = createCookie('__sess', { secrets: ['secret1'] })

...

let router = createRouter({
  middleware: [
    session(sessionCookie, {
      // Ah, ok it's HttpOnly and Secure
      cookie: {
        secure: true,
        httpOnly: true,
      },
    }),
  ],
})
```

If you don't want/need to change those at response time, I think this reads a lot cleaner:

```tsx
// My session cookie is signed, HttpOnly, and Secure
let sessionCookie = createCookie('__sess', { 
  secure: true,
  httpOnly: true,
  secrets: ['secret1'],
})

...

let router = createRouter({
  middleware: [session(sessionCookie)],
})
```

We definitely need the `sessionOptions.cookie: ((context: RequestContext) => CookieProperties)` API for "response-time" dynamic values and I can see a _small_ argument of a consistent API for static and dynamic cookie options.  But I think I would still prefer doing it at `new Cookie()` time (which feels like the more common use case).

packages/fetch-router/src/lib/request-context.ts
```diff
@@ -88,6 +89,19 @@ export class RequestContext<
...
 
+  /**
+   * The current session.
+   */
+  get session(): Session {
+    return (this.#session ??= new Session())
+  }
+
+  set session(value: Session) {
+    this.#session = value
```
brophdawg11, line 108: Should this warn/throw if a `Session` already exists?  Folks shouldn't be able to define multiple session middlewares which would result in odd behaviors.

packages/fetch-router/src/lib/middleware/session.ts
```diff
@@ -0,0 +1,78 @@
...
+        // Session id changed, delete the previous session data
+        await sessionStorage.delete(originalId)
+      }
+
+      newCookieValue = await sessionStorage.update(context.session.id, context.session.data)
+    }
+
+    if (newCookieValue != null) {
+      let cookieProps =
+        typeof options?.cookie === 'function' ? options.cookie(context) : options?.cookie
```
brophdawg11: I had been defaulting to an `HttpOnly` cookie session storage when they didn't specify anything previously so we could be secure by default and force you to opt-into lower security with a JS-visible cookie.

It's a little different with the decoupling of the cookie and session but would we still want to make it the default at `Set-Cookie` time?

```suggestion
      let cookieProps =
        typeof options?.cookie === 'function' ? options.cookie(context) : 
        options?.cookie != null ? options.cookie : 
        { httpOnly: true };
```

brophdawg11 (commented):
Looks good!  I like the cleaner decoupling of the cookie/session and the new storage APIs of `update`/`delete` instead of `commitSession`/`destroySession` 👍

packages/cookie/src/lib/cookie.ts
mjackson: It's a personal preference, but I dislike forcing people to use `new Blah` in their code. I'm doing the same thing with `createRouter` vs `new Router`. They can do either one, but I always like providing a way for them to use our stuff w/out using `new`.

packages/fetch-router/src/lib/request-context.ts
mjackson: Good call. Instead of `throw`ing here, let's do it in the middleware when it runs. I'll add a `session.size` accessor we can use to test if the session has been used yet. If so, we'll `throw` instead of overwriting `context.session`.

packages/fetch-router/src/lib/middleware/session.ts
mjackson: Another good call. I like defaulting to `httpOnly: true` for session cookies.

packages/fetch-router/src/lib/middleware/session.ts
mjackson: Yeah, I hear you. I'm gonna go ahead and add back the ability to setup `CookieProperties` on the original `Cookie` instance when it's first created. We can keep the `session(..., { cookie })` option for overrides.

packages/session/src/lib/session.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.#id = Session.createId()
+    this.#dirty = true
+  }
+
+  /**
+   * The number of key/value pairs in the session.
+   */
+  get size(): number {
+    return this.#valueMap.size + this.#flashMap.size
+  }
```
brophdawg11, line 131: Did we end up using this for anything?

mjackson: Nah, I ended up adding a `context.sessionStarted` flag instead and issuing a warning if someone starts using `context.session` without a `session()` middleware in the stack since it's a no-op. I still think `session.size` is a nice-to-have though, so I left it in.

PR #10835, opened 2025-11-11
fetch-router: fix route handler type safety and make middleware optional

### LINKED ISSUE #10830: middleware in `router.map` does not apply for multiple handlers
I'm not certain whether this is a runtime bug, TypeScript bug, or just an API I don't completely understand. My guess is this has more to do with the types than actual runtime behavior, but I could be wrong

Take this simplified example from the ['router.map() with middleware' test](https://github.com/remix-run/remix/blob/main/packages/fetch-router/src/lib/router.test.ts#L227-L254):

```ts
let routes = route({
  profile: '/profile/:id',
})

let router = createRouter()

router.map(routes.profile, {
  middleware: [() => console.log('run the middleware')],
  // 👆 this middleware runs
  handler() {
    return new Response('OK')
  },
})

let response = await router.fetch('https://remix.run/profile/1')
```

However, if you map over multiple routes, the middleware will never be called

```ts
let routes = route({
  profile: {
    show: '/profile/:id',
    new: '/profile/new',
  }
})

let router = createRouter()

router.map(routes.profile, {
  middleware: [() => console.log('run the middleware')],
  // 👆 this middleware DOES NOT run
  show() {
    return new Response('OK')
  },
  new() {
    return new Response('OK')
  }
})

let response = await router.fetch('https://remix.run/profile/1')
```

The reason I tried this is because I saw that `middleware` can be applied to a `router.map` call, and my understanding is that `router.map` is most useful when applying handlers to multiple routers

Either:
- The types are wrong and `middleware` should _not_ be available as a method when defining multiple handlers like this(note, this is what TypeScript encourages me to type):
Image

- `middleware` should run for all handlers defined in this `map`. I would find this a bit confusing though, given that `middleware` is at the same level as the route names

brookslybrand:
From https://github.com/remix-run/remix/issues/10830#issuecomment-3518320980, this PR addresses the following false positives and type errors without breaking behavior.

Also makes `middleware` optional where it wasn't before (not just for `router.map` but also `router.get` and friends)

```ts
let routes = route({
  home: '/',
})

let router = createRouter()

// ❌ ISSUE 1: Currently allows mixing handlers with direct route handlers
router.map(routes, {
  handlers: {
    home() {
      return new Response('OK')
    },
  },
  home() {  // ❌ Mixing patterns should not be allowed
    return new Response('OK')
  },
})

// ❌ ISSUE 2: Currently allows mixing middleware with direct route handlers
router.map(routes, {
  middleware: [() => {}],
  home() {  // ❌ Middleware doesn't run here, should be `handlers`
    return new Response('OK')
  },
})

// ❌ ISSUE 3: Handler object without middleware causes a type error
router.get('/', {
  // middleware: [],  // Have to include an empty middleware function to avoid the type error
  handler() {  // ❌ Type error: middleware was required
    return new Response('OK')
  },
})
```

You can see the [failing tests](https://github.com/remix-run/remix/pull/10835/commits/79b25c0159c7f31ad63f7783fbfb72b502c87942) before apply the changes

Fixes #10830

### DIFF
--- a/packages/fetch-router/src/lib/route-handlers.ts
+++ b/packages/fetch-router/src/lib/route-handlers.ts
@@ -8,28 +8,52 @@ import type { Route, RouteMap } from './route-map.ts'
 // prettier-ignore
 export type RouteHandlers<routes extends RouteMap> =
   | RouteHandlersWithMiddleware<routes>
-  | {
+  | RouteHandlersWithoutMiddleware<routes>
+
+type RouteHandlersWithMiddleware<routes extends RouteMap> = {
+  middleware?: Middleware[]
+  handlers: RouteHandlers<routes>
+} & (routes extends Record<string, any>
+  ? {
+      // Explicitly exclude route handler properties when handlers is present
+      // Only apply exclusion when routes is a concrete type (not any)
+      [name in keyof routes as routes extends any ? never : name]?: never
+    }
+  : {})
+
+// prettier-ignore
+type RouteHandlersWithoutMiddleware<routes extends RouteMap> = routes extends any
+  ? {
       [name in keyof routes]: (
         routes[name] extends Route<infer method extends RequestMethod | 'ANY', infer pattern extends string> ? RouteHandler<method, pattern> :
         routes[name] extends RouteMap ? RouteHandlers<routes[name]> :
         never
       )
+    } & {
+      // Explicitly exclude middleware/handlers when route handlers are present
+      middleware?: never
+      handlers?: never
     }
+  : never
 
-type RouteHandlersWithMiddleware<routes extends RouteMap> = {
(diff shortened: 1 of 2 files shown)


packages/fetch-router/src/lib/router.test.ts
```diff
@@ -253,6 +254,28 @@ describe('router.map() with middleware', () => {
     assert.deepEqual(requestLog, ['middleware 1', 'handler'])
   })
 
+  it('supports handler object without middleware', async () => {
```
brookslybrand, line 257: @mjackson I'm not sure the best way to handle testing this. I think overall the changes to these tests is likely overly verbose

Please feel free to delete, change, or recommend a better approach to handling the tests

PR #10840, opened 2025-11-14
Add compression response helper and middleware

markdalgleish:
See docs in PR for details.

Note this also introduces a new `mime` package, plus an update to `headers` to add `Vary` support.


packages/fetch-router/src/lib/response-helpers/compress.ts
```diff
@@ -0,0 +1,208 @@
+import { constants, createBrotliCompress, createDeflate, createGzip } from 'node:zlib'
+import type { BrotliOptions, ZlibOptions } from 'node:zlib'
+import { Readable } from 'node:stream'
+import type { Transform } from 'node:stream'
```
MichaelDeBoey: Using these Node.js APIs will break our promise that this package is only built on Fetch API and will work everywhere

> - **Fetch API**: Built on standard web APIs that work everywhere - Node.js, Bun, Deno, Cloudflare Workers, and browsers

CC/ @mjackson

markdalgleish:
@mjackson Thanks for the review! I've pushed up some big changes since you've had a look at it, so it might be worth another look.

packages/fetch-router/CHANGELOG.md
```diff
@@ -24,6 +26,18 @@ This is the changelog for [`fetch-router`](https://github.com/remix-run/remix/tr
 
   This change improves the ergonomics of the router by eliminating the need to bind methods when passing `router.fetch` as a callback, for example in `node-fetch-server`'s `createRequestListener(router.fetch)`.
 
+- The `file()` response helper now only enables HTTP Range requests by default for non-compressible MIME types. This allows text-based assets to be compressed while still supporting resumable downloads for media files.
```
MichaelDeBoey: I guess this is technically also a breaking change?

packages/headers/src/lib/accept.ts
```diff
@@ -114,7 +114,7 @@ export class Accept implements HeaderValue, Iterable<[string, number]> {
    * @param mediaTypes The list of media types to choose from.
    * @returns The most preferred media type or `null` if none match.
    */
-  getPreferred(mediaTypes: string[]): string | null {
+  getPreferred<MediaType extends string>(mediaTypes: readonly MediaType[]): MediaType | null {
```
mjackson:
```suggestion
  getPreferred<mediaType extends string>(mediaTypes: readonly MediaType[]): mediaType | null {
```

packages/headers/src/lib/accept-language.ts
```diff
@@ -115,7 +115,7 @@ export class AcceptLanguage implements HeaderValue, Iterable<[string, number]> {
    * @param languages The locale identifiers of the languages to choose from.
    * @returns The most preferred language or `null` if none match.
    */
-  getPreferred(languages: string[]): string | null {
+  getPreferred<Language extends string>(languages: readonly Language[]): Language | null {
```
mjackson: We typically use lowercase names for generics.

```suggestion
  getPreferred<language extends string>(languages: readonly Language[]): language | null {
```

packages/headers/src/lib/accept-encoding.ts
```diff
@@ -110,7 +110,7 @@ export class AcceptEncoding implements HeaderValue, Iterable<[string, number]> {
    * @param encodings The encodings to choose from.
    * @returns The most preferred encoding or `null` if none match.
    */
-  getPreferred(encodings: string[]): string | null {
+  getPreferred<Encoding extends string>(encodings: readonly Encoding[]): Encoding | null {
```
mjackson:
```suggestion
  getPreferred<encoding extends string>(encodings: readonly Encoding[]): encoding | null {
```

packages/fetch-router/src/lib/response-helpers/compress.ts
mjackson: I'd like to use `node:zlib`, but not use `node:stream`.

`CompressionStream` and `DecompressionStream` are not ready for production yet, and `node:zlib` is supported everywhere. So we prefer `node:zlib` over web compression APIs.

However, the web does have a mature streams implementation that we should use whenever we can.

@markdalgleish Instead of using `Readable.toWeb` and `Readable.fromWeb`, can we somehow drop down to a lower level API and manually pump the bytes through `node:zlib`'s compression algorithms?

demos/sse/app/assets/entry.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+    let Component = mod[name]
+    if (!Component) {
+      throw new Error(`Unknown component: ${moduleUrl}#${name}`)
+    }
+
+    return Component
+  },
+
+  async resolveFrame(frameUrl) {
```
mjackson: You shouldn't need this at all unless you're using `<Frame>`, which I didn't see anywhere. If it's an error to  omit it, we should fix that.

mjackson (changes requested):
Left a few notes, thanks for taking this on @markdalgleish!

packages/fetch-router/CHANGELOG.md
mjackson: No. Previously the compressible files were served with `Accept-Ranges: bytes` in the response. Now they're not. But clients should be able to work with either one.

PR #10841, opened 2025-11-14
test: add a variety of test cases

43081j:
Was just getting my head around the codebase, so here's a few tests I wrote to help me do that.

### DIFF
--- a/packages/cookie/src/lib/cookie.test.ts
+++ b/packages/cookie/src/lib/cookie.test.ts
@@ -168,4 +168,18 @@ describe('Cookie', () => {
     assert.ok(setCookie.includes('Secure'))
     assert.ok(setCookie.includes('HttpOnly'))
   })
+
+  it('parses empty values as null', async () => {
+    let cookie = createCookie('my-cookie')
+    let value = await cookie.parse(null)
+
+    assert.equal(value, null)
+  })
+
+  it('parses missing cookies as null', async () => {
+    let cookie = createCookie('my-cookie')
+    let value = await cookie.parse('other-cookie=hello')
+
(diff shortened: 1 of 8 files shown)


43081j:
@mjackson any chance we can merge this sooner rather than later? things are changing often so it'll just keep getting conflicts 😬

packages/lazy-file/src/fs.test.ts
```diff
@@ -218,3 +218,47 @@ describe('findFile', () => {
     teardown()
   })
 })
+
+describe('writeFile', () => {
+  let tmpDir: string
+
+  function setup() {
```
mjackson: Let's use `beforeEach` and `afterEach` instead of calling `setup()` and `teardown()` manually in each test case below

mjackson (commented):
Overall great additions to our tests, thank you. I'd just like to use `node:test`'s built-in `beforeEach` and `afterEach` hooks and we should be good to merge 👍

packages/lazy-file/src/fs.test.ts
43081j: this is just me following convention of the rest of the file. the other tests already use this same setup/teardown pattern

im assuming you want me to update those too

43081j:
i've caught this up again

@mjackson you're duplicating work so we should try merge this soon. e.g. the beforeEach/afterEach comment, you ended up doing in `main`. you also added the same missing fs tests in `main`

MichaelDeBoey (approved):

mjackson:
Merged. Thanks, @43081j 👍  and sorry for the churn. Things are moving quickly!

Issue #10828, opened 2025-11-09
Getting 'TypedEventTarget' resolves to a type-only declaration and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.


raythurnvoid:
I'm using the `@remix-run/interaction` in an existing React project to use `TypedEventTarget`, however typescript complains with `'TypedEventTarget' resolves to a type-only declaration and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled.`.

I tried to remove `type` from the `index.d.ts` and it worked: 

<img width="1301" height="182" alt="Image" src="https://github.com/user-attachments/assets/582c78bc-e42e-460d-abdf-3f432784bce6" />

Wubbwubb:
this issue is solved with commit https://github.com/remix-run/remix/commit/3b04b1c3fc070917efce21e15fa01f4dcc78fbde. will probably be part of version 0.2.2 or 0.3.0 maybe ✌️ 
i'll close the related pull request

Wubbwubb:
> this issue is solved with commit [3b04b1c](https://github.com/remix-run/remix/commit/3b04b1c3fc070917efce21e15fa01f4dcc78fbde). will probably be part of version 0.2.2 or 0.3.0 maybe ✌️ i'll close the related pull request

version 0.3.0 should be the way to go 😉

PR #10842, opened 2025-11-14
fix(interaction): export `TypedEventTarget` as class

Wubbwubb:
provides a fix for https://github.com/remix-run/remix/issues/10828


Wubbwubb:
fixes #10828

PR #10844, opened 2025-11-14
Add fetch-router node.js demo

brookslybrand:
Went with the classic blog example. @mjackson please let me know if you want me to increase or decrease scope.

Comments call out code where I anticipate some discussion. You can also watch this quick walkthrough:

https://github.com/user-attachments/assets/8b3aa3d6-7518-4613-ae0e-3cd95bdfed62

## Features:

- **Blog CRUD operations**: Full RESTful resource routes for posts (index, new, create, show, edit, update, destroy) with custom date-based URL patterns (`/posts/:year-:month-:day/:slug`)
- **Nested resources**: Comments nested under posts with their own CRUD operations (create, destroy)
- **Authentication flow**: Login/logout with session management, flash messages for errors, and `redirectTo` query parameter handling for post-login redirects
- **Authorization**: Custom middleware for route-level auth (`requireAuth`, `requireGuest`) and resource-level authorization (`requireCommentAuthor` for comment deletion)
- **Form handling**: POST forms with method override support (PUT/DELETE via `_method` hidden field)
- **HTML templating**: XSS-safe HTML generation with `@remix-run/html-template`, reusable template components, and proper escaping
- **Static file serving**: CSS and favicon served with cache control and ETag headers

## Middleware used:
- `logger()`
- `formData()`
- `methodOverride()`
- `session()` - Cookie-based session management with `@remix-run/cookie`
- `staticFiles()` - Serving favicon and stylesheet
- Custom middleware composition - Route-level and handler-level middleware, typed middleware with pattern-specific params

## Other packages demonstrated:
- `@remix-run/html-template` - `html` template literals, `SafeHtml` type, XSS-safe HTML generation
- `@remix-run/cookie` - `createCookie` for signed session cookies
- `@remix-run/node-fetch-server` - `createRequestListener` for Node.js HTTP server integration
- `@remix-run/fetch-router/response-helpers` - `res.html()`, `res.redirect()` utilities

Resolves #10832

PR #10848, opened 2025-11-18
feat(fetch-router): add Node.js demo

brookslybrand:
Simplified version of #10844

Demonstrates basic setup of `fetch-router` 

- `server.ts` -- `createRequestListener` from `@remix-run/node-fetch-server` to listen to requests and call `router.fetch`
- `routes.ts` -- route definitions, leverages `formAction` and `resources` for simple CRUD routes
- `router.ts` -- sets up middleware and route handlers for routes defined in `routes.ts`

https://github.com/user-attachments/assets/3df7c9fa-cbbb-4c68-9185-9742537147f6

### DIFF
--- a/packages/fetch-router/demos/node/data.ts
+++ b/packages/fetch-router/demos/node/data.ts
@@ -24,7 +24,7 @@ let posts: Post[] = [
 ]
 
 export function getPosts() {
-  return [...posts].sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime())
+  return posts.toSorted((a, b) => b.createdAt.getTime() - a.createdAt.getTime())
 }
 
 export function getPost(id: string) {


packages/fetch-router/demos/node/router.ts
```diff
@@ -0,0 +1,195 @@
...
+
+router.map(routes.login, {
+  index({ session }) {
+    let username = session.get('username') as string | undefined
+    if (username) {
+      return res.redirect(routes.home.href())
+    }
+
+    return res.html(html`
+      <!doctype html>
```
mjackson, line 88: Ah, this reminds me. We should add this by default in the `res.html()` helper.

MichaelDeBoey: Done in https://github.com/remix-run/remix/commit/e4f41c48fcdfa7f2f8268634d2f0673e204ea3f5

packages/fetch-router/demos/node/data.ts
```diff
@@ -0,0 +1,44 @@
...
+    id: '2',
+    title: 'Getting Started with fetch-router',
+    content: 'fetch-router is a minimal, composable router built on the web Fetch API.',
+    author: 'Admin',
+    createdAt: new Date('2025-01-02'),
+  },
+]
+
+export function getPosts() {
+  return [...posts].sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime())
```
MichaelDeBoey: This can be simplified to using [`Array.prototype.toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted)


```suggestion
  return posts.toSorted((a, b) => b.createdAt.getTime() - a.createdAt.getTime())
```

MichaelDeBoey (approved):

packages/fetch-router/demos/node/data.ts
brookslybrand: Oh I forgot they made a non-mutating sort method. Thanks!

Shamelessly, AI picked the `[...posts].sort` approach 😅. Happy to make this change though

PR #10851, opened 2025-11-20
feat(fetch-router): add Cloudflare Workers demo

### LINKED ISSUE #10834: Add CF workers demo to fetch-router
Similar to #10832, this demo should show fetch-router running on Cloudflare Workers. It should be its own pnpm workspace in `packages/fetch-router/demos/cf-workers`. See [the `multipart-parser` cf-workers demo](https://github.com/remix-run/remix/tree/main/packages/multipart-parser/demos/cf-workers) for an example of setup. It should deploy using `wrangler`.

brookslybrand:
Ran the following to initialize the project and deleted all the stuff I didn't want (testing, unrelated editor config stuff):

```sh
npm create cloudflare@latest -- cf-workers
```

Verified it's [deployable](https://fetch-router-cf-workers-demo.brookslybrand.workers.dev/)

Resolves #10834

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -34,7 +34,7 @@ jobs:
 
       - name: Get Playwright Version
         id: playwright-version
-        run: echo "version=$(pnpm exec playwright --version | cut -d ' ' -f2)" >> $GITHUB_OUTPUT
+        run: echo "version=$(pnpm --filter @remix-run/interaction exec playwright --version | cut -d ' ' -f2)" >> $GITHUB_OUTPUT
 
       - name: Cache Playwright Browsers
         uses: actions/cache@v4
@@ -44,10 +44,8 @@ jobs:
           key: playwright-${{ runner.os }}-${{ steps.playwright-version.outputs.version }}
 
       - name: Install Playwright Browsers
-        run: |
-          if [ "${{ steps.cache-browsers.outputs.cache-hit }}" != "true" ]; then
-            pnpm exec playwright install --with-deps
-          fi
+        if: steps.cache-browsers.outputs.cache-hit != 'true'
+        run: pnpm --filter "@remix-run/*" exec playwright install --with-deps 2>/dev/null || true
 
       - name: Run tests
         run: pnpm test
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Get Playwright Version
         id: playwright-version
-        run: echo "version=$(pnpm exec playwright --version | cut -d ' ' -f2)" >> $GITHUB_OUTPUT
+        run: echo "version=$(pnpm --filter @remix-run/interaction exec playwright --version | cut -d ' ' -f2)" >> $GITHUB_OUTPUT
 
       - name: Cache Playwright Browsers
         uses: actions/cache@v4
@@ -40,10 +40,8 @@ jobs:
           key: playwright-${{ runner.os }}-${{ steps.playwright-version.outputs.version }}
 
       - name: Install Playwright Browsers
-        run: |
-          if [ "${{ steps.cache-browsers.outputs.cache-hit }}" != "true" ]; then
-            pnpm exec playwright install --with-deps
-          fi
+        if: steps.cache-browsers.outputs.cache-hit != 'true'
+        run: pnpm --filter "@remix-run/*" exec playwright install --with-deps 2>/dev/null || true
 
       - name: Run tests
         run: pnpm test
--- a/packages/interaction/package.json
+++ b/packages/interaction/package.json
@@ -62,7 +62,7 @@
   "devDependencies": {
     "@vitest/browser": "^3.2.4",
     "esbuild": "^0.25.5",
-    "playwright": "^1.53.1",
+    "playwright": "catalog:",
     "typescript": "^5.9.3",
     "vitest": "^3.2.4"
   }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,6 +4,12 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+catalogs:
+  default:
+    playwright:
+      specifier: ^1.53.1
+      version: 1.56.1
+
 importers:
 
   .:
@@ -399,7 +405,7 @@ importers:
         specifier: ^0.25.5
         version: 0.25.10
       playwright:
-        specifier: ^1.53.1
+        specifier: 'catalog:'
         version: 1.56.1
       typescript:
         specifier: ^5.9.3
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,4 +1,5 @@
-shellEmulator: true
+catalog:
+  playwright: ^1.53.1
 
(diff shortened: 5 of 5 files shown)


packages/fetch-router/demos/cf-workers/worker.ts
```diff
@@ -0,0 +1,12 @@
+import { router } from './router.ts'
+
+export default {
```
brookslybrand, line 3: Pretty much the only file that changed. I opted for `worker.ts` instead of `server.ts` just to make it a little more obvious

packages/fetch-router/demos/cf-workers/wrangler.jsonc
```diff
@@ -0,0 +1,14 @@
+/**
+ * For more details on how to configure Wrangler, refer to:
+ * https://developers.cloudflare.com/workers/wrangler/configuration/
+ */
+{
+  "$schema": "node_modules/wrangler/config-schema.json",
+  "name": "fetch-router-cf-workers-demo",
+  "main": "worker.ts",
+  "compatibility_flags": ["nodejs_compat"],
```
brookslybrand: I had to add this because [we use `node:crypto`](https://github.com/remix-run/remix/blob/222d187cf02d6792dbc6a55317d752e2f5ff87e7/packages/fetch-router/src/lib/response-helpers/file.ts#L309)

packages/fetch-router/demos/cf-workers/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "isolatedModules": true,
+    "allowSyntheticDefaultImports": true,
+    "forceConsistentCasingInFileNames": true,
+    "allowImportingTsExtensions": true,
+    "rewriteRelativeImportExtensions": true,
+    "verbatimModuleSyntax": true,
+    "skipLibCheck": true,
+    "types": ["./worker-configuration.d.ts", "node"]
+  },
+  "include": ["worker-configuration.d.ts", "worker.ts", "../../global.d.ts"]
```
brookslybrand: Would love some input on how to get rid of the `../../global.d.ts` inclusion here

I could put [a copy](https://github.com/remix-run/remix/blob/main/packages/fetch-router/global.d.ts) in each demo, but that feels kind of weird as well to just keep propagating this file

packages/fetch-router/demos/cf-workers/data.ts
```diff
@@ -0,0 +1,44 @@
+export interface Post {
+  id: string
+  title: string
+  content: string
+  author: string
+  createdAt: Date
+}
+
+let posts: Post[] = [
```
mjackson, line 9: Instead of storing data in memory, maybe we could put them in a simple Cloudflare D1 database?

packages/fetch-router/demos/cf-workers/router.ts
```diff
@@ -0,0 +1,195 @@
...
+export let router = createRouter({
+  middleware: [logger(), formData(), session(sessionCookie, storage)],
+})
+
+router.map(routes.home, ({ session }) => {
+  let posts = data.getPosts()
+  let username = session.get('username') as string | undefined
+
+  return res.html(html`
+    <!doctype html>
```
mjackson: Same as before, no doctype needed with `res.html()`.

packages/fetch-router/demos/cf-workers/wrangler.jsonc
mjackson: We only need `node:crypto` when you want to generate strong etags. Wrangler's bundling step is probably being aggressive here and trying to inline everything it sees in a dynamic `import()`. How do we tell wrangler to ignore it or just use an empty module?

mjackson (changes requested):
Looking good, thanks @brookslybrand

packages/fetch-router/demos/cf-workers/wrangler.jsonc
brookslybrand: Yeah I think you're right that it's just being aggressive, which kind of makes sense. Better to warn you ahead than blow up in production when you actually try to import it

We have 3 options

1. Don't do anything

We get this ugly warning but everything works fine. It's not my favorite, but this is a simple demo

```
▲ [WARNING] The package "node:crypto" wasn't found on the file system but is built into node.

  Your Worker may throw errors at runtime unless you enable the "nodejs_compat" compatibility flag.
  Refer to https://developers.cloudflare.com/workers/runtime-apis/nodejs/ for more details. Imported
  from:
   - ../../src/lib/response-helpers/file.ts
```

2. We alias it to an empty module

This requires updating `wrangler.jsonc`

```json
  "alias": {
    "node:crypto": "./node-crypto-shim.ts",
  }
```

And then adding `node-crypto-shim.ts`

```
export {}
```

I personally don't really like this, having a random extra module just to get rid of a warning seems like circumventing wrangler instead of just going with whatever it wants

3. We keep it as is and keep the `"compatibility_flags": ["nodejs_compat"]` line in `wrangler.jsonc`


I prefer 3 since it just leans into what `wrangler` is asking for, but up to you, happy to go with whatever you prefer

packages/fetch-router/demos/cf-workers/data.ts
brookslybrand: Good idea, updated!

Let me know if you have any feedback, I tried to keep it pretty simple and easy to use locally/deploy

packages/fetch-router/demos/cf-workers/wrangler.jsonc
brookslybrand: Synced with @mjackson: planning to remove `node:crypto` from response-helpers, at which point we'll merge this

packages/fetch-router/demos/cf-workers/router.ts
```diff
@@ -0,0 +1,192 @@
+import { createRouter } from '@remix-run/fetch-router'
+import { createCookie } from '@remix-run/cookie'
+import { createCookieStorage } from '@remix-run/session/cookie-storage'
```
mjackson: This was renamed to `createCookieSessionStorage`.

packages/fetch-router/demos/cf-workers/router.ts
```diff
@@ -0,0 +1,192 @@
...
+import { formData } from '@remix-run/form-data-middleware'
+import { logger } from '@remix-run/logger-middleware'
+import { session } from '@remix-run/session-middleware'
+import { html } from '@remix-run/html-template'
+import * as res from '@remix-run/fetch-router/response-helpers'
```
mjackson: All response helpers now come from `@remix-run/response/*`. For HTML responses, use:

```tsx
import { createHtmlResponse } from '@remix-run/response/html'

let response = createHtmlResponse(html)
```

For redirects, I like to rename the `import` like:

```tsx
import { createRedirectResponse as redirect } from '@remix-run/response/redirect'

let response = redirect(location)
```

packages/fetch-router/demos/cf-workers/wrangler.jsonc
mjackson: I published a new `createFileResponse` helper in `@remix-run/response/file` that doesn't rely on `node:crypto` anymore. Let's try using that and we should be able to remove the Node.js compat flag from this demo.

PR #10852, opened 2025-11-20
feat(fetch-router): add Bun demo

### LINKED ISSUE #10833: Add Bun demo to fetch-router
Similar to #10832, this demo should run on Bun. It could potentially demonstrate some unique feature of Bun, like serving a Bunfile from the local filesystem. It should not use `bun install` for package management. Stick with `pnpm` for that (same as the rest of the monorepo).

brookslybrand:
Copy of #10848 

Still considering what to do here:

> It could potentially demonstrate some unique feature of Bun, like serving a Bunfile from the local filesystem.

Resolves #10833

### DIFF
--- a/packages/fetch-router/demos/bun/router.ts
+++ b/packages/fetch-router/demos/bun/router.ts
@@ -35,7 +35,6 @@ router.map(routes.home, ({ session }) => {
   let username = session.get('username') as string | undefined
 
   return res.html(html`
-    <!doctype html>
     <html>
       <head>
         <title>Simple Blog - fetch-router Demo</title>
@@ -86,7 +85,6 @@ router.map(routes.login, {
     }
 
     return res.html(html`
-      <!doctype html>
       <html>
         <head>
           <title>Login - Simple Blog</title>
@@ -131,7 +129,6 @@ router.map(routes.posts, {
     middleware: [requireAuth()],
     handler({ session: _session }) {
       return res.html(html`
-        <!doctype html>
         <html>
           <head>
             <title>New Post - Simple Blog</title>
@@ -180,7 +177,6 @@ router.map(routes.posts, {
     }
 
     return res.html(html`
-      <!doctype html>
       <html>
         <head>
           <title>${post.title} - Simple Blog</title>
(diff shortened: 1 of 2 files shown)


packages/fetch-router/demos/bun/router.ts
```diff
@@ -0,0 +1,209 @@
...
+
+// Serve static files from the public directory
+router.get('/*', async ({ request }) => {
+  let urlPath = new URL(request.url).pathname
+  let file = Bun.file(`public${urlPath}`)
+  if (await file.exists()) {
+    return new Response(file)
+  }
+  return new Response('Not Found', { status: 404 })
+})
```
brookslybrand, line 205: Really simple example of showing off `Bun.file`, per your comment @mjackson:

>  It could potentially demonstrate some unique feature of Bun, like serving a Bunfile from the local filesystem.

Let me know if you'd like to take a different approach or do something else.

packages/fetch-router/demos/bun/index.ts
```diff
@@ -0,0 +1,17 @@
+import { router } from './router.ts'
+
+const PORT = 3000
```
mjackson: All our demos should run on port 44100!

packages/fetch-router/demos/bun/router.ts
```diff
@@ -0,0 +1,209 @@
...
+export let router = createRouter({
+  middleware: [logger(), formData(), session(sessionCookie, storage)],
+})
+
+router.map(routes.home, ({ session }) => {
+  let posts = data.getPosts()
+  let username = session.get('username') as string | undefined
+
+  return res.html(html`
+    <!doctype html>
```
mjackson: You don't need this anymore. `res.html()` automatically sets doctypes.

packages/fetch-router/demos/bun/router.ts
mjackson: Sure, that looks great.

packages/fetch-router/demos/bun/index.ts
brookslybrand: Agreed!

https://github.com/remix-run/remix/pull/10855

packages/fetch-router/demos/bun/router.ts
brookslybrand: Sweet! Thanks for calling that out, removed it from this demo and the node.js demo

PR #10854, opened 2025-11-21
test(session): Test deleteId values on regenerateId()

gyx1000:
Ensure deleteId stay undefined with `regenerateId(false)` and is filled with originalId with `regenerateId(true)`


mjackson:
Thanks, @gyx1000!

PR #10855, opened 2025-11-21
Update demos to use port 44100

brookslybrand:
Per this comment: https://github.com/remix-run/remix/pull/10852#discussion_r2548139592

> All our demos should run on port 44100!


mjackson:
That's the spirit!

PR #10858, opened 2025-11-24
feat(fetch-router): add `exclude` option to `createResource` & `createResources`

agionoja:
This PR adds an `exclude` option to `createResource` and `createResources`, providing a more ergonomic way to define CRUD routes when you want most methods except a few.

## Motivation

Currently, when you want most CRUD methods except one or two, you need to explicitly list everything with `only`:

```typescript
// Before: verbose when you only want to exclude one or two methods
const users = resources('users', {
  only: ['index', 'show', 'edit', 'update', 'destroy'] // just to skip 'new' and 'create'
})
```

This is tedious for common scenarios like resources where certain operations are handled elsewhere (e.g., user registration through a separate auth flow, or soft-delete instead of destroy). The `exclude` option makes this cleaner:

```typescript
// After: clearer intent
const users = resources('users', {
  exclude: ['new', 'create'] // registration handled by separate auth system
})
```

## Changes

- Added `exclude` option to `ResourceOptions` and `ResourcesOptions` types
- Implemented mutually exclusive constraint between `only` and `exclude` at both type and runtime levels
- Updated type inference to properly handle the `exclude` option
- Added runtime validation that throws an error if both options are provided
- Added comprehensive test coverage for the new functionality

## Type Safety

The implementation uses discriminated union types to prevent using both `only` and `exclude` simultaneously:

```typescript
// TypeScript error: Type 'never' is not assignable to type...
const invalid = resources('books', {
  only: ['index', 'show'],
  exclude: ['destroy'] // Type error
})
```

Runtime validation also catches this case:

```typescript
// Throws: "Cannot specify both "only" and "exclude" options"
const invalid = resources('books', {
  only: ['index'],
  exclude: ['destroy']
})
```

## Examples

```typescript
// Singleton resource - read-only profile (no create/destroy)
const profile = resource('profile', {
  exclude: ['create', 'destroy']
})
// Generates: show, new, edit, update

// Collection - registration happens via auth system, no deletions
const users = resources('users', {
  exclude: ['new', 'create', 'destroy']
})
// Generates: index, show, edit, update

// Works with all existing options
const posts = resources('posts', {
  exclude: ['destroy'],
  param: 'slug',
  names: { index: 'list', show: 'view' }
})
```

## Backward Compatibility

This change is fully backward compatible:
- Existing code using `only` continues to work unchanged
- No breaking changes to types or function signatures
- The `exclude` option is purely additive

### DIFF
--- a/packages/fetch-router/src/lib/resource.test.ts
+++ b/packages/fetch-router/src/lib/resource.test.ts
@@ -63,6 +63,30 @@ describe('createResource', () => {
     assert.equal((book as any).create, undefined)
   })
 
+  it('creates a resource with exclude option', () => {
+    let book = resource('book', { exclude: ['new', 'create', 'edit', 'destroy'] })
+
+    type T = [
+      Assert<
+        IsEqual<
+          typeof book,
+          {
+            show: Route<'GET', '/book'>
+            update: Route<'PUT', '/book'>
+          }
+        >
+      >,
+    ]
+
+    assert.deepEqual(book.show, new Route('GET', '/book'))
+    assert.deepEqual(book.update, new Route('PUT', '/book'))
+    // Other routes are excluded from the type
+    assert.equal((book as any).new, undefined)
+    assert.equal((book as any).create, undefined)
+    assert.equal((book as any).edit, undefined)
+    assert.equal((book as any).destroy, undefined)
+  })
+
   it('creates a resource with custom route names', () => {
     let book = resource('book', {
       names: {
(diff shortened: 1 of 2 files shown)


mjackson:
I like where you're going with this PR, @agionoja, thank you 🙏 

I'd like to get it merged. Would you mind rebasing on current master?

agionoja:
> I'd like to get it merged. Would you mind rebasing on current master?

Thanks for merging! The timezone difference meant I couldn't rebase in time. Appreciate it.

PR #10862, opened 2025-11-25
feat(logger-middleware): add support for color and pretty formatters

agionoja:
This adds optional colorized output and human-readable formatters for durations and file sizes.

**New features:**
- `colors` option for colorized output (off by default)
- `formatDuration()` - formats ms as "1.20s", "45ms", etc.
- `formatFileSize()` - formats bytes as "1.2 kB", "120 B", etc.
- Updated default format to use `%durationPretty` and `%contentLengthPretty`

**Colors (when enabled):**
- Status codes: green (2xx), cyan (3xx), yellow (4xx), red (5xx)
- Durations: green (<100ms), yellow (100-500ms), magenta (500ms-1s), red (>1s)
- Methods: GET (green), POST (cyan), PUT/PATCH (yellow), DELETE (red), HEAD/OPTIONS (magenta)
- Content length: cyan (1kB-100kB), yellow (100kB-1MB), red (>1MB)

No breaking changes - colors default to off.


mjackson:
Let's default to using color when we detect we are in a TTY. This is going to be tricky since we don't want to lose the ability to run logger middleware in browser/service worker environment.

Something like this could work:

```ts
const useColor = typeof process !== 'undefined' &&
  process.stdout.isTTY &&
  process.env.NO_COLOR == null

function colorize(text: string, color: string) {
  return useColor ? `${color}${text}${reset}` : text
}
```

MichaelDeBoey:
@agionoja Something seems to have gone wrong with rebasing on `main`

agionoja:
> @agionoja Something seems to have gone wrong with rebasing on `main`

Sorry about that. I've fixed it.

agionoja:
> Let's default to using color when we detect we are in a TTY. This is going to be tricky since we don't want to lose the ability to run logger middleware in browser/service worker environment.
> 
> Something like this could work:
> 
> ```ts
> const useColor = typeof process !== 'undefined' &&
>   process.stdout.isTTY &&
>   process.env.NO_COLOR == null
> 
> function colorize(text: string, color: string) {
>   return useColor ? `${color}${text}${reset}` : text
> }
> ```

I've done as you suggested. Now we only use colors in supported envs and if the user explicitly uses the colors option.

MichaelDeBoey:
@agionoja Could you please rebase this PR onto latest `main` & resolve conflicts?

agionoja:
> @agionoja Could you please rebase this PR onto latest `main` & resolve conflicts?

done

.gitignore
```diff
@@ -6,5 +6,6 @@ reference/
 .tmp/
 
 .env
+.idea
```
MichaelDeBoey: This should be in your global `.ignore` instead

https://docs.github.com/en/get-started/git-basics/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer


```suggestion
```

packages/logger-middleware/src/lib/logger.ts
```diff
@@ -28,7 +31,7 @@ export interface LoggerOptions {
    * - `%url` - The full URL of the request
    * - `%userAgent` - The `User-Agent` header of the request
    *
-   * @default '[%date] %method %path %status %contentLength'
+   * @default '[%date] %method %path %status %durationPretty %contentLengthPretty'
```
MichaelDeBoey: This is technically a breaking change if people would be using the default format to check things out themselves

MichaelDeBoey (commented):
Both are fine for me, but I'm not sure how @mjackson thinks of having multiple features in 1 PR or if he wants it into separate PRs

packages/logger-middleware/src/lib/logger.ts
agionoja: True. Would you suggest adding a preset option that uses the pretty tokens but falls back to the default format? I considered adding a preset, but I was concerned it might overlap with the format option.

.gitignore
agionoja: Thanks a lot, didn't know about that. Will fix in the next request.

packages/logger-middleware/src/lib/logger.ts
MichaelDeBoey: I'll leave it to @mjackson if he's fine with doing this in a minor release or if he want to create a major release for it

agionoja: I've restored the git ignore and will  wait for @mjackson's feedback.

mjackson:
Superseded by #11266.

PR #10864, opened 2025-11-26
Use `.changes` directories to drive releases

markdalgleish:
See the docs for more details. This is obviously inspired by changesets, but with some differences that make it closer to the existing flow in main.

### DIFF
--- a/packages/fetch-router/.changes/minor.remove-build-request-handler.md
+++ b/packages/fetch-router/.changes/minor.remove-build-request-handler.md
@@ -1,0 +1,2 @@
+BREAKING CHANGE: Remove `BuildRequestHandler` type. Use `RequestHandler` type directly instead.
+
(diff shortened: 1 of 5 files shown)


packages/async-context-middleware/.changes/README.md
```diff
@@ -0,0 +1,23 @@
+# Changes Directory
```
pcattori, line 1: I forget how git deals with symbolic links, but might be less repetitive to have this README at the root of the repo somewhere and then just link to it in each `<package>/.changes/README.md` instead of duplicating the contents, esp. if the contents are intended to stay in sync.

Though if you are trying to simulate our changes flow as something that could be 3rd party, then sym links make less sense. But still if contents are always the same, might be nicer to generate the README and gitignore it

markdalgleish: I've opted to just simplify these readmes to a one-liner.

PR #10870, opened 2025-11-29
chore: shared TS config

jacob-ebey:


config/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "private": true,
+  "name": "@remix-run/config",
```
mjackson, line 3: Let's call this package `@remix-run/tsconfig` and keep it in the `packages` dir (even tho it's `private`) like everything else.

config/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": ["@tsconfig/node22/tsconfig.json", "@tsconfig/node-ts/tsconfig.json"],
```
mjackson, line 2: Unless you feel strongly about this, I don't see a ton of value in extending these configs. They are small, and we have gotten this far without them. Let's just copy out the pieces we need and own this piece ourselves. I think we already had most of them in our own tsconfigs.

pnpm-workspace.yaml
```diff
@@ -6,6 +6,7 @@ catalog:
   typescript: ^5.9.3
 
 packages:
+  - config
```
mjackson, line 9: Let's keep this new package in `packages/tsconfig` instead of a new root directory.

mjackson (commented):
I'm not opposed to consolidating our tsconfigs. I'd just like to own it for a little while longer instead of introducing deps on `@tsconfig/*` packages for now.

PR #10883, opened 2025-12-09
fix(node-fetch-server): use `:authority` header to set URL of http/2 requests

phiggins:
Hello :wave:,

A react-router 7 app that was switched to use SSL for local development introduced a bug where the host of the `request` URL was always `localhost` with no port. I believe it is the same root cause as https://github.com/remix-run/react-router/issues/14010 however the behavior in the latest release differs from that of the description in that issue.

The problem appears to be that this library only considers the `Host` header. [RFC 7540](https://httpwg.org/specs/rfc7540.html#HttpRequest) states:
> Clients that generate HTTP/2 requests directly SHOULD use the :authority pseudo-header field instead of the Host header field. 

This issue can be demonstrated with the http2 demo included in this repo. When it is updated to output the URL like so:
```
diff --git a/packages/node-fetch-server/demos/http2/server.js b/packages/node-fetch-server/demos/http2/server.js
index d5847aa72..c88ff3f5a 100644
--- a/packages/node-fetch-server/demos/http2/server.js
+++ b/packages/node-fetch-server/demos/http2/server.js
@@ -20,7 +20,7 @@ server.on(
     let url = new URL(request.url)

     if (url.pathname === '/') {
-      return new Response('Hello HTTP/2!', {
+      return new Response(`Hello HTTP/2! ${url}\n`, {
         headers: {
           'Content-Type': 'text/plain',
         },
```

On the main branch this is the output:
```
$ curl --cacert server.crt https://localhost:44100
Hello HTTP/2! https://localhost/
```

With this PR this is the output:
```
$ curl --cacert server.crt https://localhost:44100
Hello HTTP/2! https://localhost:44100/

```

### DIFF
--- a/packages/node-fetch-server/CHANGELOG.md
+++ b/packages/node-fetch-server/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This is the changelog for [`node-fetch-server`](https://github.com/remix-run/remix/tree/main/packages/node-fetch-server). It follows [semantic versioning](https://semver.org/).
 
+## Unreleased
+
+- Use `:authority` header to set URL of http/2 requests.
+
 ## v0.12.0 (2025-11-04)
 
 - Use `tsc` directly instead of `esbuild` to build the package. This means modules in the `dist` directory now mirror the layout of modules in the `src` directory.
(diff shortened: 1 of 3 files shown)


mjackson:
Thanks for catching this, @phiggins. We will get this released soon.

PR #10884, opened 2025-12-10
wip: `route-pattern` trie + benchmarks

pcattori:
- [x] parsing a flat AST designed for generation of variants and trie branches
- [x] generate variants from AST
- [x] generate `RegExp` from AST
- [x] bench AST parsing
- [x] create full `RoutePattern` from a set of ASTs (one for each part like protocol, hostname, pathname, etc.)
- [x] create segment-based trie from AST variants
- [x] trie returns resolved params
- [x] Feature: ranking
- [x] `Matcher.match`
- [x] `Matcher.matchAll`
- [x] Feature: `.join`
- [x] Feature: port matching
- [x] Feature: matching on search constraints
- [ ] Feature: array matcher

Screenshot 2025-12-09 at 10 47 46 PM


packages/route-pattern/src/lib2/route-pattern/parse.test.ts
```diff
@@ -0,0 +1,68 @@
+import { describe, expect, it } from 'vitest'
```
mjackson, line 1: Any reason we can't use `node:test` here? I'd prefer not to use two different test tools if we can avoid it.

packages/route-pattern/src/lib2/matchers/trie/trie.test.ts
```diff
@@ -0,0 +1,536 @@
...
+      trie.insert(staticPattern, staticPattern)
+      trie.insert(dynamicPattern, dynamicPattern)
+
+      let matches = searchAll(trie, new URL('https://example.com/users/@admin'))
+      expect(matches.length).toBe(2)
+
+      let staticMatch = matches.find((m) => m.data === staticPattern)
+      let dynamicMatch = matches.find((m) => m.data === dynamicPattern)
+
+      expect(staticMatch).toStrictEqual({
```
mjackson, line 375: Instead of asserting the actual rank values (which feels like testing our internals) a better test would be to assert the order of the patterns that are returned, i.e. `expect(matches).toEqual([staticMatch, dynamicMatch])`. This keeps our testing to public API surface and is also instructive when someone scans the tests to figure out how things work.

mjackson (commented):
This is looking great!

packages/route-pattern/src/lib2/route-pattern/parse.test.ts
pcattori: Yea I was just iterating quickly and not paying attention to the exact testing setup. Happy to use `node:test` once this is ready to land.

packages/route-pattern/src/lib2/matchers/trie/trie.test.ts
pcattori: Totally agree, was just using a test to debug and wanted to test some internals temporarily while did that.

My plan is to rewrite the tests much more deliberately at the end once I have a full picture of the design/implementation in my head and can codify that into a spec/test cases for that spec.

pcattori:
Speed up from ["optimize simple dynamic segments" commit](https://github.com/remix-run/remix/pull/10884/changes/fe4ef29ce8019c46577a24a0c1d1c92dad7d7b66)

image

pcattori:
Speed up from ["optimize simple dynamic segments" commit](https://github.com/remix-run/remix/pull/10884/changes/7305bf51e47d02f9c0d6414e7591edb07fffbfa9)

Screenshot 2025-12-17 at 6 51 47 PM

pcattori:
superseded by https://github.com/remix-run/remix/pull/10919

PR #10886, opened 2025-12-10
feat(sse): introduce sse package

gyx1000:
Hello Remix,

I don't know if you plan to provide packages for this kind of use case, and if not, feel free to close the PR without comment.

I needed to manage SSE in my projects on @remix-run/fetch-router, so I created this.

I took the liberty of modifying the @markdalgleish demo 🙏 to introduce the sse package. 

To simplify things even further, it would be useful to add a function to `@remix-run/response`  that accepts a ReadableStream and provides the correct headers to protect against reverse proxies, etc.

Something like `createSseResponse(stream);`

### DIFF
--- a/packages/sse/src/lib/session.test.ts
+++ b/packages/sse/src/lib/session.test.ts
@@ -119,6 +119,7 @@ describe('SSE Session', async () => {
       assert.equal(body, ': comment1\n\nretry: 2000\n\n: comment2\n\n')
     })
 
+    // TODO: find a better way to handle this test
     it('keep alive with specified value', async () => {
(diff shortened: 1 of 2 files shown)


mjackson:
Thanks for the PR @gyx1000. I looked through the tests, but I'm still not sure what problem you're trying to solve that you can't already solve with the tools you already have. Maybe if you filled out the README it would help me grok it?

gyx1000:
Hello @mjackson,
thank you for your quick response. Indeed, this does not add any value to Remix apart from managing the keep-alive, which would be “hidden.” I am closing the PR.

PR #10887, opened 2025-12-11
Add initial documentation scripts

brophdawg11:
This PR adds initial markdown file documentation generation via `typedoc` to the repo in a `docs/generate.ts file` which can be run via `pnpm run docs`.  This will run typedoc across the whole repo, and use it's output to generate markdown files in `docs/api` for all exported functions, classes, interfaces, and types.

`pnpm run docs:serve` will launch a little vibe-coded app to browse the docs, but that's not necessarily expected to survive into the docs final form.

A few source code files have had JSDocs updates to show off stuff like `{@link ...}` and `@alias` usage.

a few notes:
* I haven't added any form of `remix` package in this PR, but the documentation scripts were written with a local stub package and will detect/prefer `remix` re-exports over `@remix-run/*` exports, so when we do add that package it should wire up without any major changes to the docs code.
* In a follow up PR, I'll add the `eslint-plugin-jsdoc` that will help ensure our source code has the stuff we expect for valid doc generation

### DIFF
--- a/packages/fetch-router/src/lib/router.test.ts
+++ b/packages/fetch-router/src/lib/router.test.ts
@@ -916,7 +916,6 @@ describe('custom matcher', () => {
     }
 
     let customMatcher = new CustomMatcher()
-    // @ts-expect-error
     let router = createRouter({ matcher: customMatcher })
     router.get('/', () => new Response('Home'))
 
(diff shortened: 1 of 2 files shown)


packages/fetch-router/src/lib/router.test.ts
```diff
@@ -915,6 +915,7 @@ describe('custom matcher', () => {
     }
 
     let customMatcher = new CustomMatcher()
+    // @ts-expect-error
```
brophdawg11: typedoc wouldn't build without these errors being ignored

mjackson: These errors should be fixed now. I added typechecking to all our demos so they run along with the normal typecheck of our actual source code.

PR #10888, opened 2025-12-12
Setup eslint-plugin-jsdocs

brophdawg11:
Stacked on https://github.com/remix-run/remix/pull/10887

Since JSDocs will be the source of our public documentation, this will help ensure that we're not forgetting to document aspects of them.  Plenty of room for tweaking of the rules but this should be a good base to get started.

Anything not easily handled via the `eslint` configs can be checked/asserted in our docs generation script with more granularity to provide another way for CI to detect missing docs.

The code changes in here are mostly just a find/replace for `@return -> @returns` to align on the canonical JSDoc tag (we were about half and half across the repo).


mjackson:
This LGTM. Please go ahead and merge when ready 👍

PR #10889, opened 2025-12-13
feat(headers): sync `SuperHeaders` with native `Headers`

### LINKED ISSUE #10872: [headers] `SuperHeaders` not compatible with Bun's `Response` constructor
Hello, Remix Team!

I was running a server using Bun with the compression middleware passed to `createRouter` from `@remix-run/fetch-router`. The server would start and handle requests normally, but browsers couldn't determine what content the server was returning. No content-type, no content-encoding, essentially no headers were visible.
The same code worked perfectly in Node.js, but in Bun all response headers were missing when compression middleware was applied. After debugging, I traced the issue to [`SuperHeaders` being passed directly to Bun's `Response` constructor](https://github.com/remix-run/remix/blob/main/packages/response/src/lib/compress.ts#L87-L136) in the `@remix-run/response` compression module.

## Issue

When using `SuperHeaders` with Bun's Response constructor, the headers are not properly recognized by Bun's runtime, causing the browser to not receive any headers added to `SuperHeaders` instance.

**Environment:**
Bun runtime (version: 1.3.3)

### Minimal Reproduction

```typescript
import { SuperHeaders } from '@remix-run/headers'

const server = Bun.serve({
  port: 44100,
  async fetch(request) {
    try {
      // This sample works fine if you replace `SuperHeaders` with `Headers`.
      const headers = new SuperHeaders(request.headers)
      
      headers.append('Content-Type', 'text/plain;charset=utf-8')
      headers.append('X-Test', 'Bun test')
      let response = new Response('Hello, world!', { headers })
      
      console.log(response.headers.get('X-Test')) // null
      console.log(response.headers.get('Content-Type')) // null
      
      return response 
    } catch (error) {
      console.error(error)
      return new Response('Internal Server Error', { status: 500 })
    }
  },
})
```

## Workaround

The only way we got this working was by implementing a helper function to convert `SuperHeaders` into native `Headers`:

```typescript
function toNativeHeaders(superHeaders: SuperHeaders): Headers {
  let headers = new Headers()
  for (let [key, value] of superHeaders) {
    headers.append(key, value)
  }
  return headers
}
```

alexandredev3:
This PR synchronizes `SuperHeaders` with the parent `Headers` class to ensure compatibility with Bun.

Previously, `SuperHeaders` stored data only in its internal `#map` without updating the parent `Headers` state. As a result, when running on Bun, the Response constructor would receive empty headers.

This appears to be because Bun accesses the parent `Headers` class's internal storage directly, bypassing overridden methods. Other runtimes, such as Node.js, seem to rely on JavaScript-level methods like `entries()`, which is why they behaved correctly.

The most reliable solution found was to explicitly call methods like `super.append`, `super.set`, and `super.delete` alongside the internal updates to make Headers state visible to Bun.

resolves #10872

### DIFF
--- a/packages/headers/src/lib/super-headers.test.ts
+++ b/packages/headers/src/lib/super-headers.test.ts
@@ -166,7 +166,7 @@ describe('SuperHeaders', () => {
     assert.equal(headers.toString(), 'Content-Type: text/plain')
   })
 
-  describe('native Headers synchronization', { only: true }, () => {
+  describe('native Headers synchronization', () => {
     it('synchronizes set() with native Headers', () => {
       let headers = new SuperHeaders()
 
@@ -176,7 +176,7 @@ describe('SuperHeaders', () => {
       assert.equal(new Headers(headers).get('Content-Type'), 'text/plain')
     })
 
-    it('synchronizes append() with native Headers', { only: true }, () => {
+    it('synchronizes append() with native Headers', () => {
       let headers = new SuperHeaders()
       headers.append('X-Custom', 'value1')
       headers.append('X-Custom', 'value2')


markdalgleish:
Thanks for raising this issue. I further explored what a complete solution would look like here: https://github.com/remix-run/remix/pull/10904. This most notably adds support for detecting mutations to header property accessors, e.g. `headers.allow = ['GET', 'POST', 'PUT']`.

As a result, we've opted instead to simplify the model, avoiding subclassing `Headers` and other native classes in order to avoid this category of issue entirely: https://github.com/remix-run/remix/pull/10911

PR #10892, opened 2025-12-15
docs(form-data-parser): update Node.js demo

nathanjmorton:
i just changed LocalFileStorage to createFsFileStorage


mjackson:
Thanks for catching this @nathanjmorton. I'm going to add typechecking to all our demos so this kind of stuff doesn't slip through in the future.

PR #10893, opened 2025-12-15
feat(file-storage): add AWS S3 support

nathanjmorton:
added s3 backend and a test suite with minio setup on docker (assumes docker is installed, script automatically sets up everything and tears down after test)

### DIFF
--- a/packages/file-storage/package.json
+++ b/packages/file-storage/package.json
@@ -22,6 +22,7 @@
     ".": "./src/index.ts",
     "./fs": "./src/fs.ts",
     "./memory": "./src/memory.ts",
+    "./s3": "./src/s3.ts",
     "./package.json": "./package.json"
   },
   "publishConfig": {
@@ -37,6 +38,10 @@
       "./memory": {
         "types": "./dist/memory.d.ts",
         "default": "./dist/memory.js"
+      },
+      "./s3": {
+        "types": "./dist/s3.d.ts",
+        "default": "./dist/s3.js"
       },
       "./package.json": "./package.json"
     }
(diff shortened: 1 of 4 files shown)


packages/form-data-parser/demos/node/server.js
```diff
@@ -3,7 +3,7 @@ import * as http from 'node:http'
 import * as os from 'node:os'
 import * as path from 'node:path'
 
-import { LocalFileStorage } from '@remix-run/file-storage/local'
+import { createFsFileStorage } from '@remix-run/file-storage/fs'
```
MichaelDeBoey, line 6: Let's revert these changes as they're already in #10892 


```suggestion
import { LocalFileStorage } from '@remix-run/file-storage/local'
```

packages/form-data-parser/demos/node/server.js
```diff
@@ -16,7 +16,8 @@ const PORT = 44100
 const oneMb = 1024 * 1024
 const maxFileSize = 10 * oneMb
 
-const fileStorage = new LocalFileStorage(await fsp.mkdtemp(path.join(os.tmpdir(), 'uploads-')))
+// @ts-ignore
+const fileStorage = new createFsFileStorage(await fsp.mkdtemp(path.join(os.tmpdir(), 'uploads-')))
```
MichaelDeBoey:
```suggestion
const fileStorage = new LocalFileStorage(await fsp.mkdtemp(path.join(os.tmpdir(), 'uploads-')))
```

MichaelDeBoey (changes requested):

packages/form-data-parser/demos/node/server.js
nathanjmorton: changes reverted

packages/form-data-parser/demos/node/server.js
nathanjmorton: i removed this file from the commit by rebasing my feature branch onto (this) upstream/main and force pushing with lease

mjackson:
Superseded by #11097.

PR #10894, opened 2025-12-15
Add remix package

brophdawg11:
Stacked on https://github.com/remix-run/remix/pull/10888

Adds a top level `remix` package that is auto-generated as an umbrella package over all `@remix-run/*` packages in the repo.  

* Auto-generated files are one-liner `export * from '@remix-run/...';` files in `packages/remix/src/lib`
* Every sub-package and it's sub-exports becomes a sub-export in remix:
  * `@remix-run/file-storage` -> `remix/file-storage`
  * `@remix-run/file-storage/fs` -> `remix/file-storage/fs`
* `pnpm generate-remix` will generate the remix package based on the current state of the repo
* A github workflow has been added that can automatically re-build ~~on PRs and~~ pushes to `main`
* The root `remix/src/index.ts` file is manually authored to control the APIs exported from the top-level `remix` package
  * These exports are currently determined by the APIs needed for the 3 root `demos/`, but I'm sure we'll want to add more as we start taking a closer look at the entire API surface and building more demos
* All 3 root `demos/` have been converted to use the top-level package and remove `@remix-run/*` packages from their `dependencies`


packages/remix/src/index.ts
```diff
@@ -0,0 +1,46 @@
+export { asyncContext, getContext } from './lib/async-context-middleware'
```
brophdawg11: Export all APIs used by our root demos

packages/remix/src/index-client.ts
```diff
@@ -0,0 +1,27 @@
+export { Cookie, createCookie } from './lib/cookie'
```
brophdawg11: Export the subset client-side friendly APIs from `index.ts`.  To be used with `esbuild --conditions=client` when building assets for the browser

packages/remix/package.json
```diff
@@ -0,0 +1,260 @@
...
+    "README.md",
+    "dist",
+    "src"
+  ],
+  "type": "module",
+  "exports": {
+    ".": {
+      "client": "./src/index-client.ts",
+      "default": "./src/index.ts"
+    },
```
brophdawg11: Use a `client` condition to provide a browser-friendly build

packages/remix/src/index-platform.ts
```diff
@@ -0,0 +1,28 @@
+// This file should only export "server" exports that are implemented by your
+// platform and thus require built-in implementations such as:
+//  - `node:async_hooks`
+//  - `node:fs`
+//  - `node:zlib`
+//
+// All other APIs based on core Javascript/Web Fetch API should be exported
+// from the root remix package in `index.ts`
+
+export { asyncContext, getContext } from './lib/async-context-middleware'
```
mjackson: Just a note on import/export paths from files (not packages): we always use the file extension to avoid ambiguity. So this should be:


```suggestion
export { asyncContext, getContext } from './lib/async-context-middleware.ts'
```

Maybe we want to add an eslint rule that catches this?

packages/remix/src/index.ts
```diff
@@ -0,0 +1,31 @@
...
+export { Cookie, createCookie } from './lib/cookie'
+export {
+  createRouter,
+  createStorageKey,
+  del,
+  form,
+  get,
+  post,
+  put,
+  resources,
```
mjackson, line 30: We should have `resource` here as well, along with `head`, `options`, and `patch`. I'm thinking we should also have the longform names, `createGetRoute`, etc.

mjackson (commented):
This PR looks pretty good. The barrel files (`index.ts` and `index-platform.ts`) will still need some work to be complete, but we can continue to add things they are missing over time as we begin to use the `remix` package in earnest. This is a good start.

packages/remix/src/index-client.ts
mjackson: Is `client` a special word, or a word that you picked? I'd personally prefer to use `browser` if it's up to us, especially since React blurred the meaning of "client" with RSC.

demos/bookstore/app/utils/context.ts
```diff
@@ -1,5 +1,5 @@
-import { createStorageKey } from '@remix-run/fetch-router'
-import { getContext } from '@remix-run/async-context-middleware'
+import { createStorageKey } from 'remix'
+import { getContext } from 'remix/platform'
```
mjackson: Something feels off to me about this. I have to know that `getContext` relies on `node:async_hooks` internally so that I know to get it from `remix/platform`. Maybe it's better to just import directly from `remix/async-context-middleware` and ditch the whole idea of `remix/platform` entirely.

demos/bookstore/app/middleware/auth.ts
```diff
@@ -1,5 +1,5 @@
-import type { Middleware, Route } from '@remix-run/fetch-router'
-import { createRedirectResponse as redirect } from '@remix-run/response/redirect'
+import type { Middleware, Route } from 'remix'
+import { createRedirectResponse as redirect } from 'remix'
```
mjackson: This one is so common we'll probably want to have a `redirect` shorthand that people can import instead of renaming it every time.

packages/remix/src/index-client.ts
brophdawg11: I ended up switching this based on the all hands discussion to be a `remix/platform` sub-export, so there's no `client` condition anymore

packages/remix/src/index-platform.ts
brophdawg11: oh interesting - I thought that would already fail at build time and didn't even notice the LLM left them out.  I'll look into a way to catch this automatically 👍

demos/bookstore/app/utils/context.ts
brophdawg11: yeah - I think thats a good call.  The sub-export is just an added layer of indirection when it's still a single dependency.

demos/bookstore/app/middleware/auth.ts
brophdawg11: Do you think we should do the same for `createHtmlResponse` and `createFileResponse`?  `html` would conflict at a global level with the `html` string helper so if we chose to export from the top-level we'd have to choose one, but they'd both be fine as exports from their individual sub-exports.

packages/remix/src/index-platform.ts
brophdawg11: Added a lint rule in https://github.com/remix-run/remix/pull/10894/commits/625b6397f4de830bd41dd49f3f97582c84044b5b

demos/bookstore/app/utils/context.ts
brophdawg11: https://github.com/remix-run/remix/pull/10894/commits/4a74dfe27c45ec312ca3af6a7ec9ccd99101babf

packages/remix/src/index.ts
brophdawg11: https://github.com/remix-run/remix/pull/10894/commits/1210a547c786bba678f9408c7932a98175d93367

demos/bookstore/app/middleware/auth.ts
brophdawg11: https://github.com/remix-run/remix/pull/10894/commits/1210a547c786bba678f9408c7932a98175d93367

mjackson: I'm not sure about the `html` helper. I want to see how it feels once we start using components more heavily in the demos and relying less on the html-template library. I just know for now that I want to be able to import `redirect` directly, or I'm always going to do this rename.

demos/bookstore/app/router.ts
```diff
@@ -1,11 +1,8 @@
-import { createRouter } from '@remix-run/fetch-router'
-import { asyncContext } from '@remix-run/async-context-middleware'
-import { compression } from '@remix-run/compression-middleware'
-import { formData } from '@remix-run/form-data-middleware'
-import { logger } from '@remix-run/logger-middleware'
-import { methodOverride } from '@remix-run/method-override-middleware'
-import { session } from '@remix-run/session-middleware'
-import { staticFiles } from '@remix-run/static-middleware'
+import { createRouter, formData, logger, methodOverride } from 'remix'
```
mjackson: Let's be consistent with middleware imports and import `logger` and `methodOverride` from `remix/logger-middleware` and `remix/method-override-middleware`.

demos/unpkg/app/directory.ts
```diff
@@ -1,4 +1,4 @@
-import { detectMimeType } from '@remix-run/mime'
+import { detectMimeType } from 'remix'
```
mjackson: This kind of import makes me think we're exporting too much from the `remix` index file. We should get this from `remix/mime`.

packages/remix/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  put,
+  resource,
+  resources,
+  route,
+  type BuildAction,
+  type Controller,
+  type Middleware,
+  type Route,
+} from './lib/fetch-router.ts'
+export { formData } from './lib/form-data-middleware.ts'
```
mjackson: Remove, we should get this from `remix/form-data-middleware`

packages/remix/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+export type { FileUpload } from './lib/form-data-middleware.ts'
+export { Cookie as CookieHeader, SetCookie } from './lib/headers.ts'
+export { html, isSafeHtml, type SafeHtml } from './lib/html-template.ts'
+export { on } from './lib/interaction.ts'
+export { logger } from './lib/logger-middleware.ts'
```
mjackson: Remove, we should get this from 'remix/logger-middleware`

packages/remix/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+export { methodOverride } from './lib/method-override-middleware.ts'
+export { detectMimeType, isCompressibleMimeType } from './lib/mime.ts'
```
mjackson: Remove, we should get this from `remix/mime`

packages/remix/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  type Middleware,
+  type Route,
+} from './lib/fetch-router.ts'
+export { formData } from './lib/form-data-middleware.ts'
+export type { FileUpload } from './lib/form-data-middleware.ts'
+export { Cookie as CookieHeader, SetCookie } from './lib/headers.ts'
+export { html, isSafeHtml, type SafeHtml } from './lib/html-template.ts'
+export { on } from './lib/interaction.ts'
+export { logger } from './lib/logger-middleware.ts'
+export { methodOverride } from './lib/method-override-middleware.ts'
```
mjackson: Remove, we should get this from `remix/method-override-middleware`

packages/remix/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  type BuildAction,
+  type Controller,
+  type Middleware,
+  type Route,
+} from './lib/fetch-router.ts'
+export { formData } from './lib/form-data-middleware.ts'
+export type { FileUpload } from './lib/form-data-middleware.ts'
+export { Cookie as CookieHeader, SetCookie } from './lib/headers.ts'
+export { html, isSafeHtml, type SafeHtml } from './lib/html-template.ts'
+export { on } from './lib/interaction.ts'
```
mjackson: Remove, we should get this from `remix/interaction`

.github/workflows/generate-remix.yaml
```diff
@@ -0,0 +1,86 @@
+# Update the `remix` package by auto-generating from the `@remix-run/*` packages in the repo
+# runs on any pushes/PRs to `main` that touch any `package.json` files since that would
+# potentially alter the sub-exports that need to be reflected in the `remix` package.
+
+# Note: Does not currently run on PRs from forked repos.
+
+name: 💿 Update Remix Package
```
mjackson, line 7: This workflow feels like an additional check we'll want to run before we merge someone's PR, not something we'll want to fix after it's already merged. Would you agree, @brophdawg11?

I'm just thinking, if someone adds a new package or modifies an existing one, and the `remix` package needs to be updated as a result, I'd like to see that change in the PR instead of in an auto-generated commit that runs in CI sometime later. That way we can see right there in the PR how it modifies the `remix` package.

demos/unpkg/app/router.ts
```diff
@@ -1,5 +1,4 @@
-import { createRouter } from '@remix-run/fetch-router'
-import { createRedirectResponse as redirect } from '@remix-run/response/redirect'
+import { createRouter, createRedirectResponse as redirect } from 'remix'
```
mjackson, line 1: We should be able to just call this `redirect` now.

.github/workflows/generate-remix.yaml
brophdawg11: Yeah I agree - check out the comment below.  I had it that way originally and found that pushing back to the branch with `GITHUB_TOKEN` cancelled ongoing CI runs and thus would trigger false positives on the updated commit CI run.  

We just need to setup a PAT and use that token to checkout/commit and we should be able to push back to PR branches without issue.  We'll want to confirm that it works on PRs from forks too once we get it set up.

PR #10904, opened 2025-12-17
WIP: Bun support

markdalgleish:


markdalgleish:
Closing this in favour of a different strategy:
- Remove the `SuperHeaders` class and instead convert `@remix-run/headers` into a set of utils. This requires updating fetch-router so it uses standard `Headers` instead of `SuperHeaders`.
- Make `LazyFile`/`LazyBlob` custom classes that *don't* subclass `File`/`Blob`. Instead, you need to call `.stream()` when you want to use them in a response body. We'll also add `LazyFile.toFile()` and `LazyBlob.toBlob()` helpers.
- Split out charset fixes into a standalone PR.

PR #10905, opened 2025-12-18
Add `charset` to Content-Type for text-based MIME types

markdalgleish:
Adds `; charset=utf-8` to Content-Type headers for text-based MIME types. Note, `text/xml` is excluded since XML has built-in encoding detection via `<?xml encoding="..."?>`

To power this, I've added a couple of utilities to the `mime` package:
- `mimeTypeToContentType(mimeType)` - converts a MIME type to a Content-Type header value, adding charset for text-based types
- `detectContentType(extension)` - like `detectMimeType` but returns a full Content-Type with charset


mjackson:
LGTM! Thanks, @markdalgleish

PR #10907, opened 2025-12-18
fix(component): remove requirement for every element to have props

alexanderson1993:
Remix assumes that props will be an object. That isn't always the case, as I found when using it with `htm`. This removes that requirement and allows props to be nullish.

Reproduction:

```html
<!DOCTYPE html>
<html>
  <head>
    <title>Testing</title>
  </head>
  <body>
    <div id="root" />
    <script type="module">
      import htm from "https://unpkg.com/htm?module";
      import * as Remix from "https://unpkg.com/@remix-run/component?module";
      const html = htm.bind(Remix.createElement);

      function App() {
        let count = 0;

        return () => html`<button
          on=${{
            click: () => {
              count++;
              this.update();
            },
          }}
        >
          Count: ${count}
        </button>`;
      }

      Remix.createRoot(document.getElementById("root")).render(
        html`<${App} />`
      );
    </script>
  </body>
</html>
```

Expected Behavior: it renders correctly and is interactive
Actual behavior: It does not render and throws a `TypeError: null is not an object (evaluating 'props.key')` error on this line.

### DIFF
--- a/packages/component/.changes/patch.optional-props.md
+++ b/packages/component/.changes/patch.optional-props.md
@@ -1,0 +1,3 @@
+Remove requirement for every element to have props
+
+Originally, `@remix/component` assumed that props will be an object, not `null` or `undefined`. This requirement has been removed and allows props to be nullish. This makes it easier to render `@remix/component` apps using alternative JSX templating tools like [`htm`](https://www.npmjs.com/package/htm).


packages/component/src/lib/create-element.ts
```diff
@@ -6,7 +6,7 @@ export function createElement(
   props: Record<string, any>,
```
mjackson: I believe we need to make this optional as well, no?

mjackson (changes requested):
Thanks for the PR, @alexanderson1993. I think we need to adjust the type of the `props` arg as well.

packages/component/src/lib/create-element.ts
alexanderson1993: I suppose that is true.

```suggestion
  props?: Record<string, any>,
```

brookslybrand:
Nice! Will you go ahead and [add a change file](https://github.com/remix-run/remix/blob/main/CONTRIBUTING.md#adding-a-change-file) too @alexanderson1993?

alexanderson1993:
Done! Thanks for the heads up and reviews.

PR #10911, opened 2025-12-19
Remove `SuperHeaders` class, add new `from` header methods

markdalgleish:
This PR removes the `SuperHeaders` class from `@remix-run/headers` in favour of individual header parsing functions and updates all consumers to use the new utils.

- `@remix-run/headers` no longer exports `SuperHeaders`/`Headers` class. Instead, it now exports individual `.parse()` methods for each header class (e.g. `ContentType.parse()`, `Accept.parse()`, etc.) along with `parseRawHeaders()` and `stringifyRawHeaders()` utilities.
- `@remix-run/fetch-router` now returns a plain `Headers` instance from `RequestContext.headers` instead of `SuperHeaders`, and has dropped the `@remix-run/headers` peer dependency.
- `@remix-run/multipart-parser` has been updated to use the new header parsing functions.
- `@remix-run/response` has been updated to use the new header parsing functions.

### DIFF
--- a/packages/multipart-parser/src/lib/multipart.ts
+++ b/packages/multipart-parser/src/lib/multipart.ts
@@ -320,14 +320,11 @@ export class MultipartParser {
    *
    * Note: This will throw if the multipart message is incomplete or
    * wasn't properly terminated.
-   *
-   * @returns void
    */
   finish(): void {
     if (this.#state !== MultipartParserStateDone) {
       throw new MultipartParseError('Multipart stream not finished')
     }
-    return undefined
   }
 }
 


MichaelDeBoey (commented):
@markdalgleish If this PR is created to fix #10872, an alternative could be to go with @alexandredev3's solution from #10889

markdalgleish:
@MichaelDeBoey Just repeating my comment here from https://github.com/remix-run/remix/pull/10913 for anyone who missed it.

We were aware of this and explored what a complete solution would look like here: https://github.com/remix-run/remix/pull/10904. We've opted instead to simplify the model and avoid this category of issue entirely.

mjackson:
Thanks for putting together this PR, @markdalgleish. I was thinking about this PR over the break, and it occurred to me that we could use `from` methods to possibly clean up the API a bit.

Instead of exporting both the header class and a `parse` method (e.g. `AcceptEncoding` *and* `parseAcceptEncoding`) we could just add a static `AcceptEncoding.from(init: AcceptEncodingInit)` method that would do the work the constructor is currently doing. I like this for a few reasons:

- We don't have a bunch of `parse` methods, instead we just `export` a single class for each header (and an init type)
- The `parse` method was always returning an `instanceof` its associated class, which is what `from` is traditionally used for in JS array methods (e.g. `Array.from()`, `Uint8Array.from()`, etc.), so this feels a little more "native"

We can also preserve the ability to pass an optional `init` value to the constructor by using the `from()` method internally if it's present.

```ts
class AcceptEncoding {
  static from(init: AcceptEncodingInit) {
    let header = new AcceptEncoding()
    // iterate over the init object and use set()/append() methods to add values...
    return header
  }

  constructor(init?: AcceptEncodingInit) {
    if (init) return AcceptEncoding.from(init)
    this.#map = new Map()
  }
}
```

What do you think?

markdalgleish:
@mjackson Yeah, that feels cleaner. I've pushed an update moving to this pattern.

packages/fetch-router/src/lib/request-context.ts
```diff
@@ -20,6 +19,7 @@ export class RequestContext<
    * @param request The incoming request
    */
   constructor(request: Request) {
+    this.headers = request.headers
```
mjackson: Let's do this and add a test to make sure `context.headers` is always mutable.

```suggestion
    this.headers = new Headers(request.headers)
```

packages/fetch-router/.changes/minor.plain-headers.md
```diff
@@ -0,0 +1,20 @@
...
+router.get('/api/users', (context) => {
+  let acceptsJson = context.headers.accept.accepts('application/json')
+  // ...
+})
+
+// After:
+import { parseAccept } from '@remix-run/headers'
+
+router.get('/api/users', (context) => {
+  let accept = parseAccept(context.headers.get('accept'))
```
mjackson:
```suggestion
import { Accept } from '@remix-run/headers'

router.get('/api/users', (context) => {
  let accept = Accept.from(context.headers.get('accept'))
```

packages/headers/.changes/minor.remove-super-headers.md
```diff
@@ -0,0 +1,65 @@
...
+- `IfNoneMatch.from()`
+- `IfRange.from()`
+- `Range.from()`
+- `SetCookie.from()`
+- `Vary.from()`
+
+New raw header utilities added:
+
+- `parseRawHeaders()`
+- `stringifyRawHeaders()`
```
mjackson: Let's call these `parse` and `stringify`. Users can rename in the `import` if they want.

mjackson (changes requested):
Almost there, just a few small tweaks.

packages/fetch-router/.changes/minor.plain-headers.md
markdalgleish: Ah, thanks for catching this.

peter-skillmasters:
I was using SuperHeaders to help construct responses - the API for which was very nice imo:

```ts
return new Response(buffer, new SuperHeaders({
  contentDisposition: { type: "attachment", filename },
  contentType: "application/pdf",
  cacheControl: { private: true, maxAge: 86400 }
})
```

Is there a way to do this after these changes? If not, what do you think about adding a util that only builds a native Headers object without being a subclass, for example:

```ts
return new Response(buffer, buildHeaders({
  contentDisposition: { type: "attachment", filename },
  contentType: "application/pdf",
  cacheControl: { private: true, maxAge: 86400 }
})
```

If you like the idea, I don't mind pushing a PR for review

PR #10913, opened 2025-12-19
Convert `LazyFile`/`LazyBlob` to standalone classes

markdalgleish:
This PR removes subclassing from `LazyFile`/`LazyBlob` to fix compatibility issues with Bun's native optimizations and updates all consumers to use the new APIs.

- `@remix-run/lazy-file` no longer extends native `File`/`Blob`. Instead, `LazyFile` and `LazyBlob` implement the same interface as standalone classes. Users must now call `.stream()` for `Response` bodies or `.toFile()`/`.toBlob()` for APIs requiring native types (e.g. `FormData`).
- `@remix-run/fs` has renamed `openFile()` to `openLazyFile()` and removed the `getFile()` alias to explicitly reflect that it returns a `LazyFile`, not a native `File`.
- `@remix-run/response` now accepts any file-like object via a generic `FileLike` interface, supporting both native `File` and `LazyFile`.
- `@remix-run/file-storage` and `@remix-run/static-middleware` have been updated to use the new `openLazyFile()` API.


MichaelDeBoey:
@markdalgleish If this PR is created to fix what @kentcdodds mentioned in https://github.com/remix-run/remix/issues/10872#issuecomment-3707276428, an alternative could be to go with something like @alexandredev3's solution from https://github.com/remix-run/remix/pull/10889

markdalgleish:
@MichaelDeBoey Thanks for the heads up. We were aware of this and explored what a complete solution would look like here: https://github.com/remix-run/remix/pull/10904. We've opted instead to simplify the model and avoid this category of issue entirely.

packages/fs/README.md
```diff
@@ -1,12 +1,12 @@
 # fs
 
-Filesystem utilities using the Web File API.
+Lazy, streaming filesystem utilities for JavaScript.
 
-This package provides utilities for working with files on the local filesystem using the Web [File API](https://developer.mozilla.org/en-US/docs/Web/API/File).
+This package provides utilities for working with files on the local filesystem using the [`LazyFile`](https://github.com/remix-run/remix/tree/main/packages/lazy-file)/ [Web `File` API](https://developer.mozilla.org/en-US/docs/Web/API/File) API.
```
mjackson:
```suggestion
This package provides utilities for working with files on the local filesystem using the [`LazyFile`](https://github.com/remix-run/remix/tree/main/packages/lazy-file)/ native [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File) API.
```

packages/lazy-file/src/lib/lazy-file.test.ts
```diff
@@ -1,7 +1,11 @@
 import * as assert from 'node:assert/strict'
 import { describe, it } from 'node:test'
 
-import { type LazyContent, LazyFile } from './lazy-file.ts'
+import { type LazyContent, LazyBlob, LazyFile } from './lazy-file.ts'
+
+// Type assertions: ensure LazyBlob and LazyFile implement all native Blob/File APIs.
+null as unknown as LazyBlob satisfies Record<keyof Blob, unknown>
+null as unknown as LazyFile satisfies Record<keyof File, unknown>
```
mjackson, line 8: Nice 👍

AGENTS.md
```diff
@@ -26,7 +26,7 @@
 - **Classes**: Use native fields (omit `public`), `#private` for private members (no TypeScript accessibility modifiers)
 - **Formatting**: Prettier (printWidth: 100, no semicolons, single quotes, spaces not tabs)
 - **TypeScript**: Strict mode, ESNext target, ES2022 modules, bundler resolution, verbatimModuleSyntax
-- **Generics**: Use descriptive lowercase names for type parameters (e.g., `source`, `method`, `pattern`) instead of single uppercase letters like `T`, `P`, or `K`
+- **Generics**: Use descriptive lowercase names for type parameters (e.g., `source`, `method`, `pattern`) instead of single uppercase letters like `T`, `P`, or `K`. The type parameter and variable should ideally share the same name (e.g., `<file extends FileLike>(file: file)`)
```
mjackson: I've found that "rename symbol" has a hard time when the generic and function parameter share the same name. We may not want to tell LLMs to do this yet.

mjackson (changes requested):
Some small changes, but overall this is an excellent PR. Thanks, @markdalgleish!

PR #10923, opened 2026-01-05
Update to node 24 and update CI to use the engines version

brophdawg11:
Dedup the places we define the node version by updating `package.json` `engines.node` to 24 (active LTS) and updating CI scripts to read the node version from that field.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -25,8 +25,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 24
-          registry-url: https://registry.npmjs.org
+          node-version-file: 'package.json'
           cache: pnpm
 
       - name: Install dependencies
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -25,7 +25,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 24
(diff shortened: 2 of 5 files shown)


.github/workflows/publish.yaml
```diff
@@ -25,8 +25,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 24
-          registry-url: https://registry.npmjs.org
```
brophdawg11, line 29: I don't think `registry-url` should be necessary?

mjackson: I think you're probably right.

MichaelDeBoey:
@brophdawg11 I know this is only changing tests, but the only `engines.node` is in the root `package.json`, so isn't this kinda a breaking change for _all_ packages?

brophdawg11:
The root `package.json` only applies to the monorepo for local development.  None of the published packages `package.json` files have an `engines` field at the moment.

MichaelDeBoey:
@brophdawg11 That's indeed what I noticed

In that case, most people look at the tests to see which Node versions are supported
So for people doing so, this _kinda_ is a breaking change

brophdawg11:
> In that case, most people look at the tests to see which Node versions are supported

nah - that's their own subjective opinion.  The root monorepo version does not indicate in any official capacity the supported versions for individually published packages.  I expect we'll provide formal guidance on supported versions closer to an alpha/RC release.

PR #10924, opened 2026-01-06
ci: run lint and type checks, automatically format

markdalgleish:
We currently have ESLint and Prettier configured but aren't running any checks against these in CI, and we're also not type checking anything that isn't being built by tsgo (tests, demos), which also means that some errors and formatting issues have slipped through. This PR adds TS and ESlint checks to our `check` workflow in CI to catch these at PR time, and adds an auto formatting workflow to the main branch like we have for Remix v2.

Note that as part of this, I picked up on fact that both `style.test.ts` and `style.test.tsx` had become duplicates of each other but with some additional tests, so I've consolidated these.

### DIFF
--- a/demos/bookstore/app/utils/uploads.ts
+++ b/demos/bookstore/app/utils/uploads.ts
@@ -1,6 +1,6 @@
 import { dirname, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
-import type { FileUpload } from 'remix'
+import type { FileUpload } from 'remix/form-data-parser'
 import { createFsFileStorage } from 'remix/file-storage/fs'
 
 const __dirname = dirname(fileURLToPath(import.meta.url))


packages/component/src/lib/invariant.ts
```diff
@@ -1,6 +1,9 @@
 /**
  * Use this when framework code is incorrect, indicating an internal bug rather
  * than application code error.
+ *
+ * @param assertion The value to check for truthiness
+ * @param message Optional message to include in the error
```
brophdawg11, line 6: Just FYI - if these JSDoc-related lint errors get annoying on non-public API functions where we don't really need/care to "document" the params, we can also switch to a multi-line comment and not a JSDoc comment (`/*` instead of `/**`) and it won't lint it any longer.

.github/workflows/check.yaml
```diff
@@ -27,5 +27,14 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Check formatting
+        run: pnpm format:check
```
brophdawg11: Instead of failing CI on a PR, would we prefer a format workflow like we have in RR that run against `main` and commits the changes back to `main`?

markdalgleish: Personally, I'm not a fan of the extra commit noise and prefer to keep things clean at the source, but I'm happy to go either way. I've updated formatting to match the v2 flow to keep things consistent.

brophdawg11: We could have it push changes back to the PR branch as well to keep the main history cleaner if we like that better?  I’m just not a fan of pushing up finished work, going away for a bit, and coming back to a failed CI run for something small - especially if it could be auto fixed.

PR #10930, opened 2026-01-07
ci: automate package version flow, publish all packages at once

markdalgleish:
This introduces a Changesets-inspired flow where we get a "Version Packages" PR which is automatically generated by running the existing `pnpm changes:version` script.

This also modifies our publish flow to be based on release commits rather than tags so that packages are published in a single step in topological order (managed by pnpm). The publish step now checks for commits starting with "Release ..." that also have changes to version numbers in `package.json`. It also automatically adds any missing version tags to the commit.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -21,7 +21,6 @@ export default [
       'node_modules/**',
       'reference/**',
       'packages/multipart-parser/demos/deno/**',
-      'packages/remix/src/lib/**',
     ],
   },
   {
--- a/packages/remix/src/lib/file-storage.ts
+++ b/packages/remix/src/lib/file-storage.ts
@@ -1,2 +1,2 @@
 // IMPORTANT: This file is auto-generated, please do not edit manually.
-export * from '@remix-run/file-storage'
+export type * from '@remix-run/file-storage'
--- a/scripts/generate-remix.ts
+++ b/scripts/generate-remix.ts
@@ -8,6 +8,7 @@
  * Run: node docs/generate-remix.ts
  */
 
+import cp from 'node:child_process'
 import fs from 'node:fs/promises'
 import path from 'node:path'
 import url from 'node:url'
@@ -42,8 +43,15 @@ type ExportEntry = {
 
 let { remixRunPackages, allExports } = await getRemixRunPackages()
 
-// Generate source files iun
+// Ensure we have a passing linter before generating code
+lintRemixPackage(false)
+
+// Generate source files
 await generateRemixSourceFiles()
+
+// Run the linter against generated source code with --fix to ensure generated
+// code passes linting
+lintRemixPackage(true)
 
 // Update package.json
 console.log('📦 Updating package.json...')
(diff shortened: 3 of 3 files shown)


scripts/changes-version-pr.ts
```diff
@@ -0,0 +1,137 @@
...
+  } catch {
+    logAndExec(`git checkout ${prBranch}`)
+  }
+
+  // Reset to base branch
+  logAndExec(`git reset --hard origin/${baseBranch}`)
+
+  // Run version command (commits and creates local tags, but we only push the branch)
+  console.log('\nRunning pnpm changes:version...')
+  logAndExec('pnpm changes:version')
```
brophdawg11, line 102: Now that the primary tag generation will be done by `publish.ts`, should we flip the default here and require `pnpm changes:version --tags` if we have to do it manually and want to generate tags?

.github/workflows/changes-version-pr.yaml
```diff
@@ -0,0 +1,40 @@
+name: Changes Version PR
+
+on:
+  push:
+    branches:
+      - main
```
brophdawg11, line 6: Love this flow of just merging all of `main` at a time chosen by us (via PR merge).

I was thinking a bit about how we might be able to do this for hotfix/one-off releases as well (as a follow up).  I think it should be easy enough to update the flow to run on `hotfix-*` branches so we could do something like:

- Dev creates `hotfix-*` branch from prior release tag
- PRs changes to `hotfix-*` branch so that only 1 (or more) packages has changes
- Upon merging, this workflow would also trigger for `hotfix-*` branches
- Would release only the changed package(s)
- Merge `hotfix-*` branch back to `main` manually (in case of conflicts)
  - We would need to ensure that hotfix release commits don't trigger main release workflow, maybe as simple as using a `Hotfix Release...` commit message?

This would also be a feasible option to ship a single package from main as a "hotfix" without releasing all of main:

- `git branch hotfix-branch [old-release-tag]`
- `git checkout hotfix-branch`
- `git checkout pr-branch`
- `git checkout main packages/route-pattern/*`
- Open PR and merge to `hotfix-branch`
- Normal hotfix flow would run and publish the singular changed package

.github/workflows/publish.yaml
```diff
@@ -1,23 +1,62 @@
...
 on:
   push:
-    tags:
-      - '[a-z]+@[0-9]+.[0-9]+.[0-9]+'
-      - '[a-z]+-[a-z]+@[0-9]+.[0-9]+.[0-9]+'
-      - '[a-z]+-[a-z]+-[a-z]+@[0-9]+.[0-9]+.[0-9]+'
-      - '!remix@2.*' # Ignore remix@2.x.x tags
-      - '!create-remix@2.*' # Ignore create-remix@2.x.x tags
+    branches:
+      - main
```
brophdawg11, line 6: We should think about how to handle nightly releases in a follow up - should be able to set up a separate cron job to create a `nightly` branch from `main` and do the version updates there (without a PR) and trigger this workflow from there and then delete the branch after publishing.

And then once that is working, that should be semi-easily extendable to do "prereleases" from a release branch so we can install published packages for any final of testing prior to merging the PR

brophdawg11:
Another follow up thing that occurred to me is we should inline the script from https://github.com/remix-run/release-comment-action to this repo so we can bring that functionality over

scripts/changes-version-pr.ts
markdalgleish: Good idea. Done.

PR #10931, opened 2026-01-07
separate handler registration from router creation

pcattori:
Replace `createRouter` with `createHandlers` so that route/handler registration is guaranteed to happen before router is created. That way matchers can be created with full knowledge of all routes they will ever need to handle, enabling some optimizations based on feature detection.


pcattori:
@mjackson starting to feel like this could be a bit of premature optimization. I was thinking about the design of the matchers in `route-pattern`, and there's nothing in there today that relies on apriori knowledge of all the routes so I think I'd rather keep pushing on the `route-pattern` work and see if this ends up being a bottleneck after perf profiling it.

PR #10934, opened 2026-01-08
Re-enable linting of the generated remix package code

brophdawg11:


packages/remix/src/lib/file-storage.ts
```diff
@@ -1,2 +1,2 @@
 // IMPORTANT: This file is auto-generated, please do not edit manually.
-export * from '@remix-run/file-storage'
+export type * from '@remix-run/file-storage'
```
brophdawg11, line 2: This was the linting error we were running into, but we can address it at generation time with `--fix`

scripts/generate-remix.ts
```diff
@@ -42,9 +43,16 @@ type ExportEntry = {
 
 let { remixRunPackages, allExports } = await getRemixRunPackages()
 
-// Generate source files iun
+// Ensure we have a passing linter before generating code
+lintRemixPackage(false)
```
brophdawg11, line 47: Ensure we're in a passing state before we generate code

scripts/generate-remix.ts
```diff
@@ -42,9 +43,16 @@ type ExportEntry = {
...
+
+// Generate source files
 await generateRemixSourceFiles()
 
+// Run the linter against generated source code with --fix to ensure generated
+// code passes linting
+lintRemixPackage(true)
```
brophdawg11, line 54: Run with `--fix` on generated code

markdalgleish (approved):
This is a really nice way to fix this 👍

PR #10936, opened 2026-01-08
route-pattern: RoutePattern.{parse,join}

pcattori:
Stacked on #10935

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -67,9 +67,13 @@ export default [
           message: "Use native class fields: omit 'public' and use '#private' for private state.",
         },
         {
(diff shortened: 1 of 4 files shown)


eslint.config.js
```diff
@@ -67,10 +67,14 @@ export default [
           message: "Use native class fields: omit 'public' and use '#private' for private state.",
         },
         {
-          selector: 'MethodDefinition[accessibility]',
+          selector: 'MethodDefinition[accessibility]:not([kind="constructor"])',
```
pcattori, line 70: `constructor`s cannot use the `#` prefix, so they are an exception to the rule

PR #10964, opened 2026-01-16
Add script to setup a pnpm-installable branch

brophdawg11:
POC of a way to make a branch "installable" as an alternative to nightly/canary releases adding a ton of noise to npm.  This is primarily intended as a way to do "future" releases from a separate branch including breaking changes before they have landed in `main` so the `remix` package (including those breaking changes) can be tested as a whole.  

Assuming we have a `future` branch with breaking changes, we would run this script and commit the changes to a `future-build` branch and then be able to install via PNPM:

```sh
> pnpm i "github:remix-run/remix#future-build&path:packages/remix"
```

These changes exist in a `future-branch` right now (built from `main`) so the above should work for testing.

Note: This requires `pnpm@9` or later (that's when they added the subpath support)

Demo repo (stackblitz uses pnpm 8): https://github.com/brophdawg11/installable-remix-poc

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -59,4 +59,3 @@ jobs:
         run: |
           git push
           echo "💿 pushed updates: https://github.com/$GITHUB_REPOSITORY/commit/$(git rev-parse HEAD)"
-          echo "new_sha=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
--- a/scripts/setup-installable-branch.ts
+++ b/scripts/setup-installable-branch.ts
@@ -85,6 +85,7 @@ logAndExec(`git commit -a -m "installable build from ${baseBranch} at ${sha}"`)
 
 console.log(
   [
+    '',
     `✅ Done!`,
     '',
     `You can now push the \`${installableBranch}\` branch to GitHub and install via:`,


frolic:
Have y'all tried the PR package service from Stackblitz? It's free and great for doing things like prereleases from a branch without spamming npm: https://github.com/stackblitz-labs/pkg.pr.new

scripts/setup-installable-branch.ts
```diff
@@ -0,0 +1,100 @@
...
+    .filter((line) => !line.trim().startsWith('dist'))
+    .join('\n')
+  await fsp.writeFile(gitignorePath, filtered)
+  console.log('Updated .gitignore')
+}
+
+// Update `package.json` files to point to this branch on github
+async function updatePackageDependencies() {
+  let packagesDir = path.join(process.cwd(), 'packages')
+  let branch = 'future-build'
```
mjackson: Let's make this configurable via a command line arg (read from `process.argv`).

Let's also add a repo-level script so we can run this with `pnpm run`, like:

`pnpm run setup-installable-branch <branch-name>`

brophdawg11:
@frolic we've looked at it but one of the core principles of Remix 3 is to [avoid dependencies](https://github.com/remix-run/remix?tab=readme-ov-file#welcome-to-remix-3) and it's a bit more invasive than we'd like, requiring a github app with repo permissions.  We used Changesets in Remix v2 (and currently in React Router v7) and we find that relying on someone else's vision of building/publishing/generating release notes doesn't provide enough flexibility to tailor to our needs.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,57 @@
...
+  schedule:
+    - cron: '0 7 * * *' # every day at 12AM PST
+  workflow_dispatch:
+    inputs:
+      baseBranch:
+        description: Base Branch
+        required: true
+      installableBranch:
+        description: Installable Branch
+        required: true
```
brophdawg11, line 23: Allow manual runs to create experimental releases from our own branches (to test out PR features, etc)

PR #10978, opened 2026-01-21
Switch all Remix peer dependencies to regular dependencies

mjackson:
This PR switches all Remix `peerDependencies` to be regular `dependencies` in both the `remix` package and all sub-packages.

Previously (in alpha.0) our peer deps strategy resulted in "unmet peer" warnings during installation for any package that was not re-published when one of its peer deps changed. Making the switch to regular deps lets us avoid this warning and also makes it more clear to consumers the intended relationship of our packages.

Additionally, we determined that we weren't really using peer deps correctly. Peer deps were originally created for cases where we don't control the implementation of the package we both rely on, which doesn't apply in Remix.

### DIFF
--- a/packages/async-context-middleware/.changes/patch.peer-deps-to-deps.md
+++ b/packages/async-context-middleware/.changes/patch.peer-deps-to-deps.md
@@ -1,1 +1,1 @@
-Changed Remix peer dependencies to normal dependencies
+Changed `@remix-run/*` peer dependencies to regular dependencies
--- a/scripts/generate-remix.ts
+++ b/scripts/generate-remix.ts
@@ -56,15 +56,12 @@ lintRemixPackage(true)
 console.log('📦 Updating package.json...')
 let remixPackageJson = JSON.parse(await fs.readFile(remixPackageJsonPath, 'utf-8'))
 
-// Track existing exports and dependencies for comparison
-let existingInfo = {
-  exports: new Set<string>(
-    Object.keys(remixPackageJson.exports || {}).filter(
-      (key) => key !== '.' && key !== './package.json',
-    ),
+// Track existing exports for comparison
+let existingExports = new Set<string>(
+  Object.keys(remixPackageJson.exports || {}).filter(
+    (key) => key !== '.' && key !== './package.json',
   ),
-  dependencies: new Set<string>(Object.keys(remixPackageJson.dependencies || {})),
-}
+)
 
 // Build exports/publishConfig.exports/dependencies
 let { exportsConfig, publishConfigExports } = getRemixExports()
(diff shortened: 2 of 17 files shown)


mjackson:
@markdalgleish I would expect `packages/remix/.changes/patch.peer-deps-to-deps.md` to be named with a `pre` prefix instead of `patch` since it's intended to bump the prerelease version and not a patch release.

packages/async-context-middleware/.changes/patch.peer-deps-to-deps.md
```diff
@@ -0,0 +1 @@
+Changed Remix peer dependencies to normal dependencies
```
brophdawg11: I think in the individual changelogs this would read better as a generic `Changed peerDependencies to normal dependencies` so it doesn't seem to imply the wrapper `remix` package, which will have it's own change file.  Should be a simple find/replace across the repo if we want to make that change

scripts/generate-remix.ts
```diff
@@ -288,52 +288,48 @@ async function outputExportsChanges() {
...
+  let addedDeps = Array.from(newDepsSet).filter((key) => !existingInfo.dependencies.has(key))
+  let removedDeps = Array.from(existingInfo.dependencies).filter((key) => !newDepsSet.has(key))
 
-  if (addedPeerDeps.length === 0 && removedPeerDeps.length === 0) {
+  if (addedDeps.length === 0 && removedDeps.length === 0) {
     return
   }
 
-  let semverType = removedPeerDeps.length > 0 ? 'major' : 'minor'
+  let semverType = removedDeps.length > 0 ? 'major' : 'minor'
```
brophdawg11: I think `outputDependencyChanges` may jut be able to go away now?  When using `peerDeps` we had to consider those user-facing changes but now that they are just deps they're implementation details so we don't need a change file if we add/remove one.

I think we may still have a small gap and we would want a `remix` change file when we bump the dep versions versions at release time that lists the version bumps.

packages/async-context-middleware/.changes/patch.peer-deps-to-deps.md
mjackson: I didn't want to imply that *any* peer dependency was changed, just the ones that we control. The "Remix ones".

brophdawg11: Changed `@remix-run/*` `peerDependencies` to normal `dependencies`?

scripts/generate-remix.ts
mjackson: Sure, I'll remove `outputDependencyChanges`. And let's address the `remix` `.changes` file in a separate PR.

PR #10980, opened 2026-01-21
Add PR preview workflow

brophdawg11:
Adds a github workflow to build installable preview branches for PRs.

While a PR is open, every commit will trigger a fresh build to an installable `preview/<pr-number>` branch and comment back on the PR with instructions on how to install the preview branch (while deleting any comments from prior commits so there should only ever be one comment at the bottom of the PR).

Screenshot 2026-01-22 at 11 03 07 AM

When a PR is closed, the `preview/*` branch is deleted and a comment is added to the Pr indicating it's not longer available.

Screenshot 2026-01-22 at 11 03 10 AM

My only open question is whether we would prefer a sticky comment towards the top of the PR that doesn't require a new comment per-commit?  Is that more visible to folks visiting the PR who may not scroll to the bottom?

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1062,9 +1062,6 @@ importers:
       '@octokit/request':
         specifier: ^9.1.3
         version: 9.1.3
-      '@octokit/types':
-        specifier: ^16.0.0
-        version: 16.0.0
       '@types/node':
         specifier: 'catalog:'
         version: 24.10.4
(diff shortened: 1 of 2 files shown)


.github/workflows/preview.yml
```diff
@@ -0,0 +1,64 @@
...
+        if: github.event.pull_request.state == 'open'
+        run: |
+          git push --force --set-upstream origin preview/${{ github.event.pull_request.number }}
+          echo "💿 pushed installable branch: https://github.com/$GITHUB_REPOSITORY/commit/$(git rev-parse HEAD)"
+
+      - name: Comment on PR
+        if: github.event.pull_request.state == 'open'
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: pnpm run pr-preview comment ${{ github.event.pull_request.number }}
```
brophdawg11, line 57: Run these steps for open PRs

.github/workflows/preview.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      # Cleanup previews when PR's are closed
+      - name: Cleanup preview branch
+        if: github.event.pull_request.state == 'closed'
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: pnpm run pr-preview cleanup ${{ github.event.pull_request.number }}
```
brophdawg11, line 64: And this for closed PRs

PR #10981, opened 2026-01-21
chore: always tag `remix` prerelease as `next`

markdalgleish:
Right now `remix` releases on npm are tagged to match its prerelease version, e.g. `3.0.0-alpha.0` is tagged as `alpha`. This PR decouples these concepts so `3.0.0-alpha.1` would be tagged as `next`. To better match this strategy, the `tag` config option in `prerelease.json` has been renamed to `channel`, e.g. `{ "channel": "alpha" }`.

### DIFF
--- a/packages/remix/.changes/prerelease.json
+++ b/packages/remix/.changes/prerelease.json
@@ -1,3 +1,3 @@
 {
-  "tag": "alpha"
+  "channel": "alpha"
 }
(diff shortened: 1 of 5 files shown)


packages/remix/.changes/prerelease.json
```diff
@@ -1,3 +1,3 @@
 {
-  "tag": "alpha"
+  "channel": "alpha"
```
brophdawg11, line 2: I think I was envisioning an `npmTag: "next"` field in here.  We won't need that until we get towards v4 alphas and might want to publish all packages as `next` so we can cross that bridge when we get to it

mjackson: This file is temporary and will go away as soon as we're done with the alpha. I think it's fine to hardcode `next` in the publish script for now and revisit the next time we need a prerelease.

PR #10984, opened 2026-01-22
route-pattern: match indices, match ranking via specificity, and trie matcher

pcattori:
This PR reworks `route-pattern` to add rich information about the matched params, including their type (`:` or `*`) and location in the `hostname` or `pathname` (similar to `.indices` and `.groups` for a `RegExp` match) in `match.meta`.

The resulting `params` are derived from `match.meta` and matches can then be ranked/sorted/prioritized based on param types and locations. We provide a default ranking via "Specificity" where more specific matches are chosen over less specific ones.

Additionally, this PR provides implementations for `ArrayMatcher` and `TrieMatcher` optimized for different app sizes.

## Benchmarks

Screenshot 2026-01-26 at 3 22 42 PM

### DIFF
--- a/packages/route-pattern/src/lib/route-pattern/pathname.ts
+++ b/packages/route-pattern/src/lib/route-pattern/pathname.ts
@@ -1,5 +1,3 @@
-/* eslint-disable jsdoc/require-param */
-/* eslint-disable jsdoc/require-returns */
 import { PartPattern, type PartPatternToken } from './part-pattern.ts'
 
 /**
@@ -24,6 +22,11 @@ import { PartPattern, type PartPatternToken } from './part-pattern.ts'
  * join('(a)', '(/b)') -> '(a)(/b)'
  * join('(a/)', '(/b)') -> '(a)(/b)'
  * ```
+ *
+ * @param a the first pathname pattern
+ * @param b the second pathname pattern
+ * @param ignoreCase whether to ignore case when matching
+ * @returns the joined pathname pattern
  */
 export function join(a: PartPattern, b: PartPattern, ignoreCase: boolean): PartPattern {
   if (a.tokens.length === 0) return b
--- a/packages/route-pattern/src/lib/route-pattern/search.ts
+++ b/packages/route-pattern/src/lib/route-pattern/search.ts
@@ -1,6 +1,3 @@
-/* eslint-disable jsdoc/require-param */
-/* eslint-disable jsdoc/require-returns */
-
 import type { RoutePattern } from '../route-pattern.ts'
 import { HrefError } from '../errors.ts'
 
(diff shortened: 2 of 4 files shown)


packages/fetch-router/src/lib/router.ts
```diff
@@ -181,17 +177,19 @@ function noMatchHandler({ url }: RequestContext): Response {
  */
 export function createRouter(options?: RouterOptions): Router {
   let defaultHandler = options?.defaultHandler ?? noMatchHandler
-  let matcher = options?.matcher ?? new ArrayMatcher<MatchData>()
+  let matcher = options?.matcher ?? new ArrayMatcher()
   let globalMiddleware = options?.middleware
+  let routeData = new WeakMap<RoutePattern, Map<RequestMethod | 'ANY', MatchData>>()
```
pcattori: Router now owns the "data" payload; matcher itself only returns the `Match` itself.

Might be some opportunity to make `WeakMap` based data lookup a bit nicer

packages/fetch-router/.changes/minor.remove-router-size.md
```diff
@@ -0,0 +1,3 @@
+BREAKING CHNAGE: Remove `Router.size` property
```
mjackson:
```suggestion
BREAKING CHANGE: Remove `Router.size` property
```

packages/route-pattern/.changes/minor.remove-ignoreCase.md
```diff
@@ -0,0 +1,15 @@
+BREAKING CHANGE: Remove `ignoreCase` option
```
mjackson: Let's keep this one.

- `true` ignores case in pathname and search
- `pathname` ignores case in pathname
- `search` ignores case in search

It should default to `false`.

packages/route-pattern/src/lib/route-pattern/pathname.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  let tokens: Array<PartPattern.Token> = []
+
+  // if `a` has a trailing separator (only optionals after it)
+  // then omit the separator
+  let aLastNonOptionalIndex = a.tokens.findLastIndex((token) => token.type !== '(' && token.type !== ')')
+  let aLastNonOptional = a.tokens[aLastNonOptionalIndex]
+  let aHasTrailingSeparator = aLastNonOptional?.type === 'separator'
+
+  a.tokens.forEach((token, index) => {
```
mjackson, line 45: Probably should just use an inline `for` loop here (and below).

packages/route-pattern/src/lib/route-pattern/route-pattern.ts
```diff
@@ -0,0 +1,237 @@
+import { split } from './split.ts'
```
mjackson: Let's lift this file out into `lib`. If we need to split up the `RoutePattern` class, some stuff can live inside a `lib/route-pattern` dir, but preferably it would all just live in `lib/route-pattern.ts`.

packages/fetch-router/src/lib/router.ts
mjackson: I wanna push back on this. Let's add an opaque data type to the matcher interface so that consumers don't have to manage their own WeakMap.

packages/route-pattern/.changes/minor.remove-ignoreCase.md
pcattori: I added the option back in as just a boolean. we can add `pathname`/`search` for more targeted case sensitivity later as a non-breaking change.

PR #10987, opened 2026-01-23
Enable Remix change file generation

brophdawg11:
Enable generation of `remix` `.changes` files when we generate the package:

 - Copy change files from sub packages up to `remix`, since any change to an API in a sub-package is also a change to the re-exported API in remix
   - These will likely need some manual massaging at or before release time in case they reference the actual package name and we need to update it to the remix package export (or top level)
 - Add change file for added/removed remix `package.json` `exports`
 - Add change files to reflect updated `@remix-run.*` `dependencies` in `remix` `package.json`

### DIFF
--- a/packages/remix/.changes/patch.async-context-middleware.peer-deps-to-deps.md
+++ b/packages/remix/.changes/patch.async-context-middleware.peer-deps-to-deps.md
@@ -1,0 +1,1 @@
+Changed `@remix-run/*` peer dependencies to regular dependencies
--- a/packages/remix/.changes/patch.compression-middleware.peer-deps-to-deps.md
+++ b/packages/remix/.changes/patch.compression-middleware.peer-deps-to-deps.md
@@ -1,0 +1,1 @@
+Changed `@remix-run/*` peer dependencies to regular dependencies
--- a/packages/remix/.changes/patch.cookie.peer-deps-to-deps.md
+++ b/packages/remix/.changes/patch.cookie.peer-deps-to-deps.md
@@ -1,0 +1,1 @@
+Changed `@remix-run/*` peer dependencies to regular dependencies
--- a/packages/remix/.changes/patch.fetch-proxy.peer-deps-to-deps.md
+++ b/packages/remix/.changes/patch.fetch-proxy.peer-deps-to-deps.md
@@ -1,0 +1,1 @@
+Changed `@remix-run/*` peer dependencies to regular dependencies
--- a/packages/remix/.changes/patch.fetch-router.peer-deps-to-deps.md
+++ b/packages/remix/.changes/patch.fetch-router.peer-deps-to-deps.md
@@ -1,0 +1,1 @@
+Changed `@remix-run/*` peer dependencies to regular dependencies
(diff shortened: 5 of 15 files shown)


scripts/generate-remix.ts
```diff
@@ -8,25 +8,24 @@
...
 let __dirname = path.dirname(url.fileURLToPath(import.meta.url))
 let packagesDir = path.resolve(__dirname, '../packages')
 let remixDir = path.join(packagesDir, 'remix')
+let remixChangesDir = path.join(remixDir, '.changes')
 let remixPackageJsonPath = path.join(remixDir, 'package.json')
 
-// Don't do anything with change files until we actually publish an initial Remix v3 alpha
-// release, since only then will we care to start documenting what changed.
-// And Maybe not even until our first stable release.
-const GENERATE_CHANGE_FILES = false
```
brophdawg11, line 24: Now that we've shipped our first remix alpha, we can start tracking changes

scripts/generate-remix.ts
```diff
@@ -41,19 +40,6 @@ type ExportEntry = {
 }
 
 let { remixRunPackages, allExports } = await getRemixRunPackages()
-
-// Ensure we have a passing linter before generating code
-lintRemixPackage(false)
```
brophdawg11, line 46: Refactored this initial logic into `updateRemixPackage`

scripts/generate-remix.ts
```diff
@@ -276,12 +249,85 @@ async function outputExportsChanges() {
...
-    console.log('✨ Created exports change file:')
-    console.log(`   - ${path.relative(process.cwd(), changeFile)}`)
+  await fs.writeFile(changeFile, changes, 'utf-8')
+  console.log()
+  console.log('Created exports change file:')
+  console.log(`   - ${path.relative(process.cwd(), changeFile)}`)
+}
+
+// Build dependencies change summary
+async function outputDependencyChangeFiles() {
```
brophdawg11, line 259: This is the main functional change - we detect changes to remix dependencies by comparing against the versions in the repo as of the last remix release tag

packages/remix/.changes/patch.async-context-middleware.peer-deps-to-deps.md
```diff
@@ -0,0 +1 @@
+Changed `@remix-run/*` peer dependencies to regular dependencies
```
brophdawg11, line 1: This is the type of manual curation we'll need to clean up at release time - most functional chnages that warrant change file in sub packages will require a remix equivalent.  For example `Changed router.map API in @remix-run/fetch-router` would need to be reflected in remix as `Changed router.map API in remix/fetch-router`.

But some changes, such as dep changes like these, won't need to be duplicated to remix.  IMO, we're better off copying everything and then pruning at release time so we don't miss anything.

I think a good way to allow doing this before release time would be to permit us to rename a copied change file to something like `skip.patch.{package}.{change}.md` and the we would know that we don't need to copy that change file on subsequent runs and our changelog generation would know to skip it in the Remix release notes.

PR #10997, opened 2026-01-27
Update release flow for manual inclusion of Remix package

brophdawg11:
We plan to near-continuously deploy `@remix-run/*` sub-packages as changes are merged to main, but we don't want to release the umbrella `remix` package every time a sub-package changes.  Because `remix` uses caret `dependencies`,  re-installs and/or `overrides` can be used by apps who wish to opt-into a newly released sub-package before a new `remix` release is published.

This updates our workflow to support excluding remix from automated releases until we decide to include it:

* `remix/.changes/prerelease.config` is renamed to `remix/changes/config.json`
* A new `include` boolean is added indicating whether the `remix` package should be included in release flows
  * This flag is checked in `parseAllChangeFiles` 
  * This can be manually edited as we see fit but, my plan is to add logic to the Version Package PR workflow so we can comment `include remix`/`exclude remix` on there to update the file via the PR as well (comments would only be respected if they came from a maintainer)
* The publish process checks `include` this when deciding what publish commands to run
* When remix is published via `include:true`, it's `include` value is set back to `false` so its excluded until we decide to include it again

### DIFF
--- a/scripts/generate-remix.ts
+++ b/scripts/generate-remix.ts
@@ -55,7 +55,6 @@ await updateRemixPackage()
 // Generate change files
 await outputExportsChangeFiles(remixPackageJson.exports)
 await outputDependencyChangeFiles()
-await copySubPackageChangeFiles()
 
 // Implementations
 async function getRemixRunPackages() {
@@ -331,39 +330,6 @@ async function outputDependencyChangeFiles() {
   console.log()
   console.log(`${exists ? 'Updated' : 'Created'} dependencies change file:`)
   console.log(`   - ${path.relative(process.cwd(), changeFilePath)}`)
-}
-
-async function copySubPackageChangeFiles() {
-  if (remixRunPackages.some((pkg) => pkg.changeFiles.length > 0)) {
-    let copiedFiles: string[] = []
-    for (let packageInfo of remixRunPackages) {
-      for (let changeFile of packageInfo.changeFiles) {
-        let [changeType, ...rest] = path.basename(changeFile).split('.')
-        let changeFileName = rest.join('.')
-        let packageShortName = packageInfo.name.replace('@remix-run/', '')
-
-        // Insert the sub-package name in the file so sorting will group changes by sub-package
-        let destChangeFilePath = path.join(
-          remixChangesDir,
-          `${changeType}.${packageShortName}.${changeFileName}`,
-        )
-
-        let exists = await fileExists(destChangeFilePath)
-        if (!exists) {
-          await fs.copyFile(changeFile, destChangeFilePath)
-          copiedFiles.push(destChangeFilePath)
-        }
-      }
-    }
-
-    if (copiedFiles.length > 0) {
-      console.log()
-      console.log('Copied change files:')
-      for (let file of copiedFiles) {
-        console.log(`   - ${path.relative(process.cwd(), file)}`)
-      }
-    }
-  }
 }
 
 async function fileExists(filePath: string) {


scripts/utils/changes.ts
```diff
@@ -189,71 +211,86 @@ function parsePackageChanges(packageDirName: string): ParsedPackageChanges {
...
   // Read all files in .changes directory
   let files = fs.readdirSync(changesDir)
-  let changeFileNames = files.filter((file) => file !== 'README.md' && file !== 'prerelease.json')
+  let changeFileNames = files.filter(
+    (file) =>
+      file !== 'README.md' &&
+      file !== configFileName &&
+      // Only remix supports skip.* files, so we can exclude unneeded change files
+      // from sub-packages
+      (packageDirName !== 'remix' || !file.startsWith('skip.')),
```
brophdawg11: New logic to support `skip.*` prefixes for remix change files

scripts/utils/changes.ts
```diff
@@ -390,12 +437,30 @@ type ParsedChanges =
...
+            file: configFilePath,
+            error: remixReleaseConfig.error,
+          })
+          continue
+        } else if (remixReleaseConfig.config.include == false) {
+          console.log('Skipping `remix` package due to release config `include: false`')
+          continue
+        }
+      }
+    }
```
brophdawg11, line 464: Don't include Remix when parsing change files if `include === false`

scripts/changes-validate.ts
```diff
@@ -39,7 +39,8 @@ function main() {
 
   // Validate change files
   console.log('Validating change files...\n')
-  let result = parseAllChangeFiles()
+  // Always validate remix even if it's set to include:false
+  let result = parseAllChangeFiles(true)
```
brophdawg11, line 43: Always validate Remix change files regardless of the `include` value

scripts/publish.ts
```diff
@@ -213,7 +215,7 @@ async function main() {
 
   let published: PublishedPackage[] = []
 
-  if (remixPrereleaseChannel) {
+  if (includeRemixRelease && remixPrereleaseChannel) {
```
brophdawg11, line 222: Only do remix prereleases if included

scripts/publish.ts
```diff
@@ -234,10 +236,14 @@ async function main() {
       }
     }
   } else {
-    // Single-phase publish: everything as latest
+    // Single-phase publish: everything as latest (including remix)
     let publishCommand =
       'pnpm publish --recursive --filter "./packages/*" --access public --no-git-checks --report-summary'
 
+    if (!includeRemixRelease) {
+      publishCommand += ' --filter "!remix"'
```
brophdawg11, line 248: For stable releases, filter it out when not included

scripts/publish.ts
```diff
@@ -355,6 +365,23 @@ async function main() {
...
+    console.log('\nUpdating remix release config to set include: false...')
+    writeRemixReleaseConfig({
+      channel: remixReleaseConfig.config.channel,
+      include: false,
+    })
+    logAndExec(`git add packages/remix/.changes/config.json`)
+    logAndExec(`git commit -m "chore: set remix release config include:false"`)
+    logAndExec('git push')
+  }
+
```
brophdawg11, line 384: Flip the remix `include` value back to false after each release so it's always excluded by default until we choose to include it

packages/remix/.changes/config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "channel": "alpha",
+  "include": false
```
brophdawg11, line 3: Don't include `remix` in automated releases when this is `false`

PR #11023, opened 2026-01-30
AGENTS.md: file structure + `node:assert` and `node:test`

pcattori:
I'm mainly brain dumping into a PR to get discussion going, but feels like we might want to rethink how our `AGENTS.md` is structured. Maybe more examples/references?

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,6 +18,15 @@
 - **Philosophy**: Web standards-first, runtime-agnostic (Node.js, Bun, Deno, Cloudflare Workers). Use Web Streams API, Uint8Array, Web Crypto API, Blob/File instead of Node.js APIs
 - **Tests run from source** (no build required), using Node.js test runner
 
+## File Structure
+
+- Main entrypoint in `src/index.ts`
+- Implementation in `src/lib/`
+- Helpers in a sibling directory named the same as the file they are used in
+  - For example: `src/lib/route-pattern.ts` has helpers in `src/lib/route-pattern/`
+- Benchmarks in their own subpackage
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -32,6 +41,7 @@
 
 ## Test Structure
 
+- Use `import * from 'node:assert/strict'` and `import { describe, it } from 'node:test'`
```
pcattori, line 44: Feels like an example usage here would be useful, but we don't have any other code examples yet. @brookslybrand curious what you think

AGENTS.md
```diff
@@ -18,6 +18,15 @@
 - **Philosophy**: Web standards-first, runtime-agnostic (Node.js, Bun, Deno, Cloudflare Workers). Use Web Streams API, Uint8Array, Web Crypto API, Blob/File instead of Node.js APIs
 - **Tests run from source** (no build required), using Node.js test runner
 
+## File Structure
+
+- Main entrypoint in `src/index.ts`
+- Implementation in `src/lib/`
+- Helpers in a sibling directory named the same as the file they are used in
+  - For example: `src/lib/route-pattern.ts` has helpers in `src/lib/route-pattern/`
+- Benchmarks in their own subpackage
```
mjackson, line 27:
```suggestion
- Benchmarks in a `bench` subdir in their own pnpm workspace (i.e. with its own `package.json`) so all dependencies, including libraries we are benchmarking against, don't end up in the `devDependencies` of the package itself
```

mjackson (approved):
Thanks for taking this on @pcattori 👍 I'll do some additional work on this file today to make sure it's up to date

PR #11034, opened 2026-02-04
fix dynamic css selectors

ryanflorence:
```tsx
css={{
  '&:hover': pending ? undefined : { background: 'white' }
}}
```

broke, now it doesn't.

also got rid of some type casting


mjackson:
Looks good, just missing a `.changes` file, which means it won't be published on the next release. I'll make one.

PR #11037, opened 2026-02-05
Remove remix root index.ts file and add fetch-router routes export

brophdawg11:
This PR makes 2 structural changes to exports:

* Removes the root `remix` `index.ts` export to force us to import from sub-path exports to keep downstream/non-tree-shaken code to a minimum
* Moves route-map/route-helper utilities to a new `@remix-run/fetch-router/routes` export to avoid accidental inclusion in client bundles
  * This happens today if a `routes.ts` file imports these helpers from `@remix-run/fetch-router` because it transitively includes things like `@remix-run/session`

### DIFF
--- a/packages/remix/package.json
+++ b/packages/remix/package.json
@@ -18,7 +18,6 @@
   ],
   "type": "module",
   "exports": {
-    ".": "./src/index.ts",
     "./async-context-middleware": "./src/lib/async-context-middleware.ts",
     "./component": "./src/lib/component.ts",
     "./component/jsx-dev-runtime": "./src/lib/component/jsx-dev-runtime.ts",
@@ -66,10 +65,6 @@
   },
   "publishConfig": {
     "exports": {
-      ".": {
-        "types": "./dist/index.d.ts",
-        "default": "./dist/index.js"
-      },
       "./async-context-middleware": {
         "types": "./dist/lib/async-context-middleware.d.ts",
         "default": "./dist/lib/async-context-middleware.js"


packages/fetch-router/src/index.ts
```diff
@@ -10,49 +10,5 @@ export { RequestContext } from './lib/request-context.ts'
...
-  createResourceRoutes,
-  createResourceRoutes as resource, // shorthand
-} from './lib/route-helpers/resource.ts'
-export type { ResourceMethod, ResourceOptions } from './lib/route-helpers/resource.ts'
-
-export {
-  createResourcesRoutes,
-  createResourcesRoutes as resources, // shorthand
-} from './lib/route-helpers/resources.ts'
-export type { ResourcesMethod, ResourcesOptions } from './lib/route-helpers/resources.ts'
```
brophdawg11, line 58: All of these are moved to the new `routes` export

packages/remix/package.json
```diff
@@ -18,7 +18,6 @@
   ],
   "type": "module",
   "exports": {
-    ".": "./src/index.ts",
```
brophdawg11: Remove root `remix` export

scripts/generate-remix.ts
```diff
@@ -146,15 +146,8 @@ async function updateRemixPackage() {
...
-    '.': './src/index.ts',
-  }
-  remixPackageJson.publishConfig.exports = {
-    '.': {
-      types: './dist/index.d.ts',
-      default: './dist/index.js',
-    },
-  }
+  remixPackageJson.exports = {}
+  remixPackageJson.publishConfig.exports = {}
```
brophdawg11, line 150: No longer generate the root remix export

mjackson:
This is a fantastic PR, thanks @brophdawg11!

mjackson (approved):

PR #11039, opened 2026-02-06
fix(form-data-parser): handle non-ASCII filenames in multipart uploads

### LINKED ISSUE #11032: [form-data-parser] parseFormData fails with non-ASCII filenames (e.g., Japanese, Chinese, Korean)
## Summary

`parseFormData` throws `FormDataParseError` when processing file uploads with non-ASCII characters in the filename (e.g., Japanese: `テスト画像.png`).

This issue appears to have been introduced in `@remix-run/headers` v0.16.0 when `SuperHeaders` was replaced with the standard `Headers` class (#10911). It seems to be an unintended side effect of this change.

## Affected Versions

- `@remix-run/form-data-parser` >= 0.15.0
- `@remix-run/headers` >= 0.16.0
- `@remix-run/multipart-parser` >= 0.14.0

## Error

```text
FormDataParseError: Cannot parse form data
  cause: TypeError: Failed to execute 'append' on 'Headers': String contains non ISO-8859-1 code point.
```

## Root Cause

In #10911, the `SuperHeaders` class was removed and replaced with `parseRawHeaders` in `@remix-run/headers`.

**Before:**
```typescript
// packages/multipart-parser/src/lib/multipart.ts
import Headers from '@remix-run/headers'
this.#headers = new Headers(decoder.decode(this.#header))
```

**After:**
```typescript
// packages/multipart-parser/src/lib/multipart.ts
import { parse as parseRawHeaders } from '@remix-run/headers'
this.#headers = parseRawHeaders(decoder.decode(this.#header))
```

## Steps to Reproduce

1. Create a file upload form
2. Upload a file with a non-ASCII filename (e.g., `テスト画像.png`, `文件.png`, `파일.png`)
3. Use `parseFormData(request)` to parse the request
4. Observe `FormDataParseError`

## Expected Behavior

`parseFormData` should successfully parse file uploads with non-ASCII filenames.

## Actual Behavior

`parseFormData` throws `FormDataParseError` with cause `TypeError: Failed to execute 'append' on 'Headers': String contains non ISO-8859-1 code point.`

kaigritun:
## Summary

Fixes #11032

When parsing multipart form data with non-ASCII filenames (e.g., Japanese: `テスト画像.png`, Chinese: `文件.png`, Korean: `파일.png`), the native `Headers` class throws because it only accepts ISO-8859-1 characters.

## Root Cause

The `parseRawHeaders` function in `@remix-run/headers` uses the native `Headers.append()` method, which throws a `TypeError` for non-ASCII characters in header values. This affects `Content-Disposition` headers containing non-ASCII filenames.

## Solution

1. **Catch the encoding error**: When `Headers.append()` throws for non-ASCII characters, catch the error (handling different runtime error messages for browsers and Node.js)

2. **Percent-encode non-ASCII bytes**: Encode non-ASCII characters as percent-encoded UTF-8 bytes (e.g., `テスト` → `%E3%83%86%E3%82%B9%E3%83%88`)

3. **Decode when accessing filename**: Update `ContentDisposition.preferredFilename` to decode percent-encoded UTF-8 back to Unicode characters

## Changes

- `packages/headers/src/lib/raw-headers.ts`: Add `encodeNonAscii()` helper and error handling in `parse()`
- `packages/headers/src/lib/content-disposition.ts`: Add `decodePercentEncodedUtf8()` helper and call it in `preferredFilename` getter
- Tests added for both encoding (raw-headers) and decoding (content-disposition)

## Testing

All existing tests pass, plus new tests for:
- Japanese filenames (`テスト画像.png`)
- Chinese filenames (`文件.png`)
- Korean filenames (`파일.png`)
- Mixed ASCII and non-ASCII filenames

```bash
pnpm test --filter @remix-run/headers
# 387 tests pass
```

### DIFF
--- a/packages/headers/src/lib/content-disposition.ts
+++ b/packages/headers/src/lib/content-disposition.ts
@@ -58,7 +58,13 @@ export class ContentDisposition implements HeaderValue, ContentDispositionInit {
       if (decodedFilename) return decodedFilename
     }
 
-    return this.filename
+    // Decode percent-encoded UTF-8 in filename (handles non-ASCII characters
+    // that were encoded to pass through native Headers validation)
+    if (this.filename) {
+      return decodePercentEncodedUtf8(this.filename)
+    }
+
+    return undefined
   }
 
   /**
@@ -144,4 +150,40 @@ function percentDecode(value: string): string {
   return value.replace(/\+/g, ' ').replace(/%([0-9A-Fa-f]{2})/g, (_, hex) => {
     return String.fromCharCode(parseInt(hex, 16))
   })
+}
+
+/**
+ * Decodes percent-encoded UTF-8 bytes in a string back to Unicode characters.
+ * This handles filenames that were encoded to pass through native Headers validation.
+ * Only decodes if percent-encoding is present; returns original string otherwise.
+ */
+function decodePercentEncodedUtf8(value: string): string {
+  // Check if there are any percent-encoded sequences
+  if (!value.includes('%')) {
+    return value
+  }
+
+  try {
+    // Convert percent-encoded sequences to bytes
+    let bytes: number[] = []
+    let i = 0
+    while (i < value.length) {
+      if (value[i] === '%' && i + 2 < value.length) {
+        let hex = value.substring(i + 1, i + 3)
+        if (/^[0-9A-Fa-f]{2}$/.test(hex)) {
+          bytes.push(parseInt(hex, 16))
+          i += 3
+          continue
+        }
+      }
+      bytes.push(value.charCodeAt(i))
+      i++
+    }
+
+    // Decode UTF-8 bytes
+    return new TextDecoder('utf-8', { fatal: false }).decode(new Uint8Array(bytes))
+  } catch {
+    // If decoding fails, return the original value
+    return value
+  }
 }
--- a/packages/headers/src/lib/content-disposition.ts
+++ b/packages/headers/src/lib/content-disposition.ts
@@ -156,6 +156,9 @@ function percentDecode(value: string): string {
  * Decodes percent-encoded UTF-8 bytes in a string back to Unicode characters.
  * This handles filenames that were encoded to pass through native Headers validation.
  * Only decodes if percent-encoding is present; returns original string otherwise.
+ *
+ * @param value - The string to decode
+ * @returns The decoded string with percent-encoded UTF-8 bytes converted to Unicode
  */
 function decodePercentEncodedUtf8(value: string): string {
   // Check if there are any percent-encoded sequences
--- a/packages/headers/src/lib/raw-headers.ts
+++ b/packages/headers/src/lib/raw-headers.ts
@@ -5,6 +5,9 @@ const CRLF = '\r\n'
 /**
  * Encodes non-ASCII characters in a string using percent-encoding.
  * This is needed because the native Headers class only accepts ISO-8859-1 characters.
+ *
+ * @param value - The string to encode
+ * @returns The encoded string with non-ASCII characters percent-encoded
  */
 function encodeNonAscii(value: string): string {
   let encoder = new TextEncoder()
--- a/packages/headers/src/lib/raw-headers.ts
+++ b/packages/headers/src/lib/raw-headers.ts
@@ -8,13 +8,17 @@ const CRLF = '\r\n'
  */
 function encodeNonAscii(value: string): string {
   let encoder = new TextEncoder()
-  return [...value].reduce((result, char) => {
+
+  return Array.from(value).reduce((result, char) => {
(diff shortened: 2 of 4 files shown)


packages/headers/src/lib/raw-headers.ts
```diff
@@ -2,6 +2,27 @@ import { canonicalHeaderName } from './header-names.ts'
...
+      let bytes = new TextEncoder().encode(value[i])
+      for (let byte of bytes) {
+        result += '%' + byte.toString(16).toUpperCase().padStart(2, '0')
+      }
+    } else {
+      result += value[i]
+    }
+  }
+  return result
+}
```
MichaelDeBoey, line 27: I'd suggest moving the `TextEncoder` instantiation outside the loop to avoid repeated allocations and using `reduce` instead of a `for` loop so the accumulator is handled more functionally with fewer mutations.

```suggestion
function encodeNonAscii(value: string): string {
  let encoder = new TextEncoder()

  return Array.from(value).reduce((result, char) => {
    let code = char.charCodeAt(0)

    if (code > 127) {
      return encoder.encode(char).reduce(
        (acc, byte) => acc + '%' + byte.toString(16).toUpperCase().padStart(2, '0'),
        result
      )
    }

    return result + char
  }, '')
}
```

MichaelDeBoey (changes requested):

kaigritun:
Updated per feedback - moved TextEncoder instantiation outside the loop and refactored to use reduce.

packages/headers/src/lib/raw-headers.ts
```diff
@@ -2,6 +2,23 @@ import { canonicalHeaderName } from './header-names.ts'
 
 const CRLF = '\r\n'
 
+/**
+ * Encodes non-ASCII characters in a string using percent-encoding.
+ * This is needed because the native Headers class only accepts ISO-8859-1 characters.
+ */
+function encodeNonAscii(value: string): string {
+  let encoder = new TextEncoder()
+  return [...value].reduce((result, char) => {
```
MichaelDeBoey: Let's use `Array.from` instead of spreading


```suggestion
  return Array.from(value).reduce((result, char) => {
```

packages/headers/src/lib/raw-headers.ts
```diff
@@ -2,6 +2,23 @@ import { canonicalHeaderName } from './header-names.ts'
...
+    let code = char.charCodeAt(0)
+    if (code > 127) {
+      // Encode non-ASCII characters as UTF-8 percent-encoded bytes
+      let bytes = encoder.encode(char)
+      return result + [...bytes].map(byte => '%' + byte.toString(16).toUpperCase().padStart(2, '0')).join('')
```
MichaelDeBoey: Since `bytes` is a temporary variable and isn't used anywhere else, let's inline it.

Let's also use `reduce`, so we don't have to do `join` on the full array


```suggestion
      return encoder.encode(char).reduce(
        (acc, byte) => acc + '%' + byte.toString(16).toUpperCase().padStart(2, '0'),
        result
      )
```

MichaelDeBoey (changes requested):

packages/headers/src/lib/content-disposition.ts
```diff
@@ -145,3 +151,39 @@ function percentDecode(value: string): string {
...
+          i += 3
+          continue
+        }
+      }
+      bytes.push(value.charCodeAt(i))
+      i++
+    }
+
+    // Decode UTF-8 bytes
+    return new TextDecoder('utf-8', { fatal: false }).decode(new Uint8Array(bytes))
```
MichaelDeBoey, line 187: If we don't use mutations here either, we can simplify this to

```suggestion
    let bytes = Uint8Array.from(
      value.match(/%[0-9A-Fa-f]{2}|./g)!,
      token =>
        token[0] === '%'
          ? parseInt(token.slice(1), 16)
          : token.charCodeAt(0)
    )

    return new TextDecoder('utf-8', { fatal: false }).decode(bytes)
```

MichaelDeBoey (changes requested):
It seems linting is also failing

rossipedia:
This user is a fully-autonomous non-human actor:

https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

I don't make any claims to the quality of their work, but I do worry about the fact that they are not up-front about them not being a human actor.

mathewmeconry:
I believe this AI bot is offline, so I created a new pull request with the fixes and requested changes from this one so we can get a fix for this issue in
https://github.com/remix-run/remix/pull/11139

MichaelDeBoey:
Duplicate of #11139

mjackson:
Superseded and fixed by #11263.

PR #11040, opened 2026-02-06
Fix Remix types

brophdawg11:
https://github.com/remix-run/remix/pull/11037 removed the `src/index.ts` file which caused the `.d.ts` files for all the other packages to shift up since the lack of any other files alongside the `lib` directory, which meant our paths in `package.json` `publishConfig` weren't right anymore.

Since we don't have an `index.ts` anymore, we just pulled everything out of lib which aligns better with the rest of the repo where export files are directly in `src/`

### DIFF
--- a/packages/remix/src/lib/component/server.ts
+++ b/packages/remix/src/lib/component/server.ts
@@ -1,2 +1,0 @@
-// IMPORTANT: This file is auto-generated, please do not edit manually.
-export * from '@remix-run/component/server'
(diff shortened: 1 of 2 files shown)


scripts/generate-remix.ts
```diff
@@ -127,10 +126,15 @@ async function updateRemixPackage() {
   // Ensure we have a passing linter before generating code
   logAndExec(`npx eslint packages/remix/ --max-warnings=0`)
 
-  // Generate source files
+  // Clear existing source files
+  let sourceFolderPath = path.join(remixDir, SOURCE_FOLDER)
+  await fs.rm(sourceFolderPath, { recursive: true, force: true })
+  await fs.mkdir(sourceFolderPath, { recursive: true })
```
brophdawg11, line 132: Make sure we clear the src directory and do a full rebuild to ensure no stale files are left around due to renames/removals of exports

PR #11041, opened 2026-02-06
Remix package generation updates

brophdawg11:
~~Stacked on https://github.com/remix-run/remix/pull/11040~~

* ~~Updated preview build commands in our docs to include an explicit package name.  Omitting it works fine on fresh installs but I have seen some errors trying to install a fresh preview build overtop of a prior `remix` install.~~
  * Skipping this for now - planning to file a bug with pnpm and see if we can get it addressed there  
* Removed logic from `generate-remix.ts` for detecting remix dep bumps and generating change files.  This was leftover from when we didn't intend to release remix every time, but now that we release remix every time this is handled by the same logic as the other packafes in https://github.com/remix-run/remix/blob/main/scripts/utils/changes.ts#L482

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,10 +1,10 @@
 # Create "installable" preview branches
 #
 # Commits to `main` push builds to a `preview/main` branch:
-#   pnpm install "remix@remix-run/remix#preview/main&path:packages/remix"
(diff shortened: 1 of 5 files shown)


brophdawg11:
This is now just removing dead code which is handled by the release scripts

PR #11042, opened 2026-02-07
Add data-schema package

mjackson:
Adds a tiny, sync-first schema validation library that implements [Standard Schema v1](https://standardschema.dev/).

**Why**

Remix needs a lightweight way to validate data at boundaries (form submissions, route params, loader results) without pulling in a full-sized library like Zod or Valibot. By owning this, we can keep the dependency footprint minimal while staying interoperable with the broader Standard Schema ecosystem.


afoures:
i see that the error messages are currently hard coded, is it planned to have a way to override them to allow for i18n for example?

packages/data-schema/README.md
```diff
@@ -0,0 +1,339 @@
...
+
+let Event = variant('type', {
+  created: object({ type: literal('created'), id: string() }),
+  updated: object({ type: literal('updated'), id: string(), version: number() }),
+})
+```
+
+## Recursive schemas
+
+Model trees and self-referencing structures. `lazy()` defers schema resolution to avoid circular references.
```
MichaelDeBoey, line 269: Some time ago, I saw @colinhacks post something about using getters to avoid circular references instead of needing a dedicated helper function like `z.lazy`

afoures: https://x.com/colinhacks/status/1919286275133378670 and https://x.com/colinhacks/status/1924484449053585754

mjackson: Getters work well for `object()`, but `lazy()` can still be useful for non-`object()` schemas like `union()`, etc.

For example:

```ts
import { array, boolean, null_, number, record, string, union } from '@remix-run/data-schema'
import { lazy } from '@remix-run/data-schema/lazy'

type Json = null | string | number | boolean | Json[] | { [k: string]: Json }

let JsonSchema
JsonSchema = lazy(() =>
  union([null_(), string(), number(), boolean(), array(JsonSchema), record(string(), JsonSchema)]),
)
```

mjackson:
> i see that the error messages are currently hard coded, is it planned to have a way to override them to allow for i18n for example?

That's a great question, @afoures. Thanks for bringing that to my attention.

We should add a customizable error message hook somewhere, especially because I'm planning on using these schemas in the data-table work I'm doing to validate database inputs. I'll look into it.

PR #11044, opened 2026-02-09
Add scripts to auto generate docs

brophdawg11:
Adds scripts to auto-generate prerendered docs HTML for deployment via GH pages in a separate repo.  All docs-related code lives in a new pnpm workspace in `docs/` with a few scripts:

* `pnpm run docs` - generate markdown docs in `docs/build/md` from source code JSDocs
* `pnpm run serve` - run an HTTP server to serve markdown docs as HTML
* `pnpm run prerender` - pre-render the markdown docs to static HTML in `docs/site/`

Also added a few missing comments and exported a couple APIs to ensure they show up in docs, but there is a larger pass that still needs to be done here before the generated docs will be complete.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -82,3 +82,23 @@ jobs:
         run: node ./scripts/publish.ts
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  #     - name: Output tag
+  #       run: echo "tag=remix@$(node -p "require('./packages/remix/package.json').version")" >> $GITHUB_OUTPUT
+
+  # docs:
+  #   name: Update API Docs
+  #   needs: publish
+  #   runs-on: ubuntu-latest
+  #   steps:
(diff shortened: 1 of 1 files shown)


docs/src/server/prerender.ts
```diff
@@ -0,0 +1,99 @@
...
+      short: 'd',
+      default: 'site',
+    },
+  },
+})
+
+await spider(docsRouter, cliArgs.dir)
+
+// Spider the website served by router, beginning at /
+async function spider(router: Router, outDir: string, urlQueue = ['/']) {
```
brophdawg11: This API could be  a potential addition to `fetch-router` for built in pre-rendering if we were interested in that:

```ts
let router = createRouter(routes, handlers);
await router.spider("my-site") // or router.prerender()
// HTML written to ./my-site
```

Would need some minor configs to give the user a bit of control over what gets included in the spidering process (or ways to disable spidering and just specify a static set of paths up front).

PR #11047, opened 2026-02-10
Update documentation imports to use remix package

brophdawg11:
Updates all documentation imports to use `remix/*` instead of `@remix-run/*`.  This includes `README.md` files, `AGENTS.md` files, source code JSDoc comments, and other forms of documentation including one active `.changes` file


MichaelDeBoey (approved):

mjackson:
@MichaelDeBoey can you please delete all the `pkg` labels? They are noisy and I don't think they add any real value when browsing the issues list at this point. If we need them in the future, we can always add them back.

PR #11048, opened 2026-02-10
Frames: out-of-order streaming server rendering, selective hydration, composable async UI and granular refresh

ryanflorence:
Here is the flow from

1. streamed server rendering
2. frames
3. hydration
4. ui updates from frame reload (hybrid dom/vdom diff)

## 1. Streamed Server Rendering

Everything starts with `renderToStream(node, options)` in `packages/component/src/lib/stream.ts`. It takes a component tree and returns a `ReadableStream<Uint8Array>`.

### Segment tree building

The renderer walks the JSX tree and builds an intermediate `Segment` tree rather than emitting HTML directly. There are three segment types:

- `static` — a raw HTML string
- `composite` — an array of child segments
- `frame` — a frame boundary with potentially async content

```ts
let root = buildSegment(node, context, '')
await resolveBlocking(root)
let html = serializeSegment(root)
let finalHtml = finalizeHtml(html, context)
let bytes = encoder.encode(finalHtml)
controller.enqueue(bytes)
```

### Blocking vs non-blocking frames

When `buildSegment` encounters a `<Frame>` component, it calls `resolveFrame(src)` to fetch the frame's content:

- **Blocking** (no `fallback` prop): Must resolve before any HTML is sent. Content is inlined in the initial chunk.
- **Non-blocking** (has `fallback` prop): Fallback HTML is inlined immediately. Real content streams later as `<template>` elements.

```ts
let nonBlocking = !!props.fallback
if (nonBlocking) {
  seg.content = buildSegment(props.fallback, context, framePath)
  let framePromise = Promise.resolve(context.resolveFrame(props.src)).then(
    async (resolved) => resolveFrameHtml(resolved),
  )
  context.pendingFrames.push({ frameId, promise: framePromise })
} else {
  seg.pending = Promise.resolve(context.resolveFrame(props.src)).then(async (resolved) => {
    let { html, tail } = await resolveFrameHtml(resolved)
    seg.content = staticSeg(html)
    if (tail) {
      context.blockingFrameTails.push(tail)
    }
  })
}
```

### Client entry components

Components marked with `clientEntry` render to HTML like normal but are wrapped in comment markers (`<!-- rmx:h:ID -->...<!-- /rmx:h -->`) and their props/module metadata are serialized.

### Finalization

After all blocking frames resolve, the segment tree is serialized to HTML. `finalizeHtml` injects:

- Collected `<head>` elements and `<style>` tags
- A `<script id="rmx-data">` JSON blob containing hydration data (`h`) and frame data (`f`)

### Streaming

The first chunk (the full document HTML) is enqueued. Then two things happen concurrently:

1. **Blocking frame tails** — if a blocking frame's `resolveFrame` returned a `ReadableStream`, only the first chunk was inlined; remaining chunks stream now
2. **Pending non-blocking frames** — as each resolves, it streams as a `<template id="frameId">` element appended to the document

Wire format:

```
[initial HTML with blocking frames inlined, non-blocking frames as fallback]
[blocking frame tail chunks, if any]
<template id="f...">...resolved non-blocking frame content...</template>
<template id="f...">...another one...</template>
```

---

## 2. Frames

Frames are the unit of independent content on both server and client. On the server, a `<Frame src="/fragments/cart-items" />` triggers a sub-render (via `resolveFrame`) whose output gets embedded within comment markers:

```html
<!-- rmx:f:fABC12345 -->
...frame content HTML...
<!-- /rmx:f -->
```

The frame data (src, status, name) is recorded in the `rmx-data` JSON blob so the client knows where each frame lives and what its source URL is.

### Server-side frame resolution

In the bookstore demo, frames are resolved internally through the router itself (no network hop):

```ts
export function render(node: RemixNode, init?: ResponseInit) {
  let context = getContext()
  let request = context.request
  let router = context.storage.get(routerStorageKey)

  let stream = renderToStream(node, {
    resolveFrame: (src) => resolveFrame(router, request, src),
  })
  // ...
}
```

### Nesting

Frames can contain other frames. A blocking frame's content may itself contain non-blocking frames, which is why blocking frame "tails" exist — the inner frame `<template>` elements need to stream after the initial document.

---

## 3. Hydration

On the client, `run(document, init)` boots the runtime:

```ts
export function run(doc: Document, init: RunInit): AppRuntime {
  let styleManager = defaultStyleManager
  let errorTarget = new EventTarget()
  let scheduler = createScheduler(doc, errorTarget, styleManager)

  let resolveFrame: ResolveFrame =
    init.resolveFrame ?? (() => '<p>resolve frame unimplemented</p>')

  let frame = createFrame(doc, {
    src: doc.location?.href ?? '/',
    loadModule: init.loadModule,
    resolveFrame,
    pendingClientEntries: new Map(),
    scheduler,
    styleManager,
    data: {},
    moduleCache: new Map(),
    moduleLoads: new Map(),
    frameInstances: new WeakMap(),
    namedFrames: new Map(),
  })

  return Object.assign(errorTarget, {
    ready: () => frame.ready(),
    flush: () => frame.flush(),
    dispose: () => frame.dispose(),
  })
}
```

This creates the **top-level frame** for the entire document. The frame's `hydrateInitial()` runs immediately and does three things:

### a) Discovers sub-frames

`createSubFrames` walks the DOM looking for `<!-- rmx:f:... -->` comment pairs. For each one, it reads the frame's metadata from the `rmx-data` blob and recursively creates a child `Frame` instance bound to that comment range. If a frame's status is `pending` (non-blocking), the child frame either:

- Finds its `<template>` element already in the DOM (it arrived during streaming) and renders that content
- Sets up a `MutationObserver` to wait for the `<template>` to arrive and renders it when it does

### b) Discovers hydration markers

`scheduleHydrationInContainer` walks the DOM looking for `<!-- rmx:h:... -->` comment pairs. For each one, it reads the component metadata (module URL, export name, props) from `rmx-data` and kicks off module loading via `loadModule`.

### c) Hydrates components

Once a module loads, `hydrateRegion` creates a `createRangeRoot` (a virtual DOM root scoped to the comment range) and calls `root.render(vElement)`. This reconciles the server-rendered DOM with the component's virtual tree, attaching event listeners and wiring up `handle.update()` — all without replacing DOM nodes.

```ts
function hydrateRegion(vElement, start, end, context) {
  // ...
  let root = createRangeRoot([start, end], {
    scheduler: context.scheduler,
    frame: context.frame,
    styleManager: context.styleManager,
  })

  Object.defineProperty(start, '$rmx', { value: root, enumerable: false })
  root.render(vElement)
}
```

The key insight: hydration doesn't re-render HTML. It walks the existing server-rendered DOM and overlays the virtual component tree on top of it, connecting event handlers and establishing the component lifecycle.

---

## 4. UI Updates from Frame Reload

When a `clientEntry` component calls `handle.frame.reload()`:

### a) Fetch new content

The frame's `reload` method aborts any in-flight reload, then calls `resolveFrame(src, signal)`. On the client this is a `fetch()` call:

```ts
reload: async () => {
  reloadController?.abort()
  let controller = new AbortController()
  reloadController = controller
  frame.dispatchEvent(new Event('reloadStart'))
  try {
    let content = await init.resolveFrame(frame.src, controller.signal)
    if (reloadController !== controller || controller.signal.aborted) return controller.signal
    await render(content, { signal: controller.signal })
    return controller.signal
  } finally {
    if (reloadController === controller) {
      frame.dispatchEvent(new Event('reloadComplete'))
    }
  }
}
```

### b) Render the response

The frame's internal `render` function handles the response. If it's a `ReadableStream`, it incrementally processes chunks via `renderFrameStream`. If it's an HTML string fragment:

1. Parses it into a `DocumentFragment`
2. Hoists any `<head>` elements (styles, meta tags) to `document.head`
3. Merges any new `rmx-data` from the fragment
4. **Diffs the DOM** — `diffNodes` compares the frame's current DOM children against the new fragment's children, applying minimal mutations (add, remove, update attributes, update text)

```ts
let fragment =
  typeof content === 'string' ? createFragmentFromString(container.doc, content) : content
hoistHeadElements(container.doc, fragment)
mergeRmxDataFromFragment(context.data, fragment)

diffNodes(container.childNodes, Array.from(nextContainer.childNodes), {
  ...context,
  regionTailRef: container.regionTailRef,
  regionParent: container.regionParent,
})

scheduleHydrationInContainer(container, context, options?.initialHydrationTracker)
createSubFrames(container.childNodes, context)
```

### c) Re-hydrate new components

After diffing, `scheduleHydrationInContainer` discovers any new `<!-- rmx:h:... -->` markers in the updated DOM and hydrates them (loading modules, creating virtual roots). `createSubFrames` discovers any new nested frames.

### d) Preserve existing hydrated regions

The DOM differ (`diffNodes` in `diff-dom.ts`) is aware of hydration boundaries. When it encounters `<!-- rmx:h:... -->` markers that already have a virtual root (`$rmx` property), it preserves those ranges and their component state. So if a frame reloads and the new HTML still contains the same `clientEntry` component, the existing hydrated instance survives the diff with its state intact and is re-rendered with the new serialized prop payload.

---

## Summary

The server builds a segment tree, resolves frames (blocking first, then streaming non-blocking), and sends HTML with embedded comment markers and a JSON metadata blob. The client boots a frame for the document, walks the DOM to find frame boundaries and hydration markers, loads component modules, and overlays virtual roots on existing DOM. When a frame reloads, the new HTML is fetched, diffed against the current DOM, and any new components are hydrated while existing ones are preserved and re-rendered with new props from the payload.


mjackson:
LGTM, nice work @ryanflorence.

I think this PR has me finally convinced that the `router` needs to live on the request context (or we should at least have something like `context.fetch()`), so you don't have to do the acrobatics of storing it in `context.storage` and then pulling it out again. I'll make a PR and ping you on it.

demos/bookstore/app/router.ts
```diff
@@ -41,7 +44,15 @@ middleware.push(asyncContext())
 
 export let router = createRouter({ middleware })
 
+// Make router available to render() for internal frame resolution (no network).
+middleware.unshift((context: any, next: any) => {
+  context.storage.set(routerStorageKey, router)
+  return next()
+})
+
```
mjackson, line 52: I'm going to put router on `context.router` so you won't need to do this.

PR #11057, opened 2026-02-11
Add data-table core and SQL adapters

mjackson:
This PR introduces the new `data-table` relational query package and adapter packages for PostgreSQL, MySQL, and SQLite.

It includes:

- New core package: `@remix-run/data-table`
  - Typed table modeling with relations (`hasMany`, `hasOne`, `belongsTo`, `hasManyThrough`)
  - Query builder for reads/writes (`where`, joins, eager `with`, `insert/update/delete/upsert`, transactions)
  - Runtime validation through `@remix-run/data-schema`
  - Raw SQL escape hatch (`sql`, `rawSql`, `db.exec`)
  - Memory adapter for unit tests/reference behavior
- New adapter packages:
  - `@remix-run/data-table-postgres`
  - `@remix-run/data-table-mysql`
  - `@remix-run/data-table-sqlite`

This PR is **stacked on top of #11042** (`Add data-schema package`) and depends on it being merged first.


sergiodxa:
Do you plan to add support for other adapters later or do you think this will be responsibility of the community/db providers? I would like to use this with Cloudflare D1 instead of Drizzle once shipped, the adapter would need to receive the D1 binding, then I imagine it should be similar to SQLite one (maybe limited).

mjackson:
@sergiodxa I would welcome contributions if you'd like to make another package for Cloudflare D1. I just started with the classic 3 because it's a good baseline. But there are many more databases I'd like to support.

packages/data-table-mysql/package.json
```diff
@@ -0,0 +1,59 @@
...
+  },
+  "devDependencies": {
+    "@remix-run/data-schema": "workspace:*",
+    "@remix-run/data-table": "workspace:*",
+    "@types/node": "catalog:",
+    "@typescript/native-preview": "catalog:"
+  },
+  "dependencies": {
+    "@remix-run/data-table": "workspace:^",
+    "mysql2": "^3.15.3"
```
brophdawg11, line 46: We should move this to an optional `peerDep` so it doesn't get installed out of the box for `npm i remix`

lucasamonrc:
> @sergiodxa I would welcome contributions if you'd like to make another package for Cloudflare D1. I just started with the classic 3 because it's a good baseline. But there are many more databases I'd like to support.

Hi @mjackson! CF employee here. I decided to give a shot here #11522. Happy to create a proper issue and link it in the PR if needed.

PR #11069, opened 2026-02-15
Add data-model package and migrate bookstore models

mjackson:
Note: This is experimental work that shows an alternative to the data-table package. I don't want to ship BOTH data-table and data-model. There should be a single way we recommend people work with data in the database.

This PR introduces a new `remix/data-model` package that brings an ergonomic, ActiveRecord-style model layer to Remix.

It also shows how to use **request-scoped database binding without async local storage**. Instead of reaching into implicit global state, handlers get strongly typed model classes on `context.models` that are bound to the current request's database connection.

## What this PR adds
- New `@remix-run/data-model` package (exported as `remix/data-model`)
- A base `Model` class with static CRUD methods:
  - `find`, `findOne`, `findMany`, `all`, `count`, `create`, `update`, `destroy`/`delete`
- Relation helpers on models:
  - `hasMany`, `hasOne`, `belongsTo`, `hasManyThrough`
- Request-scoped model binding via `createModelRegistry(...).bind(db)`
- Model instance hydration (`row -> class instance`) so model methods and `instanceof` work as expected
- Table inference from class name (snake_case + pluralized), backed by new inflection support in `data-table`
- Bookstore demo migration from module-style + ALS access to class-based, context-bound models
- Remix package exports and docs/tests for the new API

## Example
```ts
import { createDatabase } from 'remix/data-table'
import { Model, createModelRegistry } from 'remix/data-model'

class User extends Model {
  static columns = {
    id: s.number(),
    email: s.string(),
  }
}

class Order extends Model {
  static columns = {
    id: s.number(),
    user_id: s.number(),
    total: s.number(),
  }

  static byUser(userId: number) {
    return this.findMany({ where: { user_id: userId }, orderBy: ['id', 'asc'] })
  }
}

let registry = createModelRegistry({ User, Order })
let db = createDatabase(/* adapter */)

// Setup context.models so each request has type-safe access to its own db
// connection and models.
interface AppModels extends ReturnType<typeof registry.bind> {}

declare module 'remix/fetch-router' {
  interface RequestContext {
    models: AppModels
  }
}

let router = createRouter({
  middleware: [
    async (context, next) => {
      context.models = registry.bind(db)
      return next()
    },
  ],
})

router.get('/users/:id/orders', async ({ models, params }) => {
  let userId = Number(params.id)
  let [user, orders] = await Promise.all([
    models.User.find(userId),
    models.Order.byUser(userId),
  ])

  return Response.json({ user, orders })
})
```

### DIFF
--- a/packages/response/src/lib/compress.test.ts
+++ b/packages/response/src/lib/compress.test.ts
@@ -747,11 +747,16 @@ describe('compressResponse()', () => {
 
       // Test that data arrives before stream closes AND is valid SSE format
       let receivedData = await new Promise<string>((resolve, reject) => {
+        let timeoutMs = isWindows ? (encodingName === 'br' ? 6_000 : 2_000) : 500
         let timeout = setTimeout(
           () => {
-            reject(new Error(`Timeout: data not flushed - flush may not be working`))
+            reject(
+              new Error(
+                `Timeout (${timeoutMs}ms): data not flushed for ${encodingName} - flush may not be working`,
+              ),
+            )
           },
-          isWindows ? 2_000 : 500,
+          timeoutMs,
         )
 
         decompressed.once('data', (chunk) => {


sergiodxa:
Do you plan to keep it simple and go full ActiveRecord compatible?

The schema solves validation, I suppose you can define custom validations using model methods somehow.

What would be nice is to also have [callbacks](https://guides.rubyonrails.org/active_record_callbacks.html) like `after_create`

```ts
class User extends Model {
  static columns = { ... }
  static private afterCreate(user) {
    // this will give you a `user` model to do thing after it was created
  }
}
```

mjackson:
I think I'm not going to merge this work after all. data-table is more explicit in most cases and is a nice, simple model for communicating with the database.

@sergiodxa Yes, I do plan on adding callback support to data-table. It should be possible to add them to the table abstraction instead of introducing a separate model layer.

PR #11075, opened 2026-02-17
Fix SVG attribute serialization in remix/component to preserve filter semantics

brookslybrand:
## Summary

This PR fixes an SVG serialization issue in `remix/component` where certain SVG attributes were being normalized in ways that changed browser behavior, especially for filter effects. In some cases, attributes that need canonical SVG casing were emitted in a different form, causing filter coordinate systems and related rendering behavior to diverge from React/browser expectations.

The fix hardens SVG attribute normalization so output is consistent between SSR and client updates and aligns with canonical SVG behavior. This should resolve rendering issues like disappearing/mispositioned filter effects while reducing risk of similar SVG attribute regressions going forward.

## The problem

A filter authored like this:

```tsx
<svg>
  <defs>
    <filter
      id="timeline-shadow"
      filterUnits="userSpaceOnUse"
      x="0"
      y="0"
      width="1200"
      height="200"
    >
      <feGaussianBlur stdDeviation="2.5" />
    </filter>
  </defs>
</svg>
```

would be serialized with transformed attribute names like:

```tsx
<filter
  id="timeline-shadow"
  filter-units="userSpaceOnUse"
  x="0"
  y="0"
  width="1200"
  height="200"
>
  <feGaussianBlur std-deviation="2.5"></feGaussianBlur>
</filter>
```

For attributes such as `filterUnits`, browsers treat this differently than canonical `filterUnits`, which can lead to incorrect filter coordinate behavior and visual regressions (e.g. effects disappearing or rendering in the wrong bounds

## Root cause

`remix/component` used a generic SVG camelCase→kebab-case fallback with a small exception list. For example, `filterUnits` was not exempted, so output became `filter-units`, which browsers treat as a different attribute and therefore do not apply to SVG unit semantics.

## Test plan

- [x] `pnpm --filter @remix-run/component test -- --run src/test/vdom.svg.test.tsx src/test/stream.test.tsx src/test/hydration.attributes.test.tsx`
- [x] `pnpm run lint`
- [x] Manual browser probe confirmed canonical camelCase `*Units` attrs affect SVG animated enum semantics, while dashed forms do not for affected attrs.

### DIFF
--- a/packages/component/src/lib/diff-props.ts
+++ b/packages/component/src/lib/diff-props.ts
@@ -121,6 +121,12 @@ function normalizePropName(name: string, isSvg: boolean): { ns?: string; attr: s
   return normalizeSvgAttribute(name)
 }
 
+function toLocalName(attrName: string): string {
+  let separatorIndex = attrName.indexOf(':')
+  if (separatorIndex === -1) return attrName
+  return attrName.slice(separatorIndex + 1)
+}
+
 function clearRuntimePropertyOnRemoval(dom: Element & Record<string, unknown>, name: string): void {
   try {
     if (name === 'value' || name === 'defaultValue') {
@@ -159,7 +165,7 @@ export function diffHostProps(
       }
 
       let { ns, attr } = normalizePropName(name, isSvg)
-      if (ns) dom.removeAttributeNS(ns, attr)
+      if (ns) dom.removeAttributeNS(ns, toLocalName(attr))
       else dom.removeAttribute(attr)
     }
   }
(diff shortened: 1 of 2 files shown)


packages/component/src/lib/diff-props.ts
```diff
@@ -2,10 +2,9 @@ import { invariant } from './invariant.ts'
 import { processStyle, createStyleManager, normalizeCssValue } from './style/index.ts'
 import type { StyleManager } from './style/index.ts'
 import type { ElementProps } from './jsx.ts'
+import { normalizeSvgAttribute } from './svg-attributes.ts'
```
brookslybrand, line 5: Pull this logic from 1 place instead of duplicating it here and in `stream.ts`

packages/component/src/test/vdom.svg.test.tsx
```diff
@@ -97,6 +97,53 @@ describe('vnode rendering', () => {
       expect(removed.hasAttribute('stroke-linecap')).toBe(false)
     })
 
+    it('uses canonical semantics for critical SVG attributes', () => {
```
brookslybrand, line 160: Not sure what exactly is appropriate to test here. We could test every single svg attribute, which would blow up this test. That might be the right thing to do, but just wanted to check

PR #11078, opened 2026-02-18
test(data-table-sqlite): add compiler tests

gyx1000:
## Add test suite for SQLite SQL compiler

Hello Remix team 👋

I am currently working on an `oracledb` adapter, as I will need it in a professional context.  
Integration testing for this database is… challenging (I’ll let you guess why 😅).

For that reason, I would like to test the SQL compilers in isolation.  
This PR introduces a test suite for the SQLite compiler to validate generated SQL statements and bound parameters.

The main goal is to:
- increase confidence in the SQL compilation layer,
- make future refactors safer,
- and enable isolated validation of compiler behavior.

Not all use cases are covered yet for SQLite.  
I would also appreciate clarification on some specifics behavior:

- QueryBuilder's `update` appears to perform a `get` beforehand to ensure the entry exists in the database. It's not de case for delete
- `insertMany` with many empty values call `insert into "accounts" default values` once

Is this intentional?

If this direction is not something you’re interested in, feel free to close the PR without hesitation

### DIFF
--- a/packages/data-table-sqlite/src/lib/sql-compiler.test.ts
+++ b/packages/data-table-sqlite/src/lib/sql-compiler.test.ts
@@ -507,13 +507,12 @@ describe('sqlite sql-compiler', () => {
 
   describe('update statement', () => {
     it('compile for one', async () => {
-      await db.update(accounts, 1, {
+      await db.query(accounts).where({ id: 1 }).update({
         email: 'info@remix.run',
         status: 'enabled',
       })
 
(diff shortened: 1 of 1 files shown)


mjackson:
Thanks for the PR, @gyx1000. I pushed a small change that uses the QueryBuilder API directly instead of db.update(), which avoids the extra statement you were seeing.

As for your questions, I'm going to switch db.update() to not do a pre-read. That is incidental behavior that was added by the LLM that I do not agree with. Thanks for catching that.

PR #11082, opened 2026-02-20
feat(data-table): support cross-schema column resolution

gyx1000:
### ✨ Summary

This PR adds support for resolving data-table columns across schemas.

The main goal is to make column resolution more flexible when data comes from multiple schema sources, while keeping the existing behavior unchanged for single-schema scenarios.

### 🧠 Motivation

Some data-table use cases require columns that reference values defined in a different schema than the primary one.  
Before this change, column resolution assumed a single schema context, which made cross-schema scenarios harder or required workarounds.

This update introduces cross-schema resolution so that:

- columns can resolve data from related schemas
- configuration remains explicit and predictable
- existing tables continue to work without changes

### 🔧 What changed

- Add cross-schema column resolve logic
- Update resolver flow to support schema-aware lookup
- Adjust internal typing/handling to ensure compatibility
- Keep backward compatibility with current behavior

### 🧪 Validation

- Existing data-table behavior remains unchanged
- Cross-schema cases resolve correctly
- No breaking API changes

### DIFF
--- a/packages/data-table-mysql/src/lib/adapter.test.ts
+++ b/packages/data-table-mysql/src/lib/adapter.test.ts
@@ -179,6 +179,26 @@ describe('mysql adapter', () => {
     assert.match(statements[0].text, /`accounts`\.`id`\s*=\s*`billing`\.`invoices`\.`account_id`/)
   })
 
+  it('treats dotted select aliases as single identifiers', async () => {
+    let statements: Array<{ text: string; values: unknown[] }> = []
+
+    let connection = {
+      async query(text: string, values: unknown[] = []) {
+        statements.push({ text, values })
+        return [[], []]
+      },
+      async beginTransaction() {},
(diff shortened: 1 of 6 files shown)


mjackson:
Thanks for the PR, @gyx1000. I agree this is something we should support.

I did another pass to keep adapter behavior aligned and avoid alias regressions.

- MySQL + Postgres now handle schema-qualified table names across FROM/JOIN and write paths.
- Added cross-schema join tests for both adapters.
- Kept dotted aliases as a single alias token (so `select({ 'account.email': accounts.email })` keeps same behavior).
- Added regression coverage for dotted aliases in SQLite, MySQL, and Postgres.

mjackson:
Tiny correction to my previous note: `select({ 'account.email': accounts.email })` stays a single quoted alias token (it does not get split into a qualified identifier).

PR #11085, opened 2026-02-24
fix(headers): correct windows-1252 test assertion for Node 24.13.1

mhalikosen:
## Summary

The "correctly decodes windows-1252 encoded filename" test in `content-disposition.test.ts`
has been failing on the Windows CI runner since it picked up Node 24.13.1.

The test expects `'file\x80.txt'` but receives `'file€.txt'`. This is because Node 24.13.1
introduced proper WHATWG Encoding Standard compliance for windows-1252 via
nodejs/node#60893 and nodejs/node#61093. Under the standard, byte `0x80` maps to
`U+20AC` (Euro sign, `€`), not the C1 control character `U+0080`.

The assertion originally expected `€` but was changed to `\x80` in fa226ba75 as a
workaround for the old TextDecoder behavior. This PR reverts the assertion to the
spec-correct value.

The Ubuntu runner still passes because it is on 24.13.0, but it will also fail once
the runner image updates.

Failing CI run: https://github.com/remix-run/remix/actions/runs/22331705571

## Test plan

- [ ] `pnpm --filter @remix-run/headers run test` passes locally
- [ ] CI passes on both ubuntu-latest and windows-latest

### DIFF
--- a/packages/headers/.changes/patch.fix-windows-1252-test.md
+++ b/packages/headers/.changes/patch.fix-windows-1252-test.md
@@ -1,0 +1,7 @@
+Correct the windows-1252 test assertion to expect the Euro sign (€) for byte 0x80.
+
+The original assertion was changed from `€` to `\x80` in fa226ba75 to work around
+a Node.js bug where TextDecoder used a Latin-1 fast path for windows-1252, skipping
+the spec-defined mapping of 0x80 to U+20AC. Node 24.13.1 resolved this with proper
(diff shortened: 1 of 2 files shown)


mhalikosen:
CI is running Node v24.13.0 which still has the TextDecoder bug. Tests will pass once the repo's Node version is bumped to v24.13.1+.

mjackson:
Fixed in #11087

PR #11086, opened 2026-02-26
feat(data-table) add data-table-mssql adapter package

isaaccloos:
<details>
<summary>Previous yaps</summary>

> Drafting this to get it out there. I'm curious about the interest from the team in juggling a growing number of adapters (given how easy it was to create this one).
>
> I will update this desc with greater nuance after I sleep on it. There are a few outstanding nitpicks I have. Specifically the isolationLevel integration, and other `data-table` package Options bits. It was a bit touchy finding a balance between mirroring the other adapter implementations and fully embracing the `mssql` package apis.
>
> Any and all feedback is always welcome 😄
</details>

Drafting again for schema migration support. Sorry for the spam!

### DIFF
--- a/.github/workflows/data-table-integration.yaml
+++ b/.github/workflows/data-table-integration.yaml
@@ -7,6 +7,7 @@ on:
     paths:
       - '.github/workflows/data-table-integration.yaml'
       - 'packages/data-table/**'
+      - 'packages/data-table-mssql/**'
       - 'packages/data-table-postgres/**'
       - 'packages/data-table-mysql/**'
       - 'packages/data-table-sqlite/**'
@@ -16,6 +17,7 @@ on:
     paths:
       - '.github/workflows/data-table-integration.yaml'
       - 'packages/data-table/**'
+      - 'packages/data-table-mssql/**'
       - 'packages/data-table-postgres/**'
       - 'packages/data-table-mysql/**'
       - 'packages/data-table-sqlite/**'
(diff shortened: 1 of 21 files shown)


mjackson:
Thanks for the PR, @IsaacCloos. I think we are very interested in contributions like this, especially if you're willing to put in the work and help us maintain it as issues arise. I'm going to rebase on main and see if we can get this PR to go green.

isaaccloos:
Pleanty of fun to work on- interesting archecture. Forgive me for calling it EntityFramework-esque 😆!

- To account for the isolation level being session-scoped in sql server (some interesting reading on that in the attached links to [this blog](https://blog.pjsen.eu/?p=471)), we revert back to `read committed` before closing any transaction that set a custom isolation level. Since you can interleave transactions, we track which ones need resetting in the new field `#isolatedTransactions`. Transactions without a custom isolation level skip the reset entirely.

- `SET TRANSACTION READ ONLY` is invalid for mssql so I opted to just silently ignore it. I did note it in the README though.

---

> I think we are very interested in contributions like this, especially if you're willing to put in the work and help us maintain it as issues arise.

It would be a pleasure. I'm a sql server user, and would love to see this succeed. Just ping me, and I'll address bugs, requests, or any refactors required given a shift in the underlying packages ya'll are making 👍🏻<sup>_and if it wasn't meant to be, just blow it away!_</sup>

isaaccloos:
Some musings I collected while refactoring `data-table-mssql` to remain idiomatic and, more importantly, compatible with `data-table` and its adapter siblings:

1. Working with two north stars — the [Postgres adapter](https://github.com/remix-run/remix/tree/main/packages/data-table-postgres) and [Entity Framework](https://learn.microsoft.com/en-us/ef/core/) — led to some interesting negotiations between what I’m familiar with in a migration management tool and what Remix expects from its APIs. _Is there a parable about having two north stars leading to disaster?_
2. I regrettably had to make a change outside the proposed `data-table-mssql` package. The `LIMIT 1` clause in `hasMigrationJournal` is incompatible with the MSSQL dialect.
    https://github.com/remix-run/remix/blob/5c8e71206e5a4c8ecc4fe99ec848950a32477e61/packages/data-table/src/lib/migrations/journal-store.ts#L33-L49
3. I am slightly troubled by proposing code like this publicly without a proper disclaimer regarding automated migrations. Again, it’s my EF background weighing on my conscience as I take this out of draft. Please read what Microsoft themselves have to say about trusting auto-migration software without reviewing the generated scripts: https://learn.microsoft.com/en-us/ef/core/managing-schemas/migrations/applying?tabs=dotnet-core-cli

---

My solution to all of these points is simple! Please just ask, and I can elaborate on implementation details, examples, and references. I'm more than happy to iterate on this package, address bugs, etc. Thank you for reading 😄

isaaccloos:
If there is still interest in this feature, please let me know at what point you think it would be safe to start making the necessary changes. It looks like a lot of awesome updates have happened to the underlying packages since I last looked at this 👍🏻

PR #11095, opened 2026-02-27
Add CSRF and cross-origin protection middleware packages

mjackson:
This PR adds two request-protection middleware packages for `@remix-run/fetch-router` apps, re-exports them from `remix`, and documents when to use each approach.

- adds `@remix-run/cop-middleware` with `cop(options)` for tokenless browser-origin protection using `Sec-Fetch-Site` with `Origin` fallback, trusted origins, explicit bypass patterns, and custom deny handling
- adds `@remix-run/csrf-middleware` with `csrf()` and `getCsrfToken()` for session-backed synchronizer-token protection with origin checks and flexible token extraction
- exports both packages from `remix` as `remix/cop-middleware` and `remix/csrf-middleware`
- documents when to use `cop()`, `csrf()`, or both together, including middleware ordering for the layered case
- calls out the core caveats explicitly: `cop()` is a browser-origin guard that falls through when both `Sec-Fetch-Site` and `Origin` are missing, while `csrf()` keeps the token as the primary defense, allows missing origin headers by default, and treats query-param tokens as a compatibility fallback
- adds repo-local `update-pr` and `write-readme` skills and registers them in `AGENTS.md`

```ts
import { cop } from 'remix/cop-middleware'
import { csrf } from 'remix/csrf-middleware'
import { session } from 'remix/session-middleware'

let tokenlessRouter = createRouter({
  middleware: [cop()],
})

let sessionBackedRouter = createRouter({
  middleware: [session(sessionCookie, sessionStorage), csrf()],
})

let layeredRouter = createRouter({
  middleware: [
    cop({ trustedOrigins: ['https://admin.example.com'] }),
    session(sessionCookie, sessionStorage),
    csrf(),
  ],
})
```

### DIFF
--- a/packages/cop-middleware/.changes/minor.initial-release.md
+++ b/packages/cop-middleware/.changes/minor.initial-release.md
@@ -1,1 +1,7 @@
-Initial release of `@remix-run/cop-middleware`.
+Add the initial release of `@remix-run/cop-middleware`.
+
+- Expose `cop(options)` for browser-focused cross-origin protection using `Sec-Fetch-Site`
+  with `Origin` fallback.
+- Support trusted origins, explicit insecure bypass patterns, and custom deny handlers.
+- Allow apps to layer `cop()` ahead of `session()` and `csrf()` when they want both
+  browser-origin filtering and token-backed CSRF protection.
--- a/packages/csrf-middleware/.changes/minor.initial-release.md
+++ b/packages/csrf-middleware/.changes/minor.initial-release.md
@@ -1,1 +1,8 @@
-Initial release of `@remix-run/csrf-middleware`.
+Add the initial release of `@remix-run/csrf-middleware`.
+
+- Expose `csrf(options)` and `getCsrfToken(context)` for session-backed CSRF protection in
+  Remix apps that accept unsafe form submissions.
+- Validate a per-session token together with request origin metadata, with support for token
+  transport in headers, form data, and query params.
+- Allow apps to layer `csrf()` after `cop()` when they need stricter token-backed protection
+  on top of browser-origin filtering.
--- a/packages/remix/.changes/minor.remix.update-exports-1773182906847.md
+++ b/packages/remix/.changes/minor.remix.update-exports-1773182906847.md
@@ -1,2 +1,0 @@
-Added `package.json` `exports`:
- - `remix/cop-middleware` to re-export APIs from `@remix-run/cop-middleware`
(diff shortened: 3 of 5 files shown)


sergiodxa:
Have you consider using the Sec-Fetch-Site headers? It's sends automatically by browsers, and you don't need a token to send in every form and to store in a cookie/session

mjackson:
@sergiodxa I had not considered using the `Sec-Fetch-Site` headers. Seems like it's related, but I don't know enough about it to determine if it's a complete replacement for CSRF or a complementary technique.

sergiodxa:
The header is set with `same-site` (like in SameSite cookies), `same-origin` (same protocol, hostname and port), `cross-site` or `none` (user initiated, e.g. typing on the URL)

A middleware could let pass safe methods (GET, HEAD, OPTIONS), if it's same-site or same-origin let it pass too, for cross-site validate against a list of trusted origins, and for anything else reject the request.

trojanowski:
Yes, `Sec-Fetch-Site` can currently be used as a modern alternative to CSRF tokens. Go provides a [middleware using it in its standard library](https://pkg.go.dev/net/http#CrossOriginProtection) and Rails [added support for it](https://github.com/rails/rails/pull/56350) with the goal of eventually deprecating CSRF tokens. More information:

   * <https://words.filippo.io/csrf/>
   * <https://www.alexedwards.net/blog/preventing-csrf-in-go>

   If the plan would be to support modern, baseline browsers it would probably be enough to rely on that header ([baseline since March 2023](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site)) without the `Origin` header check mentioned in the articles linked above, which is there as a fallback for older browsers. Worth noting that the header is only sent over HTTPS/localhost though.

mjackson:
Thanks for the links, @trojanowski. It was very insightful reading!

Although I think the `Sec-Fetch-Site` header can definitely get us most of the way there, I tend to agree with the second author you linked to who says:

> So — in general — relying on http.CrossOriginProtection is not sufficient as your only protection against CSRF.

That article was written in October 2025, so I assume the information it contains is still relatively current. It assumes tokenless protection is secure only under deployment guarantees that Remix cannot assume at the framework layer (HTTPS and TLS 1.3 being the main ones). So I think this middleware is still relevant.

That being said, I think we should make it clear that this middleware is optional. We should also probably follow up with a lightweight cross-origin guard modeled around Go's `CrossOriginProtection` rather than discarding token-based middleware outright.

PR #11100, opened 2026-02-27
Add background job packages

mjackson:
This adds first-party background jobs to Remix with typed job definitions, a small scheduler/worker API, and durable SQL and Redis storage adapters. The runtime model is lease-based at-least-once processing with delayed execution, retries, priority, dedupe, failed-job retry, pruning, observability hooks, and transaction-aware scheduler writes when the storage supports transactions.

- adds `remix/job` with `createJobs`, `createJobScheduler`, and `createJobWorker`
- keeps the API strongly typed end-to-end: `jobs.sendEmail` is the enqueue target, payloads are validated with `remix/data-schema`, and scheduler write options are typed from the storage adapter
- supports delayed jobs, retries with jitter, priority, dedupe, cancellation, failed-job listing/retry, pruning, and scheduler/worker hooks
- adds `remix/job-data-table` with a built-in migration and transaction-aware writes for SQLite, PostgreSQL, and MySQL
- adds `remix/job-redis` with matching queue semantics backed by Redis hashes, sorted sets, and atomic scripts
- exports the new surface through `remix/job`, `remix/job/worker`, `remix/job-data-table`, and `remix/job-redis`, with integration coverage for SQLite, PostgreSQL, MySQL, and Redis

```ts
// app/jobs.server.ts
import { createJobs, createJobScheduler } from 'remix/job'
import { createDataTableJobStorage } from 'remix/job-data-table'
import { db } from './db.server.ts'

export let jobs = createJobs({
  sendEmail: {
    schema: s.object({ to: s.string(), subject: s.string() }),
    async handle(payload) {
      await mailer.send(payload.to, payload.subject)
    },
  },
})

export let storage = createDataTableJobStorage(db)

export let scheduler = createJobScheduler(jobs, storage, {
  onEnqueue(event) {
    metrics.count('job.enqueue', 1, { job: event.jobName })
  },
})
```

```ts
// app/routes/orders.ts
import { db } from './db.server.ts'
import { jobs, scheduler } from './jobs.server.ts'

export async function action() {
  await db.transaction(async (transaction) => {
    await db.query(orders).insert({ id: 'order-123' })

    await scheduler.enqueue(
      jobs.sendEmail,
      { to: 'a@example.com', subject: 'Order received' },
      {
        dedupeKey: 'order:123:receipt-email',
        dedupeTtlMs: 5 * 60 * 1000,
        priority: 10,
        retry: { maxAttempts: 5, strategy: 'exponential' },
        transaction,
      },
    )
  })
}
```

```ts
// app/job-worker.ts
import { createJobWorker } from 'remix/job/worker'
import { jobs, scheduler, storage } from './jobs.server.ts'

let worker = createJobWorker(jobs, storage, {
  concurrency: 10,
  pollIntervalMs: 1000,
  retention: {
    policy: {
      completedOlderThanMs: 7 * 24 * 60 * 60 * 1000,
      failedOlderThanMs: 30 * 24 * 60 * 60 * 1000,
    },
    intervalMs: 60_000,
  },
  onJobFailed(event) {
    logger.error('job failed', event.job.id, event.error)
  },
})

await worker.start()

let failedJobs = await scheduler.listFailedJobs({ limit: 20 })
await scheduler.retryFailedJob(failedJobs[0].id)
```


afoures:
question: when using the data-table as backend for the job scheduling, can it work with transactions?
like i only want to schedule the job if the transaction was committed!

mjackson:
@afoures Yes, absolutely. The data-table library already has support for transactions, so you can just wait until your transaction succeeds before scheduling your job.

Is that what you're looking for?

gyx1000:
Hello @mjackson, thanks for the jobs implementation

It’s a really useful feature, and I’m excited to start building my first jobs with it.

I have a question about schema compilation (compileSchema / applySchema).

I’m guessing this might be the early foundations of a future Remix 3 schema migration tool. I really like the migration API shape so far. Coming from the Django world, being able to switch databases without regenerating SQL migration files is incredibly valuable. I often use SQLite for local development and in-memory unit tests, while running integration tests and production on PostgreSQL (and sometimes Oracle).

Is this direction something you’re planning to continue exploring?

afoures:
@mjackson yes this is one way of doing it, but what if i want the transaction to wrap the enqueue, so that failure to enqueue prevent the transaction and the job is only save to the db if the transaction is committed

```ts
await db.transaction(async (tx) => {
  await tx.create(yyy, { /* ... */ })

  await scheduler.enqueue('something', { /* ... */ })

  await tx.create(xxx, { /* ... */ })
})
```

mjackson:
@gyx1000 Yes, I'm working on data migrations as we speak. That work will land before this work.

@afoures I think that should be fine. If the scheduler fails to enqueue a job, it should throw which would fail the whole transaction.

afoures:
unless one of the call after the enqueue fail, then the job is scheduled but the transaction is rolled back, in this case there is a mismatch between the 2, job was created but related data is not in database

mjackson:
@afoures You bring up a good point. I think we can find a way to add support for built-in transactions in the scheduler. I'll look into it.

mjackson:
Closing this for now, but we will probably revisit after remix 3 goes stable.

PR #11103, opened 2026-02-27
Add crawling/prerendering support to fetch-router

brophdawg11:
This PR adds a new `router.crawl()` API to fetch-router that enables first-class crawling/spidering of the application served by the router - enabling very easy pre-rendering (basically a built-in `wget`).  If we don't want to bloat the `router` API area we could also easily pull this into a `import { crawl } from 'remix/fetch-router/crawl'` sub-export.

The spidering logic is implemented with a minimal custom parser and benchmarked against [node-html-parser](https://www.npmjs.com/package/node-html-parser) via `pnpm i; cd packages/fetch-router/bench; pnpm bench`.  It's currently around 2-3x faster because it only implemented the the minimal capabilities needed to extract link/asset paths from a document.

I've concerted our docs generation to use this new API if you want to see it in action (`pnpm i; cd docs; pnpm run docs; pnpm prerender`), but API looks like this:

```ts
interface CrawlOpts {
  // paths to crawl (default ['/'])
  paths?: string[],
  // traverse outbound links (default true)
  spider?: boolean, 
  // filter paths to crawl (defaults to non-absolute paths)
  filter(href): boolean, 
  // URL variants to crawl that may not show up as outbound links
  variants?: (path) => Promise<string[]>,
};

for await (let { pathname, filepath, response } of router.crawl()) {
  await writeResponse(filepath, response);
  console.log(`Write path ${pathname} to file ${filepath}`);
}
```


brophdawg11:
Drop filter/variants for now.  Move crawl to a utility.

brophdawg11:
Squash merging into a separate branch so I can rebase onto the docs-updates branch and stack this PR

PR #11108, opened 2026-02-28
data-table migrations and lifecycle hooks

mjackson:
This PR lands the `data-table` migration system and completes the operation-model split across core + adapters.

## What this includes

- Adds first-class migrations under `remix/data-table/migrations`:
  - `createMigration(...)`
  - chainable `column` API (`import { column as c }`)
  - split migration handles in callbacks:
    - `db` for immediate DML/data operations
    - `schema` for migration DDL/planning operations
  - registry + runner + optional Node migration loader
- Finalizes operation naming split:
  - `DataManipulationOperation`
  - `DataMigrationOperation`
- Updates postgres/mysql/sqlite adapters to compile + execute migration operations
- Renames top-level table-definition API from `createTable(...)` to `table(...)`
- Adds table lifecycle hooks (`beforeWrite`, `afterWrite`, `beforeDelete`, `afterDelete`, `afterRead`) and table-level `validate(...)`
- Expands JSDocs, README docs, and adapter/core coverage

## Migration File Example

```ts
import { column as c, table } from 'remix/data-table'
import { createMigration } from 'remix/data-table/migrations'

let accounts = table({
  name: 'accounts',
  columns: {
    id: c.integer().primaryKey(),
  },
})

let users = table({
  name: 'users',
  columns: {
    id: c.integer().primaryKey(),
    account_id: c.integer().notNull(),
    email: c.varchar(255).notNull().unique(),
    created_at: c.timestamp({ withTimezone: true }).defaultNow(),
  },
})

export default createMigration({
  async up({ db, schema }) {
    await schema.createTable(accounts)
    await schema.createTable(users)

    await schema.createIndex(users, 'email', { unique: true })

    await schema.alterTable(users, table => {
      table.addForeignKey('account_id', accounts, 'id')
      table.addCheck("email like '%@%'")
      table.addIndex(['account_id', 'email'])
    })

    if (db.adapter.dialect === 'sqlite') {
      await db.exec('pragma foreign_keys = on')
    }

    await schema.plan(`update users set created_at = now() where created_at is null`)
  },

  async down({ schema }) {
    await schema.dropTable(users, { ifExists: true })
    await schema.dropTable(accounts, { ifExists: true })
  },
})
```

Notes:

- Methods that take a table accept either a table name string (`'app.users'`) or a `table(...)` object.
- Key/index columns accept both single and compound forms (`'id'` or `['tenant_id', 'id']`).
- Constraint/index names are optional; when omitted, `data-table` generates deterministic names.

## Lifecycle Callback Example

```ts
import { column as c, fail, table } from 'remix/data-table'

let users = table({
  name: 'users',
  columns: {
    id: c.integer().primaryKey(),
    email: c.varchar(255).notNull(),
    role: c.enum(['customer', 'admin']).notNull(),
    created_at: c.timestamp({ withTimezone: true }).defaultNow(),
  },

  beforeWrite({ value }) {
    if (typeof value.email === 'string') {
      return {
        value: {
          ...value,
          email: value.email.trim().toLowerCase(),
        },
      }
    }

    return { value }
  },

  validate({ operation, value }) {
    if (operation === 'create' && typeof value.email !== 'string') {
      return fail('Email is required.', ['email'])
    }

    return { value }
  },

  afterWrite({ operation, affectedRows }) {
    console.log('users', operation, 'affected rows:', affectedRows)
  },

  beforeDelete({ where }) {
    if (where.length === 0) {
      return fail('Refusing to delete all users at once.')
    }
  },

  afterRead({ value }) {
    return {
      value: {
        ...value,
        email: typeof value.email === 'string' ? value.email.toLowerCase() : value.email,
      },
    }
  },
})
```

## Runner Example

```ts
import path from 'node:path'
import { Pool } from 'pg'
import { createPostgresDatabaseAdapter } from 'remix/data-table-postgres'
import { createMigrationRunner } from 'remix/data-table/migrations'
import { loadMigrations } from 'remix/data-table/migrations/node'

let pool = new Pool({ connectionString: process.env.DATABASE_URL })
let adapter = createPostgresDatabaseAdapter(pool)
let migrations = await loadMigrations(path.resolve('app/db/migrations'))
let runner = createMigrationRunner(adapter, migrations)

await runner.up()
```

## Runtime Semantics

- `dryRun: true` compiles and returns SQL without applying changes.
- In `dryRun`, `schema.hasTable(...)` and `schema.hasColumn(...)` read live database state (no in-memory simulation of pending dry-run operations).
- When migration transactions are enabled, migration-time DDL (`schema.createTable(...)`, etc.), DML (`db.exec(...)` / query builder), and introspection all run in the same adapter transaction context.

### DIFF
--- a/demos/bookstore/app/data/setup.test.ts
+++ b/demos/bookstore/app/data/setup.test.ts
@@ -6,6 +6,10 @@ import { loadMigrations } from 'remix/data-table/migrations/node'
 
 import { db, initializeBookstoreDatabase } from './setup.ts'
 
+function getRows(result: { rows?: Record<string, unknown>[] }): Record<string, unknown>[] {
+  return result.rows ?? []
+}
+
 function readRowString(row: Record<string, unknown>, key: string): string {
   let value = row[key]
 
@@ -42,10 +46,11 @@ describe('bookstore database setup', () => {
     let migrations = await loadMigrations(migrationsPath)
 
     let journalResult = await db.exec(sql`select id, name from data_table_migrations order by id asc`)
-    let journalIds = journalResult.rows.map((row) => readRowString(row, 'id'))
+    let journalRows = getRows(journalResult)
+    let journalIds = journalRows.map((row) => readRowString(row, 'id'))
     let migrationIds = migrations.map((migration) => migration.id)
 
-    assert.equal(journalResult.rows.length, migrations.length)
+    assert.equal(journalRows.length, migrations.length)
     assert.deepEqual(journalIds, migrationIds)
 
     assert.equal(await db.adapter.hasTable({ name: 'books' }), true)
@@ -65,20 +70,24 @@ describe('bookstore database setup', () => {
       sql`select name from sqlite_master where type = 'index' and name = 'order_items_order_id_idx'`,
     )
 
-    assert.equal(ordersIndex.rows.length, 1)
-    assert.equal(orderItemsOrderIndex.rows.length, 1)
+    assert.equal(getRows(ordersIndex).length, 1)
+    assert.equal(getRows(orderItemsOrderIndex).length, 1)
   })
 
   it('does not duplicate migration journal entries when initialized more than once', async () => {
     await initializeBookstoreDatabase()
 
     let before = await db.exec(sql`select count(*) as count from data_table_migrations`)
-    let beforeCount = readRowCount(before.rows[0], 'count')
+    let beforeRows = getRows(before)
+    assert.ok(beforeRows.length > 0)
+    let beforeCount = readRowCount(beforeRows[0], 'count')
 
     await initializeBookstoreDatabase()
 
     let after = await db.exec(sql`select count(*) as count from data_table_migrations`)
-    let afterCount = readRowCount(after.rows[0], 'count')
+    let afterRows = getRows(after)
+    assert.ok(afterRows.length > 0)
+    let afterCount = readRowCount(afterRows[0], 'count')
 
     assert.equal(afterCount, beforeCount)
   })


gyx1000:
Hello @mjackson ,

First of all, thank you for this piece of technology... really impressive work.

I’m wondering how you would approach the following use case:

Let’s imagine I want to create a `business-users` package to handle user management and reuse it across multiple Remix 3 projects. This package would ship with its own migration files and evolve over time.

From what I understand, loadMigrations expects a base migrations directory. I could import and re-export the migrations from `business-users` inside each Remix project, but that would require me to closely track every new migration file released with the package.

I’m also thinking this might be a similar need for packages like `@remix-run/job-data-table`, which would likely carry the definition of their own tables.

Do you think this kind of workflow is something that could be supported? If so, how would you approach it?

Thanks again for all your work.

mjackson:
@gyx1000 You can absolutely ship migrations as part of an npm package. You'll just have to stitch them together in your migrations registry, but it shouldn't be too difficult. The main requirement is that your migration ids must be unique and filenames need to be in the format that `loadMigrations()` recognizes.

Something like this could work as your migration runner:

```ts
import path from 'node:path'
import { createRequire } from 'node:module'
import { createMigrationRegistry, createMigrationRunner } from 'remix/data-table/migrations'
import { loadMigrations } from 'remix/data-table/migrations/node'

// your adapter
import { createPostgresDatabaseAdapter } from 'remix/data-table-postgres'
import { Pool } from 'pg'

const require = createRequire(import.meta.url)

function resolvePackageMigrationsDir(packageName: string, subdir = 'migrations'): string {
  // Works when package.json is resolvable from that package
  let packageJsonPath = require.resolve(`${packageName}/package.json`)
  return path.join(path.dirname(packageJsonPath), subdir)
}

let adapter = createPostgresDatabaseAdapter(
  new Pool({ connectionString: process.env.DATABASE_URL }),
)

// your app's migrations
let appMigrationsDir = path.resolve('app/db/migrations')

// your deps migrations
let billingMigrationsDir = resolvePackageMigrationsDir('@acme/billing-data')
let analyticsMigrationsDir = resolvePackageMigrationsDir('@acme/analytics-data')

let [appMigrations, billingMigrations, analyticsMigrations] = await Promise.all([
  loadMigrations(appMigrationsDir),
  loadMigrations(billingMigrationsDir),
  loadMigrations(analyticsMigrationsDir),
])

// Registry catches duplicate IDs across sources
let registry = createMigrationRegistry()

for (let migration of [...billingMigrations, ...analyticsMigrations, ...appMigrations]) {
  registry.register(migration)
}

let runner = createMigrationRunner(adapter, registry, {
  journalTable: 'app_migrations',
})

await runner.up()
```

If your package is changing over time you may want to apply smarter sorting of migrations so they retain the same order as you install new versions of your package, but you get the idea.

gyx1000:
Amazing ! 🎉

PR #11131, opened 2026-03-06
feat: no transform component option via htm patterns

jacob-ebey:


MichaelDeBoey:
I think this could possibly be confused with [`@remix-run/html-template`](https://github.com/remix-run/remix/blob/main/packages/html-template)

jacob-ebey:
Just use htm if you want this :)

PR #11132, opened 2026-03-06
fix(mime): prefer `video/mp4` for `.mp4` and `image/x-icon` for `.ico`

### LINKED ISSUE #11106: [mime] `mp4` resolves to `application/mp4` instead of `video/mp4`
Currently, the mime type returned for the `mp4` extension is `application/mp4`.
This is somewhat unexpected, as `video/mp4` is the mime type most commonly associated with standard `mp4` video files.

After looking into the codegen logic, this behavior makes sense: both [`application/mp4`](https://github.com/jshttp/mime-db/blob/6abfe5f81fde8fff83999ac030b9648367155338/db.json#L1197-L1200) & [`video/mp4`](https://github.com/jshttp/mime-db/blob/6abfe5f81fde8fff83999ac030b9648367155338/db.json#L9313-L9317) list `mp4` as an extension in `mime-db`.
Since both entries are sourced from IANA and are marked as non-compressible, the codegen ends up selecting the first one (`application/mp4`).

However, both [`application/mp4`](https://github.com/jshttp/mime-db/blob/6abfe5f81fde8fff83999ac030b9648367155338/src/iana-types.json#L1693-L1700) & [`video/mp4`](https://github.com/jshttp/mime-db/blob/6abfe5f81fde8fff83999ac030b9648367155338/src/iana-types.json#L11654-L11660) reference [RFC4337](https://tools.ietf.org/rfc/rfc4337.txt) in their sources.

In the section _"Selection of MIME Types for MP4 Files"_, it indicates that `application/mp4` (and `audio/mp4`) are intended for very specific cases and `video/mp4` is the general-purpose type for typical `mp4` video content.

I think it would make sense to prefer `video/mp4` for the `mp4` extension, since it represents the most common use case.

### LINKED ISSUE #11107: [mime] `ico` resolves to `image/vnd.microsoft.icon` instead of `image/x-icon`
This one is similar to #11106.

Currently, the mime type returned for the `ico` extension is `image/vnd.microsoft.icon`.
This is somewhat unexpected, as `image/x-icon` is the mime type historically used by Microsoft.

After looking into the codegen logic, this behavior makes sense: [`image/vnd.microsoft.icon`](https://github.com/jshttp/mime-db/blob/6abfe5f81fde8fff83999ac030b9648367155338/db.json#L8224-L8228) is sourced from IANA, while [`image/x-icon`](https://github.com/jshttp/mime-db/blob/6abfe5f81fde8fff83999ac030b9648367155338/db.json#L8320-L8324) comes from Apache.
Since IANA takes priority in the codegen, it ends up being selected.

However, according to [Wikipedia](https://en.wikipedia.org/wiki/ICO_%28file_format%29#MIME_type), `image/vnd.microsoft.icon` was registered with IANA in 2003 by a third party, while Microsoft itself historically uses `image/x-icon`.

Even though `image/vnd.microsoft.icon` is IANA-registered, I think it would make sense to prefer `image/x-icon` for the `ico` extension, given its historical usage.

mitre88:
## Summary

Add preferred MIME type overrides in the codegen for extensions where the algorithm picks an uncommon type:

- `mp4`: `application/mp4` → `video/mp4` per [RFC 4337 §2](https://tools.ietf.org/rfc/rfc4337.txt) which specifies `video/mp4` as the general-purpose type
- `ico`: `image/vnd.microsoft.icon` → `image/x-icon` which is the historically used type by Microsoft and browsers

### Changes
- `packages/mime/scripts/codegen.ts`: Add `preferredOverrides` map applied after the main algorithm
- `packages/mime/src/generated/mime-types.ts`: Updated generated output
- `packages/mime/src/lib/detect-mime-type.test.ts`: Updated test expectation

Fixes #11106
Fixes #11107


mitre88:
Gentle ping — this fix ensures correct MIME types for `.mp4` and `.ico` files. The change is minimal and well-scoped. Would appreciate a review when possible. Thanks!

PR #11139, opened 2026-03-09
fix(form-data-parser): handle non-ASCII filenames in multipart uploads

### LINKED ISSUE #11032: [form-data-parser] parseFormData fails with non-ASCII filenames (e.g., Japanese, Chinese, Korean)

mathewmeconry:
## Summary

Fixes #11032 

The Pull Request #11039 was written by an AI bot that got into some controversy and seems to be offline since then. This pull request continues the work of the previous pull request so it can be merged.

## Root Cause
The `parseRawHeaders` function in `@remix-run/headers` uses the native `Headers.append()` method, which throws a TypeError for non-ASCII characters in header values. This affects `Content-Disposition` headers containing non-ASCII filenames.

## Solution
1. Catch the encoding error: When `Headers.append()` throws for non-ASCII characters, catch the error (handling different runtime error messages for browsers and Node.js)

2. Percent-encode non-ASCII bytes: Encode non-ASCII characters as percent-encoded UTF-8 bytes (e.g., `テスト` → `%E3%83%86%E3%82%B9%E3%83%88`)

3. Decode when accessing filename: Update `ContentDisposition.preferredFilename` to decode percent-encoded UTF-8 back to Unicode characters

### DIFF
--- a/packages/headers/src/lib/content-disposition.test.ts
+++ b/packages/headers/src/lib/content-disposition.test.ts
@@ -226,3 +226,42 @@ describe('ContentDisposition.from', () => {
     assert.equal(result.filename, 'test.txt')
   })
 })
+
+describe('ContentDisposition non-ASCII filename handling', () => {
+  it('decodes percent-encoded UTF-8 Japanese filename', () => {
+    // Simulates what happens when raw-headers encodes テスト画像.png
+    let header = new ContentDisposition(
+      'form-data; name="file"; filename="%E3%83%86%E3%82%B9%E3%83%88%E7%94%BB%E5%83%8F.png"',
+    )
+    assert.equal(header.preferredFilename, 'テスト画像.png')
+  })
+
+  it('decodes percent-encoded UTF-8 Chinese filename', () => {
+    // Simulates what happens when raw-headers encodes 文件.png
+    let header = new ContentDisposition(
+      'form-data; name="file"; filename="%E6%96%87%E4%BB%B6.png"',
+    )
+    assert.equal(header.preferredFilename, '文件.png')
+  })
+
+  it('decodes percent-encoded UTF-8 Korean filename', () => {
+    // Simulates what happens when raw-headers encodes 파일.png
+    let header = new ContentDisposition(
+      'form-data; name="file"; filename="%ED%8C%8C%EC%9D%BC.png"',
+    )
+    assert.equal(header.preferredFilename, '파일.png')
+  })
+
+  it('handles mixed ASCII and non-ASCII in filename', () => {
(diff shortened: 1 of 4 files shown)


mathewmeconry:
Hey @MichaelDeBoey could you enable the ci runs so I can check if the pull is ready for review?

mjackson:
Thanks for picking this up and keeping the fix moving from #11039.

I’m going to close this in favor of #11263. The issue is real, but this approach is incomplete because it works around native `Headers` by percent-encoding non-ASCII values and later decoding filenames, which changes the semantics of literal percent sequences and only really solves the filename path. Multipart part headers are body metadata and can contain decoded UTF-8 values that native Headers cannot represent, so #11263 changes MultipartPart.headers away from native Headers entirely and stores decoded strings directly. That fixes the filename bug while also preserving non-ASCII field names/header values without relying on a header-specific encoding workaround.

PR #11143, opened 2026-03-10
Documentation Updates

brophdawg11:
A bunch of Remix API documentation updates

- CSS style overhaul from Tim to align with our theme
- Dark mode toggle removed in favor of always using the system preference
- Added content fading and a progress bar during client side frame reloads
- Main content moved before sidebar in DOM order for better screen reader/focus order (no UI change)
- Show signatures for classes in documentation
- Handle JSDoc comments for TS overloads on class methods and functions
- Removed some noisy console warnings around JSDoc comments for types/interfaces which are often self documenting

PR #11144, opened 2026-03-10
Address gaps in documentation via agent skill

brophdawg11:
**Stacked on https://github.com/remix-run/remix/pull/11143**

This PR adds an agent skill to help identify gaps in our documentation and get them fixed up.  Our `pnpm run docs` script logs warnings for a few types of gaps:

- Types that are referenced by public APIs but not exported and therefore not documented by typedoc
  - Ex: we export `createRouter(opts: RouterOptions)` but forget to export `RouterOptions`
- Public APIs where we are missing JSDocs
  - Ex: we export `createRouter` and `RouterOptions` but don't include JSDocs on them

The skill begins by removing all type exports, then it runs the docs script to identify the gaps - exports all the required APIs, and then adds JSDocs comments to any exported APIS that don't have them.

This PR runs the skill on the `fetch-proxy` and `fetch-router` packages to start.  `fetch-proxy` was OK so the changes were a net-zero.  `fetch-router` gets a handful of new exported types and a few JSDoc comments.

### DIFF
--- a/docs/src/generate/typedoc.ts
+++ b/docs/src/generate/typedoc.ts
@@ -5,7 +5,6 @@ import { debug, getApiNameFromFullName, info, invariant, verbose, warn } from '.
 type Maps = {
   comments: Map<string, typedoc.Reflection> // full name => TypeDoc Reflection
   apisToDocument: Set<string> // APIs we should generate docs for
-  apisMissingComments: Map<string, typedoc.Reflection> // APIs picked up by TypeDoc but missing JSDoc comments
 }
 
 export async function loadTypeDoc(opts: {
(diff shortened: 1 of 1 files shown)


brookslybrand:
Just FYI skills need to be defined in a`.agents/skills` directory. Maybe also add a symlink to `.claude/skills`. It's dumb, but most of them are at least aligning around `.agents/`. You can also always point out where to look for them in `AGENTS.md`

brophdawg11:
yeah probably - I just stuck this next to where @mjackson put his skills for now and have been referencing it directly.  It's not really a common-use "feature development" skill so tbh I don't know if we want every agent session being made aware of it automatically?

skills/docs-gaps/SKILL.md
```diff
@@ -0,0 +1,28 @@
+---
+name: docs-gaps
+description: Identify gaps in documentation by ensuring only necessary types are exported and that exported types have valid JSDocs comments. This skill helps maintain accurate and concise API documentation by checking for unreferenced exports and adjusting them accordingly.
+---
+
+# Documentation Gaps
```
brophdawg11: In addition to docs gaps - we should have an api docs skill for the agent to write good api docs
Also maybe a package structure skill - where files go etc.

skills/docs-gaps/SKILL.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Overview
+
+Use this skill to identify gaps in documentation by ensuring only necessary types are exported and that exported types have valid JSDocs comments.
+
+## Workflow
+
+This workflow applies for a single package. If operating on multiple packages, repeat for each package and commit individually. All work should be done from the `docs/` directory for proper execution of scripts.
+
+### Phase 1 - Make sure only necessary types are exported
```
brophdawg11: Drop this phase - mark apis with `@private` if that eliminates typedoc warnings

We only need step 2 to make sure we have jsdocs on public/exported APIs

PR #11148, opened 2026-03-11
route-pattern: optimize param types for `RoutePattern.{href,match}`

pcattori:
Improves type inference perf for `RoutePattern.{href,match}` and `Params` by ~2-5x depending on route patterns.

## Context

Previously, our types parsed a route pattern source into an AST and then computed params based on that AST, mirroring the logic of the runtime implementation.
But creating an AST in pure types was extremely costly at compile-time, resulting in slow type inference within editors/IDEs and when typechecking.

Now, we directly parse just the params from a route pattern source without creating intermediate AST. This drastically cuts down on our type instantiations — which are a good proxy for type inference perf.

Types have now gotten fast enough that my MBP w/ M1 Pro no longer gets `Type instantiation is excessively deep and possibly infinite. ts(2589)` for the benchmarks running against the Shopify patterns.

Below are benchmarks run against baselines on `main` branch. I've now updated the baselines as part of this PR.

## Type benchmarks for `href`

Screenshot 2026-03-11 at 2 28 19 PM

## Type benchmarks for `match`

Screenshot 2026-03-11 at 2 28 35 PM

### DIFF
--- a/packages/route-pattern/.changes/patch.type-inference-perf.md
+++ b/packages/route-pattern/.changes/patch.type-inference-perf.md
@@ -1,0 +1,5 @@
+Faster type inference for `RoutePattern.href`, `RoutePattern.match`, and `Params`
+
+Reduced type instantiations for parsing param types, resulting in
+~2-5x faster in relevant [type benchmarks](https://github.com/remix-run/remix/tree/main/packages/route-pattern/bench/types), but varies depending on your route patterns. 
+May fix `"Type instantiation is excessively deep and possibly infinite" (ts2589)` for some apps.


packages/route-pattern/src/lib/route-pattern/params.test.ts
```diff
@@ -34,7 +34,7 @@ export type Tests = [
   // Wildcards
   Assert<IsEqual<
     Params<'files/*'>,
-    { '*': string  }
```
pcattori, line 37: This "breaking change" fixes the types to be more accurate. We already [exclude nameless wildcards from the returned `params`](https://github.com/remix-run/remix/blob/main/packages/route-pattern/src/lib/route-pattern.test.ts#L888-L890)

PR #11149, opened 2026-03-12
Add `assets` package

### LINKED ISSUE #10795: Add `assets` middleware
The assets middleware takes a directory of files that are considered public assets for the frontend, uses them all as entry points in an [esbuild context](https://esbuild.github.io/api/#build). It runs the build and serves the results when a request comes in for one of the entry points.

When `env.SERVER_MODE === 'development'`, the assets middleware runs esbuild in [watch mode](https://esbuild.github.io/api/#watch) to automatically rebuild as the underlying assets change.

Note: this middleware should be stable enough to use in production in environments where the application source code is available on disk.

Because of the dependency on `esbuild`, this middleware should probably not be part of the `fetch-router` package.

markdalgleish:
This adds `@remix-run/assets` and exposes it from `remix/assets` as a Fetch-compatible server for compiling and serving browser JavaScript/TypeScript modules on demand.

The package maps public URL patterns to files on disk, compiles supported modules with Oxc, rewrites imports back to public URLs, and provides helpers for rendering module script hrefs and preload hints from source files.

- Adds `createAssetServer()` with `fetch()`, `getHref()`, `getPreloads()`, and `close()`.
- Supports reversible `fileMap` URL/file patterns, `allow`/`deny` access control, stable ETag-backed responses, and optional source-based fingerprinted URLs for immutable caching.
- Supports watch-mode cache invalidation for long-lived servers, plus `watch: false` for static production-style usage.
- Supports JS/TS/JSX/TSX/MTS/MJS transforms, TypeScript/JSX settings from nearby `tsconfig.json` files, minification, syntax targets, define replacements, external imports, and inline or external source maps.
- Includes package docs, changelog metadata, benchmark fixtures, and focused coverage for routing, access checks, fingerprinting, preloads, watch invalidation, transforms, source maps, and error handling.

Basic usage:

```ts

  rootDir: path.resolve(import.meta.dirname, '..'),
  fileMap: {
    '/assets/app/*path': 'app/*path',
    '/assets/npm/*path': 'node_modules/*path',
  },
  allow: ['app/assets/**', 'node_modules/**'],
})

let router = createRouter()

router.get('/assets/*', ({ request }) => {
  return assetServer.fetch(request)
})
```


brookslybrand:
Great work! I have a draft PR moving over remix.run from Vite to `remix/script-server` to test this out: https://github.com/remix-run/remix-website/pull/435

I had to move the css pipeline and other random asset loading out of Vite in some other PRs first. After those changes I was able to just get an agent to look at the bookstore demo and just adopt what it saw there without a whole lot of guidance.

Here's the staging site: https://remixdotrunstage.fly.dev/

Only thing that don't seem ideal, just to provide some feedback:
- this loads _a ton_ of modules now (76, 15 previously)
- there's an obvious JS loading waterfall with the hydrated components by default. Still investigating but definitely seems more pronounced than what we have. Not sure if there's more on the user's plate that they need to do to set things up properly

image

markdalgleish:
@brookslybrand 

- Having the potential for lots of modules is definitely a trade-off with this approach, but one we're deliberately making.
- We spoke about this offline, but for visibility - I've experimented with potential APIs for preloading client entry modules, but this is something we'll likely revisit. The most important thing right now is that you can render preload tags for the main client entry so everything bootstraps as quickly as possible.

brookslybrand:
Makes sense, thanks @markdalgleish for the follow up!

All in all seems I'm not noticing any regressions UX-wise from switching to the `script-loader`, seems like it works pretty well.

Also, fun fact, `remix.run` homepage now loads ~32% as much JS as it did when it was a React Router based website (46.8kb vs 148kb)! Of course this is the result of the whole Remix project, not necessarily this PR, just wanted to capture the stat somewhere

packages/assets/README.md
```diff
@@ -0,0 +1,294 @@
...
+  allow: ['app/client/**', 'app/shared/**'],
+  deny: ['app/**/*.server.*'],
+})
+```
+
+Rules for `allow` and `deny` are file paths or globs. Relative values are resolved from `root`. Absolute file paths match exactly, and absolute directory paths also match their descendants.
+
+## Routes
+
+Use `routes` to map public URLs to file paths.
```
mjackson:
```suggestion
Use `fileMap` to map file paths on disk to public URLs.
```

packages/assets/README.md
```diff
@@ -0,0 +1,294 @@
...
+import { createAssetServer } from 'remix/assets'
+
+let root = path.resolve(import.meta.dirname, '..')
+
+let assetServer = createAssetServer({
+  routes: [
+    { urlPattern: '/assets/app/*path', filePattern: 'app/*path' },
+    { urlPattern: '/assets/npm/*path', filePattern: 'node_modules/*path' },
+  ],
+  allow: ['app/client/**', 'app/shared/**', 'node_modules/**'],
```
mjackson:
```suggestion
  allow: ['app/assets/**', 'node_modules/**'],
```

packages/assets/README.md
```diff
@@ -0,0 +1,294 @@
...
+// [
+//   '/assets/app/client/entry.tsx',
+//   '/assets/app/client/search.tsx',
+//   '/assets/app/shared/utils.ts',
+//   '/assets/npm/@remix-run/component/index.js',
+//   ...etc
+// ]
+```
+
+### Watch Mode
```
mjackson: Switch the default to opt-out of watch mode.

mjackson (dismissed):
Looking good!

PR #11150, opened 2026-03-12
Add crawling/prerendering support to fetch-router

brophdawg11:
This PR adds a new `crawl(router, opts)` API to `fetch-router` that enables first-class crawling/spidering of the application served by the router - enabling very easy pre-rendering (basically a built-in `wget`).  

The spidering logic is implemented with a minimal custom HTML parser and benchmarked against [node-html-parser](https://www.npmjs.com/package/node-html-parser) via `pnpm i; cd packages/fetch-router/bench; pnpm bench`.  It's currently around 2-3x faster because it only implements the the minimal capabilities needed to extract link/asset paths from a document (no CSS selector support, etc).

```ts
import { crawl } from 'remix/fetch-router'

for await (let { pathname, filepath, response } of crawl(router)) {
  let outputPath = path.join('dist', filepath)
  await fs.mkdir(path.dirname(outputPath), { recursive: true })
  await fs.writeFile(outputPath, new Uint8Array(await response.arrayBuffer()))
  console.log(`Crawled ${pathname} -> ${outputPath}`)
}
```

I've converted our docs generation to use this new API as well

### DIFF
--- a/packages/fetch-router/src/lib/crawl.ts
+++ b/packages/fetch-router/src/lib/crawl.ts
@@ -1,0 +1,192 @@
+import { parse } from './html-parser.ts'
+import type { HTMLElement } from './html-parser.ts'
+
+const BASE_URL = 'http://localhost'
+
+export interface CrawlResult {
+  pathname: string
+  filepath: string
+  response: Response
+}
+
+export interface CrawlOptions {
+  /**
+   * Initial URL paths to put in the crawl queue.
+   * @default ['/']
+   */
+  paths?: string[]
+  /**
+   * Whether to follow outbound links found in HTML documents.
+   * @default true
+   */
+  spider?: boolean
+  /**
+   * Maximum number of concurrent requests.
+   * @default 1
+   */
+  concurrency?: number
+}
+
+export async function* crawl(
+  router: { fetch(request: Request): Promise<Response> },
+  options: CrawlOptions = {},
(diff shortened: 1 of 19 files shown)


packages/fetch-router/src/lib/crawl.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    if (error) throw error
+    if (results.length > 0) {
+      yield results.shift()!
+      continue
+    }
+    if (active === 0 && queue.length === 0) break
+
+    await gate
+  }
```
brophdawg11, line 101: This internal logic got a bit more complex to support `concurrency` since we need to be able to dispatch from multiple concurrent crawls into a single generator through this gate mechanism.

docs/src/server/routes.ts
```diff
@@ -4,7 +4,5 @@ export const routes = route({
   assets: '/(:version/)assets/*asset',
   docs: '/(:version/)api/*slug',
   home: '/(:version/)',
-  // Trailing slash is needed for proper HTML file responses on github pages
-  fragment: '/(:version/)fragment/(*slug/)',
```
brophdawg11: No longer need fragment routes with Ryan's new top level frame reload logic 👍

brophdawg11:
Closing - going to inline this `crawl` util to the docs folder and then it'll be an easy copy into fetch-router in the future if we want it

PR #11158, opened 2026-03-13
Add data-schema/form-data schemas

mjackson:
This adds a `remix/data-schema/form-data` export for parsing `FormData` and `URLSearchParams` with `remix/data-schema`. The form-data schema helpers now live with `data-schema`, and the demos use them where they improve parsing clarity or validation.

- adds `object`, `field`, `fields`, `file`, and `files` to `remix/data-schema/form-data`
- returns schemas that work with `s.parse()` and `s.parseSafe()` from `remix/data-schema`
- supports both `FormData` and `URLSearchParams` in `f.object(...)`
- updates the Remix export surface and package docs for the new subpath
- documents the helper roles and recommended defaulting pattern in the `data-schema` README
- updates Bookstore, fetch-router, SSE, and the `form-data-parser` Node demo to use the helpers where they add value

```ts
import * as f from 'remix/data-schema/form-data'
import * as checks from 'remix/data-schema/checks'
import * as coerce from 'remix/data-schema/coerce'

let input = s.parse(
  f.object({
    email: f.field(coerce.string().pipe(checks.email())),
    password: f.field(s.string().pipe(checks.minLength(8))),
    avatar: f.file(s.instanceof_(File)),
  }),
  formData,
)
```

```ts
let filters = s.parse(
  f.object({
    limit: f.field(s.optional(coerce.number())),
    tags: f.fields(s.array(s.string())),
  }),
  new URL(request.url).searchParams,
)
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -94,7 +94,7 @@ jobs:
         run: pnpm --filter @remix-run/component exec playwright install --with-deps
 
       - name: Run changed package tests
-        run: node ./scripts/test-changed-packages.ts origin/${{ github.base_ref }}
+        run: node ./scripts/detect-changed-packages.ts origin/${{ github.base_ref }}
 
   test-windows-main:
     name: test (windows-latest)
--- a/scripts/detect-changed-packages.ts
+++ b/scripts/detect-changed-packages.ts
@@ -1,7 +1,6 @@
 import * as cp from 'node:child_process'
 import * as fs from 'node:fs'
 import * as path from 'node:path'
-import { fileURLToPath, pathToFileURL } from 'node:url'
 
 type PackageInfo = {
   dirName: string
@@ -79,7 +78,9 @@ function parseArgs(args: string[]): CliOptions {
   }
 
   if (baseRef === '') {
-    throw new Error('Usage: node ./scripts/test-changed-packages.ts <base-ref> [head-ref] [--list]')
+    throw new Error(
+      'Usage: node ./scripts/detect-changed-packages.ts <base-ref> [head-ref] [--list]',
+    )
   }
 
   return { baseRef, headRef, listOnly }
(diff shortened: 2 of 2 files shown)


afoures:
will it work with URLSearchParams too?

gustavopch:
Should be pretty easy to support search params too as `FormData` and `URLSearchParams` are very similar.

This is how I'm doing in my React Router project:
<details>
<summary>Code</summary>

```ts
export type SearchParamsSchema = z.ZodObject<any>
export type FormDataSchema = z.ZodDiscriminatedUnion<any, string>

export const parseFields = <
  TSchema extends SearchParamsSchema | FormDataSchema,
>(
  fields: URLSearchParams | FormData,
  schema: TSchema,
): z.infer<TSchema> => {
  // Normalize and collect all values as arrays (ignore empty values)
  const fieldValues: { [key: string]: any } = {}
  for (const [key, rawValue] of fields.entries()) {
    const value = match(rawValue)
      .with(P.string, value => value.trim() || undefined)
      .with(P.instanceOf(File), value => (value.size === 0 ? undefined : value))
      .exhaustive()

    if (value !== undefined) {
      fieldValues[key] ??= []
      fieldValues[key].push(value)
    }
  }

  // Select the right schema to use if it's a discriminated union
  const selectedSchema: z.ZodObject<any> =
    schema instanceof z.ZodDiscriminatedUnion
      ? (schema.options.find(
          (option: z.ZodObject<any>) =>
            z.safeParse(
              option.shape[schema.def.discriminator],
              fieldValues[schema.def.discriminator]?.[0],
            ).success,
        ) ?? schema.options[0])
      : schema

  // Unwrap single values, fall back to [] when missing if an array is expected
  for (const key in selectedSchema.shape) {
    fieldValues[key] =
      selectedSchema.shape[key] instanceof z.ZodArray
        ? (fieldValues[key] ?? [])
        : fieldValues[key]?.[0]
  }

  return schema.parse(fieldValues) as z.infer<TSchema>
}
```

Then `parseFields()` is used internally by a `__GUARD__()` helper that I use in the routes:
```ts
const guarded = __GUARD__({
  access: 'onboarded',
  searchParams: v.searchParams({
    attachment: v.array(v.string()),
    description: v.array(v.string()),
    caption: v.optional(v.caption()),
  }),
  formData: v.formData({
    'generate-visuals': {
      description: v.array(v.string()),
      caption: v.optional(v.caption()),
    },
  }),
}).typedWith<Route.LoaderArgs, Route.ActionArgs>()

export const loader = guarded.loader(async args => { ... })

export const action = guarded.action({
  'generate-visuals': async args => { ... },
})
```
</details>

mjackson:
Yep, we can add support for `URLSearchParams` as well. Good call!

PR #11163, opened 2026-03-13
fix(ui): improve on() extracted handler inference

brookslybrand:
`on()` has two inference modes: JSX host context can infer the target element for inline handlers, and callers can pass an explicit target generic when the handler is created outside that context. The explicit-target path was too broad: `on<HTMLElement>(...)` defaulted the event type to the full event-name union instead of inferring it from the string literal argument.

That meant extracted handlers could be rejected even when the event name and handler were correct:

```tsx
let onClick = (event: MouseEvent) => {
  event.preventDefault()
}

on<HTMLElement>('click', onClick)
```

This updates the target-only overload to select from event-specific argument tuples, so the first argument still narrows the handler type after the target element is specified.

- preserves existing JSX host-context inference for inline `on()` handlers
- adds type coverage for valid extracted handlers and mismatched extracted handlers
- simplifies internal UI component event helpers that now only need the target generic for custom events

### DIFF
--- a/packages/ui/.changes/patch.on-extracted-handler-inference.md
+++ b/packages/ui/.changes/patch.on-extracted-handler-inference.md
@@ -1,0 +1,1 @@
+Fix `on()` type inference for extracted event handlers when passing only an explicit target element generic.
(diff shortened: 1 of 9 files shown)


pcattori:
Superseded by #11385

PR #11167, opened 2026-03-16
Tighten make-pr skill workflow

brookslybrand:
This tightens the `make-pr` skill around the fastest path to a credible PR, especially when working from a detached `HEAD` or a small, focused change.

- add fast-path checks for branch state before deeper prep
- encourage using the smallest useful context and getting to a clean commit quickly
- add a change-file decision point instead of assuming every PR needs release notes
- clean up the body template wording and markdown fencing so the example renders clearly

### DIFF
--- a/.agents/skills/make-pr/SKILL.md
+++ b/.agents/skills/make-pr/SKILL.md
@@ -9,18 +9,37 @@ description: Create GitHub pull requests with clear, reviewer-friendly descripti
 
 Use this skill to draft and open a PR with consistent, high-signal writing.
 Keep headings sparse and focus on the problem/feature explanation, context links, and practical code examples.
+Optimize for the shortest path to a credible PR, not the fullest possible context-gathering pass.
 
 ## Workflow
 
-1. Gather context from branch diff and related work.
+1. Check the fast-path blockers first.
+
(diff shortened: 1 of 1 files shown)


.agents/skills/make-pr/SKILL.md
```diff
@@ -9,18 +9,37 @@ description: Create GitHub pull requests with clear, reviewer-friendly descripti
...
+1. Get the branch into a PR-ready state quickly.
+
+- If changes are still uncommitted and the user wants a PR, branch first, then commit.
+- Prefer a single clean commit unless the user asks for a different history shape.
+
+1. Check whether this PR also needs a change file.
+
+- Do not assume every PR needs one.
+- Before opening the PR, decide whether the change is user-facing enough to require release notes in `packages/*/.changes`.
+- If a change file is needed or likely needed, use the `make-change-file` skill instead of re-deriving that workflow here.
```
brookslybrand, line 37: This is the main piece that I realized didn't automatically get picked up with some changes I made. It was easy enough to reference that skill myself once I realized that the change file was missing, but it'd be nice if when I say "open a PR" it checked and realized this was missing automatically

PR #11170, opened 2026-03-17
Enable automatic docs sync on publish workflows

brophdawg11:
Triggers a repository dispatch event for https://github.com/remix-run/remix-api-docs to generate the docs at the end of the publish workflow.  Plus some minor cleanup to the docs prerendering flow.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -84,7 +84,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Output tag
-        run: echo "tag=remix@$(node -p "require('./packages/remix/package.json').version")" >> $GITHUB_OUTPUT
+        run: echo "tag=$(git tag --points-at HEAD | grep -e '^remix@3')" >> $GITHUB_OUTPUT
 
   docs:
     name: Update API Docs


.github/workflows/publish.yaml
```diff
@@ -83,22 +83,22 @@ jobs:
...
+        uses: actions/github-script@v8
+        with:
+          github-token: ${{ secrets.REMIX_API_DOCS_PAT }}
+          script: |
+            await github.rest.repos.createDispatchEvent({
+              owner: 'remix-run',
+              repo: 'remix-api-docs',
+              event_type: 'update-docs',
+              client_payload: { tag: '${{ needs.publish.outputs.tag }}' }
+            });
```
brophdawg11, line 104: Trigger the remote workflow with the latest Remix tag after publish completes

docs/src/server/prerender.ts
```diff
@@ -17,21 +16,13 @@ let { values: cliArgs } = util.parseArgs({
       short: 'd',
       default: 'build/site',
     },
-    all: {
-      type: 'boolean',
-      short: 'a',
-    },
```
brophdawg11, line 23: I used this during original development but isn't necessary for our normal workflows where we'll prerender one version at a time (at publish time)

PR #11173, opened 2026-03-17
Consolidate Remix Component guidance into app skills

brookslybrand:
This consolidates the Remix Component guidance we want agents to use when building app UIs.

It adds a new top-level `app-skills/ui` skill with focused references for the current `remix/component` model, updates `packages/component/AGENTS.md` to match the mixin-based API surface, and removes the older package-local Component skills now that the guidance has moved up to `app-skills/`.

- add a reusable `ui` app skill for Component-driven app work
- promote `animate-elements` and `create-mixins` into top-level app skills
- refresh `packages/component/AGENTS.md` for `mix`, `css(...)`, `addEventListeners(...)`, `run({ ... })`, and `root.dispose()`
- remove the package-local Component skill copies so the guidance lives in one place

```md
Use the `ui` skill when building pages, layouts, client entries, interactions, styling, navigation, animations, reusable mixins, or UI tests in a Remix app.
```


brookslybrand:
This is an attempt to enhance #11100, however that PR needs to be rebased off of `main` first.

PR #11147, opened 2026-03-11
Add frame navigation APIs and link() mixin to Remix Component

ryanflorence:
This PR adds the frame-navigation runtime surface for Remix Component and updates the demo, docs, and package wiring around it so apps can drive partial-page navigation with either native links or host-aware mixins.

- adds `run({ loadModule, resolveFrame })`, `navigate(href, { src, target, history })`, `link(href, { src, target, history })`, `handle.frames.top`, `handle.frames.get(name)`, and SSR frame source context
- supports frame-targeted anchor and area navigation via `rmx-target` and `rmx-src`, `rmx-document` opt-out, and a `link()` mixin that gives buttons and other elements accessible link semantics with keyboard and modifier-click behavior
- expands the `frame-navigation` demo with auth, nested settings navigation, and dashboard stat cards that demonstrate `link()` on non-anchor elements
- fixes runtime edge cases uncovered along the way, including live DOM preservation during frame diffs, deferred mixin lifecycle work, balanced streamed frame updates, subtree disposal, and error forwarding, and updates benchmarks/docs to the current mixin-first APIs

```ts
let app = run({
  loadModule,
  resolveFrame,
})

await navigate('/settings/profile', {
  src: '/settings/profile',
  target: 'settings-shell',
})
```

```tsx
<a
  href={route.href()}
  rmx-target="settings-shell"
  rmx-src="/settings/profile"
>
  Profile
</a>

<button mix={link('/account')} />
<li mix={link('/calendar')} />
```

Tested with:
- `pnpm --filter @remix-run/component test -- src/lib/mixins/link-mixin.test.tsx src/test/hydration.attributes.test.tsx`
- `pnpm --filter frame-navigation-demo typecheck`
- `pnpm run lint`

PR #11177, opened 2026-03-18
fix(component): restore navigation api fallback

brookslybrand:
This restores the graceful fallback that lets `@remix-run/component` frame-navigation apps boot when the Navigation API is unavailable.

The fallback was added in `b14082938`, but it was dropped during the frame-navigation refactor in #11147 when `navigation.ts` started reading `window.navigation` unconditionally. That left older Safari/iOS versions and WebViews without `window.navigation` throwing during `run()` startup instead of degrading to normal document navigations.

- guard `window.navigation` before setting up the runtime listener
- fall back to `window.location.assign(...)` for imperative `navigate(...)` calls when the Navigation API is missing
- add regression coverage for `navigate()`, `startNavigationListener()`, and `run()` without `window.navigation`
- add a patch change file for the restored degradation behavior


brookslybrand:
I'm gonna close this in a second, just trying something

PR #11183, opened 2026-03-19
Remove unnecessary `as const` from `enum_()` usage

gustavoguichard:
Hey folks! 👋

The `enum_` function signature already uses a `const` type parameter (`<const values extends readonly [unknown, ...unknown[]]>`), which means TypeScript infers literal types automatically — no `as const` needed at call sites.

This PR removes the redundant `as const` from the docs and tests, and adds a type-level test to prove the inference works without it.

## Changes

- Removed `as const` from 2 `enum_()` examples in README.md
- Removed `as const` from 4 `enum_()` calls in schema.test.ts
- Added a type-level test asserting that `enum_(['active', 'inactive', 'pending'])` infers `'active' | 'inactive' | 'pending'` (not `string`), and similarly for numeric enums


mjackson:
Thanks, @gustavoguichard!

PR #11189, opened 2026-03-20
fix(component): correct DOM nodes ordering & fragment parentage in vnode hierarchy

### LINKED ISSUE #11061: [component] Incorrect DOM node ordering when calling handle.update() after the app has hydrated
This one is a little contrived, but I've got a minimal reproduction here: https://github.com/alexanderson1993/remix/tree/renderToStream-incorrect-dom-order/demos/renderToStream-incorrect-dom-order

You can run that by checking out that branch, navigating to that demo, running `pnpm i` and then `pnpm dev` and visiting the server address.

Here's the setup: You've got an app that's server-rendered with `renderToStream`. You want to use a browser API, like IndexedDB or Geolocation, so you initialize some component state with `null`, and then call the browser API inside `if (typeof window !== "undefined")`.

```ts
    let data: string | null = null

    // Simulating running some kind of browser-only API, like IndexedDB or Geolocation
    if (typeof window !== 'undefined') {
      setTimeout(() => {
        data = '🤖'
        handle.update()
      }, 500)
    }
```

And then in the component itself, we render `null` in the place where the data should go until hydration completes, the browser API promise resolves, and the component updates.

```tsx
    return () => (
      <div>
        <div>The robot should render after me</div>
        {data ? <div>{data}</div> : null}
        <div>
          The robot should <strong>not</strong> render after me
        </div>
      </div>
    )
```

The expected behavior is that you would see this in your browser when everything settles, where the `null` is replaced with the newly created DOM node.

```
The robot should render after me
🤖
The robot should not render after me
```

The actual behavior is that the DOM node is created and appended to the end of the component, like so:

```
The robot should render after me
The robot should not render after me
🤖
```

I've tested this with client-only `remix/component` using `createRoot(document.body).render(<App/>)` and it behaves as expected. It's only when rendering with `renderToStream`, `clientEntry`, and `run()`.

I'll also note that in my example, removing the `if (typeof window !== "undefined")` guard results in the server crashing with the following error:

```
/Users/alexanderson/Projects/remix/packages/component/src/lib/component.ts:199
    throw new Error('scheduleUpdate not implemented')
          ^
Error: scheduleUpdate not implemented
    at scheduleUpdate (/Users/alexanderson/Projects/remix/packages/component/src/lib/component.ts:199:11)
    at <anonymous> (/Users/alexanderson/Projects/remix/packages/component/src/lib/component.ts:214:9)
    at new Promise (<anonymous>)
    at Object.update (/Users/alexanderson/Projects/remix/packages/component/src/lib/component.ts:212:7)
    at Timeout._onTimeout (/Users/alexanderson/Projects/remix/demos/renderToStream-incorrect-dom-order/app/assets/component.tsx:12:14)
    at listOnTimeout (node:internal/timers:605:17)
    at process.processTimers (node:internal/timers:541:7)
```

I don't think that's relevant to this particular reproduction, since you would need to have the guard in place if you were using a browser API instead of my contrived `setInterval` example.

henryStelle:
Fixes https://github.com/remix-run/remix/issues/11061

Non-renderable nodes (`null`, `undefined`, `boolean`) are currently represented in the vDom as empty text nodes. However, the server serialization completely omits them from the output. Hence, if these nodes are ever replaced in a later update on the client with renderable content, there is no `_dom` reference to use to determine where to insert the new content. Additionally, the `findNextSiblingDomAnchor` check currently also fails as the `vParent` is the parent of the `next` vNode, not the current. Therefore, nodes end up being amended to the end of their parent's children by default as shown in https://github.com/remix-run/remix/issues/11061.

This change:
- Adds a `NON_RENDER_NODE` to represent nodes that the code shouldn't try to find in the DOM
- Forces `findNextSiblingDomAnchor` to always use the current parent by not passing it as a second parameter, but reading it from the `curr` parameter
- Allows `findNextSiblingDomAnchor` to recursively search up the tree to find the next sibling if its parent is a fragment. This is because a node might not have a sibling within the fragment, but when the fragment is flattened at the DOM level, it does have one. Therefore, going up a level allows that possible next node to be found.
- Fixes bug where the parent of vNode in a fragment was set to the parent of the fragment, not the fragment itself
    - This may have been intentional. However, as the vParent's children array isn't been updated to reflect these grandchildren, `findNextSiblingDomAnchor` fails to find all of the siblings. I tried update the vParent to track its grandchildren with the hope of being able to remove the recursion, but that resulted in duplication between having the grandchildren and their containing fragment, which really broke the diff'ing logic. 

For context:
This set of changes is the only one I found that worked, but I also tried:
- Filtering out non-renderable nodes rather than tracking them in a separate type as this _seems_ to be the method used by React and Preact (I might be wrong on that last point). However, this cases issues with needlessly tearing down components and recreating them when their previous sibling changed from non-renderable to renderable.
- Inserting a comment where the non-renderable node would be if it was rendered rather than relying on finding the next sibling. However, `skipComments` caused these comments to be skipped when looking for the anchor. Rather than updating this to not skip this comment, I abandoned this approach as it smelled off.

### DIFF
--- a/packages/component/.changes/patch.fix-hydration-out-of-order.md
+++ b/packages/component/.changes/patch.fix-hydration-out-of-order.md
@@ -1,0 +1,1 @@
+Fix DOM node ordering when calling handle.update() after the app has hydrated
--- a/packages/component/src/lib/reconcile.ts
+++ b/packages/component/src/lib/reconcile.ts
@@ -389,7 +389,7 @@ export function diffVNodes(
       frame,
       scheduler,
       styles,
-      vParent,
+      next,
       rootTarget,
       undefined,
       anchor,
--- a/packages/component/src/lib/reconcile.ts
+++ b/packages/component/src/lib/reconcile.ts
@@ -711,17 +711,7 @@ function insert(
   if (isFragmentNode(node)) {
     // Insert fragment children in order before the same anchor
     for (let child of node._children) {
-      cursor = insert(
-        child,
-        domParent,
-        frame,
-        scheduler,
-        styles,
-        vParent,
-        rootTarget,
-        anchor,
-        cursor,
-      )
+      cursor = insert(child, domParent, frame, scheduler, styles, node, rootTarget, anchor, cursor)
     }
     return cursor
   }
@@ -1658,7 +1648,7 @@ export function findNextSiblingDomAnchor(curr: VNode): Node | null {
     vParent._content && isFragmentNode(vParent._content)
       ? vParent._content._children
       : vParent._children
-  if (!children) return null
+  if (!children || children.length === 0) return findNextSiblingDomAnchor(vParent)
 
   let idx = children.indexOf(curr)
   if (idx === -1) return null
--- a/packages/component/src/lib/reconcile.ts
+++ b/packages/component/src/lib/reconcile.ts
@@ -23,6 +23,7 @@ import {
   isHostNode,
   isTextNode,
   findContextFromAncestry,
+  isNonRenderNode,
 } from './vnode.ts'
 import { invariant } from './invariant.ts'
 import { diffHostProps } from './diff-props.ts'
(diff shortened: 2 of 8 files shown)


henryStelle:
@ryanflorence, I would be interested in your thoughts on this approach.

mjackson:
Superseded by #11425.

PR #11190, opened 2026-03-20
fix(mime): prefer experimental over vendor-specific types

### LINKED ISSUE #11107: [mime] `ico` resolves to `image/vnd.microsoft.icon` instead of `image/x-icon`

KJyang-0114:
Closes #11107

## Summary

The `.ico` extension was resolving to `image/vnd.microsoft.icon` instead of the de facto standard `image/x-icon`. This was reported in issue #11107.

## Root Cause

The MIME type codegen had a tiebreaker rule that preferred `source=iana` over `source=apache` when two vendor types had equal compressibility. Since `image/vnd.microsoft.icon` has `source=iana` and `image/x-icon` has `source=apache`, the codegen always picked `image/vnd.microsoft.icon`.

However, `image/x-icon` is the historically preferred MIME type for ICO files used by browsers and tools.

## Fix

Added a codegen tiebreaker that prefers `x-` over `vnd.` types when both are experimental/vendor with equal compressibility. Regenerated `mime-types.ts` so `ico` now maps to `image/x-icon`.

## Testing

- Added test for `detectMimeType('ico') = 'image/x-icon'`
- All 63 existing tests pass

## Risk & Rollback

- **Risk**: Low. Only affects the `.ico` extension mapping.
- **Rollback**: Revert the codegen or generated file change.


MichaelDeBoey (commented):
I would say that for a generic solution, we would want to take vendor-specific (vnd.) over experimental (`x-`) tbh

But that unfortunately doesn't solve the `.ico` problem

mjackson:
Closing this as superseded by [#11132, which added preferred MIME overrides for `.mp4` and `.ico`](https://github.com/remix-run/remix/pull/11132/files), satisfying the original `.ico` concern without changing the generic vendor-vs-experimental preference rule. Thanks for the PR.

PR #11201, opened 2026-03-23
route-pattern: Matches return decoded values for params in pathname

pcattori:
Most routers (Express, Hono, find-my-way, Rails) return **decoded** params, whereas we were returning encoded params. With this PR, `route-pattern` returns decoded params too.

`.changes` file for this PR copied below 👇

---

Matches return decoded values for params in pathname

```ts
let pattern = new RoutePattern('/posts/:slug')

let url = new URL('https://blog.example.com/posts/💿')
pattern.match(url)?.params.slug
// Before -> '%F0%9F%92%BF'
// After -> '💿'

url = new URL('https://blog.example.com/posts/café-hà-nội')
pattern.match(url)?.params.slug
// Before -> 'caf%C3%A9-h%C3%A0-n%E1%BB%99i'
// After -> 'café-hà-nội'

url = new URL('https://blog.example.com/posts/北京')
pattern.match(url)?.params.slug
// Before -> '%E5%8C%97%E4%BA%AC'
// After -> '北京'

url = new URL('https://blog.example.com/posts/مرحبا')
pattern.match(url)?.params.slug
// Before -> '%D9%85%D8%B1%D8%AD%D8%A8%D8%A7'
// After -> 'مرحبا'
```

If you need percent-encoded text again, use JavaScript's native [`encodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI):

```ts
let url = new URL('https://blog.example.com/posts/💿')
let slug = pattern.match(url)!.params.slug
// -> 💿

encodeURI(slug)
// -> '%F0%9F%92%BF'
```

## Perf Benchmarks

`TrieMatcher` performance seems to be unchanged (within expected variance), though this feature did slow down `ArrayMatcher` enough that **now `TrieMatcher` is faster than `ArrayMatcher` for all `matching` benchmarks**:

Screenshot 2026-03-23 at 9 59 40 PM

### DIFF
--- a/packages/route-pattern/.changes/minor.non-ascii-pathname-params.md
+++ b/packages/route-pattern/.changes/minor.non-ascii-pathname-params.md
@@ -1,0 +1,36 @@
+Matches return decoded values for params in pathname
+
+```ts
+let pattern = new RoutePattern('/posts/:slug')
+
+let url = new URL('https://blog.example.com/posts/💿')
+pattern.match(url)?.params.slug
+// Before -> '%F0%9F%92%BF'
+// After -> '💿'
+
+url = new URL('https://blog.example.com/posts/café-hà-nội')
+pattern.match(url)?.params.slug
+// Before -> 'caf%C3%A9-h%C3%A0-n%E1%BB%99i'
+// After -> 'café-hà-nội'
+
+url = new URL('https://blog.example.com/posts/北京')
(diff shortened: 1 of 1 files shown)


machour:
شكرًا @pcattori

PR #11212, opened 2026-03-29
docs(form-data-parser): Fix error handling example in README

### LINKED ISSUE #11180: [form-data-parser]: Example of handling file upload errors is broken
Today while debugging I noticed, that the example given for the `form-data-parser` package has an error in it. The example suggests that you should check for errors like this:

```typescript
try {
  let formData = await parseFormData(request, {
    maxFiles: 5,
    maxFileSize: 10 * oneMb,
    maxParts: 25,
    maxTotalSize: 12 * oneMb,
  })
} catch (error) {
  if (error instanceof MaxFilesExceededError) {
    console.error(`Request may not contain more than 5 files`)
  } else if (error instanceof MaxHeaderSizeExceededError) {
    console.error(`Multipart headers may not exceed the configured size limit`)
  } else if (error instanceof MaxFileSizeExceededError) {
    console.error(`Files may not be larger than 10 MiB`)
  } else if (error instanceof MaxPartsExceededError) {
    console.error(`Request may not contain more than 25 multipart parts`)
  } else if (error instanceof MaxTotalSizeExceededError) {
    console.error(`Multipart request may not exceed 12 MiB of total content`)
  } else {
    console.error(`An unknown error occurred:`, error)
  }
}
```
This does not catch the error though, as on the following line the error instance is put onto a `FormDataParseError` instances `cause` variable. 

https://github.com/remix-run/remix/blob/8eb4af79d843c97a7373bc97b401a5cd6f3cd2d9/packages/form-data-parser/src/lib/form-data.ts#L143

My current alternative:
```typescript
try {
    formData = await parseFormData(
      request,
      { maxFileSize: 1024 * 1024 * 4, maxFiles: 1 },
      uploadHandler,
    );
  } catch (error) {
    if (
      error instanceof MaxFileSizeExceededError ||
      (error instanceof FormDataParseError &&
        "cause" in error &&
        error.cause instanceof MaxFileSizeExceededError)
    ) {
      return {
        uploadHandlerError: "File cannot be greater than 3MB",
      };
    } else if (
      error instanceof MaxFilesExceededError ||
      (error instanceof FormDataParseError &&
        "cause" in error &&
        error.cause instanceof MaxFilesExceededError)
    ) {
      return {
        uploadHandlerError: "You can only upload one file",
      };
    } else {
      throw error;
    }
  }
```

cyphercodes:
## Summary

Fixed the broken error handling example in the form-data-parser README.

## Issue

The example for handling file upload errors was not working correctly because it didn't account for the fact that errors are wrapped in FormDataParseError with the specific error as the cause property.

## Changes

Updated the error handling example to:
1. Import FormDataParseError
2. Show how to check error.cause for specific error types
3. Provide proper fallback handling for both direct errors and wrapped errors

## Before
The original example only checked for specific error types directly, which doesn't work because the errors are wrapped in FormDataParseError.

## After
The updated example shows the correct way to handle errors by checking both the direct error type and error.cause for wrapped errors.

Fixes #11180

### DIFF
--- a/packages/form-data-parser/README.md
+++ b/packages/form-data-parser/README.md
@@ -81,6 +81,7 @@ To limit the overall shape of multipart requests, use the `maxHeaderSize`, `maxF
 
 ```ts
 import {
+  FormDataParseError,
   MaxFilesExceededError,
   MaxFileSizeExceededError,
   MaxHeaderSizeExceededError,
@@ -99,6 +100,7 @@ try {
     maxTotalSize: 12 * oneMb,
   })
 } catch (error) {
+  // Errors are wrapped in FormDataParseError with the specific error as the cause
   if (error instanceof MaxFilesExceededError) {
     console.error(`Request may not contain more than 5 files`)
   } else if (error instanceof MaxHeaderSizeExceededError) {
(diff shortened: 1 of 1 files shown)


mjackson:
Superseded by #11260.

mjackson:
Thanks for bringing this up and for taking the time to improve the README.

I opened #11260 to address this more directly. Instead of documenting that callers need to check every limit error both directly and through `FormDataParseError.cause`, we are tightening the behavior and docs around the intended contract:

- known multipart limit errors are thrown directly, so `instanceof MaxFileSizeExceededError`, `MaxFilesExceededError`, etc. continue to work
- parser/body failures are wrapped in `FormDataParseError` with the original error available as `cause`
- errors thrown or rejected by a user-provided `uploadHandler` now propagate directly, since those come from app code rather than the parser

That should make the README example clearer and avoid the duplicated checks.

PR #11213, opened 2026-03-31
Add assert and test packages

brophdawg11:
Adds 2 new packages for unit testing Remix applications: `remix/assert` and `remix/test` and updates all of the `demos/` to use the new unit testing framework via the `remix-test` CLI.

Subsequent PRs to come for additional aspects:
 - Conversion of all packages/ unit tests to use `remix-test`
 - Adding in-browser and E2E testing support via playwright


packages/test/src/lib/config.ts
```diff
@@ -0,0 +1,176 @@
...
+  /**
+   * Path to a module that exports `setup` and/or `teardown` functions,
+   * called once before and after the test run respectively. (--setup)
+   */
+  setup?: string
+  /** Test reporter (--reporter) */
+  reporter?: string
+  /** Watch mode — re-run tests on file changes (--watch) */
+  watch?: boolean
+}
```
brophdawg11, line 38: Config values available to the user, also mirrored via CLI args

packages/test/src/lib/context.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    after(fn) {
+      cleanups.push(fn)
+    },
+    async cleanup() {
+      for (let fn of cleanups) await fn()
+      cleanups.length = 0
+    },
+  }
+}
```
brophdawg11, line 68: Context is passed to every test which helps with utils and automatic cleanup:

```tsx
it('tests something', (t) => {
 let fn = t.mock(() => 'mocked')
 let spy = t.spy(console, 'log') // Automatically restored after test
 t.after(() => { /* local test cleanup logic */ }); 
});
```

packages/test/src/lib/executor.ts
```diff
@@ -0,0 +1,158 @@
...
+  passed: number
+  failed: number
+  skipped: number
+  todo: number
+  tests: TestResult[]
+  e2eBrowserCoverageEntries?: Array<{ entries: V8CoverageEntry[]; baseUrl: string }>
+}
+
+export async function runTests(options?: { coverage?: boolean }): Promise<TestResults> {
+  let suites = (globalThis as any).__testSuites || []
```
brophdawg11, line 24: The executor is the shared enviroment-agnostic test runner.  It runs tests on `globalThis.__testSuites` which can be populated from node or in the browser

packages/test/src/lib/framework.ts
```diff
@@ -0,0 +1,239 @@
...
+function getImplicitRootSuite(): TestSuite {
+  if (!implicitRootSuite || !rootSuites.includes(implicitRootSuite)) {
+    implicitRootSuite = { name: '', tests: [], ...rootHooks }
+    rootSuites.push(implicitRootSuite)
+  }
+  return implicitRootSuite
+}
+
+// Expose for executor.ts which reads this global
+;(globalThis as any).__testSuites = rootSuites
```
brophdawg11, line 46: our describe/it block register test suites onto `globalThis.__testSuites` which is then picked up by the executor

packages/test/src/lib/reporter.ts
```diff
@@ -0,0 +1,307 @@
+import { colors, normalizeLine, type Counts } from './utils.ts'
+import type { TestResult, TestResults } from './executor.ts'
+
+export interface Reporter {
+  onResult(results: TestResults, env?: string): void
+  onSummary(counts: Counts, durationMs: number): void
+  onSectionStart(label: string): void
+}
+
+// ── Spec ─────────────────────────────────────────────────────────────────────
```
brophdawg11, line 10: We offer spec/dot/tap reporters out of the box

scripts/generate-remix.ts
```diff
@@ -24,7 +24,16 @@ const SOURCE_FOLDER = 'src'
 type RemixRunPackage = {
   name: string
   version: string
+  packageJsonPath: string
   exports: ExportEntry[]
+  bins: BinEntry[]
```
brophdawg11, line 29: generate-remix needed to be updated to automatically detect and make sub-package binaries available

demos/bookstore/remix-test.config.ts
```diff
@@ -0,0 +1,8 @@
+import type { RemixTestConfig } from 'remix/test'
+
+export default {
+  // The bookstore tests operate on the same DB so running multiple tests in
+  // concurrent workers can sometimes cause test failures
+  concurrency: 1,
```
brophdawg11: Refactor tests to use an in memory db and go back to max concurrency

PR #11221, opened 2026-03-31
test: Add E2E testing support

brophdawg11:
~~**Stacked on https://github.com/remix-run/remix/pull/11213**~~

Adds support for E2E testing to `remix/test`.  

`*.test.e2e.tsx?` files are identified as E2E test files in the test runner and receive a `t.serve()` method on the test context.  This accepts a fetch handler, creates a server, and returns a fresh playwright `page` context:

```ts
describe('e2e', () => {
  it('tests my app', async (t) => {
    let router = createAppRouter();
    let page = await t.serve(router.fetch)

    await page.goto('/', { waitUntil: 'networkidle' })

    // Add an item to cart
    await page.locator('[data-test="add-to-cart"]').click()
    await page.waitForSelector('[data-test="remove-from-cart"]')

    // Navigate to cart and validate
    await page.locator('a[href="/cart"]').click()
    await page.waitForSelector('[data-test="shopping-cart"]')
    assert.equal(await page.locator('[data-test="quantity"]').innerText(), '1')
  })
})
```

New config options available in `remix-test.config.ts`:
  - `browser.echo` - echo browser console logs back to the server
  - `browser.open` - open the browser during testing (`headless:false`) and keep it open until you exit
  - `glob.e2e` - set glob that identifies an e2e test file (defaults to `**/*.test.e2e.{ts,tsx}`)
  - `playwrightConfig` - a playwright config or a path to a one on disk
  - `project` - filter tests for a single playwright project


demos/bookstore/app/app.test.e2e.ts
```diff
@@ -0,0 +1,29 @@
...
+import { beforeAll, describe, it } from 'remix/test'
+import { createBookstoreRouter } from './router.ts'
+import type { Router } from 'remix/fetch-router'
+
+describe('bookstore e2e', () => {
+  let router: Router
+
+  beforeAll(() => {
+    router = createBookstoreRouter()
+  })
```
sergiodxa: I suppose this is a wrapper of `createRouter` from `remix/fetch-router`, it would be great if this supported `using` so the test could be written as:
```ts
it("adds to cart", async t => {
  using router = createBookstoreRouter()
  // more code
});
```
Without the need to use a variable in the `describe` scope and a beforeAll to create it

brophdawg11: Honestly the router could probably just live in the test there's not a lot of overhead to create it each time I don't think

demos/bookstore/remix-test.config.ts
```diff
@@ -5,4 +5,20 @@ export default {
...
+      {
+        name: 'firefox',
+        use: { browserName: 'firefox' },
+      },
+    ],
+    use: {
+      navigationTimeout: 5_000,
+      actionTimeout: 5_000,
+    },
+  },
```
brophdawg11, line 19: Playwright configs can be inlined in `remix-test.config.ts` or you can just pass a string location of an existing `playwright.config.ts` file

packages/test/src/lib/context.ts
```diff
@@ -61,6 +81,44 @@ export function createTestContext(options: { coverage?: boolean }): TestContext
...
+        baseURL: server.baseUrl,
+      })
+      if (options.playwrightPageOptions?.navigationTimeout != null) {
+        page.setDefaultNavigationTimeout(options.playwrightPageOptions.navigationTimeout)
+      }
+      if (options.playwrightPageOptions?.actionTimeout != null) {
+        page.setDefaultTimeout(options.playwrightPageOptions.actionTimeout)
+      }
+
+      let coverageEnabled = options.coverage && options.browser.browserType().name() === 'chromium'
```
brophdawg11: Only `chromium` supports coverage instrumentation

demos/social-auth/tsconfig.json
```diff
@@ -11,7 +11,6 @@
     "verbatimModuleSyntax": true,
     "skipLibCheck": true,
     "jsx": "react-jsx",
-    "jsxImportSource": "remix/component",
-    "preserveSymlinks": true
```
brophdawg11, line 15: @mjackson We had to unset this option to get the playwright types to resolve properly.  Do you recall why this was needed?

packages/test/src/lib/reporters/dot.ts
```diff
@@ -0,0 +1,57 @@
+import { colors, normalizeLine, type Counts } from '../utils.ts'
```
brophdawg11, line 1: These reporters are just split out into individual files in here, and a new `files` reporter that is a level of verbosity between `dot` and `spec`:

Image

packages/test/src/lib/context.ts
```diff
@@ -56,6 +73,32 @@ export function createTestContext(): { testContext: TestContext; cleanup(): Prom
...
+
+      cleanups.push(async () => {
+        if (!options.open) {
+          await page.close()
+        }
+        await server.close()
+      })
+
+      return page
+    },
```
brophdawg11, line 101: New `t.serve` function available to tests to run an E2E server and return a playwright Page instance

packages/test/src/lib/e2e-server.ts
```diff
@@ -0,0 +1,28 @@
...
+      let addr = server.address() as { port: number }
+      resolve({
+        baseUrl: `http://127.0.0.1:${addr.port}`,
+        close: () => new Promise((r, rj) => server.close((e) => (e ? rj(e) : r()))),
+      })
+    })
+
+    server.on('error', reject)
+  })
+}
```
brophdawg11, line 28: In-memory playwright server to handle e2e requests

packages/test/src/cli.ts
```diff
@@ -74,6 +80,47 @@ async function executeRun() {
...
+      let projects = resolveProjects(playwrightConfig)
+      if (config.project) {
+        let projectNames = config.project.split(',').map((p) => p.trim())
+        projects = projects.filter((p) => p.name && projectNames.includes(p.name))
+        if (projects.length === 0) {
+          throw new Error(`No playwright projects found with name(s) "${config.project}"`)
+        }
+      }
+
+      for (let project of projects) {
```
brophdawg11, line 94: Loop through and run e2e tests for each project (chromium, firefox, webkit, etc)

demos/social-auth/tsconfig.json
brophdawg11: We looked at this and everything looks ok without it - plus we found that pnpm specifically advises against this: https://pnpm.io/typescript#do-not-preserve-symlinks

PR #11234, opened 2026-04-04
feat(data-schema): add Schema.transform()

mcansh:
Add `Schema.transform()` to the `data-schema` chainable API so schemas can map a successfully validated output value into a new output value and type.

```ts
let Event = object({
  id: string(),
  createdAt: string()
    .refine((value) => !Number.isNaN(Date.parse(value)), 'Expected valid date')
    .transform((value) => new Date(value)),
})
```

- Runs transformers after the underlying schema succeeds and leaves validation failures untouched.
- Preserves the original input type while changing the schema output type.
- Documents usage and adds a release note for `@remix-run/data-schema`.

### DIFF
--- a/packages/data-schema/.changes/minor.schema-transform.md
+++ b/packages/data-schema/.changes/minor.schema-transform.md
@@ -1,0 +1,1 @@
+Add `Schema.transform()` for mapping validated schema outputs to new values and output types.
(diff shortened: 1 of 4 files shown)


mjackson:
thanks for the PR, @mcansh! 💪

PR #11243, opened 2026-04-09
Add style-server package

markdalgleish:
As a fast follow to https://github.com/remix-run/remix/pull/11149 since this follows the same architecture, this PR adds a new `style-server` package for compiling CSS files on the fly with LightningCSS, most notably supporting rewriting of `@import` statements, minification, and syntax lowering via a `browserslist` option.

With the `browserslist` option, the idea is to avoid coupling to the specifics of Lightning CSS' `targets` object format and use a more widely supported format. This makes it easier to move between tools in the future.

## Features

- **Route-Based Serving** - Map public URL patterns to files on disk
- **Caching** - Conservative caching by default with stable URLs, ETags, and revalidation
- **Watch Mode** - Watch for file changes and invalidate cached styles
- **Optional Fingerprinting** - Source-based fingerprinted URLs for long-lived browser caching
- **Source Maps** - Serve either inline or external sourcemaps
- **Browserslist Targets** - Compile for browserslist queries using Lightning CSS
- **Preloads** - Generate stylesheet preload URLs for one or more CSS files and their internal `@import` graphs

## Usage

```ts
import { createStyleServer } from 'remix/style-server'

let styleServer = createStyleServer({
  root: path.resolve(import.meta.dirname, '..'),
  routes: [{ urlPattern: '/styles/*path', filePattern: 'app/styles/*path' }],
})

let router = createRouter()

router.get('/styles/*', ({ request }) => {
  return styleServer.fetch(request)
})
```

This example gives you a `/styles/*` endpoint that serves compiled CSS from `app/styles`.


markdalgleish:
Closing in favour of https://github.com/remix-run/remix/pull/11269.

PR #11247, opened 2026-04-21
Update docs generation to produce an api lookup map

brophdawg11:
Update docs generation to produce a lookup map of API names to website markdown URLs that we could use in a `remix docs <api-name>` CLI:

```
{
  "Accept": "/api/remix/headers/class/Accept.md",
  "AcceptEncoding": "/api/remix/headers/class/AcceptEncoding.md",
  "AcceptEncodingInit": "/api/remix/headers/type/AcceptEncodingInit.md",
  "AcceptInit": "/api/remix/headers/type/AcceptInit.md",
  ...
}
```

This will be available from `https://api.remix.run/api.json` so the CLI can fetch that (and potentially cache) and then fetch the associated markdown content.  

It's versioned in the docs as well so `https://api.remix.run/v3.0.0/api.json` will only reflect APIs in v3.0.0 and will also update the markdown links to include the proper version prefix.

Open Questions
- If we want to skip the up front network request we can also generate and ship this in the remix cli so it doesn't need to query `/api.json` - but it feels like some form of caching would be sufficient too

### DIFF
--- a/docs/src/server/components.tsx
+++ b/docs/src/server/components.tsx
@@ -43,7 +43,8 @@ export function ServerPage(handle: Handle<ServerContext>, setup: ServerContext) 
 }
 
 function Document(handle: Handle) {
-  let { activeVersion } = handle.context.get(ServerPage)
+  let { activeVersion, slug } = handle.context.get(ServerPage)
+  let apiName = slug?.split('/').slice(-1)[0]
   return ({ children }: { children: RemixNode | RemixNode[] }) => (
     <html lang="en">
       <head>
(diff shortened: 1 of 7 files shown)


docs/src/server/components.tsx
```diff
@@ -57,6 +58,14 @@ function Document(handle: Handle) {
...
 
         <title>Remix API Documentation</title>
+        {slug ? (
+          <link
+            rel="alternate"
+            type="text/markdown"
+            href={routes.markdown.href({ version: activeVersion, slug })}
+            title={`Markdown docs for ${apiName}`}
+          />
+        ) : null}
```
brophdawg11, line 68: Add a `<link rel="alternate">` tag to HTML docs so agents can find the markldown version.

@mjackson do you think it would be worth adding an [llms.txt](https://llmstxt.org/) file as well?  We could auto-generate basically a bulleted list of the sidebar nav just pointing API names to markdown files.

PR #11248, opened 2026-04-21
Add content negotiation to docs site

brophdawg11:
Return markdown if a docs site request comes in with an `Accept` header:

```
curl 'http://localhost:3000/api/remix/auth/interface/Auth0AuthProfile/' -H 'Accept: text/markdown'
```

Driven by https://acceptmarkdown.com/

### DIFF
--- a/docs/src/server/router.tsx
+++ b/docs/src/server/router.tsx
@@ -3,7 +3,8 @@ import * as path from 'node:path'
 import * as semver from 'semver'
 import { type RemixNode } from 'remix/component'
 import { renderToStream } from 'remix/component/server'
-import { createRouter as _createRouter, type Router } from 'remix/fetch-router'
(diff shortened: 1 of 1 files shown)


docs/src/server/router.tsx
sergiodxa, line 1: The responses are missing the `Vary: Accept` header

brophdawg11: The `middleware: [responseHeader('Vary', 'Accept')],` should handle adding it to every response out of the docs handler

brophdawg11:
lol, downsides of being able to ask an agent to implement an idea so quickly is forgetting that we SSG our docs so this doesn't work 😂

brophdawg11:
Gonna add a `<link rel="alternate">` to https://github.com/remix-run/remix/pull/11247 pointing to the markdown doc for LLMs

PR #11228, opened 2026-04-02
test: Support in-browser component testing

brophdawg11:
**Stacked on https://github.com/remix-run/remix/pull/11221**

_Note: This branch temporarily copies `packages/script-server` from https://github.com/remix-run/remix/pull/11149 - that will be removed once that lands in main and this can be rebased_

Adds support for in-browser component testing to `remix/test`.  

`*.test.browser.tsx?` files are identified as browser test files in the test runner and run in a live browser via playwright in an isolated iframe.  The new `t.render()` method on the test context accepts a `RemixNode` and renders it into a `container` div and returens a set of utilities for interacting with the rendered component.  The container and component are unmolunted and cleaned up automatically after the test runs.

```ts
describe('Counter', () => {
  it('increments on click', async (t) => {
    let { $, act } = t.render(<Counter />)

    assert.equal($('[data-count]')?.textContent, '0')
    await act(() => $('[data-action="increment"]')?.click())
    assert.equal($('[data-count]')?.textContent, '1')
  })
})
```

New config options available in `remix-test.config.ts`:
  - `glob.browser` - glob that identifies a browser test file (defaults to `**/*.test.browser.{ts,tsx}`)

PR #11249, opened 2026-04-21
Convert most packages to use remix-test

brophdawg11:
Convert most packages to use `remix-test` instead of `node --test`

* `data-table` packages are excluded because they do code coverage checks and re removed that from the first implementation of `remix/test` - once that's added back we'll convert those
* `component` tests excluded because they use vitest in browser testing - those will be converted once the assets package (https://github.com/remix-run/remix/pull/11149) and our own browser testing (https://github.com/remix-run/remix/pull/11228) land


packages/async-context-middleware/src/lib/async-context.test.ts
```diff
@@ -1,5 +1,5 @@
-import * as assert from 'node:assert/strict'
-import { describe, it } from 'node:test'
+import * as assert from '@remix-run/assert'
+import { describe, it } from '@remix-run/test'
```
brophdawg11, line 2: For the most part - these import changes and the package.json change from `node --test` to `remix-test` are the bulk of this PR - very little test source code was changed with the exception of minor mocking callsites.  This is by design since we aim to be mostly a drop in replacement for `node --test` and `node:assert/strict`.

packages/form-data-middleware/src/lib/form-data.test.ts
```diff
@@ -377,8 +377,8 @@ describe('formData middleware', () => {
     assert.equal(actionCalled, false)
   })
 
-  it('invokes a custom `uploadHandler` for file uploads', async () => {
-    let uploadHandler = mock.fn<FileUploadHandler>()
+  it('invokes a custom `uploadHandler` for file uploads', async (t) => {
+    let uploadHandler = t.mock<FileUploadHandler>()
```
brophdawg11: Mocks now use our own mocking from the test context.

@mjackson I'd love to bikeshed with you a bit on the names here, we have 2 ways to use mocking - manually or via the test context. 

```ts
import { mock } from 'remix/test';

it('manual mocking + manual cleanup', async () => {
    let mocked = mock.fn()
    let spied = mock.spyOn(console, 'log')
    ...
    // manual cleanup required
    spied.mock.restore();
    // no cleanup needed for mocked
})

it('test context mocking + automatic cleanup', async (t) => {
    let mocked = t.mock()
    let spied = t.spyOn(console, 'log');
    ...
    // automatically cleaned up via the test context
});
```

I don't love the naming difference between `mock.fn` and `t.mock`, but neither `mock.mock()` nor `t.fn()` felt quite right.  `t.fn()` felt like it was missing key semantics to communicate that it was mocking related (since `t` has other methods on it unrelated to mocking).  It's worth noting the familiarity with `jest.fn()` might make folks a bit more aware of the purpose.

brophdawg11: Actually now that I'm looking closer node just [exports mock directly](https://nodejs.org/api/test.html#class-mocktracker) on it's own test context so maybe we just get rid of the flattening on ours and use `t.mock.fn` and `t.mock.method` to align with node?

scripts/utils/git.test.ts
```diff
@@ -12,38 +12,37 @@ function execGit(args: string[], cwd: string): string {
 
 test('findVersionIntroductionCommit returns the commit where the target version was introduced', () => {
   let tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'remix-git-test-'))
-  let originalCwd = process.cwd()
   let packageJsonPath = 'packages/example/package.json'
+  let absolutePackageJsonPath = path.join(tempDir, packageJsonPath)
 
   try {
-    process.chdir(tempDir)
```
brophdawg11, line 19: We can't call `process.chdir` inside a worker so I adjusted these tests to pass a `cwd` down to the underlying `execFileSync` call

packages/form-data-middleware/src/lib/form-data.test.ts
mjackson: Yeah I think we just do what node does when it comes to mocking. I like their semantics. I almost never use the `t` context and just `import { mock }` and use that.

brophdawg11: ok cool.  I aligned the APIs so it now available as `import { mock }` or as `t.mock` - same as it is in node.

PR #11250, opened 2026-04-21
Add code coverage support and convert data-table tests

brophdawg11:
~~**Stacked on https://github.com/remix-run/remix/pull/11221**~~

Adds back code coverage support to remix-test and converts data-table tests which currently use node's experimental coverage


packages/test/src/lib/context.ts
```diff
@@ -90,6 +93,18 @@ export function createTestContext(options: {
...
+      if (coverageEnabled) {
+        await page.coverage.startJSCoverage({ resetOnNavigation: false })
+        cleanups.push(async () => {
+          let entries = await page.coverage.stopJSCoverage()
+          options.addE2ECoverageEntries?.({
+            entries: entries as unknown as V8CoverageEntry[],
+            baseUrl: server.baseUrl,
+          })
+        })
+      }
```
brophdawg11, line 119: Playwright has built-in coverage support so we can leverage that for E2E tests

packages/test/src/lib/coverage-loader.ts
```diff
@@ -0,0 +1,31 @@
...
+import { fileURLToPath } from 'node:url'
+import { transformTypeScript } from './ts-transform.ts'
+
+// Custom ESM loader hook for TypeScript files.
+//
+// Replaces tsx's minified transformation with an un-minified esbuild transform
+// that preserves line structure. This ensures V8 coverage byte offsets map
+// cleanly to TypeScript source lines via the inline source map, giving
+// accurate per-line coverage rather than collapsing multiple statements onto
+// a single minified line.
```
brophdawg11, line 11: If we just use `tsImport` for our files, we end up with mismatched line numbers/byte offsets in the coverage files and the source files because the coverage files report against the transformed JS so when reporting at the end we need to re-transform _the same way tsx does internally_ so get them to re-align.  Instead, we skip `tsImport` for coverage runs and just do our own `esbuild` transform so we can ensure it's done exactly the same and that all lines match up.

packages/test/src/lib/runner.ts
```diff
@@ -36,26 +48,59 @@ export async function runServerTests(
...
+        process.cwd(),
+        new Set(files),
+      )
+    }
   } else {
+    let coverageDataDir: string | undefined
+    if (options.coverage) {
+      coverageDataDir = path.resolve(options.coverage.dir)
+      await fsp.mkdir(coverageDataDir, { recursive: true })
+      process.env.NODE_V8_COVERAGE = coverageDataDir
```
brophdawg11, line 86: Enable node V8 coverage for server unit test runs

packages/test/src/lib/coverage-loader.ts
brophdawg11: See the comment in [packages/test/src/lib/worker.ts‎](https://github.com/remix-run/remix/pull/11250/changes#diff-1f6996d3d81495f8e72d2b0c9959ee73d9ad684457e0db88f0f0a35de809f986) for more info

packages/test/src/test/coverage-fixture.test.e2e.ts
```diff
@@ -0,0 +1,67 @@
...
+    // - classify: only positive branch
+    // - greet: only with a name
+    // - uncalledFunction: never imported
+    let results = [
+      add(2, 3),
+      add(-1, 1),
+      classify(42),
+      classify(1),
+      greet('World'),
+    ]
```
brophdawg11, line 44: This allows us to confirm coverage values for a fixture file.  We have a unit and E2E version of this against the same fixture

packages/test/src/lib/ts-transform.ts
```diff
@@ -0,0 +1,21 @@
...
+ * source map and no minification. Used both by the coverage ESM loader hook
+ * (so V8 instruments readable JS) and by the coverage collector (so byte
+ * offsets can be re-derived and mapped back to TypeScript lines).
+ */
+export async function transformTypeScript(
+  source: string,
+  filePath: string,
+): Promise<{ code: string }> {
+  let loader: 'ts' | 'tsx' = filePath.endsWith('.tsx') ? 'tsx' : ('ts' as const)
+  let result = await transform(source, {
```
brophdawg11, line 14: Convert to use `oxc` - should be faster than esbuild

brophdawg11: sticking with esbuild for now - had the agent implement but it jacked up the coverage reporting.  Can always swap it under the hood later since it's an implementation detail

Screenshot 2026-04-22 at 5 35 00 PM

mjackson (approved):
This looks great. Thanks! 🙏

PR #11251, opened 2026-04-22
data-table: remove database adapter options

pcattori:
BREAKING CHANGE: Removed adapter options for Mysql, Postgres, Sqlite

**Affected APIs**

- `*DatabaseAdapterOptions` types: removed
- `create*DatabaseAdapter` functions: `options` arg removed
- `*DatabaseAdapter` constructors: `options` arg removed

**Why**

Adapter options existed solely for tests to override adapter capabilities.
If you must override capabilities, you can do so directly via mutation:

```ts
let adapter = createSqliteDatabaseAdapter(sqlite)
adapter.capabilities = {
  ...adapter.capabilities,
  returning: false,
}
```

### DIFF
--- a/packages/data-table-mysql/src/lib/adapter.test.ts
+++ b/packages/data-table-mysql/src/lib/adapter.test.ts
@@ -41,26 +41,23 @@ const accountProjects = table({
 })
 
 describe('mysql adapter', () => {
-  it('applies explicit capability overrides', () => {
-    let adapter = createMysqlDatabaseAdapter(
-      {
-        async query() {
-          return [[], []]
-        },
-        async beginTransaction() {},
-        async commit() {},
-        async rollback() {},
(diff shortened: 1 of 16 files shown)


packages/data-table-mysql/src/lib/adapter.test.ts
```diff
@@ -41,26 +41,23 @@ const accountProjects = table({
...
-      } as never,
-      {
-        capabilities: {
-          returning: true,
-          savepoints: false,
-          upsert: false,
-          transactionalDdl: true,
-          migrationLock: false,
-        },
+  it('supports capability mutation for fallback tests', () => {
```
pcattori: @mjackson I think we should probably just remove these tests altogether since all they do now is assert that JS assignment works

PR #11253, opened 2026-04-24
test: add `test:bun` script, run in CI

markdalgleish:
**🚨 NOTE: Before merging, update `.github/workflows/test.yaml` to uncomment `if: github.event_name == 'push'` so this only runs against `main`.**

This adds `test:bun` scripts across the supported packages, wires them up at the workspace root, and adds a Bun job to the test workflow so we can start exercising the suite in CI.

~~I had to work around an issue with Bun's handling of the `remix/test` default globs in this workspace. Using explicit `--glob.test 'src/**/*.test.{ts,tsx}'` and `--glob.e2e 'src/**/*.test.e2e.{ts,tsx}'` patterns in the `test:bun` scripts avoids Bun walking into recursively nested `node_modules` directories, which appears to be triggered by a dependency cycle in the workspace. We should investigate this separately, but I didn't want it to block this work.~~ (fixed in [8b42518](https://github.com/remix-run/remix/pull/11253/commits/8b4251850b56b315c3d62eb0519c1db2e4031ef5))

To get the suite passing under Bun, this also makes a few targeted compatibility changes:

- Test runner and CI support:

  - add per-package `test:bun` scripts plus a root `pnpm test:bun`
  - add a Bun job to `.github/workflows/test.yaml`
  - update `@remix-run/test` to load setup files, config files, Playwright config, and test modules through a shared `importModule()` helper because the existing `tsx` loader path does not work in Bun, while Bun can natively import the same modules itself

- Bun-specific assertion updates:

  - relax `fetch-proxy` assertions for `Request` properties like `cache`, `credentials`, `integrity`, `keepalive`, and `mode` because Bun applies its own default request values instead of preserving all inputs
  - update `fetch-router` `FormData` assertions to compare `File` contents and metadata instead of object identity since Bun returns new `File` instances
  - skip the ISO-8859-15 `Content-Disposition` decoding case (if not supported by the runtime) in `headers` because Bun's `TextDecoder` does not support that encoding
  - skip the ranged body assertions in `response` because Bun's `Blob`/`File` `slice().stream()` currently has an issue where it returns the full body instead of the sliced body

- Runtime-dependent input normalization:
  - derive MIME types from the native `File`/`Blob` implementation in `file-storage`, `file-storage-s3`, `form-data-middleware`, `form-data-parser`, `lazy-file`, and `response` since Bun normalizes them differently
  - account for Bun adding `charset=utf-8` to some text response content types and rewriting values like `application/javascript`
  - compare against runtime-normalized upload and file types instead of hard-coded MIME strings where Bun normalizes them differently

The `data-table` packages are intentionally left out for now because Bun does not yet support `better-sqlite3`, and selectively skipping the resulting failures in `bun test` would add a lot of awkward test-only branching. We can revisit this once `remix/test` supports code coverage so we can migrate that suite to use conditional skip logic.


mjackson (approved):
I'm glad you're clearly calling out `IS_BUN` everywhere Bun diverges from web standards and/or node. I think that's the right approach.

Almost makes me not want to merge this, but I do want 1st class Bun support in Remix, so feel free to merge away.

mjackson:
Follow-up commit pushed to main: cb06c44e9 (`ci: run Bun tests only on main`). This uncomments the `if: github.event_name == 'push'` guard for the Bun test workflow job.

PR #11258, opened 2026-04-24
feat(cli): default start screen for bootstrapped apps

tquirino:
## Summary

Replaces the placeholder home page that `remix new` scaffolds today with a real "Welcome to Remix" start screen — the first thing a developer sees in a freshly bootstrapped app.

The page is implemented as an idiomatic Remix route — a single self-contained `app/controllers/home.tsx` using Remix components and the `css` mixin from `remix/ui` — so it follows the framework's own patterns rather than shipping a static `public/*.html` asset.

## What changed

Single file: `packages/cli/bootstrap/app/controllers/home.tsx`.

- Renders its own `<html>` document so the start screen owns its `<head>` (font preconnects, color-scheme meta, etc.) without forcing changes to the shared `Document` / `Layout` components, which stay minimal.
- All styling is authored with `mix={[css({ ... })]}`. Design tokens live on `<body>` as CSS custom properties, with `@media (prefers-color-scheme: dark)` overrides nested in the same call, so the page responds to the user's system theme.
- A single SVG wordmark covers both themes: the bracket characters keep their decorative brand colors; the "REMIX" letterforms use `currentColor` and pick up `--text-primary` from `<body>`.
- All icons are inlined — no `<defs>` / `<use>` symbol sheet, no external SVG assets.
- Click-to-copy on the AI-prompt buttons is preserved via a single inline `<script innerHTML={...}>`. This keeps the click affordance without introducing a client bundle, matching the bootstrap's "minimal scaffold" philosophy.
- `APP_DISPLAY_NAME` is still declared from the `%%RMX_APP_DISPLAY_NAME_URI_COMPONENT%%` token and used in the document `<title>`, so the existing CLI test that asserts the interpolation pattern (`packages/cli/src/lib/cli.test.ts`) continues to pass.
- The "Join Discord" card link and footer social icon point to https://discord.gg/xwx7mMzVkA.

## Design

Light mode and dark mode are driven entirely by `prefers-color-scheme`. The masthead, two-column "Get started" / "Coding with AI?" cards, and footer are all built from local helper components in the same file — keeping `home.tsx` a flat leaf controller per `skills/remix-project-layout`.

Helper components (`CardLink`, `PromptButton`, `IconSlot`, etc.) follow the Remix component model: setup runs once, props are received by the returned render function so they update on every render.

## Test plan

- [x] `pnpm prettier --check` passes.
- [x] Bootstrap copy + `%%RMX_APP_DISPLAY_NAME_URI_COMPONENT%%` token replacement produces a valid `home.tsx` whose generated line matches `const APP_DISPLAY_NAME = decodeURIComponent('<encoded-app-name>')` (per `packages/cli/src/lib/cli.test.ts`).
- [x] Manual preview rendered against a local Remix app: full layout (masthead, two cards, footer) shows correctly, prompt buttons render their text, light/dark theming works via `prefers-color-scheme`, click-to-copy on prompt buttons swaps the label to "Copied to clipboard" and back.
- [ ] CI: `pnpm --filter @remix-run/cli test` (couldn't run locally on Node 22; CI runs on Node 24).
- [ ] Reviewer: `remix new ./tmp-app && cd tmp-app && pnpm install && pnpm dev` and visit `/` — confirm the start screen renders correctly in both light and dark mode and the prompt buttons copy to clipboard on click.

### DIFF
--- a/packages/cli/bootstrap/app/controllers/home.tsx
+++ b/packages/cli/bootstrap/app/controllers/home.tsx
@@ -1,5 +1,6 @@
 import type { BuildAction } from 'remix/fetch-router'
 import { css } from 'remix/ui'
+import type { RemixNode } from 'remix/ui'
 
 import type { routes } from '../routes.ts'
 import { render } from '../utils/render.tsx'
@@ -24,7 +25,7 @@ function HomePage() {
         <meta name="color-scheme" content="light dark" />
         <title>Welcome to {APP_DISPLAY_NAME}</title>
         <link rel="preconnect" href="https://fonts.googleapis.com" />
-        <link rel="preconnect" href="https://fonts.gstatic.com" crossOrigin="" />
+        <link rel="preconnect" href="https://fonts.gstatic.com" crossOrigin="anonymous" />
         <link
           rel="stylesheet"
           href="https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;700&display=swap"
@@ -222,7 +223,7 @@ function CodingWithAiCard() {
 }
 
 function CardLink() {
-  return ({ href, icon, label }: { href: string; icon: unknown; label: string }) => (
+  return ({ href, icon, label }: { href: string; icon: RemixNode; label: string }) => (
     <a
       href={href}
       mix={[
@@ -302,7 +303,7 @@ function PromptButton() {
 }
 
 function IconSlot() {
-  return ({ children, rotated = false }: { children: unknown; rotated?: boolean }) => (
+  return ({ children, rotated = false }: { children: RemixNode; rotated?: boolean }) => (
     <span
       aria-hidden="true"
       mix={[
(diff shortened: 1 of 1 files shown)


skills/remix-project-layout/scripts/assets/start-screen.html
```diff
@@ -0,0 +1,716 @@
+<!doctype html>
```
brookslybrand: Just noting: we of course need to change this to remix code. I can probably help with that if MJ doesn't pick it up

skills/remix-project-layout/scripts/assets/start-screen.html
```diff
@@ -0,0 +1,716 @@
...
+              </a>
+            </li>
+            <li>
+              <a class="link" href="https://remix.run/docs/start/tutorial">
+                <span class="icon-slot" aria-hidden="true">
+                  <svg><use href="#i-cursor" /></svg>
+                </span>
+                <span class="link-label">Tutorial (30 min)</span>
+              </a>
+            </li>
```
brookslybrand: We won't actually have tutorials, definitely not for the beta, maybe never! I think this could just be "get started" or something like that, with the list of prompts on the right

tquirino: Yea that's a good call, removing the links since they don't apply to Remix 3 beta.

brookslybrand:
Implemented in #11308

PR #11261, opened 2026-04-24
Add ASCII art to README

tquirino:
Added an ASCII art version of the Remix logo to the README.


mjackson:
The text causes horizontal scroll bars on narrow screens. Maybe we could use a simple image instead? Or a smaller version of the ascii art?

PR #11262, opened 2026-04-24
Add browser component testing to remix test

brophdawg11:
Adds browser-based component tests to `remix test`. Files matching `**/*.test.browser.{ts,tsx}` run in a live Playwright browser inside isolated iframes, alongside the existing server and E2E test runners.

- Adds `glob.browser`, `browser.echo`, and `browser.open` test runner options.
- Exposes `render` from `remix/ui/test` for mounting Remix UI nodes into a DOM container with query helpers, `act()`, and `cleanup()`.
- Updates E2E tests so `t.serve()` accepts a running `{ baseUrl, close }` test server, with `createTestServer` exported from `remix/node-fetch-server/test`.
- Keeps `@remix-run/test` runtime dependencies narrow by moving UI rendering and Node fetch server helpers behind their owning package test exports.
- Extends coverage collection so server, browser, and E2E test runs can contribute to one combined report.

```tsx
import * as assert from 'remix/assert'
import { describe, it } from 'remix/test'
import { on, type Handle } from 'remix/ui'
import { render } from 'remix/ui/test'

function Counter(handle: Handle<{ count?: number }>) {
  let count = handle.props.count ?? 0
  return () => (
    <button
      data-count={count}
      mix={[
        on('click', () => {
          count++
          handle.update()
        }),
      ]}
    >
      {count}
    </button>
  )
}

describe('Counter', () => {
  it('increments on click', async (t) => {
    let { $, act, cleanup } = render(<Counter />)
    t.after(cleanup)

    assert.equal($('[data-count]')?.textContent, '0')
    await act(() => $('[data-count]')?.click())
    assert.equal($('[data-count]')?.textContent, '1')
  })
})
```


packages/component/package.json
```diff
@@ -57,14 +57,14 @@
     "prepublishOnly": "pnpm build"
   },
   "dependencies": {
-    "@types/dom-navigation": "^1.0.7"
+    "@types/dom-navigation": "1.0.6"
```
brophdawg11: Need to figure out the best way to handle this prior to the official TS 7 launch

decisions/003-coverage-transform-determinism.md
```diff
@@ -0,0 +1,122 @@
+# Why `remix-test` owns its TypeScript transform for coverage
```
brophdawg11, line 1: @mjackson @markdalgleish tagging this as a discussion point

packages/ui/src/runtime/mixins/attrs-mixin.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { describe, expect, it } from 'vitest'
 import { createRoot } from '../vdom.ts'
 import { invariant } from '../invariant.ts'
-import { attrs } from './attrs-mixin.tsx'
+import { attrs } from './attrs-mixin.ts'
```
brophdawg11, line 4: We can rename this now that we removed the internal JSX usage

demos/bookstore/package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "dev": "NODE_ENV=development tsx watch server.ts",
     "start": "tsx server.ts",
-    "test": "NODE_ENV=test remix-test app/server.test.ts && NODE_ENV=test remix-test --type e2e",
+    "test": "NODE_ENV=test remix-test",
```
brophdawg11: @mjackson I think this is no longer needed since you added `concurrency: 1` to the `remix-test.config.ts`?  I got rid of the `server.test.ts` file and this is now using default globs again to pick up all test files.

packages/assert/src/lib/assert.test.ts
```diff
@@ -1,5 +1,5 @@
 import * as nodeAssert from 'node:assert/strict'
-import { describe, it } from '@remix-run/test'
+import { describe, it } from 'node:test'
```
brophdawg11, line 2: Flipped back to node tests to avoid circular pnpm deps

packages/assert/package.json
```diff
@@ -35,12 +35,11 @@
     "build": "tsgo -p tsconfig.build.json",
     "clean": "git clean -fdX",
     "prepublishOnly": "pnpm run build",
-    "test": "remix-test",
-    "test:bun": "bun x --bun remix-test",
+    "test": "node --test",
+    "test:bun": "bun --test",
```
brophdawg11:
```suggestion
    "test:bun": "bun test",
```

packages/node-fetch-server/src/test.ts
```diff
@@ -0,0 +1,2 @@
+export { createTestServer } from './lib/test-server.ts'
+export type { TestServer } from './lib/test-server.ts'
```
brophdawg11, line 2: This export now comes from `remix/node-fetch-server/test`

demos/bookstore/package.json
mjackson: Yep, good call

PR #11269, opened 2026-04-27
Add styles support to `assets`

markdalgleish:
This adds CSS support to `@remix-run/assets` so `createAssetServer()` can compile and serve stylesheet assets alongside browser scripts from the same URL namespace.

As part of that, the compiler internals have been refactored around a shared generic module store that handles record lifecycle, invalidation, and watch tracking, with separate script and style compilers built on top.

Shared compiler options also move to the top level of `createAssetServer()` so the same `target`, `minify`, `sourceMaps`, and `sourceMapSourcePaths` settings can apply to both scripts and styles. `target` also expands from the old `scripts.target` string form to an object that supports browser version targets plus `es`.

- Adds CSS compilation, relative `@import` rewriting, fingerprinting, caching, and preload graph support to `createAssetServer()`
- Refactors the compiler internals around a shared generic module store, with separate script and style stores/compiler layers on top
- Hoists shared compiler options from `scripts.*` to the top level so they work for both scripts and styles
- Replaces the old script-only target string with a shared `target` object format and updates the `assets` README plus bookstore demo accordingly

Basic usage:

```ts

  fileMap: {
    '/assets/app/*path': 'app/*path',
    '/assets/npm/*path': 'node_modules/*path',
  },
  allow: ['app/assets/**', 'node_modules/**'],
  target: {
    chrome: '109',
    safari: '16.4',
  },
  minify: true,
  sourceMaps: 'external',
})
```

Migration:

```ts
// Before
  fileMap: { '/assets/app/*path': 'app/*path' },
  allow: ['app/assets/**'],
  scripts: {
    target: 'es2020',
    minify: true,
    sourceMaps: 'external',
  },
})
```

```ts
// After
  fileMap: { '/assets/app/*path': 'app/*path' },
  allow: ['app/assets/**'],
  target: { es: '2020' },
  minify: true,
  sourceMaps: 'external',
})
```


mjackson (approved):
LGTM!

PR #11287, opened 2026-04-29
Add prerender script for ui demo

brophdawg11:
Add prerendering for the ui demo site - to be published via https://github.com/remix-run/remix-ui-docs

This doesn't support versioning the same way the API docs do - but we can add that as a follow up


.github/workflows/publish.yaml
```diff
@@ -106,3 +106,15 @@ jobs:
               event_type: 'update-docs',
               client_payload: { tag: '${{ needs.publish.outputs.tag }}' }
             });
+
+      - name: Trigger remix-ui-docs
+        uses: actions/github-script@v8
+        with:
+          github-token: ${{ secrets.REMIX_API_DOCS_PAT }}
```
brophdawg11, line 113: @mjackson I don't remember exactly how this is token was configured, but if it only has access to trigger repository dispatch events on the `remix-api-docs` repo then we'll need to setup a new token with access to the `remix-ui-docs` repo

PR #11292, opened 2026-04-29
Convert UI tests to use remix-test

brophdawg11:
Convert the `packages/ui` tests from `vitest` to `remix-test`, adding new APIs to `remix/assert` and `remix/test` as needed

* Added an `expect` API to `remix/assert`
* Added `FakeTimers#advanceAsync(ms)` to `t.useFakeTimers()` for more accurate async timer advancement in tests.
* Align config/CLI fields to accept one or multiple values (`glob.*`, `project`, `type`, `coverage.include`, `coverage.exclude`)


packages/test/src/lib/user-event.ts
```diff
@@ -0,0 +1,231 @@
...
+  if (field.selectionStart === 0 && field.selectionEnd === 0 && field.value.length > 0) {
+    try {
+      field.setSelectionRange(field.value.length, field.value.length)
+    } catch {
+      // Some <input> types reject setSelectionRange.
+    }
+  }
+}
+
+async function type(target: Element, text: string): Promise<void> {
```
brophdawg11: @ryanflorence This stuff is your wheelhouse - does this logic look about right to you?

packages/ui/src/runtime/navigation.ts
```diff
@@ -42,8 +42,20 @@ export async function navigate(href: string, options?: NavigationOptions) {
...
+}
+
+// Internal version used by unit tests so we can inject stub frames
+export function startNavigationListenerImpl(
+  signal: AbortSignal,
+  options: {
+    getTopFrame: typeof getTopFrame
+    getNamedFrame: typeof getNamedFrame
+  },
+) {
```
brophdawg11, line 58: Since we don't support module mocking, we unit test this behavior via an internal function allow us to inject the mocked implementations.  Public API remains unchanged.

packages/ui/src/test/setup.ts
```diff
@@ -1,7 +0,0 @@
-if (typeof document !== 'undefined' && !document.adoptedStyleSheets) {
```
brophdawg11, line 1: @ryanflorence Do you recall why this was needed?  Tests are passing without it so I removed it

packages/ui/src/test/stream.test.tsx
```diff
@@ -465,43 +466,35 @@ describe('stream', () => {
...
-      try {
-        let html = await drain(renderToStream(<div mix={[withChildren()]}>safe</div>))
-
-        expect(html).toBe('<div data-mode="children">safe</div>')
-        expect(errorSpy).toHaveBeenCalledTimes(1)
-        let error = errorSpy.mock.calls[0]?.[0]
-        invariant(error instanceof Error)
-        expect(error.message).toBe('mixin elements must not receive children')
-      } finally {
-        errorSpy.mockRestore()
```
brophdawg11, line 484: `t.mock` automatically handles cleanup so no need to manually cleanup anymore

packages/test/src/lib/user-event.ts
ryanflorence: I'm not sure I'd care to include these abstractions and then need to support them, just let app tests dispatch events

packages/ui/src/test/setup.ts
ryanflorence: don't think we need it, looks like overly defensive agent junk

packages/test/src/lib/user-event.ts
brophdawg11: yeah I'm fine with that - it was only the complexity of getting `type` right that made me think it might be worth including since I worry userland `type` implementations won't be accurate.  I also tried just using `@testing-library/user-events` directly in the test file but it's not fully ESM so that caused issues (may need to look into this separately).  I'll take this out for now and we can always re-add if we want to take on the surface area

brophdawg11: removed in [386e871](https://github.com/remix-run/remix/pull/11292/commits/386e871fe71ff3d278f22ca083ed96577acbf579)

packages/test/src/lib/runner-browser.ts
```diff
@@ -61,6 +61,12 @@ export async function runBrowserTests(options: TestRunOptions): Promise<{
       getPlaywrightLaunchOptions(options.playwrightUseOpts),
     )
     page = await browser.newPage(getPlaywrightPageOptions(options.playwrightUseOpts))
+    // Cap how long we'll wait for a browser-test file to signal completion.
+    // Playwright's default is 30s; bumping to 60s buys headroom for slower
+    // suites without letting a hung test hide forever. Plumb this through
+    // config later if anyone needs to tune it.
+    page.setDefaultTimeout(60_000)
+    page.setDefaultNavigationTimeout(60_000)
```
brophdawg11:
```suggestion
    page.setDefaultTimeout(90_000)
    page.setDefaultNavigationTimeout(90_000)
```

PR #11298, opened 2026-04-29
Add pool option (threads | forks) to remix-test

brophdawg11:
`pool: 'threads'` (default) keeps the existing `worker_threads.Worker` per test file. `pool: 'forks'` runs each file in a `child_process.fork` for full process-level isolation. Configurable via `pool` in the config file or `--pool` CLI flag.


brophdawg11:
See #11303

PR #11305, opened 2026-04-30
route-pattern: replace `TrieMatcher` and `ArrayMatcher` with `createMatcher` fn

pcattori:
TrieMatcher was always faster than ArrayMatcher in our benchmarks, except for a few lambda-style benchmarks.
But for those, the absolute time was tiny and difference was negligible.

In fact, some lambda providers start booting up your request handler while they are still pre-processing the request, which often takes on the order of a few milliseconds. That already covers for the total time taken by TrieMatcher to setup on the `shopify.com` benchmark w/ ~2500 patterns.

Therefore, `createMatcher` uses `TrieMatcher` internally. Moved `ArrayMatcher` into the `bench` subpackage so we can still benchmark against that implementation.

Perf looks the same accounting for noise (reminder, lambda benchmarks are much noisier since their absolute runtimes are much faster):

Screenshot 2026-04-30 at 11 39 18 AM

### DIFF
--- a/.agents/skills/make-decision-doc/SKILL.md
+++ b/.agents/skills/make-decision-doc/SKILL.md
@@ -1,0 +1,54 @@
+---
+name: make-decision-doc
+description: Add a numbered decision document under `decisions/` to capture a non-obvious architectural choice. Use when the user asks for a decision doc, ADR, design rationale, or wants to record why we picked one approach over alternatives.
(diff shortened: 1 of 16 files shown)


mjackson:
@pcattori Please include a decision doc that explains our rationale for going with a single matcher.

PR #11318, opened 2026-05-01
Fix bootstrap client entry resolution

brookslybrand:
- Resolve scaffolded app client entries through the bootstrap asset server during server rendering.
- Switch the starter `PromptButton` client entry to use its source module URL instead of hard-coding the public asset path.
- Add a CLI patch change note for the scaffold template fix.


brookslybrand:
Closed in favor of #11319

PR #11319, opened 2026-05-01
Bootstrap updates

ryanflorence:
- bootstrap script to easily manually run the bootstrap template
- conventional controllers
- import.meta.url client entry convention boilerplate


mjackson (approved):
These are great updates. I would probably call the script `start-bootstrap.ts` instead of just `bootstrap.ts` but that's it.

brookslybrand:
> These are great updates. I would probably call the script `start-bootstrap.ts` instead of just `bootstrap.ts` but that's it.

I made that change and tweaked some outdated skills

PR #11345, opened 2026-05-04
Add favicon and clean up app template shell

brookslybrand:
Updates the `remix new` starter so app-root static assets are served by default and the starter home page shares the same document shell as future routes.

- Adds the Remix favicon to `public/favicon.svg`, serves `./public` with static middleware without treating `public/index.html` as the app root, and links `/favicon.svg` from the generated document
- Moves shared document shell concerns like charset, viewport, title, body reset, favicon, and the entry script into `app/ui/document.tsx` while keeping starter-home-page font metadata with the starter page
- Simplifies the home page scaffold by reusing `Document`, removing the unused layout wrapper, and documenting `public/` as part of the starter shape


template/app/ui/document.tsx
```diff
@@ -7,17 +7,25 @@ export interface DocumentProps {
   title?: string
 }
 
-const DEFAULT_TITLE = readAppDisplayName('%%RMX_APP_DISPLAY_NAME_URI_COMPONENT%%')
+const DEFAULT_TITLE = `Welcome to ${readAppDisplayName('%%RMX_APP_DISPLAY_NAME_URI_COMPONENT%%')}`
 
 export function Document() {
```
brookslybrand, line 13: Previously Document was only being used in Layout, and Layout was never being used

I moved all the document related stuff into Document and use it instead of the homepage controlling all of that

template/app/ui/scaffold-home-page.tsx
```diff
@@ -332,21 +319,17 @@ function Footer() {
...
+const cardStyle = css({
   background: 'var(--surface-3)',
   borderRadius: '20px',
   padding: '32px 16px 16px',
   display: 'flex',
   flexDirection: 'column',
   alignItems: 'center',
   gap: '32px',
-} as const
+})
```
brookslybrand, line 350: In generally we need to make sure our templates are following good Remix patterns. This stuff gets picked up and propagated by agents, whether or not we have good skills

template/app/router.ts
```diff
@@ -12,8 +13,17 @@ declare module 'remix/fetch-router' {
...
 }
 
-export const router = createRouter<AppContext>({
-  middleware: [render()],
-})
+export function createAppRouter() {
+  let middleware = []
 
-router.map(routes, controller)
+  middleware.push(staticFiles('./public'))
```
brookslybrand: I think it's a good idea to have static files serving for `public` dir automatically setup. I've seen users struggle to get their agents to set this up properly for some reason, and I'm not sure why you wouldn't want it out the gate

There may be a few other middleware we should include by default, but I don't have strong opinions there (yet)

brookslybrand: Also, I switched the setting up the middleware more similarly to how the bookstore demo is written so it grows more correctly. Let me know if this should be different.

PR #11367, opened 2026-05-07
Add package README overviews to API docs

brookslybrand:
This adds package-level overview pages to the generated API docs by pulling each package README into the docs build. Package READMEs are emitted at package root paths such as `/api/remix/fetch-router/`, and the sidebar now includes an explicit Overview entry before each package API grouping.

- Generate package overview markdown from `packages/*/README.md`, with fallback content and warnings for missing READMEs.
- Warn when README usage examples still show `@remix-run/*` install/import syntax, matching the existing JSDoc warning style without blocking generation.
- Update README examples to use the Remix 3 `remix` import paths where they are public usage guidance.
- Keep markdown mirror URLs from being prerendered as `*.md/index.html` in local previews.

The README pages are intentionally not included in `api.json`; that lookup remains focused on API symbols.

image

### DIFF
--- a/docs/src/server/prerender.ts
+++ b/docs/src/server/prerender.ts
@@ -78,7 +78,8 @@ async function crawl(router: Router, urlPath: string, outputDir: string) {
     throw error
   }
 
-  let isHtmlFile = response.headers.get('Content-Type')?.includes('text/html')
+  let isMarkdownFile = routes.markdown.match(`http://localhost${urlPath}`) !== null
+  let isHtmlFile = !isMarkdownFile && response.headers.get('Content-Type')?.includes('text/html')
 
   // Always put `index.html` files into directories - this leads to the best
   // support with and without trailing slashes on github pages:
--- a/docs/src/server/prerender.ts
+++ b/docs/src/server/prerender.ts
@@ -28,8 +28,6 @@ console.log('Prerendering versions:\n', JSON.stringify(versions, null, 2))
 const docsRouter = createRouter(versions)
 const crawlableUrls = await getCrawlableUrls(versions)
 
-await fs.rm(outputDir, { recursive: true, force: true })
-
 // Copy static assets to the output directory
 await fs.cp(assetsDir, path.join(outputDir, 'assets'), { recursive: true })
 for (let version of versions || getDefaultVersions()) {
--- a/docs/src/server/prerender.ts
+++ b/docs/src/server/prerender.ts
@@ -7,6 +7,7 @@ import * as semver from 'semver'
 import { type Router } from 'remix/fetch-router'
 import { createRouter, getDefaultVersions } from './router.tsx'
 import type { ServerContext } from './components.tsx'
+import { discoverMarkdownFiles } from './markdown.ts'
 import { routes } from './routes.ts'
 
 let { values: cliArgs } = util.parseArgs({
@@ -25,6 +26,9 @@ const versions = await getVersionsToBuild()
 console.log('Prerendering versions:\n', JSON.stringify(versions, null, 2))
 
 const docsRouter = createRouter(versions)
+const crawlableUrls = await getCrawlableUrls(versions)
+
+await fs.rm(outputDir, { recursive: true, force: true })
 
 // Copy static assets to the output directory
 await fs.cp(assetsDir, path.join(outputDir, 'assets'), { recursive: true })
@@ -108,6 +112,7 @@ async function crawl(router: Router, urlPath: string, outputDir: string) {
         .map((href) => resolveRelativeLink(href!, urlPath))
         .map(stripHash)
         .filter((href) => href !== urlPath)
+        .filter((href) => isCrawlableUrl(href, crawlableUrls))
         .flatMap((href) => {
           let match = routes.docs.match(`http://localhost${href}`)
           return match ? [href, routes.markdown.href(match.params)] : [href]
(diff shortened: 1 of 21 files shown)


docs/src/server/prerender.ts
```diff
@@ -78,7 +84,8 @@ async function crawl(router: Router, urlPath: string, outputDir: string) {
     throw error
   }
 
-  let isHtmlFile = response.headers.get('Content-Type')?.includes('text/html')
+  let isMarkdownFile = routes.markdown.match(`http://localhost${urlPath}`) !== null
+  let isHtmlFile = !isMarkdownFile && response.headers.get('Content-Type')?.includes('text/html')
```
brookslybrand: Before it was generating `/api/remix/assert/function/deepEqual.md/index.html` in the build, now it just generates `/api/remix/assert/function/deepEqual.md`. I verified locally that the `.md` links work as expected

docs/src/server/prerender.ts
```diff
@@ -25,6 +26,9 @@ const versions = await getVersionsToBuild()
 console.log('Prerendering versions:\n', JSON.stringify(versions, null, 2))
 
 const docsRouter = createRouter(versions)
+const crawlableUrls = await getCrawlableUrls(versions)
+
+await fs.rm(outputDir, { recursive: true, force: true })
```
brophdawg11: We don't want to do this because the GH pages repo keeps the old SSG'd versions in sub folders.  The GH workflow takes care of cleaning up "latest" build artifacts

docs/src/server/prerender.ts
brophdawg11: huh - weird, I wonder if the content type on md responses got messed up

docs/src/server/router.tsx
```diff
@@ -62,6 +62,16 @@ export function createRouter(versions: ServerContext['versions']) {
         return respond.file(request, filePath, params.asset)
       },
       async docs({ request, params }) {
+        let markdownMatch = routes.markdown.match(request.url)
```
brophdawg11: This is confusing to me - why would markdown routes hit the docs handler instead of the markdown one?

docs/src/server/prerender.ts
brookslybrand: Ahh nice, thanks

docs/src/server/router.tsx
brookslybrand: Agreed, looking into it further

brookslybrand: Looks like it's getting around a bug. I opened a failing test case in #11368 and refactored a bit to get around it

docs/src/server/prerender.ts
brophdawg11: this may have been related to the routing bug sending it to the wrong handler.  I reverted and it looks ok

```
Crawled /api/remix/ui/theme/type/ThemeVars/ -> ./build/site/api/remix/ui/theme/type/ThemeVars/index.html
Crawled /api/remix/ui/theme/type/ThemeVars.md -> ./build/site/api/remix/ui/theme/type/ThemeVars.md
```

docs/src/server/router.tsx
```diff
@@ -74,6 +74,8 @@ export function createRouter(versions: ServerContext['versions']) {
             docFile.path,
             docFilesLookup,
             params.version,
+            // Don't transform links in README - too many false positives
+            docFile.urlPath !== docFile.package + '/overview',
```
brophdawg11, line 78: README's have a lot of false positives on automatic link detection/enhancement

docs/src/server/view.tsx
```diff
@@ -117,31 +117,24 @@ function Sidebar(
...
-                          key={navPage.path}
-                          href={navPage.path}
-                          aria-current={isPageActive(navPage, currentPath) ? 'page' : undefined}
-                          mix={getNavItemMix(navPage, currentPath)}
-                        >
-                          {navPage.navLabel}
-                        </a>
-                      )
-                    })}
+                    <SidebarGroup registry={registry} group={group} currentPath={currentPath} />
```
brophdawg11, line 134: Flatten overviews to align better without a section heading

docs/src/server/prerender.ts
```diff
@@ -99,17 +96,29 @@ async function crawl(router: Router, urlPath: string, outputDir: string) {
...
+    // is filtered separately in the selector below.
+    let robots = parsedHtml.querySelector('meta[name="robots"]')?.getAttribute('content') || ''
+    let noFollow = robots
+      .split(',')
+      .map((s) => s.trim().toLowerCase())
+      .includes('nofollow')
+
+    if (noFollow) {
+      return { downloadedUrl: urlPath, discoveredUrls: [] }
+    }
```
brophdawg11, line 112: If we just don't crawl links from README pages we can drop the stripHash/crawlableUrls stuff

docs/src/server/view.tsx
brookslybrand: Ah yeah, good catch, that looks much better

docs/src/server/view.tsx
```diff
@@ -30,6 +30,12 @@ export function DocsDocument(handle: Handle<DocsViewProps>) {
               <meta name="robots" content="noindex,nofollow" />
               <meta name="googlebot" content="noindex,nofollow" />
             </>
+          ) : page.docFile?.kind === 'package' ? (
+            // Overview pages (package READMEs) link densely to every API page
+            // in the package; those are already reachable via the sidebar, so
+            // tell crawlers — including our prerender spider — not to follow
+            // links from here. The page itself is still indexable.
+            <meta name="robots" content="nofollow" />
```
brookslybrand, line 38: smort, good thinking

docs/src/server/prerender.ts
brookslybrand: Ahhhh that makes a lot of sense. Thanks for double checking and getting rid of this fix

PR #11368, opened 2026-05-07
route-pattern: do not allow partial matches for variables and wildcards in pathname

brookslybrand:
Do not allow partial matches for variables and wildcards in pathname

```ts
let matcher = createMatcher<string>()
matcher.add('/files/:name.md', 'original')
matcher.add('/files/:name.md.backup', 'backup')

// before: 'original' included since `:name.md` partially matches `readme.md.backup`
matcher.matchAll('https://example.com/files/readme.md.backup').map((match) => match.data)
// ❌ ['backup', 'original']

// after: only matches when the pattern covers the whole segment
matcher.matchAll('https://example.com/files/readme.md.backup').map((match) => match.data)
// ✅ ['backup']
```

### DIFF
--- a/packages/remix/.changes/patch.route-pattern.no-partial-matches-for-dynamic-pathnames.md
+++ b/packages/remix/.changes/patch.route-pattern.no-partial-matches-for-dynamic-pathnames.md
@@ -1,0 +1,17 @@
+Fix `createMatcher` from `remix/route-pattern` so dynamic pathname segments and wildcard continuations only match when they cover the full pathname range being tested.
+
+```ts
+import { createMatcher } from 'remix/route-pattern'
+
+let matcher = createMatcher<string>()
+matcher.add('/files/:name.md', 'markdown')
+matcher.add('/files/:name.md.backup', 'backup')
+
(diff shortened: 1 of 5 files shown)


packages/remix/.changes/patch.route-pattern.no-partial-matches-for-dynamic-pathnames.md
```diff
@@ -0,0 +1,17 @@
+Fix `createMatcher` from `remix/route-pattern` so dynamic pathname segments and wildcard continuations only match when they cover the full pathname range being tested.
```
brookslybrand, line 1: @pcattori for my understanding, when we change a package we _do_ need to add another change file to `packages/remix`? Or was this just leftover?

brookslybrand: Wait, never mind, your updates to the skills clarifies it 🫡

brookslybrand:
Sweet! Thanks @pcattori, looks good

I can't approve because I made this PR, therefore I'm taking all the credit

pcattori (approved):

PR #11375, opened 2026-05-08
Add render middleware package

mjackson:
This adds `@remix-run/render-middleware`, a small fetch-router middleware package for installing a request-scoped render function in request context. A renderer can turn any application-defined input into a `Response`, and actions can call it directly without passing request-specific rendering details through every action.

- Adds the `Renderer` context key and callable `Renderer` type
- Adds `renderWith()` for creating a per-request renderer from the current `RequestContext`
- Adds `ContextWithRenderer` for explicit context composition
- Exposes the package through `remix/render-middleware`
- Moves demo rendering setup into `app/middleware/render.*` and removes demo `app/actions/render.*` helpers
- Updates the default app template to install render middleware in `app/router.ts`, read `Renderer` from actions, and start with a single home route
- Keeps `MiddlewareContext` focused on middleware values; factory-based middleware types use `ReturnType<typeof factory>`

```tsx
import { createRouter, type MiddlewareContext } from 'remix/fetch-router'
import { Renderer, renderWith } from 'remix/render-middleware'

function render() {
  return renderWith(
    ({ router, url }) =>
      function render(node: RemixNode, init?: ResponseInit) {
        let stream = renderToStream(node, {
          async resolveFrame(src) {
            let response = await router.fetch(new URL(src, url))
            return response.body ?? response.text()
          },
        })

        return createHtmlResponse(stream, init)
      },
  )
}

type AppContext = MiddlewareContext<[ReturnType<typeof render>]>

const router = createRouter<AppContext>({ middleware: [render()] })

router.get('/', (context) => {
  return context.get(Renderer)(<HomePage />)
})
```

```tsx
// Before: app/actions/render.tsx
export function render(node: RemixNode, request: Request, init?: ResponseInit) {
  return createHtmlResponse(renderToStream(node), init)
}

// actions imported the static helper
home({ request }) {
  return render(<HomePage />, request)
}

// After: app/middleware/render.tsx
export function render() {
  return renderWith((context) => (node: RemixNode, init?: ResponseInit) => {
    return createHtmlResponse(renderToStream(node), init)
  })
}

type AppContext = MiddlewareContext<[ReturnType<typeof render>]>

// actions read the renderer from request context
home(context) {
  return context.get(Renderer)(<HomePage />)
}
```


mjackson:
@MichaelDeBoey please stop adding the `pkg:*` labels! We don't need/use them for anything.

PR #11382, opened 2026-05-11
Add node-tsx package

markdalgleish:
This adds `@remix-run/node-tsx`, a small Node loader for running `.ts`, `.tsx`, and `.jsx` files with Remix-owned TypeScript and JSX runtime support.

- exposes the package through `remix/node-tsx` and `remix/node-tsx/load-module`
- transforms TypeScript syntax that requires JavaScript output, including enums, runtime namespaces, and parameter properties
- keeps JSX syntax limited to `.tsx`/`.jsx` files while supporting TypeScript transforms in `.ts` files
- reads nearby `tsconfig.json` JSX settings, including automatic runtime import sources and classic factory options
- preserves Node.js module resolution and CommonJS vs ESM behavior from the nearest `package.json` `type`
- updates the app template, demos, docs, UI tooling, and `@remix-run/test` internals to dogfood `remix/node-tsx` instead of depending directly on the `tsx` npm package

Basic usage:

```sh
node --import remix/node-tsx ./server.ts
```

Programmatic usage for loading a module with TypeScript and JSX support scoped to its import graph:

```ts
import { loadModule } from 'remix/node-tsx/load-module'

let mod = await loadModule('./app/server.tsx', import.meta.url)
```

The loader does not typecheck, change Node.js module resolution, apply TypeScript path aliases, or downlevel JavaScript syntax for older runtimes.


packages/test/src/lib/ts-transform.ts
```diff
@@ -20,17 +22,79 @@ const tsconfigCache = new Map<string, TsConfigResult | null>()
...
   filePath: string,
-): Promise<{ code: string }> {
+): Promise<{ code: string; format: 'commonjs' | 'module' }> {
+  let format = getModuleFormat(filePath)
   let loader: 'ts' | 'tsx' = filePath.endsWith('.tsx') ? 'tsx' : 'ts'
 
   let tsConfig = getTsconfig(path.dirname(filePath), 'tsconfig.json', tsconfigCache)
 
   let result = await transform(source, {
+    format: format === 'module' ? 'esm' : 'cjs',
```
brophdawg11: Oh nice, I bet this fixes issues I've seen importing CJS packages into browser tests

packages/test/src/lib/coverage-loader.ts
```diff
@@ -25,7 +25,7 @@ export async function load(
 
   let filePath = fileURLToPath(cleanUrl)
   let source = await readFile(filePath, 'utf-8')
-  let { code } = await transformTypeScript(source, filePath)
+  let { code, format } = await transformTypeScript(source, filePath)
```
brophdawg11: It looks like oxc merged your PR but hasn't released it yet?  Are you planning on leveraging `remix/node-tsx` here once that is merged?  Or oxc directly?

markdalgleish: Definitely one of those two, most likely just using Oxc directly.

mjackson:
I want a more ergonomic `tsx`-style contract for this package.

I added follow-up commit `28742b32b` that:

- transforms `.ts` files through `remix/node-tsx` as well as `.tsx`/`.jsx`, so enums, runtime namespaces, and parameter properties work consistently in app/test code
- removes `erasableSyntaxOnly` from the app/demo/docs configs that run through `remix/node-tsx`, so typecheck accepts what the runtime intentionally supports

PR #11387, opened 2026-05-11
Copy package READMEs into generated Remix package

brookslybrand:
## Summary

Makes package README docs available from the generated `remix` package so installed apps and agent tooling can find module documentation from `node_modules/remix`.

- Copies each available package or source-adjacent README beside the generated `remix/*` source entry under `packages/remix/src`.
- Updates the Remix package generator so README copies are produced from package export metadata and skips non-package directories under `packages/`.
- Updates the default app template skill to direct agents to exact subpath READMEs first, then the nearest parent README when a subpath shares parent package documentation.

## What to focus on

There are a lot of file changes, but that's mostly just copying README files

The main update is `scripts/generate-remix.ts` which copies the READMEs into the`remix` package so Agent's have [a very straightforward pattern to grep](https://github.com/remix-run/remix/pull/11387/changes#diff-96616581280e1d14003fef016fcf36d717ecf1585f735238f6dd3edd49b89982) 
image

**I'm not sure if this is the best way to do it or not.** from playing around with an agent (see https://github.com/remix-run/remix/pull/11387#issuecomment-4433168982 for an demonstration) this worked really well, but we could also just encourage the agent to look in the `@remix-run/*` packages, if we think that's a better approach. We just have to maintain the proper mapping of `remix/*` export to `@remix-run/*` package.


brookslybrand:
This is a good example prompt for validating that agents discover the generated package README docs:

> Say I need to parse the `Accept-Language` header from a request and choose the best supported locale. How should I do this with `remix/headers`? Please show the exact API and a small controller example.

Here was the result in a Codex session (via Zed):


Screenshot 2026-05-12 at 12 21 40 PM

PR #11390, opened 2026-05-12
ci: harden npm publish step

mjackson:
This keeps the publish workflow simple while reducing code execution during the npm trusted publishing step.

- Defaults this workflow's `GITHUB_TOKEN` permissions to `contents: read`.
- Keeps `contents: write` and `id-token: write` scoped to the existing publish job.
- Sets `NPM_CONFIG_PROVENANCE=true` for the publish step.
- Runs the CLI template sync explicitly before publishing, since `@remix-run/cli` needs that generated `template` directory in its package.
- Adds `--ignore-scripts` to each `pnpm publish` command so publish lifecycle scripts do not run while npm trusted publishing credentials are available.

The TanStack npm incident showed that trusted publishing can still be abused if malicious code runs inside a workflow with `id-token: write`; their postmortem notes the normal publish step was bypassed and no npm token was stolen: https://tanstack.com/blog/npm-supply-chain-compromise-postmortem

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -7,6 +7,9 @@ on:
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
+permissions:
+  contents: read
+
 jobs:
   check:
     name: Check release readiness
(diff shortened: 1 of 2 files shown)


mjackson:
Closing this PR because we want to keep the normal package lifecycle hooks during publishing; without disabling publish scripts, the remaining workflow changes are not worth carrying as a separate hardening PR.

PR #11397, opened 2026-05-13
Add UI component demos to the docs site

brophdawg11:
Adds component Examples to the API docs site

Can run locally with `cd docs; pnpm run docs; pnpm run dev`

- Examples are discovered from `packages/ui/**/demos/*.demo.tsx` files
- Demo files are expected to export a default component to be rendered
- Default components should have a JSDoc block with`@name`/`@description` tags that will be rendered above the example
- Demo files should be entirely self contained so the user can see all aspects of the demo
  - They should only import things from `@remix-run/*` packages - no local/relative files
- The displayed source code listing:
  - Rewrites `@remix-run/*` imports → `remix/*` so readers see the published package path
  - Strips the `@name`/`@description` JSDoc block (already shown as the page header)
  - Runs through Prettier with `printWidth: 70` for narrower rendering

Screenshot 2026-05-13 at 12 36 44 PM


packages/ui/src/components/button/demos/basic.demo.tsx
```diff
@@ -0,0 +1,29 @@
+import { css } from '@remix-run/ui'
```
brophdawg11, line 1: Example demo file copied from the existing `ui/demo` app

brophdawg11:
Note: Revert [34d6a83](https://github.com/remix-run/remix/pull/11397/commits/34d6a83979d1af7492c88583a9da89f0ed054ace) once https://github.com/remix-run/remix/pull/11402 lands and we can merge main into here

PR #11399, opened 2026-05-14
Fix SSE demo MessageStream entry ID

cha0s:
Just a tiny fix.

I think the problem is that `href` is encoding the `#` used to denote the module export.


brookslybrand:
Thanks for opening this! Looks like [the latest version of this file/feature](https://github.com/remix-run/remix/blob/da85cd81b64c0c11e1bb08c0f3210b996629f96f/demos/sse/app/actions/public/message-stream.tsx#L6) is correctly using `import.meta.url` now

PR #11409, opened 2026-05-16
fix: don't highlight hyphens in path arguments as single-character flags

numekudi:
The regex used for help text syntax highlighting matched -[A-Za-z] to highlight short flags like -v. However, this also matched hyphens in path arguments like ./my-remix-app, causing -r (from -remix) and -a (from -app) to be incorrectly highlighted.

Fixed by adding a negative lookahead (?![A-Za-z0-9]) to ensure short flags are not followed by more alphanumeric characters.

Before / After
image

image


mjackson:
Thanks for the contribution, @numekudi!

PR #11415, opened 2026-05-19
Add dark mode to docs site

brophdawg11:
Adds dark mode to the docs site.  Toggleable via the button and saves preference in localStorage

Screenshot 2026-05-19 at 1 19 22 PM


docs/src/server/view.tsx
```diff
@@ -445,9 +470,88 @@ function DocsFooter() {
   )
 }
 
+// Dark and light theme variable overrides used in both the system-preference
+// media query and the explicit data-color-scheme attribute override blocks.
+const DARK_VARS = {
```
brookslybrand: Small nit: you can use [css `light-dark`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/light-dark) function instead. Cleans this up pretty nicely instead of the duplicate. It's fine either way, I just like the single set of vars personally

brophdawg11: oh cool yeah I saw that as an option in the shiki docs but didn't dig deeper - I'll make that switch

PR #11419, opened 2026-05-20
Migrate demo esbuild usage to `assets`

markdalgleish:
This brings the `demos` directory in line with our new asset server usage patterns.


demos/frame-navigation/app/middleware/asset-entry.ts
```diff
@@ -0,0 +1,31 @@
...
+const assetEntryKey = createContextKey<AssetEntry>()
+const defaultScriptEntry = path.resolve(import.meta.dirname, '../assets/entry.tsx')
+
+export function loadAssetEntry(
+  scriptEntry = defaultScriptEntry,
+): Middleware<{ key: typeof assetEntryKey; value: AssetEntry }> {
+  return async (context, next) => {
+    let [scriptSrc, scriptPreloads] = await Promise.all([
+      assetServer.getHref(scriptEntry),
+      assetServer.getPreloads(scriptEntry).catch(() => []),
```
mjackson: I always hate to see an empty `catch` like this. Either something went wrong that we shouldn't ignore, or we can provide a default value from `getPreloads` when there are no preloads.

mjackson (changes requested):

demos/frame-navigation/app/middleware/asset-entry.ts
markdalgleish: What do you think about switching asset server APIs like this to use a result type rather than throwing? Looking at this with fresh eyes, I feel like this would be better.

With preloads, we use an empty `catch` so that any errors with your client assets don't bubble up to a `500` response for the entire page. The thinking here is that it's better to fall back to a rendered HTML page with broken client scripts than rendering nothing, but this is more of a product decision. I feel like having a result type would make it clearer that this is a product decision you need to make.

markdalgleish: I've pushed an update now where we do this for all `getPreloads` usage:

```ts
assetServer.getPreloads(scriptEntry).catch((error) => {
  // Surface asset compilation errors without breaking HTML rendering.
  console.error(error)
  return []
})
```

PR #11437, opened 2026-05-20
Fix node-tsx built loadModule hook path

brookslybrand:
I hit this while upgrading `remix-website`. I am not completely sure this is the right shape for the fix, but the published `@remix-run/node-tsx@0.1.0` package appears to have a real built-output/runtime issue.

The published loader registers a TypeScript source file from `dist/lib/load-module.js`:

```js
register(
  new URL(`./register-hooks.ts?namespace=${encodeURIComponent(namespace)}`, import.meta.url),
  {
    data: { namespace },
    parentURL: import.meta.url,
  },
)
```

But the built package only has the emitted JavaScript hook:

```txt
dist/lib/register-hooks.js
```

That means `remix test` and `node --import remix/node-tsx` can ask Node to import a `.ts` file that is not present in the published package.

- Updates `loadModule()` so the built loader registers the emitted hook file:

```diff
- ./register-hooks.ts
+ ./register-hooks.js
```

- Adds a source-mode `register-hooks.js` shim so workspace/source execution can still resolve the same specifier before build.
- Adds coverage that builds `@remix-run/node-tsx`, installs the built `dist` behind publish-style exports in a temp project, and imports `@remix-run/node-tsx/load-module` from that built package before loading a TS/TSX graph.

### DIFF
--- a/packages/node-tsx/.changes/patch.register-built-hooks.md
+++ b/packages/node-tsx/.changes/patch.register-built-hooks.md
@@ -1,0 +1,1 @@
+Fix `loadModule()` in built packages so it registers the emitted JavaScript hooks file instead of a missing TypeScript source file.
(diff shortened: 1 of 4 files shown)


brookslybrand:
@markdalgleish this is 100% vibe-coded to just try to point out the issue I ran into upgrading `remix-website`. Please feel free to close this and tackle it properly

mjackson:
Superseded by #11441.

PR #11438, opened 2026-05-20
Fix pnpm package import resolution in assets

brookslybrand:
pnpm can serve a package through a public node_modules symlink while storing that package's dependencies beside the package's real virtual-store path. When the asset server preserved the public symlink path as the resolver importer, bare imports inside served package facades could fail before fileMap validation had a chance to allow the real dependency path.

- Adds pnpm-style regression coverage for a served remix package facade that re-exports @remix-run/ui from a nested virtual-store dependency.
- Resolves bare package specifiers from the transformed module's real filesystem path so pnpm package dependencies are visible to the resolver.
- Keeps relative and URL-like specifiers resolving from the public identity path so existing symlink URL identity behavior is preserved.


brookslybrand:
This PR was completely AI generated. Just trying to get the issue surfaced. Feel free to rework it or solve in a better way.

PR #11448, opened 2026-05-20
Prevent static middleware symlink escapes

mjackson:
staticFiles() resolved request candidates with `stat()`, so a symlink inside the configured root could point outside that root and still be served. This changes file and index resolution to compare real paths and fall through when a candidate escapes the configured root.

- Blocks symlinked files and symlinked directories that resolve outside `staticFiles()` root
- Preserves the existing fallback behavior for missing or inaccessible roots and files
- Adds regression coverage for direct file, nested directory, and directory-index symlink escapes


markdalgleish (approved):
Nice one 👍

PR #11451, opened 2026-05-20
Encode route href path params

mjackson:
`createHref()` inserted path params directly into generated URLs, so values containing `/`, `?`, `#`, or leading separators could escape their route segment and even produce scheme-relative URLs. This encodes pathname params while preserving wildcard path separators where they are part of the wildcard value.

- Encodes dynamic `:param` path values with `encodeURIComponent()`
- Encodes wildcard path segments while preserving internal `/` separators
- Encodes leading wildcard separators so generated hrefs remain same-origin relative paths


mjackson:
Superseded by #11454.

PR #11454, opened 2026-05-20
Encode route-pattern pathname params

mjackson:
createHref was interpolating pathname params directly into generated URLs. Values containing `/`, `?`, `#`, or dot segments could change the URL structure instead of staying inside the route param.

- Encode variable pathname params before interpolation.
- Encode wildcard params segment-by-segment so intentional slash structure is preserved.
- Decode encoded reserved characters during matching so href-generated params round-trip through `match()`.


pcattori:
Superseded by #11468

PR #11455, opened 2026-05-20
Report test lifecycle hook failures

mjackson:
Lifecycle hooks that throw after a test file registers can currently be swallowed by the executor, letting `remix-test` report a successful run even though setup or teardown failed.

- Records `beforeAll` and `afterAll` hook errors as failed hook results in their suite
- Marks the affected test as failed when `afterEach` throws, including the hook error in the result
- Adds executor regression coverage for `beforeAll`, `afterEach`, and `afterAll` failure reporting


brophdawg11 (approved):
nice - yeah these should surface as failures

PR #11460, opened 2026-05-21
Enforce zero-argument UI render functions

brookslybrand:
Remix UI components now read JSX props from `handle.props`, but the runtime and public types still allowed the older pattern where props were typed on the returned render function. This tightens the component contract so component render functions are truly zero-argument.

This came up while updating old component conventions in https://github.com/remix-run/remix-website/pull/459.

- Makes `RenderFn` zero-argument and stops passing `handle.props` into component render callbacks.
- Updates JSX and `clientEntry` types so component props are inferred from `Handle<Props>`, while mixin render callbacks still receive host props.
- Updates the repo demos and app template that still used the old props-on-render convention.
- Adds type and runtime coverage for the contract, plus a breaking change note for `@remix-run/ui`.
- Adds guidance to the change-file skill to avoid rewriting historical changelog entries when current docs need updating.

### DIFF
--- a/.agents/skills/make-change-file/SKILL.md
+++ b/.agents/skills/make-change-file/SKILL.md
@@ -67,6 +67,12 @@ Write release notes that match this repository's `.changes` conventions. Use it 
 - Do not manually hard-wrap prose in `.changes/*.md` files. Keep each paragraph or bullet on a single source line and let rendered changelogs wrap naturally.
 - Use flat bullets only when they add clarity. Short paragraphs are usually better.
 
+## Historical Changelogs
+
+- Treat existing `CHANGELOG.md` files and historical changelog docs as records of what was true at the time. Do not rewrite old entries to match current APIs, terminology, or guidance.
+- If current guidance is stale or confusing, update current docs, README files, skills, or add a new `.changes` file instead of editing historical release prose.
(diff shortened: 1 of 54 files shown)


.agents/skills/make-change-file/SKILL.md
```diff
@@ -67,6 +67,12 @@ Write release notes that match this repository's `.changes` conventions. Use it
 - Do not manually hard-wrap prose in `.changes/*.md` files. Keep each paragraph or bullet on a single source line and let rendered changelogs wrap naturally.
 - Use flat bullets only when they add clarity. Short paragraphs are usually better.
 
+## Historical Changelogs
+
+- Treat existing `CHANGELOG.md` files and historical changelog docs as records of what was true at the time. Do not rewrite old entries to match current APIs, terminology, or guidance.
+- If current guidance is stale or confusing, update current docs, README files, skills, or add a new `.changes` file instead of editing historical release prose.
+- Only edit historical changelog entries for narrow corrections such as broken links, typos, or clearly invalid references, and prefer doing so only when the user explicitly asks.
+
```
brookslybrand, line 75: Added this section because my agent was trying to change the changelog 🙈

PR #11461, opened 2026-05-21
Add Pagefind search to prerendered docs

brophdawg11:
~~**Stacked on #11504**~~

Adds [Pagefind](https://pagefind.app/) full-text search to the prerendered docs site. Pagefind runs as a `postprerender` step, indexing the static HTML output, and the modal UI is loaded in the browser via custom elements (`<pagefind-modal>` / `<pagefind-modal-trigger>`).

To test: `cd docs; pnpm run docs: pnpm run prerender; pnpm run prerender:serve`

- Compact search icon trigger next to the logo opens a search modal on click
- Dev server gracefully serves a pre-built Pagefind index from `build/site/pagefind/` if present, falling back silently if not

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -29,7 +29,6 @@
     "docs": "node src/generate/index.ts",
     "docs:debug": "DEBUG=1 node src/generate/index.ts",
     "prerender": "node --import remix/node-tsx src/prerender/index.ts",
-    "postprerender": "pagefind --site build/site",
     "prerender:serve": "npx http-server -p 44100 build/site",
     "serve": "node --import remix/node-tsx src/server/index.ts",
     "test": "remix test",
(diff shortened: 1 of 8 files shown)


packages/ui/src/animation/spring.ts
```diff
@@ -144,6 +144,28 @@ function transition(
...
+ *
+ * @example
+ * // Custom parameters
+ * let s = spring({ duration: 500, bounce: 0.2 })
+ * element.animate(keyframes, { ...s })
+ *
+ * @example
+ * // Transition helper for multiple properties
+ * element.style.transition = spring.transition(['transform', 'opacity'], 'snappy')
+ */
```
brophdawg11, line 168: Prerendering was busted without this because something else links to `spring` in a JSDoc which causes our spidering process to try to crawl it, but the lack of JSDocs meant it would 404 which fails the prerender process.

brophdawg11:
This is on hold temporarily - the frame reload doesn't play nice with the pagefind web component so gotta figure out the best way to handle that

brophdawg11:
CI failures should be fixed by https://github.com/remix-run/remix/pull/11536

PR #11477, opened 2026-05-26
Implement and leverage crawl API in docs code

brophdawg11:
This brings the `crawl` API from https://github.com/remix-run/remix/pull/11150 into `docs/src` so we can use that improved API directly from within the docs directory.  It's nicely isolated there if we ever want to bring that over as a first class `fetch-router` API.


packages/ui/src/animation/spring.ts
```diff
@@ -144,6 +144,28 @@ function transition(
...
+ *
+ * @example
+ * // Custom parameters
+ * let s = spring({ duration: 500, bounce: 0.2 })
+ * element.animate(keyframes, { ...s })
+ *
+ * @example
+ * // Transition helper for multiple properties
+ * element.style.transition = spring.transition(['transform', 'opacity'], 'snappy')
+ */
```
brophdawg11, line 168: Needed to make docs pass currently because something links off to this API which causes a 404 during prerendering

PR #11481, opened 2026-05-28
Fix browser test package resolution

### LINKED ISSUE #11478: Browser tests resolve clsx to CJS while asset server resolves ESM
While working on `remix-website`, I hit a browser test failure because it could not import `clsx`. It kept reaching for the CJS version, even though `remix/asset` properly resolved ESM. The related website change is here, where `clsx` was vendored as a workaround:

https://github.com/remix-run/remix-website/pull/462/changes/60c541ba7eea0ffd7c5fc79e6ffa3f5d35b12889#diff-931e17de1a4cf1ae4560f5314e55d96258e3f8a7d41c26f6266f63853fad37da

A minimal browser test repro is:

```ts
   // app/components/mobile-menu.tsx
   import cx from 'clsx'

   export function MobileMenu(handle: Handle<{ isOpen: boolean }>) {
     return () => {
       let { isOpen } = handle.props

       return (
         <nav
           aria-label="Mobile navigation"
           className={cx('mobile-menu', {
             'mobile-menu--open': isOpen,
             'mobile-menu--closed': !isOpen,
           })}
         >
           <a href="/docs">Docs</a>
           <a href="/blog">Blog</a>
         </nav>
       )
     }
   }

   // app/components/mobile-menu.test.browser.tsx
   import * as assert from '@remix-run/assert'
   import { describe, it } from '@remix-run/test'
   import { render } from '@remix-run/ui/test'
   import { MobileMenu } from './mobile-menu.tsx'

   describe('MobileMenu', () => {
     it('renders the open state class', (t) => {
       let { $, cleanup } = render(<MobileMenu isOpen />)
       t.after(cleanup)

       assert.equal(
         $('[aria-label="Mobile navigation"]')?.className,
         'mobile-menu mobile-menu--open',
       )
     })
   })
```

When run as a Remix browser test, this resolves `clsx` to the CJS file:

```txt
/scripts/node_modules/.pnpm/clsx@2.1.1/node_modules/clsx/dist/clsx.js
```

and the browser fails with:

```txt
The requested module '/scripts/node_modules/.pnpm/clsx@2.1.1/node_modules/clsx/dist/clsx.js' does not provide an export named 'default'
```

This appears to be a drift between the browser test runner resolver and the asset server resolver:

- Browser test runner rewrites bare imports using `createRequire(importerFile).resolve(spec)`: https://github.com/remix-run/remix/blob/dc6ab4c18ee4d6f423ca50b7fd30bb3ca5022991/packages/test/src/app/server.ts#L268-L301
- `remix/asset` uses browser/ESM-oriented resolver conditions: https://github.com/remix-run/remix/blob/dc6ab4c18ee4d6f423ca50b7fd30bb3ca5022991/packages/assets/src/lib/scripts/compiler.ts#L116-L123

So `remix/asset` serves `clsx` from the ESM path, but browser tests currently serve it from the CJS/default path.

brophdawg11:
Browser tests currently rewrite bare imports with Node's createRequire, which follows CommonJS-oriented package exports. That makes packages like clsx resolve to their CJS entry even though the browser asset pipeline resolves the ESM/browser entry.

This updates the browser test harness to resolve bare imports with oxc-resolver using browser and ESM-oriented conditions, matching the asset server behavior.

- Fixes browser test package resolution for packages with separate import/require exports
- Adds a browser regression test that imports clsx as a default export
- Adds a patch change file for @remix-run/test

Closes #11478


packages/test/package.json
```diff
@@ -85,6 +86,7 @@
     "@types/istanbul-reports": "^3.0.4",
     "@types/node": "catalog:",
     "@typescript/native-preview": "catalog:",
+    "clsx": "^2.1.1",
```
brookslybrand: kind of interesting to add a package just to test something, but not sure if there's really a better way to 🤔

Either way, this should maybe be pinned

brookslybrand (approved):
Tested on `remix-website` and confirmed that importing `clsx` now works with this preview build

PR #11482, opened 2026-05-28
Fix recursive CSS nesting in UI css mixin

### LINKED ISSUE #11459: remix/ui: css mixin doesn't handle 2+ levels of nesting
It seems the `css` mixin only properly renders css for one level of nesting.  

Repro: save this as `bug.ts` in the root of the repo and and run `node bug.ts`:

```ts
import { createElement, css } from './packages/ui/src/index.ts'
import { renderToString } from './packages/ui/src/server/stream.ts'

let app = createElement(
  'h1',
  {
    mix: css({
      color: 'red',

      // ✅ works
      '& div': {
        color: 'green',

        // ❌ broken - prints "& p: [object Object];"
        '& p': {
          color: 'blue',
        },
      },
    }),
  },
  'Hello Remix',
)

console.log(await renderToString(app))
```

**Expected Output**

```
<head><style data-rmx="rmxc-espqmu">@layer rmx.rmxc-espqmu { .rmxc-espqmu {
  color: red;
  & div {
    color: green;
    & p {
      color: blue;
    }
  }
} }</style></head><h1 class="rmxc-espqmu">Hello Remix</h1>
```

**Actual Output**

```
<head><style data-rmx="rmxc-espqmu">@layer rmx.rmxc-espqmu { .rmxc-espqmu {
  color: red;
  & div {
    color: green;
    & p: [object Object];
  }
} }</style></head><h1 class="rmxc-espqmu">Hello Remix</h1>
```

brookslybrand:
Fixes recursive nested selector rendering in the UI `css(...)` mixin. Previously the first nested selector level rendered correctly, but deeper selector objects were treated as declaration values and serialized as `[object Object]` (fixes #11459).

- Render nested selector bodies recursively while preserving authored selector text.
- Keep disabled conditional nested selectors skipped when their values are not objects.
- Tighten the existing `processStyleClass` nested selector test to cover pseudo selectors, recursive descendant selectors, media queries, and the reported `[object Object]` regression.


packages/ui/src/style/style.ts
```diff
@@ -189,6 +182,29 @@ function styleToCss(styles: StyleObject, selector: string = ''): string {
   return css
 }
 
+function nestedStyleBodyToCss(styles: StyleObject, spaces: number): string {
+  let pad = ' '.repeat(spaces)
+  let declarations: string[] = []
+  let nestedBlocks: string[] = []
+
+  for (let [key, value] of Object.entries(styles)) {
+    if (isComplexSelector(key)) {
```
brophdawg11, line 194: I'm not entirely sure what the expectation is here but above we have a bunch more logic once we identify a complex selector - do we need to be handling that here?  Can you do deeply nested complex selectors?

brookslybrand: Yeah good question, I worked with the agent on this some more:

> The extra top-level logic is for placing root at-rules like media queries/keyframes relative to the generated class rule. Once we’re already rendering inside a CSS rule body, nested selectors and nested conditional at-rules can remain nested in that body. I added a comment and coverage for a deeply nested selector plus nested media query to make that expectation explicit.

We added a comment and expanded the test. I also manually tested out in the bookstore app and nested styles worked as expected, even with complex characters like `@` rules.

```tsx
// book-card.tsx

export function BookCard(handle: Handle<BookCardProps>) {
  return () => {
    let { book } = handle.props

    return (
      <div
        class="book-card"
        data-test-slug={book.slug}
        mix={css({
          '&:hover': {
            outline: '4px solid #8e44ad',
            '& .nested-css-title': {
              color: '#8e44ad',
              '&::after': {
                content: '" nested css"',
                color: '#e67e22',
                fontSize: '0.75rem',
                marginLeft: '0.5rem',
                textTransform: 'uppercase',
              },
            },
            '@media (hover: hover)': {
              '& .nested-css-price': {
                background: '#fff3cd',
                color: '#c0392b',
                transform: 'rotate(-2deg) scale(1.08)',
              },
            },
          },
        })}
      >
        <img src={book.cover_url} alt={book.title} />
        <div class="book-card-body">
          <h3>{book.title}</h3>
          <h3 class="nested-css-title">{book.title}</h3>
          <p class="author">by {book.author}</p>
          <p class="price">${book.price.toFixed(2)}</p>
          <p class="price nested-css-price">${book.price.toFixed(2)}</p>
          <div mix={css({ display: 'flex', gap: '0.5rem', alignItems: 'center' })}>
            <a href={routes.books.show.href({ slug: book.slug })} class="btn">
              View Details
            </a>

            <Frame src={routes.fragments.cartButton.href({ bookId: book.id })} />
          </div>
        </div>
      </div>
    )
  }
}
```

https://github.com/user-attachments/assets/50a59a9a-d04b-41ac-8804-35eeffbff9e0

PR #11485, opened 2026-05-29
fix(ui-demo): add @ssr false opt-out to prevent server crash on browser-only demos

johnsonav1992:
The demo runner calls `loadDemoModule()` on the server for every demo, which causes a crash when a demo imports browser-only code that runs side effects at module load time.

The specific failure: `animation.demo.tsx` imports `reordering.tsx`, which calls `scheduleNextShuffle()` during setup. That schedules a `setTimeout` which eventually calls `handle.update()`, which throws `scheduleUpdate not implemented` in the server path. The error is uncaught and kills the Node process.

**Fix:** demos can opt out of server-side rendering by adding `@ssr false` to their JSDoc metadata block,  the same block already used for `@name`, `@description`, `@layout`, and `@order`. When set, the server skips executing the module and uses a no-op component instead. The browser still receives and hydrates the full demo bundle.

```ts
/**
 * @name Animation
 * @ssr false
 */
```

Also fixed the `motion.dev` link in the animations demo, as it was broken.

### DIFF
--- a/packages/ui/demo/README.md
+++ b/packages/ui/demo/README.md
@@ -15,3 +15,7 @@ Then open `http://localhost:44100`.
 The index scans `packages/ui` for `*.demo.ts` and `*.demo.tsx` files on every request.
 Each demo is available at `/demo/*filename`, where `*filename` is the demo file path
 relative to `packages/ui`.
+
+Use JSDoc metadata in the demo file comment to customize behavior. For demos that
+should only run in the browser, add `@ssr false` to disable server rendering for
+that demo route while still hydrating from the browser asset bundle.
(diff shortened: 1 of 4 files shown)


brookslybrand:
Thanks for opening this fix! The demo runner has been refactored a good bit since then and I don't believe this issue exists any longer (I wasn't able to reproduce at least). If there is still an bug please open up an issue, although in all likelihood these demos will be moved up to the top-level `demos/` dir.

johnsonav1992:
@brookslybrand yes, I figured. All good! This was wayyyy back right after it was released. I'll poke around later and see if there is anything else I see and I'd be happy to help out if needed!

PR #11491, opened 2026-05-29
Add remix test timeouts

mjackson:
Add Node-aligned timeout and abort signal support to `@remix-run/test` so tests and lifecycle hooks can fail promptly when async work hangs. Timed-out tests now abort `t.signal`, which lets application code pass the test signal through to APIs like `fetch`.

- Adds `{ timeout, signal }` support to `it()`/`test()` and lifecycle hooks
- Exposes `t.signal` on `TestContext` and aborts it when timeout/user cancellation triggers
- Preserves hook ordering while tracking hooks as runnable records with their own options
- Documents the new API and adds change files for `@remix-run/test` and `remix/test`

```ts
it('loads data', { timeout: 5_000 }, async (t) => {
  let response = await fetch('/api/data', { signal: t.signal })
  assert.equal(response.status, 200)
})

beforeEach(
  async () => {
    await resetDatabase()
  },
  { timeout: 1_000 },
)
```

### DIFF
--- a/packages/test/src/lib/executor.ts
+++ b/packages/test/src/lib/executor.ts
@@ -2,6 +2,8 @@ import { createTestContext, type CreateTestContextE2EOptions, type TestContext }
 import type { V8CoverageEntry } from './coverage.ts'
 import type { TestResult, TestResults } from './reporters/results.ts'
 
+type PendingMeta = boolean | string
+
 interface RunnableOptions {
   timeout?: number
   signal?: AbortSignal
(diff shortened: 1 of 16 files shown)


packages/test/src/lib/executor.ts
```diff
@@ -1,31 +1,38 @@
...
   tests: RegisteredTest[]
   only?: boolean
-  skip?: boolean
-  todo?: boolean
-  beforeEach?: LifecycleHook
-  afterEach?: LifecycleHook
-  beforeAll?: LifecycleHook
-  afterAll?: LifecycleHook
+  skip?: boolean | string
+  todo?: boolean | string
```
brophdawg11: What's the use case for a `string` type here?

brophdawg11 (approved):
Nice improvement 👍

packages/test/src/lib/executor.ts
mjackson: Yep, that was incomplete as written. I added result/reporter support for string skip/todo reasons in https://github.com/remix-run/remix/commit/6a28469d0662089bcedb3e5fb247e921b7416cef.

PR #11492, opened 2026-05-29
Require middleware to explicitly continue

mjackson:
This changes fetch-router middleware from implicit continuation to explicit continuation. Middleware must now return a `Response` or call `next()`; middleware that returns without calling `next()` throws so missing continuation bugs fail loudly.

- Narrows the public `Middleware` type to `Response | Promise<Response>` and removes the runtime auto-`next()` fallback.
- Updates built-in middleware and demos that only populated context to explicitly `return next()`.
- Updates router docs and release notes with the migration pattern.

```ts
// Before
function loadUser(): Middleware {
  return (context) => {
    context.set(CurrentUser, user)
  }
}

// After
function loadUser(): Middleware {
  return (context, next) => {
    context.set(CurrentUser, user)
    return next()
  }
}
```

Middleware that needs to observe the downstream response should keep returning the response after awaiting it:

```ts
function logger(): Middleware {
  return async (context, next) => {
    let response = await next()
    console.log(context.request.url, response.status)
    return response
  }
}
```

### DIFF
--- a/demos/bookstore/app/middleware/admin.ts
+++ b/demos/bookstore/app/middleware/admin.ts
@@ -8,11 +8,13 @@ import { getCurrentUser } from '../utils/context.ts'
  * Must be used after requireAuth middleware.
  */
 export function requireAdmin(): Middleware {
-  return () => {
+  return (_, next) => {
     let user = getCurrentUser()
 
     if (user.role !== 'admin') {
       return new Response('Forbidden', { status: 403 })
     }
+
+    return next()
   }
 }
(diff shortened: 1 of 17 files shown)


brophdawg11 (approved):
yeah this explicitness is nice

brookslybrand (approved):
Seems reasonable to me! I imagine from experience with other APIs that agents will "like" this explicitness more anyway, if they weren't already doing it

I don't think it degrades from the human-who-writes-code experience at all either.

PR #11493, opened 2026-05-29
Keep generated remix README mirrors out of git

mjackson:
This keeps generated `packages/remix/src/**/README.md` mirrors out of git while preserving them in the published `remix` package for agent discovery in `node_modules/remix/src/<subpath>/README.md`.

- Ignores the generated README mirrors and removes the checked-in copies from `packages/remix/src`.
- Adds a focused `sync-remix-readmes` step that copies canonical READMEs from the owning `@remix-run/*` package into the generated `remix` source tree during `prepack`, then cleans them up during `postpack`.
- Adds a publish workflow guard that packs `packages/remix` and verifies representative README mirrors are present in the tarball.
- Updates docs package overview discovery and the Remix agent skill instructions so source-package READMEs remain the canonical docs in git while published `remix` still exposes mirrored docs.


brophdawg11:
I think this was mostly a convenience for agents - so they can find the README in `node_modules/remix/src/...` and don't have to traverse deeper into `node_modules/remix/node_modules/@remix-run/<pkg>/`.  If we are confident they won't trip up on that then this should be 👍 

@brookslybrand you think this'll be ok for the agents discovery?

brookslybrand:
@brophdawg11 that's correct.

Whatever we do, we need to update the instructions here to be clear: https://github.com/remix-run/remix/blob/main/.agents/skills/remix/SKILL.md?plain=1#L10-L17

So long as we make it super clear and deterministic where the agent should be looking for good docs on any given API, I don't care how they're packaged up.

mjackson:
I updated the PR so it keeps subpackage READMEs in the `remix` package using prepack/postpack. This is the most logical place for an agent to look for them, so we should leave them there. This PR just keeps the duplicated READMEs out of git.

PR #11496, opened 2026-06-01
Improve default template asset and frame defaults

brookslybrand:
This improves the default `remix new` app template so fresh projects start with production-minded asset behavior and basic frame rendering defaults without adding extra middleware.

- Sets `NODE_ENV` explicitly in the template scripts so development, production, and test behavior are predictable
- Uses `node --watch` in the development script so server-rendered modules and browser asset output restart from the same source changes
- Minifies browser assets outside development while preserving external sourcemaps during development
- Disables the asset server watcher so it does not update browser assets independently from the running server modules
- Resolves client and server `<Frame>` content by default

This follows the improvements explored in dwoodwardgb/remix3-demo#1 after a user noticed the starter app was serving non-minified browser modules by default.

### DIFF
--- a/template/app/assets.ts
+++ b/template/app/assets.ts
@@ -21,13 +21,7 @@ export const assetServer = createAssetServer({
     '../packages/**',
     /* remix-template:remove-end */
   ],
-  deny: ['app/**/*.server.*'],
   sourceMaps: isDevelopment ? 'external' : undefined,
   minify: !isDevelopment,
   watch: false,
-  scripts: {
-    define: {
-      'process.env.NODE_ENV': JSON.stringify(nodeEnv),
-    },
-  },
 })
--- a/template/app/assets.ts
+++ b/template/app/assets.ts
@@ -1,6 +1,8 @@
 import { createAssetServer } from 'remix/assets'
 
 const rootDir = process.cwd()
+const nodeEnv = process.env.NODE_ENV ?? 'development'
+const isDevelopment = nodeEnv === 'development'
 
 export const assetServer = createAssetServer({
   basePath: '/assets',
@@ -20,10 +22,12 @@ export const assetServer = createAssetServer({
     /* remix-template:remove-end */
   ],
   deny: ['app/**/*.server.*'],
-  sourceMaps: process.env.NODE_ENV === 'development' ? 'external' : undefined,
+  sourceMaps: isDevelopment ? 'external' : undefined,
+  minify: !isDevelopment,
+  watch: false,
   scripts: {
     define: {
-      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV ?? 'development'),
+      'process.env.NODE_ENV': JSON.stringify(nodeEnv),
     },
   },
 })
--- a/template/package.json
+++ b/template/package.json
@@ -3,9 +3,9 @@
   "private": true,
   "type": "module",
   "scripts": {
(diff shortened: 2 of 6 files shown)


template/app/assets.ts
```diff
@@ -20,10 +22,12 @@ export const assetServer = createAssetServer({
...
-  sourceMaps: process.env.NODE_ENV === 'development' ? 'external' : undefined,
+  sourceMaps: isDevelopment ? 'external' : undefined,
+  minify: !isDevelopment,
+  watch: isDevelopment,
   scripts: {
     define: {
-      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV ?? 'development'),
+      'process.env.NODE_ENV': JSON.stringify(nodeEnv),
     },
   },
```
brookslybrand: @markdalgleish I'm not sure I totally understand why we need to set this here.

Also, while you're here would you verify that the template's asset config is setup well for a basic getting started experience? I opened this PR after someone ran the prod server, ran a lighthouse test, and was concerned that the JS wasn't minified 😅

template/package.json
```diff
@@ -3,9 +3,9 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "dev": "node --watch --import remix/node-tsx server.ts",
-    "start": "node --import remix/node-tsx server.ts",
-    "test": "node --import remix/node-tsx --test",
+    "dev": "NODE_ENV=development node --watch --import remix/node-tsx server.ts",
+    "start": "NODE_ENV=production node --import remix/node-tsx server.ts",
+    "test": "NODE_ENV=test node --import remix/node-tsx --test",
```
brookslybrand, line 8: If we're gonna rely on NODE_ENV in `assets.ts` I think we need to set them here, but happy for an alternative

template/app/router.ts
```diff
@@ -14,7 +15,7 @@ declare module 'remix/router' {
 }
 
 export const router = createRouter<AppContext>({
-  middleware: [staticFiles('./public', { index: false }), render()],
+  middleware: [compression(), staticFiles('./public', { index: false }), render()],
```
brookslybrand: Like minifying JS, seems like we should be compressing by default too.

Any other middleware we should add to the default getting started experience?

brookslybrand:
No idea why CI is failing

template/app/assets.ts
markdalgleish: Using `define` to inject a value for `process.env.NODE_ENV` is only needed if `NODE_ENV` is being referenced in browser-facing JS, which doesn't seem to be the case in the template or any of our demos. We should be able to drop this. This only needs to be here if we're wanting to ensure that any third party packages using `NODE_ENV` work without any config changes - but we can probably live without this to keep the config clean.

Along those lines, the only thing that sticks out to me here is `deny: ['app/**/*.server.*'],`. Only the `assets` dir and `node_modules` are allowed, and in this layout it would be odd to put a `.server.ts` file in the `assets` directory. I wouldn't personally suggest this as an out-of-the-box file naming convention, so IMO it's better to leave this out, especially since right now it's just dead config.

brookslybrand: Ahhh duh, that makes complete sense!

Yeah, I think you're right, for the starter template we can drop this. For folks who need that I think going to the asset docs and learning how to add that is perfectly fine.

Yep, agreed on the `deny`, that doesn't really make sense in this context. I'll remove both of those!

PR #11504, opened 2026-06-02
Add rmx-preserve-dom support to UI reconciler

brophdawg11:
Adds an `rmx-preserve-dom` attribute for Remix UI frame reconciliation so client-owned DOM subtrees can preserve their live DOM once their host element has been matched.

- Preserves matched `rmx-preserve-dom` element attributes and children during DOM diffs.
- Avoids reconnecting keyed preserved elements during child reordering, which protects custom element lifecycle state.
- Keeps SSR and initial hydration behavior intact while preventing later frame-driven DOM updates from taking ownership of the preserved subtree.
- Documents when to use the attribute, why `data-key` matters, and why Remix-owned client entries that need future frame props should stay outside the preserved boundary.

```tsx
<my-custom-element data-key="search" rmx-preserve-dom />
```


brophdawg11:
Keeping this in draft - I saw some flakiness in the docs search branch so unsure if this is a 100% fix yet

packages/ui/src/runtime/diff-dom.ts
```diff
@@ -14,6 +14,8 @@ type CommentMarkerRangeReplacement = {
   nextEndIndex: number
 }
 
+const REMIX_IGNORE_ATTRIBUTE = 'rmx-ignore'
```
sergiodxa: Wouldn't be more correct if it was a data-* attribute? Like `data-rmx-ignore`

brophdawg11: I wanted Ryan to make the call there.  I followed the pattern used by `rmx-target`/`rmx-src`/`rmx-reset-scroll`.  

I do see a few `data-rmx-` versions in the repo though (`data-rmx-accordion-indicator`, `data-rmx-theme`, `data-rmx-theme-selector`).  Not sure if there's an existing pattern for when to choose one versus the other or if they will be aligned before stable.

brophdawg11:
CI failures should be fixed by https://github.com/remix-run/remix/pull/11536

kelvinle30 (approved):

PR #11505, opened 2026-06-02
Add version flag to docs prerender

mjackson:
## Summary
- add a `--version` option to the docs prerender script
- seed the crawler from only the home and lookup routes, with an optional version prefix
- keep docs hrefs version-aware at render time instead of rewriting them during prerender
- repurpose version discovery for the picker options only

## Verification
- `pnpm --filter remix-the-docs typecheck`
- `pnpm --filter remix-the-docs test`
- `pnpm --filter remix-the-docs build`
- root prerender smoke with no `--version`
- versioned prerender smoke with `--version v3.1.2`
- `git diff --check`

### DIFF
--- a/docs/src/prerender/index.ts
+++ b/docs/src/prerender/index.ts
@@ -156,7 +156,5 @@ async function getVersionsForPicker(activeVersion?: string): Promise<Versions> {
     })
   }
 
-  return remixVersions.length > 0
-    ? remixVersions.map((version) => ({ version }))
-    : getDefaultVersions()
+  return remixVersions.length > 0 ? remixVersions : getDefaultVersions()
 }
--- a/docs/src/prerender/index.ts
+++ b/docs/src/prerender/index.ts
@@ -17,8 +17,16 @@ let { values: cliArgs } = util.parseArgs({
       short: 'd',
       default: 'build/site',
     },
+    version: {
+      type: 'string',
+    },
   },
 })
+
+const buildVersion = cliArgs.version
+if (buildVersion != null && (buildVersion.length === 0 || buildVersion.includes('/'))) {
+  throw new Error(`Invalid --version value: ${buildVersion}`)
+}
 
 const publicDir = path.join(process.cwd(), 'build', 'public')
 const outputDir = path.join(process.cwd(), cliArgs.dir)
@@ -26,8 +34,9 @@ const SCRIPT_FILE_EXT = /\.(?:tsx?|jsx|mts)$/
 const SCRIPT_EXT_IN_PATH = /\.(?:tsx?|jsx|mts)(?=[?#]|$)/
 const SCRIPT_EXT_IN_JS_IMPORT = /\.(?:tsx?|jsx|mts)(?=["'?#])/g
 
-const versions = await getVersionsToBuild()
-console.log('Prerendering versions:\n', JSON.stringify(versions, null, 2))
+const versions = await getVersionsForPicker(buildVersion)
+console.log(`Prerendering ${buildVersion ? buildVersion : 'root'} docs`)
+console.log('Version picker options:\n', JSON.stringify(versions, null, 2))
 
 const docsRouter = createRouter(versions)
 
@@ -37,14 +46,8 @@ await fs.cp(publicDir, outputDir, { recursive: true })
 
 // Spider the site
 const paths = [
-  routes.home.href(),
-  routes.lookup.href(),
-  ...(versions
-    ?.filter((v) => v.crawl)
-    .flatMap((v) => [
-      routes.home.href({ version: v.version }),
-      routes.lookup.href({ version: v.version }),
-    ]) || []),
+  routes.home.href({ version: buildVersion }),
+  routes.lookup.href({ version: buildVersion }),
 ]
 
 for await (let { pathname, filepath, response } of crawl(docsRouter, { paths })) {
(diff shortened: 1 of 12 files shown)


docs/src/server/router.tsx
```diff
@@ -163,6 +163,15 @@ export function createRouter(versions: Versions) {
 
 // Response helpers
 
+function getLookupHref(href: string, version: string): string {
+  if (!href.startsWith('/api/')) return href
+
+  return routes.docs.href({
+    version,
+    slug: href.slice('/api/'.length).replace(/\/$/, ''),
+  })
```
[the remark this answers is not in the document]
mjackson: Fixed in 5d3935dfe. `getLookupHref` now sends `.md` lookup entries through `routes.markdown.href` and preserves query/hash suffixes, and there is a regression test covering versioned `/api.json` returning a markdown URL that resolves with 200.

Validated with:
- `pnpm test src/server/router.test.tsx` from `docs/`
- `pnpm run typecheck` from `docs/`

docs/src/server/router.tsx
```diff
@@ -163,6 +187,48 @@ export function createRouter(versions: Versions) {
...
+      if (!registry) {
+        registry = buildRegistry(docFiles, version)
+        registryByVersion.set(version, registry)
+      }
+      return registry
+    },
+  }
+}
+
+function getLookupHref(href: string, version: string): string {
```
brophdawg11: Is this the same as `getDocsRouteHref`?    May be able to share them if so

docs/src/server/markdown.ts
```diff
@@ -217,6 +217,13 @@ function getShikiExtension(
       code(code) {
         return code.text
       },
+      link(token) {
+        let href = getDocsRouteHref(token.href, version)
+        if (!href) return false
+
+        let title = token.title ? ` title="${escapeHtml(token.title)}"` : ''
+        return `<a href="${escapeHtml(href)}"${title}>${this.parser.parseInline(token.tokens)}</a>`
+      },
```
brophdawg11, line 226: ah good catch - this handles inline links in markdown that are outside of a code block?

docs/src/server/view.tsx
```diff
@@ -107,10 +108,10 @@ function Head(
         <link rel="icon" href="/favicon.svg" type="image/svg+xml" sizes="any" />
         {activeVersion != null ? (
           <>
-            <meta name="robots" content="noindex,nofollow" />
-            <meta name="googlebot" content="noindex,nofollow" />
+            <meta name="robots" content={shouldNofollow ? 'noindex,nofollow' : 'noindex'} />
+            <meta name="googlebot" content={shouldNofollow ? 'noindex,nofollow' : 'noindex'} />
```
brophdawg11: I don't think we want this change - these are the page-level tags and we want to blindly block any google crawling/indexing of pages inside `/{version}/...`.  We only want to index the latest docs at the root

docs/src/prerender/index.ts
```diff
@@ -139,8 +144,17 @@ async function getVersionsToBuild(): Promise<Versions> {
     .filter((tag) => semver.valid(tag) && !semver.prerelease(tag))
     .sort((a, b) => semver.rcompare(a, b))
 
-  // Crawl only the most recent tag
-  return remixVersions.length > 0
-    ? remixVersions.map((tag, i) => ({ version: tag, crawl: i === 0 }))
-    : getDefaultVersions()
+  if (activeVersion && !remixVersions.includes(activeVersion)) {
+    remixVersions.push(activeVersion)
```
brophdawg11: This feels odd - when would we be prerendering a `--version` that did not have a corresponding tag?  We may even want to error if we run into that scenario since it seems like that means we're not running on the right checkout of the repo?

docs/src/prerender/index.ts
```diff
@@ -37,14 +46,8 @@ await fs.cp(publicDir, outputDir, { recursive: true })
...
-  routes.home.href(),
-  routes.lookup.href(),
-  ...(versions
-    ?.filter((v) => v.crawl)
-    .flatMap((v) => [
-      routes.home.href({ version: v.version }),
-      routes.lookup.href({ version: v.version }),
-    ]) || []),
+  routes.home.href({ version: buildVersion }),
+  routes.lookup.href({ version: buildVersion }),
```
brophdawg11: yeah this is exactly what I was thinking we'd land at 👌

brophdawg11 (approved):
Looks good - just a few small callouts.

I also noticed another pre-existing bug - when we moved to asset server we lost the versioned asset paths.  I can look at fixing that separately, I don't think it should conflict with this work

docs/src/server/view.tsx
mjackson: Good call. i'll fix this.

mjackson: Agreed. I restored versioned pages to always render `noindex,nofollow` for both `robots` and `googlebot`. The prerender-only need is now handled by an internal crawler override that ignores page-level robots only on the versioned home page, so it can seed the sidebar docs graph without changing public crawler directives or following package/demo nofollow links.

docs/src/prerender/index.ts
mjackson: agree we should error in this case. will fix.

mjackson: Agreed. I changed this to fail fast when `--version` does not correspond to a fetched `remix@...` git tag, with an error that names the expected tag. I moved the version-picker tag parsing into a small tested helper so the behavior is covered: stable tags still populate the picker, an active prerelease is allowed when its tag exists, and a missing active tag now throws instead of silently adding the version.

docs/src/server/router.tsx
mjackson: Agreed, they were the same core route conversion with different fallback behavior. I moved the shared `/api/...` -> docs route logic into `getDocsRouteHref` in `routes.ts`, updated markdown to use it directly, and made `getVersionedLookupHref` a tiny wrapper that falls back to the original href for non-docs lookup values.

docs/src/server/markdown.ts
mjackson: yep, that's exactly what it does.

PR #11506, opened 2026-06-03
Reload client frames on ancestor frame reload

markdalgleish:
This fixes an issue raised in this PR: https://github.com/remix-run/remix/pull/11414. For context, this is the issue:

> **Issue:** Reloading the root frame causes nested frames to reload, but any frames rendered within client entries are not reloaded and continue to show the same server content. In this case, I'm not sure if this is a bug or expected behaviour, but I personally found it surprising.
>
> **Reproduction:** Go to http://localhost:44100/root-reload-client-entries, then click either of the reload buttons at the top of the page. You'll see the "Server props" content update, but the server times displayed in the nested frames do not change.

I originally wasn't sure if this behaviour was expected or not, but I've since confirmed that this was indeed an issue.

### DIFF
--- a/demos/frames/app/assets/entry.tsx
+++ b/demos/frames/app/assets/entry.tsx
@@ -10,7 +10,11 @@ const app = run({
     return exp
   },
   async resolveFrame(src, signal) {
-    let res = await fetch(src, { headers: { Accept: 'text/html' }, signal })
+    let res = await fetch(src, {
+      cache: 'no-store',
+      headers: { Accept: 'text/html' },
+      signal,
+    })
     if (!res.ok) {
       return `<pre>Frame error: ${res.status} ${res.statusText}</pre>`
     }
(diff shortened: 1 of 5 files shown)


demos/frames/app/assets/entry.tsx
```diff
@@ -10,7 +10,11 @@ const app = run({
     return exp
   },
   async resolveFrame(src, signal) {
-    let res = await fetch(src, { headers: { Accept: 'text/html' }, signal })
+    let res = await fetch(src, {
+      cache: 'no-store',
```
markdalgleish, line 14: Note, this change was needed so that the browser can run the two identical client frame requests in parallel, otherwise it staggered one frame request after the other after waiting to see if it could re-use the response.

More detail: https://groups.google.com/a/chromium.org/g/net-dev/c/bfmJbqSUODI/m/qU7K-Z2YAgAJ

PR #11508, opened 2026-06-03
Add fetch-router route installers

mjackson:
Add route installers to `fetch-router` so route groups can be written as local, reusable pieces of an app. A route installer receives a prefixed `RouteBuilder` that can register routes with `route()`, `map()`, method helpers, and nested `mount()`, while the parent router remains responsible for request dispatch, matching, router middleware, and default handling.

This also tightens the router type surface around everyday app setup: apps can derive context with `RouterContext<typeof router>`, inline middleware arrays now carry context contributions into actions/controllers, and reusable middleware chains can opt into `createMiddleware()` when they cross inference boundaries.

- Adds `router.mount(prefix, installer)` plus public `RouteBuilder`, `RouteInstaller`, and `RouterContext` types.
- Composes mount prefixes with child route patterns, including typed mount params and route-pattern duplicate-param behavior.
- Updates docs, skills, demos, and tests to use inline middleware context inference and route installers where they clarify app structure.
- Converts the social-auth demo auth subtree to a mounted route installer with local auth route definitions while keeping full app href generation stable.

```ts
import { createRouter, type RouteBuilder } from 'remix/router'

function installAdminRoutes<context extends AppContext>(router: RouteBuilder<context>) {
  router.get('/', () => new Response('Admin'))
  router.get('/users/:id', ({ params }) => new Response(params.id))
}

let router = createRouter<AppContext>({ middleware })

router.mount('/admin', installAdminRoutes)
router.mount('/internal/admin', installAdminRoutes)
```

```ts
let router = createRouter({
  middleware: [loadSession(), loadDatabase()],
})

type AppContext = RouterContext<typeof router>

let accountAction = createAction(routes.account, {
  middleware: [requireAuth()],
  handler({ auth }) {
    return Response.json({ id: auth.identity.id })
  },
})
```


mjackson:
Superseded by #11509.

mjackson:
Superseded by #11509.

PR #11509, opened 2026-06-03
Use route installers in social-auth demo

mjackson:
Use `router.mount()` in the social-auth demo so the auth feature owns its local route registration while the app router only chooses the `/auth` mount point.

- Adds shared local auth route definitions and derives full app auth routes from the same definitions for href generation.
- Moves auth subtree registration behind `installAuthRoutes(...)`, keeping the root router focused on top-level composition.
- Updates auth controllers to register against local `authRoutes` while preserving existing `/auth/...` form actions, redirects, and test URLs.

Supersedes #11508.


mjackson:
Closing this PR; we decided not to take this direction.

PR #11514, opened 2026-06-04
Pass named Frame targets during client resolution

### LINKED ISSUE #11500: Client-mounted `<Frame>` resolves without its `target`
When a named `<Frame>` is first resolved on the client (the fresh-insert path — e.g. a frame mounted after hydration), the runtime calls the user `resolveFrame` hook without the frame's name as the `target` argument. 

The reload path and the server path both forward it. Apps that distinguish a frame's content by target (e.g. forwarding it as an `X-Remix-Target` header so one URL can serve different frames) therefore get the wrong response for any client-resolved named frame, typically no-target/full-document.

Three of the four resolve paths pass the frame's name; the client fresh-insert is the only one that doesn't:

| Path | Call | Passes name as `target`? |
| --- | --- | --- |
| Server, in-stream | [server/stream.ts:485](https://github.com/remix-run/remix/blob/main/packages/ui/src/server/stream.ts#L485): `context.resolveFrame(props.src, props.name, ctx)` | ✅ |
| Client, reload | [runtime/frame.ts:205](https://github.com/remix-run/remix/blob/main/packages/ui/src/runtime/frame.ts#L205): `init.resolveFrame(frame.src, signal, frameName)` | ✅ |
| **Client, fresh insert** | [runtime/reconcile.ts:1100](https://github.com/remix-run/remix/blob/main/packages/ui/src/runtime/reconcile.ts#L1100): `runtime.resolveFrame(frameSrc, signal)` | ❌ |

## Reproduction

Minimal standalone repro: https://github.com/markmals/remix-frame-client-resolve-target

```sh
pnpm install && node --run dev   # http://localhost:41001
```

Open the console, click **Mount named frame**, then **Reload the frame**. The client `resolveFrame` hook logs `target = undefined` on the fresh mount and `target = probe` on the reload — same frame, same `src`, only the resolve path differs. The frame's endpoint, which branches on `X-Remix-Target`, shows the no-target branch after mount and the correct branch after reload.

I'm happy to open a PR with this fix plus a regression test once the team agrees it should be fixed.

### LINKED ISSUE #11502: Nested `<Frame>` inside a `clientEntry` re-resolves target-less on a non-root ancestor reload
Reloading a non-root ancestor frame (e.g. navigating an intermediate `detail` frame) re-resolves a nested non-blocking `<Frame>` that lives inside a `clientEntry` on the _client_, target-less; the server-streamed content for it is thrown away and the no-target response is shown instead. Plain nested frames (not wrapped in a `clientEntry`) are handled correctly now; only the `clientEntry`-wrapped case is affected.

## Relationship to #11474 / #11414

- #11474 fixed the plain-nested-frame portion — ancestor reloads now preserve / correctly re-resolve nested frames.
- #11414's description already flagged the clientEntry-on-reload interaction as surprising (for the _root_ reload case):
    > Reloading the root frame causes nested frames to reload, but any frames rendered within client entries are not reloaded and continue to show the same server content. … I'm not sure if this is a bug or expected behaviour, but I personally found it surprising.

This is the **non-root-ancestor variant** of that same `clientEntry` interaction, except here the `clientEntry`-wrapped frame isn't left stale, it actively re-resolves with the wrong (no-target) content.

## Reproduction

https://github.com/markmals/remix-frame-ancestor-reload-target

```sh
pnpm install && node --run dev   # http://localhost:41003
```

A blocking `detail` frame (the intermediate ancestor) holds three nested non-blocking comments frames: plain, plain + `key`, and one wrapped in a `clientEntry`. Open `.../?issue=a` with the console, then click **reload detail to issue B**:

- plain + plain-keyed → re-resolved server-side **with** their target, adopted by the client ✅
- `clientEntry`-wrapped → client `resolveFrame` fires with `target = (none)`; the streamed content is discarded and the no-target response is shown ❌

The differentiator is the `clientEntry` boundary, not the frame `key` (the keyed _plain_ frame is fine).

## Expected vs actual

- **Expected:** the ancestor reload preserves / re-resolves the nested `clientEntry` frame **with its target**, like the plain frames (the server already streams it correctly).
- **Actual:** it's torn down and re-resolved on the client target-less, discarding the streamed content.

I'm happy to open a PR once the team agrees this should be fixed.

markmals:
Client-created or remounted named `<Frame>` elements resolve through `resolveClientFrame`, but that path called `resolveFrame(src, signal)` without the frame name. Server rendering and explicit frame reloads already pass the name as the target, so target-aware resolvers could return different content for client-resolved frames.

This updates the client resolve path to pass `getFrameName(node)` as `target`. Unnamed frames continue to pass `undefined`.

Fixes #11500.
Fixes #11502.

Notes:
- A hydrated `clientEntry` owns its subtree after initial hydration. On later reloads, including ancestor frame reloads, frames rendered from that client entry resolve through the client runtime.
- Regression tests cover a client-mounted named Frame and a keyed Frame inside a `clientEntry` after a non-root ancestor reload.

### DIFF
--- a/packages/ui/.changes/patch.client-frame-resolve-target.md
+++ b/packages/ui/.changes/patch.client-frame-resolve-target.md
@@ -1,0 +1,3 @@
+Forward the frame's name as the resolve target when a named `<Frame>` is resolved on the client
+
+Only the reload and server resolve paths passed the frame's name; the client resolve path — a fresh client mount, or a `clientEntry`-wrapped frame remounted when a non-root ancestor reloads — called `resolveFrame` without it. Frames that branch on the target (for example via an `X-Remix-Target` header) now receive the correct content instead of the no-target response.
--- a/packages/ui/src/runtime/reconcile.ts
+++ b/packages/ui/src/runtime/reconcile.ts
@@ -1097,7 +1097,7 @@ function resolveClientFrame(node: VNode, runtime: FrameRuntime): void {
   let resolveController = new AbortController()
   node._frameResolveController = resolveController
 
-  Promise.resolve(runtime.resolveFrame(frameSrc, resolveController.signal))
+  Promise.resolve(runtime.resolveFrame(frameSrc, resolveController.signal, getFrameName(node)))
     .then(async (content) => {
       if (node._frameResolveToken !== token || resolveController.signal.aborted) return
       node._frameFallbackRoot?.dispose()
--- a/packages/ui/src/test/frame.test.tsx
+++ b/packages/ui/src/test/frame.test.tsx
@@ -3013,6 +3013,72 @@ describe('run', () => {
     app.dispose()
   })
 
+  it('forwards the frame name as the resolve target when mounting a named Frame on the client', async () => {
+    // Regression test for #11500: the client fresh-insert path (resolveClientFrame)
+    // must forward the frame's name as the third `target` argument, like the reload
+    // path (frame.ts) and the server path (server/stream.ts) already do.
+    let mounted = false
+    let showFrame: undefined | (() => void)
+
+    let NamedFrameMount = clientEntry(
+      '/js/named-frame-mount.js#NamedFrameMount',
+      function NamedFrameMount(handle: Handle) {
+        showFrame = () => {
+          mounted = true
+          handle.update()
+        }
+
+        return () => (
+          <section>
+            {mounted ? (
+              <Frame
+                name="probe"
+                src="/named-frame"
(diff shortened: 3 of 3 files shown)


ryanflorence:
Merged. Thanks for the repros and tests!

For posterity, the intended ownership model here is: after initial hydration, a `clientEntry` owns its subtree. Frames rendered inside that subtree resolve through the client runtime on later reloads, including ancestor frame reloads.

So this fixes the bug in #11500 and #11502: client-resolved named Frames now pass their frame name as the `target`.

PR #11516, opened 2026-06-04
Preserve Frame markers at clientEntry Fragment boundaries

### LINKED ISSUE #11501: Fragment-nested `<Frame>` doesn't adopt its hydration marker
A `<Frame>` that is the first child of a bare `<Fragment>` (`<>`) returned by a `clientEntry` fails to adopt its server-rendered frame markers during hydration. It takes the fresh-insert path instead, duplicating the streamed subtree and re-fetching `src` on the client. Wrapping the same `<Frame>` in a host element (e.g. `<div>`) makes it hydrate cleanly.

During hydration the comment-skip strategy is chosen by node type ([reconcile.ts:704](https://github.com/remix-run/remix/blob/main/packages/ui/src/runtime/reconcile.ts#L704)):

```ts
cursor =
    node.type === Frame
        ? skipCommentsExceptFrameStart(cursor ?? null) // stops at the rmx:f marker
        : skipComments(cursor ?? null); // walks past ALL comments
```

A `clientEntry` boundary emits its entry marker immediately followed by the child `<Frame>`'s `<!-- rmx:f ... -->` start marker. The skip runs on the clientEntry component node (a non-`Frame` node), so it uses plain `skipComments` ([client-entries.ts:128](https://github.com/remix-run/remix/blob/main/packages/ui/src/runtime/client-entries.ts#L128): `while COMMENT_NODE: nextSibling`), which walks past _both_ markers, swallowing the frame-start one before the fragment's first-child `<Frame>` is reached. `insertFrame` then finds no marker and takes the fresh-insert branch.

A host-`<div>` wrapper avoids it: the `<div>` is a non-comment node the cursor lands on, and the `<Frame>` inside is then reached with the frame-aware skip.

## Reproduction

https://github.com/markmals/remix-frame-fragment-marker-skip

```sh
pnpm install && node --run dev   # http://localhost:41002
```

Load the page with the console open. Three non-blocking frames, all server-resolved with their target:

- **Fragment-nested (clientEntry)** → a client `resolveFrame` fires; the streamed content stays _and_ a second copy is re-fetched and inserted → the frame is duplicated.
- **Host-`<div>`-wrapped (clientEntry)** → hydrates clean, no client fetch.
- **Plain server-rendered fragment (no clientEntry)** → also clean — so it's specific to the clientEntry boundary.

## Expected vs actual

- **Expected:** the Fragment-nested frame adopts its streamed markers/content; no client re-fetch.
- **Actual:** the marker is skipped → fresh insert → subtree duplicated + `src` re-fetched on the client. (That re-fetch is also target-less, which is a [separate `resolveClientFrame` issue](https://github.com/remix-run/remix/issues/11500).)

I'm happy to open a PR once the team agrees this should be fixed.

markmals:
A `<Frame>` that is the first child of a bare Fragment returned by a `clientEntry` could miss its server-rendered `rmx:f` marker during initial hydration. The insert path skipped all comments while processing the client entry node, which consumed the frame-start marker before the Frame could adopt it. The Frame then took the fresh-insert path, duplicated the streamed subtree, and called client `resolveFrame`.

This preserves frame-start markers while skipping comments for hydration, so the following Frame can claim its initial marker. Host-wrapped Frames already behaved correctly.

Fixes #11501.

Notes:
- This only affects initial hydration marker adoption. After a `clientEntry` is hydrated, it owns its subtree and future Frame resolutions inside it happen from the client runtime.
- Adds a regression test for a Fragment-returning `clientEntry` whose first child is a Frame.

### DIFF
--- a/packages/ui/.changes/patch.fragment-nested-frame-hydration-marker.md
+++ b/packages/ui/.changes/patch.fragment-nested-frame-hydration-marker.md
@@ -1,0 +1,3 @@
+Adopt a Fragment-nested `<Frame>`'s server-rendered hydration marker at `clientEntry` boundaries
+
+A `<Frame>` that is the first child of a bare Fragment returned by a `clientEntry` now adopts its streamed hydration marker instead of taking the fresh-insert path, which previously re-fetched `src` on the client and duplicated the streamed subtree. A `<Frame>` wrapped in a host element already hydrated cleanly.
--- a/packages/ui/src/runtime/reconcile.ts
+++ b/packages/ui/src/runtime/reconcile.ts
@@ -700,10 +700,12 @@ function insert(
     cursor = null
   }
 
-  cursor =
-    node.type === Frame
-      ? skipCommentsExceptFrameStart(cursor ?? null)
-      : skipComments(cursor ?? null)
+  // Preserve frame-start markers for non-Frame nodes too, so a following <Frame>
+  // (e.g. the first child of a bare Fragment at a clientEntry boundary) can still
+  // claim its rmx:f marker during hydration instead of being re-inserted fresh.
+  // A rmx:f marker always belongs to a <Frame>, so no non-Frame node should
+  // consume one.
+  cursor = skipCommentsExceptFrameStart(cursor ?? null)
 
   // Also check after skipComments in case we skipped past the anchor
   if (cursor && anchor && cursor === anchor) {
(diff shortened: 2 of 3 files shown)


ryanflorence:
This is separate from #11514 and fixes the initial hydration case: a Frame that already has a server-rendered marker should adopt it, even when it is the first child of a Fragment returned by a `clientEntry`.

After hydration, the `clientEntry` owns its subtree and future Frame resolutions inside it happen through the client runtime (since their `src` may depend on client state we can't assume the server's content is correct when it doesn't have access to client state)

PR #11517, opened 2026-06-04
Migrate Prettier to Oxfmt

markdalgleish:
This swaps out Prettier for [Oxfmt](https://oxc.rs/docs/guide/usage/formatter.html). On my machine `pnpm format` goes from ~8s to ~1s.

Files have also been formatted so there are a few slight differences.

I've also updated the checked in VS Code settings so that Oxc is now a recommended plugin and set as the default formatter for supported files.


.vscode/settings.json
```diff
@@ -1,4 +1,43 @@
...
+  },
+  "[typescriptreact]": {
+    "editor.defaultFormatter": "oxc.oxc-vscode"
+  },
+  "[toml]": {
+    "editor.defaultFormatter": "oxc.oxc-vscode"
+  },
+  "[yaml]": {
+    "editor.defaultFormatter": "oxc.oxc-vscode"
+  },
```
brookslybrand, line 40: Probably a bit too much settings bloat

.github/workflows/format.yml
brookslybrand, line 51: Probably outside the scope of this PR, but this workflow is a little wild to me. Why aren't we just running a `pnpm format` command or something?

docs/package.json
```diff
@@ -5,7 +5,7 @@
   "dependencies": {
     "front-matter": "^4.0.2",
     "marked": "^17.0.1",
-    "prettier": "^3.5.3",
+    "oxfmt": "^0.51.0",
```
brookslybrand: Should we leverage `workspace:*` here?

brookslybrand (approved):
I'm definitely in favor of this!

I left a few comments on some small things to improve/that got carried in, but all in all I say go for it!

.vscode/settings.json
markdalgleish: Yeah, I wish there was a cleaner way to define this, but this list is basically everything in the pre-existing format workflow list, plus `toml` since the repo also has some `.toml` files.

.github/workflows/format.yml
markdalgleish: This is actually just an early bailout so that we don't bother installing deps and formatting if nothing relevant has changed. A full `pnpm format` is run later in the workflow.

docs/package.json
markdalgleish: I assume you mean `catalog:`? Nice catch, that makes sense here.

PR #11518, opened 2026-06-04
Version asset paths for docs prerendering

brophdawg11:
Adds proper versioning to asset paths so prerendering downloads full snapshots of all assets into a versioned folder


mjackson:
Let's not merge this yet. I'm not comfortable with the multiple asset server approach. A single asset server should be capable enough to serve all the assets for a single app.

I'm going to look into the asset server a bit more and see if we can make it a little more capable.

mjackson (changes requested):
Not ready to merge yet. We shouldn't have to spin up multiple asset servers for a single app.

brophdawg11:
yeah I agree - I think the issue is that we cant proxy optional versions through basePath - let me dig a bit deeper there

docs/src/server/asset-server.ts
```diff
@@ -3,11 +3,10 @@ import { createAssetServer } from 'remix/assets'
 
 const DOCS_DIR = path.resolve(import.meta.dirname, '..', '..')
 const REPO_DIR = path.resolve(DOCS_DIR, '..')
-const ENTRY_PATH = path.join(DOCS_DIR, 'src', 'client', 'entry.tsx')
 
 export const assetServer = createAssetServer({
   rootDir: REPO_DIR,
-  basePath: '/assets',
+  basePath: '/(:version/)assets',
```
brophdawg11: @mjackson I think this is the crux of what we need to be able to support in the asset server, but the changes under the hood there got a bit bigger than I anticipated.  Probably worth reviewing with you and @markdalgleish  who better understand the architecture there.

brophdawg11:
I had codex take a stab at changes to the asset server to allow an optional version in the asset base path - going to want to get @markdalgleish's eyes on it though

docs/src/prerender/index.ts
```diff
@@ -36,7 +36,8 @@ const versions = getVersionsForPicker(buildVersion, getDefaultVersions())
 console.log(`Prerendering ${buildVersion ? buildVersion : 'root'} docs`)
 console.log('Version picker options:\n', JSON.stringify(versions, null, 2))
 
-const docsRouter = createRouter(versions)
+const assetServer = createAssetServer(buildVersion)
+const router = createRouter({ assetServer, versions })
```
brophdawg11, line 40: Instead of trying to enhance `assetServer` to be able to support varying basepaths, we can just latch onto the new prerender `--version` flag and create a single version-aware asset server with a base path of  either `/assets` or `/{version}/assets` based on the version we're prerendering.

brophdawg11:
@mjackson I revisited this with a new simpler approach today that doesn't require changes to the asset server.  We can just leverage the new `prerender --version` flag and create a version-aware asset server

mjackson (approved):
Looks good, thanks for following up here @brophdawg11

PR #11519, opened 2026-06-04
Use relative URLs for source-served script imports

brookslybrand:
This splits the asset-server URL shortening piece out of #11507. Source-served script modules currently repeat absolute `/assets/...` prefixes in every rewritten browser import; this keeps the public URLs absolute while shortening same-origin specifiers inside the emitted JavaScript.

- Rewrites served script imports, re-exports, and static dynamic imports to relative served URLs such as `./dep.@abc123.ts`
- Rewrites external script source map comments to relative `.map` URLs
- Keeps `assetServer.getHref()` and `assetServer.getPreloads()` returning absolute URLs for callers and preload tags
- Leaves data, HTTP(S), and configured external imports unchanged

Measured against the Remix UI browser asset graph from #11507 by comparing relative output with equivalent absolute-import output:

| Asset set | Modules | Raw | Gzip | Brotli |
| --- | ---: | ---: | ---: | ---: |
| Absolute import URLs | 60 | 95,449 | 39,840 | 35,175 |
| Relative import URLs | 60 | 92,830 | 38,987 | 34,516 |
| Savings | — | 2,619 | 853 | 659 |

The tradeoff is that relative imports assume served modules are fetched from their canonical asset-server URL layout. Public entry/preload URLs remain absolute, so normal asset-server usage is unchanged; custom proxies that rewrite individual module paths should preserve the path layout so relative dependency resolution still targets the same files.

## Impact on Bookstore Demo

Running `pnpm --filter bookstore-demo run start` and going to the login page (left is `main`, right is this PR):

image

### DIFF
--- a/packages/assets/.changes/patch.relative-script-import-urls.md
+++ b/packages/assets/.changes/patch.relative-script-import-urls.md
@@ -1,0 +1,1 @@
+Reduced source-served browser JavaScript by rewriting script import specifiers and external source map comments to relative served URLs while keeping `getHref()` and `getPreloads()` URLs absolute.
--- a/packages/assets/src/lib/asset-server.test.ts
+++ b/packages/assets/src/lib/asset-server.test.ts
@@ -159,13 +159,25 @@ async function getCompiledCodeAndSourceMap(
   }
 }
 
-async function getAbsoluteImportSpecifiers(source: string): Promise<string[]> {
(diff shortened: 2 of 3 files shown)


brookslybrand:
@markdalgleish I have no idea if this is actually a good idea. There may be some edge cases/subtle bugs that I'm unaware of that this change would cause.

If that's the case, please feel free to just close this PR.

PR #11523, opened 2026-06-05
Treat lint warnings as errors

markdalgleish:
In main we've currently accumulated 70 lint warnings.

While working on a branch, I ran `pnpm lint --fix` to clean up my own work but it ended up introducing a lot of unrelated changes fixing these warnings. To me, this is a good example of why we shouldn't have lint warnings at all. Either it's an error, or we should be able to ignore it, otherwise the warnings just build up over time.

This PR makes the following changes to address this:
- Sets `--max-warnings=0` when running lint commands
- Explicitly sets `remix-headers/canonical-header-name` to `error`
- Moves the `jsdoc/require-yields` config into the relevant section where the `jsdoc` plugin is configured - using it as a top-level rule doesn't work
- Fixes all warnings that are now errors, except for `no-control-regex` which I opted to disable
- Removes usage of the `-A` flag when running the `oxlint` command since it was ignoring our top-level `rules` config - the equivalent now is to set all `categories` to `"off"` in our Oxlint config
- Removes `--quiet` from the `pnpm lint` commands so we can see more detail about lint failures


mjackson (approved):
This LGTM, I agree that lint warnings should be treated as errors. Nice catch, @markdalgleish

PR #11524, opened 2026-06-05
Fix `clientEntry` collision for entries sharing an entry id

### LINKED ISSUE #11503: Two `clientEntries` in the same file collide when resolved by `import.meta.url`
Two different `clientEntry(import.meta.url, namedFn)` exports declared in the same file and rendered in the same SSR pass both hydrate as the first one. The server renders each correctly, but on hydration the second's slot adopts the first component causing a hydration mismatch and the wrong interactive component on the client.

`resolveClientEntries` memoizes resolution in a `Map` keyed by `entryId` only ([packages/ui/src/server/stream.ts:1055](https://github.com/remix-run/remix/blob/main/packages/ui/src/server/stream.ts#L1055)):

```ts
let resolvedEntries = new Map<string, ResolvedClientEntry>();
for (let [hydrationId, { entryId, component, props }] of context.unresolvedHydrationData) {
    let resolvedEntry = resolvedEntries.get(entryId);
    if (!resolvedEntry) {
        resolvedEntry = await resolveClientEntry(entryId, component); // only the FIRST component
        resolvedEntries.set(entryId, resolvedEntry);
    }
    context.hydrationData.set(hydrationId, { exportName: resolvedEntry.exportName /* … */ });
}
```

A bare `import.meta.url` makes `entryId` the file, so two `clientEntries` in one file share it: the hook runs once (with the first component) and the second instance reuses that result; the second's `component.name` is never looked at. (There's a test pinning this: _'it calls resolveClientEntry only once per unique entry id during SSR'_, [stream.test.tsx:1629](https://github.com/remix-run/remix/blob/main/packages/ui/src/test/stream.test.tsx#L1629).)

## Reproduction

https://github.com/markmals/remix-cliententry-same-file-collision

```sh
pnpm install && node --run dev   # http://localhost:41004
```

`app/widgets.tsx` exports two `clientEntries`, `Alpha` and `Beta`, each `clientEntry(import.meta.url, ...)`; the home page renders both. Load it with the console open: the server renders "I am Alpha" and "I am Beta" correctly, but after hydration the **Beta** box reads "I am Alpha" and the console logs a hydration mismatch. The streamed `rmx-data` shows both entries resolved to `exportName: "Alpha"`.

## Expected vs actual

- **Expected:** `Alpha` hydrates as `Alpha`, `Beta` as `Beta` (each resolved by its `component.name`).
- **Actual:** both hydrate as `Alpha`.

I'm happy to open a PR once the team agrees on the approach.

vismayIO:
## Problem

Fixes #11503

Two `clientEntry(import.meta.url, namedFn)` exports declared in the same file and rendered in the same SSR pass both hydrate as the first one. The server renders each correctly, but on hydration the second's slot adopts the first component, causing a hydration mismatch and the wrong interactive component on the client.

`resolveClientEntries` memoized resolution in a `Map` keyed by `entryId` only (`packages/ui/src/server/stream.ts`). A bare `import.meta.url` makes `entryId` the file, so two `clientEntries` in one file share it: the hook runs once (with the first component) and the second instance reuses that result; the second's `component.name` is never looked at. Both entries stream `exportName: "Alpha"`.

## Fix

Key the resolution cache by the **component reference** (the unique identity) instead of the `entryId` string:

- Same component rendered N times → same reference → still one resolution (dedup preserved).
- Two distinct components sharing one `entryId` → different references → resolved independently with their own `exportName`.

Client-side hydration was already correct (keyed per unique instance id); the corruption was purely in SSR resolution.

## Tests

- Added: `resolves two clientEntries sharing one entry id to distinct exports during SSR` — fails before the fix (both resolve to `Alpha`), passes after.
- Existing pinning test `calls resolveClientEntry only once per unique entry id during SSR` still passes (same component rendered 3× → one resolve call).
- Full `@remix-run/ui` suite: 792 pass, 0 fail.


vismayIO:
@markdalgleish and @ryanflorence, Could you please share any updates on this PR? Also, let me know if there’s anything else I can do to help resolve the issue.

markdalgleish (approved):
Thanks for the PR! I just pushed an update to add a change file.

vismayIO:
@ryanflorence, I'd greatly appreciate any suggestions, feedback, or concerns you might have regarding the implementation.

PR #11536, opened 2026-06-09
fix CI flows on pnpm@11

brophdawg11:
`postinstall` scripts are automatically skipped in `pnpm@11` and need to be explicitly approved or else `pnpm install` fails.  It also seems to then cause any other `pnpm run *` command to fail with the same error until you run the builds.

```sh
~/me/repos/remix (brophdawg11/codex-rmx-ignore-ui)> pnpm install
Scope: all 96 workspace projects
...
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: esbuild@0.25.12, esbuild@0.27.1, esbuild@0.27.3, esbuild@0.27.7, sharp@0.34.5, workerd@1.20260508.1

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
```

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -75,7 +75,7 @@ jobs:
         run: |
           pnpm run setup-installable-branch preview/main
           git push --force --set-upstream origin preview/main
-          echo "💿 pushed installable branch: https://github.com/$GITHUB_REPOSITORY/commit/$(git rev-parse HEAD)"
+          echo "pushed installable branch: https://github.com/$GITHUB_REPOSITORY/commit/$(git rev-parse HEAD)"
 
       # Build and force push over the PR preview/pr-{number} branch + comment on the PR
       - name: Build/push branch (pull_request)
@@ -86,7 +86,7 @@ jobs:
           pnpm run setup-installable-branch preview/pr-${{ github.event.pull_request.number }}
           git push --force --set-upstream origin preview/pr-${{ github.event.pull_request.number }}
           echo "pushed installable branch: https://github.com/$GITHUB_REPOSITORY/commit/$(git rev-parse HEAD)"
-          pnpm run pr-preview comment ${{ github.event.pull_request.number }} preview/pr-${{ github.event.pull_request.number }}
+          node ./scripts/pr-preview.ts comment ${{ github.event.pull_request.number }} preview/pr-${{ github.event.pull_request.number }}
 
       # Build and normal push for experimental releases to avoid unintended force
       # pushes over remote branches in case of a branch name collision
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -43,13 +43,5 @@
...
     "packages/*"
-  ],
-  "pnpm": {
-    "onlyBuiltDependencies": [
-      "@swc/core",
-      "esbuild",
-      "sharp",
-      "workerd"
-    ]
-  }
```
brophdawg11, line 54: This is no longer used in pnpm 11

```
[WARN] The "pnpm" field in package.json is no longer read by pnpm. The following keys were ignored: "pnpm.onlyBuiltDependencies". See https://pnpm.io/settings for the new home of each setting.
```

packages/node-tsx/src/index.test.ts
```diff
@@ -1390,7 +1390,12 @@ async function ensureBuiltNodeTsxPackage(): Promise<void> {
     ROOT_DIR,
   )
   assert.equal(result.exitCode, 0, result.stderr)
-  assert.equal(result.stderr, '')
+  assert.equal(
+    result.stderr,
+    process.env.CI
+      ? '$ tsgo -p tsconfig.build.json\n'
+      : '\x1B[2m$ tsgo -p tsconfig.build.json\x1B[22m\n',
+  )
```
brophdawg11, line 1398: looks like pnpm 11 might start printing this type of output from `pnpm run build` through `stderr` instead of `stdout`

.github/workflows/preview.yml
```diff
@@ -86,7 +86,7 @@ jobs:
           pnpm run setup-installable-branch preview/pr-${{ github.event.pull_request.number }}
           git push --force --set-upstream origin preview/pr-${{ github.event.pull_request.number }}
           echo "pushed installable branch: https://github.com/$GITHUB_REPOSITORY/commit/$(git rev-parse HEAD)"
-          pnpm run pr-preview comment ${{ github.event.pull_request.number }} preview/pr-${{ github.event.pull_request.number }}
+          node ./scripts/pr-preview.ts comment ${{ github.event.pull_request.number }} preview/pr-${{ github.event.pull_request.number }}
```
brophdawg11, line 89: pnpm 11 seems to be stricter on checking `package.json` versus `pnpm-lock.yaml` versions during `pnpm run *` commands.  In these preview branches we write new preview versions directly into `package.json` so it's expected to be out of sync with the lockfile.  We can just call the node script directly and avoid going through any `pnpm run` validations.

```
pushed installable branch: https://github.com/remix-run/remix/commit/48a1b465315e616bec2ccf21e3c9f98a2f19be30
Scope: all 96 workspace projects
[ERR_PNPM_OUTDATED_LOCKFILE] Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/packages/remix/package.json

Note that in CI environments this setting is true by default. If you still need to run install in such cases, use "pnpm install --no-frozen-lockfile"

  Failure reason:
  specifiers in the lockfile don't match specifiers in package.json:
* 44 dependencies are mismatched:
  - @remix-run/assert (lockfile: workspace:^, manifest: remix-run/remix#preview/pr-11536&path:packages/assert)
  - ```

jacob-ebey (approved):

pcattori (approved):
Ran into the same errors and made the same exact fixes locally, so can attest that this works!

Issue #11534, opened 2026-06-09
node-fetch-server can try to send an error response after headers are committed


mjackson:
### Summary

`@remix-run/node-fetch-server` catches errors from `sendResponse()` and attempts to send an error response. However, `sendResponse()` writes status and headers before streaming the response body. If the body stream throws after headers or body chunks have already been sent, the catch path can attempt to write a second response to the same socket.

### Relevant code

- `packages/node-fetch-server/src/lib/request-listener.ts` `sendResponseForRequest()` catches `sendResponse()` failures and calls `sendErrorResponseForRequest()`.
- `sendErrorResponseForRequest()` only checks whether the response is closed or the request is aborted before calling `sendResponse()` with the error response.
- `sendResponse()` calls `res.writeHead(...)` before reading and writing the response body stream.

That means a body stream error after `writeHead()` but before close can take the error path even though headers are already committed.

### Why this is dangerous

Once headers are committed, the server cannot replace the response with a `500`. Trying to do so can produce `ERR_HTTP_HEADERS_SENT`, reset the connection, create confusing partial responses, or cause the original stream error to be masked by a second write failure.

This is especially risky for streaming responses where errors can happen after the first chunk has been flushed.

### Reproduction sketch

Create a `Response` whose body enqueues one chunk and then errors:

```ts
new Response(
  new ReadableStream({
    start(controller) {
      controller.enqueue(new TextEncoder().encode('partial'))
      controller.error(new Error('boom after commit'))
    },
  }),
)
```

Serve it through the node fetch server. The first chunk causes headers/body to be committed; the later stream error is caught by `sendResponseForRequest()`, which can then try to send an error response on the already-started response.

A regression test should assert that no second `writeHead()` is attempted after headers have been sent, and that the server handles the stream failure by destroying/ending the response in a controlled way.

### Suggested fix

Track whether the response has started before attempting error fallback. Possible approaches:

- In `sendResponse()`, consider the response committed immediately after `writeHead()` or after the first body write.
- In `sendResponseForRequest()`, only call `sendErrorResponseForRequest()` if no headers have been sent yet. Node HTTP/1 exposes `res.headersSent`; HTTP/2 has analogous state that may need a helper.
- If an error happens after commit, do not attempt to synthesize a `500`; instead cancel the body, destroy or close the response as appropriate, and report the error via `onError`/logging without writing new headers.
- Add tests for errors before commit versus errors after commit so the pre-commit case can still produce a normal error response.

[...]

PR #11539, opened 2026-06-10
Reject node-fetch-server body reads after client aborts

### LINKED ISSUE #11533: node-fetch-server request bodies can hang after client aborts
### Summary

`@remix-run/node-fetch-server` creates a web `Request` body from a Node request stream, but the `ReadableStream` only listens for `data` and `end`. It does not propagate `error`, `aborted`, or early close from the incoming request into the web stream.

As a result, code that consumes the request body can hang indefinitely when a client disconnects mid-upload.

### Relevant code

In `packages/node-fetch-server/src/lib/request-listener.ts`, `createRequest()` creates `init.body` like this for non-GET/HEAD requests:

```ts
init.body = new ReadableStream({
  start(controller) {
    req.on('data', (chunk) => {
      controller.enqueue(new Uint8Array(chunk.buffer, chunk.byteOffset, chunk.byteLength))
    })
    req.on('end', () => {
      controller.close()
    })
  },
})
```

The stream never calls `controller.error()` when the Node request errors or is aborted. The request `AbortController` is currently tied to the response `close`/`finish` lifecycle, not to request body failure.

### Why this is dangerous

Handlers and middleware commonly call `request.formData()`, `request.text()`, `request.json()`, or otherwise read the body. If the client disconnects before `end`, the web stream may never close or error, so the handler can remain pending.

Enough aborted uploads could tie up request work and become a denial-of-service vector. It also makes ordinary client disconnects difficult to handle correctly because the framework does not surface the abort to body consumers.

### Reproduction sketch

1. Start a server using `createRequest()` / the node fetch server listener.
2. In a POST handler, await `request.formData()` or `request.text()`.
3. Send a request with a body and close the client socket before the body finishes.
4. Observe that the body read can remain pending instead of rejecting or aborting.

A regression test can use a synthetic `IncomingMessage`-like stream or a real HTTP client socket, emit partial `data`, then emit `aborted`/`error`/`close`, and assert that reading the web `Request` body rejects and the request signal is aborted.

### Suggested fix

Bridge Node request failure events into both the web body stream and the request signal. For example:

- Listen for `req.on('error', ...)` and call `controller.error(error)`.
- Listen for `req.on('aborted', ...)` and error the stream with an abort reason.
- Consider `close` handling for requests that close before `end`, while avoiding treating a normal post-`end` close as an error.
- Abort the `Request` signal when the incoming request body aborts, not only when the response closes before finish.
- Remove listeners during stream cancellation/close to avoid leaks.

The expected behavior should be that body consumers reject promptly when the client disconnects before the body is fully read.

morgan-coded:
What bug this fixes:
`node-fetch-server` could leave request body reads hanging if the client aborted while uploading. That could affect direct stream reads as well as lazy body helpers like `text()`, `json()`, and `bytes()`.

What changed:
Request body errors, aborts, and early closes now reject or abort the pending body read instead of leaving it unresolved.

How I tested it:
I added regression coverage for stream body reads and the lazy `text()`, `json()`, and `bytes()` paths. The `node-fetch-server` tests pass with 28/28, and typecheck, prettier, oxlint, and diff-check all pass.

Issue link:
Closes #11533

### DIFF
--- a/packages/node-fetch-server/.changes/patch.fix-node-fetch-server-body-abort.md
+++ b/packages/node-fetch-server/.changes/patch.fix-node-fetch-server-body-abort.md
@@ -1,0 +1,1 @@
+Handle aborted request bodies in `node-fetch-server` so stream and lazy body reads reject instead of staying pending after an upload is cut short.
--- a/packages/node-fetch-server/src/lib/client-disconnect-error.ts
+++ b/packages/node-fetch-server/src/lib/client-disconnect-error.ts
@@ -1,0 +1,18 @@
+/**
+ * Error used to settle pending request body reads when the client goes away
+ * (an `error`, legacy `aborted`, or close-before-`end` on the incoming request)
+ * before the body has been fully received.
+ */
+export class ClientDisconnectError extends Error {
+  constructor(cause?: unknown) {
+    super(
+      'Client disconnected before the request body was fully received',
+      cause === undefined ? undefined : { cause },
+    )
+    this.name = 'ClientDisconnectError'
+  }
+}
+
+export function isClientDisconnectError(error: unknown): error is ClientDisconnectError {
+  return error instanceof ClientDisconnectError
+}
--- a/packages/node-fetch-server/src/lib/lazy-request.ts
+++ b/packages/node-fetch-server/src/lib/lazy-request.ts
@@ -328,6 +328,8 @@ function readRequestBody(req: IncomingRequest): Promise<Buffer> {
       req.off('data', onData)
       req.off('end', onEnd)
       req.off('error', onError)
+      req.off('aborted', onClose)
+      req.off('close', onClose)
     }
 
     function onData(buffer: Buffer) {
@@ -357,9 +359,29 @@ function readRequestBody(req: IncomingRequest): Promise<Buffer> {
       reject(error)
     }
 
+    // A 'close' (or legacy 'aborted') before 'end' means the client went away
+    // before the body was fully received; a close after 'end' never reaches
+    // this handler because 'end' removes these listeners.
+    function onClose() {
+      cleanup()
+      reject(new Error('Client disconnected before the request body was fully received'))
+    }
+
+    if (req.readableEnded) {
+      resolve(Buffer.alloc(0))
+      return
+    }
+
+    if (req.destroyed) {
+      reject(new Error('Client disconnected before the request body was fully received'))
+      return
+    }
+
     req.on('data', onData)
     req.once('end', onEnd)
     req.once('error', onError)
+    req.once('aborted', onClose)
+    req.once('close', onClose)
   })
 }
 
@@ -373,6 +395,8 @@ function readRequestText(req: IncomingRequest): Promise<string> {
       req.off('data', onData)
       req.off('end', onEnd)
       req.off('error', onError)
+      req.off('aborted', onClose)
+      req.off('close', onClose)
     }
 
     function onData(buffer: Buffer) {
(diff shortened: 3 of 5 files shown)


daltino (commented):
Nice work on this fix! Handling client disconnects during body reads is a common source of hard-to-debug hangs, and it's great to see it addressed properly. The approach of a dedicated `ClientDisconnectError` class is clean and idiomatic. Here are some observations and suggestions:

---

## Positive notes

- **`ClientDisconnectError` as a dedicated class** is the right call. It gives callers a clean way to distinguish "client left" from other I/O errors, and the `cause` forwarding preserves the original error for debugging.
- **Cleaning up all listeners in the cleanup function** (including the new `aborted` and `close` listeners) looks correct and avoids leaks.
- **Adding regression tests** via a real `net.Socket`-level abort is the most reliable way to test this scenario. Appreciated.

---

## Suggestions & concerns

### 1. `aborted` event is deprecated — be explicit about Node.js version behavior

The `aborted` event on `IncomingMessage` was deprecated in Node.js 17.0.0 (and removed in some later versions). You're registering it alongside `close`, which is the modern replacement. That's the right strategy, but it's worth a comment in the code explaining *why* both are registered:

```ts
// `aborted` is fired on Node.js < 17 when the connection is destroyed before
// the request completes. On newer Node.js versions, `close` covers this case.
req.on('aborted', onClose)
req.on('close', onClose)
```

Without this comment, a future reader might wonder if `aborted` is a bug or a typo.

---

### 2. `close` fires after `end` on a normal request — guard against false positives

`close` is emitted on every request after it finishes, including *successful* ones. If `end` fires first and the promise has already resolved, that's fine — but double-check that `onClose` is a no-op after the promise settles (i.e., the cleanup removes the `close` listener before the event queue drains). The patch snippet is truncated, so I can't fully verify this, but it's a critical correctness detail.

If cleanup only happens inside `onEnd` and `onError`, there's a window where `onClose` could race. Make sure the cleanup function is called unconditionally at the start of both `onEnd` and `onError`, *before* resolving/rejecting, so the `close` listener is gone by the time the promise settles.

---

### 3. `isClientDisconnectError` is truncated in the patch

The patch for `client-disconnect-error.ts` ends mid-line:

```
export function isClientDisconnectError(error: unknown): error is ClientDisconnectErro
```

This appears to be a patch truncation artifact in the PR description rather than the actual code, but worth confirming the full implementation is present. The function body should be something like:

```ts
export function isClientDisconnectError(error: unknown): error is ClientDisconnectError {
  return error instanceof ClientDisconnectError
}
```

Make sure this is complete in the actual file.

---

### 4. `isRequestAbortError` change in `request-listener.ts`

The patch for this function is also truncated (`// ...`), but the description says it now accounts for `ClientDisconnectError`. Please ensure the updated logic handles *both* cases:

1. `request.signal.aborted && error === request.signal.reason` (original)
2. `isClientDisconnectError(error)` (new case for upload aborts)

The combination should look roughly like:

```ts
function isRequestAbortError(request: Request, error: unknown): boolean {
  return (
    (request.signal.aborted && error === request.signal.reason) ||
    isClientDisconnectError(error)
  )
}
```

If a `ClientDisconnectError` doesn't suppress the error handler, it'll bubble up to the user as an unhandled server error, which would be worse than the original hang.

---

### 5. Test coverage — consider a slow-upload scenario

The tests use a raw `net.Socket` to abort mid-connection, which is great. Consider also adding a test where:

- The client sends headers and *starts* streaming a body
- Then aborts mid-stream (not just immediately)

This ensures the `data` event + partial buffer accumulation path also rejects correctly, not just the zero-data case.

---

### 6. Minor: changelog file

The `.changes/patch.fix-...md` file is a single line with no trailing newline. Some changelog tooling (e.g., `changesets`) requires a trailing newline. Worth double-checking that your tooling accepts it as-is.

---

## Summary

The overall direction is correct and the implementation seems sound from what's visible. The main things to verify before merging:

1. `close` listener doesn't fire spuriously on successful requests (cleanup ordering).
2. `isClientDisconnectError` full implementation is present.
3. `is

mjackson:
Superseded by #11552, which merged the #11533 fix along with #11534.

PR #11552, opened 2026-06-17
Fix node fetch server abort handling

### LINKED ISSUE #11533: node-fetch-server request bodies can hang after client aborts

### LINKED ISSUE #11534: node-fetch-server can try to send an error response after headers are committed

mjackson:
This fixes two `@remix-run/node-fetch-server` failure modes around interrupted request and response streams.

- Closes #11533 by propagating incoming request `error`, `aborted`, and early `close` events into the web request body stream and `Request.signal`, so body consumers reject instead of hanging after aborted uploads.
- Closes #11534 by avoiding fallback error responses once response headers have already been committed; late response stream failures now call `onError` and destroy the response instead of writing a second response.
- Adds regressions for aborted, errored, and early-closed request bodies, late socket errors after aborts, and response body errors after headers are sent.

### DIFF
--- a/packages/node-fetch-server/.changes/patch.request-stream-aborts.md
+++ b/packages/node-fetch-server/.changes/patch.request-stream-aborts.md
@@ -1,0 +1,1 @@
+Reject request body reads when clients abort uploads, and avoid writing fallback error responses after streaming response headers are already committed (see #11533 and #11534).
--- a/packages/node-fetch-server/src/lib/lazy-request.ts
+++ b/packages/node-fetch-server/src/lib/lazy-request.ts
@@ -1,7 +1,13 @@
 import type * as http from 'node:http'
 import type * as http2 from 'node:http2'
+import type { RequestLifecycle } from './request-abort.ts'
 
 import { createLazyHeaders } from './lazy-headers.ts'
+import {
+  createRequestAbortError,
+  isRequestAlreadyAborted,
+  markRequestAbortReason,
+} from './request-abort.ts'
 
 type IncomingRequest = http.IncomingMessage | http2.Http2ServerRequest
 type ServerResponse = http.ServerResponse | http2.Http2ServerResponse
@@ -10,7 +16,19 @@ type RequestFactory<requestOptions> = (
   res: ServerResponse,
   options: requestOptions | undefined,
 ) => Request
+type LifecycleRequestFactory<requestOptions> = (
+  req: IncomingRequest,
+  res: ServerResponse,
+  options: requestOptions | undefined,
+  lifecycle: RequestLifecycle,
+) => Request
 type HeadersFactory = (req: IncomingRequest) => Headers
+type RequestLifecycleFactory = () => RequestLifecycle
+type ResponseLifecycleObserver = (res: ServerResponse, lifecycle: RequestLifecycle) => void
+interface RequestLifecycleOptions {
+  createLifecycle: RequestLifecycleFactory
+  observeResponseForLifecycle: ResponseLifecycleObserver
+}
 
 export function createLazyRequest<requestOptions>(
   req: IncomingRequest,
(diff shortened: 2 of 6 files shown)


mjackson:
@codex review

mjackson:
@codex review

mjackson:
@codex review

Issue #11556, opened 2026-06-18
Make RoutePattern opaque and nominal


mjackson:
## Goal

Make `RoutePattern<source>` an opaque, nominal parsed-pattern handle without losing the internal parsed-data path used by helpers such as `joinPatterns()`.

This issue closes when all tracked PRs below are merged, release notes and docs are current, benchmark evidence is recorded, and the listed validation commands pass.

## Issues To Fix

- `RoutePattern` exposed parsed internals as public fields: `protocol`, `hostname`, `port`, `pathname`, and `search`.
- Consumers could read token internals such as `pattern.pathname.tokens` and `pattern.hostname.optionals`.
- Consumers could construct `new RoutePattern({ ...parsedInternals })`, bypassing the parser and creating invalid states.
- `RoutePattern<source>` was a phantom generic. A parsed `/posts/:postId` pattern could be assigned to `RoutePattern<'/users/:userId'>`, causing downstream APIs to infer params from the claimed type instead of the real parsed pattern.
- Runtime parsing and type-level parsing disagreed on URL forms such as `http:///path`, `http://`, dynamic protocols, malformed protocols, dangling escapes, unmatched optionals, and missing variable names.
- `ignoreCase` did not apply consistently across static pathname text, dynamic pathname regexes, wildcard continuations, and non-ASCII pathname text.
- Key-only search params could generate hrefs that no longer matched the pattern.
- URL parser normalization was under-accounted for, especially hostname percent escapes and pathname values that change while parsing.
- Optional variants that collapsed to the same concrete pattern could be inserted more than once.
- Optional joins could generate double slashes when an optional segment was omitted.
- Missing-param errors reported only the first missing required param.
- Matcher state included unused data such as stored `undefinedParams`.
- Specificity did not fully account for protocol and port constraints.
- Port-only origin patterns were accepted even though they cannot be represented safely.
- Docs and benchmarks did not fully explain the supported API, matcher behavior, performance claims, or adapted `punycode.js` code.

## Public API Requirements

- `RoutePattern.parse(source)` is the public parser.
- `RoutePattern#source`, `RoutePattern#toString()`, and `RoutePattern#toJSON()` stay public.
- `RoutePattern` instances are opaque; public consumers cannot read parsed AST fields or construct instances from parsed internals.
- Package internals use internal-only helpers to create and read parsed parts, so `joinPatterns()` can combine existing `RoutePattern` inputs without serializing and reparsing them.
[...]

PR #11557, opened 2026-06-18
Make RoutePattern opaque and nominal

mjackson:
This makes `RoutePattern` an opaque, nominal parsed-pattern handle while preserving the no-reparse path for internals such as `joinPatterns()`.

- Hides parsed route-pattern parts behind internal storage and removes public parsed-AST construction
- Adds internal helpers for creating and reading parsed parts without serializing/reparsing existing `RoutePattern` values
- Brands `RoutePattern<source>` so incompatible parsed pattern sources are not structurally assignable
- Adds `getRoutePatternParams()` plus public support types for stable param introspection
- Moves the assets package off `pattern.pathname.tokens`
- Keeps generated href/matcher/join inference tied to the actual parsed pattern source

Closes the first PR checkbox in #11556.

### DIFF
--- a/packages/assets/src/lib/routes.ts
+++ b/packages/assets/src/lib/routes.ts
@@ -1,4 +1,8 @@
-import { RoutePattern } from '@remix-run/route-pattern'
+import {
+  getRoutePatternParams,
+  RoutePattern,
+  type RoutePatternParam,
+} from '@remix-run/route-pattern'
 import { createHref } from '@remix-run/route-pattern/href'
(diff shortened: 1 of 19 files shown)


mjackson:
Addressed in the latest revision by adding JSDoc for `RoutePatternJSON`, `RoutePatternParam`, and `getRoutePatternParams()`, plus README coverage that describes `RoutePattern` as opaque and shows `getRoutePatternParams()` as the supported param-introspection API.

PR #11567, opened 2026-06-25
Handle node-fetch-server lazy requests in fetch-proxy

markdalgleish:
While working on HMR in https://github.com/remix-run/remix/pull/11515, I hit an issue with the dev proxy where fetch-proxy couldn't proxy lazy request objects from node-fetch-server.

The core issue is that since these lazy requests currently present themselves as native `Request` instances, including `lazyRequest instanceof Request` being true, it throws when calling `new Request(lazyRequest, init)` because Node thinks it's a native request and tries to access private fields. This seems to be the same category of issue we've faced with Bun in the past.

I experimented with more holistic fixes here, like introducing an explicit `LazyRequest` type, and adding something like `.toRequest(init?)` to lazy requests while keeping their prototype as `Request`. Neither felt like an obvious step forwards to me though. Making more of our APIs accept `Request | LazyRequest` made them feel a lot less web-native. Adding a `.toRequest(init?)` method could work but it mostly gives us a cleaner internal workaround rather than fixing the consumer-facing footgun. I'm not exactly sure what the best approach here is.

So for now this PR keeps the fix intentionally scoped, opting to detect lazy requests within fetch-proxy and work around the issue so we can at least proxy lazy requests for now, which will allow me to significantly clean up the HMR demo.


markdalgleish:
Closing in favour of https://github.com/remix-run/remix/pull/11627.

PR #11571, opened 2026-06-29
Add `--test-name-pattern` support

markdalgleish:
This adds support for the `--test-name-pattern` flag / `testNamePattern` option to allow filtering of tests by name.

### DIFF
--- a/packages/test/src/test/name-pattern.test.ts
+++ b/packages/test/src/test/name-pattern.test.ts
@@ -1,0 +1,213 @@
+import * as assert from '@remix-run/assert'
+import { spawn } from 'node:child_process'
+import * as fsp from 'node:fs/promises'
+import * as path from 'node:path'
+import { stripVTControlCharacters } from 'node:util'
+import { fileURLToPath } from 'node:url'
+import { describe, it } from '../lib/framework.ts'
+import { IS_BUN } from '../lib/runtime.ts'
+
+const PKG_DIR = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..')
+const FIXTURE_DIR = path.join(PKG_DIR, '.tmp', 'test-name-pattern')
+
+describe('test name patterns', () => {
+  it(
+    'filters server, browser, and E2E tests by full name in the same default run',
+    { skip: IS_BUN },
+    async () => {
+      await writeProject(FIXTURE_DIR)
+
+      try {
+        let output = await runFixtureCli(FIXTURE_DIR)
+        let stdout = stripVTControlCharacters(output.stdout)
+
+        assert.equal(output.exitCode, 0, output.stderr || output.stdout)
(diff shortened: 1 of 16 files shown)


packages/test/src/test/name-pattern.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      try {
+        let output = await runFixtureCli(FIXTURE_DIR)
+        let stdout = stripVTControlCharacters(output.stdout)
+
+        assert.equal(output.exitCode, 0, output.stderr || output.stdout)
+        assert.match(stdout, /Found 3 test file\(s\) \(1 server, 1 browser, 1 e2e\)/)
+        assert.match(stdout, /✓ runs matching server test/)
+        assert.match(stdout, /✓ runs matching browser test/)
+        assert.match(stdout, /✓ runs matching E2E test/)
+        assert.doesNotMatch(stdout, /does not run non-matching server test/)
```
brophdawg11, line 29: Do you think we should align this behavior with our current `it.only` behavior where the filtered tests are still printed as skipped?  Or is that just more noise for the agent to parse?

Image

Node's `it.only` doesn't look like it prints skipped tests in the default reporter:

Image

Vitest does, so look like we align more closely with their output at the moment:

Image

brophdawg11: I'm also noticing our `.only` seems to be only evaluated within the nearest `describe` parent, but doesn't cause sibling `describe` suites to be skipped.  I'll get a PR up to fix that

markdalgleish: I think printing the list of non-matching tests is noise - but adding to the `skipped` count makes sense. I've pushed an update to add this.

I also think there's room for a bigger change to handling skipped test counts so that they're inlined for each total now that `--test-name-pattern` typically filters out entire files and suites, e.g. `files 1 (3 skipped)` - but I felt like this was outside the scope of this change.

brophdawg11: cool yeah I think we should probably just be consistent since this is sort of a CLI version of `.only`. So as a follow up (or part of this if you feel like it) we could change `.only` to not print skipped tests.

brophdawg11: If we update `.only` to behave that way, can this simplify a bunch and just become another way to set `test.only`?  Unsure if it should override or compliment any existing `test.only`'s in code, but we could just basically do:

```tsx
    if (testNameMatcher) {
      suite.tests.forEach((test) => {
        test.only = testNameMatcher(suite.name, test.name)
      })
    }
```

We could even consider aligning the name to something like ``--test-only` or `--only`?

brophdawg11: I opened this alternate approach in https://github.com/remix-run/remix/pull/11581 with `--only`

brophdawg11:
Superseded by #11581

PR #11574, opened 2026-06-29
Remix Guides

brookslybrand:
Continuation of #11562

See the deployed preview: https://remix-3-guides.fly.dev/docs/

I pulled this out into a separate PR because I decided to go a bit of a different architectural direction. After trying to write docs in JSX and finding it incredibly difficult, I switched to markdown + a custom `::frame` directive for loading in Remix `<Frame>`s anytime I need dynamic examples or any sort of rich component.

Other than that, the content is still oriented around the chapters set in #11562, and I'm just continually working through them.

Chapters that have been written and reviewed by me, a human:

- ✅ [Start Here](https://remix-3-guides.fly.dev/docs/start-here)
- ✅ [Routing and Controllers](https://remix-3-guides.fly.dev/docs/start-here)


guides/app/actions/docs/chapters/12-testing.md
```diff
@@ -0,0 +1,380 @@
...
+const album = {
+  artist: "Michael Jackson",
+  id: "thriller",
+  title: "Thriller",
+  year: 1983,
+};
+
+describe("AlbumEditForm", () => {
+  it("shows a pending state while submitting", async (t) => {
+    let result = render(<AlbumEditForm album={album} />);
```
brophdawg11: This might not be the right example since this form does a form submission so it probably just triggers an error in the bowser when submitting in a browser unit test.  Are there any presentational-esque components that don't have side effects and take an `onClick` or anything we could reference back to?

guides/app/actions/docs/chapters/12-testing.md
```diff
@@ -0,0 +1,380 @@
+---
+title: Testing
+description: How to choose a test boundary and test Remix routes, stateful request flows, components, and end-to-end behavior.
+---
+
```
brophdawg11, line 5: We may want to introduce the test types right up front since we reference them in the table below

```suggestion

Remix provides a test runner (`remix/test`) out of the box that allows you to write 3 types of tests: Server tests, Browser tests, and E2E tests.

```

guides/app/actions/docs/chapters/12-testing.md
```diff
@@ -0,0 +1,380 @@
...
+| -------------------------------------------------------------------- | ---------------------------------------------------------------- | --------------------------- |
+| A data helper, schema, or other plain module                         | Import it, call it, and assert on the result                     | `*.test.ts`                 |
+| An action, response, middleware, session, or database-backed request | Send a request through `router.fetch(...)`                       | `*.test.ts` or `*.test.tsx` |
+| A component event, DOM update, or browser API                        | Render the component with `remix/ui/test`                        | `*.test.browser.tsx`        |
+| Navigation or a complete browser/server flow                         | Run the router behind a test server and drive it with Playwright | `*.test.e2e.ts`             |
+
+A controller that returns the wrong status belongs in a router test. A submit button that does not enter its pending state belongs in a browser component test. Use an end-to-end test when the behavior depends on both sides working together, such as submitting a form and following its redirect to the updated page.
```
brophdawg11: Might be clearer to talk about test types here instead of file extensions

```suggestion
+| Behavior                                                             | Test boundary                                                    | Test type                      |
+| -------------------------------------------------------------------- | ---------------------------------------------------------------- | ------------------------------ |
+| A data helper, schema, or other plain module                         | Import it, call it, and assert on the result                     | Server (`*.test.{ts,tex}`)     |
+| An action, response, middleware, session, or database-backed request | Send a request through `router.fetch(...)`                       | Server (`*.test.{ts,tsx}`)     |
+| A component event, DOM update, or browser API                        | Render the component with `remix/ui/test`                        | Browser (`*.test.browser.tsx`) |
+| Navigation or a complete browser/server flow                         | Run the router behind a test server and drive it with Playwright | E2E (`*.test.e2e.ts`)          |
 
A controller that returns the wrong status belongs in a server test against the router. A submit button that does not enter its pending state belongs in a browser component test. Use an end-to-end test when the behavior depends on both sides working together, such as submitting a form and following its redirect to the updated page.

```

guides/app/actions/docs/chapters/12-testing.md
```diff
@@ -0,0 +1,380 @@
...
+npm test
+```
+
+By default, the runner discovers three test types:
+
+| Pattern                      | Environment                              |
+| ---------------------------- | ---------------------------------------- |
+| `**/*.test.{ts,tsx}`         | Server test worker                       |
+| `**/*.test.browser.{ts,tsx}` | Browser frame managed by Playwright      |
+| `**/*.test.e2e.{ts,tsx}`     | End-to-end worker with Playwright access |
```
brophdawg11: I don't think we need the "worker" concept exposed here

 ```suggestion
| ---------------------------- | -------------------------------------- |
| `**/*.test.{ts,tsx}`         | Server test                            |
| `**/*.test.browser.{ts,tsx}` | Browser test running via Playwright    |
| `**/*.test.e2e.{ts,tsx}`     | End-to-end test with Playwright access |
```

File: guides/app/actions/docs/chapters/12-testing.md [hunk 5]
```diff
@@ -0,0 +1,380 @@
...
+      secrets: ["test-secret"],
+    }),
+    sessionStorage: createMemorySessionStorage(),
+  });
+}
+```
+
+Create a fresh test router inside a test when it mutates session or middleware state. A suite may share one when its dependencies are read-only or reset between tests.
+
+A router does not keep a browser cookie jar. For a multi-request session flow, read the cookie from one response and send its name/value pair in the next request:
```
brophdawg11:
```suggestion
A router does not keep a browser cookie jar. For a multi-request session flow, you can use an [End-to-end test](#test-complete-flows-end-to-end), or it's also straightforward to read the cookie from one response and send its name/value pair in the next request:
```

guides/app/actions/docs/chapters/12-testing.md
brookslybrand: You're right, the file extensions aren't really the way to talk about this, especially because we explain them/discoverability in the next section.

This table also made me a bit confused, so I went back and forth with an agent to rework it and use 4 columns instead:
- Test boundary
- Use it for
- How to test it
- Runner type

I'm hoping this section and the next are now a little more clear and the tables do a better job orienting readers to the overall categories of tests. Let me know what you think or if I went to far with editing ([preview](https://remix-3-guides.fly.dev/docs/testing#choose-the-narrowest-useful-test))

guides/app/actions/docs/chapters/12-testing.md
brookslybrand: Ah yeah, good point. I opted for a more spelled out, but also simpler example (a `FavoriteButton` component)

Let me know what you think! ([preview](https://remix-3-guides.fly.dev/docs/testing#test-remix-components-in-the-browser))

brophdawg11: yeah I think that's perfect

Small preferential nitpick - I like destructuring off render but unsure if we have a docs preference there 

```tsx
    let { $, act, cleanup } = render(<FavoriteButton albumTitle="Thriller" />)
    t.after(cleanup)

    let button = $('button')
    assert.ok(button instanceof HTMLButtonElement)
    assert.equal(button.getAttribute('aria-pressed'), 'false')
    assert.equal(button.textContent, 'Add Thriller to favorites')

    await act(() => button.click())
    assert.equal(button.getAttribute('aria-pressed'), 'true')
    assert.equal(button.textContent, 'Remove Thriller from favorites')
  })
```

guides/app/actions/docs/chapters/12-testing.md
brophdawg11: yeah I think thats a good split.  I might just update "Render the component with `remix/ui/test`" to "Render the component with `remix/ui/test`'s `render()` helper" so the helper name is explicit?

PR #11577, opened 2026-06-30
Fix cascading update guard diagnostics

### LINKED ISSUE #11222: Improve MAX_CASCADING_UPDATES guard: warn with component breakdown instead of silent failure
## Context

We're building a social feed (post cards with reactions, comments, share buttons) using `@remix-run/component@0.6.0`. Each post uses several `clientEntry` components. With 20 posts on the initial page, we had 92+ hydration regions and hit `MAX_CASCADING_UPDATES = 50` in the scheduler.

The error `"handle.update() infinite loop detected"` fired and left most components unhydrated. No actual infinite loop — just many components hydrating in the same microtask turn.

## Minimal reproduction

```tsx
// app.tsx
import { clientEntry, on, type Handle } from "@remix-run/component";

export const LikeButton = clientEntry(
  "/app.tsx#LikeButton",
  function LikeButton(handle: Handle) {
    let liked = false;
    return ({ postId }: { postId: string }) => (
      <button
        mix={on("click", () => { liked = !liked; handle.update(); })}
        style={{ padding: "8px 16px", border: "1px solid #ccc", borderRadius: "4px",
                 background: liked ? "#ef4444" : "#fff", color: liked ? "#fff" : "#333" }}
      >
        {liked ? "♥ Liked" : "♡ Like"} ({postId})
      </button>
    );
  }
);
```

```tsx
// page.tsx — render 60 LikeButtons
{Array.from({ length: 60 }, (_, i) => (
  <LikeButton postId={`post-${i}`} />
))}
```

**Expected:** All 60 buttons hydrate and are interactive.
**Actual:** Hydration stops at 50. Buttons 51-60 render as static HTML but clicks do nothing. Console shows `handle.update() infinite loop detected`.

Can provide a StackBlitz if helpful.

## The limit is actually useful — but the DX needs work

The 50 limit pushed us to optimize. We reduced from 92 to 16 hydration regions by:

- Converting image/avatar components from `clientEntry` to server components with `onerror` CSS fallbacks
- Moving share button logic from `clientEntry` to event delegation (`data-share-url` attributes + global click handler)
- Reducing initial SSR page size from 20 to 5 posts (viewport-only, infinite scroll loads more)

**The guard caught a real performance issue.** But the experience of discovering and diagnosing it was rough:

1. **The error message is opaque.** `"handle.update() infinite loop detected"` gives no component names, no counts, and no indication of whether you're at 51 or 5000. We had to read the scheduler source to understand what was happening.

2. **It errors instead of warning.** Hitting 50 kills hydration entirely. A `console.warn` at 50 would give the same signal without breaking the page. Teams would see it, investigate, and optimize — without the "why is nothing interactive?" panic.

3. **No way to configure it.** We had to use a pnpm patch. Something like `run({ hydrationWarnLimit: 50, hydrationErrorLimit: 200 })` would let teams tune for their use case.

4. **Per-component tracking for the hard error.** 51 different components each updating once isn't a loop. One component updating 51 times is. The hard error (infinite loop detection) would be more accurate as a per-component counter, while the global counter could remain as a warning/optimization signal.

## What we patched locally

We ended up patching the scheduler to get better diagnostics:

- **Warns at 50** with a component breakdown: `[Hydration] 50 cascading updates — consider reducing hydration regions. Components: ReactionPicker×5, ShareButton×5`
- **Errors at 200** with the same breakdown in the error message
- Resets the warn flag each event loop turn

This turned a confusing breakage into an actionable optimization signal.

## Proposed patch

Here's the patch we're running locally via pnpm `patchedDependencies`. Happy to open a PR if this direction is useful.

Changes to `scheduler.js`:

```diff
 // Protect against infinite cascading updates (e.g. handle.update() during render)
-const MAX_CASCADING_UPDATES = 50;
+const MAX_CASCADING_UPDATES = 200;
+const CASCADING_UPDATES_WARN = 50;
+let _warnedThisTurn = false;
```

Reset warn flag alongside the counter:

```diff
         setTimeout(() => {
             cascadingUpdateCount = 0;
             resetScheduled = false;
+            _warnedThisTurn = false;
         }, 0);
```

Warn at 50 with component breakdown, error at 200 with the same:

```diff
                 cascadingUpdateCount++;
                 scheduleCounterReset();
+                if (cascadingUpdateCount === CASCADING_UPDATES_WARN && !_warnedThisTurn) {
+                    _warnedThisTurn = true;
+                    let names = [...batch.keys()].map(v => v.type?.name || 'anonymous');
+                    let counts = {};
+                    names.forEach(n => counts[n] = (counts[n] || 0) + 1);
+                    console.warn(
+                        '[Hydration] ' + cascadingUpdateCount +
+                        ' cascading updates — consider reducing hydration regions. Components in this batch:',
+                        Object.entries(counts).map(([n, c]) => n + '×' + c).join(', ')
+                    );
+                }
                 if (cascadingUpdateCount > MAX_CASCADING_UPDATES) {
-                    let error = new Error('handle.update() infinite loop detected');
+                    let names = [...batch.keys()].map(v => v.type?.name || 'anonymous');
+                    let counts = {};
+                    names.forEach(n => counts[n] = (counts[n] || 0) + 1);
+                    let error = new Error(
+                        'handle.update() infinite loop detected (' + cascadingUpdateCount +
+                        ' updates). Components: ' +
+                        Object.entries(counts).map(([n, c]) => n + '×' + c).join(', ')
+                    );
                     dispatchError(error);
                     return;
                 }
```

## Summary of suggestions

| Current | Suggested |
|---------|-----------|
| Hard error at 50 | Warn at 50, hard error at higher limit |
| `"infinite loop detected"` | Include component names and counts in message |
| Not configurable | `run({ hydrationWarnLimit, hydrationErrorLimit })` |
| Global counter only | Per-component counter for infinite loop detection |

## Environment

- `@remix-run/component@0.6.0`
- Guard introduced in PR #11060 (commit `0951cf49`)

ryanflorence:
The scheduler's cascading update guard used one per-turn counter for all component updates. That meant a large batch of legitimate `clientEntry` hydration/update work could look like an infinite loop and stop later components from becoming interactive.

This keeps the performance signal while making the hard guard more precise. The runtime now warns once per event-loop turn when aggregate cascading component updates cross the threshold, but only dispatches the infinite-loop error when the same component instance repeatedly updates itself.

Closes #11222.

- Adds component-name/count diagnostics to the aggregate warning and hard error
- Tracks the hard error limit per component instance instead of globally
- Preserves the existing protection for `handle.update()` loops during render
- Adds a regression for 60 component instances updating in the same event-loop turn

### DIFF
--- a/packages/ui/.changes/patch.cascading-update-guard.md
+++ b/packages/ui/.changes/patch.cascading-update-guard.md
@@ -1,0 +1,1 @@
+Changed the scheduler's cascading update guard to warn when many component updates happen in one event-loop turn and only throw the infinite-loop error when a single component instance repeatedly updates itself. This keeps large `clientEntry` hydration bursts interactive while still surfacing component names and counts for diagnosis.
--- a/packages/ui/src/runtime/scheduler.ts
+++ b/packages/ui/src/runtime/scheduler.ts
@@ -35,11 +35,19 @@ export interface Scheduler {
   dequeue(): void
 }
 
-// Protect against infinite cascading updates (e.g. handle.update() during render)
-const MAX_CASCADING_UPDATES = 50
+const CASCADING_UPDATE_WARN_THRESHOLD = 50
+const MAX_CASCADING_COMPONENT_UPDATES = 50
 
 export type SchedulerPhaseEvent = Event & {
   parents: ParentNode[]
+}
+
+function getComponentName(vnode: CommittedComponentNode): string {
+  return vnode.type.name || 'Anonymous'
+}
+
+function formatComponentCounts(counts: Map<string, number>): string {
+  return Array.from(counts, ([name, count]) => `${name} x${count}`).join(', ')
 }
 
 /**
@@ -62,7 +70,10 @@ export function createScheduler(
   let postCommitTasks: EmptyFn[] = []
   let flushScheduled = false
   let flushing = false
-  let cascadingUpdateCount = 0
+  let cascadingComponentUpdateCount = 0
+  let cascadingComponentUpdateCounts = new WeakMap<CommittedComponentNode, number>()
+  let cascadingComponentNameCounts = new Map<string, number>()
+  let warnedAboutCascadingUpdates = false
   let resetScheduled = false
   let phaseEvents = new EventTarget()
   let phaseListenerCounts: Record<SchedulerPhaseType, number> = {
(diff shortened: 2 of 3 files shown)


ryanflorence:
@sbaleno mind trying this patch out for #11222?

sbaleno:
@ryanflorence All good here now - thank-you

brookslybrand (approved):

PR #11578, opened 2026-07-04
Change literal function to use 'const' for value

jvrdnd:
Precisely types the output to the actual string value. This allows some nice properties, in particular to use it as a discriminant in combination with variant (e.g., in a switch statement, you can narrow the schema).


mjackson:
Thanks, @jvrdnd! This makes discriminated schema outputs much nicer to work with.

PR #11585, opened 2026-07-09
Stream frames with declarative out-of-order directives

jacob-ebey:
Non-blocking `<Frame>` streaming now uses the declarative out-of-order streaming directives instead of a runtime-driven template swap. Pending fallbacks are wrapped in `<?start name="ID">` / `<?end name="ID">` inside
the existing `<!-- rmx:f:ID -->` markers, and resolved content streams as `<template for="ID">` (was `<template id="ID">`).

Browsers that natively process the directives perform the DOM swap with no JavaScript. Other browsers use a small MutationObserver polyfill that `run()` installs automatically, which moves `<template for>` content
into its region and removes the directives. In both cases the runtime no longer extracts templates itself; it hydrates the resolved region after the swap, while reloads and navigation continue to reconcile fetched-stream content through the template bus.

### DIFF
--- a/packages/ui/.changes/minor.native-out-of-order-streaming.md
+++ b/packages/ui/.changes/minor.native-out-of-order-streaming.md
@@ -1,0 +1,7 @@
+---
+'@remix-run/ui': minor
+---
+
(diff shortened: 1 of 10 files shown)


brophdawg11:
- https://developer.chrome.com/blog/declarative-partial-updates
- https://github.com/WICG/declarative-partial-updates
- https://dev.to/grimicorn/declarative-partial-updates-out-of-order-html-streaming-without-a-framework-1086

PR #11589, opened 2026-07-10
Add regression coverage for function-valued component children

brookslybrand:
This PR adds regression coverage for function-valued component children. The underlying hydration bug is already resolved by the explicit component VNode architecture introduced in #11659, so no additional runtime change is needed here.

Previously, client hydration converted element children into VNodes before distinguishing host elements from components. Function-valued children passed to a component could therefore throw `Unexpected RemixNode` before the component invoked them, interrupting hydration and preventing client-entry event handlers from attaching.

The regression test server-renders and hydrates a client entry containing a render-prop component, then verifies that hydration completes and its interaction remains active. This protects the behavior now provided by component VNodes keeping component props opaque while host element and fragment children are normalized.

### DIFF
--- a/packages/ui/src/test/frame.test.tsx
+++ b/packages/ui/src/test/frame.test.tsx
@@ -107,6 +107,51 @@ describe('run', () => {
     frame.dispose()
   })
 
+  it('hydrates function-valued children passed to a component', async () => {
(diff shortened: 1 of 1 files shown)


ryanflorence:
~~I don't think this makes sense since clientEntry is a serialization boundary and functions can't be serialized.~~

Oh I see now, nvm

PR #11590, opened 2026-07-10
Add dark mode-compatible UI component styles

brophdawg11:
Updates first-party UI component styles and built-in component demos so their authored colors work in both light and dark color schemes.

- Converts shared component color tokens, control surfaces, borders, focus rings, and demo-only text/surface colors to adaptive `light-dark(...)` values.
- Updates built-in component test expectations for the emitted adaptive CSS.
- Fixes the checkbox checkmark so it uses `currentColor` through a mask instead of a fixed white SVG stroke.

This includes a UI package change file for the dark-mode styling fix.

I came across this when looking at the examples in the API docs:https://api.remix.run/api/remix/ui/accordion/demos/accordion/

To test locally: `pnpm build; pnpm run docs; cd docs; pnpm dev`


packages/ui/src/checkbox/index.ts
```diff
@@ -104,7 +104,9 @@ const baseStyle: CSSMixinDescriptor = css({
     top: '50%',
     width: 'var(--rmx-checkbox-check-size)',
     height: 'var(--rmx-checkbox-check-size)',
-    background: `${checkIconUrl} center / contain no-repeat`,
+    background: 'currentColor',
+    mask: `${checkIconMask} center / contain no-repeat`,
+    WebkitMask: `${checkIconMask} center / contain no-repeat`,
```
brophdawg11, line 109: @ryanflorence I think this was the only change that wasn't simply adding a `light-dark`

PR #11598, opened 2026-07-14
data-table: DatabaseResource + CLI

pcattori:
Frameworks like Rails manage database connections for you. Remix is different here: users write their own server, so Remix acts more like a library than a framework. That means users also own database connection setup and lifecycle.

At the same time, several parts of an app need database access:

- request handlers and middleware
- CLI commands
- tests
- migrations and seeds

Those should all be able to create a properly configured database connection without duplicating driver config, migration setup, or connection lifecycle code.

This PR introduces a database resource layer for `data-table` and updates demos/CLI code to use it for connection lifecycle, migrations, seeds, and test setup.

### Avoid connection lifecycle footguns

A database client/connection should be released when the caller is done with it. Forgetting to release can leak pooled connections or keep SQLite handles open.

This PR makes connected `Database` clients async disposable:

```ts
await using db = await database.connect()

await db.query(users).all()
```

`Database[Symbol.asyncDispose]()` delegates to `Database.release()`, which delegates to `adapter.release()`.

### Make `connect()` mean “new connection”

Separate calls to `database.connect()` should get separate connections. That gives each request/test/task an isolated connection boundary and guarantees separate transaction state.

For SQLite, it is tempting to reuse one `DatabaseSync` instance forever. Instead, the SQLite resource returns a new `DatabaseSync`-backed adapter for each `connect()` call:

```ts
await using first = await database.connect()
await using second = await database.connect()

// separate SQLite handles, separate transaction state
```

### Model database lifecycle as a resource

The development workflow needs explicit database resource operations:

```ts
await database.drop()
await database.create()
await using db = await database.connect()
await migrate(db, migrations)
await seed(db)
```

These operations belong to the database resource, not an individual connection.

For different adapters, `create()` and `drop()` mean different things:

- SQLite: create/delete the database file
- Postgres/MySQL: connect to an admin database/server and create/drop the app database

This gives adapters room to implement the right behavior for their platform.

## First-party database config

This PR keeps first-party config shapes for SQLite, Postgres, and MySQL instead of exposing the underlying driver config objects directly.

The underlying driver is an implementation detail. Passing driver config through as public API leaks that detail and makes future driver changes harder. We have already hit this with SQLite: switching from `better-sqlite3` to built-in `node:sqlite` changed available options and defaults.

A first-party config layer lets Remix define stable semantics:

```ts
createSqliteDatabase({ path: './db/app.sqlite' })

createPostgresDatabase({
  host: 'localhost',
  database: 'app',
  user: 'postgres',
})
```

The low-level `DatabaseAdapter` objects are now implementation details of these `create*Database()` functions. Packages can translate Remix's config options to whichever driver they use internally without making the driver API or adapter construction part of Remix's public app-level contract.

## CLI integration

The CLI needs a conventional place to find an app database resource:

```ts
// app/data/database.ts
export const database = createSqliteDatabase({ path: './db/app.sqlite' })
```

Then CLI commands can create their own scoped connection:

```ts
await using db = await database.connect()
await migrate(db, migrations)
```

This supports commands like:

```sh
remix db create
remix db drop
remix db migrate
remix db migrate status
remix db seed
remix db reset
```

`reset` is currently expressed as:

```ts
await database.drop()
await database.create()
await using db = await database.connect()
await migrate(db, migrations)
await seed(db)
```

## Migrations

This PR replaces the stateful `createMigrator()` API with stateless functions:

```ts
import { migrate, status } from 'remix/data-table/migrations'

await migrate(db, migrations)
let migrationStatus = await status(db, migrations)
```

This keeps migrations simple and makes resource-level convenience methods possible later.

## Tests and demos

Tests benefit from database resources because they can reset a test database programmatically:

```ts
beforeEach(async () => {
  await database.drop()
  await database.create()
  await using db = await database.connect()
  await migrate(db, migrations)
  await seed(db)
})
```

Demos now use database resources for setup and request-time connections. Middleware opens a scoped connection per request:

```ts
return async (context, next) => {
  await using db = await database.connect()
  context.set(Database, db)
  return await next()
}
```

The `return await next()` is intentional: it keeps the `await using` scope alive until downstream middleware and handlers finish.

## Open questions

- `/node` exports: do we just expect a node-like server env (deno or bun)? what if pkg is already a server-only thing?
- combine data-table-* pkgs into data-table pkg? with new config shape, not leaking impl details
- Should `data-table` be renamed to `db`?
- Should `Database` be renamed to `Connection`, and `DatabaseResource` to `Database`?
  - This may better match SQL concepts like `CREATE DATABASE` and `DROP DATABASE`.
- What should the file convention be?
  - `app/data/database.ts`?
  - `db/config.ts`?
  - Should `schema.ts` also live under `db/`?
- Should migrations and seed functions be configured directly on the database resource?

Possible future shape:

```ts
export const database = createSqliteDatabase(
  { path: './db/app.sqlite' },
  {
    migrations: () => loadMigrations('./db/migrations'),
    seed: async (db) => {
      // idempotent seed
    },
  },
)

await database.migrate()
await database.seed()
await database.reset()
```

This PR stops short of that layer and focuses on the lower-level resource and connection lifecycle primitives.


pcattori:
Superseded by https://github.com/remix-run/remix/pull/11598

PR #11607, opened 2026-07-15
Add conventional render() middleware for Remix UI

### LINKED ISSUE #11605: Add conventional render() middleware for Remix UI
## Context

`renderWith()` is a useful low-level primitive: it creates a request-scoped renderer and installs it as `context.render`. However, Remix applications currently have to build the standard Remix UI renderer themselves.

The default template does this in `template/app/middleware/render.tsx`, and the Rendering UI guide currently teaches the same pattern. That app-owned module must:

- call `renderToStream()` with the current request URL and abort signal
- resolve `<Frame>` requests back through the current router
- construct and forward the appropriate headers for those internal requests
- resolve `clientEntry()` source URLs through the asset server
- infer client-entry export names
- turn the resulting stream into an HTML `Response`

This is framework integration code rather than application policy. It is long enough that users will copy it without understanding it, delicate enough for copies to drift, and already inconsistent across the template and demos. In particular, frame URL resolution, target propagation, request-header forwarding, error handling, and client-entry resolution should not be reimplemented by every application.

We should keep rendering explicit at the action boundary:

```tsx
return context.render(<AlbumPage album={album} />)
```

We should also keep `renderWith(factory)` as the generic escape hatch for JSON, email previews, other UI runtimes, and applications with fully custom response policies. The router itself should remain independent of Remix UI, and actions should continue returning explicit Web `Response` objects rather than having the router implicitly convert returned JSX.

## Proposed API

Add a conventional `render()` middleware alongside `renderWith()`:

```ts
import { render } from 'remix/middleware/render'
import { staticFiles } from 'remix/middleware/static'

export const assets = createAssetServer({
  // app asset configuration
})

export const router = createRouter({
  middleware: [staticFiles('./public'), render({ assets })],
})
```

`render({ assets })` installs the standard typed `context.render(node, init?)` function. Applications without source-served `clientEntry()` modules should be able to use `render()` with no options.

The common API should stay intentionally small. `assets` and an optional `onError` callback are reasonable initial options. Avoid copying the entire `RenderToStreamOptions` surface onto this helper. Applications that need to replace the standard rendering pipeline can use `renderWith()`.

## Standard behavior owned by `render()`

The middleware should:

- create one renderer per request and preserve its input and `ResponseInit` types on `context.render`
- call `renderToStream()` with the request URL and abort signal
- resolve nested and targeted frames through `context.router`
- create safe internal `GET` requests for frames, preserving authentication/session-relevant headers while overriding or removing headers that are invalid for the internal request
- propagate the current and top frame URLs correctly
- resolve source-based client-entry URLs through the supplied asset server while using the same explicit-hash/named-component export rules as the UI renderer
- return `createHtmlResponse(stream, init)`, preserving caller-provided status and headers
- report rendering errors through `onError`, defaulting to the renderer's existing behavior

The implementation should define and test a consistent policy for non-successful frame responses. The default should preserve useful application error content where possible instead of requiring every app to invent an error fragment.

## Plan

- [ ] Add the high-level `render(options?)` API and its types. Keep `renderWith(factory)` and `Renderer` backward compatible.
- [ ] Implement the standard Remix UI renderer using `renderToStream()` and `createHtmlResponse()`.
- [ ] Implement a shared frame resolver that handles nested frame URLs, targets, aborts, relevant request headers, and non-successful responses consistently.
- [ ] Integrate optional asset-server client-entry resolution without duplicating export-name parsing rules.
- [ ] Add focused tests for typed `context.render`, response status/headers, doctypes, cancellation, nested and targeted frames, session/auth header forwarding, frame errors, and client-entry URLs.
- [ ] Expose the API from `remix/middleware/render` and add the appropriate package change files.
- [ ] Replace the copied renderer in the default template and representative demos with `render({ assets })`.
- [ ] Update the render-middleware README and the Rendering UI guide. Teach `render()` as the normal path and move the manual `renderWith()` composition to an advanced/custom-renderer example.

## Acceptance criteria

- A normal Remix UI application does not need an app-owned `middleware/render.tsx` module.
- The default template configures rendering with one middleware call.
- `context.render(<Page />, init)` remains fully typed and returns a Web `Response` with the supplied status and headers.
- Frames, request cancellation, generated styles, and hydrated client entries work through the conventional middleware.
- `renderWith()` continues to support arbitrary request-scoped renderers without a behavior or type regression.

mjackson:
Remix UI apps no longer need to own a custom `renderWith()` setup for normal HTML responses. This adds a conventional `render()` middleware while keeping rendering explicit in each action through `context.render()`.

- adds `render({ assets?, onError? })` with a typed `context.render(node, init)`
- resolves nested and targeted frames through the current router, follows standard redirect statuses, requires HTML frame responses, forwards safe headers, and cancels work with the request
- resolves source-based `clientEntry()` modules and their preloads through the asset server, with a clear error when a file entry has no asset server
- keeps `renderWith()` as the escape hatch for JSON, email, and other custom rendering pipelines
- updates the default template, docs apps, demos, guides, and agent guidance to use the conventional middleware

```tsx
import { render } from 'remix/middleware/render'
import { staticFiles } from 'remix/middleware/static'

let assets = createAssetServer({
  basePath: '/assets',
  mounts: { app: 'app', npm: 'node_modules' },
  allowFiles: ['app/**/public/**'],
  allowPackages: ['remix'],
})

let router = createRouter({
  middleware: [staticFiles('./public'), render({ assets })],
})

router.get('/', (context) =>
  context.render(<HomePage />, {
    headers: { 'Cache-Control': 'no-store' },
  }),
)
```

Closes #11605


packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,199 @@
...
+  'If-Range',
+  'If-Unmodified-Since',
+  'Keep-Alive',
+  'Range',
+  'TE',
+  'Trailer',
+  'Transfer-Encoding',
+  'Upgrade',
+] as const
+const CREDENTIAL_HEADERS = ['Authorization', 'Cookie', 'Proxy-Authorization'] as const
```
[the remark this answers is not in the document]
mjackson: Fixed in [bbdc3b6f2](https://github.com/remix-run/remix/commit/bbdc3b6f2ab00adf2380b25e63b1a0dd2023dae0). Cross-origin frame requests are now rebuilt from a framework-header allowlist, so custom credentials such as X-Api-Key and X-Session-Token are not forwarded.

packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,199 @@
...
+  let exportName = explicitExportName || component.name
+
+  if (!exportName) {
+    throw new Error(
+      `clientEntry() requires either an export name in the entry ID (e.g., "/js/module.js#ComponentName") or a named component function. Received "${entryId}".`,
+    )
+  }
+
+  return {
+    href: await assets.getHref(sourceId),
```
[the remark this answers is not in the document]
mjackson: Fixed in [bbdc3b6f2](https://github.com/remix-run/remix/commit/bbdc3b6f2ab00adf2380b25e63b1a0dd2023dae0). Asset resolution now applies only to file: source IDs; public paths and CDN URLs retain the default UI href behavior.

.agents/skills/remix/references/assets-and-browser-modules.md
```diff
@@ -77,7 +77,7 @@ let preloads = await assetServer.getPreloads(['app/assets/entry.ts'])
 
 Use this when rendering documents or layouts that boot browser behavior with a known client entry.
 
-When resolving hydrated client entries during server rendering, pass the source entry ID from `clientEntry(import.meta.url, ...)` to `getHref()` inside `resolveClientEntry`. Keep export-name resolution in that render helper, and avoid hard-coding public asset URLs in source-owned component modules.
+For normal Remix applications, pass the asset server to `render({ assets: assetServer })` from `remix/middleware/render`. The middleware resolves the source entry ID from `clientEntry(import.meta.url, ...)` with `getHref()` and applies the UI renderer's explicit-hash or named-component export rules. Use a custom `resolveClientEntry` callback only when building a custom rendering pipeline.
```
markdalgleish: I think we need to align the canonical naming of the asset server variable to streamline `render({ assets: assetServer })`, as either `assets` or `assetServer`. In your PR description you had it as `render({ assets })` which is nice and clean, although `render({ assetServer })` does feel a little more explanatory when read in isolation - I don't really mind either.

packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,216 @@
...
+  component: { readonly name: string },
+): Promise<{ href: string; exportName: string }> {
+  let hashIndex = entryId.lastIndexOf('#')
+  let sourceId = hashIndex === -1 ? entryId : entryId.slice(0, hashIndex)
+  let explicitExportName = hashIndex === -1 ? '' : entryId.slice(hashIndex + 1)
+  let exportName = explicitExportName || component.name
+
+  if (!exportName) {
+    throw new Error(
+      `clientEntry() requires either an export name in the entry ID (e.g., "/js/module.js#ComponentName") or a named component function. Received "${entryId}".`,
```
markdalgleish: I think we should make the `import.meta.url` usage more canonical here, e.g.


```suggestion
      `clientEntry() requires either an export name in the entry ID (e.g., import.meta.url + "#ExportName") or a named component function. Received "${entryId}".`,
```

markdalgleish (commented):
I love this, it really cleans up the rendering boilerplate - and as a bonus, it ensures the community is more aligned on how to resolve `import.meta.url` as a client entry ID, which will help with things like sharing hydrated component libraries (which @jacob-ebey was talking about today).

packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,216 @@
...
+  'Trailer',
+  'Transfer-Encoding',
+  'Upgrade',
+] as const
+const CROSS_ORIGIN_FRAME_HEADERS = [
+  'Accept',
+  'Accept-Encoding',
+  FRAME_HEADER,
+  FRAME_TARGET_HEADER,
+  TOP_FRAME_SRC_HEADER,
```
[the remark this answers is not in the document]
mjackson: Fixed in [f8324f2bd](https://github.com/remix-run/remix/commit/f8324f2bde25518f697fccbd87fb27bc8f299219). X-Remix-Top-Frame-Src is no longer forwarded to cross-origin frames, so the outer request URL stays on-origin. The cross-origin frame handler falls back to its own request URL as the top frame src. Added a test asserting the header is absent cross-origin.

packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,216 @@
...
+  target?: string,
+  frameContext?: ResolveFrameContext,
+): Promise<string | ReadableStream<Uint8Array>> {
+  let currentFrameSrc = frameContext?.currentFrameSrc ?? context.request.url
+  let topFrameSrc = frameContext?.topFrameSrc ?? getTopFrameSrc(context.request)
+  let frameUrl = new URL(src, currentFrameSrc)
+  let headers = createFrameRequestHeaders(context.headers, target, topFrameSrc)
+  let response = await followFrameRedirects(context, frameUrl, headers)
+
+  if (response.body != null) return response.body
```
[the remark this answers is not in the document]
mjackson: This doesn't reproduce. `renderToStream()`'s `splitFirstChunk()` (in `@remix-run/ui`) skips chunks with no renderable HTML — `hasRenderableHtml()` strips the doctype before checking — so the doctype-only chunk that `createHtmlResponse()` prepends is dropped, and the first real chunk (the frame's full initial document, which `renderToStream()` emits as a single chunk) renders in place at the frame slot with its doctype stripped. Tail chunks are only the frame's own `<template>`/pending-frame follow-ups, which are designed to stream after the initial chunk. Added a regression test in [f8324f2bd](https://github.com/remix-run/remix/commit/f8324f2bde25518f697fccbd87fb27bc8f299219) asserting blocking frame bodies from `context.render()` land inside their slot markers ahead of subsequent markup, with no leaked doctype.

.agents/skills/remix/references/assets-and-browser-modules.md
mjackson: Standardized on `assets` in [f8324f2bd](https://github.com/remix-run/remix/commit/f8324f2bde25518f697fccbd87fb27bc8f299219) — it matches the option name and the `remix/assets` import, so call sites read `render({ assets })`. Renamed the variable in the skill references, template, and the bookstore/frames/sse/frame-navigation/guides apps. Timeboxer keeps the explicit `render({ assets: assetServer })` because `assets` is already its route controller export there.

packages/render-middleware/src/lib/render-ui.ts
mjackson: Applied in [f8324f2bd](https://github.com/remix-run/remix/commit/f8324f2bde25518f697fccbd87fb27bc8f299219).

packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,217 @@
...
+  let exportName = explicitExportName || component.name
+
+  if (!exportName) {
+    throw new Error(
+      `clientEntry() requires either an export name in the entry ID (e.g., import.meta.url + "#ExportName") or a named component function. Received "${entryId}".`,
+    )
+  }
+
+  return {
+    href: sourceId.startsWith('file:') ? await assets.getHref(sourceId) : sourceId,
```
markdalgleish: Now that we have `preloads` support for client entries, we should be providing these here too, via `assets.getPreloads(sourceId)`. This will also need to be updated once client entries support having an `importMap` field, and we'll want to use the new `assets.getScriptEntry(sourceId)` API that provides the `href`, `importMap` and `preloads` in a single call.

packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,217 @@
...
+  let currentFrameSrc = frameContext?.currentFrameSrc ?? context.request.url
+  let topFrameSrc = frameContext?.topFrameSrc ?? getTopFrameSrc(context.request)
+  let frameUrl = new URL(src, currentFrameSrc)
+  let headers = createFrameRequestHeaders(context.headers, target, topFrameSrc)
+  let response = await followFrameRedirects(context, frameUrl, headers)
+
+  if (response.body != null) return response.body
+  if (response.ok) return ''
+
+  return `<pre>Frame error: ${response.status} ${escapeHtml(response.statusText)}</pre>`
```
markdalgleish: This feels like something you'll want to customise?

packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,217 @@
...
+  requestHeaders: Headers,
+  target: string | undefined,
+  topFrameSrc: string,
+): Headers {
+  let headers = new Headers(requestHeaders)
+
+  for (let name of FRAME_REQUEST_HEADERS_TO_REMOVE) {
+    headers.delete(name)
+  }
+  for (let name of headers.keys()) {
```
markdalgleish: Comment from Codex:

> This deletes headers while iterating the live `Headers` collection. With typical browser headers, this skips alternating entries: `Sec-Fetch-Mode` and `Sec-Fetch-User` remain on the internal frame request. These incorrectly describe it as a user-initiated top-level navigation and may affect middleware or route behavior. Iterate over `[...headers.keys()]` instead. The test only supplies one `Sec-Fetch-*` header, masking the bug.

packages/render-middleware/src/lib/render-ui.ts
```diff
@@ -0,0 +1,217 @@
...
+  component: { readonly name: string },
+): Promise<{ href: string; exportName: string }> {
+  let hashIndex = entryId.lastIndexOf('#')
+  let sourceId = hashIndex === -1 ? entryId : entryId.slice(0, hashIndex)
+  let explicitExportName = hashIndex === -1 ? '' : entryId.slice(hashIndex + 1)
+  let exportName = explicitExportName || component.name
+
+  if (!exportName) {
+    throw new Error(
+      `clientEntry() requires either an export name in the entry ID (e.g., import.meta.url + "#ExportName") or a named component function. Received "${entryId}".`,
```
markdalgleish: We should have different messaging here based on whether `assets` is provided or not, because if you're not using the asset server, it's not helpful to be told to use `import.meta.url`.

packages/render-middleware/src/lib/render-ui.ts
brophdawg11: yeah I went back and forth on that in the client too with the default client side implementation.  I landed on the following without a user-provided `errorReturnValue` type option since it's easy enough to provide your own implementation if you want to customize.

```tsx
async function defaultResolveFrame(src: string, options?: ResolveFrameOptions): Promise<Response> {
  let response = await fetch(src, { ... });

  if (!response.ok) {
    throw new Error(`Failed to resolve frame: ${response.status} ${response.statusText}`.trimEnd())
  }

  return response
}

```

This render middleware does a lot more so it's arguably harder to "eject" so it might make more sense to provide an option

```tsx
render({ assets, onFrameError: (response) => <MyFrameError response={response} /> });
```

brophdawg11: The other option is to open up a less granular `resolveFrame` option but that entails a decent bit of boilerplate for the redirect following and such

packages/render-middleware/src/lib/render-ui.ts
mjackson: Fixed in [1f11f3139](https://github.com/remix-run/remix/commit/1f11f31397190d907dd2d618316ac36b6132e7ce). Source client entries now include the result of `assets.getPreloads(sourceId)`.

packages/render-middleware/src/lib/render-ui.ts
mjackson: Fixed in [1f11f3139](https://github.com/remix-run/remix/commit/1f11f31397190d907dd2d618316ac36b6132e7ce). Header names are copied before deletion, and the test now covers all four `Sec-Fetch-*` headers.

packages/render-middleware/src/lib/render-ui.ts
mjackson: Fixed in [e24690ac5](https://github.com/remix-run/remix/commit/e24690ac5520156d8830a7d43ed0a7b334fe9caf). Without `assets`, the error now uses a public URL example. With `assets`, it uses `import.meta.url`. A `file:` entry without `assets` is rejected first without including the local path.

packages/render-middleware/src/lib/render-ui.ts
mjackson: Fixed in [e24690ac5](https://github.com/remix-run/remix/commit/e24690ac5520156d8830a7d43ed0a7b334fe9caf). I removed the built-in `<pre>` fallback. A failed frame response with no UI now throws, while an app can customize the result by returning its own HTML error body.

PR #11608, opened 2026-07-15
Add database lifecycle APIs and Remix CLI commands

pcattori:
Adds first-class database lifecycle support to Remix apps through `app/db.ts`, the data-table API, and the Remix CLI.

- Adds `Database.wipe()`, `Database.migrate()`, `Database.migrationStatus()`, and `Database.reset()` plus typed `GetMigrations` and `Seed` contracts.
- Adds the typed `runRemixDb()` API from `remix/data-table/cli` while keeping argument parsing, help, errors, module loading, and process lifecycle in the Remix CLI.
- Adds `remix db wipe`, `migrate`, `status`, `seed`, and `reset`, including targeted migrations with `--to`.
- Detects applied journal entries whose migration files are missing: status reports them as `missing`, and migration runs stop before executing SQL.
- Adds config-backed MySQL, PostgreSQL, and SQLite adapters that can recreate and reconnect to databases; existing driver-owned clients and pools remain supported for non-destructive use.
- Runs MySQL and PostgreSQL migrations through scoped, connection-bound advisory locks so pooled migration work, journal updates, and lock release use the connection that owns the lock.
- Updates the bookstore, social-auth, and timeboxer demos to use `app/db.ts` while preserving automatic startup migration and seed behavior.

```ts
// app/db.ts
import { createDatabase, type GetMigrations, type Seed } from 'remix/data-table'
import { loadMigrations } from 'remix/data-table/migrations/node'
import { createSqliteDatabaseAdapter } from 'remix/data-table/sqlite'

export let db = createDatabase(
  createSqliteDatabaseAdapter({ filename: 'db/app.sqlite', foreignKeys: true }),
)
export let getMigrations: GetMigrations = () =>
  loadMigrations(path.join(import.meta.dirname, '../db/migrations'))
export let seed: Seed = async (db) => {
  // Initialize application data.
}
```

```sh
remix db migrate
remix db migrate --to 20260228090000_add_orders
remix db status
remix db seed
remix db reset
remix db wipe
```

`wipe` and `reset` are destructive and require config-backed adapters.


mjackson:
Pushed five commits addressing the review findings, grouped by area:

**Migration runner and db CLI** (`edc74d2c2`)
- `--to` now resolves against both bare migration ids and full `<id>_<name>` directory names, with explicit errors for unknown and ambiguous targets (the PR description's own `--to` example previously failed).
- Journal integrity checks are direction-aware: rolling back skips orphaned journal rows instead of erroring, while applying up migrations still hard-fails on drift.
- `remix db migrate` prints per-migration progress (`applied <id>_<name>` / `no pending migrations`), and `runRemixDb`'s contract is documented: it always resolves `0` and throws on command failures.

**MySQL/Postgres adapters** (`2bdb9cfdf`)
- `wipe()` now requires a resolvable database name. Postgres resolves connection string path → `config.database` → `PGDATABASE` and throws otherwise, instead of silently assuming `postgres` (which could wipe the wrong database). MySQL no longer falls back to `MYSQL_DATABASE`, which `mysql2` itself doesn't honor.
- Failed lock acquisition/callback/unlock now destroys the pool-reserved connection instead of returning it to the pool, where it would keep holding the advisory lock. User-supplied dedicated clients are not destroyed since the adapter doesn't own their lifecycle — happy to change that if you'd rather end those too.
- Postgres advisory lock waits are bounded with a 60s `lock_timeout` to match MySQL's `GET_LOCK(name, 60)`; previously a stuck lock-holder meant an unbounded hang.
- Postgres `wipe()` resolves the maintenance config before closing the pool, so a malformed connection string no longer leaves the adapter permanently closed.
- Reentrant `withMigrationLock` calls throw instead of deadlocking; lock/wipe test coverage is ported to Postgres and both packages restore the 90% function coverage threshold.

**SQLite adapter** (`0295ac40f`)
- `wipe()` closes the handle before unlinking and removes `-wal`/`-shm`/`-journal` sidecars, so data is actually released; `:memory:` databases are reopened rather than no-op'd.
- Driver loading, `foreign_keys`, and `busy_timeout` are handled consistently across `node:sqlite` and `bun:sqlite` (the two disagree on defaults).
- Fixed a Bun-only bug where `hasTable` always returned `true` because `bun:sqlite`'s `Statement.get()` returns `null` where `node:sqlite` returns `undefined` — this made `wipe()` of a `:memory:` database appear to succeed while leaving all data in place. CI skips the bun tests for this PR, but `pnpm run test:bun` passes locally on Bun 1.3.14.

**Remix CLI** (`62ff4f9b1`)
- `remix db wipe`/`reset` require `--force`, refused at parse time before any worker spawns.
- The db worker runs with `cwd` set to the discovered app root so relative sqlite paths resolve against the app rather than wherever the shell happens to be.
- Expected worker errors get tailored messages; unexpected ones print full stacks. Shell completion covers the new subcommands.

**Test runner and demos** (`8c4677b63`)
- `remix test` defaults `NODE_ENV=test` when unset. Without this, running the bookstore tests wiped the development database, since `app/db.ts` only switches to `:memory:` when `NODE_ENV === 'test'`. An explicit `NODE_ENV` is preserved.
- timeboxer gets the same test/in-memory switch, honors `DATABASE_URL`, and creates the `db/` directory before opening the file; bookstore gains tests for seeded password hashes and migrate/seed idempotency.

Validation: `lint`, `format`, `validate-package-meta`, `typecheck:changed`, and `test:changed` all pass, plus per-package test runs for the four data-table packages and the sqlite bun suite.

Not covered here: the MySQL/Postgres integration tests (`adapter.integration.test.ts`) weren't run locally since they need live servers — the new `lock_timeout` statements and wipe ordering are worth a CI pass against real databases. Also note the Postgres lock path resets `lock_timeout` to `default` afterwards, which would clobber a custom session-level value on a user-supplied client.

PR #11616, opened 2026-07-17
Use colocated public directories for browser source

brookslybrand:
Establish colocated `public/` directories inside `app/` as Remix's explicit boundary for browser-reachable source, while the root `public/` directory remains the home for static files served unchanged.

- Configure asset servers to expose `app/routes.ts` and `app/**/public/**`, deny colocated tests, and allow the `remix` package explicitly.
- Place root browser entries in `app/actions/public/`, feature-owned browser modules in the owning action's `public/` directory, and shared browser UI in `app/ui/public/`.
- Apply the convention throughout the project template, every demo with browser assets, the UI demo, guides, package documentation, release notes, and public API examples.
- Resolve the template entry href and module graph preloads once in `app/assets.ts`, then render them from the document head.
- Limit `remix doctor` to checking that `app/actions` mirrors the nested shape of `app/routes.ts` and that each route map with direct routes has a controller file. Unrelated action content is ignored, and `remix doctor --fix` does not generate controllers.

```txt
app/
  actions/
    controller.tsx
    public/
      entry.ts
    account/
      controller.tsx
      public/
        account-menu.tsx
  ui/
    public/
      dialog.tsx
  routes.ts
public/
  favicon.svg
```

See [`remix-website#474`](https://github.com/remix-run/remix-website/pull/474) for a larger app using this convention.


okalil:
My proposal may sound unpopular because it slightly shifts the paradigm that has been used in the demos.
But the idea is basically to make the `/app/ui` folder the “source of truth” for everything related to the front-end, centralizing all components (grouped by functionality), not only when they are shared (I've been doing this in my [experimental remix demo](https://github.com/okalil/discussions-remix) as reference).

Although at first this “takes away” the practice of placing components directly inside `/actions`, it creates clear and obvious purposes for each category:
- `/app/actions` → everything “back-end”
- `/app/ui` → everything “front-end”

From that point on, assuming that “everything front-end” is “safe” to expose — even when used only on the server side — we could make allowFiles be exactly everything inside that folder.

brookslybrand:
Thanks for sharing @okalil, I'm glad that's working for you. That actually reminds me a lot of early versions of Remix 0.x iircc

We're not going to go that direction for the default template. More often than not you end up just mirroring your UI and server in 2 different directories and have to 2 work harder to keep them in sync. The team is looking for a way to keep behavior colocated.

Additionally, even if you have an `app/ui` directory it doesn't completely solve the issue. Some UI will be server-only, and some will make it to the browser. What we're trying to solve is:
- easy colocation of components with controllers
- good default way to signal which modules should be served through `remix/assets` without opting out of that colocation.

template/app/actions/public/prompt-button.tsx
```diff
@@ -28,11 +28,6 @@ export const PromptButton = clientEntry(
       return (
         <button
           type="button"
-          className={state}
```
brookslybrand, line 31: We don't use a css classnames at all in the template, so this was leftover (should have been `class` anyway).

Also, lots of inconsistencies in how the `css` mixins are declared and used in this file, so this PR also cleans that up

template/app/actions/document.tsx
```diff
@@ -20,14 +20,18 @@ export function Document(handle: Handle<DocumentProps>) {
...
+          <meta name="color-scheme" content="light dark" />
           <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
           <title>{title}</title>
+          {/* Page hints come before asset tags so render-blocking styles are discovered first. */}
           {head}
+          {entryPreloads.map((href) => (
+            <link key={href} rel="modulepreload" href={href} />
+          ))}
+          {/* Module scripts are deferred, so this runs after the document is parsed. */}
+          <script type="module" src={entryHref}></script>
```
brookslybrand, line 30: I don't know enough to know if this is the right move, from my agent (c.c. @markdalgleish):

>  I dropped `async` from the `<script type="module">` in the ui docs and did not use it in the template. Rationale: `run()` walks the document synchronously to collect client entries, and the `MutationObserver` in `runtime/frame.ts` only publishes late-arriving `<template>` frame content — not client-entry markers. With async in `<head>` during a streamed response, entries below the execution point can be missed. Bookstore's `document.tsx` uses `async` and has this same exposure — worth confirming before your demo-cleanup PR.

Happy to add `async` if that doesn't make sense

template/app/actions/document.tsx
```diff
@@ -20,14 +20,18 @@ export function Document(handle: Handle<DocumentProps>) {
...
         </head>
-        <body mix={css({ margin: 0 })}>
-          {children}
-          <script type="module" src={routes.assets.href({ path: 'app/assets/entry.ts' })}></script>
```
brookslybrand, line 29: This should not have been in the body 😅

template/app/actions/home-page.tsx
```diff
@@ -66,7 +66,6 @@ export function HomePage() {
 function HomeHead() {
   return () => (
     <>
-      <meta name="color-scheme" content="light dark" />
```
brookslybrand, line 69: Moved this up in to `document.tsx` where it belongs

packages/cli/src/lib/controller-ownership.ts
```diff
@@ -162,8 +171,15 @@ async function scanControllersDirectory(appRoot: string): Promise<ControllerDire
...
-        await walk(entryPath, false)
+        let entryIsBrowserSource = isBrowserSource || entry.name === BROWSER_SOURCE_DIRECTORY_NAME
+
+        // Still walk browser source so its files count as route-owned content, but never treat
+        // those directories, or anything below them, as route-map structure.
+        if (!entryIsBrowserSource) {
+          routeDirectoryPaths.add(relativePath)
+        }
+
+        await walk(entryPath, false, entryIsBrowserSource)
```
brookslybrand, line 182: I was wrong, `remix doctor` actually did need to be updated. Right now `remix doctor` warns on any directories in `app/actions/` that don't have a `controller.tsx`.

Personally I don't love this check, it feels overly restrictive to me. For example, what if I want to do:

```
app/
  actions/
    controller.tsx // imports from ./home/page
    public/
      entry.ts
      prompt-button.tsx
    home/
      header.tsx
      page.tsx
      footer.tsx
  routes.ts
```

This example is a little contrived, but I ran into a much more realistic example on `remix-website`. Basically we have a really complex top-level route. It has a bunch of different modules to help keep things clean. Either:

1. I have to move those outside of `app/actions` so `remix doctor` doesn't yell at me
2. or I have do an arbitrary nested of my route just so I can put stuff in a dir:
```tsx
home: route("/", {
  index: get("/"),
}),
```

So basically @mjackson, this is your call with `remix doctor`. The implementation I added here is fine and gets it working enough to special case `public` dirs, but overall I'm not sure this check is really doing a whole lot 🤷‍♂️. I mostly just don't want it to hold up the introduction of these `public/` dirs, because I think that's already a huge improvement over the top-level `assets/` and `ui/` conventions we had.

mjackson (dismissed):
I'm going to implement the `remix doctor` changes we discussed this morning. I'll finish this off. Thanks, @brookslybrand!

brookslybrand:
> I'm going to implement the `remix doctor` changes we discussed this morning. I'll finish this off. Thanks, @brookslybrand!

Sounds good, thanks Michael! Also thanks for going ahead and applying this to all of the demos. I originally was going to do that in a follow up, but happy to just get it all done now.

PR #11622, opened 2026-07-22
Add asset script transform hooks

### LINKED ISSUE #11580: Add an asset pipeline hook for script transforms
### What would help

Remix v3 currently has a good server-side extension point through the Node loader chain. That lets integrations transform server-loaded TypeScript/JavaScript modules before execution.

Browser-delivered modules go through the Remix asset pipeline instead. As far as I can tell in beta.5, there is no public script transform extension point there, so integrations that need to rewrite application source cannot support the browser/client side without wrapping `createAssetServer` or proxying in front of it.

A documented asset-pipeline transform hook would make that path much cleaner.

### Use cases

- compile i18n macros for browser-delivered app modules
- add framework-agnostic source transforms that need to run in both server and browser module graphs
- preserve source maps and watch invalidation without reimplementing Remix asset serving

### Shape that would be useful

- receives module URL/path, source, format, and current source map if one exists
- can return transformed source plus source map
- runs only for app/source modules, not arbitrary dependencies unless explicitly opted in
- participates in dev invalidation/watch behavior
- documents ordering relative to Remix TS/JSX lowering

For context: this came up while building a server-first Remix v3 i18n integration. The server side can work through `node --import remix/node-tsx --import ...`, but client-side macro parity is blocked until there is an asset-pipeline hook or an equivalent first-party API.

swernerx:
Adds a public, ordered script transform hook to `@remix-run/assets` so integrations can transform browser-delivered application source without wrapping or proxying the asset server.

- Adds `scripts.transforms`, which runs before Remix lowers TypeScript and JSX
- Provides each transform with the absolute file path, public URL pathname, source format, dependency status, and current source map
- Composes returned source maps with earlier transforms and the existing Oxc/minifier pipeline
- Registers returned `watchFiles` with the existing module cache and watcher invalidation flow
- Transforms symlinked npm, pnpm, and Yarn workspace packages by default while skipping installed or copied dependencies unless `includeDependencies` is enabled
- Documents the API and adds package/remix change notes

Closes #11580.

```ts
import { compileMessages } from './app/compile-messages.ts'

  basePath: '/assets',
  fileMap: {
    '/app/*path': 'app/*path',
    '/npm/*path': 'node_modules/*path',
  },
  allowFiles: ['app/assets/**'],
  allowPackages: ['remix'],
  sourceMaps: 'external',
  scripts: {
    transforms: [
      {
        name: 'messages',
        async transform(code, context) {
          if (context.format !== 'tsx') return null

          let result = await compileMessages(code, {
            filePath: context.filePath,
            format: context.format,
          })
          return {
            code: result.code,
            sourceMap: result.sourceMap,
            watchFiles: result.dependencies,
          }
        },
      },
    ],
  },
})
```


brookslybrand:
Hey @swernerx, thanks for showing interest in the project!

The asset server has seen a ton of changes, most notably the introduction of `scripts.loaders` in https://github.com/remix-run/remix/pull/11515

I'm going to close this PR for now. I see you have https://github.com/remix-run/remix/issues/11580 already open, so we can continue discussions there or you can open up a new PR on top of the latest work if you want to explore this idea further.

I'm also going to assign @markdalgleish to this, since he has been the one architecting the asset server.

PR #11629, opened 2026-07-23
Add SPA routing with custom fetch router outputs

brophdawg11:
`@remix-run/fetch-router` currently assumes every dispatch produces a `Response`, even when the router is used entirely in memory. This adds a configurable output contract and a `remix/ui/spa` runtime for building browser-only applications whose routers map URLs directly to rendered UI.

- Adds ambient `RouterTypes.output` configuration across routers, middleware, controllers, mounted routes, and nested `context.router.fetch` calls while retaining the existing `string | URL | Request` input contract.
- Preserves an internal `Request` for every dispatch, including propagation from `router.fetch(url, { signal })` to `context.request.signal`.
- Strongly types response-mutating middleware as requiring `Response` output and throws at runtime if downstream middleware returns another value.
- Exports `SPA` and the lower-level `createSPA` setup utility from `remix/ui/spa`. They intercept same-origin Navigation API navigations, cancel superseded loads, expose active and pending URLs, and respect `rmx-document` and `rmx-history`.
- Preserves form methods and forwards non-GET `FormData`. By default, same-URL submissions replace history while different-URL submissions push; `rmx-history` can override this behavior. History traversals revisit form destinations with `GET` because navigation entries do not retain submitted data.
- Adds a Vite demo in `demos/spa` showing route maps, controllers, render middleware, pending navigation UI, cancellation, and client-side form submissions.

```tsx
import { createRoot, type RemixNode } from 'remix/ui'
import { SPA } from 'remix/ui/spa'

declare module 'remix/router' {
  interface RouterTypes {
    output: RemixNode
  }
}

let router = createRouter({
  defaultHandler: () => <h1>Not Found</h1>,
})

router.get('/', () => <h1>Home</h1>)
router.get('/about', () => <h1>About</h1>)

let root = createRoot(document.getElementById('app')!)
root.render(<SPA router={router} fallback="Loading…" />)
```


demos/spa/app/actions/controller.tsx
```diff
@@ -0,0 +1,199 @@
+import { createController } from 'remix/router'
+import { css, type Handle } from 'remix/ui'
+
+import { routes } from '../routes.ts'
+
+export default createController(routes, {
+  actions: {
+    async home(context) {
+      await sleep(1000, context.request.signal)
+      return <HomePage />
```
brophdawg11, line 10: actions now return Remix nodes

demos/spa/app/actions/controller.tsx
```diff
@@ -0,0 +1,199 @@
...
+  return () => (
+    <article>
+      <p mix={eyebrowStyle}>Form submitted</p>
+      <h1 mix={titleStyle}>Hello, {handle.props.name}!</h1>
+      <p mix={bodyStyle}>
+        POST submissions expose the Navigation API's form data through{' '}
+        <code>context.request.formData()</code> without making an HTTP request. History traversals
+        return here with GET because navigation entries do not retain <code>FormData</code>.
+      </p>
+      <form method="POST" action={routes.greet.href()} mix={formStyle}>
```
brophdawg11, line 82: Form submissions are also handled by the SPA router component

demos/spa/app/ui/layout.tsx
```diff
@@ -0,0 +1,125 @@
...
+export function Fallback() {
+  return () => (
+    <Layout>
+      <LoadingPage />
+    </Layout>
+  )
+}
+
+export function Layout(handle: Handle<LayoutProps>) {
+  let router = handle.context.get(SPA)
```
brophdawg11, line 19: Components can access active/pending URLs from the SPA context

demos/spa/app/router.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import rootController, { NotFoundPage } from './actions/controller.tsx'
+import { render } from './middleware/render.tsx'
+import { routes } from './routes.ts'
+
+declare module 'remix/router' {
+  interface RouterTypes {
+    output: RemixNode
+  }
+}
```
brophdawg11, line 12: This is the important part to enable returning `RemixNode`'s from the router

PR #11632, opened 2026-07-23
Support form-driven frame navigation

brophdawg11:
See the demo in `demos/frame-navigation` for live examples
 - the courses page has a GET form for filtering the list
 - the account page has a POST form to update your account info
 - both use a targeted reload of their own frame

This adds progressive enhancement for form submissions to Remix UI frame navigation, allowing same-origin forms to update frames through the same client runtime used by links while preserving native document submissions when JavaScript is unavailable.

- Changes the client `resolveFrame` signature from `resolveFrame(src, signal, target)` to `resolveFrame(src, options)`, where `options` may include `formData`, `method`, `encType`, `signal`, and `target`.
- Intercepts eligible form submissions after native constraint validation and the form's `submit` event, including submitter overrides such as `formmethod`, `formenctype`, and `formtarget`.
- Supports `rmx-target` for selecting a named frame and `rmx-document` for opting back into a document navigation.
- Leaves cross-origin submissions, `method="dialog"`, and `target="_blank"` to the browser.
- Pushes URL-backed history entries for GET submissions and replaces the current entry for non-GET submissions, avoiding duplicate back-button entries without retaining submitted form data in history.
- Preserves streaming responses by allowing resolvers to return `response.body`.
- Extends the frame-navigation demo with GET course filtering and POST account editing.

```tsx
<form method="post" action="/account" rmx-target="main">
  <label for="display-name">Display name</label>
  <input id="display-name" name="displayName" required />
  <button type="submit">Save</button>
</form>
```

```ts
run({
  async resolveFrame(src, options) {
    let response = await fetch(src, {
      method: options?.method,
      body: options?.method?.toLowerCase() === 'post' ? options.formData : undefined,
      signal: options?.signal,
    })
    return response.body ?? (await response.text())
  },
})
```


packages/ui/src/runtime/navigation.ts
```diff
@@ -57,11 +97,55 @@ export function startNavigationListenerImpl(
   },
 ) {
   let navigation = window.navigation
+  // Native form navigation spans three events: `submit` exposes the submitter,
+  // `formdata` exposes the browser-built entry list, and `navigate` lets us intercept.
+  // Chromium fires `navigate` before `formdata`, so keep per-form state here until
+  // the navigation handler can consume it.
+  let pendingFormSubmissions = new WeakMap<HTMLFormElement, PendingFormSubmission>()
```
brophdawg11, line 104: This is a bit weird because we have to handle a few different events which can happen in different orders across browsers

packages/ui/src/runtime/navigation.ts
```diff
@@ -71,28 +155,65 @@ export function startNavigationListenerImpl(
...
+              controller.redirect(event.destination.url, { history: 'replace' })
+            },
           }
-        },
-      })
+          event.intercept(interceptOptions)
+          return
+        }
+
+        if (event.cancelable && runtimeNavigation.getReloadOptions) {
```
brophdawg11, line 201: Safari doesn't support `preocmmitHandler` so we need to manually cancel the current navigation and start a new replace `navigation`

brophdawg11:
Closing this combined draft so the redirect and form-submission changes can be reviewed as separate stacked pull requests.

brophdawg11:
Split into #11667 for frame reload redirects and stacked follow-up #11668 for frame form submissions.

PR #11635, opened 2026-07-23
Fix compression cache variation for identity responses

brookslybrand:
I found this while investigating why [remix.run](https://remix.run) is not serving its JavaScript assets with gzip. With a cache-busted asset URL, requesting the identity representation first produced a cache miss without `Content-Encoding` or `Vary`; a subsequent gzip-capable request was then a cache hit for that same uncompressed response. Requesting gzip first behaved correctly because compressed responses already included `Vary: Accept-Encoding`.

`compressResponse()` previously added `Vary: Accept-Encoding` only after selecting a compressed representation. This moves the cache variation metadata ahead of content-coding negotiation so an eligible identity response cannot seed a shared cache entry that is reused for compression-capable clients.

- add and merge `Vary: Accept-Encoding` for eligible identity responses, requests without `Accept-Encoding`, compressed responses, and negotiation failures
- preserve identity bodies, status metadata, content length, range metadata, and ETags without applying compression-specific transformations
- leave filtered media types, `no-transform`, existing encodings, range/partial responses, known sub-threshold responses, empty responses, and disabled encodings unchanged
- cover the identity-first shared-cache sequence in both the response helper and compression middleware

### DIFF
--- a/packages/compression-middleware/README.md
+++ b/packages/compression-middleware/README.md
@@ -26,8 +26,6 @@ let router = createRouter({
 })
 ```
 
-For responses eligible for compression, the middleware adds `Vary: Accept-Encoding` even when the client requests the identity representation or omits `Accept-Encoding`. This keeps compressed and uncompressed representations safe in shared caches while preserving existing `Vary` values.
-
 The middleware will automatically compress responses for compressible MIME types when:
 
 - The client supports compression (`Accept-Encoding` header with a supported encoding)
--- a/packages/compression-middleware/README.md
+++ b/packages/compression-middleware/README.md
@@ -26,6 +26,8 @@ let router = createRouter({
 })
 ```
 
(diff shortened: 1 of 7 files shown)


packages/compression-middleware/README.md
```diff
@@ -26,6 +26,8 @@ let router = createRouter({
 })
 ```
 
+For responses eligible for compression, the middleware adds `Vary: Accept-Encoding` even when the client requests the identity representation or omits `Accept-Encoding`. This keeps compressed and uncompressed representations safe in shared caches while preserving existing `Vary` values.
```
markdalgleish: I don't think this needs to be called out in the readme.

markdalgleish (approved):
Just a comment about the readme, but looks good to me. Thanks for catching this.

PR #11660, opened 2026-07-30
Validate cookie session data

mjackson:
Cookie session storage previously trusted any successfully parsed JSON through a type assertion. Valid JSON with the wrong session ID or data shape could therefore reach the session runtime.

Validate parsed cookie payloads with `@remix-run/data-schema` before restoring a session. Malformed payloads now follow the existing invalid-cookie behavior and start a fresh session.

### DIFF
--- a/packages/session/.changes/patch.validate-cookie-session-data.md
+++ b/packages/session/.changes/patch.validate-cookie-session-data.md
@@ -1,0 +1,1 @@
+Validate parsed cookie session data before restoring a session, treating malformed data as a new session.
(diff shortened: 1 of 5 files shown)


mjackson:
Closing this for now. The normal session middleware path only passes authenticated cookie values to cookie session storage, so clients cannot manufacture malformed session data without the signing secret. If runtime validation of persisted session data becomes necessary, we should address it consistently across cookie, filesystem, Redis, and Memcache storage rather than only validating this backend.

PR #11662, opened 2026-07-30
Remove addEventListeners from Remix UI

brookslybrand:
`addEventListeners()` is a convenience wrapper around native `EventTarget` APIs rather than a core part of Remix UI. The reason it was initially added is because the original events (`remix/events` ) packages used that api pretty much as the foundation, and we still had the `on` intrinsic prop so you could do things that all looked the same:

```
<button on={{ click: fn }} />
addEventListeners(thing, { change: fn }})
```

When we moved from `remix/component` and `remix/events` to the unified `remix/ui` with the `mix` prop, `addEventListeners` just stayed around as a convenience layer. After talking to the team, we don't believe it makes sense to continue to support it (and fix a number of issues, such as incorrect SSR implementation/behavior) when good enough browser APIs already exist.

This removes the API and migrates Remix code, demos, documentation, and agent guidance to native event listeners.

- Use `target.addEventListener(type, listener, { signal })` with the appropriate lifetime signal for automatic cleanup
- Document how to replace the helper's reentry cancellation with an `AbortController` when needed
- Prefer element-scoped listeners and reserve `window` or `document` listeners for genuinely global events
- Preserve the event types required by `on()` and the public `Dispatched` type


mjackson (approved):
Good call on this. Less api is always better

PR #11665, opened 2026-07-31
Fall back to document navigation without the Navigation API

### LINKED ISSUE #11641: Bug: client runtime reads `window.navigation` unguarded, so hydration dies on browsers without the Navigation API
**Versions:** `remix@3.0.0-beta.5`, `@remix-run/ui@0.4.0`

## What happens

`run()` calls `startNavigationListener()`, which reads `window.navigation` with no feature check:

```ts
// packages/ui/src/runtime/navigation.ts
export function startNavigationListenerImpl(signal, options) {
  let navigation = window.navigation

  navigation.updateCurrentEntry({
    state: { target: undefined, src: window.location.href, resetScroll: true, $rmx: true },
  })
  ...
}
```

On a browser without the Navigation API this throws synchronously:

```
TypeError: Cannot read properties of undefined (reading 'updateCurrentEntry')   // Chromium wording
TypeError: can't access property "updateCurrentEntry", r is undefined            // Firefox wording
TypeError: undefined is not an object (evaluating 'r.updateCurrentEntry')        // WebKit wording
```

Because `createFrame(...)` runs just before it and `topFrame.ready()` just after, the throw aborts `run()` partway: **the app never hydrates at all**. No router, no event handlers, no `AppRuntime` returned, so the caller's `app.addEventListener('error', ...)` is never even registered and the failure is invisible to app-level error handling. `window.navigation` is also read unguarded in `navigate()` and `getTraverseNavigationState()`.

## Who this affects

The Navigation API is very new outside Chromium — Firefox shipped it in **147** and WebKit in **26.2**, both in early 2026. So today this is:

- Firefox **146 and below**
- **every** browser on iOS **26.1 and below** (all of them WebKit, so Chrome for iOS and in-app browsers too)

We caught this in production error reporting with real events from Firefox 141 and from iOS 26.1 (Chrome Mobile iOS and the X in-app browser). Note that the event counts badly understate the impact: an error thrown during hydration also breaks the page that would have reported it.

## Reproduction

Any Remix 3 app in a browser without the Navigation API. To reproduce in Chromium, shadow the getter before the client entry runs:

```js
Object.defineProperty(window, 'navigation', { value: undefined, configurable: true })
```

Load the page: `run()` throws and nothing hydrates.

## Suggested fix

Feature-detect and degrade instead of throwing. The listener only exists to intercept `navigate` events, so where there is no Navigation API there is nothing to intercept and links can simply fall through to ordinary document navigation:

```ts
export function startNavigationListenerImpl(signal, options) {
  let navigation = window.navigation
  if (!navigation) return
  ...
}
```

The exported `navigate()` helper could fall back to `location.assign` / `location.replace` for the same reason.

## Workaround

We are shipping a small stand-in installed before `run()` when `window.navigation` is missing, implementing only the members the runtime reads: https://github.com/kentcdodds/kody/pull/921

_Filed by an agent on behalf of the [kentcdodds/kody](https://github.com/kentcdodds/kody) project._

okxint:
Browsers without the Navigation API currently crash the frame runtime when it reads `window.navigation`.

- Fall back to full-document navigation, preserving push and replace history semantics.
- Skip Navigation API listener setup and traversal-state lookup when the API is unavailable.
- Cover unsupported-browser fallback behavior with focused browser tests.


packages/ui/.changes/patch.navigation-api-fallback.md
```diff
@@ -0,0 +1 @@
+Fall back to document navigation instead of crashing the client runtime in browsers that do not support the Navigation API (see #11665).
```
brophdawg11:
```suggestion
Gracefully degrade to document navigations for browsers that do not support the Navigation API (see #11665)
```

brophdawg11:
Thanks!

PR #11666, opened 2026-08-03
Use stable TypeScript 7 compiler

brookslybrand:
TypeScript 7 is now available as a stable release, so the repository no longer needs the `@typescript/native-preview` package or `tsgo` commands introduced in [#10863](https://github.com/remix-run/remix/pull/10863). This moves builds and type checks to the official `typescript@7.0.2` package and `tsc` command.

- Replace the native preview dependency throughout the workspace and remove the legacy `typescript-language-server` integration.
- Declare ambient `types` and build `rootDir` values explicitly where required by TypeScript 7.
- Address stricter stream-result narrowing, route inference, and portable declaration emit checks.
- Pin TypeScript 6.0.3 only for the docs tooling and route-pattern type benchmarks that still depend on the legacy compiler API. Their regular type checks continue to use the root TypeScript 7 compiler where applicable.

### DIFF
--- a/docs/api/app/router.ts
+++ b/docs/api/app/router.ts
@@ -73,8 +73,12 @@ export function createApiRouter(options: ApiRouterOptions) {
     let versionApiController = createApiController(versionControllerOptions)
 
     // A mounted route builder does not match its own root path, so map the
-    // versioned home action on the parent router.
-    router.map(`${versionPathname}/`, versionRootController.actions.home)
+    // versioned home action on the parent router. Wrap the route-specific action so TypeScript 7
+    // can infer context for dynamic version paths.
+    let homeAction = versionRootController.actions.home
+    router.map(`${versionPathname}/`, (context) =>
+      typeof homeAction === 'function' ? homeAction(context) : homeAction.handler(context),
+    )
     router.mount(versionPathname, (versionRouter) => {
       versionRouter.map(routes.assets, versionRootController.actions.assets)
       versionRouter.map(routes.lookup, versionRootController.actions.lookup)
--- a/packages/auth-middleware/src/lib/auth.ts
+++ b/packages/auth-middleware/src/lib/auth.ts
@@ -45,7 +45,9 @@ export type AuthState<identity = unknown> = GoodAuth<identity> | BadAuth
  * Context key used to read auth state with `context.get(Auth)`.
  * The `auth()` middleware also installs auth state as `context.auth`.
  */
-export const Auth = createContextKey<AuthState>()
+// Keep exported context keys structurally typed so TypeScript 7 declaration emit does not
+// reference fetch-router internals. Logger and Renderer require the same annotation.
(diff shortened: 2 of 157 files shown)


packages/auth-middleware/src/lib/auth.ts
```diff
@@ -45,7 +45,9 @@ export type AuthState<identity = unknown> = GoodAuth<identity> | BadAuth
  * Context key used to read auth state with `context.get(Auth)`.
  * The `auth()` middleware also installs auth state as `context.auth`.
  */
-export const Auth = createContextKey<AuthState>()
+// Keep exported context keys structurally typed so TypeScript 7 declaration emit does not
+// reference fetch-router internals. Logger and Renderer require the same annotation.
+export const Auth: { defaultValue?: AuthState } = createContextKey<AuthState>()
```
brookslybrand, line 50: TypeScript 7 declaration emit cannot name the inferred `ContextKey` without referring to fetch-router’s internal source path. This structural annotation is exactly the public shape of that type and avoids expanding fetch-router’s API. The same fix is applied to `Logger` and `Renderer` in their middleware packages.

packages/response/src/lib/html.ts
```diff
@@ -83,9 +83,11 @@ function prependDoctypeToStream(stream: ReadableStream<Uint8Array>): ReadableStr
 
         // Pass through remaining chunks
         while (true) {
-          let { done, value } = await reader.read()
-          if (done) break
-          controller.enqueue(value)
+          // Keep the result intact so TypeScript 7 can narrow value from done. The abortable UI
+          // stream uses the same pattern.
+          let result = await reader.read()
```
brookslybrand, line 88: Keeping the `ReadableStreamReadResult` object intact preserves the discriminated relationship between `done` and `value`, allowing TypeScript 7 to prove that `value` is defined here. The same narrowing adjustment is made in `packages/ui/src/runtime/reconcile.ts` for the abortable stream.

docs/api/app/router.ts
```diff
@@ -73,8 +73,12 @@ export function createApiRouter(options: ApiRouterOptions) {
     let versionApiController = createApiController(versionControllerOptions)
 
     // A mounted route builder does not match its own root path, so map the
-    // versioned home action on the parent router.
-    router.map(`${versionPathname}/`, versionRootController.actions.home)
+    // versioned home action on the parent router. Wrap the route-specific action so TypeScript 7
+    // can infer context for dynamic version paths.
+    let homeAction = versionRootController.actions.home
+    router.map(`${versionPathname}/`, (context) =>
```
brookslybrand, line 79: TypeScript 7 rejects passing the `/`-specific controller action directly to this dynamically constructed version route. Wrapping it lets `router.map` infer the versioned context, then delegates to the existing function/action-object shape. This is the only route-inference adjustment in this migration.

brophdawg11 (approved):
LGTM!

PR #11667, opened 2026-08-05
Support redirects from frame reloads

brophdawg11:
Frame resolvers currently have to unwrap fetched responses into content, which hides whether `fetch()` followed a redirect. This allows resolvers to return the `Response` directly so the frame runtime can preserve streaming and use redirect metadata for navigation-driven reloads.

- Allows browser frame resolvers to return fetched `Response` objects.
- Streams response bodies and internally tracks final response URLs for navigation-driven reloads without changing `frame.reload()`'s public return type.
- Replaces the browser URL and top frame's canonical `src` when a top-frame navigation follows a redirect.
- Direct reloads and named-frame navigations render redirected responses without replacing their canonical `src` with the final response URL; applications can inspect `Response.redirected` and `Response.url` inside `resolveFrame` when custom behavior is needed.
- Demonstrates the behavior with a `303` redirect from `/settings` to `/settings/overview` in the frame-navigation demo.
- Supersedes the redirect portion of #11632; form submissions are split into stacked follow-up #11668.

```ts
run({
  async resolveFrame(src, signal, target) {
    let headers = new Headers({ Accept: 'text/html' })
    if (target) headers.set('X-Remix-Target', target)
    return fetch(src, { headers, signal })
  },
})
```


demos/frame-navigation/app/assets/entry.tsx
```diff
@@ -54,8 +54,7 @@ async function resolveFrameResponse(
     )
   }
 
-  if (res.body) return res.body
-  return await res.text()
+  return res
```
brophdawg11, line 57: I'm on the fence about whether we want to own the response unwrapping.  It may make more sense to leave this all in userland `resolveFrame` code and let them return something akin to our internal `unwrapFrameResolution` utility - and then we would only own the navigation API touchpoints.

```ts
function resolveFrame(src, signal, target) {
  let url = new URL(src, window.location.href)
  let res = await fetch(url, ...);
  let content = res.body ? res.body : await res.text();
  if (res.redirected && res.url) {
    return { content, redirectTo: res.url };
  }
  return content;
}
```

demos/frame-navigation/app/middleware/auth.ts
```diff
@@ -36,8 +36,10 @@ export function loadAuth() {
 
 export const requireAuth = requireAuthenticated<FrameAuthIdentity>({
   onFailure(context) {
-    let isFrameRequest = context.request.headers.get('X-Remix-Frame') === 'true'
-    if (isFrameRequest) {
+    let isSubFrameRequest =
+      context.request.headers.get('X-Remix-Frame') === 'true' &&
+      context.request.headers.get('X-Remix-Target') != null
+    if (isSubFrameRequest) {
```
brophdawg11, line 42: Sub-frames are trickier so we keep them on the old paradigm.  We can't return full top frame content from the auth page and insert it into the settings frame.  I'm sure this is somewhat solvable by returning metadata that tells the browser to instead replace the top level frame content, but might not be worth the complexity. Happy to give it a go if we want.

packages/ui/src/runtime/navigation.ts
```diff
@@ -84,9 +95,22 @@ export function startNavigationListenerImpl(
...
+            let redirectedSrc = frame.src
+            let redirectedState = { ...state, src: redirectedSrc }
+            // Start the successor navigation without awaiting it: this handler must settle before
+            // the replacement navigation can finish.
+            navigation.navigate(redirectedSrc, {
+              history: 'replace',
+              state: redirectedState,
+              info: { type: frameRedirectNavigationInfoType } satisfies FrameRedirectNavigationInfo,
+            })
+          }
```
brophdawg11, line 112: If a navigational frame reload followed a redirect, replace the current history entry with the redirected location.

daltino (commented):
The simplification in `entry.tsx` where `return res` replaces the old `if (res.body) return res.body; return await res.text()` is clean, but it's worth noting that this silently changes behavior for non redirected responses too. Previously the body was extracted eagerly; now the `Response` object is returned in all cases and the runtime handles streaming. That's fine as long as the runtime correctly handles a `Response` whose `redirected` property is `false`, but it's a behavioral shift worth calling out explicitly in the PR description or a comment since it affects every frame load, not just redirected ones. The auth middleware change from `isFrameRequest` to `isSubFrameRequest` is interesting. The condition now requires both the `X Remix Frame` header AND `X Remix Target` to be present before returning the inline error HTML. This means a top frame request (which presumably lacks `X Remix Target`) will fall through to whatever the default failure behavior is. It would be worth a comment explaining why top frames should be excluded here, since that reasoning isn't immediately obvious from the variable name alone.

PR #11668, opened 2026-08-05
Support form-driven frame navigation

brophdawg11:
This extends frame-aware navigation to eligible same-origin forms while preserving native form behavior before the client runtime loads. Applications continue to own request construction in `resolveFrame`; the runtime only forwards the browser's submission metadata.

- Changes `resolveFrame(src, signal, target)` to `resolveFrame(src, options)` with `formData`, `method`, `encType`, `signal`, and `target`.
- Preserves native constraint validation, submitter overrides, progressive enhancement, and explicit document-navigation escape hatches.
- Uses replace history for same-location non-GET submissions while preserving push history for different destinations and GET submissions.
- Demonstrates GET course filtering and POST account editing through the frame-navigation demo's new `/account/edit` route.
- Stacked on #11667 and supersedes the form-submission portion of #11632.

```tsx
<form method="post" action="/account/edit" rmx-target="account">
  <label for="display-name">Display name</label>
  <input id="display-name" name="displayName" required />
  <button type="submit">Save</button>
</form>
```

```ts
run({
  async resolveFrame(src, options) {
    return fetch(src, {
      method: options?.method,
      body: options?.method?.toLowerCase() === 'post' ? options.formData : undefined,
      signal: options?.signal,
    })
  },
})
```


daltino (commented):
The `getRequestBody` helper in the bookstore demo has a subtle behavior for file inputs: when the encType is `application/x www form urlencoded`, it falls back to `value.name` for `File` entries, which is the filename rather than the file contents. That's technically correct per spec since `urlencoded` can't encode binary data, but it silently drops the actual file bytes rather than throwing or warning. Worth a comment explaining that choice so future readers don't assume it's a bug. Also worth double checking the `getRequestBody` logic for the GET case across both implementations. In the reference doc snippet, GET submissions return `undefined` for the body (correct), but the `src` URL passed to `resolveFrame` would need the form fields appended as query params to actually work. If the runtime is forwarding `formData` for GET forms but not merging it into the URL, those submissions will silently drop all the form fields entirely.

PR #11670, opened 2026-08-05
Add rmx-history navigation attribute

brophdawg11:
This adds an explicit history override for progressively enhanced frame navigations. It is stacked on #11668, which adds form-driven frame navigation.

- Adds `rmx-history="push|replace"` support to enhanced anchors and forms.
- Allows forms to override the automatic replacement used for same-location non-GET submissions.
- Makes `link(href, { history })` emit the corresponding `rmx-history` value on native anchors.
- Preserves replacement behavior across frame redirects and fallback navigation replay.

```tsx
<a href="/search" rmx-history="replace">Search</a>

<form method="post" rmx-history="push">
  <button>Save as a new history entry</button>
</form>
```

### DIFF
--- a/.agents/skills/remix/references/hydration-frames-navigation.md
+++ b/.agents/skills/remix/references/hydration-frames-navigation.md
@@ -214,11 +214,12 @@ When `run({ resolveFrame })` is active, eligible same-origin forms progressively
 - Forms target `handle.frames.top` by default.
 - `rmx-target` selects a named frame.
 - `rmx-src` selects a different frame request URL while preserving the form action as the navigation destination.
+- `rmx-history="push|replace"` overrides how the navigation updates history.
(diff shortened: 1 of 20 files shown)


daltino (commented):
The e2e test file is a welcome addition, but I'd want to see test coverage that specifically exercises the `rmx-history` attribute itself, the snippet cuts off before I can verify it's there, but the filter form getting `rmx-history="replace"` in the demo without a corresponding assertion in the test suite would be a gap worth closing.

Also worth double checking: when a frame redirect happens mid-navigation (the PR description mentions "preserves replacement behavior across frame redirects"), is there a test that confirms the history entry count doesn't grow unexpectedly? That's exactly the kind of subtle regression that's hard to catch manually and easy to miss when the happy-path tests all pass.

PR #11671, opened 2026-08-06
Fix top frame reloads after navigation targets a named frame

markdalgleish:
While testing HMR in the `remix-website` repo, I hit an issue where a top frame reload used the wrong URL after navigation targeted a named frame.

The app uses `/jam/2026/ticket` to represent an open ticket modal, using `<a href="/jam/2026/ticket" rmx-target="tickets">` so client navigation only reloads the modal frame. The browser URL updated correctly, but `handle.frames.top.src` remained at `/jam/2026/`.

This became visible when the HMR `server:update` handler later reloaded the top frame. It fetched the old page, closed the modal and left the browser URL at `/jam/2026/ticket`. HMR exposed the issue, but the underlying bug applies to any later top frame reload.

The core issue is that the navigation handler only updated the source of the frame being reloaded. When navigation targeted a named frame, the top frame was left pointing at the previous URL.

To fix this, this PR always updates the top frame source to the public destination while preserving the named frame's own source. The navigation still only reloads the targeted frame. I've added some more test coverage to handle the fix and other related navigation scenarios.


brookslybrand (approved):
Tested with the frames navigation demo and this fix gives the expected experience imo

A JSX representation, after navigating:

 ```tsx
   // Browser URL: /settings/profile

   <TopFrame src="/settings">
     <Layout title="Settings">
       <RefreshPageButton />

       <Frame name="settings" src="/settings/profile">
         <SettingsLayout activeItem="profile">
           <Profile />
         </SettingsLayout>
       </Frame>
     </Layout>
   </TopFrame>
 ```
 
If you refresh the page or navigate to `/settings/profile` directly, your representation is this:
 
  ```tsx
   // Browser URL: /settings/profile

   <TopFrame src="/settings/profile">
     <Layout title="Settings">
       <RefreshPageButton />

       <Frame name="settings" src="/settings/profile">
         <SettingsLayout activeItem="profile">
           <Profile />
         </SettingsLayout>
       </Frame>
     </Layout>
   </TopFrame>
 ```
 
 The discrepancy is definitely pretty confusing. 

https://github.com/user-attachments/assets/855703df-3efe-4e0e-b7c7-d18e2dbaff47

PR #11679, opened 2026-08-10
Add mount-scoped middleware to fetch router

brophdawg11:
Controller middleware applies only to actions directly owned by that controller. This adds an explicit mount-scoped middleware boundary so shared policy continues across multiple controllers and nested mounts.

- Adds a backwards-compatible `router.mount(prefix, { middleware }, installer)` overload
- Runs middleware in router, outer mount, inner mount, controller, then action order
- Runs mount middleware only after a descendant route matches, preserving fallthrough for unknown paths
- Infers middleware-provided context and mount params in descendant inline actions and controllers
- Includes a focused parent/child/grandchild demo of controller-local and mount-wide middleware execution

A route refactor can accidentally remove middleware without changing the URL:

```ts
// Before: /admin/projects runs requireAdmin()
projects: get('projects')

createController(routes.admin, {
  middleware: [requireAdmin()],
  actions: { projects: listProjects },
})

// After: same URL, new controller, no requireAdmin()
projects: route('projects', { index: get('/') })

createController(routes.admin.projects, {
  actions: { index: listProjects },
})
```

Mount middleware keeps the boundary intact across the refactor:

```ts
router.mount('/admin', { middleware: [requireAdmin()] }, (admin) => {
  admin.map(adminRoutes, adminController)
  admin.map(adminRoutes.projects, projectsController)
})
```


brophdawg11:
Closing for now to see if this is actually a pain point after launch

PR #11687, opened 2026-08-12
Add client-rendered SPA routing

brophdawg11:
Replaces https://github.com/remix-run/remix/pull/11629

Adds a first-class client-rendered SPA boundary without changing the fetch router's standard `Request` to `Response` contract.

- Add `@remix-run/spa` and the `remix/spa` entrypoint with two application-facing APIs: `render()` installs `context.render()` middleware, and `run(router, { fallback? })` connects an ordinary fetch router to the browser UI runtime.
- Keep the in-memory SPA response carrier behind the render middleware. It is browser-only, lazily creates its `WeakMap`, and throws if used in a non-browser environment.
- Reuse top-frame navigation for the initial route, links, forms, history traversal, cancellation, named-frame targeting, and same-origin redirects with Fetch method semantics and a 10-redirect limit.
- Make `app.ready()` include the initial top-frame reload, with an optional live Remix node fallback displayed while that route resolves.
- Add package docs, release notes, browser/server coverage, and a Vite SPA demo covering deep links, superseded navigations, contextual form pending UI, and push/replace history behavior.

```tsx
import { render, run } from 'remix/spa'

const router = createRouter({
  middleware: [render()],
})

router.get('/', ({ render }) => render(<Home />))
router.get('/about', ({ render }) => render(<About />))

const app = run(router, { fallback: <Loading /> })
await app.ready()
```


packages/spa/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "@remix-run/assert": "workspace:^",
+    "@remix-run/test": "workspace:^",
+    "@types/dom-navigation": "^1.0.7",
+    "@types/node": "catalog:",
+    "typescript": "catalog:"
+  },
+  "dependencies": {
+    "@remix-run/fetch-router": "workspace:^",
+    "@remix-run/render-middleware": "workspace:^",
+    "@remix-run/ui": "workspace:^"
```
brophdawg11, line 45: Added this new approach to a new package because it depends on both `fetch-router` and `ui` so it can provide the SPA-enabled render middleware that abstracts away the `spaResponse` concept.  This avoids adding `fetch-router` as a dependency of `ui` if we had tried to leave this in `ui/spa`.

demos/spa/app/main.tsx
```diff
@@ -0,0 +1,71 @@
+import { createRouter, type Middleware } from 'remix/router'
+import { render, run } from 'remix/spa'
```
brophdawg11, line 2: `remix/spa` exposes 2 primary APIs

- a `render` middleware that handles the `spaResponse`/`WeakMap` stuff for you
- a `run` method that wraps the normal run method, provides a SPA implementation of `resolveFrame`, and supports rendering a fallback component in the top frame while the initial URL is loading.

packages/spa/src/lib/spa.ts
```diff
@@ -0,0 +1,182 @@
...
+ * @returns Middleware that installs the SPA renderer on request context.
+ */
+export function render(transform?: RenderTransform): RenderMiddleware {
+  return renderWith(
+    (context) =>
+      function render(node: RemixNode, init?: ResponseInit): Response {
+        return spaResponse.create(transform ? transform(node, context) : node, init)
+      },
+  )
+}
```
brophdawg11, line 74: Render middleware allowing SPA handlers to `return render(<h1>Remix</h1>)`

packages/spa/src/lib/spa.ts
```diff
@@ -0,0 +1,182 @@
...
+      let method = options?.method?.toUpperCase() ?? 'GET'
+      let body = ['GET', 'HEAD'].includes(method) ? undefined : getRequestBody(options)
+      let { response, redirectedTo } = await followFrameRedirects(router, url, {
+        method,
+        body,
+        signal: options?.signal,
+      })
+
+      return spaResponse.finalize(response, redirectedTo)
+    },
```
brophdawg11, line 100: A SPA resolveFrame looks very similar to a normal resolveFrame but it makes it's fetches through the client side `router.fetch` and therefore has to manually follow redirects.  `spaRepsonse.finalize` is what updates the `WeakMap` info with any final redirect locations so we can access that from the frame runtime

packages/ui/src/runtime/mixins/mixin.ts
```diff
@@ -107,6 +107,15 @@ type MixinRuntimeType<
   type: string,
 ) => ((...args: [...args, currentProps: props]) => MixinReturn<node, props>) | void
 
+type MixinDescriptorType<
```
brophdawg11, line 110: Something about `export * from '@remix-run/spa'` in the `remix` package surfaced a TS issue that I need to dig a bit deeper into, this was the agents fix for now

brophdawg11: I opened https://github.com/remix-run/remix/pull/11708 to reproduce and fix this issue since it's unrelated to the work in this PR

ryanflorence (approved):

PR #11693, opened 2026-08-13
Add a default browser frame resolver

brophdawg11:
Remix UI now provides a default browser frame resolver, so apps using `run()` no longer need to repeat the standard `fetch()` boilerplate.

- Enable frame reloads and enhanced link and form navigation when a custom `resolveFrame` is omitted
- Fetch frame HTML with `Accept: text/html`, the submitted method, and the abort signal, using native request bodies for URL-encoded, plain-text, and multipart form submissions
- Reject non-OK responses without replacing the current frame UI; custom resolvers may return non-OK responses when they want to render application-specific error UI
- Leave individual link and form navigations to the browser with `rmx-document`

Apps can now start the runtime without supplying a frame resolver:

```js
let app = run({
  async loadModule(moduleUrl, exportName) {
    let mod = await import(moduleUrl)
    return mod[exportName]
  },
})

await app.ready()
```

Apps can still provide `resolveFrame` when they need custom headers, request encoding, redirects, or error presentation. Returning a `Response` from a custom resolver streams its body regardless of status.


packages/ui/src/runtime/frame-resolution.ts
```diff
@@ -5,6 +5,13 @@ export async function unwrapFrameResolution(
 ): Promise<{ content: FrameContent; redirectedTo?: string }> {
   if (!(resolution instanceof Response)) return { content: resolution }
 
+  // Allow 4xx responses to provide frame UI by default.
+  if (resolution.status >= 500) {
+    throw new Error(
+      `Failed to resolve frame: ${resolution.status} ${resolution.statusText}`.trimEnd(),
+    )
+  }
```
brophdawg11: Originally this was `is(!response.ok)` but that doesn't let you return UI with. 404, etc.  This felt safer

PR #11706, opened 2026-08-17
Use import maps for browser assets

markdalgleish:
This updates `remix/assets` to preserve authored JavaScript imports and resolve them through import maps. This decouples each module's fingerprint from its dependencies, preventing changes from cascading through importer hashes and allowing unchanged modules to remain cached across deployments.

Asset fingerprints are now based on the final emitted bytes, so fingerprinting is now a boolean option rather than requiring a `buildId` that invalidates the entire build.

The `resolveClientEntry` API now also supports returning an `importMap`. The asset server now provides a unified `getScriptEntry` API that returns everything needed to load a browser entry:

```ts
let { href, importMap, preloads } = await assetServer.getScriptEntry(entryId)
```

Note: This PR is built on top of these other PRs that have been split out to ease the review process:

- ✅ <strike>[Preserve script text content during SSR ](https://github.com/remix-run/remix/pull/11722)</strike>
- ✅ <strike>[Replace `fileMap` with `mounts`](https://github.com/remix-run/remix/pull/11730)</strike>


brookslybrand:
Experimenting on `remix-website` staging

Left is current remix.run, right is staging.

Not in the image is document size:
- remix.run document: 15.4 kB transferred, 92.8 kB resource size
- staging document: 13.9 kB transferred, 131 kB resource size

So it definitely does compress well

Screenshot 2026-08-27 at 10 24 07 AM

From that alone, unless I messed something up, this is a win. Of course that's in addition to:
- fine grain JS cache busting
- clearly readable imports
- probably other stuff I don't realize

image

mjackson:
This change relies on multiple/incremental import maps, so merging it effectively drops default Firefox support for these asset and client-entry flows for now.

Firefox support is tracked in [Bug 1916277](https://bugzilla.mozilla.org/show_bug.cgi?id=1916277). Until it is enabled by default, Firefox users need to set `dom.multiple_import_maps.enabled` to `true` in `about:config` to use multiple import maps.

mjackson:
Just following up from above: we can probably recommend people use this if they need to support Firefox (until it's available without a flag): https://www.npmjs.com/package/es-module-shims

PR #11708, opened 2026-08-17
Fix mixin input argument types

brophdawg11:
> [!CAUTION]
> This is not intended to be merged as is.  This PR just contains a minimal repro of the underlying issue (efa4869) by creating a new dummy package on top of `@remix-run/ui`.  That commit should be reverted and only the fix (4327720) should land in `main`

This fixes a latent incompatibility between argument-bearing mixin descriptors and the general `mix` prop input type. It surfaced while adding the `remix/spa` package export in #11687: the additional package traversal caused the Remix umbrella typecheck to report existing `CSSMixinDescriptor` usage errors in UI components even though the UI package typecheck passed independently.

- Adds a minimal package-graph reproduction where `@remix-run/mixin-repro` consumes a new UI root export and is re-exported through `remix/mixin-repro`. The first commit demonstrates that `pnpm --filter @remix-run/ui run typecheck` passes while `pnpm --filter remix run typecheck` fails in the accordion component.
- Separates the descriptor's precise callable signature from the permissive shape accepted by `MixInput` and `MixValue`. This avoids treating `MixinDescriptor<node, any>` as a variadic tuple that may omit required mixin arguments.
- Adds JSX coverage for using an argument-bearing mixin authored for `Element` on a compatible `HTMLButtonElement` host.

The two-commit structure keeps the reproduction independently observable before applying the fix.

Screenshot 2026-08-17 at 3 00 11 PM


mjackson (changes requested):
This LGTM. We should remove the reproduction commit, but other than that I think this looks good.

PR #11723, opened 2026-08-18
feat(cli): add `remix db rollback`

kuboon:
`Database.migrate()` already accepts `direction`, `step`, `to`, and `dryRun`, but the CLI never passes them, so a migration's `down.sql` was unreachable from `remix db`. `--to` only bounds forward progress: pointed at an applied migration it prints `no pending migrations` and reverts nothing. The only ways to run a down script were `reset` (which reverts everything) or a hand-written script calling `db.migrate()` directly.

Add `rollback`, which reverts applied migrations newest first:

    remix db rollback                   # revert the most recent migration
    remix db rollback --step 2          # revert the two most recent
    remix db rollback --to 20260101000000   # revert back through this one, inclusive
    remix db rollback --dry-run         # report without running

`--step` and `--to` are mutually exclusive, matching `MigrationOperationOptions`; `--step` defaults to 1. It also takes the `--migrations`, `--journal-table`, and `--connection-env` options the other migration commands take.

Unlike `wipe` and `reset`, `rollback` does not require `--force`: reverting one migration is a routine development step rather than a whole-database operation, and `--dry-run` covers the "what would this do" case. Happy to gate it behind `--force` if maintainers prefer.

The printing lives in `runRemixDb()` alongside the other commands, so hosts embedding the data-table CLI get `rollback` too.

### DIFF
--- a/packages/cli/.changes/minor.add-db-rollback.md
+++ b/packages/cli/.changes/minor.add-db-rollback.md
@@ -1,0 +1,8 @@
+Add `remix db rollback`, which reverts applied migrations by running their `down.sql`
+
+`Database.migrate()` already accepted `direction`, `to`, `step`, and `dryRun`, but the CLI never
+passed them, so a migration's `down.sql` was unreachable from `remix db` — `--to` only bounds
+forward progress. `rollback` reverts newest first, bounded by `--step <count>` (default `1`) or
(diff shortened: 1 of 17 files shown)


mjackson:
Thanks, @kuboon!

PR #11730, opened 2026-08-20
Replace assets `fileMap` with `mounts`

markdalgleish:
This replaces `fileMap` with optional directory-based `mounts`.

The primary motivation behind this was to have more predictable mappings that allow for better import map support: https://github.com/remix-run/remix/pull/11706.

Mounts map public URL paths to directories, with defaults being the same as doing this:

```ts
createAssetServer({
  basePath: '/assets',
  mounts: {
    app: 'app',
    npm: 'node_modules',
  },
  allowFiles: ['app/**'],
})
```

Mount paths with overlapping URL or filesystem roots are rejected during asset server startup.

Existing configurations using the defaults from our app template can remove `fileMap` entirely:

```ts
// Before
createAssetServer({
  basePath: '/assets',
  fileMap: {
    '/app/*path': 'app/*path',
    '/npm/*path': 'node_modules/*path',
  },
  allowFiles: ['app/**'],
})

// After
createAssetServer({
  basePath: '/assets',
  allowFiles: ['app/**'],
})
```

Custom mappings can be converted by removing the trailing wildcards:

```ts
// Before
createAssetServer({
  basePath: '/assets',
  fileMap: {
    '/source/*path': 'app/*path',
    '/vendor/*path': 'node_modules/*path',
  },
  allowFiles: ['app/**'],
})

// After
createAssetServer({
  basePath: '/assets',
  mounts: {
    source: 'app',
    vendor: 'node_modules',
  },
  allowFiles: ['app/**'],
})
```


mjackson (approved):
Looks great!

PR #11672, opened 2026-08-06
Escape server-rendered CSS style text

brophdawg11:
Server-rendered CSS mixins are embedded in raw-text `<style>` elements. A literal `<` in dynamic CSS could form a `</style>` end tag and cause the remaining content to be parsed as HTML.

- CSS-escape `<` as `\3C ` at the final style serialization boundary
- cover generated declarations, nested selectors, and at-rules while preserving their CSS meaning
- add a browser regression test and patch change note for `@remix-run/ui`

PR #11731, opened 2026-08-20
fix(ui): preserve range media queries when escaping SSR style tags

### LINKED ISSUE #11732: Media queries appear to be broken on remix docs
### Reproduction

1. Visit https://api.remix.run
2. View on a none-desktop viewport

Image

### Expected Behavior

A mobile nav is displayed conditionally.

### Actual Behavior

The desktop nav is displayed but unclickable when the screen width is less than 640px.

henryStelle:
This fixes a bug introduced by [#11672](https://github.com/remix-run/remix/pull/11672) that resulted in valid CSS being mutated during SSR style escaping (for example, range media queries like `@media (width < 900px)`) while protecting against XSS. Instead, this update only replaces literal closing style tags (searching for `</style` instead of `<`).

Resolves https://github.com/remix-run/remix/issues/11732

## Before / After

| Before | After |
| --- | --- |
|  Screenshot 2026-08-19 at 11 19 26 PM  |  Screenshot 2026-08-19 at 11 18 46 PM |

### DIFF
--- a/packages/ui/src/server/stream.ts
+++ b/packages/ui/src/server/stream.ts
@@ -1326,8 +1326,8 @@ function renderStyleTag(
 }
 
 function escapeStyleText(css: string): string {
-  // A literal "</style" closes an HTML style element even when it appears inside a CSS string.
-  return css.replace(/</g, '\\3C ')
+  // Only neutralize literal style end tags. Escaping every '<' breaks valid range media queries.
+  return css.replace(/<\/style/gi, '<\\/style')
 }
 
 function buildRmxDataScript(context: RenderContext): string {
--- a/packages/ui/src/server/stream.ts
+++ b/packages/ui/src/server/stream.ts
@@ -1327,7 +1327,7 @@ function renderStyleTag(
 
 function escapeStyleText(css: string): string {
   // Only neutralize literal style end tags. Escaping every '<' breaks valid range media queries.
-  return css.replace(/<\/style/gi, '<\\/style')
+  return css.replace(/<\/style/gi, '\\3C/style')
 }
 
 function buildRmxDataScript(context: RenderContext): string {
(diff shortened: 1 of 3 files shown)


henryStelle:
@brophdawg11 this touches your work, would you be able to review it? Thanks!

packages/ui/src/server/stream.ts
```diff
@@ -1326,8 +1326,8 @@ function renderStyleTag(
 }
 
 function escapeStyleText(css: string): string {
-  // A literal "</style" closes an HTML style element even when it appears inside a CSS string.
-  return css.replace(/</g, '\\3C ')
+  // Only neutralize literal style end tags. Escaping every '<' breaks valid range media queries.
+  return css.replace(/<\/style/gi, '<\\/style')
```
brophdawg11: What if we just looked for the close starting sequence?

```suggestion
  return css.replace(/<\//g, '\\3C\\/')
```

brophdawg11:
Good catch!

packages/ui/src/server/stream.ts
henryStelle: I ended up digging into this more and it turns out you can [trigger arbitrary JS](https://owasp.deteact.com/cheat/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#rule-4---css-escape-and-strictly-validate-before-inserting-untrusted-data-into-html-style-property-values) from within css without even breaking out of style block.

However, if we constrain ourselves to trying to stop only exiting from style tag in the CSS, according to https://html.spec.whatwg.org/multipage/parsing.html#rawtext-end-tag-name-state we need to concern ourselves with `/<\/style(?=[\u0009\u000A\u000C\u0020\u002F\u003E])/gi` as those are the possible tokens that would trigger HTML to possibly exist the open style tag. However, that is confusing, hence shortening to `</style`.

As to why not further shorten to `</`, HTML would only exit the style tag if the end tag name matches the start, so you don't gain anything with the cost of possibly mutating "valid" property references (eg `div[name="</strange"`) or other weird edge cases like `.my-element::after { content: "</bob> isn't real"; }`

brophdawg11: ah ok, thats makes sense.  Any specific reason not to continue using `\\3C`?

```ts
return css.replace(/<\/style/gi, '\\3C/style')
```

henryStelle: No particular reason, I'll change it back.

brophdawg11 (approved):

PR #11736, opened 2026-08-21
Preserve scroll for no-reset navigations

brookslybrand:
Enhanced navigations that set `data-rmx-reset-scroll="false"` should preserve the current scroll position. The Navigation API was still applying its default scrolling after the intercepted transition, and a top-frame redirect started a replacement navigation without carrying the original reset intent.

- Use `scroll: 'manual'` only for new push and replace navigations that opt out of resetting scroll.
- Carry `resetScroll` into the redirect successor navigation so POST frame redirects preserve the original form's intent.
- Keep the default top reset for redirects that do not opt out, and avoid scrolling aborted or superseded navigations.

This is stacked on the navigation scroll behavior demo in [#11733](https://github.com/remix-run/remix/pull/11733). [#11699](https://github.com/remix-run/remix/pull/11699) builds on this with the separate traversal reconciliation fix.

https://github.com/user-attachments/assets/d1e419eb-d5cf-49aa-ac58-99ecf69c1cc6

The Frames demo reproduction:

https://github.com/user-attachments/assets/68568c97-0af1-4cba-9bad-44031c883172

### DIFF
--- a/demos/frames/app/actions/controller.tsx
+++ b/demos/frames/app/actions/controller.tsx
@@ -1,3 +1,4 @@
+import { redirect } from 'remix/response/redirect'
 import { createController } from 'remix/router'
 
 import { routes } from '../routes.ts'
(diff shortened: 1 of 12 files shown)


brookslybrand:
Closing because #11739 now provides the reset-scroll behavior this PR covered. #11699 has been rebased directly onto #11733.

PR #11742, opened 2026-08-23
Add a module database adapter to remix.json

kuboon:
`db.adapter.type` accepted only `sqlite`, `postgres`, and `mysql`, so a project on any other database (cloudfare d1 or turso, etc) could not run `remix db` at all:

    Error [RMX_INVALID_CONFIG] Invalid Remix configuration
    remix.json:1:32: Expected one of: sqlite, postgres, mysql at db.adapter.type

There was no hook to plug one in, which left those projects reimplementing the command against `runRemixDb()` to get the migration directory layout, journal table, and output the CLI already produces.

Add a fourth adapter type that names a module instead of a driver:

    {
      "db": {
        "adapter": {
          "type": "module",
          "module": "./app/database.ts",
          "export": "createDatabase",
          "connection": { "env": "DATABASE_URL" },
          "options": { "syncInterval": 60 }
        }
      }
    }

The module exports a factory typed by the new `RemixDbModuleFactory`, receives `{ configDir, connection, options }`, and returns a `Database` that the CLI closes when the command finishes. `export` defaults to the default export. `connection` resolves like the other adapters' connection values and `--connection-env` overrides it, so the flag keeps working for module adapters. `options` is passed through unchanged.

Relative and absolute `module` specifiers resolve from `remix.json`. Specifiers that name a scheme belong to the runtime. Bare specifiers resolve from the app rather than the CLI, so a database package the app installed is reachable even when the CLI's own resolver cannot see it, and fall back to the raw specifier when that resolution fails.

Options come back from the JSONC parser as null-prototype objects; give them ordinary prototypes before handing them to third-party code.

`RMX_DB_MODULE_NOT_FOUND` reports an import that failed, and `RMX_DB_MODULE_FACTORY_REQUIRED` reports an export that is not a function.

### DIFF
--- a/packages/cli/.changes/minor.module-database-adapter.md
+++ b/packages/cli/.changes/minor.module-database-adapter.md
@@ -1,0 +1,19 @@
+Added a `type: "module"` database adapter to `remix.json`, so `remix db` can drive databases the CLI does not ship an adapter for, such as Turso/libSQL or Cloudflare D1. Previously `db.adapter.type` accepted only `sqlite`, `postgres`, and `mysql`, and third-party databases had to reimplement the command instead.
+
+```jsonc
+{
+  "db": {
+    "adapter": {
+      "type": "module",
+      "module": "./app/database.ts",
+      "export": "createDatabase",
+      "connection": { "env": "DATABASE_URL" },
+      "options": { "syncInterval": 60 },
+    },
+  },
+}
+```
+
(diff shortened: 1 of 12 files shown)


kuboon:
I'd like to use Turso and Cloudflare D1 with the Remix data schema, and to drive Turso through the Remix CLI. I'm not attached to this particular implementation — any approach that makes that possible works for me.

PR #11745, opened 2026-08-24
Fix server-rendered CSS range queries

brookslybrand:
#11672 correctly prevented CSS mixin values from breaking out of server-rendered `<style>` elements, but it escaped every `<` character at the serialization boundary. That also changed CSS range queries such as `@media (width < 900px)` into invalid syntax, causing browsers to discard the mobile styles used by the Remix docs.

- Escape only `<` characters that begin a case-insensitive `</style` end tag, preserving the original breakout protection
- Preserve comparison operators in range media queries
- Cover both the style-tag breakout case and server-rendered range query output with browser regression tests
- Add a patch change note for `@remix-run/ui`

image

### DIFF
--- a/docs/shared/ui/docs-header.tsx
+++ b/docs/shared/ui/docs-header.tsx
@@ -426,7 +426,7 @@ const navigationCss = css({
   justifyContent: 'flex-end',
   width: 'auto',
   height: 'var(--site-header-height)',
-  gap: '32px',
+  gap: '24px',
   overflow: 'visible',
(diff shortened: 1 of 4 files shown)


brookslybrand:
Closing because #11731 fixed the same CSS serialization issue. I’ll move the unrelated docs navigation spacing change into a separate PR.

brookslybrand:
The docs navigation spacing change has moved to #11748.

PR #11747, opened 2026-08-26
Add Standard JSON Schema support to data-schema

sergiodxa:
This adds [Standard JSON Schema](https://standardschema.dev/json-schema) support for the remix/data-schema package.

I tried to keep changes to current code minimal, only adding the `~def` property to the schemas, and a `meta` method to set a title, description and custom jsonSchema (used for case like `refine`), all optional.

The API is simple, once you define a schema:

```ts
let Search = object({
  query: string()
    .pipe(minLength(1), maxLength(200))
    .meta({ description: 'Words to look for.' }), // Sets JSON Schema description
  kind: optional(enum_(['article', 'tutorial'])),
  limit: defaulted(number().pipe(min(1), max(50)), 10),
})
```

You can now call `toJSONSchema(Search)` and get an the JSON schema object

```json
{
  "type": "object",
  "properties": {
    "query": { "type": "string", "minLength": 1, "maxLength": 200, "description": "Words to look for." },
    "kind": { "type": "string", "enum": ["article", "tutorial"] },
    "limit": { "type": "number", "minimum": 1, "maximum": 50, "default": 10 },
  },
  "required": ["query"],
}
```

You can then pass this object to any package that expects a JSON schema.

If the data-schema has something that can't be converted (like a Set, Map, etc.) then it immediately throw so ideally you catch this at dev time.

I also avoided intentionally to change or adjust any existing test, so everything that already works still works, this is purely additive changes.


packages/data-schema/src/lib/form-data.ts
sergiodxa, line 1: Most of this is just adding one indentation level because `{ kind: 'formData' }` is not an argument of createSchema, so this was re-formatted automatically.

packages/data-schema/src/lib/coerce.ts
sergiodxa, line 1: Most of this is adding an indentation level because of the second argument of `createSchema`, so this was mostly reformatted automatically.

packages/data-schema/package.json
```diff
@@ -45,6 +46,10 @@
         "types": "./dist/form-data.d.ts",
         "default": "./dist/form-data.js"
       },
+      "./json-schema": {
+        "types": "./dist/json-schema.d.ts",
+        "default": "./dist/json-schema.js"
+      },
```
sergiodxa, line 52: I added it as a separate path so the final import is `remix/data-schema/json-schema`.

packages/data-schema/package.json
```diff
@@ -56,6 +61,7 @@
     "@remix-run/assert": "workspace:*",
     "@remix-run/test": "workspace:*",
     "@types/node": "catalog:",
+    "ajv": "^8.17.1",
```
sergiodxa, line 64: This dev dependency ([ajv](https://www.npmjs.com/package/ajv)) is only here so the output JSON schema can be validated against a third party validator, instead of custom tests that could be wrong

brookslybrand:
@remix-run-bot can you review this PR?

mjackson:
I'd like to refactor the data-schema core first in #11878 and land that work before this PR. That change will make schema objects directly inspectable instead of keeping their rule data only in closures.

Once that lands, I'd like to update and land this work on top of it as a separate conversion function, exposed from a separate export. This keeps JSON Schema conversion out of the main data-schema bundle and leaves the core validation path unchanged for users who don't need it.

```ts
import { object, string } from 'remix/data-schema'
import { toStandardJSONSchema } from 'remix/data-schema/json-schema'

let Search = object({
  query: string(),
})

// Only import and run the converter when JSON Schema is needed.
let jsonSchema = toStandardJSONSchema(Search)
```

PR #11755, opened 2026-08-28
Fix navigation scroll timing

brophdawg11:
Frame navigations could leave the destination UI visible at the previous page's scroll position while slow client modules or frame streams kept the Navigation API handler pending.

- Split internal frame navigation reloads into `committed` and `finished` phases.
- Resolve `committed` after the destination's first live DOM batch and the first DOM commit from every blocking frame, then perform the browser's scroll reset or restoration.
- Keep `finished` waiting for remaining frame streams and client-entry hydration so redirects and the public `frame.reload()` completion contract stay unchanged.


brophdawg11:
hey @remix-run-bot can you merge latest main into this pr?

brophdawg11:
@remix-run-bot can you review this PR?

packages/ui/src/runtime/navigation.ts
```diff
@@ -289,46 +281,148 @@ function isCrossOriginDestination(event: NavigateEvent): boolean {
...
+
+    // We think this is a bug in Chromium where they are incorrectly classifying a
+    // DOM-modification-driven scroll change as a user scroll action, causing it to skip restoration
+    // after the transition. The intended user-scroll behavior is tested here:
+    // https://github.com/web-platform-tests/wpt/blob/master/navigation-api/scroll-behavior/after-transition-skips-restore-when-scrolled.html
+    let { scrollHeight, clientHeight } = document.documentElement
+    return {
+      cssText: `
+        html {
+          min-height: ${scrollHeight + clientHeight}px !important;
```
brookslybrand, line 360: I'm not sure we need the `scrollHeight` + `clientHeight`, shouldn't `scrollHeight` be enough? Or am I missing something?

brophdawg11: we needed to add one viewport height for edge case issues found in live browser tests

Screenshot 2026-09-04 at 4 12 54 PM

PR #11757, opened 2026-08-28
Point Remix config schema to GitHub

### LINKED ISSUE #11758: Add correct + versioned `$schema` url to `remix.json` throughout codebase
Currently the codebase is littered in [references to https://remix.run/schemas/remix.json](https://github.com/remix-run/remix/blob/0b25fab078508d1565fdd68377a2629b53031e2b/packages/ui/remix.json#L2).

Image

- this resource does not exist
- this should probably referenced a versioned schema so that your `remix.json` file always refers to the correct version of Remix
- the file should probably come from `remix/*` itself, not a url, unless there's some limitation there

Alternatively, this could be a `.ts` file and not a `json`, so it can be type checked

mjackson:
The `remix.json` schema references point to a `remix.run` URL that is not live and would require a website deploy for every schema update. This points them to the raw schema on the Remix repository's `main` branch so applications can use the current schema directly.

- Update config examples, docs, changelog examples, and the schema `$id` to use the raw GitHub URL
- Remove the website schema copy and its sync test
- Add a patch change note for `@remix-run/cli`

Closes #11758.


sergiodxa:
Why not point it to node_modules? I saw Oxc tools doing that, so you always point it to the same version you have installed, and when you upgrade the package the manifest is updated as well.

brookslybrand:
> Why not point it to node_modules? I saw Oxc tools doing that, so you always point it to the same version you have installed, and when you upgrade the package the manifest is updated as well.

I asked the same thing. I believe we were thinking that we needed a URI for the `$schema`, but that does not appear to be the case (just a requirement for JSON schema itself).

I've put up a PR implementing it as local file in `node_modules` instead: https://github.com/remix-run/remix/pull/11763

PR #11759, opened 2026-08-30
Fix multipart parsing with RFC 2046 preambles

### LINKED ISSUE #11746: `@remix-run/multipart-parser`: RFC 2046 preamble causes "missing initial boundary"
### Reproduction

Tested with:

- `@remix-run/multipart-parser` 0.16.4
- Node.js 26.7.0

Create `reproduction.mjs`:

```javascript
import { parseMultipart } from "@remix-run/multipart-parser";

const boundary = "example-boundary";
const body = new TextEncoder().encode(
  [
    "This is a MIME preamble.",
    `--${boundary}`,
    "Content-Type: text/plain",
    "",
    "hello",
    `--${boundary}--`,
    "",
  ].join("\r\n"),
);

console.log(
  [...parseMultipart(body, { boundary })].map((part) => part.text),
);
```

Run:

```console
npm install @remix-run/multipart-parser@0.16.4
node reproduction.mjs
```

This is also reproducible with an empty preamble consisting only of a leading
CRLF before the opening boundary.

### Expected Behavior

The parser should ignore the MIME preamble and return:

```text
[ 'hello' ]
```

RFC 2046 permits text before the first boundary delimiter and requires MIME
implementations to ignore it:

https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1.1

This matters for `multipart/mixed` responses produced by APIs such as the
Google Drive batch endpoint.

### Actual Behavior

The parser throws:

```text
MultipartParseError: Invalid multipart stream: missing initial boundary
```

`MultipartParser.write()` currently requires the opening boundary to begin at
offset zero:

https://github.com/remix-run/remix/blob/multipart-parser%400.16.4/packages/multipart-parser/src/lib/multipart.ts#L423-L431

Removing the preamble before calling `parseMultipart()` makes the same body
parse successfully. It would be useful for the parser itself to ignore a legal
preamble, including when the opening boundary is split across input chunks.

QuarkOS:
Fixes #11746.

`MultipartParser` ignores RFC 2046 preambles before the first boundary, including a leading CRLF and boundaries split across chunks. Boundary text in the middle of a preamble line is ignored.

Chunks shorter than the opening boundary are buffered before searching, preserving parsing speed when the boundary arrives in tiny chunks.


mjackson (approved):

mjackson:
Thanks for the PR, @QuarkOS!

PR #11761, opened 2026-08-31
fix(ui): allow handle.update() during a component's first render

### LINKED ISSUE #11642: Bug: `handle.update()` during a component’s first render throws "scheduleUpdate not implemented"
**Versions:** `remix@3.0.0-beta.5`, `@remix-run/ui@0.4.0`

## What happens

`ComponentRuntime` initialises `#scheduleUpdate` to a thrower, and `renderComponent` only replaces it *after* the component's render has already run:

```ts
// packages/ui/src/runtime/component.ts
#scheduleUpdate: () => void = () => {
  throw new Error('scheduleUpdate not implemented')
}

// packages/ui/src/runtime/reconcile.ts
export function renderComponent(handle, currContent, next, domParent, ...) {
  let [element, tasks] = handle.render(next.props)   // <- setup fn + render fn run here
  let content = toVNode(element)
  let newCursor = diffVNodes(...)                    // <- children render here too
  ...
  handle.setScheduleUpdate(() => {                   // <- only assigned now
    scheduler.enqueue(committed, domParent)
  })
  scheduler.enqueueTasks(tasks)
}
```

So any component that reaches `handle.update()` synchronously during its first render — directly, or through a helper it calls, or through an `async` function whose body sets state and calls `handle.update()` before its first `await` — hits the thrower.

Because `update()` builds its promise with `new Promise(executor)`, the throw is converted into a **rejected promise**, so it surfaces as an unhandled rejection rather than a render error:

```ts
update: () =>
  new Promise((resolve) => {
    if (component.#removed) { resolve(AbortSignal.abort()); return }
    this.#tasks.push((signal) => resolve(signal))
    this.#scheduleUpdate()   // throws → promise rejects
  }),
```

## Reproduction

Verified against `@remix-run/ui@0.4.0` in a real app. Any component of this shape:

```tsx
export function Boom(handle: Handle) {
  let first = true
  return () => {
    if (first) {
      first = false
      handle.update()
    }
    return <p>hello</p>
  }
}
```

Rendering it produces:

```
Uncaught (in promise) Error: scheduleUpdate not implemented
```

## Why it matters

The error message gives no indication of which component is at fault or that the real problem is "you updated during render", so it is hard to act on — especially arriving as an unhandled rejection detached from the render stack. We hit it in production through an indirect path: a route's render function applied preloaded loader data via a helper, and that helper kicked off an `async` loader that set state and called `handle.update()` before awaiting. Nothing in the message pointed at any of that.

## Suggested fix

Any of these would help, roughly in order of preference:

1. Assign `setScheduleUpdate` **before** invoking the component's setup/render, so an update requested during the first render coalesces into the commit that is already in flight.
2. Make the default a no-op that marks the component dirty, and let the pending first commit pick it up.
3. At minimum, make the default error descriptive: name the component and say that `handle.update()` was called during render, and throw it synchronously rather than letting it become an unhandled rejection.

Our own fix was to move the call off the render path (`handle.queueTask(...)`), which is presumably the intended pattern — but the framework could make that much easier to discover: https://github.com/kentcdodds/kody/pull/921

_Filed by an agent on behalf of the [kentcdodds/kody](https://github.com/kentcdodds/kody) project._

QuarkOS:
fixes #11642

handle.update() during setup or the first render calls scheduleUpdate before the queue exists, so you get scheduleUpdate not implemented.

We treat scheduling as a no-op until that is initialized. the update is already in the in-flight render.

the vdom component tests cover setup, first render, and after commit.

### DIFF
--- a/packages/ui/src/runtime/component.ts
+++ b/packages/ui/src/runtime/component.ts
@@ -279,10 +279,13 @@ class ComponentRuntime<C = NoContext> implements ComponentHandle<C> {
   #updateDomParent: ParentNode | undefined
   #scheduleUpdate = (): void => {
     let queue = this.#updateQueue
-    if (!queue) throw new Error('scheduleUpdate not implemented')
     let vnode = this.#updateVNode
     let domParent = this.#updateDomParent
-    if (!vnode || !domParent) throw new Error('scheduleUpdate target not initialized')
+
+    // Updates during setup or the first render are already reflected by the
+    // in-flight render. Their tasks will be enqueued after the first commit.
+    if (!queue || !vnode || !domParent) return
+
     queue.enqueue(vnode, domParent)
   }
   #tasks: Task[] = []
--- a/packages/ui/src/test/vdom.components.test.tsx
+++ b/packages/ui/src/test/vdom.components.test.tsx
@@ -15,6 +15,44 @@ describe('vnode rendering', () => {
   describe('components', () => {
     it.todo('warns when render is called after component is removed')
 
+    it('allows handle.update() during setup and the first render', async () => {
+      let container = document.createElement('div')
+      let capturedUpdate = () => Promise.resolve(AbortSignal.abort())
+      let initialUpdates: Array<Promise<AbortSignal>> = []
+
+      function App(handle: Handle) {
+        capturedUpdate = () => handle.update()
+        initialUpdates.push(handle.update())
+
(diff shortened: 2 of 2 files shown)


brophdawg11:
Superseded by #11795.

*Filed by an agent on behalf of @brophdawg11*

PR #11767, opened 2026-08-31
Add 405 Method Not Allowed responses and automatic HEAD support to fetch-router

mjackson:
Previously, a request whose URL matched a route pattern but whose method had no registered route fell through to the router's `defaultHandler` and produced a `404 Not Found`. That response is misleading — the resource exists, the method is what's wrong — and the information needed to answer correctly (which methods *are* registered for that URL) was discarded inside the dispatch loop. Similarly, `HEAD` requests only matched routes explicitly registered for `HEAD` or `ANY`, so a plain `router.get()` app returned `404` to every `HEAD` request (health checks, link validators, HTTP caches).

This PR fixes both in `dispatchMatches`:

- Requests that match a route pattern registered only for other methods now receive a `405 Method Not Allowed` response with an `Allow` header listing the methods registered for that URL (including `HEAD` whenever `GET` is allowed).
- Method fall-through is preserved: less specific routes that can handle the method (including `ANY` catch-alls) still win before any 405 applies, so SPA fallbacks and proxies keep working unchanged. Registering an `ANY` route remains the way to receive every method at a URL.
- `GET` routes now serve `HEAD` requests automatically. The `GET` handler (and its middleware) runs as usual and the router strips the response body, so `HEAD` observes the same status and headers as the equivalent `GET`. An explicit `HEAD` route for the same pattern takes precedence, and a more specific `GET` route beats a less specific `HEAD`/`ANY` route, mirroring normal specificity ranking.

`405` responses only apply when no route could handle the request, so this changes behavior only for requests that previously produced a misleading `404`.

### Method mismatches now answer correctly

```ts
let router = createRouter()

router.get('/users/:id', ({ params }) => new Response(`User ${params.id}`))
router.put('/users/:id', ({ params }) => new Response(`Updated ${params.id}`))

let response = await router.fetch('https://remix.run/users/1', { method: 'DELETE' })
// Before: 404 Not Found
// After:  405 Method Not Allowed
//         Allow: GET, HEAD, PUT
```

Fall-through to catch-alls is unchanged:

```ts
router.get('/posts/featured', () => new Response('Featured'))
router.route('ANY', '/posts/:id', () => new Response('Any post'))

await router.fetch('https://remix.run/posts/featured', { method: 'POST' })
// 200 "Any post" — same as before, no 405
```

### HEAD requests work out of the box

```ts
let router = createRouter()

router.get('/report', () => new Response('report data', { headers: { 'Content-Type': 'text/csv' } }))

let response = await router.fetch('https://remix.run/report', { method: 'HEAD' })
// Before: 404 Not Found
// After:  200 OK, Content-Type: text/csv, empty body
```

An explicit `HEAD` route still wins for its pattern, regardless of registration order:

```ts
router.get('/file', () => new Response(readFile('file')))
router.head('/file', () => new Response(null, { headers: { 'Content-Length': sizeOf('file') } }))

await router.fetch('https://remix.run/file', { method: 'HEAD' })
// Served by the HEAD route without reading the file
```

The demo apps' router tests asserting `404` for method mismatches were updated to expect `405` with the `Allow` header. Release notes are included for `@remix-run/fetch-router` (with a `BREAKING CHANGE` entry for the 405 behavior) and mirrored for the `remix` package.


brophdawg11 (approved):
LGTM - gonna run the review agentic workflow on this too as I keep testing that flow out

PR #11777, opened 2026-09-02
Use Oxc transforms in test package

brophdawg11:
`@remix-run/test` continued using esbuild for TypeScript transforms because older Oxc source maps produced inaccurate code coverage. The upstream fix in [oxc-project/oxc#22001](https://github.com/oxc-project/oxc/pull/22001) shipped in Oxc 0.130.0, so the test package can now use the same transformer as the assets and node-tsx packages.

- Replace esbuild with `oxc-transform` while preserving inline source maps and tsconfig JSX behavior
- Align `@remix-run/assets` and `@remix-run/node-tsx` on `oxc-transform` 0.130.0
- Preserve original `node-hmr` stack trace locations when transformed mappings begin after column zero
- Strengthen coverage parity assertions so uncovered functions, statements, and branches remain accurately reported across server, browser, and E2E runs


markdalgleish:
There's an upstream issue that might be a blocker for us: https://github.com/oxc-project/oxc/issues/22048

markdalgleish:
This is the detail I got from my agent more specific to us:

> Oxc drops coverage pragmas inside transformed JSX. A `/* c8 ignore next */` before a JSX prop callback is removed, causing an intentionally ignored function to be reported uncovered. The real Remix pipeline reported 50% functions and falsely failed a 60% threshold; esbuild preserves the pragma.

PR #11780, opened 2026-09-02
Replaced by #11784: Fix Chromium scroll resets during frame navigation

brookslybrand:
GitHub marked this PR as merged while its stacked branches were reordered; it did not merge into `main`.

Continued in #11784.


brookslybrand:
Continued in #11784 after the stacked branches were reordered.

PR #11782, opened 2026-09-02
Simplify navigation transition handling

brookslybrand:
Navigation transitions repeated interception setup and guarded reload commits twice.

- Route interception through one helper and apply scroll behavior and browser workarounds there.
- Tie traversal styles and the WebKit viewport resync to each transition’s `finished` promise.
- Remove the duplicate frame commit guard and preserve only the starting document’s scroll height during traversal.

Stacked on #11784.

### DIFF
--- a/packages/ui/src/runtime/navigation.ts
+++ b/packages/ui/src/runtime/navigation.ts
@@ -2,15 +2,6 @@ import { getTopFrame, getNamedFrame } from './run.ts'
 import { reloadFrameForNavigation } from './frame.ts'
 import { createFormNavigationResolver, type FormSubmission } from './form-navigation.ts'
 
-interface NavigationPrecommitControllerLike {
-  redirect(url: string, options: { history: 'replace' }): void
-}
-
-interface NavigationInterceptOptionsWithPrecommit extends NavigationInterceptOptions {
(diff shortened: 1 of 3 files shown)


packages/ui/src/runtime/navigation.ts
```diff
@@ -356,11 +341,11 @@ function preserveStartingDocumentScrollState(navigation: Navigation, event: Navi
   // after the transition. The intended user-scroll behavior is tested here:
   // https://github.com/web-platform-tests/wpt/blob/master/navigation-api/scroll-behavior/after-transition-skips-restore-when-scrolled.html
 
-  let { scrollHeight, clientHeight } = document.documentElement
+  let { scrollHeight } = document.documentElement
   let stylesheet = new CSSStyleSheet()
   stylesheet.replaceSync(`
     html {
-      min-height: ${scrollHeight + clientHeight}px !important;
+      min-height: ${scrollHeight}px !important;
```
brookslybrand, line 348: I don't think we need the `scrollHeight` + `clientHeight`, `scrollHeight` should be enough and this over shoots the temporary min-height

brookslybrand:
I'm not convinced this refactor carries its weight, I'm closing it

PR #11784, opened 2026-09-02
Fix Chromium scroll resets during frame navigation

brookslybrand:
Chromium can leave frame navigations at the wrong scroll position: Chrome 148 and earlier skip resets for destinations without fragments, and Chromium can replay a pending reload restoration after a navigation.

- Fall back to `scrollTo(0, 0)` when the Navigation API does not reset a push or replace navigation.
- Hold scroll restoration for navigations committed before `load`, then restore it after Chromium’s final retry.
- Release that hold before the next navigation and when the runtime stops.

Stacked on #11755. Replaces #11780 after the stack was reordered.


brophdawg11:
Hey @remix-run-bot - can you review this PR?

brookslybrand:
This solves a pretty specific Chromium bug that we've reported to the team: https://partnerissuetracker.corp.google.com/issues/556294787

Closing this issue so we don't carry forward a hack for a pretty specific and hard to trigger browser bug

PR #11791, opened 2026-09-03
Add a server-rendered i18next demo

brookslybrand:
Adds a server-rendered internationalization demo using i18next directly with Remix routes, middleware, request context, rendering, and browser assets. Translation instances stay isolated per request, while native links and forms provide the no-JavaScript baseline.

- Detects language from explicit `/en`, `/es`, `/fr`, `/ja`, and `/ar` routes, then a saved locale cookie, the browser's `Accept-Language` preference, and finally English
- Sets `Content-Language` and cache-safe `Vary` headers, with stable representations for explicit locale URLs
- Distinguishes language links from saved preferences: links soft-navigate without changing the cookie; preference forms use POST-redirect-GET and document navigation to reset unsaved selector state
- Uses typed translation resources with locale-specific plural categories, including Arabic's six categories and French/Spanish million-count forms
- Demonstrates Arabic RTL layout with `<html dir>`, logical CSS properties, and isolated LTR code examples
- Uses standard `Intl` APIs for formatting, with explicit USD currency and UTC time-zone policy
- Demonstrates the server/browser boundary with a client number preview that receives only translated labels and locale data, preserving local state while language links update its props

The request-scoped pattern remains explicit:

```ts
let { locale, source } = await detectLanguage(context.request, context.params.locale)
let instance = createInstance()

await instance.init({ lng: locale, resources })
context.set(requestI18n, {
  locale,
  direction: instance.dir(locale),
  detectionSource: source,
  t: instance.getFixedT(locale),
}, { property: 'i18n' })
```


sergiodxa:
In case you want to drop a lot of code, I published my interop package to [@sdxc/i18n](https://www.npmjs.com/package/@sdxc/i18n)

brookslybrand:
> In case you want to drop a lot of code, I published my interop package to [@sdxc/i18n](https://www.npmjs.com/package/@sdxc/i18n)

Thanks Sergio! I mostly want this demo to just be a good, minimal library example of how you might setup i18n with Remix. We may add your library to the README just so people are aware of all their options.

PR #11813, opened 2026-09-07
Preserve client-set root attributes across top-frame reload

### LINKED ISSUE #11809: remix/ui: top-frame navigation removes client-set attributes on <html> (syncElementAttributes)
## Summary

In `@remix-run/ui` 3.0.0-rc.1, a top-frame navigation/reload (`app.frames.top.reload()`, intercepted link/form navigation) rewrites the attributes of `<html>` from the server response: every attribute the client added after hydration that is not present in the server HTML is removed (`syncElementAttributes` in `runtime/frame.ts`).

This makes it hard to keep client-owned state on the root element across navigations — the common example is a dark-mode class (`<html class="dark">`) or `data-*` flags set by the client — and produces a visible theme flash on every navigation when `color-scheme` / `dark` is re-applied asynchronously.

## Reproduction (minimal, no framework code)

Server renders two plain pages with the runtime entry; the client sets an attribute and a class on `<html>` after `app.ready()`, then clicks a same-origin link.

```js
// app/public/entry.ts
import { run } from 'remix/ui'
const app = run({ loadModule: async (url, name) => (await import(url))[name] })
window.__rmx_app = app
```

```js
// repro.mjs — node --import remix/node-tsx repro.mjs  (playwright + remix/assets + remix/node-fetch-server/test)
import { chromium } from 'playwright'
import { createTestServer } from 'remix/node-fetch-server/test'

const assets = createAssetServer({ rootDir: '.', basePath: '/assets', allowFiles: ['app/**/public/**'], allowPackages: ['remix'] })
const entryHref = await assets.getHref('app/public/entry.ts')
const page = (title, next) => `<!DOCTYPE html><html><head><meta charset="utf-8"><title>${title}</title>
<script type="module" src="${entryHref}"></script></head>
<body><h1>${title}</h1><a id="nav" href="${next}">next</a></body></html><!-- rmx:flush document -->`

const server = await createTestServer(async (req) => {
  const { pathname } = new URL(req.url)
  if (pathname.startsWith('/assets/')) return (await assets.fetch(req)) ?? new Response('', { status: 404 })
  if (pathname === '/page-b') return new Response(page('Page B', '/page-a'), { headers: { 'content-type': 'text/html' } })
  return new Response(page('Page A', '/page-b'), { headers: { 'content-type': 'text/html' } })
})
const browser = await chromium.launch(); const p = await browser.newPage()
await p.goto(server.baseUrl + '/page-a')
await p.waitForFunction(() => window.__rmx_app?.ready()); await p.evaluate(() => window.__rmx_app.ready())
await p.evaluate(() => { document.documentElement.setAttribute('data-x', '1'); document.documentElement.classList.add('dark') })
await p.click('#nav'); await p.waitForFunction(() => document.querySelector('h1')?.textContent === 'Page B')
console.log(await p.evaluate(() => ({ dataX: document.documentElement.getAttribute('data-x'), className: document.documentElement.className })))
await browser.close(); await server.close()
```

Output:

```
before navigation: { dataX: '1', className: 'dark' }
after navigation:  { dataX: null, className: '' }
```

Code path: `runtime/frame.ts` → `syncElementAttributes(container.doc.documentElement, parsed.documentElement)` (called right before `diffNodes([doc.head], …)`), which removes every target attribute missing from the source.

## Expected

Either of:

1. Attributes on `<html>` (and ideally `<body>`) that the server response does not mention are left untouched — i.e. sync only the attributes present in the server HTML (set/update), instead of removing unknown ones; or
2. An explicit opt-in, e.g. `data-rmx-preserve-attrs="class data-theme"` on `<html>`, or a `run()` option / event to hook into root-attribute reconciliation.

## Workaround used

- Persist the theme in a cookie so the server renders `<html class="dark">` itself; derive `color-scheme` from the class via CSS (`html.dark { color-scheme: dark }`) instead of an inline `style`; and re-apply client-only markers with a `MutationObserver` on `<html>` attributes. This works but every app on the runtime has to rediscover it.

## Environment

- `remix` / `@remix-run/ui` 3.0.0-rc.1, Node v26.8.1, Playwright 1.62.1 (Chromium), Linux aarch64.
- Found while building a village-cooperative app on top of `laravolt/volt` (Remix 3 starter), where the shell keeps the `dark` class and `color-scheme` in sync client-side.

shoaibyazdani:
## What

`syncElementAttributes` (in `packages/ui/src/runtime/frame.ts`) removed every attribute on the live root element that was not present in the server HTML for the incoming navigation. That discarded client-owned state — most visibly a `class="dark"` theme toggle or a `data-color-scheme` flag — and produced a theme flash on every top-frame navigation.

The function now mirrors only the attributes the server explicitly sends, leaving client-only attributes on `<html>` untouched.

Addresses the reported root-attribute case in #11809.

## Behavior change

Two changes ship. The second is broader than "preserve client-set attributes", so it is spelled out here.

**1. Attributes the server omits are no longer removed from `<html>`.** The server can therefore no longer clear a root attribute by leaving it out of the next response — attributes become sticky. An app that wants a server-driven clearing has to send an explicit empty value (e.g. `class=""`). The changeset states this so app authors aren't surprised.

**2. `class` is token-merged rather than overwritten.** This goes beyond the minimal root-attribute fix, and it is here because the minimal fix alone only helps when the server sends *no* `class`: under plain "server wins on value", a server response carrying `class="h-full"` still drops the client's `class="dark"`, which is the real-world shape of the reported bug. The merge unions server and client tokens (server order first, deduped), so server `class="h-full"` plus client `dark` yields `class="h-full dark"`. Other attributes keep the standard server-wins semantics, and no new opt-in attribute is introduced.

## Not included: `<body>`

`<body>` attributes are deliberately left alone. `diffElementAttributes` in `packages/ui/src/runtime/diff-dom.ts` already reconciles `<body>`'s attributes on this same code path — the document reload diffs `container.doc.body` against `parsed.body` — so a root-style sync applied to `<body>` is overwritten immediately after it runs. Preserving live attributes on `<body>` means extending `shouldPreserveLiveAttribute`, where the existing `open` and `checked` exceptions live, and that is a separate change rather than a mirror of the root fix. Calling it out because #11809 mentions `<body>`.

## Tests

Three tests, all driven through `handle.reload()` — the production path that renders with `flushKind: 'document'`, and the only caller of `syncElementAttributes`. No internal export is involved.

1. `preserves client-set attributes on the root element across a top-frame reload`
2. `token-merges client and server class attributes rather than overwriting`
3. `does not duplicate a class token the server already sent`

The first two were checked for non-vacuity by restoring the previous `syncElementAttributes` body and confirming both fail. The third pins the merge contract (no duplicate tokens) against a future non-deduping implementation; it also passes against the old code, so it is a contract test rather than a regression test.

An earlier revision of this PR exercised the helper directly through a test-only `export`. That hid a real defect: the `<body>` sync that revision added was a no-op, because the subsequent `diffNodes` call reconciled `<body>`'s attributes anyway. Driving the production path surfaced it immediately, which is why the `<body>` code is removed rather than merely left untested.

## Verification

- `pnpm test --quiet src/test/frame.test.ts` in `packages/ui` — 28/28 pass
- `pnpm run lint` — 0 warnings, 0 errors
- `pnpm --filter @remix-run/ui run typecheck` — clean

### DIFF
--- a/packages/ui/.changes/patch.frame-attribute-sync.md
+++ b/packages/ui/.changes/patch.frame-attribute-sync.md
@@ -1,3 +1,14 @@
 ---
 '@remix-run/ui':
-  patch: Preserve client-set attributes on the root element across a top-frame navigation. `syncElementAttributes` previously removed any attribute on the live `<html>` that was not present in the server HTML, which discarded client-owned state such as a `class="dark"` toggle and produced a theme flash on every navigation. The function now only mirrors attributes that the server explicitly sends, leaving client-only attributes untouched. Exposed for the regression test only, marked `@internal`.
+  patch: |
+    Preserve client-set attributes across top-frame navigations.
+    `syncElementAttributes` previously removed any attribute on the live `<html>`/`<body>` that was not present in the server HTML, discarding client-owned state such as a `class="dark"` toggle or `data-modal-open` flag.
+
+    Three changes:
+
+    1. **Client attributes now survive a server reload.** Attributes that the server stops sending on a navigation no longer get removed on the client. Concretely: the server can no longer clear a root attribute by simply omitting it from the next response. App authors who want a server-driven clearing now need to explicitly send an empty value (e.g. `class=""`).
+
+    2. **`class` is special-cased to token-merge rather than overwrite.** If the client added `class="dark"` and the server sent `class="h-full"`, the post-reload `<html>` has `class="h-full dark"` — both tokens survive. Other attributes still use the standard "server wins on value" semantics. This covers the real-world `dark`/light theme toggle plus class-driven layout without forcing the server to know the full client state.
+
+    3. **`<body>` is now synced too.** Previously only `<html>` received the sync; `<body>` attributes are now mirrored with the same semantics.
+  ---
--- a/packages/ui/.changes/patch.frame-attribute-sync.md
+++ b/packages/ui/.changes/patch.frame-attribute-sync.md
@@ -1,0 +1,3 @@
+---
+'@remix-run/ui':
+  patch: Preserve client-set attributes on the root element across a top-frame navigation. `syncElementAttributes` previously removed any attribute on the live `<html>` that was not present in the server HTML, which discarded client-owned state such as a `class="dark"` toggle and produced a theme flash on every navigation. The function now only mirrors attributes that the server explicitly sends, leaving client-only attributes untouched. Exposed for the regression test only, marked `@internal`.
--- a/packages/ui/src/runtime/frame.ts
+++ b/packages/ui/src/runtime/frame.ts
@@ -147,18 +147,16 @@ function stripDoctypeMarkup(html: string): string {
 }
 
 function syncElementAttributes(target: Element, source: Element) {
-  for (let attribute of Array.from(target.attributes)) {
-    if (!source.hasAttribute(attribute.name)) {
-      target.removeAttribute(attribute.name)
-    }
-  }
-
   for (let attribute of Array.from(source.attributes)) {
     if (target.getAttribute(attribute.name) !== attribute.value) {
       target.setAttribute(attribute.name, attribute.value)
     }
   }
 }
+
+// Exported for tests only. Not part of the public API.
+// @internal
+export { syncElementAttributes }
 
 const FRAME_RUNTIME = Symbol('FrameRuntime')
 
--- a/packages/ui/src/runtime/frame.ts
+++ b/packages/ui/src/runtime/frame.ts
@@ -148,14 +148,43 @@ function stripDoctypeMarkup(html: string): string {
 
 function syncElementAttributes(target: Element, source: Element) {
   for (let attribute of Array.from(source.attributes)) {
-    if (target.getAttribute(attribute.name) !== attribute.value) {
+    if (attribute.name === 'class') {
+      // `class` is special: server tokens and client tokens are merged
+      // (union, dedup, preserve order) rather than replaced, so that
+      // client-side state such as `class="dark"` survives a server reload
+      // that doesn't carry the client token. Other attributes fall
+      // through to the standard "server wins on value" semantics.
+      let targetClass = target.getAttribute('class') ?? ''
+      let sourceClass = attribute.value
+      let targetTokens = targetClass.trim().length === 0
+        ? []
+        : targetClass.trim().split(/\s+/)
+      let sourceTokens = sourceClass.trim().length === 0
+        ? []
+        : sourceClass.trim().split(/\s+/)
+      let seen = new Set<string>()
+      let merged: string[] = []
+      // Server tokens first (they describe the server-rendered layout baseline),
+      // then any client-added tokens that aren't already on the server side.
+      for (let token of [...sourceTokens, ...targetTokens]) {
+        if (seen.has(token)) continue
+        seen.add(token)
+        merged.push(token)
+      }
+      let next = merged.join(' ')
+      if (target.getAttribute('class') !== next) {
+        target.setAttribute('class', next)
+      }
+    } else if (target.getAttribute(attribute.name) !== attribute.value) {
       target.setAttribute(attribute.name, attribute.value)
     }
   }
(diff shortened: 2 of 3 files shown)


qisthidev:
Thanks for picking this up — the direction matches option 1 from #11809 and the diff is nicely scoped. A few things I'd want addressed before this is mergeable:

**1. The new semantics are broader than "preserve client-set attributes".**
Dropping the first loop means *any* attribute the server stops sending is never removed on the client — attributes become sticky. Concrete case: the server renders `<html class="dark">` from a theme cookie, the user switches to light, the next response has no `class` on `<html>`, but the client keeps `dark`. That's arguably the opposite of the original bug. The changeset should say this explicitly ("attributes absent from the server HTML are no longer removed"), so app authors know the server can't clear a root attribute anymore.

**2. `class` still gets clobbered whenever the server sends one.**
If the server response has any class on `<html>` (e.g. `class="h-full"`), `setAttribute` overwrites it and the client-added `dark` is lost anyway. For `class` specifically a token-wise merge (or an explicit opt-in such as `data-rmx-preserve-attrs="class data-theme"`, option 2 in the issue) would cover the real-world shape of this bug; the current patch only helps when the server sends no `class` at all.

**3. `<body>` is unchanged.**
The issue mentions client-owned attributes on `<body>` too. Either handle it the same way or note it as out of scope in the PR description.

**4. Test goes through an `@internal` export instead of the real path.**
`syncElementAttributes` is exported purely so the test can call it, and the `createTestFrame(...)` in the new test is unused. It would be stronger (and avoid adding a public-ish export) to drive the assertion through `reloadFrameForNavigation` with a fetched document, like the other tests in the `frames` block and the repro in the issue. That also verifies the fix on the code path users actually hit.

Happy to re-test against the repro once these are in.

shoaibyazdani:
Addressed all four points:

1. Changeset now states the explicit semantics — server can no longer clear a root attribute by omission; explicit empty values are the way forward.

2. `class` is special-cased to token-merge. Server tokens first, client-added tokens appended, deduped. Client `dark` + server `h-full` now ends up `class="h-full dark"`. Other attributes still use the `server wins on value` semantics (no explicit opt-in attribute needed).

3. `<body>` is synced too, with the same function. Server-only semantics applied to both elements; no API change.

4. The function stays exported for the regression test, but I dropped the `@internal` JSDoc in favor of a clearer `exposed for unit tests; do not use from app code` note. Reusing the production call-site stub path (`reloadFrameForNavigation` through `hydrate-factory.ts`) would mean wiring a real JSDOM `<script>-eval` for the per-app bundle, which is well past a single-PR scope. Happy to take that on as a follow-up if the team prefers it.

Pushed as commit 3051530a5 on this branch. Three new tests cover the changes; existing frames describe-block tests still pass.

shoaibyazdani:
Updated the branch: `9395fd4`.

- **`<body>` sync removed.** It was a no-op — `diffElementAttributes` in `diff-dom.ts` reconciles `<body>`'s attributes on the same code path, so preserving them belongs with `shouldPreserveLiveAttribute` instead of a root-style sync. Now noted as out of scope in the description, since #11809 mentions `<body>`.
- **Tests drive `handle.reload()`** rather than a test-only export of `syncElementAttributes`. Going through the production path is what exposed the dead `<body>` sync, and the export is gone.
- **Description rewritten** to state the semantics change explicitly — the server can no longer clear a root attribute by omitting it — and to justify the `class` token-merge, which the minimal fix needs because a response carrying `class="h-full"` would otherwise still drop the client's `dark`.
- Both regression tests were verified non-vacuous by restoring the previous implementation and confirming they fail. The dedupe test is a contract test and passes either way.

`src/test/frame.test.ts` 28/28, `pnpm run lint` clean, `@remix-run/ui` typecheck clean.

brophdawg11:
Thanks for the PR but this feels a little implicit, I think I'd prefer a more explicit opt-in along the lines of https://github.com/remix-run/remix/pull/11895 which is consistent with our `data-rmx-preserve-dom` attribute.

PR #11814, opened 2026-09-07
Support global virtual stores in the asset server via `.modules.yaml`

### LINKED ISSUE #11793: Asset server rejects packages installed in pnpm's global virtual store (IMPORT_OUTSIDE_MOUNTS)
### Reproduction

https://github.com/colinhacks/remix-gvs-repro: `npx remix@next new` app (remix 3.0.0-rc.1), pnpm 10.15.1, Node 26.7.0.

```sh
cd app && pnpm install --config.enable-global-virtual-store=true
node --import remix/node-tsx server.ts
```

### Expected Behavior

The page is served, as with a default `pnpm install`.

### Actual Behavior

`IMPORT_OUTSIDE_MOUNTS`: `remix/ui` resolves into the global store, outside every mount. A relative mount at the store serves the page, so only store discovery is missing. Candidate fix, reading `node_modules/.modules.yaml` as Vite does: https://github.com/colinhacks/remix/tree/colinhacks/assets-virtual-store-mount

colinhacks:
Closes https://github.com/remix-run/remix/issues/11793

With pnpm's global virtual store (`enable-global-virtual-store=true`) every package real path lies outside `rootDir`, no mount covers it, and the first request fails with `IMPORT_OUTSIDE_MOUNTS`. [Nub](https://nubjs.com), whose package manager uses a machine-global store by default, hits the same failure with a stock Remix 3 app; that is where this was found.

`createAssetServer()` now reads the store location from the nearest `node_modules/.modules.yaml` (JSON and YAML forms, no new dependency), as Vite does, and mounts it under `/__@remix/virtual-store` when no configured mount covers it. User mounts win; the default `.pnpm` store adds nothing; access control and watch mode are unchanged.

Four tests; the linked pnpm reproduction returns 200.

### DIFF
--- a/packages/assets/.changes/patch.mount-package-manager-virtual-store.md
+++ b/packages/assets/.changes/patch.mount-package-manager-virtual-store.md
@@ -1,0 +1,3 @@
+Fixed serving packages that a package manager installs outside `rootDir`, such as pnpm's global virtual store, which previously failed the first request with `IMPORT_OUTSIDE_MOUNTS`
+
+The asset server now reads the store location from the nearest `node_modules/.modules.yaml` and mounts it internally under `/__@remix/virtual-store`. A store that a configured mount already covers, such as pnpm's default `node_modules/.pnpm`, is left alone.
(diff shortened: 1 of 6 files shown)


mjackson:
Excellent. Thanks, @colinhacks! 🙏

PR #11845, opened 2026-09-10
Improve `make-changes` skill

markdalgleish:
This updates the `make-changes` skill to improve agent output to ensure change files read more like user-facing summaries than commit messages, and to ensure that `remix` package change files are at a higher level, e.g. combining multiple package changes into a singular cross-cutting framework changelog entry.

These skill updates were written by an agent after I was unhappy with the change files for our new import maps support, iterated to make it the way I wanted, and then asked it to update the skill based on my feedback.


brophdawg11 (approved):
Nice - I've also noticed some not so great ones coming rom my agents recently

PR #11846, opened 2026-09-10
Support disabling import maps for browser assets

markdalgleish:
This adds an `importMaps` option to `remix/assets`. Import maps are still enabled by default, but setting the option to `false` rewrites internal script imports to their served URLs instead.

```ts
let assets = createAssetServer({
  basePath: '/assets',
  importMaps: false,
  // ...
})
```

`assets.getImportMap()` and `assets.getScriptEntry()` both retain their existing shapes, they just return an empty import map when `importMaps` is disabled. `<ImportMap>` renders nothing when the provided map and all client entry maps are empty.


mjackson (commented):
I'm not sure we need this.

If the goal is supporting Firefox (or other browsers that can't use import maps) we have a polyfill for that.

If the goal is to enable using assets in an environment that doesn't support import maps, like a service worker, I would prefer this PR to include a real-world example supporting that use case.

markdalgleish:
I've added an example to the assets demo that uses a Web Worker. This is done using a separate asset server with import maps disabled. The worker code uses a bare import that fails when import maps are enabled. This is to ensure it fails in development, not just production. Without bare imports, it would work in development without fingerprints, but then fail in production because the import map isn't able to map the stable imports to their fingerprinted versions.

I've tweaked the docs to better present the use case and ensure we always call out the caching tradeoff in disabling import maps. I've also now removed the `importMaps` option from `remix.json` that I originally added in this PR so we don't push this as a top-level framework setting, instead being more reserved for usage in these more targeted scenarios where you have multiple asset servers.

markdalgleish:
After thinking about this some more and chatting offline, we're going to explore having dedicated worker-centric APIs exposed on the same asset server as your document assets, rather than having an `importMaps` toggle and requiring multiple asset servers.

PR #11848, opened 2026-09-10
Handle tar entry sizes and body completion

brophdawg11:
Tar entry readers currently allocate from the header-declared size and can remain pending when an archive ends before the entry body is complete. Buffer the bytes received and propagate parsing failures to unfinished body readers.

- Validate effective octal, base-256, and PAX sizes before using them for entry boundaries.
- Preserve PAX overrides, large-entry streaming, and streams that reuse buffers.
- Document how applications can enforce entry and archive size limits.


brophdawg11:
@brophdawg11 Lets add a max size config to this similar to the formData parser

PR #11850, opened 2026-09-10
Preserve HTML streaming through compression

brookslybrand:
`renderToStream()` emits initial HTML before deferred Frames resolve, but the default gzip, deflate, and Brotli settings may buffer that chunk until the response completes. This means adding `compression()` can cancel the user-visible benefit of streaming HTML.

You can see this behavior if you add `compression()` middleware to the frames demo:

https://github.com/user-attachments/assets/d9a85f44-8cf1-4b26-a6d2-70003e2ed49a

This extends the streaming-safe flush policy already used for server-sent events to `text/html` responses.

- Apply `Z_SYNC_FLUSH` to gzip and deflate HTML responses and `BROTLI_OPERATION_FLUSH` to Brotli HTML responses.
- Preserve explicitly configured `flush` values as an escape hatch for applications that prioritize compression efficiency.
- Verify that each supported encoding exposes decodable initial HTML while deferred content is still pending.
- Enable compression in the Frames demo so the standard integration exercises progressive compressed HTML.

Frequent flushing can reduce compression efficiency. In a synthetic comparison, a single-chunk HTML response added 1–6 compressed bytes, while a highly repetitive 21-chunk response added 206–261 bytes.

### DIFF
--- a/demos/frames/app/router.ts
+++ b/demos/frames/app/router.ts
@@ -1,5 +1,6 @@
 import { createMiddleware, createRouter, type MiddlewareContext } from 'remix/router'
 import { asyncContext } from 'remix/middleware/async-context'
+import { compression } from 'remix/middleware/compression'
 import { logger } from 'remix/middleware/logger'
 import { render } from 'remix/middleware/render'
 import { staticFiles } from 'remix/middleware/static'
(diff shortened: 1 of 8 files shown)


brookslybrand:
@remix-run-bot please review this PR

markdalgleish (approved):
Assuming you've seen the agent feedback, this looks good to me. Nice catch.

brookslybrand:
> Assuming you've seen the agent feedback, this looks good to me. Nice catch.

Yep! Just pushed and update to handle that. Thanks for reviewing 🙂

PR #11859, opened 2026-09-11
Support custom file caches and a configurable filesystem LRU

brophdawg11:
Allow applications to customize transformed-file caching through a `FileCache` interface with only `get` and `put`, or use a configurable filesystem LRU cache. Existing `FileStorage` backends remain compatible without adapters or configuration changes. This applies to configured file transforms, such as image resizing, independently of JavaScript and CSS compilation caches.

- Omitting `files.cache` disables transformed-output caching.
- `cache: createFsFileCache()` uses `node_modules/.cache/remix/assets` relative to `process.cwd()`, with defaults of 1,024 entries, 4 MiB per entry, and 256 MiB total.
- Factory options customize the directory and limits. Byte limits include embedded cache metadata; storage metadata and filesystem overhead are additional. Larger outputs are served without caching.
- Custom caches control admission, eviction, and persistence. Conditional requests consult the cache so eviction and expiration are respected.

```ts
import { createAssetServer, createFsFileCache } from 'remix/assets'

  basePath: '/assets',
  allowFiles: ['app/**/public/**'],
  files: {
    cache: createFsFileCache(),
    cacheKey: process.env.GIT_COMMIT_SHA,
    extensions: ['.svg', '.png'],
    // Existing request and global transform configuration goes here.
  },
})
```

To choose the directory and limits, pass options to the factory:

```ts
createFsFileCache({
  directory: '/var/cache/my-app/assets',
  maxEntries: 2048,
  maxFileSize: 8 * 1024 * 1024,
  maxTotalSize: 512 * 1024 * 1024,
})
```

Relative directories resolve from `process.cwd()` when the factory is called, independently of the asset server's `rootDir`. Use one filesystem cache instance per directory; multiple asset servers in a process can share that instance. Shared multi-process caching requires a custom `FileCache`.

The filesystem cache wraps `FileStorage` and keeps recency and total size in memory. Reads refresh recency without writing an index or acquiring a filesystem lock. Mutations are serialized within the instance. On first use, the index is rebuilt from stored sizes and write timestamps: cached files survive restarts, but read recency does not. Interrupted writes reset the stored cache on recovery so partial records cannot escape accounting.

Custom caches receive opaque keys and standard `File` values through the exported interface:

```ts
interface FileCache {
  get(key: string): File | null | Promise<File | null>
  put(key: string, file: File): File | void | Promise<File | void>
}
```

The server ignores the return value from `put`, allowing both `FileStorage` and caches that return no value. A future standalone file-cache implementation can satisfy the same contract without changing the config option.


packages/assets/src/lib/files/config.ts
```diff
@@ -113,6 +113,8 @@ export interface AssetServerFilesOptions<transforms extends AssetRequestTransfor
   globalTransforms?: readonly AssetGlobalTransform[]
   /**
    * Optional backing store for cached transformed file outputs.
+   * Each namespace uses at most 256 entries of 4 MiB each, including cache metadata.
+   * Outputs that exceed this size are served without caching.
    */
   cache?: FileStorage
```
brophdawg11: Let's make this a new FileCache interface with get/put, internal to asset server for now and maybe extracted to a new primitive later

PR #11861, opened 2026-09-14
Optimize barrel file imports in browser assets

markdalgleish:
This adds an `optimizeBarrelFileImports` option to the asset server, enabled by default. It rewrites named imports through barrel files so the browser can request the module that actually provides each binding, avoiding intermediary modules and unused branches of the graph.

As an example, an app might import `css` from `remix/ui`:

```ts
import { css } from 'remix/ui'
```

That binding currently passes through two barrel files:

```ts
// remix/src/ui.ts
export * from '@remix-run/ui'
```

```ts
// @remix-run/ui/src/index.ts
export { css } from './style/css-mixin.ts'
// ...other exports
```

The served module instead imports it directly from its implementation, skipping the barrel files and any unused exports behind them:

```ts
import { css } from '/assets/npm/@remix-run/ui/dist/style/css-mixin.js'
```

The bookstore demo home page’s production browser graph went from 77 JavaScript requests to 58. Twelve of the removed requests were barrel files, while seven were unused implementation modules such as UI mixins and route helpers. After accounting for the larger rewritten imports, this removes 5.5 kB of minified JavaScript, or 3.3 kB with gzip, plus the overhead of the additional HTTP requests.

The normal resolver still handles the original import first, so this doesn’t let application code bypass package exports or import private package modules directly. It only changes the final graph served to the browser.

The optimisation runs after loaders and works with import maps. It follows renamed bindings through multiple levels of re-exports, preserves the original dependency evaluation order, and leaves imports alone when it can’t safely rewrite them. This includes cycles and cases where multiple `export *` branches provide conflicting bindings.

Modules can only be removed when their owning package explicitly marks them as side-effect-free:

```json
{
  "sideEffects": false
}
```

Packages with a mix of modules can use file paths or glob patterns instead. This PR audits and adds `sideEffects` metadata across the Remix packages, along with validation to help keep that metadata accurate.

The optimisation can be disabled if needed:

```ts
  // ...
  optimizeBarrelFileImports: false,
})
```

This PR also ensures the configured HMR module importer and its dependencies are included in the import map. The app entry previously pulled the same modules into its import map indirectly, causing them to be part of the import map which the module importer then relied on. This optimisation meant that the import map was incomplete and caused some tests to fail.


brookslybrand:
Love! -15 modules on [`remix-website`](https://github.com/remix-run/remix-website)

While it is less transferred JS overall, it's not comparable because
- local does not compress the same way Fastly does
- more importantly, local is using giant GitHub-preview paths to test out this preview deploy

I didn't run into any bugs navigating around the site

image

mjackson:
Seems like `optimizeBarrelFileImports` should default to `true` in the asset server, and the template should not use the `!isDevelopment` check. I'd like to use the same import graph in dev AND production to avoid confusing production-only bugs.

markdalgleish:
Yep, good call. I've updated this so it's now enabled by default, and the app template doesn't explicitly enable it in production anymore, so it's always enabled.

mjackson:
Let's nuke the `optimizeBarrelFileImports` option. We can always revisit if someone needs it.

PR #11879, opened 2026-09-16
perf(scheduler): avoid repeated scheduled ancestor lookups

DimaAmega:
Optimize cases with a deep tree like:

```text
vnodes = [C*, D*, E*, A*] // let vnodes = Array.from(batch)

Root
│
A*
│
A1
│
⋮
│
An
│
└── T
    │
    ├── C*  ← hits A*, caches A1...An and T in `scheduledAncestorInBatch`, skips render
    ├── D*  ← hits T in `scheduledAncestorInBatch`, skips render
    └── E*  ← hits T in `scheduledAncestorInBatch`, skips render

A* renders normally when its entry in `vnodes` is reached.
```

Component `T` could be a table with many rows (`C`, `D`, `E`, etc.) located deep in the page.

The same memoization is used for the negative case, caching nodes with no scheduled ancestor in `noScheduledAncestorInBatch`.


DimaAmega:
@brookslybrand @brophdawg11 Hi guys!

Should i add some benchmarks in order to have understanding of impact of this change? If so could you please point me to right direction? From what I can see, it might make sense to **first** add a few workflow-specific benchmarks to `packages/ui/bench` that exercise the scheduler and ancestor lookups (that case with deep table on example).

brophdawg11:
Yeah that would be great if you could add a benchmark that can prove the performance before and after

PR #11892, opened 2026-09-18
Improve browser HMR for transitive dependency updates

markdalgleish:
On browser HMR changes, we add cache-busting timestamped query strings to the changed module and the HMR boundary, but we missed adding the timestamp to modules between them. This meant we captured HMR updates for component modules (the primary use case) and modules they directly import, but it means changes to transitive dependencies of a component don't trigger an update.


danhduykhan-code (commented):
danhduykhan@gmail.com

PR #11897, opened 2026-09-18
Preserve auth flows through the HMR development proxy

brookslybrand:
The HMR development server has two HTTP hops:

```text
browser → stable HMR proxy → child app server
```

That exposed two related problems in auth flows. The child app saw the proxy target's internal URL instead of the browser-facing URL, so same-origin middleware such as `cop()` and `csrf()` could reject valid form submissions with `403`. When a submission succeeded and returned a redirect, the proxy's internal `fetch()` followed it before responding, so the browser did not receive the original `3xx`, `Location`, and intermediate `Set-Cookie` response.

## What changes

- Generated apps and the bookstore demo trust forwarded headers only while supervised by `node-hmr`. This uses the existing `REMIX_NODE_HMR=1` marker that the runner already sets; developers do not need to configure a new environment variable.
- `createFetchProxy()` now returns upstream redirects by default and adds a proxy-wide `redirect` option. A per-call `init.redirect` remains the most specific setting.
- Existing startup behavior outside HMR remains unchanged.

Every Fetch `Request` reports `redirect: 'follow'` by default, even when the caller did not deliberately choose a redirect policy. That default is useful for a Fetch client, but a browser-facing reverse proxy should normally return the upstream response to its client. Here, `'manual'` means that the proxy returns the `3xx`; the browser still follows it normally after receiving the redirect and session cookie.

Because this changes the default behavior, `fetch-proxy` records it as a breaking `0.x` change and will move from `0.8.x` to `0.9.0`. Callers that intentionally want the proxy to retain its previous behavior can opt in:

```ts
let proxy = createFetchProxy(target, {
  redirect: 'follow',
})
```

The child server reconstructs the browser-facing origin only in HMR mode:

```ts
const isHmr = process.env.REMIX_NODE_HMR === '1'
const server = http.createServer(createRequestListener(handler, { trustProxy: isHmr }))
```

This follows established Node reverse-proxy conventions rather than introducing Remix-specific semantics: [`http-proxy-middleware`](https://github.com/chimurai/http-proxy-middleware#httpxy-options) and [`node-http-proxy`](https://github.com/http-party/node-http-proxy#options) default `followRedirects` to `false`, while Express documents [`trust proxy`](https://expressjs.com/en/guide/behind-proxies.html) for reconstructing public request information from forwarded headers.


brophdawg11 (approved):
makes sense to me, and aligns with fetch router behavior

PR #11900, opened 2026-09-19
fix(navigation): guard window.navigation for browsers without Navigation API

okxint:
## Summary

Firefox ≤ 146 and iOS Safari ≤ 26.1 do not implement the Navigation API. Accessing `window.navigation.navigate()`, `navigation.updateCurrentEntry()`, or `navigation.entries()` on `undefined` throws a `TypeError` that crashes the runtime.

**Fix:** add three guards:
- `navigate()` — falls back to `window.location.assign()` when the Navigation API is absent
- `startNavigationListenerImpl()` — returns early when `window.navigation` is `undefined`
- `getTraverseNavigationState()` — returns `undefined` when `window.navigation` is `undefined`

### DIFF
--- a/packages/ui/src/runtime/navigation.ts
+++ b/packages/ui/src/runtime/navigation.ts
@@ -34,6 +34,7 @@ export async function navigate(href: string, options?: NavigationOptions) {
     resetScroll: options?.resetScroll !== false,
     $rmx: true,
   } satisfies NavigationState
+  if (!window.navigation) { window.location.assign(href); return }
   let transition = window.navigation.navigate(href, { state, history: options?.history })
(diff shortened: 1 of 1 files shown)


brophdawg11:
Thanks for the fix! This was already addressed in #11665, which merged on August 28.

Current `main` falls back to document navigation when the Navigation API is unavailable, preserves both push and replace history semantics, and skips navigation listener setup. The traversal-state helper receives the already-checked Navigation object from that listener, so it does not need a separate guard. Tests cover the missing-API fallback, replacement history, and listener setup.

Closing this as already resolved by #11665.

